PR #31146, opened 2020-10-18
[4.0] Cleanup jstools

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Move the build.js file into the build folder
- Move the . eslintrc file into the build folder
- Move the .stylelintrc.json file into the build folder
- Remove redundant  drone-package.json (this might be wrong, let's see if things break bad)
- Fix couple of styling issues in Media Manager (Oops, blame said that I was responsible)
- Make sure that Drone is running the same commands for linting js and scss

### Testing Instructions
If Drone is ok that will be all the tests needed here (there is no real functionality changed here except changing some paths)

Eslint and Stylelint are using different commands in the dev CLI and Drone. Although things seem to be green in reality they're not

Everything is Green


dgrammatiko:
@richard67 is there some special ritual for the drone.yml ?
Screenshot 2020-10-18 at 18 32 06

HLeithner:
> @richard67 is there some special ritual for the drone.yml ?
> Screenshot 2020-10-18 at 18 32 06

Yes if you modify the drone.yml we have to sign it, I started it by hand.

about the *.css linter thing, we have some .es files (introduced by me...)

dgrammatiko:
> about the *.css linter thing, we have some .es files (introduced by me...)

I'm not changing any of the logic here, just moving the configuration files from the root folder to the build folder and making sure that both the CLI and Drone are executing the same commands

HLeithner:
Ok then I missunderstood the changes... I signed your done.yml.

Thanks

Issue #31165, opened 2020-10-20
[4.0] Revert PR #29571


N6REJ:
PR #29571 breaks admin login.

### Steps to reproduce the issue
log into admin on working install
pull latest commits from 4.0-dev
try logging into admin
notice nothing happens.

### Expected result

can login admin

### Actual result
login button does nothing.

### System information (as much as possible)

4.0-dev 19oc20 1900cst pull

### Additional comments
reverting commit allows logging in again
ran both composer install & npm ci but didn't fix the issue.

SharkyKZ:
I can't reproduce the issue.

richard67:
@SharkyKZ On a new install of current 4.0-dev I can't reproduce either.

But as @N6REJ wrote, he has pulled the changes on an existing installation and then had the issue.

So maybe we will have an issue when not making a new installation but updating?

=> It needs to test if a J4 Beta 4 can be updated to latest nightly J4 build (which includes #29571 ) to see if there is an issue when not having made a new installation.

In addition it should also be tested in a 3.10 can be properly updated to latest nightly J4 build without running into the issue.

I'll try to find time later for testing that, but if someone else could test that, too, it would be safer because I might make mistakes.

SharkyKZ:
My guess you need to clear browser cache. If you're using some weird browser, https://github.com/joomla/joomla-cms/pull/31169 might help.

brianteeman:
Can not confirm

N6REJ:
idk what happened.. its working fine now.

PR #31169, opened 2020-10-20
[4.0] Add type attribute to button

SharkyKZ:
### Summary of Changes

Adds missing `type` attribute to backend login button.

### Testing Instructions

Code review or test that you can still login to backend.

No.

### DIFF
--- a/administrator/modules/mod_login/tmpl/default.php
+++ b/administrator/modules/mod_login/tmpl/default.php
@@ -122,8 +122,7 @@ Text::script('JHIDEPASSWORD');
 		</div>
 		<?php endforeach; ?>
 		<div class="form-group">
-			<button class="btn btn-primary btn-block btn-lg mt-4"
-				id="btn-login-submit"><?php echo Text::_('JLOGIN'); ?></button>
+			<button type="submit" id="btn-login-submit" class="btn btn-primary btn-block btn-lg mt-4"><?php echo Text::_('JLOGIN'); ?></button>
 		</div>
 		<input type="hidden" name="option" value="com_login">
 		<input type="hidden" name="task" value="login">


richard67:
I have tested this item :white_check_mark: successfully on 3b7919382d435c953ff85da7c77985ec564ecfdd<br /><br />Code review: The PR adds the right type attribute to the button. See [https://www.w3schools.com/tags/att_button_type.asp](https://www.w3schools.com/tags/att_button_type.asp "https://www.w3schools.com/tags/att_button_type.asp"):

> Tip: Always specify the type attribute for the <button> element. Different browsers may use different default types for the <button> element.

The "submit" type is the right one.

Real test: As I did not have problems with login like they were reported in issue #31165 , I could only check that login still works as well as before and that the type attribute really has been added.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31169">issues.joomla.org/tracker/joomla-cms/31169</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on 3b7919382d435c953ff85da7c77985ec564ecfdd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31169">issues.joomla.org/tracker/joomla-cms/31169</a>.</sub>

chmst:
I had the problems as reported in  #31165 but not always, and cannot say when it occurred. So I can confirm that this is the right attribute and the problem does not occur with this change.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31169">issues.joomla.org/tracker/joomla-cms/31169</a>.</sub>

brianteeman:
This should not be required as any button inside a form has the default value of "type=submit"

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

richard67:
> This should not be required as any button inside a form has the default value of "type=submit"
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

@brianteeman But does it do harm to specify the type attribute?

brianteeman:
No - but my point is that if everyone is saying this is an "intermittent" problem then I doubt very much that adding the default value for the button is actually resolving the problem

richard67:
> No - but my point is that if everyone is saying this is an "intermittent" problem then I doubt very much that adding the default value for the button is actually resolving the problem

Agree.

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31169">issues.joomla.org/tracker/joomla-cms/31169</a>.</sub>

infograf768:
tks

PR #31182, opened 2020-10-21
Update checksum not validated when in different character case

innato:
If update server's XML has checksum in capitals, it will be marked invalid because the Joomla!-generated $hashPackage will be lower case.
With the proposed change (line 398), the process will be more foregiving. I used the MD5 & SHA Checksum Utility from Raymond Lin (https://raylin.wordpress.com) to generate the checksum in the update server's XML file. This utility generates an upper case checksum and is therefore incompatible with the Joomla! process.

Pull Request for Issue https://github.com/joomla/joomla-cms/issues/31178 .

### DIFF
--- a/libraries/src/Installer/InstallerHelper.php
+++ b/libraries/src/Installer/InstallerHelper.php
@@ -395,7 +395,7 @@ abstract class InstallerHelper
 				$hashRemote  = $updateObject->$hash->_data;
 				$hashOnFile  = true;
 
-				if ($hashPackage !== $hashRemote)
+				if (strtolower($hashPackage) !== strtolower($hashRemote))	
 				{
 					return self::HASH_NOT_VALIDATED;
 				}
--- a/libraries/src/Installer/InstallerHelper.php
+++ b/libraries/src/Installer/InstallerHelper.php
@@ -395,7 +395,7 @@ abstract class InstallerHelper
 				$hashRemote  = $updateObject->$hash->_data;
 				$hashOnFile  = true;
 
-				if (strtolower($hashPackage) !== strtolower($hashRemote))	
+				if ($hashPackage !== strtolower($hashRemote))	
 				{
 					return self::HASH_NOT_VALIDATED;
 				}


zero-24 (approved):

jiweigert:
I'm not happy with that solution. Your PR weaken the checksum security.
Uppercase is different to lowercase. Period. 

It's the update-server maintainers job to met the requirements of the Manifest's XML Checksum.
If it don't match, it don't match and rejecting it is the correct way.


 <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31182">issues.joomla.org/tracker/joomla-cms/31182</a>.</sub>

zero-24:
From what i found the resulting hash is case insensitive: https://stackoverflow.com/questions/11892053/is-sha-256-case-insensitive
https://stackoverflow.com/questions/59871996/is-the-file-hashing-checksum-value-case-insensitive

So why should that weeken the security?

jiweigert:
>
>
> From what i found the resulting hash is case insensitive:
> https://stackoverflow.com/questions/11892053/is-sha-256-case-insensitive
>
> https://stackoverflow.com/questions/59871996/is-the-file-hashing-checksum-value-case-insensitive
>
> So why should that weeken the security?
>
Simply because you exclude All uppercases from being used inside a
checksum, in case the characters match, but not the sizing.

Having Ae56GHffFF678 should treated be differently to aE56gHFFff678 easing
out by comparing them all lower (or upper-) case should not be done, when
the Joomla's XML requirements  stating, that the checksum should be
lowercase (which is still wrong but at least a requirement)

And the best "answer" (from Stackoverflow, seriously?) stating "the algorithm is case-sensitive".

And so should be the input, everything else is weaking security.

alikon:
on every paper i can have memory, on the lecterature, the exadecimal letters are always expressed with uppercase
just as a silly example ascii to ebcdic conversion table always use upper case ....
it could be for historic reason or for whatever else reason, but they are allways  expressed with uppercase
so i don't like to change this  "convention"  for the sake of change

zero-24:
> >
> >
> > From what i found the resulting hash is case insensitive:
> > https://stackoverflow.com/questions/11892053/is-sha-256-case-insensitive
> >
> > https://stackoverflow.com/questions/59871996/is-the-file-hashing-checksum-value-case-insensitive
> >
> > So why should that weeken the security?
> >
> Simply because you exclude All uppercases from being used inside a
> checksum, in case the characters match, but not the sizing.
> 
> Having Ae56GHffFF678 should treated be differently to aE56gHFFff678 easing
> out by comparing them all lower (or upper-) case should not be done, when
> the Joomla's XML requirements  stating, that the checksum should be
> lowercase (which is still wrong but at least a requirement)
> 
> And the best "answer" (from Stackoverflow, seriously?) stating "the algorithm is case-sensitive".
> 
> And so should be the input, everything else is weaking security.

Just to get it right the Algorithmen is case sensitive. And nothing else would make sense but the resulting string or better the hash sum, is a hex value and that is case insensitive.

> Hashes and checksums are often presented in hexadecimal notation. Although it is common to use upper case A-F instead of lower case a-f, it does not make any difference.

https://stackoverflow.com/questions/59871996/is-the-file-hashing-checksum-value-case-insensitive

PHP will always return the hash hex value in lowercase as documented here:
https://www.php.net/manual/en/function.hash.php

> raw_output
> When set to TRUE, outputs raw binary data. FALSE outputs lowercase hexits. 

(Where false is the default value as we usually just compare the hex value and not the binary data)

zero-24:
> on every paper i can have memory, on the lecterature, the exadecimal letters are always expressed with uppercase
> just as a silly example ascii to ebcdic conversion table always use upper case ....
> it could be for historic reason or for whatever else reason, but they are allways  expressed with uppercase
> so i don't like to change this  "convention"  for the sake of change
> 

Well as mention above php is defaulting to lowercase for the resulting hash values. I personally dont care which case we use i more like the lowercase as i think that easier to read but in this case its not displayed anyway.

With the change proposed here you can use any style that you prefere in the update.xml's right now you can just use the lowercase version that php provides.

jiweigert:
>
>
> With the change proposed here you can use any style that you prefere in
> the update.xml's right now you can just use the lowercase version that php
> provides.
>
I still don't see why Joomla should accept both.

If update server deliver the checksum in uppercase, that's a problem of
update server maintainer and can be easy fixed by send the checksum in
lowercase.

That should not be the job of Joomla.

When Joomla switches to case-sensitive Checksums of more secure algorithms
or other procedures (and that is just a matter of time, having the growing
power of Cloud capacity in mind..) it's another thing to be have to changed
back.

There is no real reason to accept this change, just for the lazyness not to
met the requirement of lowercase the checksum in the first place.

The question is:
It is already documented, that the checksum has to be in lowercase?

If yes, reject this pr.
If no, document the requirement and reject or accept this pr.

>

zero-24:
My point is that a checksum is a hex value which is case insensitive by definition, right? So why should we thread them case sensitive?

jiweigert:
Out from docs.joomla.org, the example of the manifest.xml shows the
checksums all in lower case.

The checksums are optional, but it's documented, that Joomla will reject
them, if they don't match.

```

   - *sha256*, *sha384*, *sha512* – Optional. Since Joomla 3.9.0 you can
   add file checksums in these hash formats. Be aware that in Joomla 3 only a
   notice is shown on *updates* if a checksum is not correct. That's all.
   The update doesn't stop then. In Joomla 4 updates and installations will
   stop if a provided checksum doesn't match.

```
I interpret that, that it is intended, that Joomla reject checksums in
uppercase as designed and documented.

Documented by example and behaviour description if they don't match.

Again, in my opinion its lowering the security, when Joomla would accept
this PR.
What would come next?
Oh, your submitted password don't match?
Let's lowercase what you entered and let's try again?
There are so many open systems out there because of those security design
decisions, because system developers decided to make it easier to hack in.
And on security every little step counts.
I agree, that there is no difference on Hex-level, but why to make it
easier?

The behaviour is described.

Tobias Zulauf <notifications@github.com> schrieb am Do., 22. Okt. 2020,
01:26:

jiweigert:
Additionally, when checksums are based on the hex-system, why don't Joomla
do not compare hex-wise, but instead implement a string comparison?

Edit: Aaah... the "magic type conversion" of PHP ... 

I still have the impression, that the developer who wrote the check, intentionally used the case-sensitive check..

libraries/src/Installer/InstallerHelper.php
```diff
@@ -395,7 +395,7 @@ public static function isChecksumValid($packagefile, $updateObject)
 				$hashRemote  = $updateObject->$hash->_data;
 				$hashOnFile  = true;
 
-				if ($hashPackage !== $hashRemote)
+				if (strtolower($hashPackage) !== strtolower($hashRemote))
```
HLeithner: The hash_file function returns only lowercase characters so we need to convert them.
```suggestion
				if ($hashPackage !== strtolower($hashRemote))	
```

HLeithner:
We compare hashes and we know our hashes are always in lowercase so also a attacker know this so an attacker wouldn't use uppercase hashes but on the other hand a ci that creates a hash automatically maybe better creating hashes in uppercase because used later somewhere else.

Anyway it's a silly discussion to compare a binary string that was converted to hex in upper or lower case if only one programmer doesn't spent one hour to find out the he/she needs to supply the string in lowercase instead of camelCase or what ever.

so please @innato update the pr and we can merge it if we get 2 tests.

Thanks

rdeutz:
I agree with @HLeithner let's move on

jiweigert:
Well, I still not agree.

It's clearly visible in the example for an update-server manifest.xml on docs.joomla.org as lowercase, 
it's documented that Joomla in v4 will stop installation when comparision fail,
PHP own function produce as default lowercase,

so why change it?

Yes, an attacker who read the docs or analyse a manifest.xml will use lower case, all others who are not that smart, may fail because they try uppercase checksums.

So by allowing case-insensitive checksums, you give away 50% of security for every script-kid.
I don't think that's a silly discussion at all. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31182">issues.joomla.org/tracker/joomla-cms/31182</a>.</sub>

bembelimen:
I have tested this item :white_check_mark: successfully on 4059d6a7b4a509530596dea7915ffe02f32bbc94<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31182">issues.joomla.org/tracker/joomla-cms/31182</a>.</sub>

SniperSister:
> So by allowing case-insensitive checksums, you give away 50% of security for every script-kid.

That assumption is just wrong: there security in this concept is created by the actual hashes, not by the question if the strings are compared case sensitive or case insensitive. The current case-insensitive compare is just wasting peoples time trying to debug issues related to it.

jiweigert:
No developer would waste time if he take care of the documented example and also that Jommla will fail otherwise.
And that assumption is not wrong,

Either you compare checksum value AND case
OR
you just compare the value only.

Simple logic.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31182">issues.joomla.org/tracker/joomla-cms/31182</a>.</sub>

bembelimen:
I've added the issue link into the post to get the issue.

richard67:
I have tested this item :white_check_mark: successfully on 4059d6a7b4a509530596dea7915ffe02f32bbc94<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31182">issues.joomla.org/tracker/joomla-cms/31182</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31182">issues.joomla.org/tracker/joomla-cms/31182</a>.</sub>

HLeithner:
Thanks @innato for your first time contribution!

zero-24:
Thanks @innato :+1:

innato:
@HLeithner @zero24 Thanks for your kind remarks. It was an interesting first-time experience.

Issue #31184, opened 2020-10-21
Wrong font size in backend, revisited


nikosdion:
As I had already reported months ago, the default font size in Atum is too big. Yes, I understand that's the default Bootstrap and browser size albeit technically it's not exactly that.

First of all, Bootstrap 4's SCSS does allow you to set a different base font size instead of 1 rem. It's documented in it. Setting it to 0.875rem seems to be a better default and in line with literally every other site out there. I tried to file a PR to do that on the fly as we had discussed but I was told it's wrong even though that's exactly what we had agreed on AND what Bootstrap 4 recommends doing.

Furthermore, changing the font size from the browser's settings is not a viable solution for two reasons:

1. It affects all sites, not just Joomla. Since all other sites using rems instead of absolute pt/px sizes set a base font size between 0.8 rem and 0.9 rem they end up with type too tiny for me to read.
2. Chromium-based browsers such as Google Chrome, Microsoft Edge, Opera, Brave etc DO NOT have the option to change the based font size beyond "small", "medium", "large". The step from "medium" (16pt) to "small" is massive. One is too big to fit in the laptop display, the other is too small to read without a magnifying glass.

I honestly think that the right solution is changing the base font size as Bootstrap recommends. Since this is not something you want to consider I am asking you to find a magic solution. I am not going to participate further in the discussion as I've already told you why it's a problem and you have already rejected my solution (even though you agreed to it initially – and then you have the gall to tell me I'm "difficult" on Twitter, sigh). All I want is to be able to freaking use Joomla 4's backend WITHOUT having to add custom CSS. Is it too much to ask?

brianteeman:
see #31191

nikosdion:
Your PR doesn't change the font size, it only changes the calculations in SCSS. You'd still need to recompile Atum's SCSS every time you want a different font size and it'd apply globally (all users) which we said it's wrong when we discussed in length about it between April and September.

The only _viable_ way to make it all work correctly that *I* can see is having a base font size given as a client-side variable e.g. `var(--joomla-base-font-size)` with a default of `1 rem` set **in the template options**. A control in the accessibility settings could change that variable the same way you've been able to change colors in Joomla 3 (and Atum, only that user-facing controls were yanked away).

[...]

PR #31158, opened 2020-10-19
[4.0] More cleanup on the jstools

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Remove the .eslintignore
- Fix the eslint errors on the remaining WebComponents
- Add a try/catch for any Json.parse()
- Favour functional forEach over for loops

### Testing Instructions
All fields are working as before

PR #31191, opened 2020-10-21
[4.0] Atum font sizes

brianteeman:
PR for #31184

I took a look at this to create a PR the correct way for bootstrap scss

We should just have to change one variable
`$font-size-base:             	    0.875rem; //instead of 1rem`

But the rest of the font variables have not been defined correctly (probably based on bs4 beta1?) and they should be changed as well to be relative to this variable.

See the commit to bootstrap 3 years ago where this was done https://github.com/twbs/bootstrap/pull/24060

There may be other font sizes in our scss that need to be updated to use this variable - ~I have not checked yet~ _i believe I have done them all now but I deliberately ignored anything to do with icon fonts_

_**I have no opinion on a base font size of 16px (before) and 14px (with this PR)**_

To test you will need to run `npm ci` or `node build.js --compile-css`


administrator/templates/atum/scss/_variables.scss
```diff
@@ -110,21 +110,22 @@ $sidebar-width-closed:             3rem;
 $main-brand-height:                3rem;
 
 // Fonts
+$font-size-base:             	     0.875rem; // Instead of default 1rem
```
infograf768: CS => change to `.875rem`

richard67:
@brianteeman Could you update the branch for this PR with latest 4.0-dev of the CMS? Currently scss-cs is failing in Drone, but not due to code style of your PR. It seems the scss linter can't be started due to recently merged changes in the 4.0-dev branch of the CMS. In case if your 4.0-dev branch was not up to date with this when you have created the branch for this PR, this might be the reason. Thanks in advance.

bembelimen:
The changes looks good, is there any a11y requirement based on font size? I know, that is should be possible to resize up to 200% without broken layout, but is there any min size?

PS: did you add other PRs into your PR? The JS stuff looks irrelevant?

brianteeman:
grrh I updated the branch as requested and it looks like I did something wrong at the same time.

Regarding accessibility there are no rules or even recommendations about font sizes either minimum or maximum. The rules that applies are regarding contrast and a recommendation to use rem units to help scale content with a user’s personal preferences.

brianteeman:
@bembelimen @richard67 the unrelated changes are coming from #31158 but they are already merged so I dont understand why they are showing up here

richard67:
No idea either why they show up.

brianteeman:
Closing and will create a new pr with a new branch

brianteeman:
see #31207

Issue #31091, opened 2020-10-14
[4.0] SCSS in Joomla


thednp:
### Problem identified
I've looked into the default J4 template Cassiopeia and I didn't find any options to change colors or anything regarding styling, instead I found some SCSS files to push customized style and variables into Bootstrap.

I don't think this is the right approach to end-user "ease to use" promise.

### Open questions
* What/how is the SCSS compiled in Joomla 4? Is it php/js?
* How can we push custom SCSS variables, mixins or styling from template style administration panel?
* Is this a temporary/permanent state or an "under development" feature? (to be fair, it doesn't make sense to have color picker form field if we don't actually use it)

ReLater:
> What/how is the SCSS compiled in Joomla 4? Is it php/js?

I´t's compiled during the build process of an installable Joomla 4 package. It's JS using npm/node.

See the /buid/ directory. For example https://github.com/joomla/joomla-cms/blob/4.0-dev/build/build-modules-js/compilecss.es6.js

> How can we push custom SCSS variables, mixins or styling from template style administration panel?

No way to do that with an installed core package. There is no on-the-fly SCSS compilation from Joomla core.

> Is this a temporary/permanent state or an "under development" feature?

It's the current state ;-)

> (to be fair, it doesn't make sense to have color picker form field if we don't actually use it)

It makes sense if the selected colors are inserted as CSS.

At the moment I'm testing the Astroid framework. It's available already for Joomla 4: https://github.com/joomdev/astroid-framework

It comes with a PHP SCCS compiler. Maybe that's something for you?

thednp:
Yes and no. We need something that works with Joomla code and consistent with its features, correct? Why should I use a framework?

Bakual:
I've asked in the maintainer chat. The files are there for people who want to make an own template based on Cassiopeia or Atum. This way they can work from the SCSS files if they want.
Obviously, they need to have their own tooling for compiling their own custom template then. Chances are high that people dealing with SCSS files already have that tooling anyway.

SharkyKZ:
That doesn't work because some SCSS files are missing.

Bakual:
~Which ones are missing?~
Just found, you mean the ones in the build folder I guess. That looks wrong indeed.

SharkyKZ:
Any SCSS in `build/media_source`. I guess we can just copy them to to `media` directory when compiling CSS.

Bakual:
[...]

PR #31199, opened 2020-10-22
[4.0] Some SCSS files not shipped

SharkyKZ:
Fixes https://github.com/joomla/joomla-cms/issues/31091, at least in some part.

### Summary of Changes

Moves some SCSS files back to media folder so all SCSS files are shipped and CSS can be compiled in production environment.

### Testing Instructions

Testing BEFORE result requires a built package like nightly or beta. For testing AFTER result download the package from the link below this comment.

Download scssphp from https://github.com/scssphp/scssphp (bad idea, I know, but it's fine for testing purposes).
Use scssphp to compile some file, e.g. `templates/cassiopeia/scss/template.scss`.

CSS compilation fails because of missing files:

>../../../build/media_source/system/scss/icomoon` file not found for @import: ../templates/cassiopeia/scss/template.scss on line 21, at column 1

CSS file compiled.

No.


dgrammatiko:
Please DON'T reintroduce the media folder in the repo, there's another way to move the scss files to media (in the init module)

SharkyKZ:
I wanted to avoid having duplicate files in the working directory. This makes things complicated because assets needs to be copied after affected SCSS files are modified (because other SCSS files reference files from `media` directory). But OK.

dgrammatiko:
> This makes things complicated because assets needs to be copied after affected SCSS files are modified (because other SCSS files reference files from media directory)

I'll have to update the code for the tooling anyhow, but good to know what the requirements (for scss) are...

dgrammatiko:
@SharkyKZ please undo the changes in the build tools modules. I 've updated them in https://github.com/joomla/joomla-cms/pull/31221 to always ship and update as necessary the scss files per compilation

PR #31221, opened 2020-10-24
[4.0] Allow scss compilation for any template

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
Allow the scss compiler to transpile all the scss files [given that they exist in a folder named `scss`] in any template (frontend or administration). This allows the usage of the tools for developing custom templates.

### Testing Instructions
Run `npm run build:css` and check that both Cassiopeia and Atum are not visually broken. Also rename the configuration.php and check that the installation is also not visually broken

No.


thednp:
So now we expect end users to install npm and stuff on their server to do `run build:scss`?

brianteeman:
no

dgrammatiko:
> So now we expect end users to install npm and stuff on their server to do run build:scss?

No, unless you're developing on a live server: 

> This allows the usage of the tools for developing custom templates.

ghost:
I like that but:
I think you have to do something with the css files in the template folders, otherwise something will happen to those from line 61.
(Also, it wouldn't be bad to adjust the description in the header area)

thednp:
Also how is the end user customization working with Joomla?

dgrammatiko:
> Also how is the end user customization working with Joomla?

That's up to the template developer. FWIW you REALLY don't need SCSS anymore in 2020, just use css variables and plain css. The platform provides way better tools than the outdated Bootstrap...

> I think you have to do something with the css files in the template folders, otherwise something will happen to those from line 61.

Won't break anything although will be a good idea to **only** run the scss to css for templates (this needs some work on the system templates and I won't be doing it in this PR)

SharkyKZ:
Please use `git mv` to move files.

SharkyKZ:
Nevermind, Github UI shows files as being deleted but at least Git blame is still there 😕 .

dgrammatiko:
@SharkyKZ you can always blame me no matter what Github says 😃

astridx:
> @thednp Also how is the end user customization working with Joomla?

Even if I agree with @dgrammatiko and like to work as uncomplicated as possible ("just use css variables and plain css"), some things are historically based :)
I wrote a small plugin in which you can specify directories and files that are to be translated from scss to css when you open the website in the frontend. Maybe you like to take a look at this. I just uploaded it here: https://github.com/astridx/plg_system_agscsscompiler
It is based on: https://github.com/scssphp/scssphp
There is certainly a lot of room for improvement, but for my purposes it is sufficient. It works in Joomla 3 and 4

astridx:
I have tested this item :white_check_mark: successfully on dad8a81a18b66887702ff0990b8bbfdc7a07c539<br /><br />With the current changes everything works fine for me. I applied this branch, run `composer i` and `npm ci` and opened some pages in backend and frontend. I noticed nothing what is wrong.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31221">issues.joomla.org/tracker/joomla-cms/31221</a>.</sub>

dgrammatiko:
> There is certainly a lot of room for improvement, but for my purposes it is sufficient. It works in Joomla 3 and 4

Sorry but there's no room for improvement. PHP people have to realise that front end is based on node js. There aren't any tools in the PHP land that come close to what exists in the js land. It's so simple, deal with it. Also JS **is** one of the languages that make up the web...

thednp:
@dgrammatiko how about a `onSaveSettings` callback we can hook into and push our own code to do SCSS compile?

dgrammatiko:
> how about a onSaveSettings callback we can hook into and push our own code to do SCSS compile?

For the nth time: the solution **needs** to be something that runs in the client side (Joomla cannot suddenly require npm as a hosting requirement) and quite honestly the PHP solution for compiling scss to css is just hot garbage...

Once again I'll point you to my proposal for utilising wasm: https://github.com/dgrammatiko/scss-wasm-compile
Creating a simple Form Field with a simple button that will invoke the wasm part on the browser and then (if compilation is successful) will save it to the appropriate file is all it takes. Less than an hour to do it...

dgrammatiko:
@thednp @astridx FWIW today libSass just got deprecated: https://sass-lang.com/blog/libsass-is-deprecated

thednp:
Wonderful news. What's next? Joomla 2050?

HLeithner:
@dgrammatiko since we have to use dart sass in the future which have a native javascript version we could consider to use this in the browser even if I still don't know why we have to compile sass in the backend (this belongs to the other issuse).

dgrammatiko:
> I still don't know why we have to compile sass in the backend

Unfortunately this is a misconception. You can compile **ONCE** bootsctrarp with your needed/exposed css variables and then never have to recompile from scss. AFAIK even the current templates do that...

thednp:
@dgrammatiko it's about the end user customization, probably since Joomla 1.5 we used to have PHP powered CSS added for end-user customization purposes, now with modern tech, we are expected to deliver.

I would consider this a critical issue. The end-user (at least from my experience) expects to have color and logo customization options, everything working with modern and optimized code, I don't see where is my miss-conception.

dgrammatiko:
As I said before **use css variables**. The platform already solved this problem for us. Actually we don't need css preprocessor, the only reason we're using one is bootstrap but bootstrap is a monolithic framework and once you define all the scss vars as css vars you don't need to recompile... Is it so hard to get around what I'm saying...

thednp:
@dgrammatiko or anyone, perhaps you know, what is the JavaScript event on template settings saved? I will try and hook into that.

Thanks

dgrammatiko:
@thednp on form submit seems a good bet

thednp:
@dgrammatiko no, I mean the event triggered after the settings have been saved. I need to know the response, is it a failure/success so I know I can push valid SCSS variables into the compiler.

dgrammatiko:
> no, I mean the event triggered after the settings have been saved.

@thednp the response is a page reload. The form is not submitted via AJAX so there is no response other than what is returned from the server-side (in this case it's either the same edit page or the list of items, eg category view)

wilsonge:
@dgrammatiko can you rebase this one please. I think this is mergeable

dgrammatiko:
@wilsonge should be fine now

wilsonge:
Thanks!

thednp:
My question now is: is it possible to hook into this script to somehow push template developer's SCSS variable defined values into the grinder?

dgrammatiko:
>  is it possible to hook into this script to somehow push template developer's SCSS variable defined values into the grinder?

This script is available only when you clone the Joomla repo it's not part of the deliverables on the **installable** package. Joomla **IS NOT REQUIRING** node, this script is a node JS script. This is available only to developers. If for whatever reason someone needs to compile SCSS to css, let me repeat this for the nth time: the only realistic solution is to use JS+WASM as the PHP version of the SCSS compiler is hot garbage. I've also mentioned many times that SCSS compilation shouldn't be required as the templates are using CSS Variables so you could change (with inline styles) those styles eg: `:root { --primary: red; }` without the need for a preprocessor.

PR #31246, opened 2020-10-27
Use ZipArchive in JArchiveZip

SharkyKZ:
### Summary of Changes

`zip_*` functions are deprecated on PHP 8. Use `ZipArchive` instead.

### Testing Instructions

Test that extracting zip archive using `JArchiveZip` still works. E.g. use this code:

```
$zip = new JArchiveZip;
$zip->extract($file, $destination);
```
Where `$file` is path to zip archive and `$destination` is destination directory.

Works like before.

No.

### DIFF
--- a/libraries/joomla/archive/zip.php
+++ b/libraries/joomla/archive/zip.php
@@ -189,7 +189,7 @@ class JArchiveZip implements JArchiveExtractable
 	 */
 	public static function hasNativeSupport()
 	{
-		return function_exists('zip_open') && function_exists('zip_read');
+		return extension_loaded('zip');
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


alikon:
you should do the same fix on https://github.com/joomla/joomla-cms/blob/staging/libraries/vendor/joomla/archive/src/Zip.php too

SharkyKZ:
Already done.

alikon:
what i'm missing ?
https://github.com/joomla/joomla-cms/blob/staging/libraries/vendor/joomla/archive/src/Zip.php#L276-L315
is full of `zip_*`

HLeithner:
I tested the Zip Archive in our framework (which uses the same code) successfully.

Thanks

PR #19842, opened 2018-03-05
[4.0] Send the container in the onBeforeExecute event

laoneo:
With the introduction of service providers and the DI container, class overrides should be avoided in extensions. This pr adds the container to the `onBeforeExecute` plugin event. Like that an extension developer can override the services in the container.

The following code example registers it's own form factory:
```
public function onBeforeExecute(BeforeExecuteEvent $event)
{
	$event->getContainer()->share(
		FormFactoryInterface::class,
		function (Container $container)
		{
			return new MyAwesomeFormFactory;
		},
		true
	);
}
```

NOTE:
Not sure if a new event should be created for that, was lazy and did use an existing one.

PR #31251, opened 2020-10-27
[4.0] protected $_router instead of private

sakiss:
### Summary of Changes
Set the `$_router` variable as protected instead of private in the class: `Joomla\CMS\Router\Route`
By doing that we can extend the Route class and define our own Router in that.

This way we can provide additional functionalities in our Router.

### Testing Instructions
It can only be tested by extending the `Joomla\CMS\Router\Route` and defining our own Router

No Router can be defined in sub-classes.

You can define your own Router in sub-classes.

DKN

### DIFF
--- a/libraries/src/Router/Route.php
+++ b/libraries/src/Router/Route.php
@@ -49,7 +49,7 @@ class Route
 	 * @var    Router[]
 	 * @since  3.0.1
 	 */
-	private static $_router = array();
+	protected static $_router = array();
 
 	/**
 	 * Translates an internal Joomla URL to a humanly readable URL. This method builds links for the current active client.


HLeithner:
Can you give me an example what you want to achieve?

sakiss:
@HLeithner In my case i want to override the [Router::createUri](https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Router/Router.php#L432), in order to encode the special characters included in the url's query vars.

But i think that can be used in a more broad way e.g. set validation and escape rules.

HLeithner:
Maybe I'm wrong but wouldn't it be easier to add a preprocessor?

sakiss:
@HLeithner Thanks for the suggestion.
Any idea on how this can be implemented?

HLeithner:
Create a plugin and add a buildRule in the onAfterInitialise event. somehing like this:

```php
class plgSystemMyRouterExtgension extends JPlugin {

public function onAfterInitialise() {
  $app = JFactory::getApplication();
  if ($app->isSite()) {
    $router = $app->getRouter();
    // attach build rules for translation on SEF
    $router->attachBuildRule(array($this, 'buildRule'), JRouter::PROCESS_BEFORE);
  }
}

public function buildRule(&$router, &$uri) {
  /* modify $uri here */
}

}
```

sakiss:
@HLeithner Thank you!
The [actual problem is in the Uri class](https://github.com/joomla-framework/uri/issues/22), hence it requires more than adding rules.

In any case i think that providing the ability to set our own Router could add extra flexibility.

HLeithner:
Can you send me the code you want to use in your router if this variable is not private?

sakiss:
I am overwriting the `Router::createUri` by calling my own [Uri class](https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Router/Router.php#L436) within that fn.

dgrammatiko:
@sakiss if you're essentially overriding the Uri class probably you should consider doing it like the example here: https://github.com/joomla/joomla-cms/pull/19842

sakiss:
@dgrammatiko Valid. The only reason i don't like that, is that is sets an object in the container which will be used globally. I prefer having something with limited scope. But i understand the point that the same class should be used on input and output (build/parse).

Issue #26299, opened 2019-09-14
[4.0] Missing language strings in installer


roland-d:
### Steps to reproduce the issue
1. Install an extension that you know will fail the installation
2. Notice you get a popup with a message like this:
![image](https://user-images.githubusercontent.com/359377/64907180-6fa31f80-d6ef-11e9-8d3b-6407c2fbab0c.png)

The issue is that we have this code in the installer `'JLIB_INSTALLER_' . $this->route` but there are no translations for them.

We need to find out which routes we have, to know which messages we need.

### Expected result
The message is translated

### Actual result
The message is untranslated

### Additional comments
I am putting this issue here as a reminder so we don't forget about them

brianteeman:
@roland-d are you sure that this is specific to Joomla 4 as it looks the same code as in joomla 3

roland-d:
@brianteeman I am not sure it is specific to J4 as I did not look at J3, I just ran into this and wanted to make sure we don't forget about it.

ceford:
I have had a look at this issue by searching the code for occurrences of JLIB_INSTALLER_' and I think there is only one string missing: JLIB_INSTALLER_DISCOVER_INSTALL and it is the same in J3. I could do a PR for J4. However, I don't know how to test it. Clearly I need to unpack a component and discover it. I tried that with the search component from J3 and it installed. I don't know how to set the PHP minimum requirement as suggested by the screenshot above. Also, I noticed that a lot of the instances of JLIB_INSTALLER_ were not followed by strtoupper - in ModuleAdapter.php, PackageAdapter.php, PluginAdapter.php and more. Should that be a separate issue? Or does it not matter?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/26299">issues.joomla.org/tracker/joomla-cms/26299</a>.</sub>

alikon:
please test #31315

PR #31276, opened 2020-10-30
[4.0] Discover string fix

ceford:
Pull Request for Issue #26299 .

### Summary of Changes

Added strings JLIB_INSTALLER_DISCOVER_INSTALL="Discover Install"
In admin and site lib_joomla.ini

I noticed that there places where strtoupper is missing: e.g  - Text::_('JLIB_INSTALLER_' . $this->route), - leaving that as a separate issue but could fix if anyone would like to confirm - look in libraries/src/Installer/InstallerAdapter.php

### Testing Instructions

See #26299 
I do not know how to test this PR! I checked the code and I am fairly sure there is only one string missing from the set discover_install, install, update and uninstall. 

Check the code: the strings are missing.

Check the code: the strings are present.

None


chmst:
I think that the string 'discover_install" is wrong and should be only "discover" in installer_adapter.php. Then the Language Key is in the language file like all the others.

SharkyKZ:
@ceford This PR is fine but can you make it against staging instead?

ceford:
> @ceford This PR is fine but can you make it against staging instead?

Being relatively new to github, I don't know how to change a pr. And I don't know how changes in staging make their way to 4-Dev. I don't seem to have staging on my local clone. A few pointers would be helpful.

SharkyKZ:
@ceford Depends on what you're using. If branch is not available, you may need to use `git checkout` command to check out staging branch.

brianteeman:
You can change the branch here

ceford

ceford:
Closing this to create a new request with new base. I don't get a list of my requests so I can't select this one to edit.

PR #31297, opened 2020-11-02
[4.0] Fixes issue with no bold (strong) font in cassiopeia

regularlabs:
Making any text bold in cassiopeia has no effect, because only 1 weight of the Fira Sans font is being loaded in.
This fixes that.


brianteeman:
This is a very valid issue and PR - however as cassiopeia is undergoing development in a separate branch before being merged back here you might want to check it there. https://github.com/joomla/cassiopeia

regularlabs:
Ok, seems they dropped using Fira...

Issue #31138, opened 2020-10-17
[4.0] Empty Alt descriptions [a11y]


brianteeman:
An alt text provides an accessible description of an image. Most of the time you should always have one. You should never not have any alt because then assistive technology such as a screenreader will announce the filename. See the October Joomla Magazine article on this ( published on Tuesday, 20 October 2020)

In the rarer situation that the image is purely decorative then it is acceptable to have an empty alt text which should be `alt=""`. 

Currently there are several places in Joomla where if you are too lazy to enter a text description of the image joomla will insert `alt=""` While this does prevent the filename being read aloud it is not good practice as you are preventing a significant % of your readers being able to access all the content. If the content author performs an accessibility check of their site then _most_ checkers will not report any errors or warnings about missing descriptions as it sees the `alt=""`

We should be doing everything to encourage the use of alt text descriptions and one method used by some other cms is that the  alt text input field is required so you have to enter a description in order to insert the image OR you have to check a box that says its is a decorative image and should use an empty `alt=""`

I am not sure about making it required (although I have done on a site I manage) but I am definitely in favour of 
1. Not defaulting to `alt=""` if there is no text entered
2. Having a check box that must be checked if it is a decorative image and then outputting `alt=""`

This way we still allow the user to have an empty alt AND we allow the user to check their site for missing descriptions.

![chrome_2020-10-17_13-38-39](https://user-images.githubusercontent.com/1296369/96350961-6afa2900-10b0-11eb-9f21-2e0263451951.png)

richard67:
Makes sense to me.

adj9:
On this wake I realized that it is necessary to redefine the Random image module.
Now the alt is spelled out by the file name, but it is not necessarily appropriate for the multimedia content. If it is possible to have the reader of the accessibility plugin read the alt is nice. I haven't seen the technical part yet. :)

chmst:
It surely makes sense to ask for attention on alt - texts. 
But do we really want to add an additional input field to every image? So that blind users or keyboard users have more to do? 
I suggest to add a hint to the field, but no additional checkbox.

brianteeman:
@chmst without the extra field the problem is as follows. Maybe you have a better solution to fixing this

1. User creates a description
output is alt="the description"

2. User deliberately does not add a description as its decorative only
output is alt=""

3. User too lazy to add a description
here there must be no alt at all

I do not see how you can achieve point 2 and point 3 without the extra field

[...]

PR #31318, opened 2020-11-05
[4.0] Intro and Full image accessibility

brianteeman:
An alt text provides an accessible description of an image. Most of the time you should always have one. You should never not have any alt because then assistive technology such as a screenreader will announce the filename. See the October Joomla Magazine article on this

In the rarer situation that the image is purely decorative then it is acceptable to have an empty alt text which should be alt="".

Currently there are several places in Joomla where if you are too lazy to enter a text description of the image then joomla will insert alt="" While this does prevent the filename being read aloud it is not good practice as you are preventing a significant % of your readers being able to access all the content.

Research conducted by the EU shows that content creators want to (need to) create accessible content but don't really know how. The main tool that they have available to them is an accessibility checker. And they then act upon the findings of that tool.

**If the content author performs an accessibility check of their site then most checkers will not report any errors or warnings about missing descriptions as it sees the alt="" and ignores it.**

Adding alt="" is perfectly valid for a decorative image only. It should be a conscious/active decision to mark an image as decorative.

Comment from @brucel confirming this https://github.com/joomla/joomla-cms/issues/31138#issuecomment-711789514

Comment from @mgifford confirming this https://github.com/joomla/joomla-cms/issues/31138#issuecomment-712222061

Adding an additional help text to the description is not enough

Making this a required field is not possible in Joomla with the current codebase as it would only be required IF there is an image and we dont have support for that condition.

Suggesting that you have to put `""` in the description to mark something as decorative is not very user friendly and it is actually adding more keystrokes etc than skipping to the next field and checking the box - as explained to @carcam

The correct use of alt text is **not** advanced accessibility. It is very basic accessibility and addressed at level A

There is no point in making the commitment and statement that Joomla itself is accessible if it at the same time makes it harder to create an accessible site.

### Testing
Testing is really easy. No npm, js or css involved.
Merge the pull request and then go to edit an article on both the admin and site. For the site you might need to enable the frontend images and urls option in the component.

You now have a new field and tweaked text

Please test three scenarios when adding an intro image

1. Image Description (Alt Text) = Empty
No Description = unchecked

2. Image Description (Alt Text) = "some description"
No Description = unchecked

3. Image Description (Alt Text) = Empty
No Description = checked

The expected behaviour for each of these tests is

1. ``
2. `some description`
3. ``
or ``

PLEASE do not comment on what you think the code does but apply the PR and test it. Getting very tired of people blocking PR with their comments without actually testing the code.

PS we will need to do similar changes elsewhere


C-Lodder:
Seems good.

Question though. Why have another parameter for this (checkbox) as opposed to completely omitting the `alt` attribute if no description is provided?

ceford:
I have tested this item :white_check_mark: successfully on c5dd8cb777f9b7b082238013144e512275fc467f<br /><br />It does what it says. But, case 4: Description present and checked = Case 2.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31318">issues.joomla.org/tracker/joomla-cms/31318</a>.</sub>

brianteeman:
@C-Lodder there are two different scenarios. An empty ALT by ignorance and a blank ALT by choice. alt="" is valid and the correct thing to do for a purely decorative image. Instead of the checkbox it could be done by saying you have to enter "" in the alt text field to indicate it is a decorative image. Pre-existing user research and testing on this says that this would not be a user friendly or intuitive option. 

@ceford I have no  idea what you are saying - there is no case 4

C-Lodder:
> An empty ALT by ignorance and a blank ALT by choice

Ahh yes, sorry.

C-Lodder:
I have tested this item :white_check_mark: successfully on c5dd8cb777f9b7b082238013144e512275fc467f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31318">issues.joomla.org/tracker/joomla-cms/31318</a>.</sub>

ceford:
In anticipation of this being merged soon I have changed the Help screen to provide a new screen shot and this in the text:

>**Image Description (Alt Text).** Set the alt attribute for this image. A few descriptive words for screen readers.
>**No Description.** Check in the rare instance of a purely Decorative Image - no description required. Note that if the Image Description is empty and the No Description checkbox is unchecked then the image will fail to meet accessibility criteria. If an image description is present this check box has no effect.

It will take a while for this change to pass through the translation system. Please come back if my understanding is incorrect.

carcam:
@brianteeman thank you very much for this commit and I do appreciate you consider the options I suggested about double quotes and making the field required (which I also know it was our [yours and mine] preferred options).

I don't see anything against this and as per the motives we discussed it looks like a great addition to Joomla A11y.

I want to discuss this with the team but remembering an old saying from my university times: "Better ask forgiveness than permission" (sorry for the bad translation), I would go for requesting RTC status (not even sure if this is something I can do).

Also I will propose JAT to review the preview PR that was rejected as for providing an homogeneus UI, it looks like it is the correct path (though as you know I'm against adding more fields to the form), but that is another story.

Thank you very much indeed.

RTC

Best!!

infograf768:
@carcam 
If you have tested this pr ok, please add your test on issues.joomla.org

carcam:
@infograf768 it already has 2 tests, right?

infograf768:
a 3rd test adds value to the Pr ;)

Rtc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31318">issues.joomla.org/tracker/joomla-cms/31318</a>.</sub>

infograf768:
Note: I guess a similar pr will be done for category or is it useless there?

brianteeman:
Already done hours ago

layouts/joomla/content/intro_image.php
```diff
@@ -14,22 +14,24 @@
 
 $params  = $displayData->params;
 $images  = json_decode($displayData->images);
+
 ?>
 <?php if (!empty($images->image_intro)) : ?>
 	<?php $imgclass = empty($images->float_intro) ? $params->get('float_intro') : $images->float_intro; ?>
+	<?php $alt = empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : 'alt="'. htmlspecialchars($images->image_intro_alt, ENT_COMPAT, 'UTF-8') .'"'; ?>
```
Quy, line 21:
```suggestion
	<?php $alt = empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : 'alt="' . htmlspecialchars($images->image_intro_alt, ENT_COMPAT, 'UTF-8') . '"'; ?>
```

brianteeman: cs

chmst:
Thanks!

brianteeman:
Thanks

PR #31328, opened 2020-11-06
[4.0] https updates and fix @link

brianteeman:
This pr changes the url for some commented links to https AND updates the non-working w3c links to use working links to the living standard

code review

~for reviewers the only change in the svg files is `xmlns="http://www.w3.org/2000/svg` changed to https~


manojLondhe:
I have tested this item :red_circle: unsuccessfully on 5ceaed7ab5238fc1144c54104abb18b463f4d92b<br /><br />While code changes look fine, after applying patch using patch-tester backend logo is not getting loaded

Before
before

After
after

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31328">issues.joomla.org/tracker/joomla-cms/31328</a>.</sub>

brianteeman:
I have removed the svg files from this PR

manojLondhe:
I have tested this item :white_check_mark: successfully on e83032f8ee91a2843968ccb4cae8b80543dd1f63<br /><br />For some reasons applying this patch using patch tester was not working. Not all files were getting patch applied to (eg - no changes were seen in /templates/cassiopeia/offline.php)

So I tested by applying [patch](https://patch-diff.githubusercontent.com/raw/joomla/joomla-cms/pull/31328.diff "patch")  manually as

`git apply 31328.diff`
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31328">issues.joomla.org/tracker/joomla-cms/31328</a>.</sub>

administrator/components/com_redirect/src/Field/RedirectField.php
```diff
@@ -36,7 +36,7 @@ class RedirectField extends ListField
 	 *
 	 * @var    object
 	 * @since  3.4
-	 * @link   http://www.iana.org/assignments/http-status-codes/
+	 * @link   https://www.iana.org/assignments/http-status-codes/
```
PhilETaylor: use `https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml` as the old link is just redirected 302 to the non trailing slash https://www.iana.org/assignments/http-status-codes and then 302 again to https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

libraries/src/Document/Document.php
```diff
@@ -1043,7 +1043,7 @@ public function getModifiedDate()
 	 *
 	 * @since   1.7.0
 	 *
-	 * @link    http://www.w3.org/TR/xhtml-media-types
+	 * @link    https://www.w3.org/TR/xhtml-media-types
```
PhilETaylor: add trailing slash to prevent 302 redirect to https://www.w3.org/TR/xhtml-media-types/

libraries/src/Form/Filter/UrlFilter.php
```diff
@@ -50,7 +50,7 @@ public function filter(\SimpleXMLElement $element, $value, $group = null, Regist
 		$value = InputFilter::getInstance()->clean($value, 'html');
 		$value = trim($value);
 
-		// <>" are never valid in a uri see http://www.ietf.org/rfc/rfc1738.txt.
+		// <>" are never valid in a uri see https://www.ietf.org/rfc/rfc1738.txt.
```
PhilETaylor: trailing period in the comment makes the URL invalid and 404

brianteeman:
thanks @PhilETaylor

Quy:
I have tested this item :white_check_mark: successfully on 6133e42abe0439b290409fd85ff656cfb27bf8f5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31328">issues.joomla.org/tracker/joomla-cms/31328</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on 6133e42abe0439b290409fd85ff656cfb27bf8f5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31328">issues.joomla.org/tracker/joomla-cms/31328</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31328">issues.joomla.org/tracker/joomla-cms/31328</a>.</sub>

HLeithner:
This is wrong because http://www.w3.org/2001/XMLSchema-instance is a name space not an URI, so changing it would break the schema. Revert this part or find an updated document with an https URI, I didn't found any.

brianteeman:
change made as requested by @HLeithner

HLeithner:
Thanks

brianteeman:
Not going to pretend that I 100% understood

PhilETaylor:
that's because @HLeithner is actually wrong.

http://www.w3.org/2001/XMLSchema-instance actually is hard redirected to https://www.w3.org/2001/XMLSchema-instance when you visit it, the non-https version no longer exists as a namespace.

HLeithner:
> PhilETaylor
> that's because @HLeithner is actually wrong.
> http://www.w3.org/2001/XMLSchema-instance actually is hard redirected to https://www.w3.org/2001/XMLSchema-instance when you visit it, the non-https version no longer exists as a namespace.

why did you deleted your comment @PhilETaylor

PhilETaylor:
because I was drafting another one, with more details... but as this is already merged there is no point. 

Using the http or https version of the name space is perfectly acceptable. There was no reason for it to be reverted. 

http://www.w3.org/2001/XMLSchema-instance redirects to https://www.w3.org/2001/XMLSchema-instance

Their content is the same. It makes no difference if the namespace is loaded over https or http, the result is the same. 

infuse the response contains `content-security-policy: upgrade-insecure-requests` so most secure clients/browsers would auto upgrade to https anyway

The actual ` targetNamespace="http://www.w3.org/2001/XMLSchema-instance"` remains HTTP and not HTTPS regardless of the http protocol used to retrieve the namespace document. 

https://www.w3.org/TR/xmlschema-1/ states:
> 4.3.1 Standards for representation of schemas and retrieval of schema documents on the Web

> For interoperability, serialized ·schema documents·, like all other Web resources, may be identified by URI and retrieved using the standard mechanisms of the Web (e.g. http, https, etc.) Such documents on the Web must be part of XML documents (see clause 1.1), and are represented in the standard XML schema definition form described by layer 2 (that is as <schema> element information items).

PR #31359, opened 2020-11-07
[4.0] tags menu items image alt [a11y]

brianteeman:
Continues the work of  #31318 and #31323 by correcting the behaviour of the alt text and changing the strings to match the work done in #31318

This is very slightly different to the other similar PR as before this PR there were no alt text at all not even empty ones

### Background
To read why this change is very important see #31318 

### Testing Part 1
Testing is really easy. No npm, js or css involved.
Merge the pull request and then create one of each of the three types of tag menu items and add an image

#### List All Tags
image

#### Tagged Items & Compact Tagged Items
image

### Testing Part 2

1. Image Description (Alt Text) = Empty
No Description = unchecked

2. Image Description (Alt Text) = "some description"
No Description = unchecked

3. Image Description (Alt Text) = Empty
No Description = checked

The expected behaviour for each of these tests is

1.``
2. `some description`
3. ``
or ``

PLEASE do not comment on what you think the code does but apply the PR and test it. Getting very tired of people blocking PR with their comments without actually testing the code.

cc @carcam


Quy:
Tested successfully. Listed as a discussion in issue tracker.

brianteeman:
Conflicts resolved

brianteeman:
Looks like this doesnt appear on the issue tracker. It would be great if those on github could test it and we can get it merged as no one using patchtester will see it

brianteeman:
@Quy @richard67 @chmst  It would be great if you could test and merge this - apparently its missing from JIssues so no testers see it

Quy:
Tested successfully.

chmst:
The PR works when the user sets values correct. But we have 4 cases 

```
alt text       |   decorative only |  Result
--------------------------------------------
not empty |   checked             |   alt text is displayed (is this intended?)
not empty |   not checked       |  alt text is displayed (correkt)
empty       |   checked              |  alt = "" (correct)
empty       |   not checked       |  no alt text at all - as much as i know this is not a11y 
```

brianteeman:
This is not about making all images have alt text or "".

Its all about making a conscious decision if an image has alt text or if it has "" which is correct for marking it as decorative

Its important that "empty | not checked " exists as thats the equivalent to not making a conscious decision and will now mean that an accessibility audit will identify the image as needs review to be given either an alt text or to be marked as decorative



empty       |   not checked       |  no alt text at all - as much as i know this is not a11y

chmst:
ok, I forgot. Tested successfully.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31359">issues.joomla.org/tracker/joomla-cms/31359</a>.</sub>

brianteeman:
Thanks all

Issue #30793, opened 2020-09-28
Templating - fontawesome not loaded correctly


ced1870:
### Steps to reproduce the issue
create a new template using the Web Assets
$wa  = $this->getWebAssetManager();
$wa->useStyle('template.test.ltr');

in the joomla.asset.json file :smile: 
"assets": [
    {
      "name": "template.test.ltr",
      "type": "style",
      "uri": "template.css",
      "dependencies": [
        "fontawesome"
      ]
    },

### Expected result
to load the library Fontawesome from the joomla core

### Actual result
load the media/vendor/fontawesome-free/css/fontawesome.min.css file, but the @font-face is missing

no FA icon is working in the page, because the font is not declared :smile: 
@font-face{font-family:Font Awesome\ 5 Free;  ... also for the Brand

in Cassopeia, the font-face is included in the template.min.css file, but if we load FA from the core, it shall be loaded correctly also with the font-face. Else we shall declare the font-face each time, for each template/extension and this is not correct

### System information (as much as possible)
4.0.0 beta 4

N6REJ:
thanks for reporting this.
This is happening because they (fa ) expect you to load the fontset yourself.
```
  <!-- reference your copy Font Awesome here (from our CDN or by hosting yourself) -->
  <link href="/your-path-to-fontawesome/css/fontawesome.css" rel="stylesheet">
  <link href="/your-path-to-fontawesome/css/brands.css" rel="stylesheet">
  <link href="/your-path-to-fontawesome/css/solid.css" rel="stylesheet">
```

ReLater:
> but if we load FA from the core, it shall be loaded correctly also with the font-face. Else we shall declare the font-face each time, for each template/extension and this is not correct

I agree. Joomla has the webfonts packed. Therefore their should be additional CSS that can be used as additional dependencies to decide which fonts to load (or not load) from /media/ folder.

ced1870:
> This is happening because they (fa ) expect you to load the fontset yourself.
may be but in joomla asset there is 1 dependency, "fontawesome". It shall call all the needed CSS to make it work. Not only a part of it ...
as per the FA doc, you must copy and load the 3 files to make it work. They don't say to load only one file

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/30793">issues.joomla.org/tracker/joomla-cms/30793</a>.</sub>

[...]

PR #31364, opened 2020-11-08
[4.0] fix: fontawesome with font-face

Fedik:
Pull Request for Issue #30793 .

### Summary of Changes
This loads fontawesome.css with font-face

### Testing Instructions
run `npm install`  and please look #30793 for detail

font-face not loaded

font-face loaded

none


richard67:
@ced1870 Please test. Thanks in advance.

dgrammatiko:
Loading **all** the fonts by default results to 3 fonts that need to be fetched with a size (at least) over 164kb +(59kb for the css file). The problem is that this css is a render-blocking asset, meaning nothing is painted in the screen before the css and all the dependencies are fetched, parsed and evaluated. In sort say bye-bye to any meaningful lighthouse scores destroying any SEO, UX and eventually, the ROI of anything build with Joomla...
There are 2 ways out of this:
- Move to svg for any rendered icon, which is what everybody is doing anyways
- Lazily load the fonts (has caveats, eg icons will popup out of nowhere **after** the page has already has been painted)

Screenshot 2020-11-08 at 13 18 54
Screenshot 2020-11-08 at 13 19 56

Fedik:
A modern browsers will load the font file only when it really used.
The PR fine as it is.

dgrammatiko:
> A modern browsers will load the font file only when it really used.

True, but the css is still render-blocking so bye-bye meaningful performance. I mean the PR is fine but font awesome just destroys performance and this has nothing to do with the PR rather the higher-order decision to stick with the outdated practice of using fonts for the icons.

Anyways my comment was more to indicate that Joomla is drifting away from being a viable and competitive solution due to the enforced Bootstrap, Jquery and Font Awesome. People will, eventually, figure this out the hard way, sooner or later...

ced1870:
Hi, sorry but I don't understand in which file I shall make the changes ...

Fedik:
you have to run `npm install` 
or download "prebuild" attached at end of this page:
image

ced1870:
I have downloaded this version and wanted to check the icons in the article using my own template. But, there is no Fontawesome icons in this version ! it still uses the Icomoon !
how am I supposed to test that ?

Fedik:
how you did it in #30793? that how you should test

drmenzelit:
I tried to test this PR: applied the patch, run npm, removed the `fontawesome.min.css` file from Cassiopeia. I see now that the `media/vendor/fontawesome-free/css/fontawesome.min.css` is being loaded. The file loads the font-face, but the font is not there, I think the path to the webfonts is not correct, e.g. `url(../webfonts/fa-solid-900.woff2) format("woff2")`. And the definitions of the Joomla icons (e.g. <span class="icon-search"></span>) are missing.

Fedik:
that correct, it a fix for default `fontawesome` build, for whom want to use it (in own template), it does not include joomla icons :wink: 

if you try to use some fa-blabla icon you will get it loaded:
Screenshot_2021-02-28_15-13-57

HLeithner:
> for whom want to use it (in own template), it does not include joomla icons 😉

this hit me yesterday, so I'm in favor for merging this.

as @Fedik already said "fonts are only loaded when used", of course if you use them they will be loaded and svg maybe better but it's too late for that.

For the b/c layer in atum and cassopeia I would move the icomoon css to it's own joomla asset so it can be loaded only when needed. (I don't know why cassopeia loads b/c icons but I'm sure there is a reason). for the Backend 3rd party extensions that needs icon could simply include the b/c package with the asset manager.

drmenzelit:
> I don't know why cassopeia loads b/c icons but I'm sure there is a reason

Cassiopeia doesn't load b/c icons, in the Joomla views we still have the old classes "icon-edit", "icon-calendar", etc. What was done, is a conversion / mapping from the classes with Fontawesome content (icons). The file `media/system/css/joomla-fontawesome.css` contains these mapping and loads the webfonts.
This PR will load Fontawesome / webfonts without the Joomla icons.

HLeithner:

drmenzelit:
Closing this in favor of using `media/system/css/joomla-fontawesome.css` in own templates, because that file includes the mapping of Joomla icons to Fontawesome.

PR #31382, opened 2020-11-11
[4.0] Fixed titles for 3rd party component dashboards

Harmageddon:
### Summary of Changes
In version 4, extensions can create their own dashboards. While this is a really cool feature, it seems to be mostly undocumented, so I had to do a bit of research on how to use this, by reading PRs like #28027, #27773 and similar.
The code for `com_cpanel` contains parts to define the title and icon used for the dashboard. However, these don't seem to work. There is a test that checks if the `$extension` string starts with `com_`. But previous filtering converts `com_something` into `com-something`.

### Testing Instructions
1. Install a component that uses an own dashboard. For this purpose, I created https://github.com/Harmageddon/com_demoj4 which you can use.
  If you develop your own extension, this is the perfect occasion to test dashboards. ;-)
2. In backend, click on the dashboard icon next to the component's menu item.
3. Look at the title on top of the page.

### Actual result BEFORE applying this Pull Request
There is no icon and the title says "Home Dashboard".

### Expected result AFTER applying this Pull Request
An icon is shown and the title says "DemoJ4 Dashboard" (or what you have specified as `COM_<YOUR COMPONENT>_DASHBOARD_TITLE`.

### Documentation Changes Required
The whole feature lacks documentation. I'm not even sure if it is correct to name the dashboard `com_demoj4` for a component with the name `com_demoj4` or it should be just `demoj4`.

Pinging @Bakual as I saw you used the dashboard already for your extension. Of course, it would be great to have other extension developers here as well who might be already using this feature as well.

PR #31392, opened 2020-11-12
[4.0] Move language strings for dashboard titles to respective component.

Bakual:
Inspired by discussion in https://github.com/joomla/joomla-cms/pull/31382

Currently, the component specific dashboards of com_menus, com_content, com_users and com_privacy take their dashboard title from a string in com_cpanel.
This makes no sense and is a special threatment of core components which shouldn't happen.

Also current code requires that 3rd party dashboard are named as "com_foo" unlike core dashboards which are just named "foo". This PR will work with both variants for 3rd party.

### Summary of Changes
Moving the strings to the correct component language file
Adjusting the title lookup code so it first tries to load the string from a component language file. It prepends the dashboard name with "com_" (eg "content" becomes "com_content) if not already present to guess the related component.
If no string is found it will try with a com_cpanel string.

I removed the string COM_CPANEL_DASHBOARD_CONTENT_WORKFLOW_TITLE="Content Workflow Dashboard" as I couldn't find a workflow related dashboard.

### Testing Instructions
Test the various dashboards if the title is still the same.

Titles are shown, but taken from com_cpanel

In core all looks fine, titles come from the component language files

Maybe if dashboard documentation exists somewhere.


Bakual:
On a sidenote, the icon code is very strange as well, using a language string to define the icon used. However language strings are for translating purposes, not for design.
For core this was already changed to a hardcoded list. Which is better but still not good and not useable for 3rd parties.

If someone has a good idea how to solve this, I'm all ears.

administrator/language/en-GB/com_cpanel.ini
```diff
@@ -8,13 +8,8 @@ COM_CPANEL_ADD_DASHBOARD_MODULE="Add module to the dashboard"
 COM_CPANEL_ADD_MODULE_MODAL_TITLE="Add Module"
 COM_CPANEL_DASHBOARD_BASE_TITLE="Home Dashboard"
 COM_CPANEL_DASHBOARD_COMPONENTS_TITLE="Components Dashboard"
-COM_CPANEL_DASHBOARD_CONTENT_TITLE="Content Dashboard"
-COM_CPANEL_DASHBOARD_CONTENT_WORKFLOW_TITLE="Content Workflow Dashboard"
```
infograf768, line 12: Is this one used?

ceford:
I have tested this item :white_check_mark: successfully on 6483d07737ac505259cb4b3383375b6dadf34089<br /><br />Could the icon name be in the manifest xml file? Or in config.xml as a hidden field?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31392">issues.joomla.org/tracker/joomla-cms/31392</a>.</sub>

administrator/language/en-GB/com_cpanel.ini
Bakual: I haven't found any use of that, that's why I deleted it (mentioned in the PR description).

Bakual:
> Could the icon name be in the manifest xml file? Or in config.xml as a hidden field?

Could be an idea, I'll have a look at it. Maybe even the title could be taken from there, would make things much more intuitive for sure.

Bakual:
Did an alternativ PR which makes use of the manifest file for title and icon specifications: https://github.com/joomla/joomla-cms/pull/31397

Bakual:
I'm closing this one, as the other basically became an extended version. 😄

Issue #30225, opened 2020-07-30
[4.0] Joomla\CMS\Session\MetadataManager::checkSessionRecordExists() extra SQL query


Denitz:
### Is your feature request related to a problem? Please describe.

If the session handler is database, `Joomla\CMS\Session\MetadataManager::checkSessionRecordExists()` performs an extra database query which is is actually actually useless.

### Describe the solution you'd like

This method is executed `onAfterSessionStart`, hence the record in `#__session` always exists if a session is stored in the database.

For database session storage we can omit the check for table record presence because it always exists at this stage (sure if the session is in Active state).

Denitz:
Sorry, no time for PR, here is the sample code for `Joomla\CMS\Session\MetadataManager::createOrUpdateRecord()`

```
	public function createOrUpdateRecord(SessionInterface $session, User $user)
	{
		// Active session with database storage always exists at this stage.
		if ($this->app->get('session_handler') === 'database' && $session->isActive())
		{
			$exists = self::$sessionRecordExists;
		}
		else
		{
			$exists = $this->checkSessionRecordExists($session->getId());
		}

		// Do not try to touch the database if we can't determine the record state
		if ($exists === self::$sessionRecordUnknown)
		{
			return;
		}

		if ($exists === self::$sessionRecordDoesNotExist)
		{
			$this->createSessionRecord($session, $user);

			return;
		}

		$this->updateSessionRecord($session, $user);
	}
```

Denitz:
https://github.com/joomla/joomla-cms/pull/31421

PR #31421, opened 2020-11-17
[4.0] Remove extra session query

Denitz:
Pull Request for Issue #30225 .

### Summary of Changes

If the session handler is database, `Joomla\CMS\Session\MetadataManager::checkSessionRecordExists()` performs an extra database query which is is useless because if we have active session at this stage, it means that the data was loaded from `#__session` table before and the session record exists.

We can check the session presence for:
1. Non-database handler.
2. Inactive session.
3. New session (calling isNew() method results in getting session data, which also results in auto-writing the session record in the database for the new record, next it's updated via `updateSessionRecord()` ).

### Testing Instructions

Use database session handler, load page multiple times, see that we always have 3rd query:

```
SELECT `session_id`
FROM `jos_session`
WHERE `session_id` = :session_id LIMIT 1
```
which is executed right after:
```
SELECT `data`
FROM `jos_session`
WHERE `session_id` = ?
```

The 3rd query always exists.

The 3rd query from above is only executed once on a new session. You can test the new session creation and expiration by truncating the `#__session` table, or deleting cookies, or setting past time in `#__session` table.

No.

### DIFF
--- a/libraries/src/Session/MetadataManager.php
+++ b/libraries/src/Session/MetadataManager.php
@@ -120,19 +120,24 @@ final class MetadataManager
 	 */
 	public function createOrUpdateRecord(SessionInterface $session, User $user)
 	{
-		$exists = $this->checkSessionRecordExists($session->getId());
-
-		// Do not try to touch the database if we can't determine the record state
-		if ($exists === self::$sessionRecordUnknown)
+		// Active non-new session with database handler always exists at this stage.
+		// We should only check session presence for non-database or inactive or new sessions.
+		if ($this->app->get('session_handler') !== 'database' || !$session->isActive() || $session->isNew())
 		{
-			return;
-		}
+			$exists = $this->checkSessionRecordExists($session->getId());
 
-		if ($exists === self::$sessionRecordDoesNotExist)
-		{
-			$this->createSessionRecord($session, $user);
+			// Do not try to touch the database if we can't determine the record state
+			if ($exists === self::$sessionRecordUnknown)
+			{
+				return;
+			}
 
-			return;
+			if ($exists === self::$sessionRecordDoesNotExist)
+			{
+				$this->createSessionRecord($session, $user);
+
+				return;
+			}
 		}
 
 		$this->updateSessionRecord($session, $user);


ceford:
I have tested this item :white_check_mark: successfully on e1c1eaced322bced141ca7d5ef9cbb71f2e32268<br /><br />Confirm: on second page load the SELECT session_id query is not present.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31421">issues.joomla.org/tracker/joomla-cms/31421</a>.</sub>

jwaisner:
I have tested this item :white_check_mark: successfully on e1c1eaced322bced141ca7d5ef9cbb71f2e32268<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31421">issues.joomla.org/tracker/joomla-cms/31421</a>.</sub>

jwaisner:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31421">issues.joomla.org/tracker/joomla-cms/31421</a>.</sub>

Fedik:
I think this need to be reviewed by someone experienced

jfeedback:
This might look like a duplicate query, but from an engineering perspective, the current API is correct and a system engineer from the Joomla team should reject this pull request.

The changes in this pull request introduce a hidden dependency for this metadata manager class to function properly, the assumption that the session handler has promptly written a record to the database before this metadata manager can be properly used.  This is the type of assumption that leads to broken code.

For background, the database session handler in Joomla 3 and earlier have this same fundamental problem in that it assumes that it can only perform UPDATE queries, which means it relies on another part of the system to run the INSERT query and if that doesn't happen for some reason then the session handler will critically fail.  Luckily, someone fixed this fatal flaw in the 4.0 API.

Fedik:
@wilsonge please review

Denitz:
@jfeedback The session is always started before the metadata manager runs the code (because the session ID is used by the manager), hence the data is written.
And 4.0 database session storage does `INSERT` in `write()`, am I wrong?

wilsonge:
The comment from @jfeedback is correct from an architectural perspective. Honestly 99% of the time this code probably is fine - but there are going to be edge cases where you can get issues happening

Issue #31388, opened 2020-11-12
[4.0] Option "Number of columns" in List All Tags doesn't work


drmenzelit:
### Steps to reproduce the issue
Create some tags.
Create a menu item of type "List All Tags".
Play with the option "Number of Columns".

### Expected result
The tags are showed in columns as set in the options.

### Actual result
The tags are all in one column, but depending of the settings there is 1 or more ul lists holding the tags.

Columns: 7 (one ul with 3 li elements)
![grafik](https://user-images.githubusercontent.com/9153168/98932328-7d188d00-24df-11eb-9471-2a02db78321a.png)

Columns: 1 (3 ul with 1 li element each)
![grafik](https://user-images.githubusercontent.com/9153168/98932375-8d306c80-24df-11eb-9e1b-fb448537bed1.png)

adj9:
<b> HTML inspect </b>
![Schermata 2020-11-14 alle 16 40 54](https://user-images.githubusercontent.com/5918444/99151032-a3355d00-2698-11eb-9200-931415d96cd7.png)


In this case the row tab5 is wider. It's Bootstrap problem?

N6REJ:
for w/e reason the columns value is treated as rows.  So when you say 3 columns, it's returning 3 ROWS.  This is being caused by .css
![image](https://user-images.githubusercontent.com/1850089/99373641-42b84280-2887-11eb-9ede-37f1aa54b86f.png)
I'm looking into the best fix.

N6REJ:
layout looks totally different in chrome then in FF.  This is effecting https://github.com/joomla/cassiopeia/issues/209 also.
Started a PR.. https://github.com/joomla/joomla-cms/pull/31422

PR #31422, opened 2020-11-17
[4.0] Change tags "list all" behavior

N6REJ:
Pull Request for Issue # https://github.com/joomla/cassiopeia/issues/209 & #31388 

### Summary of Changes
Changes flow of tags in list all view to columnar to match options.
Supports responsive flow.

### Testing Instructions
If your using the test data sample then there is an "all tags" item otherwise...
Create some tags.
Create a menu item of type "List All Tags".
Play with the option "Number of Columns". ( in menu item )
apply pr
run npm build:css
verify that the number of columns matches the option chosen.
works based on # of columns.  If columns = 0 then its flex flow so everything just flows & sizes automatically, if columns = 1 it's vertical layout, else its columnar layout.

image

Firefox
Chrome
tags

works based on # of columns.  If columns = 0 then its flex flow so everything just flows & sizes automatically, if columns = 1 it's vertical layout, else its columnar layout.


N6REJ:
totally borked in chrome but as above in FF
image

infograf768:
Works fine in FF and Safari.

Screen Shot 2020-11-17 at 11 25 14

In chrome the issue comes from the form control height:
```
.form-control {
    display: block;
    width: 100%;
    height: calc(1.5em + 1.2rem + 2px);  // HERE is the issue.
    padding: .6rem 1rem;
    font-size: 1rem;
    font-weight: 400;
    line-height: 1.5;
    color: #495057;
    background-color: #fff;
    background-clip: padding-box;
    border: 1px solid #ced4da;
    border-radius: .25rem;
    transition: border-color .15s ease-in-out,box-shadow .15s ease-in-out;
}
```
If we comment the height, all is fine.

Screen Shot 2020-11-17 at 11 32 42

N6REJ:
Chrome now renders properly 
image

infograf768:
I have tested this item :white_check_mark: successfully on 072b56cd3348cbf4981823f2eebf9ab1f6323d56<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31422">issues.joomla.org/tracker/joomla-cms/31422</a>.</sub>

brianteeman:
Unless I am reading this wrong this file will generate invalid markup if this condition is not met
`<?php if ($n === 1 || $i === 0 || $bscolumns === 1 || $i % $bscolumns === 0) : ?>`

As we will end up a with a list of `li` without any wrapping `ul`

components/com_tags/tmpl/tags/default_items.php
```diff
@@ -83,6 +83,9 @@ class="inputbox" onchange="document.adminForm.submit();"
 		<?php endif; ?>
 	</form>
 
+	<!-- Fix for chrome. -->
```
sandewt: > `<!-- Fix for chrome. -->`

Fix is not only for `Chrome`.

This fix is also needed for `Firefox` and `MS Edge` on Windows.

Comment could be something like this: `<!-- Clear floated content -->`

N6REJ: Didn't see the problem in ff but did in edge.

N6REJ:
> Unless I am reading this wrong this file will generate invalid markup if this condition is not met
> `<?php if ($n === 1 || $i === 0 || $bscolumns === 1 || $i % $bscolumns === 0) : ?>`
> 
> As we will end up a with a list of `li` without any wrapping `ul`

Correct.  But thats as desired.  We want it to fail because then we're still parsing tag items.  We only want it to succeed on new groups.
$i=0 guarantee's we'll ALWAYS use it at least once.

components/com_tags/tmpl/tags/default_items.php
sandewt: > Didn't see the problem in ff but did in edge.

I have tested in Chrome, Firefox and Edge (Windows).

Explanation:
I have deleted in the patch `<div class="clearfix"></div>`, and looked at what would happen.
For the result in ff, see the image. 
So the fix is not only for Chrome but also for other browsers.

brianteeman:
So why not move the ul out of the for each item loop and then you dont need any of the conditions

N6REJ:
I didn't write the code, however, to answer your question, because each nth item will require a new row ( ul ) because the number of optional columns has been reached.  I know its confusing, I had to spend a few minutes wrapping my head around it.  I probably would've done it differently myself.

brianteeman:
If you are saying that the code generates multiple lists then that is wrong markup - it should be one list

chmst:
Agree. It must be one list to be semantically correct. @N6REJ it is not your fault.

I have another question: The column param - isn't it old fashioned in times of mobile first? 
Why not choose "vertical" or "horizontal" and then let it be flexible?

N6REJ:
ok, I'll examine the code and see how I can make it better... 
@brianteeman they have it as several lists to contain it to n items per row.

N6REJ:
CAN we change this behavior?  Doing so will break b/c as far as I can tell.
image
And if we do do it, how/what changes do we make do the db? @richard67

brianteeman:
We can break b/c but ideally we should aim not to. In this case if J3 also has this incorrect code then you are fixing a bug not breaking b/c

I usually do this in css with `column-count`

if tag_columns = default
    echo the list
else 
    echo inline style 

but @chmst approach is also valid

N6REJ:
ok, so how do we come to a consensus as to the best plan of action.  I put a short survey out on glip but so far no responses.
Talking to alison she wants control on desktop and flow on mobile.  Seems like christianne wants flow period.

I'm open to changing things just want to all get on the same page and do whats best for the project.

N6REJ:
What if we give them an additional parameter to turn columnar sizing on/off.  If off it would fit as many in a row as possible.

N6REJ:
is 
image 

supposed to limit the # of tags?  If so its broken

N6REJ:
@brianteeman @chmst this thing is taking on a life of its own :P
should we merge this one and create another for the changes that will be needed/desired?  I've got a poll out on the fb group right now and its producing some interesting feedback.

N6REJ:
works based on # of columns.  If columns = 0 then its flex flow so everything just flows & sizes automatically, if columns = 1 it's vertical layout, else its columnar layout.

brianteeman:
As long as it is only one list I dont care

N6REJ:
it is

adj9:
I can't apply this PR, with J! Patch Tester
Schermata 2020-11-21 alle 11 51 54

richard67:
@adj9 Testing sample data is not available in a normal installation which has been made with an installation package. It is only available if you are working on a development environment, i.e. a git clone. That's why patchtester can't apply the patch for testing sample data on an installation where this kind of sample data is not included.

richard67:
@N6REJ Drone reposts SCSS code style errors here: [https://ci.joomla.org/joomla/joomla-cms/37681/1/20](https://ci.joomla.org/joomla/joomla-cms/37681/1/20). Please fix. Thanks in advance.

ceford:
Looked at this today - works on Firefox, Chrome and Safari on Mac. But items don't wrap on narrow screens - so if you set 4 columns you get four really narrow columns. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31422">issues.joomla.org/tracker/joomla-cms/31422</a>.</sub>

N6REJ:
@ceford thats by design.  If you set 0 then it will use responsive behaviors.  If you set 4 it assumes you know its 4 columns regardless of screen size.

Quy:
@N6REJ Please fix conflicts. Thanks.

PR #31428, opened 2020-11-18
Update composer packages j/archive, j/string and j-backports/lessphp

HLeithner:
Update 3rd party packages to be php 8 compatible


HLeithner:
Finally fixed the composer issue and updated our php8 framework packages

PR #31442, opened 2020-11-20
Fix "Uncaught ValueError: Path cannot be empty ..." in com_joomlaupdate restore.php

richard67:
Pull Request for Issue #31308 .

### Summary of Changes

Adds a check if the path is empty.

### Testing Instructions

See issue #31308 .

To reproduce, update e.g. a 3.9.22 to 3.9.23 RC.

To test the fix of this PR, update e.g. a 3.9.22 to the update package built by Drone for this PR.

Verify that the update has run completely by checking the Joomla Update log file `administrator/logs/joomla_update.php`. The following logs should be there:
```
2020-11-20T18:39:21+00:00       INFO 192.168.98.1       update  Update started by user Super User (342). Old version is 3.9.22.
2020-11-20T18:39:21+00:00       INFO 192.168.98.1       update  Downloading update file from https://ci.joomla.org/artifacts/joomla/joomla-cms/staging/31442/downloads/37665/Joomla_3.9.23-rc+pr.31442-Release_Candidate-Update_Package.zip.
2020-11-20T18:39:22+00:00       INFO 192.168.98.1       update  File Joomla_3.9.23-rc+pr.31442-Release_Candidate-Update_Package.zip downloaded.
2020-11-20T18:39:22+00:00       INFO 192.168.98.1       update  Starting installation of new version.
2020-11-20T18:39:25+00:00       INFO 192.168.98.1       update  Finalising installation.
2020-11-20T18:39:25+00:00       INFO 192.168.98.1       update  Deleting removed files and folders.
2020-11-20T18:39:26+00:00       INFO 192.168.98.1       update  Cleaning up after installation.
2020-11-20T18:39:26+00:00       INFO 192.168.98.1       update  Update to version 3.9.23-rc+pr.31442 is complete.
```

### Actual result BEFORE applying this Pull Request

See issue #31308 .

### Expected result AFTER applying this Pull Request

No PHP error like described in issue #31308 .

### Documentation Changes Required

None.

PR #29689, opened 2020-06-18
Converted copyright statement to standard format

nibra:
### Summary of Changes

All copyright statements have been changed to the standard format
`© YEAR OWNER <CONTACT>`
as discussed and decided in the Production Department meeting at June 2nd, 2020:
"The 3.10 and 4.x, framework repos. Introduce into 3.10 and merge into 4.0 and then do missing files for 4 beta2. Niels in charge."

### Testing Instructions

No functionality has been touched

### Expected result

Everything works like before

### Actual result

Everything works like before

### Documentation Changes Required

No. However, samples showing the copyright statement *SHOULD* be updated accordingly.

### Scripts used to process the copyright statement

(Also see [nibra/fix-copyright](https://github.com/nibra/fix-copyright) for progress on this)

#### `fix-copyright.sh`

```bash
#!/usr/bin/env zsh

ROOT=${PWD}
GREP_PATTERN="(Copyright )?\(C\) .* Open Source Matters.*All rights reserved\.?"
SED_PATTERN="\(Copyright \)\?(C) .* Open Source Matters.*All rights reserved\.\?"
OWNER="Open Source Matters, Inc."
CONTACT="https://www.joomla.org"

main() {
	echo "Processing files in ${ROOT}"

	COUNT=0

	for FILE in $(grep -rilP "${GREP_PATTERN}" --exclude-dir="vendor" "${ROOT}"); do
		YEAR=$(php fix-copyright.php ${FILE})

		if [[ ${FILE} == *.xml ]]; then
			REPLACEMENT="(C) ${YEAR:-2005} ${OWNER}"
		else
			REPLACEMENT="(C) ${YEAR:-2005} ${OWNER} <${CONTACT}>"
		fi

		COUNT=$((COUNT + 1))

		printf '%5.5s| %7.7s => %s\n' "${COUNT}" "${YEAR:-default}" "${FILE}"

		sed -i -e "s|${SED_PATTERN}|${REPLACEMENT}|" "${FILE}"

	done

	echo "Fixed ${COUNT} copyright notices."
}

main
```

#### `fix-copyright.php`

```php
<?php

class GitHistory
{
	/**
	 * An implementation of file history algorithm with renames detection.
	 * (This comment block is from the IntelliJ source code at https://github.com/JetBrains/intellij-community/blob/ea20241265f9fb956c5a99b1023765aa0e941979/plugins/git4idea/src/git4idea/history/GitFileHistory.java)
	 *
[...]

PR #31447, opened 2020-11-21
Updated installation language files

Bakual:
Installation language files updates coming from the TTs


Bakual:
@HLeithner As usual there will be more files coming over the next days. I will update this PR with any new files.

HLeithner:
Thanks, please create a new PR if we have new translations

installation/language/fr-FR/fr-FR.ini
```diff
@@ -1,5 +1,5 @@
 ; Joomla! Project
-; Copyright (C) 2005 - 2020 Open Source Matters. All rights reserved.
+; (C) 2005 Open Source Matters, Inc. <https://www.joomla.org>
```
PhilETaylor, line 2: Is it ok that the Copyright statements were changed in this commit?

installation/language/fr-FR/fr-FR.xml
```diff
@@ -3,10 +3,10 @@
 	version="3.9"
 	client="installation">
 	<name>French (France)</name>
-	<version>3.9.17</version>
+	<version>3.9.23</version>
 	<creationDate>2020-03-24</creationDate>
 	<author>Joomla.fr</author>
-	<copyright>Copyright (C) 2005 - 2020 Open Source Matters. All rights reserved.</copyright>
+	<copyright>(C) 2005 Open Source Matters, Inc.</copyright>
```
PhilETaylor, line 9: (and here) Is it ok that the Copyright statements were changed in this commit?

HLeithner:
Yes thats a preparation for 3.10

HLeithner:
btw. @PhilETaylor could you do some careful update tests with 3.9.23rc2 in different php versions?

installation/language/fr-FR/fr-FR.ini
Bakual: Yes, because it's the same change that has already been done for 3.10.
See https://github.com/joomla/joomla-cms/blob/8758c0a3d25b2d6a603cd5c94babfcb770dc0c57/installation/language/en-GB/en-GB.ini#L2

I'm not enforcing it for all files in J3.9, but it doesn't hurt if TTs change it already.

PhilETaylor: But why are these changing? Why was that decision made? It goes against everything before it and https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

Bakual: PR was https://github.com/joomla/joomla-cms/pull/29689

PhilETaylor: Wow another huge direction mistake by the production team. Oh well too late now. 

This goes against everything documented about copyright headers in GPL projects.

HLeithner: The reason is that we don't touch all files one the first release in the new year.

PhilETaylor: You mean like literally every other GPL project does... wrong decision. Period. 

Now joomla doesn't follow the guidelines...

Too late now. Just another nail in the wall.

PhilETaylor:
@HLeithner what specifically are you worried might break?

HLeithner:
In rc2 we changed the restore.php https://github.com/joomla/joomla-cms/pull/31442 so I want to be sure it's all ok. I did several tests without problems but that doesn't mean it has no side effect

PhilETaylor:
That makes no sense to test at the moment then, because the Joomla 3.9.22 restore.php will be used to perform the upgrade to Joomla 3.9.23 by extracting the Joomla 3.9.23 zip file ... you will only know if its all broken when using Joomla 3.9.23 restore.php to upgrade to 3.9.24... 

This is the reason com_joomlaupdate was pulled out to be its own extension that could be independently distributed, updated before a Joomla upgrade etc... but it seems lately that was abandoned

HLeithner:
You can upgrade from rc2 to rc2 or 3.10 (don't know if it's possible at the moment) I'm also happy with a code review.

Also you can "update" to all prebuilt PRs

installation/language/fr-FR/fr-FR.xml
infograf768: same reason as the other one, alas...

PR #31456, opened 2020-11-21
[4.0] Refactoring some lib_joomla language strings

Bakual:
Currently, the language file lib_joomla.ini contains several strings which reference a PHP method. There are several issues with that:
* The method name should not be translated, but is part of the language string.
* The method is written in a format `Classname: :Functionname` (with a space between the : :). This isn't how a method is properly written.
* The classname references the old JFoo class names, not the new namespaced ones.
* The same message exists multiple times, just for different classes/functions. This is redundant work for TTs and not reuseable.

### Summary of Changes
I've started with the GET_NAME strings:
JLIB_APPLICATION_ERROR_APPLICATION_GET_NAME="JApplication: :getName() : Can't get or parse class name."
JLIB_APPLICATION_ERROR_CONTROLLER_GET_NAME="JController: :getName() : Can't get or parse class name."
JLIB_APPLICATION_ERROR_MODEL_GET_NAME="JModel: :getName() : Can't get or parse class name."
JLIB_APPLICATION_ERROR_VIEW_GET_NAME="JView: :getName() : Can't get or parse class name."

I have replaced them with a new generic one:
JLIB_APPLICATION_ERROR_GET_NAME="%s: Can't get or parse class name."

I've changed all calling places so the current method name (`__METHOD__`) is passed as a variable.

JLIB_APPLICATION_ERROR_APPLICATION_GET_NAME actually wasn't used anymore.
I also removed the string JLIB_APPLICATION_ERROR_VIEW_GET_NAME_SUBSTRING which also isn't used anymore in J4.

### Testing Instructions
This should work with code review as there actually should be no way with a properly written extension to trigger this error.
If someone knows how to trigger the error, feel free to share.

Error message would be for example
JController: :getName() : Can't get or parse class name.

Error message would be for example
Joomla\CMS\MVC\Controller\FormController::__construct: Can't get or parse class name.

None


Bakual:
If this PR is accepted, I will do more of the same kind with other instances of the same issue.

brianteeman:
Other than thinking it might be better as JLIB_APPLICATION_ERROR_GET_NAME  ie remove the word GENERIC it all seems a good idea to me

Bakual:
You're right, the GENERIC part isn't really needed. I've changed it.

ceford:
I have tested this item :white_check_mark: successfully on 6f37bc4a52a711daaf0a8d00e58a8d62cafa21fe<br /><br />I applied the patch and looked through the code and it looks good. I also inserted a simple call to enqueueMessage and (separately) to throw new exception in the article display HtmlView.php file. That gave the expected behaviour - a warning message and a fatal error message.

Code ```$app->enqueueMessage(Text::sprintf('JLIB_APPLICATION_ERROR_GET_NAME', __METHOD__), 'warning');```

Result ```Joomla\Component\Content\Site\View\Article\HtmlView::display: Can't get or parse class name.```

Code ```		throw new \Exception(Text::sprintf('JLIB_APPLICATION_ERROR_GET_NAME', __METHOD__), 500);```

Result ```500 Joomla\Component\Content\Site\View\Article\HtmlView::display: Can't get or parse class name. ```
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31456">issues.joomla.org/tracker/joomla-cms/31456</a>.</sub>

wilsonge:
LGTM Thanks!

HLeithner:
@wilsonge thats a b/c break

Bakual:
@HLeithner Yes it is, and an allowed one for J4.0.

Issue #31348, opened 2020-11-07
[4.0]Admin view -Article Filter-Tag Filter - Unable to view pre-defined tag or assigned tags to articles as Jooma 3x 


jaisheela:
### Steps to reproduce the issue
Go to Admin view
Move to Content Menu
Go to Articles
Check Article filters
Tag filter shows no tags (No results match) 

### Expected result

It should show the tags created for filtering the article as in it should have predefined tags

### Actual result
 
I have created  8 tag but it says "no results match" 

### System information (as much as possible)

Joomla 4
![notag](https://user-images.githubusercontent.com/16459454/98439217-5e4c7c00-2116-11eb-8dab-71ac51d5b878.png)
![tag](https://user-images.githubusercontent.com/16459454/98439230-7d4b0e00-2116-11eb-8989-3277f5a2ceac.png)
 
MySQL 5.7.32-0ubuntu0.18.04.1

PHP Version | 7.3.12-1+ubuntu18.04.1+deb.sury.org+1

jaisheela:
![Uploading joom3x.png…]()

jaisheela:
![Uploading joom3x.png…]()

infograf768:
Confirmed.
In J3 we have

<img width="418" alt="Screen Shot 2020-11-07 at 18 29 27" src="https://user-images.githubusercontent.com/869724/98447694-3ca7c080-2127-11eb-8d32-1741983a977f.png">

@wilsonge

infograf768:
Setting as release blocker

ghost:
Are the tags shown if you enter 3 letters
 https://github.com/joomla/joomla-cms/issues/30385#issuecomment-674540757 ?

infograf768:
yes, only the tags which fit, very different from J3

SharkyKZ:
Expected behavior as already explained in https://github.com/joomla/joomla-cms/issues/30385 and in many other issues.

richard67:
Expected behaviour regarding how these "fancy selects" work, i.e. it is not a bug what has been observed here.

But that doesn't mean there is no issue. You have to look up the names of tags somewhere else to know which 3 characters could be suitable. Or can you type ANY 3 characters, matching or not, to see the available tags? I think no. It's a functional downgrade compare to "normal "multiselects, and it's bad UX, too.

regularlabs:
This is SOOO user-unfriendly!

[...]

PR #22263, opened 2018-09-19
[4.0] Replace chosen.js to choice.js

Fedik:
### Summary of Changes
This a try to to replace chosen.js to choice.js.

- [x] Module position field 
- [x] Tags field
- [x] Category field
- [x] Filter fields

@wilsonge is any else field is missed in the list?

#### Testing:

Module editing: create custom position, search existing positions
Article editing: test Tag field, test category field
Articles filter: test the filter by multiple category, Tag, etc, 

Everything should work as before the patch.

### Documentation Changes Required
maybe

PR #31481, opened 2020-11-24
[4.0] Prefill tags with most used items

HLeithner:
Pull Request for Issue #31348 .

### Summary of Changes
Pre-fill the tags field with (default) 30 items based on the most used count.
Also reduce the min term length to 1, if there is a performance issue the site owner has to tweak this value.

### Testing Instructions
* Open an article
* Look at the tags field, you will see no tags
* Create 40 tags (or reduce the configuration entry "Initial number of shown tags" in com_tags config to a lower number
* Open an article
* You will see 30 tags
* Create several articles
* Add the always the same 10 tags
* You will always see the 10 tags and 20 additional tags
* Also test non ajax mode

The tags field was empty

You have some (hopefully) tags to select.

N/A

Thanks to @bembelimen for helping


richard67:
@HLeithner PHPCS [https://ci.joomla.org/joomla/joomla-cms/37795/1/6](https://ci.joomla.org/joomla/joomla-cms/37795/1/6)

administrator/language/en-GB/com_tags.ini
```diff
@@ -26,6 +26,7 @@ COM_TAGS_CONFIG_SELECTION_SETTINGS_LABEL="Item Selection"
 COM_TAGS_CONFIG_SHARED_SETTINGS_DESC="These settings apply to all tag layouts unless they are changed for a specific menu item."
 COM_TAGS_CONFIG_SHARED_SETTINGS_LABEL="Shared Layout"
 COM_TAGS_CONFIG_TAG_MIN_LENGTH_LABEL="Minimum Search Length"
+COM_TAGS_CONFIG_TAG_PREFILL_LIMIT_LABEL="Initial number of shown tags"
```
brianteeman: This description does not reflect that the list is the list of most used tags

richard67: And as a label it should be capitalized.

HLeithner: I was hoping that brian makes a suggestion for the label ;-)

wilsonge:
So looks like this blocker was intended. The remoteSearch option seems to have been added as part of the move to the choices.js element https://github.com/joomla/joomla-cms/pull/22263 - first of all I guess it's worth asking whether we should roll back some of that code rather than adding in additional queries

I definitely don't see the need to allow people to select the number of items returned? This seems like our favourite parameter creep

Finally I think reducing the min term length should be done outside the scope of this PR. There's no need to really change that as part of this change

Good work at figuring out this one though!

administrator/language/en-GB/com_tags.ini
brianteeman: You should have asked. But as I agree with @wilsonge that this is not the solution to the problem I wont waste my time

HLeithner:
@wilsonge I updated the PR by reverting the default min length and removed the limit parameter

libraries/src/Form/Field/TagField.php
```diff
@@ -205,16 +198,74 @@ protected function getOptions()
...
+				if ($count > 0)
+				{
+					$query->whereNotIn($db->quoteName('a.id'), ArrayHelper::getColumn($options, 'value'));
+				}
+			}
 		}
-		catch (\RuntimeException $e)
+
+		// Only execute the query if we need more tags not already loaded by the $preQuery query
+		if (!$isRemoteSearch || $prefillLimit > 0)
```
wilsonge, line 256: Isn't this code always going to run because `$prefillLimit` is set to 30 (and even in the previous version of your code if this was a parameter unless that param had been set to 0)

HLeithner: `$prefiilLimit` is set in line 244

wilsonge: Yes and 30 is always greater than 0?

punambaravkar:
I have tested this item :white_check_mark: successfully on aea7162b480a253143deaa7cc3c78a1e0646f51a<br /><br />some tags to select.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31481">issues.joomla.org/tracker/joomla-cms/31481</a>.</sub>

vaibhavsTekdi:
I have tested this item :white_check_mark: successfully on aea7162b480a253143deaa7cc3c78a1e0646f51a<br /><br />Yes, After adding the patch, it works correctly & I can see the previously added tags in the tags filter on the article list view.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31481">issues.joomla.org/tracker/joomla-cms/31481</a>.</sub>

wilsonge:
Finally tested this and it's completely fine. Thanks!

PR #31493, opened 2020-11-26
[4.0] Pre-update check Codestyle

brianteeman:
This PR is only to address codestyle issues resulting from the recent merge by @wilsonge

There are other issues with this code in J4 but they will be addressed later in their own PR


wilsonge:
Thanks!

brianteeman:
thanks - should be easier to see the real stuff to change now

Issue #31506, opened 2020-11-27
J3.9.23 -> Extensions installation fails because of Timeouts


ufuk-avcu:
### Steps to reproduce the issue
Use the newest Joomla 3.9.23
Install for example Yootheme Pro Package. (important: This package includes more Zips within the zip File and this seems to be the problem. If the zip doesn't contains other zips, the installation times are as usual.)

### Expected result
Installation complete after several seconds.
Tested with Joomla 3.9.22 -> 15 seconds for installation.

### Actual result
Installation takes minutes or fail:
Maximum execution time of 60 seconds exceeded in /libraries/vendor/joomla/archive/src/Zip.php on line 301
With higher exexcution Time settings the installation takes 2:05 minutes.

### System information (as much as possible)
Joomla 3.9.23
Tested it with Yootheme Pro 2.3.19 (7,3 MB) -> takes long. Zip in zips.
Tested with oder extensions without zips in zip ist fast and as excepted.

### Conlusion
Joomla 3.9.23 -> 2:05 Minutes
Joomla 3.9.22 -> 0:15 Minutes
-> for the same extension.

nibra:
The [joomla-framework/archive](https://github.com/joomla-framework/archive) package has been updated to use the PHP [ZipArchive](https://www.php.net/manual/en/class.ziparchive.php), because the [zip functions](https://www.php.net/manual/en/ref.zip.php) we used before are deprecated in favor of the Object API.

This is an issue that we inherited from PHP, so we can't do much about it. Nevertheless we'll investigate the possibility of a workaround. Can't promise anything, though.

ufuk-avcu:
Tested it also with Akeeba Backup Core. The main zip contains, 6 other zips. But there are no problems here. The installation speed is more or less the same.

steffans:
This resolves the issue for me, use directly `ZipArchive::getFromIndex` instead of first getting and reading the file with `ZipArchive::getStream`. This should happen with all larger zip files because the read operation takes longer.

```
	/**
	 * Extract a ZIP compressed file to a given path using native php api calls for speed
	 *
	 * @param   string  $archive      Path to ZIP archive to extract
	 * @param   string  $destination  Path to extract archive into
	 *
	 * @return  boolean  True on success
	 *
	 * @since   1.0
	 * @throws  \RuntimeException
	 */
	protected function extractNative($archive, $destination)
	{
		$zip = new \ZipArchive;

		if ($zip->open($archive) !== true)
		{
[...]

PR #31513, opened 2020-11-28
Update of joomla-framework/archive

nibra:
Pull Request for Issue #31506.

### Summary of Changes

For zip files, load the buffer directly instead of utilising the stream.

### Testing Instructions

Install an extension that contains zip files in the zip archive without increasing the max execution time..

You'll get an error like
*Gateway Timeout - The gateway did not receive a timely response from the upstream server or application.*
because of the timeout.

Installation works as it should.

No.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     "require": {
         "php": ">=5.3.10",
         "joomla/application": "~1.9",
-        "joomla/archive": "^1.1.7",
+        "joomla/archive": "dev-master",
         "joomla/data": "~1.2",
         "joomla/di": "~1.5",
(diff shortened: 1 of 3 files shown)


wilsonge (changes requested):
You need to tag a new version of the package. Updating to dev-master isn't acceptable. Also in the 3.x branch you need to commit the updated lock file - which is why tests currently fail

nibra:
New PR: #31514

xillibit:
Why now when i update one specific package with composer it update and install the package joomla/archive from 1.1.7 to 1.1.8 ?

Issue #26595, opened 2019-10-14
[4.0] content history broken by workflows


brianteeman:
### Steps to reproduce the issue
Create an article and save it
Edit the article and change its state from unpublished to published
Use versions to compare the two

### Expected result
Translated row called 
![image](https://user-images.githubusercontent.com/1296369/66784998-e04a8100-eed3-11e9-9854-389f4b8c27c2.png)

### Actual result
![image](https://user-images.githubusercontent.com/1296369/66784878-919ce700-eed3-11e9-81ef-5fdadf9ef6ac.png)

richard67:
Could be caused by #26295 ? If so: I will make a PR anyway for making com_contenthistory compatible with real null dates, but this can’t be merged before PR’s for nulldates for all other components relevant for content history have been merged.

richard67:
I’ll check this tonight German time and make PR if it is the nulldate thing causing this.

brianteeman:
Completely unrelated - this is nothing to do with dates

richard67:
I see ... well, if you discover for com_contenthistory or somewhere else some issues which **are** or **might be** related to dates, let me know.

alikon:
is this still an issue with the workflow v3?

brianteeman:
Yes it is still a problem. 
![menu](https://user-images.githubusercontent.com/1296369/84536332-878eea80-ace5-11ea-947c-c847d801f51d.gif)

brianteeman:
it looks like the change is not being written to the db until you close the article?

bembelimen:
The problem is how the workflow is implemented. The flow is the following:
1. article gets saved
2. workflow adjust the article based on the given plugins (transition)

Now the history is written in step 1, so it does not get the updated values.

The solution would be either to write the history after execution of the workflow or to update the history entry when transition is through.

brianteeman:
@bembelimen the reason isnt that important. What matters is that without the content history the workflow loses a major bit of functionality. Arguably its more importnat with workflows than before

HLeithner:
OK I looked at it, 2 problems one solution.

As benjamin already said the history is saved before the workflow triggers.
Solution for this, we trigger the history save a secondtime after the workflow finished adding a "version note" -> "Workflow"

Second problem, we can't revert a version and shouldn't if the workflow is active, so solution for this. 
Deactivate the revert button (functionality) in the version history if workflow is active.
[...]

PR #31518, opened 2020-11-28
Fix broken versioning in workflow items

bembelimen:
Pull Request for Issue #26595 .

### Summary of Changes
After discussing with @HLeithner we decide to go for the following approach:
when workflow is active, the values from the items which are covered by the workflow plugins (in this case "state" + "featured") should not be put into the versioning table.

This happens by removing this values from the hash + the values itself via the workflow plugins.

### Testing Instructions
- Create an article
- Change it

- Values handled by the workflow are only versioned after a second save

- Values handled by the workflow are never versioned.

### DIFF
--- a/plugins/workflow/featuring/featuring.php
+++ b/plugins/workflow/featuring/featuring.php
@@ -427,7 +427,7 @@ class PlgWorkflowFeaturing extends CMSPlugin implements SubscriberInterface
 	}
 
 	/**
-	 * We remove the publishing field from the versioning
+	 * We remove the featured field from the versioning
 	 *
 	 * @param   EventInterface  $event
 	 *
--- a/plugins/workflow/featuring/featuring.php
+++ b/plugins/workflow/featuring/featuring.php
@@ -11,6 +11,7 @@ defined('_JEXEC') or die;
 
 use Joomla\CMS\Application\CMSApplicationInterface;
 use Joomla\CMS\Event\AbstractEvent;
+use Joomla\CMS\Event\Table\BeforeStoreEvent;
 use Joomla\CMS\Event\View\DisplayEvent;
 use Joomla\CMS\Event\Workflow\WorkflowFunctionalityUsedEvent;
 use Joomla\CMS\Event\Workflow\WorkflowTransitionEvent;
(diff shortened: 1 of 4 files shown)


brianteeman:
I have tested this and it does what it says but for me its a major loss of functionality to just remove the versioning of the state. I do not consider this as a fix for the release blocker.

bembelimen:
Thank you @brianteeman for testing and giving feedback.

Do we agree, that a user should not change values which are handled by the workflow?

The issue is, that the value are not saved at all and the solution should be, that the values are saved but only recovered when the workflow is disabled?

brianteeman:
> Do we agree, that a user should not change values which are handled by the workflow?

honestly not sure. it would depend on the workflow settings

richard67:
Just by the way, the failure of system tests in Drone seems not to be related to the PR but to a problem with testing environment ("Failed to connect to 127.0.0.1 port ...: Connection refused", like we currently have it for other J4 PRs.

bembelimen:
> 
> 
> > Do we agree, that a user should not change values which are handled by the workflow?
> 
> honestly not sure. it would depend on the workflow settings

If you disable the e.g. the publishing plugin, then it's versioned. So the settings have an impact.

brianteeman:
Editor submits a new version. Admin check it and says no so they need to revert to the previous version

can they still do that?

bembelimen:
> 
> 
> Editor submits a new version. Admin check it and says no so they need to revert to the previous version
> 
> can they still do that?

Yes, they can revert every value they could also change in the normal edit form. So e.g. title, text, dates etc. but not "state" and "featured" if (!) they are enabled within the workflow.

plugins/workflow/featuring/featuring.php
```diff
@@ -421,6 +426,82 @@ public function onContentBeforeSave(EventInterface $event)
 		return true;
 	}
 
+	/**
+	 * We remove the publishing field from the versioning
```
richard67: Shouldn't it be "We remove the featuring field from the versioning" here?

bembelimen: Thx

wilsonge:
I'm going to merge this because this is at least more consistent. However I'm not going to close the original issue for now because potentially there could be better fixes. 

> Editor submits a new version. Admin check it and says no so they need to revert to the previous version

In this case if the article is already published the editor version is also live before admin review. If the article isn't live then state won't have changed. If the state has changed between the new version and the admin publishing then you're reverting multiple versions and potentially yes you wouldn't see a rollback. Potentially I guess if there is a path in workflow to revert the workflow state maybe we should execute it (or else not allow the rollback??) But super unsure which is why I'm happy to keep the existing issue open to discuss these kinds of concrete use cases.

@HLeithner also made a good point that right now hitting publish in the list view doesn't trigger versioning (I mean for sure I'd see that as a bug but...) state is already inconsistent in j3/j4 - probably just needs more work overall.

PR #31523, opened 2020-11-29
Fix image selection in media field

bembelimen:
# Summary of Changes

When creating a 3rd party filesystem plugin and selecting an image in a media file selection (e.g. in the article form view), the image will not be displayed, because the variable is wrong.

### Testing Instructions
- Apply patch
- Run npm
- Install a 3rd party media filesystem plugin (like: https://github.com/bembelimen/joomla-cms/tree/4.1/media-manager ) and select an image in a media field.

- Image not selected/displayed

- Image selected/displayed

@dgrammatiko perhaps you can check?

### DIFF
--- a/build/media_source/system/js/fields/joomla-image-select.w-c.es6.js
+++ b/build/media_source/system/js/fields/joomla-image-select.w-c.es6.js
@@ -100,13 +100,8 @@
 
           // eslint-disable-next-line prefer-destructuring
           Joomla.selectedMediaFile.url = resp.data[0].url.split(rootFull)[1];
-          if (resp.data[0].thumb_path) {
-            Joomla.selectedMediaFile.thumb = resp.data[0].thumb_path;
-          } else {
-            Joomla.selectedMediaFile.thumb = false;
-          }
         } else if (resp.data[0].thumb_path) {
-          Joomla.selectedMediaFile.thumb = resp.data[0].thumb_path;
+          Joomla.selectedMediaFile.url = resp.data[0].thumb_path;
         }
       } else {
         Joomla.selectedMediaFile.url = false;


dgrammatiko:
Is `thump_path` always the same as the `url`? If so then remove it in the Media manager Api, but I think that thump is for a thumbnail provided by some cloud storage providers (s3, dropbox)

bembelimen:
> 
> 
> Is `thump_path` always the same as the `url`? If so then remove it in the Media manager Api, but I think that thump is for a thumbnail provided by some cloud storage providers (s3, dropbox)

No, but that is not the fix here. Thumbpath is for [non local adapters](https://github.com/joomla/joomla-cms/pull/31523/files#diff-fbd97535fb6ae6b893ea29264408e2964dc28abcfea9dfb25a2385fca1a6bcb8R103). This fix is the call, when someone either insert an image into the editor (via the image plugin) or uses the image media field (like the intro image). In this case [only the "url" is relevant](https://github.com/joomla/joomla-cms/pull/31523/files#diff-fbd97535fb6ae6b893ea29264408e2964dc28abcfea9dfb25a2385fca1a6bcb8R110). I search the whole Joomla for ```Joomla.selectedMediaFile.thumb``` which is never used, I assume, that is is a leftover or a typo?

So the value filled in has to be ```Joomla.selectedMediaFile.url``` to functional.

dgrammatiko:
> So the value filled in has to be Joomla.selectedMediaFile.url to functional.

Yes, you're right. FWIW the thumb part shouldn't be there but rather in https://github.com/joomla/joomla-cms/blob/acb3b6d66582d6b3467731680b4813536eccf53e/build/media_source/system/js/fields/joomla-field-media.w-c.es6.js#L159-L187 something like:

```js
updatePreview() {
  if (['true', 'static'].indexOf(this.preview) === -1 || this.preview === 'false' || !this.previewElement) {
    return;
  }

  // Reset preview
  if (this.preview) {
    const { value } = this.inputElement;

    if (!value) {
      this.previewElement.innerHTML = '<span class="field-media-preview-icon"></span>';
    } else {
      let thumb = Joomla.selectedMediaFile.url;
      this.previewElement.innerHTML = '';
      const imgPreview = new Image();

      if (Joomla.selectedMediaFile.thumb) {
        thumb = Joomla.selectedMediaFile.thumb;
      }

      const mediaType = {
        image() {
          imgPreview.src = /http/.test(thumb) ? thumb : Joomla.getOptions('system.paths').rootFull + thumb;
          imgPreview.setAttribute('alt', '');
        },
      };

      mediaType[this.type]();

      this.previewElement.style.width = this.previewWidth;
      this.previewElement.appendChild(imgPreview);
    }
  }
}
```

brianteeman:
Is this on hold or should it still be tested?

ghost:
Can a maintainer please set the label "Conflicting Files"?

dgrammatiko:
This should be already fixed by now

HLeithner:

rdeutz:
What should we do with this PR @bembelimen @dgrammatiko

dgrammatiko:
If https://github.com/joomla/joomla-cms/pull/38805 is merged then also local adapters would have a thumb so this PR would break the other one.

PR #31542, opened 2020-11-30
[4.0] Rename sidebar

brianteeman:
This PR renames the module positions in the Cassiopeia template from sidebar-left and -right to sidebar-primary and -secondary

This is because when the language is changed from LTR to RTL the positions are moved and the names -left and -right are no longer applicable.

thanks to @scrabble96 for spotting this


richard67:
Maybe we should provide an update SQL for updating previous J4 Beta versions?

brianteeman:
I thought about it and rejected it as potentially that is changing users data

richard67:
> I thought about it and rejected it as potentially that is changing users data

@brianteeman Then it should be documented and linked in the release announcement.

Scrabble96:
I will test this tomorrow.

infograf768:
-TPL_CASSIOPEIA_POSITION_SIDEBAR-LEFT="Sidebar-left"
-TPL_CASSIOPEIA_POSITION_SIDEBAR-RIGHT="Sidebar-right"
+TPL_CASSIOPEIA_POSITION_SIDEBAR-PRIMARY="Sidebar-primary"
+TPL_CASSIOPEIA_POSITION_SIDEBAR-SECONDARY="Sidebar-secondary"

is imho totally arbitrary and counter intuitive. Why would left be primary and right secondary?...
Looks like this is just splitting hairs for no real gain.
RTL users are not so idiots that they can't understand left and right will be reverted in rtl display.

brianteeman:
Why should they have to think that left is right and right is left. Not surprised that you once again find a spurious reason to object

richard67:
> is imho totally arbitrary and counter intuitive. Why would left be primary and right secondary?...

Am just thinking about it .. maybe `sidebar-1` and `sidebar-2`?

richard67:
@brianteeman I hope closing was not due to my last comment. I only thought about naming, not more.

brianteeman:
@richard67 no not you

PR #31561, opened 2020-12-03
Fix session's moving from Joomla 3 to Joomla 4

wilsonge:
@HLeithner did basically all of the debugging work on this so massive massive credit to him! Both me and @zero-24 had unsuccessfully tried to debug this in the past.

Pull Request for Issue #28465 .

### Summary of Changes
- Double encodes the session name for compat with session names in Joomla 3
- Ensures the code that is stored within a session namespace (concept removed in the framework) continues to function during the lifetime of J4 to allow stable upgrades from J3 (it's worth noting that in the future if users go from Joomla 3 to Joomla 4 and straight a J5 where this migration code has been removed there could be issues - but I'm taking this as another bridge for another day)
- Removes the session migration step that was left over from 3.4.7 session security issues as we require users to go via 3.10 all users will have had their session migrated so this code is useless anyhow.

### Testing Instructions
Test by upgrading from Joomla 3.x to this 4.x repo. Before you'll get logged out during the upgrade. After patch you will stay logged in

None

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -102,10 +102,6 @@ class JoomlaInstallerScript
 		$this->clearStatsCache();
 		$this->convertTablesToUtf8mb4(true);
 		$this->cleanJoomlaCache();
-
-		// VERY IMPORTANT! THIS METHOD SHOULD BE CALLED LAST, SINCE IT COULD
-		// LOGOUT ALL THE USERS
-		$this->flushSessions();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


wilsonge:
Well I didn't mean to merge this - but I managed to forget to swap branches when i started on the next batch on conflicts and git push'd it to 4.0. So I guess this is merged.... At least it get's into nightlies for testing :/

infograf768:
It totally breaks my test site here.

PR #30910, opened 2020-10-03
Add  vertical ordering to blog layouts

chmst:
Pull Request for Issue # .

### Summary of Changes
Proof of Concept 
There are many complaints of users who don't know how to add modifier classes for Blog Layouts. 
This PR is a proof of concept where the main classes for the modifiers are added to the blog layout params. 
I would like to have some feedback and if it is accepted can add it to featured. 

Open questions are, for example:
- Is it self explaining and helpful? 
- Do we need global params for that ?

### Testing Instructions
You need several articles in a category.

Now make a menu item for a category blog. In the Tab Blog Layout There are two new selections: Default classes. 

grafik
One for leading articles (the screenshot), the same for intro articles.

If the user selects "yes", the shown presents new input fields: 

grafik
The same for intro articles

Select here what you want, play with different combinations and see what happens with your blog. For example this: 

grafik

### Actual result BEFORE applying this Pull Request
There is one input field for the classes of the leading articles and one for the intro articles. 

### Expected result AFTER applying this Pull Request
There is one input field for the classes of the leading articles and one for the intro articles. 
There are input fields for default classes.

### Documentation Changes Required
yes

Issue #27478, opened 2020-01-11
[4.0] Blog Layout Ordering


200MPHMEDIA:
### Steps to reproduce the issue

When setting up a Blog Layout, the choice to align by parameters and/or by row or column order is missing in Version 4.  RE: Align by date horizontal vs. Align by Title Vertical.  This option should be restored as selectors vs. "classes."

Why?  How man beginning users 1) know what classes are, and 2) know what class to place where for a desired result.

Can anyone shed light on fixing this issue to restore the options to A) Order by ... and B) Flow by Row or Column options?

ankurshr01:
can I work on this issue?

200MPHMEDIA:
Have no objection to anyone working on the issue.

 

We are not code capable, so appreciate your help.

 

Joomla! 3 had this option for the Blog Layout, yes?

 

If you refer back to 3, you will find a better configuration of choices for the Category Blog Layout.

 

We cannot figure out “why” the option choices were removed.

 

Understand that “classes” may have been added to configure this, but what beginning user knows anything about classes, let along which class is needed for which layout.

 

Replicating the same options on J! 3 seems to be a wiser course of action for Category Blog Layouts.

ReLater:
There were already requests like yours. The decision was made that J4 will not have your requested settings.

It depends on the CSS "framework" that a template provides which CSS classes the user has to use.

Thus, it's the job of template developers to add a documentation to their templates or provide an additional framework plugin that makes the requested settings possible or to provide ready-to-use menu items overrides for this or that.

Joomla 4 core isn't tied as strictly as Joomla 3 to a specific base CSS system (like BS 2).

(I don't say that I liked this decision concerning classes instead of "settings" that could be "translated" to classes in the background according to template requirements. In a scenario where you use different templates the new approach is annoying for users. On the other hand it has some advantages (Maybe. Depends on the used CSS system.)).

200MPHMEDIA:
OK, got all that.  Like you, do not agree with that decision to “code up” layouts, which is instinctive of programmers and code guys.

 

They would rather write 100 lines of code vs. clicking on a backend button.

 

Anyway, for the default template, what are the “classes” to enter (and where to do it), to create some sort of flow options.

 

Appreciate your response.  Thanks.

ReLater:
> Anyway, for the default template, what are the “classes” to enter (and where to do it), to create some sort of flow options.

I don't know if there's already a documentation. Maybe @ciar4n can help.

[...]

PR #18319, opened 2017-10-12
[4.0] Modifier Classes - Category Blog

ciar4n:
Pull Request for Issue #18149 .

### Summary of Changes
This PR allows users to build blog layouts using classes. The intent here is to demonstrate how by using classes we can allow for almost limitless possibilities for the end user and remove the restrictions resulting from some of the current layout options.

In this PR I have created 2 fields (Category Blog Menu Item -> Blog Layout). The first field adds a class to the leading blog items. A second for the remaining blog items. The ‘# Columns’ field has been removed.

This PR also moves com_content/category/blog from bootstrap grid to CSS grid. 

For now I have not touched com_content/featured nor have I created a CSS grid fallback for IE11. Both easily achievable if this idea is favoured.

### Testing Instructions

I have divided classes in to 4 groups. One class from each of these groups can be added, allowing a user to build the layout/style of their choice. Classes are added to the **Blog Class (Leading Items)** or **Blog Class** fields in the **Blog Layout** tab (Category Blog menu item). There is a known issue affecting the admin tabs. This is not related to this PR.

Note that the default layout/styling is 1 column, a top image and no box

**Column Layout**

- columns-2
- columns-3
- columns-4
- masonry-2
- masonry-3
- masonry-4

**Styling**

- boxed

**Image Location**

- image-left
- image-right
- image-bottom

**Image Variation**

- image-alternate 
_alternates the image position on every second item. If an image is set to the left, every second item will display the image on the right. If an image is set to the bottom, every second item will display the image at the top._

Other classes can be easily added but for now the above will allow a good proof of concept.

Following are some example class combinations and their result. Obviously you can experiment here in creating your own. Note that this is a concept so there may be some small issues but nothing that can’t be easily resolved. 

### image-left
image

### boxed image-left image-alternate
image

### boxed columns-3
image

### columns-2 image-left
image

### boxed masonry-3 image-bottom
image

The 2 separate fields **Blog Class (Leading Items)** and **Blog Class** allow you to set alternative classes to the leading items compared to the remaining items...

### Blog Class: boxed image-left image-alternate
image

PR #31570, opened 2020-12-03
[4.0] Upgrade column blog layouts to according CSS classes

Harmageddon:
Pull Request for Issue #27478. This is an alternative approach to #30910. 

### Summary of Changes
Instead of re-adding the parameters that were removed with #18319, this approach searches for the "column" and "ordering" parameters from J3 and converts the values into the according CSS classes, i.e. `columns-X` in the case of horizontal ordering, or `masonry-X` in the case of vertical ordering.

### Testing Instructions

As this PR modifies the update mechanism, you need to test it by upgrading a J3 site to J4. Note that you might not be able to update afterwards to higher J4-beta versions on the same site (don't know how this behaves with PR update packages). So I'd recommend to set up a clean, new installation (or a copy from an existing one) only for this PR.
1. Set up a local 3.10-alpha installation.
2. Create a set of articles, at least 4, in the same category. They should have different text lengths and at least 4 of them should also be featured.
3. For each of the three menu item types "Category Blog", "Featured" and "List All Categories", create several menu items that differ in the parameters "Number of Columns" and "Multi-Column Order". Some should be using only one column, some two or three. Some should use the global order, some horizontal and some vertical.
4. To be able to repeatedly test the upgrade process, you might want to create a backup of files and database at this point.
5. Upgrade your site to J4 using the update package of this PR.
  - At "Components - Joomla! Update", click on "Options". 
  - Update Channel: Custom URL
  - Minimum Stability: Development
  - Custom URL: At the GitHub page of this PR, scroll down to the bottom. Click on "Show all checks". Next to "Download — Prebuilt packages are available for download.", click on "Details". Search for the part "You can test this package as an update by using this custom update server:" and copy the URL from there.

#### How to test the "List All Categories" menu item type

- Create a menu item of type "List All Categories"
  - Select a Top Level Category: Root
  - In the tab "Blog Layout", you can configure the parameters "Number of Columns" and "Multi-Column Order".
- Either create a new category, again with at least 4 articles, that has no own menu item, or unpublish all menu items for the category that you created above in step 2.
- To test, navigate to frontend, click on your "List All Categories" menu item and then click on the title of your category. The category should now be displayed in blog layout according to the parameters you specified above.

("before applying this PR" = using an unpatched update package, such as the normal beta6 or nightly build)

All blog views are rendered in only one column, all with the same order. The above-mentioned parameters are gone without a trace.

The number of columns should stay the same compared to before the upgrade. Depending on whether the according menu item used horizontal or vertical ordering, the articles should now be arranged in a similar manner.

More precisely (with "local" = set in the menu item, "global" = in the configuration of com_content):

<table>
<thead>
<tr><th colspan="4">Before Upgrade (J3)</th><th colspan="2">After Upgrade (J4)</th></tr>
<tr><th>columns (local)</th><th>order (local)</th><th>columns (global)</th><th>order (global)</th><th>CSS class (local)</th><th>CSS class (global)</th></tr>
</thead>
<tbody>
<tr><td>3</td><td>horizontal</td><td>2</td><td>vertical</td><td>columns-3</td><td>masonry-2</td></tr>
<tr><td>3</td><td>vertical</td><td>2</td><td>vertical</td><td>masonry-3</td><td>masonry-2</td></tr>
<tr><td>3</td><td><em>(global)</em></td><td>2</td><td>vertical</td><td>masonry-3</td><td>masonry-2</td></tr>
<tr><td><em>(global)</em></td><td>horizontal</td><td>2</td><td>vertical</td><td>columns-2</td><td>masonry-2</td></tr>
<tr><td><em>(global)</em></td><td><em>(global)</em></td><td>2</td><td>vertical</td><td><em>(global)</em></td><td>masonry-2</td></tr>
<tr><td>3</td><td>horizontal</td><td>1</td><td>vertical</td><td>columns-3</td><td></td></tr>
<tr><td><em>(global)</em></td><td><em>(global)</em></td><td>2</td><td>vertical</td><td>(global)</td><td>masonry-2</td></tr>
<tr><td><em>1</em></td><td><em>(global)</em></td><td>2</td><td>vertical</td><td><strong>(global)</strong></td><td><strong>masonry-2</strong> <sup><a href="#issuecomment-754564227">[1]</a></sup></td></tr>
</tbody>
</table>

Obviously, this table is not exhaustive because listing all possible combinations would take some time and space. So please feel free to come up with your own combinations!

Not sure where to document this. We definitely need to document what classes are available for layouts in the Cassiopeia template.

### Open Points
1. I'm not sure how to handle the "categories" view. In J3, it has parameters to control the blog layouts which are displayed when users click on one of the categories. In J4, this tab exists as well, but I don't see a "blog class" field there. Should we add such a field there? (This is the reason, why this PR is created as a draft. After fixing this, the PR should be ready to test)
2. The masonry layout breaks articles in two if it fits the column layout. This is a different display from J3. And looking at this part, which seems to have no effect at all, I doubt if this behaviour is intended or if it is preferable to avoid breaking articles there. https://github.com/joomla/joomla-cms/blob/8053386a7c9c1c1f1766748aae3c5161662aaf2d/templates/cassiopeia/scss/blocks/_modifiers.scss#L214-L224
  I could add another modifier CSS class that could be used to avoid breaking. In that case, the next question would be whether this class should be included in the conversion from J3 to J4 that is done with this PR. Or we could change this for the masonry layout in general, if breaking inside articles is not intended at all. Comments?

### Screenshots

order | J3 | J4
------------ | ------------- | -------------
horizontal | blog-horizontal-j3 | blog-horizontal-j4
vertical | blog-vertical-j3 | blog-vertical-j4

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -7047,19 +7047,14 @@ class JoomlaInstallerScript
 		$query = $db->getQuery(true)
 			->select(
 				[
-					$db->quoteName('id'),
-					$db->quoteName('link'),
-					$db->quoteName('params'),
+					$db->quoteName('m.id'),
+					$db->quoteName('m.link'),
+					$db->quoteName('m.params'),
 				]
 			)
-			->from($db->quoteName('#__menu'));
-
-		// Restrict selection to com_content items.
-		$query2 = $db->getQuery(true)
-			->select($db->quoteName('extension_id'))
-			->from($db->quoteName('#__extensions'))
-			->where($db->quoteName('element') . ' = ' . $db->quote('com_content'));
-		$query->where($db->quoteName('component_id') . ' = (' . $query2 . ')');
+			->from($db->quoteName('#__menu', 'm'))
+			->leftJoin($db->quoteName('#__extensions', 'e'), $db->quoteName('e.extension_id') . ' = ' . $db->quoteName('m.component_id'))
+			->where($db->quoteName('e.element') . ' = ' . $db->quote('com_content'));
 
 		$menuItems = $db->setQuery($query)->loadAssocList('id');
 
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -17,6 +17,7 @@ use Joomla\CMS\Installer\Installer;
 use Joomla\CMS\Language\Text;
 use Joomla\CMS\Log\Log;
 use Joomla\CMS\Table\Table;
+use Joomla\CMS\Uri\Uri;
 use Joomla\Component\Fields\Administrator\Model\FieldModel;
 use Joomla\Database\ParameterType;
 
@@ -6668,6 +6669,8 @@ class JoomlaInstallerScript
 			}
 		}
 
+		$this->convertBlogLayouts();
+
 		return true;
 	}
 
@@ -7027,6 +7030,85 @@ class JoomlaInstallerScript
 
 			// Execute the query.
 			$db->execute();
+		}
+	}
+
+	/**
+	 * Converts layout parameters for blog / featured views into the according CSS classes.
+	 *
+	 * @return void
+	 *
+	 * @since 4.0.0
+	 */
+	private function convertBlogLayouts()
+	{
+		$db = Factory::getDbo();
+		$query = $db->getQuery(true)
+			->select(
+				[
+					$db->quoteName('id'),
+					$db->quoteName('link'),
+					$db->quoteName('params'),
+				]
+			)
+			->from($db->quoteName('#__menu'));
+
+		// Restrict selection to com_content items.
+		$query2 = $db->getQuery(true)
+			->select($db->quoteName('extension_id'))
+			->from($db->quoteName('#__extensions'))
+			->where($db->quoteName('element') . ' = ' . $db->quote('com_content'));
+		$query->where($db->quoteName('component_id') . ' = (' . $query2 . ')');
+
+		$menuItems = $db->setQuery($query)->loadAssocList('id');
+
+		foreach ($menuItems as $id => $menuItem)
+		{
+			$view = Uri::getInstance($menuItem['link'])->getVar('view');
+
+			if (!in_array($view, ['category', 'categories', 'featured']))
+			{
+				continue;
+			}
+
+			$params = json_decode($menuItem['params'], true);
+
+			if (!isset($params['num_columns']))
+			{
+				continue;
+			}
+
+			$nColumns = (int) $params['num_columns'];
+			unset($params['num_columns']);
+			$order = 0;
+
+			if (isset($params['multi_column_order']))
+			{
+				$order = (int) $params['multi_column_order'];
+				unset($params['multi_column_order']);
+			}
+
+			if ($nColumns > 1)
+			{
+				// Convert to the according CSS class depending on order = "down" or "across".
+				$layout = ($order === 0) ? 'masonry-' : 'columns-';
+
+				if (strpos($params['blog_class'], $layout) === false)
+				{
+					$params['blog_class'] .= ' ' . $layout . $nColumns;
+				}
+			}
+
+			$newParams = json_encode($params);
+
+			$query = $db->getQuery(true)
+				->update($db->quoteName('#__menu'))
+				->set($db->quoteName('params') . ' = :params')
+				->where($db->quoteName('id') . ' = :id')
+				->bind(':params', $newParams, ParameterType::STRING)
+				->bind(':id', $id, ParameterType::INTEGER);
+
+			$db->setQuery($query)->execute();
 		}
 	}
 }
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -9,6 +9,7 @@
 
 defined('_JEXEC') or die;
 
+use Joomla\CMS\Component\ComponentHelper;
 use Joomla\CMS\Extension\ExtensionHelper;
 use Joomla\CMS\Factory;
 use Joomla\CMS\Filesystem\File;
@@ -7078,13 +7079,15 @@ class JoomlaInstallerScript
 				continue;
 			}
 
-			$nColumns = (int) $params['num_columns'];
+			$nColumns = $params['num_columns'] !== '' ? (int) $params['num_columns']
+				: (int) ComponentHelper::getParams('com_content')->get('num_columns', '1');
 			unset($params['num_columns']);
 			$order = 0;
 
 			if (isset($params['multi_column_order']))
 			{
-				$order = (int) $params['multi_column_order'];
+				$order = $params['multi_column_order'] !== '' ? (int) $params['multi_column_order']
+					: (int) ComponentHelper::getParams('com_content')->get('multi_column_order', '0');
 				unset($params['multi_column_order']);
 			}
 
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -19,6 +19,7 @@ use Joomla\CMS\Language\Text;
 use Joomla\CMS\Log\Log;
 use Joomla\CMS\Table\Table;
 use Joomla\CMS\Uri\Uri;
+use Joomla\Component\Config\Administrator\Model\ComponentModel;
 use Joomla\Component\Fields\Administrator\Model\FieldModel;
 use Joomla\Database\ParameterType;
 
@@ -7057,6 +7058,7 @@ class JoomlaInstallerScript
 			->where($db->quoteName('e.element') . ' = ' . $db->quote('com_content'));
 
 		$menuItems = $db->setQuery($query)->loadAssocList('id');
+		$contentParams = ComponentHelper::getParams('com_content');
 
 		foreach ($menuItems as $id => $menuItem)
 		{
@@ -7069,24 +7071,45 @@ class JoomlaInstallerScript
 
 			$params = json_decode($menuItem['params'], true);
 
+			// Don't update parameters if num_columns is unset.
 			if (!isset($params['num_columns']))
 			{
 				continue;
 			}
 
-			$nColumns = $params['num_columns'] !== '' ? (int) $params['num_columns']
-				: (int) ComponentHelper::getParams('com_content')->get('num_columns', '1');
+			$useLocalCols = $params['num_columns'] !== '';
+
+			if ($useLocalCols)
+			{
+				$nColumns = (int) $params['num_columns'];
+			}
+			else
+			{
+				$nColumns = (int) $contentParams->get('num_columns', '1');
+			}
+
 			unset($params['num_columns']);
+
 			$order = 0;
+			$useLocalOrder = false;
 
 			if (isset($params['multi_column_order']))
 			{
-				$order = $params['multi_column_order'] !== '' ? (int) $params['multi_column_order']
-					: (int) ComponentHelper::getParams('com_content')->get('multi_column_order', '0');
+				if ($params['multi_column_order'] !== '')
+				{
+					$useLocalOrder = true;
+					$order = (int) $params['multi_column_order'];
+				}
+				else
+				{
+					$order = (int) $contentParams->get('multi_column_order', '0');
+				}
+
 				unset($params['multi_column_order']);
 			}
 
-			if ($nColumns > 1)
+			// Only add CSS class if columns > 1 and a local value was set for columns or order.
+			if ($nColumns > 1 && ($useLocalOrder || $useLocalCols))
 			{
 				// Convert to the according CSS class depending on order = "down" or "across".
 				$layout = ($order === 0) ? 'masonry-' : 'columns-';
@@ -7107,6 +7130,51 @@ class JoomlaInstallerScript
 				->bind(':id', $id, ParameterType::INTEGER);
 
 			$db->setQuery($query)->execute();
+		}
+
+		// Update global parameters for com_content.
+		$nColumns = $contentParams->get('num_columns');
+
+		if ($nColumns !== null || true)
+		{
+			$nColumns = (int) $nColumns;
+			$order  = (int) $contentParams->get('multi_column_order', '0');
+			$params = $contentParams->toArray();
+
+			if (!isset($params['blog_class']))
+			{
+				$params['blog_class'] = '';
+			}
+
+			// Convert to the according CSS class depending on order = "down" or "across".
+			$layout = ($order === 0) ? 'masonry-' : 'columns-';
+
+			if (strpos($params['blog_class'], $layout) === false && $nColumns > 1)
+			{
+				$params['blog_class'] .= ' ' . $layout . $nColumns;
+			}
+
+			unset($params['num_columns']);
+
+			$app = Factory::getApplication();
+			/** @var ComponentModel $configModel */
+			$configModel = $app->bootComponent('com_config')
+				->getMVCFactory()
+				->createModel('Component', 'Administrator', ['ignore_request' => true]);
+
+			$query = $db->getQuery(true)
+				->select($db->quoteName('extension_id'))
+				->from($db->quoteName('#__extensions'))
+				->where($db->quoteName('element') . ' = ' . $db->quote('com_content'));
+
+			$componentId = $db->setQuery($query)->loadResult();
+
+			$data = array(
+				'id'     => $componentId,
+				'option' => 'com_content',
+				'params' => $params,
+			);
+			$configModel->save($data);
 		}
 	}
 }
--- a/administrator/components/com_content/config.xml
+++ b/administrator/components/com_content/config.xml
@@ -715,6 +715,22 @@
 		description="COM_CONTENT_CONFIG_BLOG_SETTINGS_DESC"
 		>
 		<field
+			name="blog_class_leading"
+			type="text"
+			label="JGLOBAL_BLOG_CLASS_LEADING"
+			size="3"
+			validate="CssIdentifier"
+		/>
+
+		<field
+			name="blog_class"
+			type="text"
+			label="JGLOBAL_BLOG_CLASS"
+			size="3"
+			validate="CssIdentifier"
+		/>
+
+		<field
 			name="num_leading_articles"
 			type="number"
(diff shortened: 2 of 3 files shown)


chmst:
@richard67 please have a look here.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

brianteeman:
> The masonry layout breaks articles in two if it fits the column layout. 

I am assuming this is referring to article 4 in the screenshot for cassiopeia vertical ?

I agree that the scss implies the intention was to prevent the column break. If you are saying it doesn't work then wont simply fixing it prevent the need for any extra classes

Harmageddon:
> I am assuming this is referring to article 4 in the screenshot for cassiopeia vertical ?

Yes, exactly.

> I agree that the scss implies the intention was to prevent the column break. If you are saying it doesn't work then wont simply fixing it prevent the need for any extra classes

I guess so. Maybe @ciar4n or somebody who is more involved with the Cassiopeia template can tell whether the current behaviour is intended or not? I don't want to break anything just because I assume it's a bug.

brianteeman:
Are you using firefox?

Harmageddon:
> Are you using firefox?

Ohh. Yes I am. And in Chromium, the breaking is indeed avoided. Good point! 

So this seems to be some weird Firefox thing that it doesn't respect `break-inside: avoid` for elements that have a `display: flex`. Too tired now, but I'll do some research tomorrow if this is a known bug or behaviour.

brianteeman:
It is a known bug which is why I asked ;)

See https://www.smashingmagazine.com/2019/02/css-fragmentation/

Harmageddon:
> It is a known bug which is why I asked ;)
> 
> See https://www.smashingmagazine.com/2019/02/css-fragmentation/

Hmm, if I'm not missing anything, that article says that Firefox has issues with `break-avoid: column`, `break-after` and many other things. But that it should support `break-inside: avoid` / `page-break-inside: avoid`, which we are using here.
I filed a bug for Firefox, because I couldn't find any information on this behaviour: https://bugzilla.mozilla.org/show_bug.cgi?id=1680720
So for the moment, I'd treat this as a separate browser issue that we might need to keep in mind, but don't need to fix with this PR.

Harmageddon:
For the "categories" view, I now chose the simplest option, that is, adding the "blog class leading" and "blog class" field to the "Blog Layout" tab of this view. 
One question I'm asking myself right now: We have these two fields which include the "useglobal" flag, but we don't have a global option to set the blog class for intro or leading articles. Should we add this to the config of com_content? If we don't, this PR is going to replace global options for the number of columns or the order with local options, i.e. a class specified in the respective menu item.

administrator/components/com_admin/script.php
```diff
@@ -7029,4 +7033,85 @@ private function updateContentTypes(): void
...
+			)
+			->from($db->quoteName('#__menu'));
+
+		// Restrict selection to com_content items.
+		$query2 = $db->getQuery(true)
+			->select($db->quoteName('extension_id'))
+			->from($db->quoteName('#__extensions'))
+			->where($db->quoteName('element') . ' = ' . $db->quote('com_content'));
+		$query->where($db->quoteName('component_id') . ' = (' . $query2 . ')');
+
```
chmst, line 7059: Maybe a join is easier to read.

richard67: And a join possibly has a better performance than a subquery.

chmst:
Looks really good. It resolves the migration problem in an elegant way without needs for extra params in J4. 

About the categories view: 
I don't remember that categories are displayed in columns in j3 and think that these params in j3 were not used at all and can be ignored.

ChristineWk:
> About the categories view:
> I don't remember that categories are displayed in columns in j3 and think that these params in j3 were not used at all and can be ignored.

Checking something on a 3.10-0-alpha4-dev:

a) Articles Options Blog/Featured Layouts > Multi Column Order: down or across.

b) Menu Item Typ: Category List > Choose a Category > there are  no columns.

c) Here I'm not sure:
Menu Item Typ: Featured Articles > Show all featured articles from one or multiple categories in a single or multi column layout.
But it's the same as a)

administrator/components/com_admin/script.php
Harmageddon: Good point, I didn't think of this. Replaced it with a join. Thank you! :-)

Harmageddon:
> About the categories view:
> I don't remember that categories are displayed in columns in j3 and think that these params in j3 were not used at all and can be ignored.

No, they are used indeed. I'm going to update the test instructions with this.
If you have a "List all categories" view, you can choose how a single of these categories should be displayed when you click on one of them (given that the category doesn't have its own menu item). Quoting the hint on that parameter tab:
> These options are also used when you select one of the category links, on the first page and/or thereafter, unless they are changed for a specific menu item.

This includes the columns parameters. So if you select to display two columns there, this does not apply directly to the "List all categories" view, but to category views accessible from there.

@ChristineWk This PR treats cases a) and c) (plus d) the menu item type "List all categories", as explained above).

> b) Menu Item Typ: Category List > Choose a Category > there are no columns.

Correct, the category list doesn't include columns and thus, needs no changes here.

N6REJ:
Where are you getting masonry from?  in bs4 its card-columns. https://getbootstrap.com/docs/4.5/components/card/#card-columns

chmst:
At the moment it is bootstrap4.  I would prefer a pure css solution, but this not necessary in scope of this pr

Harmageddon:
> Where are you getting masonry from? in bs4 its card-columns. https://getbootstrap.com/docs/4.5/components/card/#card-columns

Masonry classes were introduced in #18319. But yes, card-columns might be an alternative.

adj9:
I have done the various steps from J! 3.10, but then didn't apply the PR and upgraded to J! 4-dev with Joomla! Update.
After the update I have this screen.
Schermata 2020-12-12 alle 12 26 24

The content has remained the same. About the modules you have to reposition them, but it's part of the upgrade steps J3->J4.

With PR enabled, I can't find the "Number of Columns" and "Order of Multiple Columns" parameters in "Blog Layout".
Screan: 
Schermata 2020-12-12 alle 12 51 25


Is npm necessary?

ChristineWk:
> 
> 
> I have done the various steps from J! 3.10, but then didn't apply the PR and upgraded to J! 4-dev with Joomla! Update.
> After the update I have this screen.
> Schermata 2020-12-12 alle 12 26 24
> 
(?) https://github.com/joomla/joomla-cms/issues/28465
https://github.com/joomla/joomla-cms/pull/31561

Harmageddon:
Thank you for testing!

>  After the update I have this screen.

That's a known issue, see comment by @ChristineWk. Not sure if it's fixed by #31561, but we can ignore it for this PR. 

In order to update based on this PR here, you need to use a custom update package. I added some details to the testing instructions. If it's not clear what to do, feel free to ask! :-)

> With PR enabled, I can't find the "Number of Columns" and "Order of Multiple Columns" parameters in "Blog Layout".

This PR does not re-introduce the parameters, because in #18319, the choice was made to remove these parameters and use CSS classes instead. This PR adds a step to the update to 4.0 (which is why you need to use the update package of this PR instead of applying the PR after the update). There, the parameters from J3 are converted to the according CSS classes for J4. 
So after you updated with an update package from this PR, the field "Blog Class" in your screenshot should contain something, and in frontend, the layout should look similar to that from J3.

> Is npm necessary?

No, not for this PR.

adj9:
- Reset del sito, base 3.9.24-dev
- Upgrade to 3.10
- Create 4 articles in a category
- Create a Blog category
- Upgrade to 4  from URL custom of PR

The position of the items remains and is correct.
Schermata 2020-12-12 alle 16 58 24


Blog Layout of the BackEnd is this
Schermata 2020-12-12 alle 16 58 46

I did not understand how you act on the parameters  "Number of Columns" and "Multi-Column Order".

adj9:
Sorry

The parameter Blog Class works.

adj9:
I have tested this item :white_check_mark: successfully on 905d5092f9578dbd43379e76881cff9b6f71f86c<br /><br />With a prev steps, the par Blog Class works. :)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on 905d5092f9578dbd43379e76881cff9b6f71f86c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

ChristineWk:
@Harmageddon Could you please wait with merging, as I'm trying to start the update from J 3.10.0.alpha4-dev to J 4.0.0 (with the update package) Thks. I hope it will work :-)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

ChristineWk:
a) Update was successful:

screen shot 2020-12-13 at 15 31 25::

b) Your site has been updated. Your Joomla version is now ‎4.0.0-beta6-dev+pr.31570

c) but: It's broken.
An error has occurred.
0 Call to undefined method Joomla\CMS\Application\AdministratorApplication::isAdmin() 

Then I made some checks etc:

There are tables not up to date!  > Then Repair button > Update Structure > OK now.

Then I checked via "System" > Templates
Administrator: Atum
Site: Cassiopeia and Protostar (Copy)

Maintenance: Global Check in of:
categories_table
extensions_table
modules_table
Result: 219 items check in.

Templates Style: Changed Cassiopeia to default.

System Dashboard = OK
Home Dashboard = Broken (see above response).
Frontend: No Cassiopeia<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

ChristineWk:
Update from my experiment :-)

- Home Dashboard was broken (see response previous comment)
Solution: 1 Plugin I had to remove. In addition: Some JCE Plugins.

- Frontend: No Cassiopea, even it was set under Style to cassiopea as default.
Solution: I had to change it on each menu separately. Reason: It was a copy of previous Protostar.

Module: Main Menu (Top Menu) changed from position-1 to: menu!

I had created (previous Joomla 3.10) 2 menus as featured (top menu), 2 menus (bottom menu) as blog & 1 menu List all Categories.

Joomla 4.0.0 now: as above: looks OK. I also tried 1 menu with: masonry-2 > it works.



<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

administrator/components/com_admin/script.php
```diff
@@ -7029,4 +7033,80 @@ private function updateContentTypes(): void
...
+			$view = Uri::getInstance($menuItem['link'])->getVar('view');
+
+			if (!in_array($view, ['category', 'categories', 'featured']))
+			{
+				continue;
+			}
+
+			$params = json_decode($menuItem['params'], true);
+
+			if (!isset($params['num_columns']))
```
chmst, line 7075: Stop also if num_columns <=1 ?

chmst:
About the Open Points: My suggestion
1) Adding both input fields into the categories layout is correct.
2) We add these classes to config.xml 
3) Question is, which class to use during migration for a vertical "down" ordering. A class "column-x" means that the articles appear "across".

- masonry-x (bootstrap v4) means ordering "down" but has some disadvantages 
- card-column is the correct and newer implementation for masonry but does not say how many columns.
- something other like columns-down-x or card-columns-x

I think that users would understand best the masonry-x. Otherwise we must explain our own class. We can use this now as is or override the whole in our scss / layouts.

Would be nice to have mor meanings about that.

richard67:
As this PR here is still in draft status, and the alternative PR #30910 has been closed, should the issue #27478 be re-opened and the newer duplicate issue #31801 by the same author be closed? Or vice versa?

@Harmageddon Any plans about when you will mark the PR as ready for review?

ChristineWk:
@richard67 
My personal opinion is: #27478 should not be re-openend and #31801 should be closed, as both refers to this Draft.

administrator/components/com_admin/script.php
N6REJ: to do so I set the .xml field filter to not allow below 0 in my pr.

Harmageddon:
> @Harmageddon Any plans about when you will mark the PR as ready for review?

I planned to allow for a discussion here regarding the open points (global parameter for Blog Class and the things @chmst pointed out). But if everyone implicitly agrees in silence, I'm going to add the global parameter, resolve the review above, and mark it as ready for review in the next few days. ;-)

chmst:
@Harmageddon yes, please do so. I will be happy to test it and think that it is the best solution we can get.

administrator/components/com_admin/script.php
Harmageddon: @chmst Indeed, below in the code, the CSS classes are only added if `$nColumns > 1`. The rest however (including removing the num_columns and multi_column_order parameter values) is run if `num_columns` is set, even in the case <= 1. Are there cases where this could lead to problems?

@N6REJ I don't understand your comment. Which PR are you talking about?

Harmageddon:
I added the global parameters for the blog class fields, along with a table in the testing instructions above that indicates which setting should lead to which result.
There is one inconsistency that I found (see last row of the table): If in J3, we set the global number of columns to 2, and the local number of columns for one menu item to 1, the update script now uses "columns-2"/"masonry-2" as the global option, but leaves the local class field empty, as one column would be equivalent to not adding any CSS class. However, if this field is empty, the global value of "columns-2"/"masonry-2" is used. 
The only solution I see right now would be to use some nonsense class like "columns-1" in the case of a menu item having only one column specified. Does this make sense?

ChristineWk:
> 
> 
> I added the global parameters for the blog class fields, along with a table in the testing instructions above that indicates which setting should lead to which result.

Thats works.  Thank you!

> The only solution I see right now would be to use some nonsense class like "columns-1" in the case of a menu item having only one column specified. Does this make sense?

Hmm. Tested something:

Menus Edit Item: Blog Layout it shows: 
Use Global masonry-2 (per default)

a) if you want to delete the: masonry-2. It‘s (was) not possible, it will be changed to: "Use Global (masonry-2)" and becomes inactive.

b) To change that menu to 1 column, you have to enter (overwrite) with e.g: column-1. Thats works.

All other menus remains unchanged (= still showing e.g. masonry-2 Layout) = OK.

administrator/components/com_admin/script.php
chmst: @HarmageddonI have no use-case where it could make problems. So think it is ok

richard67:
@Harmageddon Could you remove the sentence about draft status at the beginning of the testing instructions? Thanks in advance.

administrator/components/com_admin/script.php
Harmageddon: Okay, thanks! :-)

Harmageddon:
@richard67 Oh, sorry. Done! Thank you for the reminder. :-)

drmenzelit:
I have tested this item :white_check_mark: successfully on ef5468872557c26e367a82ce43b7d4ff335d35ad<br /><br />All columns parameters I had set up in J 3.10 were modified in the corresponding class (columns-x or masonry-x) after update to J4<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

ChristineWk:
Thanks @drmenzelit for testing.

In principle, the test was also successful for me, but I waited for a comment, because of the topic of column-1, as written above. Ev. to enter manually: column-1 (if desired) was just a suggestion & shouldn't be a problem.

drmenzelit:
@ChristineWk sorry, I missed your comment...

drmenzelit:
> a) if you want to delete the: masonry-2. It‘s (was) not possible, it will be changed to: "Use Global (masonry-2)" and becomes inactive.
> 
> b) To change that menu to 1 column, you have to enter (overwrite) with e.g: column-1. Thats works.

a) That is the same behaviour as in columns parameters in J3, if you don't set a specific number of columns in the menu item, the global configuration will be taken
b) Related to a), I think it is ok to have to change manually to columns-1 if you want to have 1 column ( you have to do the same in J3)

ChristineWk:
Thanks Viviana for your clarification/help

> b) Related to a), I think it is ok to have to change manually to columns-1 if you want to have 1 column ( you have to do the same in J3)

That's what I meant.

ChristineWk:
I have tested this item :white_check_mark: successfully on ef5468872557c26e367a82ce43b7d4ff335d35ad<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31570">issues.joomla.org/tracker/joomla-cms/31570</a>.</sub>

Harmageddon:
@ChristineWk @drmenzelit Thank you for testing! :-)
I think I agree with you: People who want this very specific setting of having two columns globally, and one column for a single menu item just have to use the workaround with "column-1", or need to remove the global class and set the "column-2" locally. This should be okay, if we document the available classes somewhere.

wilsonge:
This makes sense to me. Nice job - thanks!

Issue #26750, opened 2019-10-21
[4.0] Alternative Solution for media modal adapter in url


wilsonge:
### Steps to reproduce the issue
https://github.com/joomla/joomla-cms/pull/21189 solved the issue with inserting some images by hardcoding the adapter path. Whilst that fixed an immediate issue it's absolutely not good for when we support images hosted in the cloud.

### Expected result
No need to hardcode adapter meaning we can't use 3rd party adapters

### Actual result
Hardcoded adapter.

brianteeman:
As stated in that PR it also means that we can't type a url in a field url

wilsonge:
There isn't a URL field in j4. So yes - but I've no clue if that's going to continue working anyhow - needs a separate issue

Fedik:
I wonder why do we have this fancy adapters, instead of use a stream wrapper PHP feature?

https://www.php.net/manual/en/class.streamwrapper.php
https://www.hashbangcode.com/article/php-custom-stream-wrappers

dgrammatiko:
@wilsonge please update the description to reflect that a similar approach solution be also added in tinyMCE: https://github.com/joomla/joomla-cms/blob/59a15a035f7f7e7baf6bc014bf4ea6bde24ee83a/plugins/editors/tinymce/tinymce.php#L526

bembelimen:
In my opinion, there are three options:
1. we have to store always 2 values:
- the path to the displayed file (like external provider have their own URL)
- the internal path (with local adapter etc.)
Problem: all outputs will break (3rd party)
2. we calculate the internal path based on the display path
Problem: could be impossible as some provider do not have this mapping or the display path is completely different to the internal path, so each file would have to be checked.
3. Remove ```&amp;path=local-0:/' . $folder```
Problem: if you press "select" you will always start from the root and not in the folder from the selected file.

I would prefere the 3rd option

dgrammatiko:
There is a 4th option: do something similar to the lazyloading approach, eg attach the adapter as a url param

BUT essentially the problem is the media field itself:
- it stores the data as a string although a rich format (json) will be more appropriate
- although it implies that all media types are covered truth is that in reality it is an image field

bembelimen:
> 
> 
> There is a 4th option: do something similar to the lazyloading approach, eg attach the adapter as a url param

Yes, but then you have to attach the adapter + the internal URL, otherwise you're running into the 2. solution and have to recalculate it (which is not always possible).

> 
[...]

PR #31599, opened 2020-12-06
[4.0] Fix local adapter, keep fixed ordering, fix default value

Fedik:
Pull Request for Issue from comment https://github.com/joomla/joomla-cms/issues/26750#issuecomment-739246507 .

### Summary of Changes

Do not allow User to change ordering in the plugin parameters.
local-0:/, local-1:/ ... local-x:/ always should stay in original order.

I more like idea with named prefix `<provider>-<account>`, as @dgrammatiko  suggested in https://github.com/joomla/joomla-cms/issues/26750#issuecomment-739248621.
However this will not work, I not checked in deep but it seems that whole thing hardcoded to use `<account>` as integer.

Also I have fixed default value.

### Testing Instructions

Apply path.
Go to Plugins: FileSystem - Local.
Add couple Directories, and try change their order.

You can change ordering

You cannot change ordering

none

### DIFF
--- a/plugins/filesystem/local/local.xml
+++ b/plugins/filesystem/local/local.xml
@@ -29,8 +29,8 @@
 					label="PLG_FILESYSTEM_LOCAL_DIRECTORIES_LABEL"
 					multiple="true"
 					layout="joomla.form.field.subform.repeatable-table"
-					buttons="add,remove,move"
-					default='{"directories0":{"directory":"images"}}'
+					buttons="add,remove"
+					default='[{"directory":"images"}]'
 				>
 					<form>
 						<field
--- a/plugins/filesystem/local/local.php
+++ b/plugins/filesystem/local/local.php
@@ -79,14 +79,13 @@ class PlgFileSystemLocal extends CMSPlugin implements ProviderInterface
 	public function getAdapters()
 	{
 		$adapters = [];
-		$directories = $this->params->get('directories', '[{"directory":{"directory": "images"}}]');
+		$directories = $this->params->get('directories', '[{"directory": "images"}]');
 
 		// Do a check if default settings are not saved by user
 		// If not initialize them manually
 		if (is_string($directories))
 		{
 			$directories = json_decode($directories);
-			list($directories) = $directories;
 		}
 
 		foreach ($directories as $directoryEntity)


dgrammatiko:
@laoneo can you have a look at this one?

laoneo:
I didn't test, but looks fine for me...

adj9:
I have created folders with PR applied, but they remain fixed. 
I don't know how to vary the order.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31599">issues.joomla.org/tracker/joomla-cms/31599</a>.</sub>

plugins/filesystem/local/local.xml
```diff
@@ -29,8 +29,8 @@
 					label="PLG_FILESYSTEM_LOCAL_DIRECTORIES_LABEL"
 					multiple="true"
 					layout="joomla.form.field.subform.repeatable-table"
-					buttons="add,remove,move"
```
bembelimen, line 32: I see what you did here, but I'm not sure, if this is needed, because one can "destroy" the value mapping by just selecting another folder or removes it and create new (to change ordering). So I think this change is unnecessary.

Fedik: Yes and no,
with editing of the value, it kind of obvious that the thing will be broken,
however, with reordering User may expect that the thing will work as before, but it will not.

User do not know how thing works in detail (and should not).

bembelimen: Ok, the argument works for me 👍

ceford:
I have tested this item :white_check_mark: successfully on f53a54fed7e3a72abc09ebfb3ebeb89a472f7885<br /><br />I can confirm the pull request removes the ordering thingy. But:

Can someone explain why this matters and how this plugin is supposed to be used? The Help screen is rubbish - I will fix it when I know more.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31599">issues.joomla.org/tracker/joomla-cms/31599</a>.</sub>

Fedik:
> Can someone explain why this matters 

It a bit tricky, as whole Adapters thing :smile: 
There a bit https://github.com/joomla/joomla-cms/issues/26750#issuecomment-739246507
User can add some own folders here (many of them), that the system will see like 0, 1, 2 ... n  (it is actual a Row indexes). If User change ordering, then association "Row Index" => "Row value"  also will change, that will break previous stored values.

> how this plugin is supposed to be used?

That a good question :wink: 
It something for Media manager. I not searched, but I hope there a documentation.

ghost:
I have tested this item :white_check_mark: successfully on f53a54fed7e3a72abc09ebfb3ebeb89a472f7885<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31599">issues.joomla.org/tracker/joomla-cms/31599</a>.</sub>

ceford:
I have changed the Media Help screen to read as follows:

> LOCAL Panel
> 
> This area shows the structure of the local folders of your site, by default the images folder. 
> If you wish to keep images and documents in separate local folders: first create a folder in 
> the root of your site, say files, then go to the Filesystem - Local plugin and add files under 
> Select directories. Back in the Media component you will see images and files as separate 
> items in the Local panel.
> 

Which means that the Media Options descriptors are incorrect.

It will take a while for the change to pass through the translation system.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31599">issues.joomla.org/tracker/joomla-cms/31599</a>.</sub>

wilsonge:
Thanks!

PR #31575, opened 2020-12-03
Add Joomla 3 session namespace to has function

HLeithner:
Pull Request for login screen on joomla 4 update. Complete the last fix for Joomla update.

### Summary of Changes
Check also the joomla 3 namespace "__default" in the session `has` function.

### Testing Instructions
update from j3.10

### Actual result BEFORE applying this Pull Request
Shows the upgrade login screen on upgrade

### Expected result AFTER applying this Pull Request
No upgrade login screen

PR #31601, opened 2020-12-06
[4.0] Regression and b/c fix for session handling

HLeithner:
Pull Request for Issue noticed by @nikosdion in https://github.com/akeeba/fof/commit/d69b179bbefcc03b4dabd5e82a04d9c0d092b2ca

The Joomla 4 session manager doesn't handle namespaces anymore. For j3 b/c joomla includes a b/c layer which was doesn't handle existing sessions correctly. In PR #31561 we tried to fix the upgrade double login issue. This PR got merged too early and a bugfix PR #31575 fixed a part of the problem. The session PRs introduced a new bug handling none "default" namespace which is addressed in this PR.

### Summary of Changes
Try the prefixed namespace used in session migrated from j3 after the none prefixed parameter in the following functions:

* `Joomla\CMS\Session\Session::get()`
* `Joomla\CMS\Session\Session::has()`
* `Joomla\CMS\Session\Session::clean()`

We don't change the `Joomla\CMS\Session\Session::set()` function because new parameters will always be stored in a none prefixed namespace.

### Testing Instructions
* Upgrade from J3.10 to j4, this should still work
* Save values to the session with and without namespace

```php
$session = JFactory::getApplication()->getSession();
# Clear session
$session->clear('__ns');
$session->clear('ns');

echo "Set wow without namespace ns\n";
$session->set('wow', 'xxx');
echo "Check wow without namespace\n";
if ($session->has('wow')) {
  echo "Good\n";
} else {
  echo "FAIL\n";
}
var_dump($session->get('wow', 'xxx'));
$session->clear('wow');

echo "Set wow with namespace ns and j3 syntax \n";
$session->set('wow', 'xxx', 'ns');
echo "check wow with j3 syntax \n";
if ($session->has('wow','ns')) {
  echo "Good\n";
} else {
  echo "FAIL\n";
}
var_dump($session->get('wow', 'not found', 'ns'));
if ($session->has('ns.wow')) {
  echo "Good\n";
} else {
  echo "FAIL\n";
}
var_dump($session->get('ns.wow', 'not found'));
$session->clear('wow', 'ns');
$session->clear('ns');

echo "Set __ns.wow for j3 session simulation \n";
$session->set('__ns.wow', 'xxx');
echo "check wow with j3 syntax \n";
if ($session->has('wow','ns')) {
  echo "Good\n";
} else {
  echo "FAIL\n";
}
var_dump($session->get('wow', 'not found', 'ns'));
echo "check wow with j4 syntax and underscore prefixed (normally not used) \n";
if ($session->has('__ns.wow')) {
  echo "Good\n";
} else {
  echo "FAIL\n";
}
var_dump($session->get('__ns.wow', 'not found'));
echo "check wow with j4 syntax\n";
if ($session->has('ns.wow')) {
  echo "Good\n";
} else {
  echo "FAIL\n";
}
var_dump($session->get('ns.wow', 'not found'));
```

Working with the session using the old syntax and names doesn't work

Still all works including a better b/c layer for old syntax and namespaces

N/A

### DIFF
--- a/libraries/src/Session/Session.php
+++ b/libraries/src/Session/Session.php
@@ -194,24 +194,36 @@ class Session extends BaseSession
 					'deprecated'
 				);
 
-				$name = '__' . $args[2] . '.' . $name;
+				$name       = $args[2] . '.' . $name;
+				$legacyName = '__' . $name;
 			}
+		}
+
+		if (parent::has($name))
+		{
+			// Parent is used because of b/c, can be changed in Joomla 5
+			return parent::get($name, $default);
+		}
+
+		/*
+		 * B/C for retrieving sessions that originated in Joomla 3.
+		 * A namespace before Joomla 4 has a prefix of 2 underscores (__).
+		 * This is no longer the case in Joomla 4 and will be converted
+		 * when saving new values in `self::set()`
+		 */
+		if (isset($legacyName) && parent::has($legacyName))
+		{
+			return parent::get($legacyName, $default);
 		}
 
 		// More b/c for retrieving sessions that originated in Joomla 3. This will be removed in Joomla 5
 		// as no sessions should have this format anymore!
-		if (parent::has($name))
-		{
-			return parent::get($name, $default);
-		}
-		elseif (parent::has('__default.' . $name))
+		if (parent::has('__default.' . $name))
 		{
 			return parent::get('__default.' . $name, $default);
 		}
-		else
-		{
-			return $default;
-		}
+
+		return $default;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


libraries/src/Session/Session.php
```diff
@@ -194,24 +194,35 @@ public function get($name, $default = null)
 					'deprecated'
 				);
 
-				$name = '__' . $args[2] . '.' . $name;
+				$name       = $args[2] . '.' . $name;
```
Quy:
```suggestion
				$name = $args[2] . '.' . $name;
```

libraries/src/Session/Session.php
```diff
@@ -272,7 +283,7 @@ public function has($name)
 					'deprecated'
 				);
 
-				$name = $args[1] . '.' . $name;
+				$name       = $args[1] . '.' . $name;
```
Quy:
```suggestion
				$name = $args[1] . '.' . $name;
```

HLeithner:
thx @Quy fixed cs

wilsonge:
Rather than this - all namespaces in our b/c mapping just need to be given a prefix by default. also we should try the get fallback if a namespace has been provided.

wilsonge:
Thanks!

Issue #29215, opened 2020-05-25
[4.0] Remaining mails to be converted to mail templates


richard67:
### Steps to reproduce the issue

This is just a reminder on the remaining mail types to be converted to mail templates after PR #28722 has been merged.

### Expected result

All kinds of mails sent by the core are using mail templates.

### Actual result

Done:
- [x] Testmail in global configuration (done with PR #22126 )
- [x] Contact form in com_contact (done with PR #28722 )
- [x] com_contact API controller to send a mail (done with PR #28722 )
- [x] Joomla Updatenotification (done with PR #28722 )
- [x] Massmailer mail (done with PR #28722 )
- [x] Mail to the new user when creating a new user in the backend (done with PR #28722 )
- [x] Password reset for users in frontend (done with PR #28722 )
- [x] Username reminder for users in frontend (done with PR #28722 )

To be done:
- [ ] Notification mail from com_actionlogs (administrator\components\com_actionlogs\src\Model\ActionlogModel.php line 174)
- [ ] com_messages notification (administrator\components\com_messages\src\Model\MessageModel.php line 435)
- [ ] com_privacy send export to user (administrator\components\com_privacy\src\Model\ExportModel.php line 240)
- [ ] com_privacy notification about a new request (administrator\components\com_privacy\src\Model\RequestModel.php line 362)
- [ ] com_privacy creating a new request (components\com_privacy\src\Model\RequestModel.php line 188/190 ATTENTION!! HERE IS A BUG! The mail is send twice!)
- [ ] reminder of expiring privacy consent (plugins\system\privacyconsent\privacyconsent.php line 604)
- [ ] User registration (components\com_users\src\Model\RegistrationModel.php line 198, 254, 650, 711)

richard67:
I'll try to do at least some of the remaining mail templates 'cause I know now how it has to be done.

richard67:
Closing as having a pull-request which implements all missing mail templates listed in the description. Thanks @Hackwar for that work. Unfortunately I won't be able to test due to health problems (nothing serious, only a bit painful). When I'm fit again and it's still necessary I'll test of course.

PR #31621, opened 2020-12-09
[4.0] Converting additional mails to mail templates

Hackwar:
Pull Request for Issue #29215.

### Summary of Changes
This converts the mails of the actionlogs component, the privacy component and the privacy consent plugin to the mail template system.

* [X]  Notification mail from com_actionlogs (administrator\components\com_actionlogs\src\Model\ActionlogModel.php line 174)
* [X]  com_messages notification (administrator\components\com_messages\src\Model\MessageModel.php line 435)
* [X]  com_privacy send export to user (administrator\components\com_privacy\src\Model\ExportModel.php line 240)
* [X]  com_privacy notification about a new request (administrator\components\com_privacy\src\Model\RequestModel.php line 362)
* [X]  com_privacy creating a new request (components\com_privacy\src\Model\RequestModel.php line 188/190 ATTENTION!! HERE IS A BUG! The mail is send twice!)
* [X]  reminder of expiring privacy consent (plugins\system\privacyconsent\privacyconsent.php line 604)
* [X]  User registration (components\com_users\src\Model\RegistrationModel.php line 198, 254, 650, 711)

The mentioned bug has also been fixed.

### Testing Instructions
You have to trigger the feature to send these mails to test this.
- Go to your user profile in the backend and enable to receive actionlog updates. You should get a mail when you login.
- Create a request for export or removal in the frontend privacy component to trigger the frontend mails.
- Create a request for export or removal in the backend privacy component to trigger the backend mails.
- Send the export of data in the backend via mail to trigger that one...
- Enable renewable consents, give consent and then edit it in the database to be close to expiration. That should trigger that last mail.
- Register users with all the different settings we have.


brianteeman:
Dont you need to delete this file if it is no longer being used ?
`administrator\components\com_actionlogs\layouts\logstable.php`

brianteeman:
You've removed the instructions to translators
`You can use the following merge codes for all COM_PRIVACY_EMAIL strings:`

But shouldnt that be replaced with a note to translators about the placeholders in the new strings?

richard67:
@Hackwar I see 2 new update SQL scripts `4.0.0-2020-12-08.sql`, but I can't see any changes in `supports.sql` for new installations in this PR. Forgotten?

Hackwar:
@brianteeman I deleted the file. Regarding the placeholders: Since those aren't fixed anymore and can be extended at will, I don't think the notes are helpfull enough to keep them. I understand your point, but I think that will create more confusion than be helpfull. If you disagree hard on that, I can put those back in, but I would prefer to leave that out.

@richard67 Fixed.

Hackwar:
Ok, now all mails have been converted to templates. I don't want to have to do that again...

I encountered one issue: When you are in the mail template component, you can filter by extension. When you do that, you get only a reduced number of strings translated in that list, since we only load the extension translation files for which we currently have templates displayed. I'm not yet sure where to put code to load those files every time...

Hackwar:
BTW: That makes 25 mail templates so far in the system.

richard67:
Adding the release blocker label as inherited from the issue #29215 .

wilsonge:
I'm merging this without tests so it's in the beta for this evening

PR #31635, opened 2020-12-11
[4.0] A faster gzipping npm command

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Remove the current script for gzipping .min.css && .min.js files
- Remove the zopfli dependency

side effect
- Place fontawesome in the dependencies (not in the dev dependencies)
- Remove path dependency (path is a core module)

### Testing Instructions
Git clone the Repo to a temp folder somewhere in your HDD
run `composer install`
Run `npm install` (set a stopwatch to benchmark the process)
Then run `npm run gzip`

Git clone the Repo to another temp folder somewhere in your HDD
Checkout this PR
Run `composer install`
Run `npm install` 
Then run `npm run gzip` (set a stopwatch to benchmark the process)

Compare the 2 benchmarks
Compare the 2 folders

Gzipping the assets is a very long task 

Gzipping the assets is snappy

@wilsonge @HLeithner I know this might seems like I selflessly plugin my code here but there are couple benefits here:
- dropping dependencies
- The code has 0 dependencies
- The code supports Brotli (just add `-b` at the end of each command)
- You can still fork the code if you need to modify it
- Feel free to close this if it breaks any rules about code ownership, etc


HLeithner:
If I understand you package right it simply uses zlib, so we could simply add the lines adding the compression to our build.js? I mean at the moment your script get loaded 3 times from github (maybe cached) but looks like a waste of resources. 

Or I'm wrong?

dgrammatiko:
> your script get loaded 3 times from github (maybe cached) but looks like a waste of resources.

You could combine the 3 commands to 1. Also since this is done with npx there is no real dependency added (mind that the command is only executed by the release leader, it's a one off execution just before the release of a package)
But yes you could fork the code in the project if you like. (having it as a node package is better imo as 3rd PD can use it as well for their own packages 😉 )

wilsonge:
I’m down - although is there a reason not to have you package as a dev dependency? Just so it’s version locked with everything else

dgrammatiko:
> although is there a reason not to have you package as a dev dependency

The idea is to reduce the dependencies that need to be installed (both for CI and local development). Maybe another approach is to create a repo for the npm tools and fork the code there and also transfer the rest of the build tools there as well (yeah I know breaking things to different repos makes it more difficult to maintain but then again those tools should be available to all the 3rd PD, right now this is true only through the 4.0-dev repo). Hard for me to justify what fits better the project but I guess having one line npm commands for any dev should be handy

brianteeman:
Unable to test as the command `npm run gzip` does nothing on my install

dgrammatiko:
@brianteeman interesting, this was already tested on windows, can you try running`cd media` and then `npx --no-install github:dgrammatiko/compress` ? do you get any message? Were any .gz files created?

brianteeman:
```
`C:\htdocs\joomla-cms>npm run gzip 

> joomla@4.0.0 gzip C:\htdocs\joomla-cms
> node -e 'require("./build/build-modules-js/gzip-assets.es6.js").gzipFiles()'

```

npx --no-install github:dgrammatiko/compress`  works perfectly

`

dgrammatiko:
I see, somehow you're executing the old command `> node -e 'require("./build/build-modules-js/gzip-assets.es6.js").gzipFiles()'`, this was before the changes in this PR. I guess it has to do with the conflict?

brianteeman:
Yes I am trying to do the first part of the test instructions **before** applying the pr

Sorry that wasn't clear

dgrammatiko:
> Yes I am trying to do the first part of the test instructions before applying the pr

Maybe my instructions are bad, you need to clone and check out the `4.0-dev` repo for the first step

brianteeman:
thats what I am using - the clone that I do all my dev on

brianteeman:
Just to confirm I did a completely new clone and the output of `npm run gzip` is exactly the same as before

dgrammatiko:
>` node -e 'require("./build/build-modules-js/gzip-assets.es6.js").gzipFiles()`

This probably won't run on windows, can you change `./build/build-modules-js/gzip-assets.es6.js` to `C:\htdocs\joomla-cms\build\build-modules-js\gzip-assets.es6.js`

brianteeman:
I could if I could find the file to change :(

dgrammatiko:
It's the package.json but you should be able to run the command directly in the terminal: 
`node -e 'require("C:\htdocs\joomla-cms\build\build-modules-js\gzip-assets.es6.js").gzipFiles()'`

brianteeman:
I had already tried that but just double double checked
no change

Maybe its me, maybe its windows. Perhaps another windows user can confirm

brianteeman:
found the bug
changing the code in package.json from
`    "gzip": "node -e 'require(\"./build/build-modules-js/gzip-assets.es6.js\").gzipFiles()'",`

to 

`    "gzip": "node -e require('./build/build-modules-js/gzip-assets.es6.js').gzipFiles()",`

note the change from "\ to ' and the removal of the ' around require

Issue #31657, opened 2020-12-15
Form Field Type Note


venomdev:
### Steps to reproduce the issue
Create a form field of type note

### Expected result
A valid html div block with a label and description

### Actual result
a closing div tag before the opening div tag

### System information (as much as possible)
Here's the line in question:
https://github.com/joomla/joomla-cms/blob/staging/libraries/joomla/form/fields/note.php#L60

### Additional comments
Also should we wrap the description inside some p tags?

PR #31658, opened 2020-12-15
JFormFieldNote fix

venomdev:
* Move closing div tag to end of HTML output
* Add p tags to the field description

Pull Request for Issue #31657

### DIFF
--- a/libraries/joomla/form/fields/note.php
+++ b/libraries/joomla/form/fields/note.php
@@ -55,9 +55,9 @@ class JFormFieldNote extends JFormField
 		}
 
 		$html[] = !empty($title) ? '<' . $heading . '>' . JText::_($title) . '</' . $heading . '>' : '';
-		$html[] = !empty($description) ? JText::_($description) : '';
+		$html[] = !empty($description) ? '<p>'. JText::_($description). '</p>' : '';
 
-		return '</div><div ' . $class . '>' . implode('', $html);
+		return '<div ' . $class . '>' . implode('', $html). '</div>';
 	}
 
 	/**


ghost:
@venomdev you can close #31657. It don´t need an issue if a pr is created.

venomdev:
No worries, I created the issue first and decided to fix it myself.

brianteeman:
Thanks for your fisrt contributuion. Unfortunately it doesnt work. When tested on the note field in the tinymce editor plugin we can see the problem.

### Before this PR
image
`
### After this PR
image

venomdev:
Wouldn't that mean the tiny MCE has an error?

The field is now valid html

Fedik:
I cannot reproduce your issue, with note field I have markup:
```html
<div class="control-group">
  <div class="control-label">
  </div><div ><h4>test</h4></div>
  <div class="controls">
  </div>
</div>
```

The closing div inside the filed is hacky way to make the note full width: close **previous** **div**, and **reopen** it **after** the note again.
Changes in this PR break it.

venomdev:
> The closing div inside the filed is hacky way to make the note full width: close **previous** **div**, and **reopen** it **after** the note again.

That makes sense for how it currently works. I'm using these fields on the front end in a twig template and they aren't wrapped in the usual divs for admin pages. Looks like I'll need to create a custom note form field.

Something to consider for an update on this field, would it be possible to add an attribute that defaults to the current output and when the attribute is set switches to the stand alone html?

Currently I'm thinking something like

html=strict (standard valid html)

Or

html=wrapped (set by default)

Fedik:
Do not need any extra attributes, there already `layout` attribute, that allows to use a custom markup.
However the Note field does not support it currently. The solution would be is to make support a custom layout in this field.

venomdev:
@Fedik Ok, yeah a layout file looks the way to go, I didn't know these existed for fields.

I'll work on that when I get a chance, I still needed to create a custom field for now.

PR #31660, opened 2020-12-15
[4.0] Refactor JInstaller strings

Bakual:
Currently, the language file lib_joomla.ini contains several strings which basically have the same error message, but for the different types of extensions.
Many of the strings already aren't used anymore since a generic "extension install failed..." type of the string exists.

### Summary of Changes
* Removed the unused strings.
* Changed the ones still used to the more generic "extension" string
* Used the variable for the type of installation (install, uninstall, update, discover_install) which allowed to reduce the string amount as well

### Testing Instructions
You would need to have an extension package which is broken (eg files missing) or a server with wrong folder permissions (not able to write) to get errors 
Imho it should be fine to do a code review as I'm not changing any functionality itself.

Error messages are sometimes specific to the type of extension (template, file, library, component, plugin, module, package)

Error messages are more generic.


installation/language/en-GB/joomla.ini
```diff
@@ -231,7 +231,7 @@ JLIB_FORM_FIELD_INVALID="Invalid field:&#160;"
 JLIB_FORM_VALIDATE_FIELD_INVALID="Invalid field: %s"
 JLIB_FORM_VALIDATE_FIELD_REQUIRED="Field required: %s"
 JLIB_INSTALLER_ABORT="Aborting language installation: %s"
-JLIB_INSTALLER_ABORT_PACK_INSTALL_CREATE_DIRECTORY="Package Install: Failed to create folder: %s."
+JLIB_INSTALLER_ABORT_CREATE_DIRECTORY="Extension %1$s: Failed to create folder: %2$s"
```
brianteeman, line 234:
```suggestion
JLIB_INSTALLER_ABORT_CREATE_DIRECTORY="Extension %1$s: Failed to create folder: %2$s."
```

Bakual: Thanks, fixed it.

brianteeman: This is in the installation. Is it really an extension that failed to create the folder. Surely the only thing it could be here was a language pack.  Same for line 235.

??

richard67:
@Bakual The phpcs failure in Drone is not related to your PR, it comes from the 4.0-dev branch having had code style errors when you had created your branch based on it. But if you want to get rid of the Drone failure, update your branch to latest 4.0-dev, where the errors meanwhile have been corrected.

Bakual:
Rebased PR on latest 4.0-dev

wilsonge:
Makes lots of sense to me!

installation/language/en-GB/joomla.ini
Bakual: I actually haven't checked the usage of the string this deeply. I just checked if the string is used at all.
Keep in mind that technically, a language pack is just an extension as well.

PR #31668, opened 2020-12-15
[4.0] en-GB style guide issues

brianteeman:
Updates e.g. to eg as documented in the style guide plus one other string spotted at the same time that was not correct grammar


chmst:
Just because I am curious. 
I have learnt e.g. at school (long time ago) - but never would recognize eg (looks like a misspelled egg) as e.g.
Why is this better?

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31668">issues.joomla.org/tracker/joomla-cms/31668</a>.</sub>

brianteeman:
It looks odd to me as well - an age thing perhaps

> Why is this better?

It is based on advice from the UK The Government Digital Service style guide which covers style, spelling and grammar conventions for all content published on GOV.UK

wilsonge:
Looks odd to me too :)

Issue #31671, opened 2020-12-15
Passwords in configuration can't be unset


HLeithner:
Another problem which was oversight by the JSST is that we can not unset a password.

_Originally posted by @HLeithner in https://github.com/joomla/joomla-cms/issues/31515#issuecomment-735371752_

brianteeman:
Yes you can. You edit the configuration.php file

HLeithner:
Closed with or #31672

Issue #31452, opened 2020-11-21
Database Settings - Password DB


Razzo1987:
### Steps to reproduce the issue
Go to the Backend > Global Configuration > Server > Database Settings
Try to use a different DB (Example: a DB of a empty Joomla installation).
You can't specify the DB password (in FTP configuration you can)

### Expected result
![image](https://user-images.githubusercontent.com/906604/99882311-bf4c7780-2c1f-11eb-9a19-5457d2ac0ae8.png)

### Actual result
![image](https://user-images.githubusercontent.com/906604/99882316-ca070c80-2c1f-11eb-88f5-21f684a4a5f5.png)

### Additional comments
On save Joomla must check if the new connection parameters work correctly (like in the installation setup) and the save the info.
The issue is for J3 and J4.

I think without the password the other DB fields is pointless.

ne4lh:
joomla 4 added this code
administator/components/com_config/forms/application.xml

 <field
     name="db_password"
     type="password"
     label="COM_CONFIG_FIELD_DATABASE_PASSWORD_LABEL"
     required="true"
     validate="password"
     force="on"
     filter="raw"
  />

admin language file
COM_CONFIG_FIELD_DATABASE_PASSWORD_LABEL="Database Password"
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31452">issues.joomla.org/tracker/joomla-cms/31452</a>.</sub>

ne4lh:
[j4.txt](https://issues.joomla.org/uploads/1/17ca05627038a122f28e0836ec5f177c.txt)

did post the first time<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31452">issues.joomla.org/tracker/joomla-cms/31452</a>.</sub>

HLeithner:
Closing because we have a pr #31672

PR #31672, opened 2020-12-15
[3] Allow to set password empty passwords and add database password back

HLeithner:
Pull Request for Issue #31671 and #31452 .

### Summary of Changes
Add an attribute `lock` to the password field which does the following 3 things:

* set the `disabled` attribute to the password field (this means the field will not be send to the server)
* add a button with the label "Modify" to the password field (this toggles the disable attribute of the password field)
* removes the value from the password field

Also the database password field is back, since we have a database connection checker we can be safe that we don't break the installation on save.

### Testing Instructions
* Change change passwords like database password, smtp password, ftp password and others in the global configurations.
* set the password to an empty value
* Test other password fields like user passwords

* Changing passwords work
* Setting an empty password doesn't work
* Other Password fields at the global configuration work

* Changing passwords work
* *Setting an empty password will set the password to an empty value now*
* Other Password fields at the global configuration still work

https://docs.joomla.org/Password_form_field_type


administrator/language/en-GB/en-GB.com_config.ini
```diff
@@ -58,6 +58,8 @@ COM_CONFIG_FIELD_DATABASE_HOST_DESC="The hostname for your database entered duri
 COM_CONFIG_FIELD_DATABASE_HOST_LABEL="Host"
 COM_CONFIG_FIELD_DATABASE_NAME_DESC="The name for your database entered during the installation process. Do not edit this field unless absolutely necessary (eg the transfer of the database to a new hosting provider)."
 COM_CONFIG_FIELD_DATABASE_NAME_LABEL="Database Name"
+COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC="The password for access to your database entered during the installation process. Do not edit this field unless absolutely necessary (eg the transfer of the database to a new hosting provider)."
```
brianteeman: > The password for access to your database entered during the installation process

Obviously that makes sense the first time someone reads it. But once they have changed it once it is no longer a true statement

layouts/joomla/form/field/password.php
```diff
@@ -42,6 +42,7 @@
  * @var   array    $options         Options available for this field.
  * @var   array    $inputType       Options available for this field.
  * @var   string   $accept          File types that are accepted.
+ * @var   boolean  $lock            File types that are accepted.
```
brianteeman: file types ??

brianteeman:
The parts I have tested (global config db password and user password) work correctly

I do not understand why the lock/modify button is present

There must be a clear feedback that the button has been clicked and the input is now disabled. The subtle background change on the input is not sufficient

### Locked
image

### Unlocked
image

HLeithner:
Don't know how to make it more clear because that's how a disabled field looks like

layouts/joomla/form/field/password.php
HLeithner: Thanks will fix it

administrator/language/en-GB/en-GB.com_config.ini
HLeithner: It's copy paste from the username so this would be wrong too... Can you make a suggestion please?

brianteeman: COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC="The password for access to your database. Do not edit this field unless absolutely necessary (eg after the transfer of the database to a new hosting provider)."

brianteeman:
> Don't know how to make it more clear because that's how a disabled field looks like

By changing the button. 

Look at the password login field where it changes from an open to a closed eye - although in this case I would try something like using a padlock in open and closed state and then also toggle the text between "protected" and "modify" or something like that

brianteeman:
I assume that you realise that as soon as you change the database you will be immediately logged out as the default sessionhandler is database

HLeithner:
> I assume that you realise that as soon as you change the database you will be immediately logged out as the default sessionhandler is database

Yes but that's true for all database settings.

brianteeman:
Was just checking. Personally that's a great reason for all of these database fields to be readonly as I cant think of a real world scenario where I have a working site that I will ever want to change the database of. In the extremely rare scenario that someone might come up with you can always edit the configuration.php.

HLeithner:
I use them regularly while doing joomla updates and I see no problem doing this also because we have a protection against miss configuration.

On the other side you can can break your site with many other settings in the global configuration. But at the moment I want to fix the unset problem so I can build the rc....

layouts/joomla/form/field/password.php
```diff
@@ -80,8 +96,13 @@
...
+<?php if ($lock): ?><span class="input-append"><?php endif; ?>
 <input type="password" name="<?php
 echo $name; ?>" id="<?php
 echo $id; ?>" value="<?php
 echo htmlspecialchars($value, ENT_COMPAT, 'UTF-8'); ?>" <?php
 echo implode(' ', $attributes); ?> />
+<?php if ($lock): ?>
+<button type="button" id="<?php echo $id; ?>_lock" class="btn btn-info" data-toggle="button"><?php echo JText::_('JMODIFY'); ?></button>
+</span>
+<?php endif; ?>
```
brianteeman:
```suggestion
<?php if ($lock): ?>
    <span class="input-append">
<?php endif; ?>

<input 
    type="password" 
    name="<?php echo $name; ?>" 
    id="<?php echo $id; ?>" 
    value="<?php echo htmlspecialchars($value, ENT_COMPAT, 'UTF-8'); ?>" 
    <?php echo implode(' ', $attributes); ?>
/>
<?php if ($lock): ?>
    <button type="button" id="<?php echo $id; ?>_lock" class="btn btn-info" data-toggle="button"><?php echo JText::_('JMODIFY'); ?></button>
    </span>
<?php endif; ?>
```

wilsonge:
I think this works well for the database field.

My only comment would be that I think the fields that are hidden by showon (e.g. ftp/proxy) need to be in edit mode by default when first enabled somehow :/ Because otherwise they look like they contain values when they are actually empty and will probably confuse our users. Example field with no value

Screenshot 2020-12-16 at 00 23 33

Especially as we're using the string modify which kinda implies something already exists.

HLeithner:
> My only comment would be that I think the fields that are hidden by showon (e.g. ftp/proxy) need to be in edit mode by default when first enabled somehow :/ Because otherwise they look like they contain values when they are actually empty and will probably confuse our users. Example field with no value
> 
> Screenshot 2020-12-16 at 00 23 33
> 
> Especially as we're using the string modify which kinda implies something already exists.

I would remove the hint completely

wilsonge:
Also I think if you enter a default value for the password field then currently the value is set into the session and hitting save subsequently as the password field is disabled constantly uses the value from the session until you either hit cancel or remodify it. However it's not clear from the UI (obviously) that you're using the session value as opposed to the original value of the field because the field switches back to the disabled mode on the page reload (maybe some value in session storage or something?)

HLeithner:
Now it looks like this in protected mode
image
and like this in unprotected mode
image

HLeithner:
as alternative a one-way "modify" button without return is possible too

wilsonge:
> as alternative a one-way "modify" button without return is possible too

I think this would make more sense to me - although no sure if the absence of the button would constitute valid a11y - @brianteeman any feedback on that? Combined with automatic unlock if the field has been submitted with an invalid value preventing save (i.e. the password value is stuck in the session as opposed to the configuration.php value)

brianteeman:
Sorry I'm not commenting on accessibility. Ask @carcam and the accessibility team. It's a big enough team that I'm sure one of them could find the time to respond.

SniperSister:
> I would remove the hint completely

Can't we add the hint conditionally? If a non-empty value is set, add the hint; if the value is empty, leave the field blank. That way we have a clear indicator for the user.

HLeithner:
@wilsonge 
> > as alternative a one-way "modify" button without return is possible too
> 
> I think this would make more sense to me - although no sure if the absence of the button would constitute valid a11y - brianteeman any feedback on that? Combined with automatic unlock if the field has been submitted with an invalid value preventing save (i.e. the password value is stuck in the session as opposed to the configuration.php value)

I see no clean way to address this, yes the password is saved in the session but will not be set in the field again because it's replaced by an empty string.

@SniperSister 
> > I would remove the hint completely
> 
> Can't we add the hint conditionally? If a non-empty value is set, add the hint; if the value is empty, leave the field blank. That way we have a clear indicator for the user.

I renamed the button to "Previous" after pressing "Modify" and added the hint back with the count of characters of the original password. Also if you press "Previous" the password field is set to an empty string again and the hint (placeholder) is shown again.

chmst:
I have tested the functions and it works well. 

my2cent for the buttons: I suggest "edit" (if it is disabled) and "protect" (when enabled) instead of "modify" and "protected".

For a11y the  button could have a title-attribute "open the input field for edit"  and "protect the input field" for example.

HLeithner:
> my2cent for the buttons: I suggest "edit" (if it is disabled) and "protect" (when enabled) instead of "modify" and "protected".

"edit" is fine for me, I changed "protected" to "Previous" could be "cancel" too

> For a11y the button could have a title-attribute "open the input field for edit" and "protect the input field" for example.

This is a title correct? So a description like "Allow to change the password" and "Cancel the password change" or so is too much?

administrator/language/en-GB/en-GB.com_config.ini
```diff
@@ -58,6 +58,8 @@ COM_CONFIG_FIELD_DATABASE_HOST_DESC="The hostname for your database entered duri
 COM_CONFIG_FIELD_DATABASE_HOST_LABEL="Host"
 COM_CONFIG_FIELD_DATABASE_NAME_DESC="The name for your database entered during the installation process. Do not edit this field unless absolutely necessary (eg the transfer of the database to a new hosting provider)."
 COM_CONFIG_FIELD_DATABASE_NAME_LABEL="Database Name"
+COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC="The password for access to your database. Do not edit this field unless absolutely necessary (eg after the transfer of the database to a new hosting provider)"
```
Quy:
```suggestion
COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC="The password for access to your database. Do not edit this field unless absolutely necessary (eg after the transfer of the database to a new hosting provider)."
```

wilsonge:
I have tested this item :white_check_mark: successfully on d3698cd7becae9b43056126ffe32db41e21b11ae<br /><br />I think this is good enough to ship. Definitely can improve it further. But fixes the "regression" in functionality for the next release. I'd consider at least being able to change SMTP credentials between local dev and prod sites essential (when I'm restoring via akeeba or whatever).<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31672">issues.joomla.org/tracker/joomla-cms/31672</a>.</sub>

webgras:
I have tested this item :white_check_mark: successfully on d3698cd7becae9b43056126ffe32db41e21b11ae<br /><br />I tested password fields (FTP, SMTP and DB in global configuration - could not find others) to set a new one, change it and empty it.
I tested also the normal user registration in frontend and creating a user in the backend.
I changed passwords successfully in user management and in account details.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31672">issues.joomla.org/tracker/joomla-cms/31672</a>.</sub>

PhilETaylor:
Ping. you might want to fix this bug... https://github.com/joomla/joomla-cms/issues/31716

HLeithner:
I already see you report and try to reproduce it and of course I will fix it

PR #31674, opened 2020-12-16
[4.0] Cleanup and refactor JLIB_DATABASE_ERROR_... strings

Bakual:
The `lib_joomla.ini` contains a lot of JLIB_DATABASE_ERROR strings which are nowhere used in code. That is just unneeded translator work.
There are also some strings where the class was injected, but the method was part of the string. The method itself shouldn't be translated. This PR adds a new generic string where the method and error message is injected.
In the nested table class, there is a function which requires an error message, but never uses it. I've changed the method so the argument is optional and removed the error messages. The strings used wouldn't work as error message anyway since they expect a parameter.

### Summary of Changes
* The unused strings are removed from the INI files.
* `JLIB_DATABASE_ERROR_MOVE_FAILED` and `JLIB_DATABASE_ERROR_DELETE_FAILED` and  are changed to the new generic `JLIB_DATABASE_ERROR_METHOD_FAILED`
* The function `_runQuery` in libraries/src/Table/Nested.php was changed to have an optional error message instead of required. The messages strings are removed.
* As a bonus, the language class now uses the correct (existing) string for when the SEF code is already present.

### Testing Instructions
As with the other PRs, it's hard to actually test this. Code review should do.
You can try to see if the strings are indeed not used in the CMS.

None


Bakual:
Statistic: 38 unused strings are removed, two are replaced by a new one. Not that bad I think 😃

api/language/en-GB/joomla.ini
```diff
@@ -881,15 +881,9 @@ JHELP_USERS_USER_MANAGER_EDIT="Users:_Edit_Profile"
...
 JHELP_USERS_USER_NOTES_EDIT="User_Notes:_New_or_Edit"
 
-; If there is an error connecting database before initialisation, en-GB.lib_joomla.ini can't be loaded
-; we therefore have to load the strings from en-GB.ini
-
-JLIB_DATABASE_ERROR_ADAPTER_MYSQL="The MySQL adapter 'mysql' is not available."
-JLIB_DATABASE_ERROR_ADAPTER_MYSQLI="The MySQL adapter 'mysqli' is not available."
-JLIB_DATABASE_ERROR_CONNECT_DATABASE="Unable to connect to the Database: %s."
-JLIB_DATABASE_ERROR_CONNECT_MYSQL="Could not connect to MySQL."
+; If there is an error connecting database before initialisation, lib_joomla.ini can't be loaded
```
brianteeman:
```suggestion
; If there is an error connecting to the database before initialisation, lib_joomla.ini can't be loaded
```

administrator/language/en-GB/joomla.ini
```diff
@@ -907,15 +907,9 @@ JHELP_USERS_USER_MANAGER_EDIT="Users:_Edit_Profile"
...
 JHELP_USERS_USER_NOTES_EDIT="User_Notes:_New_or_Edit"
 
-; If there is an error connecting database before initialisation, en-GB.lib_joomla.ini can't be loaded
-; we therefore have to load the strings from en-GB.ini
-
-JLIB_DATABASE_ERROR_ADAPTER_MYSQL="The MySQL adapter 'mysql' is not available."
-JLIB_DATABASE_ERROR_ADAPTER_MYSQLI="The MySQL adapter 'mysqli' is not available."
-JLIB_DATABASE_ERROR_CONNECT_DATABASE="Unable to connect to the Database: %s."
-JLIB_DATABASE_ERROR_CONNECT_MYSQL="Could not connect to MySQL."
+; If there is an error connecting database before initialisation, lib_joomla.ini can't be loaded
```
brianteeman:
```suggestion
; If there is an error connecting to the database before initialisation, lib_joomla.ini can't be loaded
```

language/en-GB/joomla.ini
```diff
@@ -362,15 +362,9 @@ JGRID_HEADING_LANGUAGE_DESC="Language descending"
...
 JGRID_HEADING_ORDERING_DESC="Ordering descending"
 
-; if there is an error connecting database before initialisation, en-GB.lib_joomla.ini can't be loaded
-; we therefore have to load the strings from en-GB.ini
-
-JLIB_DATABASE_ERROR_ADAPTER_MYSQL="The MySQL adapter 'mysql' is not available."
-JLIB_DATABASE_ERROR_ADAPTER_MYSQLI="The MySQL adapter 'mysqli' is not available."
-JLIB_DATABASE_ERROR_CONNECT_DATABASE="Unable to connect to the Database: %s"
-JLIB_DATABASE_ERROR_CONNECT_MYSQL="Could not connect to MySQL."
+; if there is an error connecting database before initialisation, lib_joomla.ini can't be loaded
```
brianteeman:
```suggestion
; If there is an error connecting to the database before initialisation, lib_joomla.ini can't be loaded
```

libraries/src/Table/Nested.php
```diff
@@ -1805,7 +1805,7 @@ protected function _logtable($showData = true, $showQuery = true)
 	 * @since   1.7.0
 	 * @throws  \Exception on database error.
 	 */
-	protected function _runQuery($query, $errorMessage)
+	protected function _runQuery($query, $errorMessage = '')
```
SharkyKZ, line 1808: Might as well remove it.

Bakual: I *think* it would have to be deprecated first, but not sure since it's a protected method. I didn't want to go through that hassle in this PR.
@wilsonge Would it be fine to remove the $errorMessage argument completely? It isn't used at all (even Doc Bloc says so).

SharkyKZ: It's going to error on PHP 8 either way.

Bakual: Why that?

SharkyKZ: https://3v4l.org/6Y3EK.

Bakual: Ah, I see. But that is only the case if an extending class overrides this protected _runQuery function.

RickR2H:
I have tested this item :white_check_mark: successfully on c18193650f7448f0339f686b75c6594565b535bf<br /><br />Checked all the removed language strings with a global search trough the system. It appears all the removed strings are not used anymore is the system. Thanks for the cleanup! <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31674">issues.joomla.org/tracker/joomla-cms/31674</a>.</sub>

drmenzelit:
I have tested this item :white_check_mark: successfully on c18193650f7448f0339f686b75c6594565b535bf<br /><br />Searched for removed strings and found only in ini files.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31674">issues.joomla.org/tracker/joomla-cms/31674</a>.</sub>

brianteeman:
Is there any reason that this is not merged? It has two successful tests and it would be a shame for it to miss the language freeze<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31674">issues.joomla.org/tracker/joomla-cms/31674</a>.</sub>

Bakual:
Since we are now in language freeze, I'm closing this PR.

PR #31675, opened 2020-12-16
[4.1] Actionlogs: Added parameter to disable relative timestamps.

Harmageddon:
Based on a request in the German support forum: https://forum.joomla.de/thread/12928-joomla-benutzeraktivit%C3%A4ten-datum-und-uhrzeit-protokollieren/

The actionlogs component displays date and time of an action relative to the current time ("X minutes/hours/days ago"). While this might be useful to some people, others prefer to see only the exact timestamp when the action happened. In J3, this timestamp is hidden in a tooltip. In J4, the display is better, as the timestamp is in small font below the relative one. But this becomes confusing and redundant again, at least for actions that happened more than 4 weeks ago. In this case, we have two times the same date in a different format:
joomla-pr-actionlog-date-relative-old

### Summary of Changes

As some users might be fine with the relative date format, I introduced a parameter in the configuration of com_actionlogs, that allows to disable the relative display. To not change current experience, it defaults to the current behaviour, i.e., displaying both relative and absolute date.

### Testing Instructions

1. Make sure you have the action logs plugins enabled and actions are logged.
2. Perform some actions in backend.
3. Navigate to "Users - User Actions Log".
4. Look at the list. It should look like the image at "actual result" below.
5. Apply this PR.
6. Reload the action log list. It should still look the same.
7. Click on "Options" and set the new parameter "Show Relative Date/Time" to "No".
8. Save & Close
9. Look at the action log list again.

joomla-pr-actionlog-date-relative-yes

joomla-pr-actionlog-date-relative-options

joomla-pr-actionlog-date-relative-no

Should be documented at https://docs.joomla.org/Help39:Components_User_Actionlogs_Options


brianteeman:
Do we really need to add another option :(

chmst:
I like to see the date and time, but think we just can display date and time. I see no advantage in the relative presentation - this relative presentation seems to be inspired by forum displays, where it makes more sense.

Harmageddon:
> Do we really need to add another option :(

Not necessarily - I'm open for alternative proposals. And I'm in favour of reducing our number of parameters wherever possible.

I don't really like such relative date displays, especially when it's not "5 minutes ago", but "around 2 weeks ago", which might be 2 weeks, might be almost 4 weeks, nobody knows. And looking at the forum post and the comment by @chmst, I'm not alone. So my personal choice would be to get rid of the relative display and only show the exact timestamp.
But as I see this kind of display more and more across the internet, I guess there are people who prefer it like this and might find it easier to read. 

Should we go for one of the two ways and ignore the other user group? Or introduce a new parameter (in a form that right now has only 5 parameters, so it's not crowded - yet) to allow for a choice?

HLeithner:
Just change it to absolute dates, I see no reason to add an option for this.

brianteeman:
I find it very useful to see "5 minutes ago" etc so I dont have to do mental arithmetic and check my phone for the date.

ChristineWk:
I have tested this item :white_check_mark: successfully on ff32892ea3338311d2556abb8c4e1db2a3e7ea95<br /><br />I als prefer, NOT to see e.g.: "5 minutes ago".....<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31675">issues.joomla.org/tracker/joomla-cms/31675</a>.</sub>

ReLater:
I have tested this item :white_check_mark: successfully on ff32892ea3338311d2556abb8c4e1db2a3e7ea95<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31675">issues.joomla.org/tracker/joomla-cms/31675</a>.</sub>

ReLater:
I think with the additional param this pr is fine like it is.

Fedik:
it can be both at same time, without extra option, kind of:
if difference less than a 12h then show "relative", and "tolltip" (`title` attribute) shows real time

brianteeman:
> if difference less than a 12h then show "relative", and "tolltip" (title attribute) shows real time

tooltip is not accessible

Fedik:
> tooltip is not accessible

`title` attribute

ReLater:
When I have a statistical list in front of me I don't want to hover, sorry, stupid text values to see the real date in a tooltip that I can't copy. There's a date filter in that view. Simply add additional options there.

administrator/components/com_actionlogs/config.xml
```diff
@@ -56,5 +56,16 @@
...
 			<option value="PATCH">COM_ACTIONLOGS_FIELD_VALUE_PATCH</option>
 		</field>
+		<field
+			name="date_relative"
+			type="radio"
+			label="COM_ACTIONLOGS_DATE_RELATIVE_LABEL"
+			layout="joomla.form.field.radio.switcher"
+			default="1"
+			filter="integer"
+		>
```
Quy:
```suggestion
			>
```

administrator/components/com_actionlogs/config.xml
```diff
@@ -56,5 +56,16 @@
...
+		<field
+			name="date_relative"
+			type="radio"
+			label="COM_ACTIONLOGS_DATE_RELATIVE_LABEL"
+			layout="joomla.form.field.radio.switcher"
+			default="1"
+			filter="integer"
+			>
+			<option value="0">JNO</option>
+			<option value="1">JYES</option>
```
Quy:
```suggestion
			<option value="0">JHIDE</option>
			<option value="1">JSHOW</option>
```

administrator/language/en-GB/com_actionlogs.ini
```diff
@@ -14,6 +14,7 @@ COM_ACTIONLOGS_COMMA="Comma"
 COM_ACTIONLOGS_CONFIGURATION="User Actions Log: Options"
 COM_ACTIONLOGS_CSV_DELIMITER_LABEL="CSV Delimiter"
 COM_ACTIONLOGS_DATE="Date"
+COM_ACTIONLOGS_DATE_RELATIVE_LABEL="Show Relative Date/Time"
```
Quy:
```suggestion
COM_ACTIONLOGS_DATE_RELATIVE_LABEL="Relative Date/Time"
```

Harmageddon:
I think the whole discussion here underlines the reason for the additional parameter, so both sides can choose their way without restricting the others.

> it can be both at same time, without extra option, kind of:
> if difference less than a 12h then show "relative", and "tolltip" (`title` attribute) shows real time

As @ReLater wrote: Hiding the information in a tooltip / title (the way it is in J3) is even worse, because you have to hover over each single item to access the information.

Fedik:
> I think the whole discussion here underlines the reason for the additional parameter,

Yeap, exactly :smile: 
It really makes no sense, for such simple thing.
Either it should be a global parameter to switch all the backend dates to displaying to "relative", or no parameter at all.

> so both sides can choose their way without restricting the others.

well, it is not possible to fulfill everyone's needs anyway :wink: 

Just drop the parameter, and display the date as everywhere on the site.

ChristineWk:
I have tested this item :white_check_mark: successfully on bb6238ca68437a7c543e2136f9396a6779646ad5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31675">issues.joomla.org/tracker/joomla-cms/31675</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on bb6238ca68437a7c543e2136f9396a6779646ad5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31675">issues.joomla.org/tracker/joomla-cms/31675</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31675">issues.joomla.org/tracker/joomla-cms/31675</a>.</sub>

Harmageddon:
Is there anything that keeps this PR from being merged? Anything I could or should add here?

rdeutz:
I am moving this to 4.1, we are in beta and this is a new feature. Thank you all who have been working on this and bringing it to the state it is now. But if we want to get 4.0 out of the door we need to make a cut.

Side note: this decision has be discussed in the production department and the maintainers team. It's not just mine even if I support the decision.

Harmageddon:
Okay, I see you (PLT) have to draw the line somewhere. Looking forward to 4.0 and 4.1 then. :-)

bembelimen:
Thanks

Issue #30969, opened 2020-10-07
[4] Incorrect integer value: 'Mr Hacker' for column 'created_by' at row 1


PhilETaylor:
### Steps to reproduce the issue

Joomla 4 -> Add Article -> Save
Edit same article 
Edit HTML of the page with inspector tools

change the value of `jform_created_by_id ` field

`<input type="hidden" id="jform_created_by_id" name="jform[created_by]" value="Mr Hacker" class="field-user-input " data-onchange="">`

### Expected result

Validation before hitting the database with a "string" 

### Actual result

`Save failed with the following error: Incorrect integer value: 'Mr Hacker' for column 'created_by' at row 1`

<img width="958" alt="Screenshot 2020-10-07 at 02 06 13" src="https://user-images.githubusercontent.com/400092/95275916-b0f9f600-0841-11eb-9c79-f4b5f3f57efd.png">

### Additional comments

Do the same again, but this time use a non valid integer number like 999999 

`<input type="hidden" id="jform_created_by_id" name="jform[created_by]" value="999999" class="field-user-input " data-onchange="">`

Save.

NOTE THAT 999999 PERSISTS IN THE DATABASE!!!!!!!! This id is invalid, its not a user id of an existing user.

<img width="783" alt="Screenshot 2020-10-07 at 02 07 57" src="https://user-images.githubusercontent.com/400092/95276035-0209ea00-0842-11eb-9ce0-26e48e7018d1.png">

brianteeman:
Like all validation in Joomla it is broken. I've banged my head against a brick wall for 18 months over this trying to convince people

Quy:
Related #29855

PhilETaylor:
> Related #29855

In so much as the user not found message only - but the real reason for opening this issue was not for that - it was to highlight the complete lack of validation that is allowing a string to get as far as an integer database column, unfiltered and un checked. I should also NOT be able to set a new user id, of a user that doesn't exist. 

Probably should have been an email to @joomla/security but I had already posted before I realised the true depth of this issue.

wilsonge:
I assume this isn't specific to J4?

wilsonge:
Pull request for J4 anyhow https://github.com/joomla/joomla-cms/pull/31698

PR #31698, opened 2020-12-18
[4.0] Add User ID validation to fields

wilsonge:
Pull Request for Issue #30969 .

### Summary of Changes
Adds a new validation rule for having a valid user ID and applies it to all form fields in the CMS of type user.

### Testing Instructions
Joomla 4 -> Edit any article (sample data or manually create one and hit save)

Edit HTML of the edit page with inspector tools

change the value of jform_created_by_id field

<input type="hidden" id="jform_created_by_id" name="jform[created_by]" value="Mr Hacker" class="field-user-input " data-onchange="">

Before patch: If you hit save you'll get a database validation error `Save failed with the following error: Incorrect integer value: 'Mr Hacker' for column 'created_by' at row 1` (see original issue for screenshot). After patch:

Screenshot 2020-12-18 at 04 57 12

You can also try and do the same thing with an invalid user id (e.g. 999999) - note before patch an integer id even if not valid was saved to the DB. After the patch it will not be.

### Backwards Compatibility

There is one issue here which is that articles which were created by deleted users now will NOT save to the database until they are changed to point to a user that does exist. I'm unsure how serious to rank this as an issue (you will get an error like the following :-

Screenshot 2020-12-18 at 04 59 00

but it will still save)

Possibly relating to the deleted users comment.

/cc @joomla/security

### DIFF
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -43,7 +43,7 @@ class UserIdRule extends FormRule
 		// Check if the field is required.
 		$required = ((string) $element['required'] === 'true' || (string) $element['required'] === 'required');
 
-		// If the value is empty, null or has the value 0 and the field is not required return True else return false
+		// If the value is empty, null or has the value 0 and the field is not required return true else return false
 		if (($value === '' || $value === null || (string) $value === '0'))
 		{
 			return !$required;
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -19,7 +19,7 @@ use Joomla\Registry\Registry;
 /**
  * Form Rule class for the Joomla Platform.
  *
- * @since  1.7.0
+ * @since  __DEPLOY_VERSION__
  */
 class UserIdRule extends FormRule
 {
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -43,8 +43,8 @@ class UserIdRule extends FormRule
 		// Check if the field is required.
 		$required = ((string) $element['required'] === 'true' || (string) $element['required'] === 'required');
 
-		// If the value is empty and the field is not required return True.
-		if (($value === '' || $value === null) && !$required)
+		// If the value is empty, null or has the value 0 and the field is not required return True.
+		if (($value === '' || $value === null || (string) $value === '0') && !$required)
 		{
 			return true;
 		}
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -24,7 +24,7 @@ use Joomla\Registry\Registry;
 class UserIdRule extends FormRule
 {
 	/**
-	 * Method to test the username for uniqueness.
+	 * Method to test the validity of a Joomla User.
 	 *
 	 * @param   \SimpleXMLElement  $element  The SimpleXMLElement object representing the `<field>` tag for the form field object.
 	 * @param   mixed              $value    The form field value to validate.
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -60,8 +60,6 @@ class UserIdRule extends FormRule
 			->bind(':userId', $value, ParameterType::INTEGER);
 
 		// Set and query the database.
-		$db->setQuery($query);
-
-		return (bool) $db->loadResult();
+		return (bool) $db->setQuery($query)->loadResult();
 	}
 }
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -2,7 +2,7 @@
 /**
  * Joomla! Content Management System
  *
- * @copyright  (C) 2017 Open Source Matters, Inc. <https://www.joomla.org>
+ * @copyright  (C) 2020 Open Source Matters, Inc. <https://www.joomla.org>
  * @license    GNU General Public License version 2 or later; see LICENSE.txt
  */
 
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -43,10 +43,10 @@ class UserIdRule extends FormRule
 		// Check if the field is required.
 		$required = ((string) $element['required'] === 'true' || (string) $element['required'] === 'required');
 
-		// If the value is empty, null or has the value 0 and the field is not required return True.
-		if (($value === '' || $value === null || (string) $value === '0') && !$required)
+		// If the value is empty, null or has the value 0 and the field is not required return True else return false
+		if (($value === '' || $value === null || (string) $value === '0'))
 		{
-			return true;
+			return !$required;
 		}
 
 		// Get the database object and a new query object.
--- a/libraries/src/Form/Rule/UserIdRule.php
+++ b/libraries/src/Form/Rule/UserIdRule.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  (C) 2017 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license    GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\CMS\Form\Rule;
+
+\defined('JPATH_PLATFORM') or die;
+
+use Joomla\CMS\Factory;
+use Joomla\CMS\Form\Form;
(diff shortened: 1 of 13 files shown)


libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  (C) 2017 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman:
```suggestion
 * @copyright  (C) 2020 Open Source Matters, Inc. <https://www.joomla.org>
```

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,64 @@
...
+use Joomla\CMS\Factory;
+use Joomla\CMS\Form\Form;
+use Joomla\CMS\Form\FormRule;
+use Joomla\Database\ParameterType;
+use Joomla\Registry\Registry;
+
+/**
+ * Form Rule class for the Joomla Platform.
+ *
+ * @since  1.7.0
```
brianteeman:
```suggestion
 * @since  __DEPLOY_VERSION__
```

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @since   1.7.0
+	 */
+	public function test(\SimpleXMLElement $element, $value, $group = null, Registry $input = null, Form $form = null)
+	{
+		// Get the database object and a new query object.
+		$db = Factory::getDbo();
+		$query = $db->getQuery(true);
+
+		// Build the query.
+		$query->select('COUNT(*)')
```
Denitz, line 48: Better use `$query->select('id')`, counting all users with the same ID is strange for checking the single unique ID presence.

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,64 @@
...
+			->from($db->quoteName('#__users'))
+			->where($db->quoteName('id') . ' = :userId')
+			->bind(':userId', $value, ParameterType::INTEGER);
+
+		// Set and query the database.
+		$db->setQuery($query);
+		$userId = (bool) $db->loadResult();
+
+		if ($userId !== 1)
```
Denitz: `$userId` is type-casted to boolean in line 55, it will never be equal to INT 1. Simple `return $userId;` is enough.

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @param   ?Registry          $input    An optional Registry object with the entire data set to validate against the entire form.
+	 * @param   ?Form              $form     The form object for which the field is being tested.
+	 *
+	 * @return  boolean  True if the value is valid, false otherwise.
+	 *
+	 * @since   1.7.0
+	 */
+	public function test(\SimpleXMLElement $element, $value, $group = null, Registry $input = null, Form $form = null)
+	{
+		// Get the database object and a new query object.
```
Denitz, line 43: What will be if a `$value` is empty? No need to query the database in this case.

wilsonge:
OK I pushed up the actual working version to the wrong remote last night 🤦  Some of this fixes your comments @Denitz (like the required field properties and the value returned).

libraries/src/Form/Rule/UserIdRule.php
HLeithner: count(*) actually returns 1 (or more) in $db->loadResult() which is casted to boolean. 1 will be `true` and no result would be `false` the solution is ok.

wilsonge: I actually pulled this code from the username form field actually. The count to bool seemed kinda smart. I'm not overly attached to it though

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,67 @@
...
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function test(\SimpleXMLElement $element, $value, $group = null, Registry $input = null, Form $form = null)
+	{
+		// Check if the field is required.
+		$required = ((string) $element['required'] === 'true' || (string) $element['required'] === 'required');
+
+		// If the value is empty and the field is not required return True.
+		if (($value === '' || $value === null) && !$required)
```
Denitz: What about zero? Maybe, `if (((int)$value === 0) && !$required)` ?

Besides, if field IS required and value IS empty - we still query the database for non-existing ID.

wilsonge: If the user form field doesn't have a user selected it returns an empty string rather than 0. I think generally we store this as null in the database too rather than the integer 0

Denitz: Actually, this should cover all cases:
```
if ((int) $value === 0)
{
    return !$required;
}
```

Denitz: 3rd-party extensions can have zeros, and even core tables have, i.e. `#__banners` - `created_by int(10) unsigned NOT NULL DEFAULT 0,`

wilsonge: Fair enough :)

Denitz: OK, what is the purpose of querying the database if a field is required but the value is empty?

wilsonge: We aren't. We are returning true?

wilsonge: Oh wait I see what you mean 🤦

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Form Rule class for the Joomla Platform.
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class UserIdRule extends FormRule
+{
+	/**
+	 * Method to test the username for uniqueness.
```
Quy: Correct description?

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,67 @@
...
+	 * @return  boolean  True if the value is valid, false otherwise.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function test(\SimpleXMLElement $element, $value, $group = null, Registry $input = null, Form $form = null)
+	{
+		// Check if the field is required.
+		$required = ((string) $element['required'] === 'true' || (string) $element['required'] === 'required');
+
+		// If the value is empty, null or has the value 0 and the field is not required return True else return false
```
Quy:
```suggestion
		// If the value is empty, null or has the value 0 and the field is not required return true else return false
```

libraries/src/Form/Rule/UserIdRule.php
```diff
@@ -0,0 +1,67 @@
...
+		// Build the query.
+		$query->select('COUNT(*)')
+			->from($db->quoteName('#__users'))
+			->where($db->quoteName('id') . ' = :userId')
+			->bind(':userId', $value, ParameterType::INTEGER);
+
+		// Set and query the database.
+		$db->setQuery($query);
+
+		return (bool) $db->loadResult();
```
bembelimen:
```suggestion
		return (bool) $db->setQuery($query)->loadResult();
```

ceford:
I have tested this item :white_check_mark: successfully on 44f8b35eced0d90a57a4b52c8c65954c13e990b5<br /><br />I can confirm:
before the patch an in valid user ID is saved - two messages: Article Saved (success) and Unable to find user with the ID: 800 (warning). 
with the patch an invalid user ID is not saved - two messages combined as one warning: Invalid field: Created By and Unable to find user with the ID: 800.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31698">issues.joomla.org/tracker/joomla-cms/31698</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on 44f8b35eced0d90a57a4b52c8c65954c13e990b5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31698">issues.joomla.org/tracker/joomla-cms/31698</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31698">issues.joomla.org/tracker/joomla-cms/31698</a>.</sub>

richard67:
Thanks!

PR #31706, opened 2020-12-18
Hashing the simpleXMLElement doesn't produce an unique hash per element

Bakual:
Currently the field `LimitboxField` generates a hash so it doesn't have to calculate the options with each appearance of the same element. This is fine.
Howvere the has is created from `$this->element` which is a simpleXMLElement. We all know that stuff behaves very strange and in unexpected ways.
In this case, `$hash = md5($this->element);` doesn't create a hash based on the whole element with all its attributes. The hash is instead created on an empty simpleXMLElement (ignoring attributes) which ends up the same hash for each instance of the field.

### Summary of Changes
This PR changes the hash so it is created from the XML string instead.

### Testing Instructions
Simplest test would be to add two fields to the article form (administrator/components/com_content/models/forms/article.xml)
Within the fieldset "basic" add this lines:
````
		<field
			name="list_limit"
			type="limitbox"
			label="List"
			useglobal="true"
		/>
		<field
			name="feed_limit"
			type="limitbox"
			label="Feed"
			useglobal="true"
		/>

````
Now check the options tab in an article. And see the value in "Use Global"
If your installation has the default values, then the List field should show 20 and the Feed field should show 10 (values come from global configuration).

image

Both fields show the same value, because the hash is actually the same for both fields

image

Fields are showing correct values since hashes are different.

None

### DIFF
--- a/libraries/src/Form/Field/LimitboxField.php
+++ b/libraries/src/Form/Field/LimitboxField.php
@@ -54,7 +54,7 @@ class LimitboxField extends \JFormFieldList
 	protected function getOptions()
 	{
 		// Accepted modifiers
-		$hash = md5($this->element);
+		$hash = md5($this->element['name']);
 
 		if (!isset(static::$options[$hash]))
 		{
--- a/libraries/src/Form/Field/LimitboxField.php
+++ b/libraries/src/Form/Field/LimitboxField.php
@@ -54,7 +54,7 @@ class LimitboxField extends \JFormFieldList
 	protected function getOptions()
 	{
 		// Accepted modifiers
-		$hash = md5($this->element['name']);
+		$hash = md5($this->element->asXML());
 
 		if (!isset(static::$options[$hash]))
 		{

### CODE REVIEW

**Comment by alikon** [comment]
I have tested this item :white_check_mark: successfully on bfac765a52e95ca895b8202cb95176cf5a288cb3<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31706">issues.joomla.org/tracker/joomla-cms/31706</a>.</sub>

File: libraries/src/Form/Field/LimitboxField.php [hunk 1]
```diff
@@ -54,7 +54,7 @@ class LimitboxField extends \JFormFieldList
 	protected function getOptions()
 	{
 		// Accepted modifiers
-		$hash = md5($this->element);
+		$hash = md5($this->element['name']);
```
HLeithner:
```suggestion
		$hash = md5($this->element->asXML());
```

ghost:
I have tested this item :white_check_mark: successfully on bfac765a52e95ca895b8202cb95176cf5a288cb3<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31706">issues.joomla.org/tracker/joomla-cms/31706</a>.</sub>

libraries/src/Form/Field/LimitboxField.php
Bakual: Thanks, that's perfect! Didn't know this method exists.

Bakual:
@alikon and @gostn Can you please retest? I applied a nice suggestion by @HLeithner

viocassel:
I have tested this item :white_check_mark: successfully on e02b0f5bae8f830eb0b71cc5aec90012f7caa7cc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31706">issues.joomla.org/tracker/joomla-cms/31706</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on e02b0f5bae8f830eb0b71cc5aec90012f7caa7cc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31706">issues.joomla.org/tracker/joomla-cms/31706</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31706">issues.joomla.org/tracker/joomla-cms/31706</a>.</sub>

HLeithner:
thanks

PR #31707, opened 2020-12-18
[Staging] Issue Templates

brianteeman:
No point in having the rfc or feature request templates as we have discussions


HLeithner:
Shouldn't they refer to discussion instead of removing it?

PR #31708, opened 2020-12-18
[4.0] Remove issue templates

brianteeman:
No point in having the RFC or Feature Request templates

Same as #31707 - wasn't sure if it needs to be done in both


wilsonge:
It's fine I'll merge it up if the 3.x one is merged. But thankyou!

PR #31712, opened 2020-12-18
[4.0] Fix merge error

brianteeman:
Fixes a merge error with this file in #31221

It's clearly wring as it is double the size ;)


wilsonge:
I'm unsure whether this is or isn't an error. You're downgrading the lock file from v2 to v1 (see the diff right at the top). v2 came out with npm 7 in October this year. https://github.com/npm/rfcs/blob/latest/implemented/0013-no-package-json-_fields.md I think it might be related to this - but still trying to make my way through the npm 7 release notes to try and figure it out for sure

brianteeman:
I just know that I was getting errors so ran npm i and saw that the file was half the size

brianteeman:
As the files from 6 and 7 are not compatible with each other then we will need to update
```

  "engines": {
    "node": ">=10.19",
    "npm": ">=6.13.4"
  },
```

Otherwise whenever someone commits with a different npm the file will change etc??

wilsonge:
> package-lock.json is updated to a newer format, using "lockfileVersion": 2. This format is backwards-compatible with npm CLI versions using "lockfileVersion": 1, but older npm clients will print a warning about the version mismatch.

https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major

So they are compatible but yes they generate different files (I'm still running npm v6 and everything still installs/updates for me - but yes running npm install will revert to the old-style file). I'm not sure how much it matters 👍 

Current node LTS version ships with npm v6, but the "current" node version ships with npm v7 - there won't be a node LTS version until October 2021 that contains npm v7 - so I guess we keep with npm 6 and lockfile v1 for now

brianteeman:
just checked and the package settings are coded to say npm 6 not 7 so they all should be lockfile=1

Issue #29425, opened 2020-06-04
[4.0] Can't start installation on linux


joecurtis:
When trying to install 4.0 beta 1 on both a Raspberry pi 3 running buster and  also a Fedora 23 
installation on a separate machine I get the following error as soon as I try to open the site using the extracted files from https://github.com/joomla/joomla-cms/releases/download/4.0.0-beta/Joomla_4.0.0-beta1-Beta-Full_Package.tar.bz2.

Oops! An Error Occurred
The server returned a "500 Whoops, looks like something went wrong.".

Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused.

skurvish:
@joecurtis take a look at this ticket, it may help: https://github.com/joomla/joomla-cms/issues/29424<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/29425">issues.joomla.org/tracker/joomla-cms/29425</a>.</sub>

richard67:
@joecurtis Please make sure your PHP meets the technical requirements [https://downloads.joomla.org/technical-requirements](https://downloads.joomla.org/technical-requirements) and has a sufficient max_execution_time and memory_limit.

richard67:
P.S. ... and the updated requirements for Joomla 4 you can find here: [https://docs.joomla.org/Potential_backward_compatibility_issues_in_Joomla_4](https://docs.joomla.org/Potential_backward_compatibility_issues_in_Joomla_4).

dgrammatiko:
@richard67 that's not a Joomla error. Probably a misconfigured Apache/Nginx - PHP/PHP-FPM

richard67:
@dgrammatiko Agree. Just wanted to point to a few docs, maybe some php module is missing. Will give it a bit time and then close it.

dgrammatiko:
> maybe some php module is missing. 

IIRC we did spend some time to get the initial checks right, eg even a PHP4 will echo some nice html instead of some cryptic text. If there are cases that the current code isn't covering then there should be an effort to get some code for those.
Anyways this one here is not a case that Joomla can handle, eg if you're running php-fpm and the socket is not found I think you get this 500 error by Apache/Nginx/IIS. Joomla never executed a single line in this case neither there is something that can do about it.

Anyways my 2p

dgrammatiko:
[...]

PR #31735, opened 2020-12-20
[4.0] check for libxml

brianteeman:
PR for #29425

This is a real edge case discovered by @fedik https://github.com/joomla/joomla-cms/issues/29425#issuecomment-640010134

libxml is enabled by default when compiling php and it cannot be disabled etc in a php.ini

The check I have put in place is the same one that myBB uses for the same issue.

The only questions are
1. Did I put the check early enough
2. Do we really need it for such an edge case

### DIFF
--- a/installation/src/Application/InstallationApplication.php
+++ b/installation/src/Application/InstallationApplication.php
@@ -334,7 +334,7 @@ final class InstallationApplication extends CMSApplication
 		if (!function_exists('xml_parser_create'))
 		{
 			$output   = 'Joomla requires the libxml PHP extension. This should be enabled by default.';
-			$template = JPATH_ROOT . '/media/system/noxml.html';
+			$template = JPATH_ROOT . '/media/system/html/noxml.html';
 
 			if (file_exists($template))
 			{
--- a/installation/src/Application/InstallationApplication.php
+++ b/installation/src/Application/InstallationApplication.php
@@ -333,7 +333,15 @@ final class InstallationApplication extends CMSApplication
 	{
 		if (!function_exists('xml_parser_create'))
 		{
-			die('Joomla requires the libxml PHP extension. This should be enabled by default.');
+			$output   = 'Joomla requires the libxml PHP extension. This should be enabled by default.';
+			$template = JPATH_ROOT . '/media/system/noxml.html';
+
+			if (file_exists($template ))
+			{
+				$output = file_get_contents($template);
+			}
+
+			die($output);
 		}
 
 		$xml = simplexml_load_file(JPATH_INSTALLATION . '/localise.xml');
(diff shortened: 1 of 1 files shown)


dgrammatiko:
@brianteeman although this would work I think that the check needs to be in the Initial checks (compatibility layer): https://github.com/joomla/joomla-cms/blob/77bcd7198380ea033c77eecfd07ceb839805fe5f/installation/src/Model/ChecksModel.php#L66

Also that would result to a better error view

HLeithner:
@brianteeman can you fix the cs plz

Fedik:
maybe better to place it in `installation/includes/app.php` or even `installation/index.php`

with some nice message like:
https://github.com/joomla/joomla-cms/blob/77bcd7198380ea033c77eecfd07ceb839805fe5f/installation/index.php#L20-L26

> I think that the check needs to be in the Initial checks (compatibility layer)

The problem the App does not even start if `xml` not available ;) 

upd: the text in `templates/system/incompatible.html` can be updated to show more info, than just minimum php version

dgrammatiko:
> maybe better to place it in installation/includes/app.php or even installation/index.php

Since all the checks from the Model could end up being foobar if any of these fail I guess beefing up these files is a more realistic approach. Yeah I know the code will end up being god awful but functional code is better than beautiful non functional code.

My 2c

installation/src/Application/InstallationApplication.php
```diff
@@ -333,7 +333,15 @@ public function getLocalise()
 	{
 		if (!function_exists('xml_parser_create'))
 		{
-			die('Joomla requires the libxml PHP extension. This should be enabled by default.');
+			$output   = 'Joomla requires the libxml PHP extension. This should be enabled by default.';
+			$template = JPATH_ROOT . '/media/system/noxml.html';
+
+			if (file_exists($template ))
```
dgrammatiko:
```suggestion
			if (file_exists($template))
```

dgrammatiko (changes requested):

installation/src/Application/InstallationApplication.php
```diff
@@ -331,6 +331,19 @@ private function executeController()
 	 */
 	public function getLocalise()
 	{
+		if (!function_exists('xml_parser_create'))
+		{
+			$output   = 'Joomla requires the libxml PHP extension. This should be enabled by default.';
+			$template = JPATH_ROOT . '/media/system/noxml.html';
```
dgrammatiko:
```suggestion
			$template = JPATH_ROOT . '/media/system/html/noxml.html';
```

dgrammatiko (changes requested):

Issue #29741, opened 2020-06-21
[4.0] Feed Display Word Count Broken


PhilETaylor:
### Steps to reproduce the issue

Create a new feed display site module with RSS feed url of https://www.nytimes.com/svc/collections/v1/publish/https://www.nytimes.com/section/world/rss.xml

Set the word count to 0 - load front page to see the module

<img width="328" alt="Screenshot 2020-06-21 at 18 24 58" src="https://user-images.githubusercontent.com/400092/85231118-8a2bc700-b3ec-11ea-885e-a6bb1e45882e.png">

set the word count to 5 - load the front page and compare.

<img width="1188" alt="Screenshot 2020-06-21 at 18 24 15" src="https://user-images.githubusercontent.com/400092/85231110-71231600-b3ec-11ea-9670-1ca258e5e25a.png">

### Expected result

5 words of each RSS feed item

### Actual result

`...` in each item.

<img width="317" alt="Screenshot 2020-06-21 at 18 24 40" src="https://user-images.githubusercontent.com/400092/85231114-7d0ed800-b3ec-11ea-8ebc-b9c8382125e5.png">

SharkyKZ:
The label says word count but the code handling this is for character count.

PhilETaylor:
Well even if that is the case, it’s not working as a char count as regardless what it’s set to all I get is ...

SharkyKZ:
You do get "The..." in one of the entries. So that's working fine.

SharkyKZ:
Before https://github.com/joomla/joomla-cms/commit/e3ccc0d680a3207112e148b421a1ae4144c4af2f this was indeed used as word count.

brianteeman:
string.truncate counts characters not words

ghost:
> string.truncate counts characters not words

`Word Count` > ` Character Count`

bembelimen:
Closed as having a PR

PR #31737, opened 2020-12-20
[4.0 newsfeed character count

brianteeman:
It is not a word count. string.truncate is a character count (and a trim to prevent partial words)

Create a newsfeed module with the following example feed `https://www.nytimes.com/svc/collections/v1/publish/https://www.nytimes.com/section/world/rss.xml`

Leave the character count at 0 and you have the full feed
retest with the character count at 10 and you have approx 10 characters depenant on word breaks

Pull Request for Issue #29741


modules/mod_feed/mod_feed.xml
```diff
@@ -125,10 +125,10 @@
 				</field>
 
 				<field
-					name="word_count"
+					name="char_count"
 					type="text"
```
Quy, line 129: Change to `number` type?

toivo:
I have tested this item :white_check_mark: successfully on 5f9b20744bbc4eb1333e30a894438b97120cf5cc<br /><br />Tested successfully in Beta6-dev of 21 December.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31737">issues.joomla.org/tracker/joomla-cms/31737</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on 5f9b20744bbc4eb1333e30a894438b97120cf5cc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31737">issues.joomla.org/tracker/joomla-cms/31737</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31737">issues.joomla.org/tracker/joomla-cms/31737</a>.</sub>

infograf768:
This works, but I guess you are aware that this is not B/C because the field name is modified and therefore existing modules will default to 0.

Not my decision to merge.
@wilsonge @HLeithner

brianteeman:
Good point about the changed field name. I didnt consider that. At least its only a cosmetic b/c change

I am happy to change it back but would prefer to keep the change so that the code is more meaningful

I leave it to others to decide

modules/mod_feed/tmpl/default.php
```diff
@@ -127,7 +127,7 @@
 						<?php
 							// Strip the images.
 							$text = OutputFilter::stripImages($text);
-							$text = HTMLHelper::_('string.truncate', $text, $params->get('word_count', 0));
+							$text = HTMLHelper::_('string.truncate', $text, $params->get('char_count', 0));
```
bembelimen, line 130:
```suggestion
							$text = HTMLHelper::_('string.truncate', $text, $params->get('char_count', $params->get('word_count', 0) * 5));
```

I would suggest to use the average letter count for english words and multiply it with the old parameter, as default for the new parameter. ("5" is the result of a quick internet search)

brianteeman: That is not a good idea as it is tied to english and it would confuse a user who inputs 5 words and sees 3 words

HLeithner: It's better then a string with `0` characters for upgraded sites.

In this context word count is not optimal anyway. But if you don't like it you can easily fix it by explode the string using whitespace and only take the word count.

brianteeman: If you are concerned about the upgraded sites then I can always revert the field name change as already stated. FYI the 0 means all not none

HLeithner: I suspect that we already have enough "silent" b/c breaks of this kind without automatically migration path.

brianteeman:
Please advise how to proceed to resolve and fix this bug

brianteeman:
No point in keeping this open

PR #31750, opened 2020-12-21
[4.0] Fix for subfields values not outputting 0 values

AndySDH:
### Summary of Changes

Very quick fix.

It turns out that having a 0 as the value of a subfield wasn't outputting the value. This fixes it.

### Testing Instructions

Create a subfields instance containing for example an integer field, or a text field, or a radio field, etc. that can contain a 0 value.

In the article form, give this field a 0 value inside the subfields.

The 0 values are not output, they're turned into an empty string.

The 0 values are output correctly.

None.


wilsonge:
Merging this on review

PR #31761, opened 2020-12-22
[4.0] contact xtd-editor button

brianteeman:
Corrects the svg icon that is used in the button so that its the same as used on the menus etc.

### before
image

image

### after
image


wilsonge:
LGTM - Thanks!

brianteeman:
thanks

Issue #31609, opened 2020-12-07
[4.0] add aria-label / aria-labelledby to module rendering


brianteeman:
@hans2103 introduced code with #29886 which correctly adds aria values to the module if it is using a landmark. However it also adds the value if it is a div and it has no effect here at all as assistive tech can not _see_ it

Reference https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/

Summary 
If you use aria-label, aria-labelledby, or aria-describedby with any other elements (like div, span, p, blockquote, or strong etc.), they generally won’t work across all browser/assistive technology combinations.

Confirmed that this is still valid advice with the author https://twitter.com/LeonieWatson/status/1335905663113900033

brianteeman:
Closed as we have two pr to resolve this

PR #31772, opened 2020-12-23
[4.0] html5 chrome fix [a11y] and markup

brianteeman:
With this PR the aria values are only added if the module is NOT a div ie its a section etc

Additionally there was no check for the presence of a header class as a result by default there would be an empty class

### Testing Instructions
Create a module and make sure it has the html5 chrome

```
<div class="moduletable " aria-labelledby="mod-1">
	<h3 class="" id="mod-1">Title</h3>
```

```
<div class="moduletable ">
	<h3>Module Title</h3>
```

#### With header class
```
<div class="moduletable ">
	<h3 class="example">Module Title</h3>
```

#### Not a div
```
<section class="moduletable " aria-labelledby="mod-1">
	<h3 id="mod-1">Title</h3>
```

### Note
This is Partial Pull Request for Issue #31609 

If approved I will make similar PR for the same mistakes in the cassiopeia chromes

### DIFF
--- a/layouts/chromes/html5.php
+++ b/layouts/chromes/html5.php
@@ -29,14 +29,21 @@ $moduleAttribs['class'] .= $bootstrapSize !== 0 ? ' col-md-' . $bootstrapSize : 
 $headerTag              = htmlspecialchars($params->get('header_tag', 'h3'), ENT_QUOTES, 'UTF-8');
 $headerClass            = htmlspecialchars($params->get('header_class', ''), ENT_QUOTES, 'UTF-8');
 $headerAttribs          = [];
-$headerAttribs['class'] = $headerClass;
 
-if ($module->showtitle) :
-	$moduleAttribs['aria-labelledby'] = 'mod-' . $module->id;
-	$headerAttribs['id']             = 'mod-' . $module->id;
-else:
-	$moduleAttribs['aria-label'] = $module->title;
-endif;
+// Only output a header class if one is set
+if ($headerClass != '') {
+	$headerAttribs['class'] = $headerClass;
+}
+
+// Only add aria if the moduleTag is not a div
+if ($moduleTag != 'div') {
+	if ($module->showtitle) :
+		$moduleAttribs['aria-labelledby'] = 'mod-' . $module->id;
+		$headerAttribs['id']             = 'mod-' . $module->id;
+	else:
+		$moduleAttribs['aria-label'] = $module->title;
+	endif;
+}
 
 $header = '<' . $headerTag . ' ' . ArrayHelper::toString($headerAttribs) . '>' . $module->title . '</' . $headerTag . '>';
 ?>
--- a/layouts/chromes/html5.php
+++ b/layouts/chromes/html5.php
@@ -31,13 +31,13 @@ $headerClass            = htmlspecialchars($params->get('header_class', ''), ENT
 $headerAttribs          = [];
 
 // Only output a header class if one is set
-if ($headerClass != '')
+if ($headerClass !== '')
 {
 	$headerAttribs['class'] = $headerClass;
 }
 
 // Only add aria if the moduleTag is not a div
-if ($moduleTag != 'div')
+if ($moduleTag !== 'div')
 {
 	if ($module->showtitle) :
 		$moduleAttribs['aria-labelledby'] = 'mod-' . $module->id;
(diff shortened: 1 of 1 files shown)


layouts/chromes/html5.php
```diff
@@ -29,14 +29,21 @@
...
-	$moduleAttribs['aria-label'] = $module->title;
-endif;
+
+// Only output a header class if one is set
+if ($headerClass != '') {
+	$headerAttribs['class'] = $headerClass;
+}
+
+// Only add aria if the moduleTag is not a div
+if ($moduleTag != 'div') {
```
Quy:
```suggestion
if ($headerClass != '')
{
	$headerAttribs['class'] = $headerClass;
}

// Only add aria if the moduleTag is not a div
if ($moduleTag != 'div')
{
```

layouts/chromes/html5.php
```diff
@@ -29,14 +29,23 @@
...
+{
+	$headerAttribs['class'] = $headerClass;
+}
+
+// Only add aria if the moduleTag is not a div
+if ($moduleTag != 'div')
+{
+	if ($module->showtitle) :
+		$moduleAttribs['aria-labelledby'] = 'mod-' . $module->id;
+		$headerAttribs['id']             = 'mod-' . $module->id;
```
Quy:
```suggestion
		$headerAttribs['id']              = 'mod-' . $module->id;
```

layouts/chromes/html5.php
```diff
@@ -29,14 +29,23 @@
...
-
-if ($module->showtitle) :
-	$moduleAttribs['aria-labelledby'] = 'mod-' . $module->id;
-	$headerAttribs['id']             = 'mod-' . $module->id;
-else:
-	$moduleAttribs['aria-label'] = $module->title;
-endif;
+
+// Only output a header class if one is set
+if ($headerClass != '')
```
hans2103: @brianteeman wouldn't it be better to code `if(!empty($headerClass))` ?

layouts/chromes/html5.php
```diff
@@ -29,14 +29,23 @@
...
+{
+	$headerAttribs['class'] = $headerClass;
+}
+
+// Only add aria if the moduleTag is not a div
+if ($moduleTag != 'div')
```
hans2103: @brianteeman wouldn't it be better to code `if($moduleTag !== 'div')` instead?

hans2103 (commented):
only did a code review and added some comments for @brianteeman

layouts/chromes/html5.php
brianteeman: No idea as they both do the same thing dont they?

layouts/chromes/html5.php
brianteeman: No idea as you dont say why it would be better

layouts/chromes/html5.php
hans2103: @brianteeman 
I did some research and just found out that:

```
<?php
	$var01 = ''; // totally empty
	$var02 = ' '; // contains a space

	if (!empty($var01)) {
		echo 'var01 is not empty';
	}
	if (!empty($var02)) {
		echo 'var02 is not empty';
	}
?>
```

after rendering I saw 'var02 is not empty' on my screen... So it is not empty.
I space will be seen as not empty. Not the result we want to achieve I think.

You are right... we have to stick to the `!=` or `!==` comparison.

layouts/chromes/html5.php
hans2103: Although it does not matter in this case. It makes more since when compare `0` and `'0'`. The first is a string, the latter a number. 
But... using `!==` and `===` throughout the entire project, instead of `!=` and `==`, makes the code more consistent.

brianteeman: Looking at the current codebase its about 50/50 usage between != and !==

brianteeman:
@wilsonge @bembelimen if you can tell me if it should be changed to !== from != or not then this can be tested and merged as its a silly bug that will trigger every a11y checker

HLeithner:
>if you can tell me if it should be changed to !== from != or not then this can be tested and merged as its a silly bug that will trigger every a11y checker

in this case it's easy, you expect a string 'div' so any other value is wrong and it doesn't matter if the type is integer so you have to use !==

brianteeman:
thanks

brianteeman:
Changes made as requested

HLeithner:
Would it make sense to change the default value for "Module Tag" to section?

HLeithner:
Thanks

brianteeman:
> Would it make sense to change the default value for "Module Tag" to section?

No not really

> If approved I will make similar PR for the same mistakes in the cassiopeia chromes

Doing that now

PR #31781, opened 2020-12-25
[4.0] Clean up the system templates (assets)

dgrammatiko:
Pull Request for Issue # . Part of the Child template effort

### Summary of Changes
- Move all the scss files to `/media/system/scss`
- Adjust all the stylesheet includes

### Testing Instructions
- You need to run `npm install`
- for the front end you need to add `$template->template = 'system';` after line 546 in the file https://github.com/joomla/joomla-cms/blob/7935d489c876155aa3ba1ccfa911a2ad6dd5a79e/libraries/src/Application/SiteApplication.php#L591 then visit your local site (eg localhost/)
- for the back end you need to add `$template->template = 'system';` after line 270 in the file https://github.com/joomla/joomla-cms/blob/7935d489c876155aa3ba1ccfa911a2ad6dd5a79e/libraries/src/Application/AdministratorApplication.php#L269 then visit your administrator part of your local site (eg localhost/administrator)
- For both cases check the assets loaded in the head of the document

Screenshot 2020-12-25 at 13 06 00
Screenshot 2020-12-25 at 13 09 49

It works exactly as before, the static assets were just moved into the `/media` folder

IDK the system templates are supposed to act as fallback if something goes horrible wrong with the current active template (but the fallback is not pretty or functional...)


brianteeman:
You are deleting `administrator/templates/system/images/calendar.png` but it is still referenced here i think https://github.com/joomla/joomla-cms/blob/9a467db9cba707782612541918e473a71691e16c/build/media_source/system/scss/general.scss#L163

administrator/templates/system/error.php
```diff
@@ -15,7 +15,7 @@
 /** @var Joomla\CMS\Document\ErrorDocument $this */
 
 // Load template CSS file
-$this->getWebAssetManager()->registerAndUseStyle('template.system.error', 'administrator/templates/system/css/error.css');
+$this->getWebAssetManager()->registerAndUseStyle('template.system.error', 'error.css');
```
Fedik, line 18: The full path was for purpose. I am not sure in current state, but the idea is:
If requested file `error.php`  `component.php` etc not found then it fallback to `system/error.php` etc, hovewer the template name stays default (eg cassiopeia, atum) and if you have here  'error.css' then override will look in the default template instead of `system`.

Just want to point out. This need to be checked. If it still works as expected, then good.

dgrammatiko: AFAIK template overrides always win

dgrammatiko:
> You are deleting `administrator/templates/system/images/calendar.png` but it is still referenced here

Yes the reference is fine because in the media/system/images the file already exists. So this pr just removed the duplicate...

brianteeman:
> Yes the reference is fine because in the media/system/images the file already exists. So this pr just removed the duplicate...

Isn't the path wrong then?

dgrammatiko:
> Isn't the path wrong then?

The scss will compile this to `url("../images/calendar.png")` which is relative to `/media/system/css/general.css` (the 2 dots dictate to go one level up and then use the `images/calendar.png`), so I'm pretty sure this is fine

EDIT: btw this is totally FOOBAR. The calendar is not using png images for the icon in J4, in J3 there was an instance in the Beez and the Hathor templates

administrator/templates/system/error.php
Fedik: The issue when template does not have it.
If template does not have error.php (and error.css, same for component.php), it should be loaded system version.

Currently I cannot check it. Just thoughts.

dgrammatiko: @Fedik ah, ok now I get it but this has nothing to do with the logic of the template loading files which obviously I'm not touching here. If it's broken (I haven't tested either) I'll do another PR for that

EDIT: just checked and the fallbacks are not the system but either cassiopeia or atum (and that's not my fault)...

brianteeman:
Can you please confirm the line numbers and changes for the test instructions

dgrammatiko:
@brianteeman oops there was an error in the instructions

brianteeman:
I dont seem to be able to trigger the use of the system template

dgrammatiko:
@brianteeman maybe it's easier to comment out the whole `public function getTemplate` in both the Site and Administrator apps so the system kicks in (it's the default on the Web app?

brianteeman:
So the problem was that I was editing one site and viewing another 🤦 

It appears to work BUT in your screenshots there is a data-asset-name which i dont get, is that from an earlier version?

image

dgrammatiko:
> It appears to work BUT in your screenshots there is a data-asset-name which i don't get

That's just the extra bits added whenever the debug is enabled. It's fine if you don't have them

brianteeman:
I have tested this item :white_check_mark: successfully on 9a467db9cba707782612541918e473a71691e16c<br /><br />got there in the end<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31781">issues.joomla.org/tracker/joomla-cms/31781</a>.</sub>

dgrammatiko:
@brianteeman thanks for testing, appreciated and Merry Xmas

brianteeman:
I have not tested this item.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31781">issues.joomla.org/tracker/joomla-cms/31781</a>.</sub>

brianteeman:
Sorry I had to remove my successful test.

The PR moves the css to the media folder BUT tinymce still refers to the /templates/ folder when it is looking for an editor.css otherwise it gives an error everytime you open an editor.

```
			if ($use_content_css)
			{
				// First check templates folder for default template
				// if no editor.css file in templates folder, check system template folder
				if (!file_exists($templates_path . '/' . $template . '/css/editor.css'))
				{
					// If no editor.css file in system folder, show alert
					if (!file_exists($templates_path . '/system/css/editor.css'))
					{
						Log::add(Text::_('PLG_TINY_ERR_EDITORCSSFILENOTPRESENT'), Log::WARNING, 'jerror');
					}
					else
					{
						$content_css = Uri::root(true) . '/templates/system/css/editor.css';
					}
				}
				else
				{
					$content_css = Uri::root(true) . '/templates/' . $template . '/css/editor.css';
				}
			}
		}
```<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31781">issues.joomla.org/tracker/joomla-cms/31781</a>.</sub>

dgrammatiko:
@brianteeman good catch! https://github.com/joomla/joomla-cms/pull/31781/commits/65418596b973f13b2ec1ebf145e1769154f75301 should fix that

brianteeman:
The last patch solved the error message but I'm not sure if the editor.css is being used. And shouldnt it be the minified file?

dgrammatiko:
> And shouldnt it be the minified file?

Joomla just hands in the url for the css to tinyMCE which is doing all the work here (it's in an iframe so it needs a bit more digging to get to it). Could it be minified? Sure but since we're doing all the checks ourselves (in the tinyMCE plugin, the PHP part) we have to do that ourselves. The part that I reverted was using `HTMLHelper::stylesheet` which does all the work in the background for us but it was wrong as it needs to refer to the active front end template (by default it refers to the current active template which will be wrong for the backend part)

dgrammatiko:
I will redo this PR with a twist, we can do better here

PR #30192, opened 2020-07-25
[4.0] Child templates 

dgrammatiko:
Pull Request is the foundation for #30149

### Summary of Changes
Introduces Inheritable templates solving numerous problem with the current state of templates. Read more on the linked issue above

This PR is complete in the sense that contains all the changes (even the upgrade path from J3) to introduce the new mode but without introducing the GUI and neither is moving any of the templates to the new mode. These can be done in following PR's

### Testing Instructions

Do not try to test this with patch tester, it won't work.
Install the PR's installable, you can d/l it here (at the bottom of this page)
If you want to use git command lines or the desktop app make sure that you'll do a fresh install or apply the updates (sql files)

Make sure you have the browser's console open so you can observe any missing assets.

If this went smooth, then it's a good sign as the installation is a seperate application with it's own template and although this PR is not touching the installation template everything works as before. B/C check 1

### Front End step 1

Go to the front end and vist the following urls:

- /index.php/blog
- /index.php/blog/3-welcome-to-your-blog
- /index.php/author-login (login here)
- /index.php?option=com_config&view=modules&id=109&Itemid=104 (or just try to edit the top left module)

Everything should be fine, no console logs or missinng icons, etc. B/C check 2

### Front End step 2

Download this [cassiopeia_overrides_v2.zip](https://github.com/joomla/joomla-cms/files/4977969/cassiopeia_overrides_v2.zip) and extract the contents and copy them in the folder `templates/cassiopeia/html`. Revisit the pages of the previous step. You should see in the pages some plain text (that will also break a bit the design) `Legacy Component view override`. You have to look for `Component`, `Module`, `Plugin` and `JLayout`. If so then the cascading of the templating is working correctly, in the previous step there was no overrides in this one the overrides were treated as top priority. Also cassiopeia has some overrides for the static assets so that was already tested (if you want you might open a normal 4.0 version in another winndow and check the number of assets per page between the 2). B/C check 3

### Front End step 3

Install the first new mode template (basically cassiopeia with another name: siteparent) from here [siteparent_v2.zip](https://github.com/joomla/joomla-cms/files/4977972/siteparent_v2.zip)

Go to the admin and make the siteparent template default and also assign all the menus to this template!!! Go to the front end and repeat the Front End step 1 and confirm that everything is as before. check 4

### Front End step 4

[...]

PR #31785, opened 2020-12-26
[4.0] Tinymce adopts child templates

dgrammatiko:
Pull Request for Issue # .  Part of the Child template effort

### Summary of Changes
TinyMCE respects the paths for child templates

### Testing Instructions
You will need to install the front end template https://github.com/joomla/joomla-cms/files/4977972/siteparent_v2.zip from https://github.com/joomla/joomla-cms/pull/30192
The template needs to be default for all the menus
Then copy the file /templates/system/css/editor.css to /media/site/siteparent/css/editor.css

Edit an article (either backend or front end) and observe that the editor.css is loaded in the iframe of tinyMCE

No, feature parity, backward compatibility not broken, Logic is not radically changed, just adopted for the child able templates, also there shouldn't be any performance degradation (it's almost the same amount of commands executed prior or after this PR)


plugins/editors/tinymce/tinymce.php
```diff
@@ -290,8 +290,26 @@ public function onDisplay(
...
+		// Default to legacy
+		$templates_path = 'templates' . '/' . $template->template;
 		$content_css    = null;
-		$templates_path = JPATH_SITE . '/templates';
+
+		if ("" !== $template->parent)
+		{
+			$templates_path = 'media/templates/site/' . $template->template;
+
+			if (!file_exists(JPATH_ROOT. '/' . $templates_path))
```
wilsonge: Do we not need to also check for a template override in the child template here

wilsonge:
I don't think the JPATH_ROOT usage is going to work for files in the media directory for admin templates

plugins/editors/tinymce/tinymce.php
dgrammatiko: The logic is like: we set the path to $tempalate->template (which is the child) then we check if the file exists there and if not we set it to the parent. Finally the path is verified or failing bad to the system one at line 341-361

dgrammatiko:
> I don't think the JPATH_ROOT usage is going to work for files in the media directory for admin templates

I had the same idea BUT the logic never uses the Admin template, it always refers to the active front end template the one that is used (check the where clause in the db query I modified, btw this is not changed here)

brianteeman:
Tinymce should be using the site template editor.css whether its in the front end or the admin

plugins/editors/tinymce/tinymce.php
```diff
@@ -290,8 +290,26 @@ public function onDisplay(
 			return '';
 		}
 
+		// Default to legacy
+		$templates_path = 'templates' . '/' . $template->template;
 		$content_css    = null;
-		$templates_path = JPATH_SITE . '/templates';
+
+		if ("" !== $template->parent)
```
```suggestion
		if ($template->parent !== '')
```

plugins/editors/tinymce/tinymce.php
dgrammatiko: This is hot garbage, meaning will never work as indented, my bad. I will come up with the appropriate logic later on

pabloarias:
I have tested this item :white_check_mark: successfully on e71d3e6e1886898e62c378c75585c24e5b124dca<br /><br />I've installed Siteparent template and made it default.

I've copied the file /templates/system/css/editor.css to /media/**templates**/site/siteparent/css/editor.css

Before install path, editor iframe loads /templates/system/css/editor.css but after install the path editor iframe loads /media/templates/site/siteparent/css/editor.css<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31785">issues.joomla.org/tracker/joomla-cms/31785</a>.</sub>

BertaOctech:
I tried to test this issue but I got an error whe I installed the template.

I installed the template, I selected it as default for all menu items, I emptied joomla cache and the browser cache.

When I loaded the frontend I received the following error

> Joomla\CMS\WebAsset\Exception\UnsatisfiedDependencyException:
> Unsatisfied dependency "bootstrap.js.bundle" for an asset "template.cassiopeia" of type "script"
> at at /libraries/src/WebAsset/WebAssetManager.php:1037<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31785">issues.joomla.org/tracker/joomla-cms/31785</a>.</sub>

dgrammatiko:
> Unsatisfied dependency "bootstrap.js.bundle" for an asset "template.cassiopeia" of type "script"

I think you have an outdated Joomla Beta version. Sor sometime now there's no `bootstrap.js.bundle`

BertaOctech:
> > Unsatisfied dependency "bootstrap.js.bundle" for an asset "template.cassiopeia" of type "script"
> 
> I think you have an outdated Joomla Beta version. Sor sometime now there's no `bootstrap.js.bundle`

Thank you for the tip, however, 

I updated my test site to ‎4.0.0-beta7, before testing.

I have now updatated to 4.0.0-beta8-dev. 

Some problem.

I will try on a fresh install if I have some time later on

dgrammatiko:
> I will try on a fresh install if I have some time later on

Maybe the provided template needs some update, so give it some time till I get something for you

brianteeman:
Sorry but this is not good.

Before this PR the editor.css file provided by the template would be editable in the template manager. By moving it to the media folder it is no longereditable.

dgrammatiko:
> Sorry but this is not good.

Sorry, but you totally missing the point here. For legacy templates (eg existing templates) the assets live in the template folder and the script will look there but for the new mode since all the static assets are moved into media/templates/... the script will look there. Nothing is broken here and also the editing will be possible once I finish that com_templates refactoring which is always get postponed due to other more urgent PRs...

brianteeman:
Definitely too late for such a major change in 4.0

dgrammatiko:
> Definitely too late for such a major change in 4.0

The code is already merged, the only thing missing is the UI, eg the refactoring of com_templates to expose these changes

socke300:
Is there still a test needed here?

dgrammatiko:
Yes but the code needs to be updated, so it can be tested (right now it can't)

plugins/editors/tinymce/tinymce.php
```diff
@@ -1124,4 +1037,152 @@ private function removeElementWithValue($array, $key, $value)
...
+	 * Helper function to resolve the given editor css
+	 *
+	 * @param   Object  $template  The ACTIVE FRONT END template object
+	 * @param   string  $cssFile   The filename for the editor css stylesheet
+	 *
+	 * @return string
+	 *
+	 * @since  __DEPLOY_VERSION__
+	 */
+	private function resolveTemplateEditorCSS($template, $cssFile): string
```
HLeithner, line 1143: Doesn't HTMLHelper::stylesheet with option pathOnly does the same?

dgrammatiko: Not in this context. We are replicating the same functionality but the difference is that the reference template IS NOT the current active template (eg the current active template in the back end is Atum but we want to get the files for the Front End active template). It could be way less code if https://github.com/joomla/joomla-cms/blob/a72b834b303d07bb51575394f0235dbba7f0679d/libraries/src/HTML/HTMLHelper.php#L382 was public and allowed to get the template as an argument, but that's not the case...

Fedik: It will not work, because `includeRelativeFiles` and so `HTMLHelper::stylesheet` is stick to active client and template.
However you could simplify the code a bit, define array of potential files, and loop:

```php
$potentialFiles = [
 'templates/' . $template->template . '/css/' . $cssFile,
 'media/templates/site/' . $template->template . '/css/' . $cssFile,
];

if ($template->parent)
{
    $potentialFiles[] = 'media/templates/site/' . $template->parent . '/css/' . $cssFile;
    $potentialFiles[] = 'templates/' . $template->parent . '/css/' . $cssFile;
}

$finalFile = null;

foreach ($potentialFiles as $potentialFile)
{
    if (is_file(JPATH_ROOT . '/' . $potentialFile))
    {
        $finalFile = $potentialFile;
        break;
    }
}

return $finalFile ? Uri::root(true) . '/' . $finalFile : $this->getSystemEditorCss();
```

dgrammatiko: @Fedik @HLeithner I think this PR needs some more touches but I will wait till there's a decision for https://github.com/joomla/joomla-cms/pull/35669 as in that one the editor.css could be created by the template instead of a hardcoded file so the whole logic will be a bit more different

dgrammatiko:
Replaced by https://github.com/joomla/joomla-cms/pull/36011

PR #31692, opened 2020-12-17
[4.0] run gzip

brianteeman:
There is a node command to gzip the assets

`npm run gzip`

When testing on windows it doesn't work (no idea about linux/mac)

After applying this PR then running the same command and it will work. (note it will appear to hang at the end - just be patient its slow)

Discovered while testing #31635

PR #31800, opened 2020-12-29
[4.0] Move gzip command to work on windows

wilsonge:
Successor PR to https://github.com/joomla/joomla-cms/pull/31692 which doesn't work on Mac

There is a node command to gzip the assets

`npm run gzip`

When testing on windows it doesn't work although does on Mac

After applying this PR then running the same command and it will work. (note it will appear to hang at the end - just be patient its slow)

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -3,14 +3,14 @@
  *
  * To get the complete functional media folder please run
  *
- * npm install
+ * npm ci
  *
  * For dedicated tasks, please run:
  * node build.js --build-pages  === will create the error pages (for incomplete repo build PHP+NPM)
  * node build.js --copy-assets  === will clean the media/vendor folder and then will populate the folder from node_modules
  * node build.js --compile-js   === will transpile ES6 files and also uglify the ES6,ES5 files
  * node build.js --compile-css  === will compile all the scss defined files and also create a minified version of the css
- *
+ * node build.js --gzip  === will create gzip files for all the minified stylesheets and scripts.'
  */
 
 // eslint-disable-next-line import/no-extraneous-dependencies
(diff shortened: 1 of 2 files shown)


build/build.js
```diff
@@ -102,3 +104,8 @@ if (Program.compileJs) {
 if (Program.watch) {
   watch.run();
 }
+
+// Compress/transpile the javascript files
```
dgrammatiko:
```suggestion
// Gzip js/css files
```

dgrammatiko:
I have tested this item :white_check_mark: successfully on fa39d9fd3cd0fbff7211c567e26dc010f6b108db<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31800">issues.joomla.org/tracker/joomla-cms/31800</a>.</sub>

build/build.js
```diff
@@ -42,6 +43,7 @@ Program
   .option('--compile-js, --compile-js path', 'Handles ES6, ES5 and web component scripts')
   .option('--compile-css, --compile-css path', 'Compiles all the scss files to css')
   .option('--watch', 'Watch file changes and re-compile (ATM only works for the js in the media_source).')
+  .option('--gzip', 'Precompress all the minified stylesheets and scripts.')
```
brianteeman: Not quite sure what the correct word is but its definitely not precompress. Pre- is something that takes before.

```suggestion
  .option('--gzip', 'Compress all the minified stylesheets and scripts.')
```

brianteeman:
>  (note it will appear to hang at the end - just be patient its slow)

Beyond slow - a tortoise is faster. Can't there be any console output?

Otherwise it works ok

dgrammatiko:
> Beyond slow - a tortoise is faster. Can't there be any console output?

Well if it was way faster the output would be irrelevant (unless there's an error). BTW https://github.com/joomla/joomla-cms/pull/31635 is many times faster than the synchronous WASM code used atm (WASM was used because zopfli didn't had a 32bit binary so the Drone was failing, a sad story but then again this was the only place that Joomla actually used WASM)...

PR #30802, opened 2020-09-29
Rename incorrectly cased files

SharkyKZ:
Closes https://github.com/joomla/joomla-cms/issues/23725 and  https://github.com/joomla/joomla-cms/issues/30736 .

### Summary of Changes

This PR attempts to implement case sensitive file rename and rename some incorrectly cased files.

### Testing Instructions

Coming soon.

### Documentation Changes Required

No.

Issue #29794, opened 2020-06-27
[4.0] Composer Deprecation Notices


mbabker:
Run `composer install` with Composer 1.10.7 (current stable as of this post).  There are deprecation notices regarding 3 CMS classes not complying with PSR-4 standards that will not resolve correctly when using Composer 2 (2 are because of inconsistent casing between the FQCN and file path, one is because a class is stuffed into a folder unexpectedly).  These issues should be addressed before Composer 2 is released.

PhilETaylor:
```
Deprecation Notice: Class Joomla\Tests\Unit\Libraries\Cms\Image\ImageInspector located in ./tests/Unit/Libraries/Cms/Image/stubs/ImageInspector.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer.phar/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class Joomla\CMS\Form\Rule\SubformRule located in ./libraries/src/Form/Rule/SubFormRule.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer.phar/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class Joomla\CMS\Filesystem\Support\StringController located in ./libraries/src/Filesystem/Support/Stringcontroller.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer.phar/src/Composer/Autoload/ClassMapGenerator.php:201


Also (but upstream) 


Deprecation Notice: Class Flow\JSONPath\Test\JSONPathSliceAccessTest located in ./libraries/vendor/flow/jsonpath/tests/JSONPathSliceAccessTest.php does not comply with psr-0 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer.phar/src/Composer/Autoload/ClassMapGenerator.php:201
Deprecation Notice: Class Flow\JSONPath\Test\JSONPathTest located in ./libraries/vendor/flow/jsonpath/tests/JSONPathTest.php does not comply with psr-0 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer.phar/src/Composer/Autoload/ClassMapGenerator.php:201
[...]

PR #31804, opened 2020-12-30
[4.0] Fix Composer Deprecation Notices

bembelimen:
Pull Request for Issue #29794 .

### Summary of Changes
Fixes the composer deprecations using the 3.x code from @SharkyKZ here: #30802 

### Testing Instructions
Test 1: Update a Joomla! 3.10 installation with the 4.0 installation of this PR (see at the end for download link)
Test 2: Run ```composer install``` for this PR
Test 3: Test unit tests

Result 1: (especially on windows systems) the renamed files from this PR have the incorrect case (or old files are still there)
Result 2: composer does throw deprecation notices

Result 1: (especially on windows systems) the renamed files from this PR have the correct case
Result 2: composer does not throw any deprecation notices
Result 3: unit test are working

Thanks @SharkyKZ for the 3.0 code
Thanks @HLeithner for helping with this PR


administrator/components/com_joomlaupdate/restore_finalisation.php
```diff
@@ -72,7 +72,10 @@ function finalizeRestore($siteRoot, $restorePath)
 			// Remove obsolete files - prevents errors occurring in some system plugins
 			if (class_exists('JoomlaInstallerScript'))
 			{
-				(new JoomlaInstallerScript)->deleteUnexistingFiles();
+				$installer = new JoomlaInstallerScript;
+
+				$installer->deleteUnexistingFiles();
+				$installer->fixFilenameCasing();
```
wilsonge: Not quite - it's a protected function :)

bembelimen: Thx

astridx:
Is composer version 1 important or can I test it with composer 2?

administrator/components/com_joomlaupdate/restore_finalisation.php
HLeithner: I don't think that this should be here, there is no reason to have it on 2 places. I thinks It's also wrong in the databasemodel (both deleteUnexistingfiles and fix filename casing).

bembelimen:
> 
> 
> Is composer version 1 important or can I test it with composer 2?

If you get errors/deprecations there, too, then yes.

administrator/components/com_joomlaupdate/restore_finalisation.php
wilsonge: I think having it here is the correct place because it runs before the first page load on the new J4 codebase - better than loading the first page as a general rule (as we may always need to add more files to this list in the future for composer things)

HLeithner:
@wilsonge drone is broken now^^

wilsonge:
This test is in a different class to the one I modified!?

HLeithner:
The current error message expains it better:
```
1) Joomla\Tests\Unit\Libraries\Cms\Image\ImageTest::testGetFilterInstance
RuntimeException: The Filterinspector image filter is not available.
```

Since you removed the Instance creating it seems the 2 files are related.

administrator/components/com_installer/src/Model/DatabaseModel.php
```diff
@@ -321,6 +321,7 @@ public function fix($cids = array())
 			{
 				$installer = new \JoomlaInstallerScript;
 				$installer->deleteUnexistingFiles();
+				$installer->fixFilenameCasing();
```
HLeithner:
```suggestion
```

richard67:
@bembelimen Could you solve conflicts? Maybe it needs to adapt to the recently merged "dry run" feature.

In addition, it needs to adapt to changes I've made with my PR #32006 , which replaces PR #30802 and contains additional changes on comments and a fix for OSX. If you want I can help with that and make a PR for you after you have done the rest.

richard67:
@bembelimen I was checking the conflict shown here for the script.php file and have noticed something.

With PR #31943 , some logging and a dry run option have been added to the files and folder deletion.

This PR here now calls at the end of that function the new one for moving files with wrongly cased file names.

It might make sense to refactor this PR a bit to do the same (logging and dry run option) for the files moving function.

The dry run later could be used e.g. in 4.1 to have something like the database schema check also for the "are there files or folders to be deleted or files to be renamed" checks, with corresponding "Fix" buttons maybe, to de-couple that a bit from database schema fixing.

richard67:
@bembelimen I've allowed myself to fix the conflict.

administrator/components/com_admin/script.php
```diff
@@ -7142,4 +7144,69 @@ private function convertBlogLayouts()
...
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function fixFilenameCasing()
+	{
+		$files = array(
+			// 3.10 changes
+			'libraries/src/Filesystem/Support/Stringcontroller.php' => 'libraries/src/Filesystem/Support/StringController.php',
+			'libraries/src/Form/Rule/SubFormRule.php' => 'libraries/src/Form/Rule/SubformRule.php',
+			// __DEPLOY_VERSION__
+			'media/vendor/skipto/js/skipTo.js' => '/media/vendor/skipto/js/skipto.js',
```
```suggestion
			'media/vendor/skipto/js/skipTo.js' => 'media/vendor/skipto/js/skipto.js',
```

wilsonge:
Thanks!

Issue #30950, opened 2020-10-06
[4][mobile] Mail Templates Needs Mobile Review


PhilETaylor:
### Steps to reproduce the issue

Visit com administrator/index.php?option=com_mails&view=templates on a mobile device

### Expected result

like the rest of Joomla 4, nicely laid out and responsive to the screen size 

### Actual result

unresponsive table 

![Screen Recording 2020-10-06 at 11 08 14 pm](https://user-images.githubusercontent.com/400092/95265322-e0e8cf80-0828-11eb-989e-fe6987fcb8c0.gif)

brianteeman:
Please test #31823

PR #31823, opened 2021-01-01
[4.0] Mail Templates mobile

brianteeman:
Pull Request for Issue #30950

Simple PR to hide two columns from the mobile view. The mobile view can never contain everything and the two columns removed are just alternative ways to create/edit templates per language

### Before
image

### After
image


wilsonge:
Thanks!

brianteeman:
Thx

PR #31844, opened 2021-01-02
[4.0] correction for 31832

brianteeman:
There was no need to change the language key. It was consistent with all the similar strings which followed a set structure. (and if changed it should have been _HEADING not _HEADLINE)

All that was needed was to change the value of the string


wilsonge:
Thanks!

brianteeman:
thanks

"merge in haste repent at leisure"

PR #31854, opened 2021-01-04
[4.0] Skip-to plugin major upgrade [a11y]

brianteeman:
This is a draft PR to give a heads up about this new version especially as there are new language strings.

Amongst other things it addresses @infograf768 comments about not everything be translatable and an old comment from @clodder about the dropdown script which is no longer being used.

In the process of creating this PR I discovered a few bugs in the upstream code (pr submitted) and that there github is ahead of the release on npm yet with a lower package number.

As a result this PR is not completed and its not working (hence its a draft).

This release is much much better under the hood. I have only implemented some of the features (ie not the important and all distinction) as they're more suitable to a content rich site than an admin ui.

For now you can only do a code review. As soon as I get a response to https://github.com/paypal/skipto/issues/68 I will be able to complete the release

More information http://paypal.github.io/skipto/

### DIFF
--- a/build/build-modules-js/settings.json
+++ b/build/build-modules-js/settings.json
@@ -548,30 +548,21 @@
         "name": "skipto",
         "licenseFilename": "LICENSE.md",
         "js": {
-          "src/js/skipTo.js": "js/skipTo.js",
-          "src/js/dropMenu.js": "js/dropMenu.js"
+          "src/js/skipto.js": "js/skipto.js"
         },
         "css": {
-          "src/css/SkipTo.css": "css/SkipTo.css"
+          "src/css/skipTo.css": "css/skipTo.css"
         },
         "provideAssets": [
           {
             "name": "skipto",
             "type": "style",
-            "uri": "SkipTo.css"
-          },
-          {
-            "name": "skipto.dropmenu",
-            "type": "script",
-            "uri": "dropMenu.js",
-            "attributes": {
-              "defer": true
-            }
+            "uri": "skipTo.css"
           },
           {
             "name": "skipto",
             "type": "script",
-            "uri": "skipTo.js",
+            "uri": "skipto.js",
             "attributes": {
               "defer": true
             }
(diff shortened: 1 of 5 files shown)


plugins/system/skipto/skipto.php
```diff
@@ -52,24 +52,46 @@ public function onAfterDispatch()
 			return;
 		}
 
+		// Are we in a modal?
+		$input   = Factory::getApplication()->input;
+		if ($input->get('tmpl', '', 'cmd') === 'component')
```
richard67:
```suggestion

		if ($input->get('tmpl', '', 'cmd') === 'component')
```

build/build-modules-js/settings.json
```diff
@@ -548,30 +548,21 @@
         "name": "skipto",
         "licenseFilename": "LICENSE.md",
         "js": {
-          "src/js/skipTo.js": "js/skipTo.js",
-          "src/js/dropMenu.js": "js/dropMenu.js"
+          "src/js/skipto.js": "js/skipto.js"
         },
         "css": {
-          "src/css/SkipTo.css": "css/SkipTo.css"
+          "src/css/skipTo.css": "css/skipTo.css"
```
HLeithner, line 554: if the case is changed we need https://github.com/joomla/joomla-cms/pull/31804 first and add the corresponding files to the rename function

brianteeman: This is still up in the air until upstream fixes their npm etc.

They do the css differently to the way we did it so this line will probably be deleted anyway

HLeithner: OK but the changed the .js file too

brianteeman: yes I know. I cant do anything about it until they fix their end

dgrammatiko:
@brianteeman fwiw if https://github.com/paypal/skipto/pull/69 is ever gets merged then the whole inline part of the plugin is useless (not to mention that the option type="module" is not doing what you think it does, in sort the plugin will NEVER execute in browsers without module support)

brianteeman:
@dgrammatiko thanks for #69 that will really make it much better

As for the module bit, I don't actually have a clue what it does.  @Fedik wrote that

brianteeman:
Good news is that I have it working locally

Bad news is that they made even more breaking changes and they put them in a patch release. So as soon as that is fixed and they release 4.0 I will be ready

brianteeman:
closed see #32043

Issue #30691, opened 2020-09-19
Skip To behaviour on Popups


adj9:
I briefly introduce the Skip To which is not very well known. it is a pop-up menu to use Joomla from the keyboard and get access to the area of interest faster. Entering the administrative area
1. Click tab 
2. Click the down arrow
<img width="303" alt="1" src="https://user-images.githubusercontent.com/5918444/93670439-b2829a80-fa9b-11ea-96b9-ac025bfa142c.png">

This menu appears at the top left

### Steps to reproduce the issue
In a PopUp, for example, to insert the image, you need to double click the tab button.

<img width="1181" alt="2" src="https://user-images.githubusercontent.com/5918444/93670552-78fe5f00-fa9c-11ea-95b0-9fd9f2258d8c.png">

### Expected result
I expect that it is not necessary to type the tab button twice to go to the second step.

### Actual result
In a PopUp, you need to double click the tab button.

richard67:
@brianteeman Could you have a look on this issue as you are very familiar with the Skip To thing? Thanks in advance.

brianteeman:
The main problem is that we are NOT using real modals but are instead loading an iframe and as a result loading everything again

richard67:
Is there any way to fix it without changing from iframes to real modals (what I expect to happen with version 5 maybe ;-) )?

brianteeman:
Not as far as I am aware. This is a longstanding architectural problem which no one has ever been interested in fixing. Finally now that people can SEE the issue they are waking up to the fact that using the iframe is wrong

PhilETaylor:
Also you can see the Skip To Menu on mobile by pulling down... unexpected.

![Screen Recording 2020-10-06 at 10 55 02 pm](https://user-images.githubusercontent.com/400092/95264334-0bd22400-0827-11eb-9ad9-bfe1dee45c92.gif)

brianteeman:
@Fedik is there perhaps something we can do with webassetsmanager here. The problem is that the modal is an iframe of the component.php in the template and this loads a full set of assets

I'm wondering if there is a way to unset an asset on the component.php page

This problem is not just for the skipto plugin but some other system plugins as well

Fedik:
> I'm wondering if there is a way to unset an asset on the component.php page

yes sure, If the asset was loaded before template file (before rendering of component.php):
```php
if ($wa->assetExists('script', 'blabla'))
{
  $wa->disableScript('blabla');
}
```

[...]

PR #31863, opened 2021-01-05
[4.0] Disable accessibility plugin in modal

brianteeman:
To test enable the system accessibility plugin
In an article in the images & links tab click on select to open the modal

### Before
image

### After
image

In the new skipto plugin I will apply the same fix and that will address #30691

### DIFF
--- a/plugins/system/accessibility/accessibility.php
+++ b/plugins/system/accessibility/accessibility.php
@@ -52,6 +52,13 @@ class PlgSystemAccessibility extends CMSPlugin
 			return;
 		}
 
+		// Are we in a modal?
+		$input   = Factory::getApplication()->input;
+		if ($input->get('tmpl', '', 'cmd') === 'component')
+		{
+			return;
+		}
+
 		// Load language file.
 		$this->loadLanguage();
 
--- a/plugins/system/accessibility/accessibility.php
+++ b/plugins/system/accessibility/accessibility.php
@@ -53,8 +53,7 @@ class PlgSystemAccessibility extends CMSPlugin
 		}
 
 		// Are we in a modal?
-		$input   = Factory::getApplication()->input;
-		if ($input->get('tmpl', '', 'cmd') === 'component')
+		if ($this->app->input->get('tmpl', '', 'cmd') === 'component')
 		{
 			return;
 		}


plugins/system/accessibility/accessibility.php
```diff
@@ -52,6 +52,13 @@ public function onBeforeCompileHead()
 			return;
 		}
 
+		// Are we in a modal?
+		$input   = Factory::getApplication()->input;
+		if ($input->get('tmpl', '', 'cmd') === 'component')
```
HLeithner:
```suggestion
		if ($this->app->input->get('tmpl', '', 'cmd') === 'component')
```

chmst:
I have tested this item :white_check_mark: successfully on 767a6638fe93bb78c7722f2190023f168eecd36b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31863">issues.joomla.org/tracker/joomla-cms/31863</a>.</sub>

plugins/system/accessibility/accessibility.php
richard67: Plus an empty line before the "if" for fixing PHP CS in drone.

ghost:
I have tested this item :white_check_mark: successfully on 767a6638fe93bb78c7722f2190023f168eecd36b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31863">issues.joomla.org/tracker/joomla-cms/31863</a>.</sub>

plugins/system/accessibility/accessibility.php
HLeithner: This isn't needed after this change, because the line is now a comment.

brianteeman:
Thanks @HLeithner

plugins/system/accessibility/accessibility.php
```diff
@@ -52,6 +52,12 @@ public function onBeforeCompileHead()
 			return;
 		}
 
+		// Are we in a modal?
+		if ($this->app->input->get('tmpl', '', 'cmd') === 'component')
```
Fedik, line 56: I think better idea do not load it when any `tmpl` are requested. 
Then it will be more safe for Users custom `tmpl`

brianteeman: not convinced - lets leave it as this for now. It can always be improved at a later date if required

wilsonge:
Thanks!

brianteeman:
Thanks

PR #31885, opened 2021-01-08
[4.0] Minor Cleanup of workflow code

wilsonge:
### Summary of Changes
- Allows injection of the CMSApplication + DatabaseDriver classes into the Workflow class (this is to allow future unit testing of the class).
- Cleanup constructor - the options array (and setOptions method) weren't used by the class anymore - they've been removed in the large refactor. It now just requires a string of the extension name.

### Testing Instructions
Workflow related functionalities unchanged. Install sample data and create and modify workflows 

Possibly. Not sure whether the documentation covers the constructor. It's likely it only covers using the trait.


bembelimen (approved):
LGTM 👍

PR #31886, opened 2021-01-08
[4.0] Setup legacy CLI app with proper dispatcher and container

laoneo:
### Summary of Changes
The CLI app gets deprecated by the new Console app. Still a lot of extensions rely on it. To be as much backwards compatible as possible, some setup work is needed. This pr adds the functionality to load the extensions namespace map, and setup properly the dispatcher and some session variables, copied from the console app.

### Testing Instructions
- Copy the discover app from here https://github.com/Digital-Peak/DPDocker/blob/master/webserver/scripts/discoverapp.php to the joomla root.
- Remove the lines 25 to 37 in the file discoverapp.php of your joomal root
- Link an extension with a symlink
- Execute `php discoverapp.php` in your Joomla root folder

It crashes and the extension doesn't get installed.

The extension is installed.

### DIFF
--- a/libraries/src/Application/CliApplication.php
+++ b/libraries/src/Application/CliApplication.php
@@ -116,16 +116,25 @@ abstract class CliApplication extends AbstractApplication implements DispatcherA
 
 		$container = $container ?: Factory::getContainer();
 		$this->setContainer($container);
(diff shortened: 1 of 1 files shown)


wilsonge:
Thanks!

Issue #31901, opened 2021-01-08
[4][CLI] update:joomla:remove-old-files incorrectly states "files removed"


PhilETaylor:
### Steps to reproduce the issue

Repeatedly run the CLI command 

`php cli/joomla.php update:joomla:remove-old-files`

### Expected result

After the first run (or even on the first run depending on the actual result) it should say NO files removed, because the command has NOT removed files. 

### Actual result

Command's success message always states that "Files removed" which is factually incorrect. Nothing was removed. 

<img width="852" alt="Screenshot 2021-01-08 at 17 36 58" src="https://user-images.githubusercontent.com/400092/104046534-21653a00-51d8-11eb-9926-40a658f3853f.png">

### System information (as much as possible)

```
PHP 7.4.13 (cli) (built: Dec 23 2020 11:21:05) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.13, Copyright (c), by Zend Technologies
```

alikon:
closing as we have a pr #31943

PR #31914, opened 2021-01-08
[4.0] cli remove old files

brianteeman:
Quick fix to change the success message

Pull Request for Issue #31901

### DIFF
--- a/libraries/src/Console/RemoveOldFilesCommand.php
+++ b/libraries/src/Console/RemoveOldFilesCommand.php
@@ -51,7 +51,7 @@ class RemoveOldFilesCommand extends AbstractCommand
 
 		(new \JoomlaInstallerScript)->deleteUnexistingFiles();
 
-		$symfonyStyle->success('Files removed');
+		$symfonyStyle->success('There are no old files left to remove');
 
 		return 0;
 	}


PhilETaylor:
the tools doesn't only remove files, it[ removes folders too ](https://github.com/joomla/joomla-cms/blob/4c434000c9ba717602846a144b709e0a66a3631d/administrator/components/com_admin/script.php#L2310)- so there is much more that could change...

The administrator/components/com_admin/script.php also outputs error messages that are not caught by the console command output streamer as well.

brianteeman:
Folders is a level of pedantry I'm not going to entertain. The pr addresses the reported issue.

ceford:
I have tested this item :white_check_mark: successfully on 60aaf53e2ffcbedbbd4f0722ccadd2d89bcbde74<br /><br />Before the PR the command response is:
 [OK] Files removed                                                                                                     
After the PR the command response is:
 [OK] There are no old files left to remove                                                                             
So just changing the response message works as described.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31914">issues.joomla.org/tracker/joomla-cms/31914</a>.</sub>

alikon:
this is not correct
put a dummy file at `'/administrator/components/com_actionlogs/actionlogs.php',`
run `php cli/joomla.php update:joomla:remove-old-files`

you'll get always the message
 [OK] There are no old files left to remove

brianteeman:
@alikon thats not related. The file you added is not an "old file"

alikon:
it is an old file https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_admin/script.php#L476

brianteeman:
In that case there is a problem with the script if its not removing files. All this PR does is to modify the string the script outputs on completion. Any issues with the script not working is completely off topic and should have its own issue.

alikon:
> All this PR does is to modify the string the script outputs on completion. Any issues with the script not working is completely off topic and should have its own issue.



which is wrong imho they are strictly related
the correct command output should be 

- files/folder removed if 1 or more file/folder file was removed
- no files/folders to remove if no file/folder file was removed

and yes the `script.php` must be modifyed to return at least a boolean

PhilETaylor:
If the script returned an array of files and their statues, and information on the action then the command could show WHICH files it deleted, WHICH it checked etc. 

Then the correct use of command verbose level could be used (-v -vv -vvv) to display a summary, more detail and debug information

Verbose is what makes a command so useful.

The command deletes folders too and therefore this PR just makes an inaccurate string just as inaccurate

PhilETaylor:
The command should also have a --dry-run option

brianteeman:
Got to love bikeshedding

alikon:
still need to fully understand what _bikeshedding_ means 
:smile:  but...

PhilETaylor:
Basically he's using it as an abusive term towards me because he doesn't have the skills to correctly address the issue and wants to belittle me for wanting the best from joomla and for offering my valid comments on how joomla can do much better, which he doesn't see as valid - unless I invest my time in writing the PR myself. I've come to expect nothing less. 

All the commands in joomla are currently a joke and don't fully implement Symfony command logic in a way that people that have been using Symfony based commands for years come to expect as a minimum. This is just the latest example. 

The commands do the bare minimum currently and don't honour the built in modes like verbosity 

"Doing something right" is better than ignoring the issue. 

And like you say, it seems this command doesn't even do what it should - delete files!

brianteeman:
I am using it as a term to indicate that the conversation you are having is unrelated to the actual issue. As for your consistent criticism of others people work and only being willing to contribute a spellcheck - everyone knows that you are just a leach who takes from joomla without giving back.

PhilETaylor:
Screenshot 2021-01-09 at 13 20 23

PR #31962, opened 2021-01-09
[4.0] PHP support status.

brianteeman:
Updates the php version check plugin to make it aware of php 8.0

Code review

PS should we still be recommending 7.3? @wilsonge


wilsonge:
Can you do this to staging instead please and I'll ensure it get's merged up. It belongs in the 3.x branch too.

ceford:
I see the code change - but where does the result appear if I am using an out of date PHP? Is there a way to simulate that?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/31962">issues.joomla.org/tracker/joomla-cms/31962</a>.</sub>

brianteeman:
@wilsonge See #31977

Closing this here

PR #31988, opened 2021-01-10
s/deletion/uninstall

PhilETaylor:
### Summary of Changes

Change the word deletion to be uninstall

There is no concept of "deleting" or "deletion" of a plugin or extension in Joomla. 

It is consistently referred to as "Uninstalling" or "Uninstall" throughout Joomla 

Screenshot 2021-01-10 at 17 35 36

### Testing Instructions

Text change only


HLeithner:
Thanks

PR #32002, opened 2021-01-11
[4.0] Remove old sample data images

brianteeman:
Removes the parks and fruit folders from the images/sampledata folder as they are no longer used.

They are already excluded from update packages.

They do not need to be added to the list of files to remove on upgrade as they might be used by a user.

(They are used in the test sampledata plugin but thats not shipped with Joomla and is already pretty broken)


PhilETaylor:
Something wrong with this PR. Its not merged correctly.

brianteeman:
closed. will redo and re-open later

PR #32003, opened 2021-01-11
[4] Remove return after an exception (copy and paste legacy issue)

PhilETaylor:
### Summary of Changes

On code review.

Before Joomla 4 this was a `JError::raiseError(...)` call and not an `Exception` so the `return` made sense then

Now we throw `Exception` there is no way to reach the `return false;` and therefore its redundant unreachable code


wilsonge:
Thanks!

infograf768:
👍

PR #31997, opened 2021-01-10
[4] remove redundant code in frontend template manager

PhilETaylor:
### Summary of Changes

remove code smell and correctly load the form for frontend template config
$form was created twice, directly overwriting itself as soon as it was loaded.

### Testing Instructions

Install testing data on Joomla 4
login to FRONTEND 
click "Template Settings" in the right hand side menu

### Actual result BEFORE applying this Pull Request

Page loads

### Expected result AFTER applying this Pull Request

Page loads - but with less code to power it

Screenshot 2021-01-10 at 19 38 00

### Documentation Changes Required

none - code review only

PR #32004, opened 2021-01-11
[4.0] Fix cs in #31997

brianteeman:
I guess @wilsonge missed that the tests were failing when he merged this

If it passes the tests (and it should) then it can be merged


wilsonge:
I did. Still can’t see it in the GitHub diff even now I know it’s there o.O

Phpcs passes so merged. Thanks!

PR #32007, opened 2021-01-11
[4.0] remove images

brianteeman:
Removes the parks and fruit folders from the images/sampledata folder as they are no longer used.

They are already excluded from update packages.

They do not need to be added to the list of files to remove on upgrade as they might be used by a user.

(They are used in the test sampledata plugin but thats not shipped with Joomla and is already pretty broken)


HLeithner:
That files are used in the testing sampledata, if you remove the files you also have to remove the sampledata doesn't make sense to keep only the half of the package.

brianteeman:
testing Sampledata plugin is not shipped with Joomla so no point in shipping the images with joomla

HLeithner:
> testing Sampledata plugin is not shipped with Joomla so no point in shipping the images with joomla

that's maybe true but removing only a part of it makes no sense, 2 possible solutions
1. fixing the testing sample data
1. deleting not only a part of it and delete everything (I suspect its only the directory plugins/sampledata/testing and the reference in the build script)

of course moving the images to another folder and fixing the path in the sampledata and exclude them from build process is also an option

brianteeman:
> fixing the testing sample data

Even when fixed there is no need to ship the images

> deleting not only a part of it and delete everything (I suspect its only the directory plugins/sampledata/testing and the reference in the build script)

That is what is currently done in the build scripts ;)

To keep you happy I will just leave these images as they are and remove them in the build scripts.

PR #32009, opened 2021-01-11
[4.0] Remove sampledata images

brianteeman:
Add the sampledata parks and fruitshops image folders to the list of things not to be included in a build

This means they are still available for use in the sampledata testing plugin as that is only available in git clones


HLeithner:
thx

PR #32010, opened 2021-01-11
Let's use original assets from parent template

NetAssoPro:
Cannot use the asset's heritage of a parent folder without make a copy of all assets in /media/templates/<application>/<parent_template>

### Summary of Changes

Change the path to use the asset's heritage of a parent template

### Testing Instructions

- Create a simple sub template with only an xml file
- Try to run the website with this subtemplate

- The assets of the parent are not loaded

- The assets are corretly loaded

### DIFF
--- a/libraries/src/HTML/HTMLHelper.php
+++ b/libraries/src/HTML/HTMLHelper.php
@@ -465,7 +465,7 @@ abstract class HTMLHelper
 
 							if (empty($found))
 							{
-								$found = static::addFileToBuffer("$templaPath/$template->parent/$folder/$file", $ext, $debugMode);
+								$found = static::addFileToBuffer(JPATH_THEMES . "/$template->parent/$folder/$file", $ext, $debugMode);
 							}
 						}
 						else


dgrammatiko:
What are the contents of your XML? 
You need to define the parent template in order the cascading will work as expected. Eg
```xml
<extension type="template" client="site">
  <name>somename</name>
  <version>1.0</version>
  <creationDate>July 2020</creationDate>
  <author>dGrammatiko</author>
  <copyright>Copyleft</copyright>
  <description>Demo template to test Child templates PR</description>

// This part is ESSENTIAL
  <inheritable>0</inheritable>
  <parent>theNameOfTheParentTemplate</parent>
</extension>
```

you can check this child template https://github.com/joomla/joomla-cms/files/4977974/sitechild_v2.zip

that is inheriting from https://github.com/joomla/joomla-cms/files/4977972/siteparent_v2.zip

NetAssoPro:
Hi,
My main it's to overide some part of the template ATUM. 
My XML is 
```xml<?xml version="1.0" encoding="utf-8"?>
<extension type="template" client="administrator">
	<name>xxx</name>
	<version>1.0</version>
	<creationDate>2021</creationDate>
	<author>NAP</author>
	<description>Template ADMIN xxx</description>
	<parent>atum</parent>
	<inheritable>0</inheritable>
	<media destination="templates/administrator/xxx" folder="media">
		<folder>css</folder>
		<folder>scss</folder>
	</media>
</extension>
```

dgrammatiko:
@NetAssoPro sorry but the child templates **EXPECTS** the static assets **stored at the media folder in a path like: `media/templates/admin/templatename` or  `media/templates/site/templatename`** check the provided templates above.

NetAssoPro:
It's sad :(
You lost all the advandage of overiding if you need to copy main files in a directory to override them in an other diretory...
You don't think so ?

dgrammatiko:
> You lost all the advandage of overiding if you need to copy main files in a directory to override them in an other diretory...
> You don't think so ?

Not really. The child templates **should** work **only** with compatible templates. Atum is not inheritable (yet), check the xml file, there's no line `<inheritable>1</inheritable>`. This means that **you shouldn't create a child from this template**. So basically you tried to do something that we explicitly wanted to avoid: conflicting old templates with the new ones (that support Childs). This would been easier if the GUI part was in place (I'm still working on it, although it takes longer than I expected)

NetAssoPro:
Ok, i understand. Really thanks for your fast feedback. I will let my hack ,waiting the ATUM inheritable.
I'm not sure the GUI will be enought in my case, i need some advanced things (like use Fontawesome pro instead of Fontawesome free) and extend the possibilities to custom template colors.

PR #31999, opened 2021-01-10
[4.0] Fix smtp test mail and migrate related jquery code to es6

HLeithner:
Pull Request for Issue #31981 .

### Summary of Changes
* Make password lock and show exclusive
* Fix upmerge typo issue
* Replace jquery code with es6 and move to passwordview js file

### Testing Instructions
* Test password fields
* Test smtp email settings as in #31724

### Actual result BEFORE applying this Pull Request
Smtp test mail with password didn't worked
View Password Icon and Modify button was visible at the same time which doesn't work.

### Expected result AFTER applying this Pull Request
Test E-Mail settings work and only "Modify" or "Show Password" Icon exists and work

PR #32017, opened 2021-01-12
[4.0] Remove jQuery from password field

laoneo:
Pull Request to remove jQuery, as introduced in #31672.

### Summary of Changes
Changes the password field lock functionality to plain vanilla. It also doesn't depend on the Bootstrap class "active" but on the disabled state.

### Testing Instructions
- Open the back end global configuration
- Click on the Server tab
- Modify the password

Password can be changed.

Password can be changed.


wilsonge:
Sorry Harald got there first in https://github.com/joomla/joomla-cms/pull/31999 - are you able to test that?

PR #32043, opened 2021-01-15
[4.0] skipto accessibility plugin major upgrade v4.0

brianteeman:
SkipTo is a replacement for your old classic "Skipnav" link, (so please use it as such)! The SkipTo script creates a drop-down menu consisting of the links to important landmarks and headings on a given web page. The menu makes it easier for keyboard and screen reader users to quickly jump to the desired location by simply choosing it from the list of options.

### Benefits
- All users can get an outline of the content on the page.
- Screen reader users can get a higher level navigation menu without having to use the screen reader landmark and header navigation commands which typically include longer lists of lower level headings and less used landmarks.
- Keyboard only users can more efficiently navigate to content on a page.
- Speech recognition users can use the menu to more efficiently navigate to content on a page.

### How it works
- The SkipTo menu button is the first tabbable element on the page, and it is configured not to be visible when the page is loaded, the menu button becomes visible when it receives focus.
- Once the keyboard focus is on the menu button, pressing the ENTER or the SPACEBAR key will pull down the list of important landmarks and headings on the page.
- If you decide to reach the menu again, simply press the built-in access key alt+9

This plugin is enabled by default for the admin and can optionally be enabled for the frontend
This version addresses previous concerns about an additional dropdown menu js.
This version addresses previous concerns that the landmarks were not translatable.

To facilitate upgrades from previous beta releases I have added the list of removed files.

Thanks to the help of the people at https://github.com/paypal/skipto for adding some joomla specific changes

### Testing
If testing from github you will need to run `npm i` to get the updated scripts

image


toivo:
I have tested this item :white_check_mark: successfully on a8c9f816e763f1e507b9451cab01f7fba4424516<br /><br />Testing successfully using the downloaded package Joomla_4.0.0-beta7-dev+pr.32043-Development-Full_Package.zip.

Test environment Windows 10, Wampserver 3.2.4
Apache 2.4.46a, MySQL 8.0.22, PHP 8.0.1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32043">issues.joomla.org/tracker/joomla-cms/32043</a>.</sub>

wilsonge:
Thanks!

brianteeman:
@wilsonge did you double check the code for the upgrades as there is case sensitive changes there

wilsonge:
I'd completely forgotten that's why we'd sat on this in the old PR 🤦 Thanks for reminder. I've added it into here https://github.com/joomla/joomla-cms/pull/31804/files#diff-db7eb77540ff419bd7e6557d2779f4cf1e31158c20cb4b63dbbe8d6c426385adR6983 as I think this should ship in the next few days

richard67:
The change in `administrator/components/com_admin/script.php` made by this PR is wrong. It adds the files to be deleted to the array with the list of folders, but they are definitely files.

Will make a PR to fix that, eventually together with adding the missing files and folder deletions from Beta 6 and 7.

brianteeman:
oops

richard67:
PR is #32582 . Will do other fixes (missing stuff) separately, to keep things reviewable and testable.

PR #32053, opened 2021-01-15
[4] Correct case of class name s/SimpleXmlElement/SimpleXMLElement

PhilETaylor:
Correct case of class name s/SimpleXmlElement/SimpleXMLElement


wilsonge:
Thanks!

PR #32054, opened 2021-01-15
[4] Remove unset call as $xml is not set when here

PhilETaylor:
Remove unset call as $xml is not set when here

Code review. 

`$xml` is only set in the nested foreach statement, and already unset before the return and continue within that.


HLeithner:
I'm wrong when I say that the $xml is unset automatically on return if not referenced to the object it self or used outside this function? Even the unset in the loop before continue look use less because the $xml will be reassigned in the next run.

HLeithner:
Thanks btw

PR #32060, opened 2021-01-16
[4.0] Form label [a11y]

brianteeman:
Each `<form>` element and form role that needs to be exposed as a landmark must be given an accessible name. This name will allow an assistive technology user to be able to quickly understand the purpose of the form landmark.

You can easily test this with the skipto plugin. Before this PR the item is just _Form_ After this PR the item is either _Form: New xxxx_ or _Form: Edit xxx_

- [x] com_banners
- [x] com_categories
- [x] com_contact
- [x] com_content
- [x] com_fields 
- [x] com_finder
- [x] com_installer
- [x] com_languages
- [x] com_mails
- [x] com_menus
- [x] com_messages
- [x] com_modules
- [x] com_newsfeeds
- [x] com_plugins
- [x] com_privacy 
- [x] com_redirect
- [x] com_tags
- [x] com_templates
- [x] com_users
- [x] com_workflow


administrator/components/com_menus/tmpl/menu/edit.php
```diff
@@ -20,7 +20,7 @@
 
 Text::script('ERROR');
 ?>
-<form action="<?php echo Route::_('index.php?option=com_menus&layout=edit&id=' . (int) $this->item->id); ?>" method="post" name="adminForm" id="item-form" class="form-validate">
+<form action="<?php echo Route::_('index.php?option=com_menus&layout=edit&id=' . (int) $this->item->id); ?>" method="post" name="adminForm" id="item-form" aria-label="<?php echo Text::_('COM_MENUS_MENU_FORM_' . ( (int) $this->item->id === 0 ? 'NEW' : 'EDIT'), true); ?>"class="form-validate">
```
Quy:
```suggestion
<form action="<?php echo Route::_('index.php?option=com_menus&layout=edit&id=' . (int) $this->item->id); ?>" method="post" name="adminForm" id="item-form" aria-label="<?php echo Text::_('COM_MENUS_MENU_FORM_' . ( (int) $this->item->id === 0 ? 'NEW' : 'EDIT'), true); ?>" class="form-validate">
```

Quy: There are more to fix.

carcam:
Weird enough the patch does not update the language file so the aria-label shows just the key string...

brianteeman:
Thats probably patchtester bug - i think it has issues when you have a lot of files.

carcam:
It looks like it. Changes in code reflect the language string so I will test it again and will take that into consideration.

wilsonge:
Merging on review - as if patchtester fails we're unlikely to get many tests and this is a pretty simple patch

brianteeman:
thanks

PR #32062, opened 2021-01-16
[4.0] Introduce a new render

dgrammatiko:
Pull Request for pending todos 

### Summary of Changes

- Create a new renderer bodybottom which renders just before the closing body tag
- Add a new private variable to the Document
- Add setter/getter for the previous private variable

- Fix the @todos for the toolbar modals

### Testing Instructions

Apply patch, and check that the modals in the Article Edit form (both the Versions and Preview) are still functional

@wilsonge although the problem was pretty obvious in the BS5 transition this solution and the shenanigans (the ability to enqueue html for the bottom of the page) should be pretty valid for J4 regardless of the BS5 decision

### DIFF
--- a/layouts/joomla/toolbar/modal.php
+++ b/layouts/joomla/toolbar/modal.php
@@ -33,7 +33,7 @@ $icon     = isset($displayData['icon']) ? $displayData['icon'] : 'icon-download'
 $text     = isset($displayData['text']) ? $displayData['text'] : '';
 
 // Enqueue the modal at the body bottom
-Factory::getDocument()->setBodyBottom(
+Factory::getDocument()->enqueueBodyEnd(
 	HTMLHelper::_('bootstrap.renderModal',
 		'modal_' . $selector,
 		[
--- a/layouts/joomla/toolbar/modal.php
+++ b/layouts/joomla/toolbar/modal.php
@@ -59,7 +59,7 @@ Factory::getDocument()->appendBodyEnd(
 		class="btn btn-primary"
 		type="button"
 		data-toggle="modal"
-		onclick="document.getElementById('modal_<?php echo $selector; ?>').open();">
+		data-target="#modal_<?php echo $selector; ?>">
 		<span class="<?php echo $icon; ?>" aria-hidden="true"></span>
 		<?php echo $text; ?>
 	</button>
--- a/layouts/joomla/toolbar/modal.php
+++ b/layouts/joomla/toolbar/modal.php
@@ -33,7 +33,7 @@ $icon     = isset($displayData['icon']) ? $displayData['icon'] : 'icon-download'
 $text     = isset($displayData['text']) ? $displayData['text'] : '';
 
 // Enqueue the modal at the body bottom
-Factory::getDocument()->enqueueBodyEnd(
+Factory::getDocument()->appendBodyEnd(
 	HTMLHelper::_('bootstrap.renderModal',
 		'modal_' . $selector,
 		[
--- a/layouts/joomla/toolbar/modal.php
+++ b/layouts/joomla/toolbar/modal.php
@@ -31,38 +31,36 @@ $id       = isset($displayData['id']) ? $displayData['id'] : '';
 $class    = isset($displayData['class']) ? $displayData['class'] : 'btn btn-primary';
 $icon     = isset($displayData['icon']) ? $displayData['icon'] : 'icon-download';
 $text     = isset($displayData['text']) ? $displayData['text'] : '';
-?>
 
-<!-- Render the button -->
+// Enqueue the modal at the body bottom
+Factory::getDocument()->setBodyBottom(
+	HTMLHelper::_('bootstrap.renderModal',
+		'modal_' . $selector,
+		[
+			'url'         => $displayData['doTask'],
+			'title'       => $text,
+			'height'      => '100%',
+			'width'       => '100%',
+			'modalWidth'  => 80,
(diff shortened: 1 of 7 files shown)


brianteeman:
If you can think of a better word than queue it will be better in the long term as its so easy to make a typo with it

dgrammatiko:
> If you can think of a better word than queue it will be better in the long term as its so easy to make a typo with it

I'm not good at naming things but I'm open to anything here

SharkyKZ:
This produces invalid HTML.

dgrammatiko:
> This produces invalid HTML.

~Interested to see how you came up with this conclusion~

Fixed

todordev:
It is a good idea but it will be better if the assets are managed by the assets manager.
We can provide an option **position**  with values **_bodystart_** and **_bodyend_**. We should rework the [script render](https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Document/Renderer/Html/ScriptsRenderer.php) to parser jdoc blocks depending on the attribute **position**.


```php
Factory::getDocument()->getWebAssetManager()
	->addInlineScript(
		HTMLHelper::_('bootstrap.renderModal',
			'modal_' . $selector,
			[
				'url'         => $displayData['doTask'],
				'title'       => $text,
				'height'      => '100%',
				'width'       => '100%',
				'modalWidth'  => 80,
				'bodyHeight'  => 60,
				'closeButton' => true,
				'footer'      => '<button class="btn btn-secondary" data-bs-dismiss="modal" type="button"'
				        . ' onclick="window.parent.Joomla.Modal.getCurrent().close();">'
				        . Text::_('COM_BANNERS_CANCEL') . '</button>'
				        . '<button class="btn btn-success" type="button"'
				        . ' onclick="Joomla.iframeButtonClick({iframeSelector: \'#modal_downloadModal\', buttonSelector: \'#exportBtn\'})">'
				        . Text::_('COM_BANNERS_TRACKS_EXPORT') . '</button>',
			]
		),
		['position' => 'bodyend']
	);
```

```html
<head>
     <jdoc:include type="scripts" />
</head>

<body>
     <jdoc:include type="scripts" position="bodystart" />
      .....
      ....
     <jdoc:include type="scripts" position="bodyend" />
</body>
```

dgrammatiko:
> It is a good idea but it will be better if the assets are managed by the assets manager.

This PR is not affecting the script position, it's a way to delegate a chunk of HTML to the end of the document.

Could the scripts refactored to take advantage of the new position? 

Definitely but that's for another PR. This one solves a problem that all the Toolbar buttons that trigger a Modal have issues with the z-index (conflicting dropdown and modal) so the rendered Modal HTML needs to be delegated to another place in the document (and since they are Modals could happily exists as the last elements in the page)

Hope this clears things up

PS the `HTMLHelper::_('bootstrap.renderModal', [])` returns an HTML string not a script 😉

ghost:
What still needs to be done to merge this PR?

dgrammatiko:
> What still needs to be done to merge this PR?

It needs 2 successful tests as every other Pull Request

ghost:
I see 3 successful checks for this PR, thats why I asking what still needs to be done to merge this PR?

dgrammatiko:
> I see 3 successful checks for this PR, thats why I asking what still needs to be done to merge this PR?

Thumbs up is not the way a test result is counted. You need to use https://issues.joomla.org/tracker/joomla-cms/32062 to mark your test as successful or failed

layouts/joomla/toolbar/modal.php
```diff
@@ -31,38 +31,36 @@
...
+			'width'       => '100%',
+			'modalWidth'  => 80,
+			'bodyHeight'  => 60,
+			'closeButton' => true,
+			'footer'      => '<button class="btn btn-secondary" data-bs-dismiss="modal" type="button"'
+							. ' onclick="window.parent.Joomla.Modal.getCurrent().close();">'
+							. Text::_('COM_BANNERS_CANCEL') . '</button>'
+							. '<button class="btn btn-success" type="button"'
+							. ' onclick="Joomla.iframeButtonClick({iframeSelector: \'#modal_downloadModal\', buttonSelector: \'#exportBtn\'})">'
+							. Text::_('COM_BANNERS_TRACKS_EXPORT') . '</button>',
```
HLeithner, line 52: Maybe a bit unreleated but some questions about this code.

* Don't we have an alternative way to add the onclick events?
* Why do we use here COM_BANNERS_CANCEL
* And way do we render an com_banner export button here?

At least the filename let me expect that this is a generic layout `layouts/joomla/toolbar/modal.php`

dgrammatiko: That file is already deleted in the main repo, thus the conflicts. About the totally legit question if we have a way to add onclick events: the answer is no but maybe we can beef up the html helper to do that (in combo with the initialization of the modal js part). A simple idea is to pass a json from the php with the css selector and the on-event name and the string to call (needs to be a function name or we getting xss problems). In sort doable but needs a bit abstraction work upfront

HLeithner: ah sorry didn't checked the conflicts first... thanks for the event handler explanation

dgrammatiko:
I guess this is not going anywhere

ghost:
Why you closed this pull?

dgrammatiko:
Two reasons:
- no tests after 2+ weeks
- this has an odd edge case

ghost:
for test you should leave open this

for edge case I don't know

PR #32067, opened 2021-01-17
[4.0] Fix edit account URI fragment

C-Lodder:
### Summary of Changes

Fix the URI fragment for the `Edit Account` so that it always opens the first tab

### Testing Instructions

1. Open the User Menu (top right)
2. Click `Edit Account`
3. Navigate to any tab...e.g `Joomla API Token`
4. Click `Close` in the toolbar
5. Re-do steps 1 and 2

The tab which was opened as last before the last closing is shown, e.g. Joomla API Token view

My Profile Details view is shown

### DIFF
--- a/administrator/modules/mod_user/tmpl/default.php
+++ b/administrator/modules/mod_user/tmpl/default.php
@@ -37,7 +37,7 @@ $hideLinks = $app->input->getBool('hidemainmenu');
 			<?php echo Text::sprintf('MOD_USER_TITLE', $user->name); ?>
 		</div>
 		<?php $uri   = Uri::getInstance(); ?>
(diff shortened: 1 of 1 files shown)


richard67:
I've tested with success, but the issue tracker doesn't show the PR yet, so I can't mark my test result.

There's a small mistake in the "Actual result BEFORE applying this Pull Request" section. It says "Joomla API Token view is shown", but it should be something like "The tab which was opened as last before the last closing is shown, e.g. Joomla API Token view.".

C-Lodder:
https://issues.joomla.org/tracker/joomla-cms/32067

richard67:
Yes, meanwhile it appears in the tracker, but it hasn't noticed yet that it's a PR and so there is no test button. Will wait a while.

richard67:
The tracker shows also the wrong build (staging instead of 4.0-dev).

HLeithner:
Thanks

PR #32072, opened 2021-01-18
[4.0] image xtd-editor icon

brianteeman:
Change the icon to use the same as the icon used in the media manager

### before
image

### after
image


wilsonge:
Thanks!

brianteeman:
thanks

PR #32120, opened 2021-01-23
[4.0] External link icon placement

brianteeman:
As identified by @infograf768 the placement of the icon to indicate an external link is incorrect in RTL.

Unlike his proposal which creates additional css just for the RTL to wap the class from ::before to ::after this PR does it the correct way.

The icon is still ::before the beginning of the text and now correctly identifies the direction.

No need for a separate class in the css at all. ;)

PR #32101 should be reverted and this used instead as its the correct semantic way

### DIFF
--- a/administrator/templates/atum/scss/blocks/_global.scss
+++ b/administrator/templates/atum/scss/blocks/_global.scss
@@ -166,8 +166,9 @@ body .container-main {
   text-align: end !important;
 }
 
-// extern links with icons
+// external links with icons
 a[target="_blank"]::before {
+  display: inline-block;
   padding-right: 3px;
   font-family: "Font Awesome 5 Free";
   font-size: 14px;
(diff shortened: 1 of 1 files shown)


infograf768:
testing

infograf768:
we still have a padding problem as it still uses padding-right instead of left

infograf768:
Screen Shot 2021-01-23 at 11 55 43

brianteeman:
That would be resolved when the build tools are updated to use rtlcss

infograf768:
> That would be resolved when the build tools are updated to use rtlcss

Then let's wait to revert https://github.com/joomla/joomla-cms/pull/32101 and merge this.

brianteeman:
Let's not as your pr is wrong and merging the two would end up with two icons. The use of inline-block is correct. That's why the menu icons work without an rtl override

brianteeman:
Padding resolved by using padding-inline-end which is therefore the same for LTR and RTL

wilsonge:
Thanks!

PR #28589, opened 2020-04-06
[4.0][RTL] CSS logical properties initial

ciar4n:
Pull Request for Issue #28546 .

Also removed some redundant rules ( #28586 & #28585).

### Summary of Changes
Implements some logical properties to reduce on RTL CSS.

### Testing Instructions
Apply this patch and run `node build.js --compile-css` to update SCSS. Check RTL styling.

PR #32127, opened 2021-01-23
[4.0] Remove header page title css

brianteeman:
This css should have been removed a long time ago when @ciar4n changed the _header.scss in  #28589


wilsonge:
thanks!

Issue #32116, opened 2021-01-23
[4.0][BS5]Bootstrap tab is broken


joomdonation:
### Steps to reproduce the issue
Use the below code to create tab
```php
echo JHtml::_('bootstrap.startTabSet', 'tab', ['active' => 'tab1']);
echo JHtml::_('bootstrap.addTab', 'tab', 'tab1', 'Tab 1 Title');
echo 'Tab 1 Content';
echo JHtml::_('bootstrap.endTab');
echo JHtml::_('bootstrap.addTab', 'tab', 'tab2', 'Tab 2 Title');
echo 'Tab 2 Content';
echo JHtml::_('bootstrap.endTab');
echo JHtml::_('bootstrap.endTabSet');
```

### Expected result
Tabs displayed

### Actual result
Tabs are not being displayed

### System information (as much as possible)
Joomla 4 nightly build

### Additional comments
Maybe we need to wait for PR https://github.com/joomla/joomla-cms/pull/31990/ merged to have it works ?

joomdonation:
@wilsonge I tried again with nightly build today and the issue is still not fixed yet.

dgrammatiko:
@joomdonation I think the selector needs to be a CSS selector and you're passing a string (which will result to an element tag `tab` which obviously doesn't exist)

The example code should make it clear what's broken here: https://github.com/dgrammatiko/bs5/blob/cbd103ddb7e28632d563bdec0a3af6fee4a813f9/test_component/site/tmpl/tab/default.php#L19-L33

joomdonation:
@dgrammatiko The selector actually becomes part of id of the Tabset. We have been using it for years, see the code from staging https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_content/views/article/tmpl/edit.php#L65

(and just for information, that code has been working at least until recently)

dgrammatiko:
@joomdonation but does it work if you pass it as a class? The B/C break is something new after the Bootstrap V5 and it might be reverted or patched in a way that will be B/C (undecided at this point).

joomdonation:
@dgrammatiko Yes. If I pass a class selector, then it works

```php
echo JHtml::_('bootstrap.startTabSet', '.tab', ['active' => 'tab1']);
echo JHtml::_('bootstrap.addTab', '.tab', 'tab1', 'Tab 1 Title');
echo 'Tab 1 Content';
echo JHtml::_('bootstrap.endTab');
[...]

PR #32146, opened 2021-01-24
[4.0] Mitigate some B/C breaks

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/32116

### Summary of Changes
With the introduction of Bootstrap v5, the selector argument in all the methods has changed so it can accept any valid CSS selector. The three methods that create the HTML markup for Tabs, Accordions and Modals are Joomla specific and can continue to function as before, eg: the selector is a word and the method treats it as a class. This PR applies a simple check that if the selector is  **not** a class name will modify it to become one (the old code was doing something similar already). 

### Testing Instructions
Grab the testing component from https://github.com/dgrammatiko/bs5/blob/main/BS5Tests.zip
Modify the layouts for tab, accordion and modal by removing the `.` from the second argument of the method. 

B/C broken

Interactive components work fine

@wilsonge

### DIFF
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -570,11 +570,11 @@ abstract class Bootstrap
 	 *
 	 * @since   3.0
 	 *
-	 *                             Options for the tooltip can be:
-	 *                             - parent  selector  If selector then all collapsible elements under the specified parent will be closed when this
-	 *                                                 collapsible item is shown. (similar to traditional accordion behavior)
-	 *                             - toggle  boolean   Toggles the collapsible element on invocation
-	 *                             - active  string    Sets the active slide during load
+	 * Options for the tooltip can be:
+	 * - parent  selector  If selector then all collapsible elements under the specified parent will be closed when this
+	 *                     collapsible item is shown. (similar to traditional accordion behavior)
+	 * - toggle  boolean   Toggles the collapsible element on invocation
+	 * - active  string    Sets the active slide during load
 	 */
 	public static function startAccordion($selector = '.myAccordian', $options = []) :string
 	{
@@ -582,6 +582,12 @@ abstract class Bootstrap
 		if (isset(static::$loaded[__METHOD__][$selector]))
 		{
 			return '';
+		}
+
+		// For B/C
+		if (substr($selector, 0, 1) !== '.')
+		{
+			$selector = '.' . $selector;
 		}
 
 		// Include Bootstrap component
@@ -632,6 +638,12 @@ abstract class Bootstrap
 	 */
 	public static function addSlide($selector, $text, $id, $class = '') :string
 	{
+		// For B/C
+		if (substr($selector, 0, 1) !== '.')
+		{
+			$selector = '.' . $selector;
+		}
+
 		$in        = static::$loaded[__CLASS__ . '::startAccordion'][$selector]['active'] === $id ? ' show' : '';
 		$collapsed = static::$loaded[__CLASS__ . '::startAccordion'][$selector]['active'] === $id ? '' : ' collapsed';
 		$parent    = static::$loaded[__CLASS__ . '::startAccordion'][$selector]['parent'] ?
@@ -699,6 +711,12 @@ HTMLSTR;
 			return '';
 		}
 
+		// For B/C
+		if (substr($selector, 0, 1) !== '.')
+		{
+			$selector = '.' . $selector;
+		}
+
 		// Initialise with the Joomla specifics
 		$options['isJoomla'] = true;
 
@@ -732,6 +750,12 @@ HTMLSTR;
 
 		if (!isset(static::$loaded[__METHOD__][$sig]))
 		{
+			// For B/C
+			if (substr($selector, 0, 1) !== '.')
+			{
+				$selector = '.' . $selector;
+			}
+
 			// Setup options object
 			$opt['active'] = (isset($params['active']) && ($params['active'])) ? (string) $params['active'] : '';
 
@@ -775,6 +799,12 @@ HTMLSTR;
 	public static function addTab($selector, $id, $title) :string
 	{
 		static $tabLayout = null;
+
+		// For B/C
+		if (substr($selector, 0, 1) !== '.')
+		{
+			$selector = '.' . $selector;
+		}
 
 		$tabLayout = $tabLayout === null ? new FileLayout('libraries.html.bootstrap.tab.addtab') : $tabLayout;
 		$active = (static::$loaded[__CLASS__ . '::startTabSet'][$selector]['active'] == $id) ? ' active' : '';


laoneo:
I have tested this item :white_check_mark: successfully on a0ed44c94db1ed93fb9c71744ffc2e065a19b254

Tested it with the free version of DPCalendar 8.0.3.

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32146">issues.joomla.org/tracker/joomla-cms/32146</a>.</sub>

wilsonge:
Can we be a bit smarter with this? At least check for an id starting with # maybe even not prepend if it starts with data (and assume it’s a data attribute)

ghost:
I have tested this item :white_check_mark: successfully on a0ed44c94db1ed93fb9c71744ffc2e065a19b254<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32146">issues.joomla.org/tracker/joomla-cms/32146</a>.</sub>

laoneo:
In J3 we do not have a starting #. See https://github.com/Digital-Peak/DPCalendar-Free/blob/main/com_dpcalendar/admin/views/event/tmpl/default.php#L46

rdeutz:
@richardfrederik67 Please stop creating fake accounts, say you have tested something you haven't tested. You are wasting our time we can use to bring this project forward. You might guessed it, your account is blocked.

dgrammatiko:
> Can we be a bit smarter with this? At least check for an id starting with # maybe even not prepend if it starts with data (and assume it’s a data attribute)

@wilsonge I'm not fixing all the selectors for all the methods. This PR affected only the 3 methods that can be called to produce some HTML wrapping text around some existing markup. It might be a good idea to revert to the J3 version of the selectors for these methods. BTW the rest of the functions still can handle any valid CSS selector (which effectivelly is a B/C break)

laoneo:
Revert to the J3 version would be nice.

dgrammatiko:
> Revert to the J3 version would be nice.

I think for the 3 methods (accordion, tab, modal) this is ok, the reasons I didn't did it in this PR were:
- reverting needed a bit more files touched
- and I didn't know if this was valid so didn't want to waste my time

But I'm fine doing that if there's an agreement on how to handle (I guess at this point we agree that this is an unneeded break)

wilsonge:
> In J3 we do not have a starting #. See https://github.com/Digital-Peak/DPCalendar-Free/blob/main/com_dpcalendar/admin/views/event/tmpl/default.php#L46

I know but there we only accept classes which isn’t ideal either. Ideally we want to transition to allow any Dom selector. So the question is is how to we get there with the best b/c

Fedik:
This will not fly.

`div.blabla`
`#foo-bar`
`custom-element`
etc.

Is valid selectors that will be broken

Fedik:
I think should be this:
```php
if (strpos($selector, '.') === false && strpos($selector, '#') === false  && strpos($selector, ' ') === false)
{
	$selector = '.' . $selector;
}
```

Also somewhere need to be noted that "element selector" will not work. 
And of course need a "fixed time", when this hack will be removed.

laoneo:
At the end if the day we need a solution where we can use the same call in J3 and J4.

HLeithner:
> At the end if the day we need a solution where we can use the same call in J3 and J4.

Does it need to be the same call, I mean yes the same call but maybe adding an option for a full css selector in j4, in some of the functions we already have an $options parameter that could be used for it.

Alternative is to use new function names for j4 and write migration functions for j3 which simple always add the "."

dgrammatiko:
> At the end of the day, we need a solution where we can use the same call in J3 and J4.

The method is the same, the only difference is the second argument is expected to be a real CSS selector in j4 where is J3 was a word interpreted either as class name or id.

> Alternative is to use new function names for j4 and write migration functions for j3 which simple always add the "."

something similar is what I had proposed to George when we were discussing the JS PR (eg handle the break in the J3 with a deprecation notice and an additional message in the log). I understand that this opens a can of worms but the current (J3) API will not be sufficient for J4. In J3 you could have called `bootstrap.framework` to load the JS and then add inline js to initilise any component. With J4 due to all the other changes this isn't needed anymore as long as you call the method for the component (eg `HTMLHelper::_('bootstrap.button', 'button[data-something-valid]')` will do that without any js from the DEV)

laoneo:
> Does it need to be the same call, I mean yes the same call but maybe adding an option for a full css selector in j4, in some of the functions we already have an $options parameter that could be used for it.
> 
> Alternative is to use new function names for j4 and write migration functions for j3 which simple always add the "."

Yes we need, this is called backwards compatibility. Adding an option for this is ridiculous. @dgrammatiko already made it BC, so we either use it as it is or do look for improvements.

HLeithner:
> Yes we need, this is called backwards compatibility. Adding an option for this is ridiculous. @dgrammatiko already made it BC, so we either use it as it is or do look for improvements.

The Option I thought was in j4 only to activate full CSS selector (or the other way around to disable b/c mode)

But I think function with a clean structure and a b/c wrapper would be better

dgrammatiko:
I'm not really sure if we're all on the same page here, let me recap:
- The functions `endTab, addTab, endTabSet, startTabSet, renderModal, endSlide, addSlide, endAccordion, startAccordion` are Joomla specific, thus we can use the limited way (accepts a word and interprets it as classname or ID). Not great but we're not breaking B/C

- BUT `toast, tooltip, tab, scrollspy, popover, modal, dropdown, collapse, carousel, button, alert` are Bootstrap coupled and should be able to accept any valid CSS selector. Also, the implementation rn is not correct, the $selector should default to null and in this case, only the js component is loaded (nothing initialised) and if a $selector is set then also the component will initialise that automatically (no extra js required from the devs in this case). Here the B/C is expected (also kinda impossible to patch it in a way to cover all CSS selectors without writing some spec-compliant CSS selector validator)

Fedik:
Selectors is selectors, no mater where they used. Joomla 1.5 used only ID everywhere. Current issue is kind of very old legacy.
I would not do restriction for Tabs/Accordion, but if it will help, well then :)

image

laoneo:
So guys, can we not just merge this one and open an issue to find the gold solution. ATM tons of extensions are broken on J4.

wilsonge:
@dgrammatiko can you add the proposed code by Fedik into these accordion/tabs elements so we effectively will accept class/id’s with prefix’s

Fedik:
I just noticed that I have missed "attribute selector", here is update:

```php
if (strpos($selector, '.') === false && strpos($selector, '#') === false  
  && strpos($selector, ' ') === false &&  strpos($selector, '[') === false)
{
	$selector = '.' . $selector;
}
```

dgrammatiko:
@wilsonge @laoneo @Fedik I tried another approach: https://github.com/joomla/joomla-cms/pull/32357

richard67:
@dgrammatiko Does that other PR #32357 replace this one here? Or is it an addition to this here?

dgrammatiko:
@richard67 keep both for the time. I have no clue how we want to move forward here

richard67:
@dgrammatiko But at the end we chose only 1 of them, right? Just asking to be sure, no problem with it.

dgrammatiko:
@richard67 yup at the end there will be one only solution, but it might not be any of these PRs 🤣

laoneo:
Can you rebase this branch with 4.0-dev please.

dgrammatiko:
This is not quite right according to https://github.com/joomla/joomla-cms/issues/32368#issuecomment-776612255 So I'm closing please test https://github.com/joomla/joomla-cms/pull/32357

PR #32159, opened 2021-01-25
[4.0] Use a bootstrap global object

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
Use a global object named `bootstrap` instead of the `Joomla.Bootstrap`

## WHY?

This is my fault, let me explain: Joomla has 2 instances (tabs and modals) where it kinda uses it's own constructor (eg some specific tasks executed before initialising the interactive elements). When I was implementing this JS part of BS5 I thought a `Joomla` registry was a good idea. **Never ever ship code that was a nice idea**. The problem is that Bootstrap documentation in such case is referring to objects that don't actually exist in the document. **I'm sorry, this was a terrible idea**

### Testing Instructions
Check the dropdowns in the dashdoard
Check the modals in the Article Edit
Check the Accordion in the Menu edit (select new component)
Check the tabs in the tinyMCE plugin (set0... )

Bootstrap docs don't actually reflect the methods of the interactive components

Code is the same but the Bootstrap docs reflect exactly what exists in our codebase

### Example of the code that reflects BS docs
Screenshot 2021-01-25 at 18 31 58
Screenshot 2021-01-25 at 18 36 40
Screenshot 2021-01-25 at 18 39 41

The Bootstrap Docs **is** our documentation but we still need to document the Bootstrap.php methods 

@wilsonge sorry mate for the mess, this needs to be merged ASAP

### DIFF
--- a/build/.eslintrc
+++ b/build/.eslintrc
@@ -21,7 +21,8 @@
   // Additional global variables your script accesses during execution
   "globals": {
     "Joomla": true,
-	"MediaManager" : true
+	"MediaManager" : true,
+	"bootstrap": true,
   },
   // Rule overrides
   "rules": {
--- a/build/media_source/mod_multilangstatus/js/admin-multilangstatus.es6.js
+++ b/build/media_source/mod_multilangstatus/js/admin-multilangstatus.es6.js
@@ -21,7 +21,7 @@
       document.body.appendChild(clone);
 
       // Modal was moved so it needs to be re initialised
-      Joomla.Bootstrap.Initialise.Modal(clone);
+      Joomla.initialiseModal(clone);
     }
   });
 })();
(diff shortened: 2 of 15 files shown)


ceford:
I applied the patch, ran npm ci and checked that the various things mentioned still worked. They did. But I don't really understand what I am looking for. Should I expect to see visible changes in the rendered html? I think this is a pass but I am not sure!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32159">issues.joomla.org/tracker/joomla-cms/32159</a>.</sub>

dgrammatiko:
> But I don't really understand what I am looking for

The changes affect only the exposed global object so it's only for very advanced developers. Thanks for testing

wilsonge:
LGTM and tested myself locally too. Thanks!

dgrammatiko:
Thanks and sorry for the mess

wilsonge:
Don't worry :)

Issue #32113, opened 2021-01-22
[4.0][BS5] TODO RTL css


brianteeman:
all rtl needs to be reviewed etc

brianteeman:
ALL the rtl scss that we have needs to be reviewed to see if it is even needed now.

As well as the start/end stuff there is also magic taking process when the compilation takes place. Bootstrap 5 uses rtlcss methodology which provides magic transformations and the use of rtl variables if necessary. 

We should end up with nothing in the template-rtl.scss file

### magic
This css 

```
.list-group .published {
    padding-left: .9rem;
    border-left: 5px solid var(--success);
}

```
will magically be written in the template-rtl.css file as
```

.list-group .published {
    padding-right: .9rem;
    border-right: 5px solid var(--success);
}
```

### Variables
```
h1 {
  font-weight: 700 /* rtl:600 */;
}
```
will be written in the template-rtl.css file as
```
h1 {
  font-weight: 600;
}
```

I suspect that we will be able to completely remove the template-rtl.scss file eventually.

### More information :
https://getbootstrap.com/docs/5.0/getting-started/rtl/

https://rtlcss.com/learn/

https://rtlcss.com/playground/

brianteeman:
Note that for this all to work we will need to change our build scripts

wilsonge:
Yeah this is a build script change and a lot of SCSS changes (to add all the relevant comments). I did actually do basic testing of the RTL views - and I think most the stuff continues to work as before for now. There's one bug JM raised with the menu now staying sticky on scroll. But I think by and large RTL continues to work with the hacky overrides. I *think* this may be something we don't have to prioritise for 4.0 stable but more like something we target for 4.1.

infograf768:
RTL admin menu is working fine now with Dimitris patch

infograf768:
`float-start` is really broken in RTL: it always gives a float: left

wilsonge:
Float-start previously was float-left which also always did a float left. This won’t have changed

brianteeman:
Please give an example

ciar4n:
I think its rename to `float-start` in BS5 is a bit deceiving. It gives the impression that it is a logical property and therefore switches direction in RTL which it does not. Float as a logical property is only supported in Firefox (`float: inline-start;`).

Ideally, floating elements should be avoided. Its only real use these days is where it was originally designed for and that is within an article where you want text to wrap around an element (eg. an image). 

[...]

PR #32177, opened 2021-01-27
[4.0] RTL input groups rounded corners

brianteeman:
Since the BS5 upgrade some of the rounded corners are incorrect. I have put the corrections in the vendor/bootstrap/_bootstrap-rtl.scss that @infograf768 created.

The existing rounded corners code in the template-rtl.scss file may or may not be needed I didnt check.

scss really isnt my thing

## Before
chrome_2021-01-27_22-22-14

chrome_2021-01-27_22-22-54

chrome_2021-01-27_22-23-36

## After
chrome_2021-01-27_22-26-02

chrome_2021-01-27_22-25-34

chrome_2021-01-27_22-25-03

### DIFF
--- a/administrator/templates/atum/scss/vendor/bootstrap/_bootstrap-rtl.scss
+++ b/administrator/templates/atum/scss/vendor/bootstrap/_bootstrap-rtl.scss
@@ -142,3 +142,35 @@
   padding-left: 0 !important;
 }
 
+
+.input-group {
+  &:not(.has-validation) {
+    > :not(:last-child):not(.dropdown-toggle):not(.dropdown-menu),
+    > .dropdown-toggle:nth-last-child(n + 3) {
+      @include border-end-radius($border-radius);
+    }
+  }
+
+  &.has-validation {
+    > :nth-last-child(n + 3):not(.dropdown-toggle):not(.dropdown-menu),
+    > .dropdown-toggle:nth-last-child(n + 4) {
+      @include border-end-radius(0);
+    }
+  }
+  $validation-messages: "";
+  @each $state in map-keys($form-validation-states) {
+    $validation-messages: $validation-messages + ":not(." + unquote($state) + "-tooltip)" + ":not(." + unquote($state) + "-feedback)";
+  }
+
+  > :not(:first-child):not(.dropdown-menu)#{$validation-messages} {
+    margin-left: -$input-border-width;
+    @include border-start-radius(0);
+    @include border-end-radius(0);
+  }
+
+  > :last-child:not(.dropdown-menu)#{$validation-messages} {
+    margin-left: -$input-border-width;
+    @include border-start-radius($border-radius);
+    @include border-end-radius(0);
+  }
+}


bembelimen:
Isn't this more a BS5 RTL issue and should be reported there?

infograf768:
> Isn't this more a BS5 RTL issue and should be reported there?

We are not using the bootstrap5 rtl possibilities right now. Thus why we had to create overrides.

ceford:
Is there an easy way to invoke rtl behaviour? I tried fiddling with direction in the Atum index.php but that did not do anything.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32177">issues.joomla.org/tracker/joomla-cms/32177</a>.</sub>

HLeithner:
I thought we will use bs5 right to left support, shouldn't we work on that instead of do patchwork? I mean it's a bit waste of time or? @dgrammatiko @wilsonge what's missing to use the native RTL support?

wilsonge:
You run this https://rtlcss.com/ as postcss against the generated css file from SCSS (full info https://getbootstrap.com/docs/5.0/getting-started/rtl/) . See https://github.com/joomla/joomla-cms/issues/32113#issuecomment-765903379 for some more info

brianteeman:
@ceford you change the value of <rtl> in the langmetadata.xml file in the admin languages folder

@HLeithner until the build scripts are updated then there is no choice. And @wilsonge has said that they wont be done yet so this just moves us forward.

HLeithner:
Ok just read the comment thx

wilsonge:
> And @wilsonge has said that they wont be done yet so this just moves us forward.

Just to be clear if @dgrammatiko or someone contributes it then I'll definitely consider it. I just don't personally have the capacity because I wanna focus on the release blockers and there's a lot more people who can work on scss

dgrammatiko:
@wilsonge will do after the joomla update pr

brianteeman:
LEt me know if you need a hand on the update

ceford:
Fixed as illustrated if Debug System is Yes but not if Debug System is No. I guess that must be in the compilation of the min file. So is this a pass? Really weird to see English ltr layout! Thanks for the tip on how to set this up.

Another odd-ball: In ltr mode in Firefox and Chrome on Mac there is an enormous empty area to the left of the visible page. I can scroll left, and scroll and scroll and scroll but nothing comes into view.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32177">issues.joomla.org/tracker/joomla-cms/32177</a>.</sub>

brianteeman:
Anything holding back this PR from being merged and in the beta 7?

wilsonge:
No - good enough for a first pass

brianteeman:
Thanks

PR #32178, opened 2021-01-27
[4.0] RTL help and options button

brianteeman:
Looks like the values for the rtl settings were the wrong way around

## Before
image

## After
image


infograf768:
Please wait on this one as it could mean there was an error for all ms- class overrides.
Can’t look at it until end of week.

brianteeman:
If there is an error it doesnt have any impact on this PR and can be addressed in its own pr

wilsonge:
This is now the opposite of `me-auto` which seems logical. I think this is good to merge.

brianteeman:
Thanks

infograf768:
I checked. That one was the only error in the `ms-` classes. oof

Issue #32185, opened 2021-01-29
[4.0] Uncaught Reference Error - boostrap not definded


chmst:
### Steps to reproduce the issue
Fresh installation 4.0 dev, after npm ci. Open Global Configuration, watch Console

<img width="852" alt="bootstrap-not-defined" src="https://user-images.githubusercontent.com/1035262/106259117-145bc980-621f-11eb-9873-e77298be67de.PNG">

On Xampp, win10.

dgrammatiko:
That's probably because of a wrong order of the scripts. Try moving https://github.com/joomla/joomla-cms/blob/83d1e49f45c5d93688a7024b0dbd2795cfe9380e/layouts/joomla/form/field/media.php#L119-L120

before https://github.com/joomla/joomla-cms/blob/83d1e49f45c5d93688a7024b0dbd2795cfe9380e/layouts/joomla/form/field/media.php#L158

PS. Nope, that's not the problem. Change https://github.com/joomla/joomla-cms/blob/83d1e49f45c5d93688a7024b0dbd2795cfe9380e/build/media_source/system/js/fields/joomla-field-media.w-c.es6.js#L192
to 
```js
})(customElements, Joomla, bootstrap);
```

and https://github.com/joomla/joomla-cms/blob/83d1e49f45c5d93688a7024b0dbd2795cfe9380e/build/media_source/system/js/fields/joomla-field-media.w-c.es6.js#L1 to 

```js
((customElements, Joomla, bootstrap) => {
```

chmst:
With these changes 

((customElements, Joomla, bootstrap) => {
...
})(customElements, Joomla, bootstrap);

the first two errors have gone, it remains the third one at line 247.






<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32185">issues.joomla.org/tracker/joomla-cms/32185</a>.</sub>

dgrammatiko:
Try changing
```js
&& bootstrap.Modal.getInstance(this.modalElement) === undefined)
```

to

```js
&& bootstrap.Modal.getInstance
&& bootstrap.Modal.getInstance(this.modalElement) === undefined)
```

chmst:
[...]

PR #32193, opened 2021-01-30
[4.0] Wrong order of scripts loading

dgrammatiko:
Pull Request for Issue #32185 .

### Summary of Changes

The fields Media and User have a dependency on the Bootstrap Modal but the order of the execution is wrong. The modal helper function needs to be called before including the field's own script

### Testing Instructions
- Login to the administrator area
- Go to Global Configuration and confirm that there are no console errors
- Edit an article and confirm no console errors and the `Created By`, `Intro Image` and `Full Article Image` fields still working

No


ceford:
I did not understand the testing instructions concerning User field in Article edit. However, in Users / Add User Note I get **Uncaught ReferenceError: bootstrap is not defined** which goes away after applying the patch. But I get this warning in the form: Notice: Undefined variable: uri in /Users/ceford/Sites/joomla-cms-4/layouts/joomla/form/field/user.php on line 61. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32193">issues.joomla.org/tracker/joomla-cms/32193</a>.</sub>

dgrammatiko:
> But I get this warning in the form: Notice: Undefined variable: uri in /Users/ceford/Sites/joomla-cms-4/layouts/joomla/form/field/user.php on line 61.

This should be fixed now with https://github.com/joomla/joomla-cms/pull/32193/commits/bf8f7ecdae1d02ca9cc9f1d22af30520bfd003f8

> I did not understand the testing instructions concerning User field in Article edit

Got to the Publishing tab and check the field `Created By`

chmst:
The notice in user field is ok now, but I stll have the error "uncaught.." in media field.

dgrammatiko:
> The notice in user field is ok now, but I stll have the error "uncaught.." in media field.

What URL?

chmst:
New article

dgrammatiko:
@chmst should be ok after https://github.com/joomla/joomla-cms/pull/32193/commits/4355940a6a00c914ef9536fc849cb1f91ede21aa

Fedik:
@dgrammatiko a better fix is to to call `Joomla.initialiseModal(this.modalElement, {isJoomla: true });`
not in `connectedCallback()`, but on first interaction, lazylly, and have an internal flag to mark that `Joomla.initialiseModal` was called.

Fedik:
Additionally, instead of callback I would use event:
```
Joomla.Event.dispatch(this.modalElement, 'joomla:initModal', {isJoomla: true });
```
So it does not strict to have `Joomla.initialiseModal`, and anyone can listen to this event.

You see, it useful :wink:

dgrammatiko:
> but on first interaction, lazylly, and have an internal flag to mark that Joomla.initialiseModal was called.

True, that a better implementation (btw this PR was done on Github UI, so...)

> Additionally, instead of callback I would use event:

In general, I will agree with you here that events are a better way. There are couple caveats here:
- Bootstrap.Modal itself is a constructor and all the code `Joomla.initialiseModal` is kinda a wrapper around it
- The callback is closer to the `bootstrap.Modal(element)`, so I guess easier for Joomla devs (might be wrong here)
- A way better model is Custom Elements, the constructor is on the actual element so no event no callbacks. The problem is that instead of `<bootstrap-modal>` we try to apply all the interactivity on a `<div>`...

Fedik:
> A way better model is Custom Elements,

Well, it already Custom Element :smile: 

I understood what you mean, however it not really a fix, example when on frondend I want to use FooBarModal.js instead of BotCrap.js.
Event or Callback still better in this case, because I can disable botcrap.js, and hook up my script.

> btw this PR was done on Github UI

Not bad as for Github UI :wink:

chmst:
I have tested this item :white_check_mark: successfully on 4355940a6a00c914ef9536fc849cb1f91ede21aa<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32193">issues.joomla.org/tracker/joomla-cms/32193</a>.</sub>

dgrammatiko:
> I understood what you mean, however it not really a fix, example when on frondend I want to use FooBarModal.js

That's the big problem of `HTMLHelper` and `FormField` they need to become client-specific. Actually this is one of the major problems of Joomla and somehow it needs a realistic solution so finally the backend could be decoupled from the frontend.

Fedik:
> That's the big problem of HTMLHelper

I tried to reduce use of `HTMLHelper::_('bootstrap.foobar')` (and such) calls to  `$wa->useScript('bootstrap.foobar')`, but I see someone added much more fancy code there :smile: 

But not much important.

dgrammatiko:
> I see someone added much more fancy code there

If you have a better way then go ahead and change the code there. All I wanted to do there was to modularise the JS, eg include only the JS you need per route (instead of bringing both jQuery and the whole Bootstrap library even if all you wanted was a collapse component).

Fedik:
> If you have a better way then go ahead and change the code there. All I wanted to do there was to modularise the JS

As I said it is not much important for now. 
For now I made this #32204, so later we can improve it, when will have some more time.

dgrammatiko:
> when will have some more time.

Yeah let's patch only the really breaking or bad things for now and hopefully, we can improve things later on

Fedik:
I have tested this item :white_check_mark: successfully on 4355940a6a00c914ef9536fc849cb1f91ede21aa<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32193">issues.joomla.org/tracker/joomla-cms/32193</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32193">issues.joomla.org/tracker/joomla-cms/32193</a>.</sub>

wilsonge:
Thanks!

Issue #32182, opened 2021-01-29
[4] Login: WebAuthn-Img shows alt


ghost:
### Steps to reproduce the issue

![Untitled](https://user-images.githubusercontent.com/71960554/106245322-9098e180-620c-11eb-924d-31516f56111b.png)

### System information (as much as possible)

Test on launch.joomla.org, installation of today.

![image](https://user-images.githubusercontent.com/71960554/106246944-2df51500-620f-11eb-904f-d7a1fa712d28.png)

chmst:
I cannot replicate this issue.

ghost:
Safari:
![Screen Shot 2021-01-29 at 10 55 48](https://user-images.githubusercontent.com/71960554/106260360-a0babc00-6220-11eb-8aa6-fa9e9752bdce.png)

infograf768:
Can't reproduce
<img width="303" alt="Screen Shot 2021-01-29 at 17 21 22" src="https://user-images.githubusercontent.com/869724/106300230-7edc2c00-6256-11eb-830e-d7e00bd979f5.png">

ghost:
menu, not form as shown in screenshots on firefox and safari.

![106260360-a0babc00-6220-11eb-8aa6-fa9e9752bdce](https://user-images.githubusercontent.com/71960554/106301713-52c1aa80-6258-11eb-9c6b-79749ca91b26.png)

infograf768:
That was not clear in the instructions.

Issue confirmed.

ghost:
> That was not clear in the instructions.

The firefox-screenshot show clear the layout of menu, the safari-screenshot show clear the breadcrumbs.

infograf768:
Please do not argue:
the breadcrumbs does not mean anythying.
also the module could have been displayed elsewhere...

Firefox:
<img width="1256" alt="Screen Shot 2021-01-29 at 18 03 20" src="https://user-images.githubusercontent.com/869724/106304982-5e16d500-625c-11eb-8986-441f7887e7c4.png">

infograf768:
Safari

<img width="1234" alt="Screen Shot 2021-01-29 at 18 17 48" src="https://user-images.githubusercontent.com/869724/106306434-59532080-625e-11eb-8efa-d48e86bc9ffb.png">

infograf768:
@N6REJ 

Looks like we have an issue here. Can you look at it?

ghost:
@infograf768 Please do not argue.

[...]

PR #31545, opened 2020-12-01
[4.0] modify #31190

N6REJ:
Pull Request for Issue # .

### Summary of Changes
addresses https://github.com/joomla/joomla-cms/pull/31516
removes api and replaces it with overrideable call

### Testing Instructions
apply pr
run npm ci
check logins via https and verify logo shows.

### Expected result AFTER applying this Pull Request
image

image

### Documentation Changes Required
remove documentation documenting htmlhelper::icon.svg

PR #32196, opened 2021-01-30
[4.0] Correcting WebAuthn display in login menu item

infograf768:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/32182

### Summary of Changes
https://github.com/joomla/joomla-cms/pull/31545 forgot to modify com_users login default_login tmpl and login modules

### Testing Instructions
Set webautn to display when login.
Create a login menu item.

Screen Shot 2021-01-29 at 18 03 20

Screen Shot 2021-01-30 at 10 08 50

### Note

This PR does not solve the tooltip display in the menu as well as in the login module
`PLG_SYSTEM_WEBAUTHN_LOGIN_DESC="Login without a password using the W3C Web Authentication (WebAuthn) standard in compatible browsers. You need to have already set up WebAuthn authentication in your user profile."`

### DIFF
--- a/components/com_users/tmpl/login/default_login.php
+++ b/components/com_users/tmpl/login/default_login.php
@@ -79,7 +79,7 @@ $usersConfig = ComponentHelper::getParams('com_users');
 				<div class="com-users-login__submit control-group">
 					<div class="controls">
 						<button type="button"
-								class="btn btn-secondary <?php echo $button['class'] ?? '' ?>"
+								class="btn btn-secondary w-100 <?php echo $button['class'] ?? '' ?>"
 								<?php foreach ($dataAttributeKeys as $key): ?>
 								<?php echo $key ?>="<?php echo $button[$key] ?>"
 								<?php endforeach; ?>
@@ -92,11 +92,7 @@ $usersConfig = ComponentHelper::getParams('com_users');
 							<?php if (!empty($button['icon'])): ?>
 								<span class="<?php echo $button['icon'] ?>"></span>
 							<?php elseif (!empty($button['image'])): ?>
-								<?php echo HTMLHelper::_('image', $button['image'], Text::_($button['tooltip'] ?? ''), [
-									'class' => 'icon',
-								], true) ?>
-							<?php elseif (!empty($button['svg'])): ?>
-								<?php echo $button['svg']; ?>
+								<?php echo $button['image']; ?>
 							<?php endif; ?>
 							<?php echo Text::_($button['label']) ?>
 						</button>
--- a/components/com_users/tmpl/login/default_login.php
+++ b/components/com_users/tmpl/login/default_login.php
@@ -93,6 +93,8 @@ $usersConfig = ComponentHelper::getParams('com_users');
 								<span class="<?php echo $button['icon'] ?>"></span>
 							<?php elseif (!empty($button['image'])): ?>
 								<?php echo $button['image']; ?>
+							<?php elseif (!empty($button['svg'])): ?>
+								<?php echo $button['svg']; ?>
 							<?php endif; ?>
 							<?php echo Text::_($button['label']) ?>
 						</button>
--- a/components/com_users/tmpl/login/default_login.php
+++ b/components/com_users/tmpl/login/default_login.php
@@ -92,7 +92,9 @@ $usersConfig = ComponentHelper::getParams('com_users');
 							<?php if (!empty($button['icon'])): ?>
 								<span class="<?php echo $button['icon'] ?>"></span>
 							<?php elseif (!empty($button['image'])): ?>
-								<?php echo $button['image']; ?>
+								<?php echo HTMLHelper::_('image', $button['image'], Text::_($button['tooltip'] ?? ''), [
+									'class' => 'icon',
+								], true) ?>
 							<?php elseif (!empty($button['svg'])): ?>
 								<?php echo $button['svg']; ?>
 							<?php endif; ?>
--- a/administrator/modules/mod_login/tmpl/default.php
+++ b/administrator/modules/mod_login/tmpl/default.php
@@ -112,6 +112,8 @@ Text::script('JHIDEPASSWORD');
 					<span class="<?php echo $button['icon'] ?>"></span>
 				<?php elseif (!empty($button['image'])): ?>
 					<?php echo $button['image']; ?>
+				<?php elseif (!empty($button['svg'])): ?>
+					<?php echo $button['svg']; ?>
 				<?php endif; ?>
 				<?php echo Text::_($button['label']) ?>
 			</button>
--- a/modules/mod_login/tmpl/default.php
+++ b/modules/mod_login/tmpl/default.php
@@ -120,6 +120,8 @@ Text::script('JHIDEPASSWORD');
 						<span class="<?php echo $button['icon'] ?>"></span>
 					<?php elseif (!empty($button['image'])): ?>
 						<?php echo $button['image']; ?>
+					<?php elseif (!empty($button['svg'])): ?>
+						<?php echo $button['svg']; ?>
 					<?php endif; ?>
 					<?php echo Text::_($button['label']) ?>
 				</button>
--- a/plugins/system/webauthn/src/PluginTraits/AdditionalLoginButtons.php
+++ b/plugins/system/webauthn/src/PluginTraits/AdditionalLoginButtons.php
@@ -162,7 +162,7 @@ trait AdditionalLoginButtons
 				'id'                 => $randomId,
 				'data-webauthn-form' => $form,
 				'data-webauthn-url'  => $url,
-				'image'              => $image,
+				'svg'                => $image,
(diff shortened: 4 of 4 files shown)


infograf768:
@nikosdion 
Does this break your former PR?

components/com_users/tmpl/login/default_login.php
```diff
@@ -92,9 +92,7 @@ class="btn btn-secondary <?php echo $button['class'] ?? '' ?>"
 							<?php if (!empty($button['icon'])): ?>
 								<span class="<?php echo $button['icon'] ?>"></span>
 							<?php elseif (!empty($button['image'])): ?>
-								<?php echo HTMLHelper::_('image', $button['image'], Text::_($button['tooltip'] ?? ''), [
-									'class' => 'icon',
-								], true) ?>
+								<?php echo $button['image']; ?>
```
nikosdion: This change is wrong and I consider it a blocker to merging this PR.

The `image` parameter returned by additional login button plugins is NOT meant to be an absolute URL. It's supposed to provide a URL relative to the `media` folder which can be overridden in the template. That's why it goes through `HTMLHelper::image`.

This is an important feature to preserve. By going through HTMLHelper the site integrator can drop in an image in `media/templates/site/YOUR_TEMPLATE/...` to override the image used by the additional login button plugin. With your change it's no longer possible. The site integrator would have to core hack the plugin itself(!!!) to change the image or do other unholy things, like hiding the image with CSS and using some sort of `::before` magic to render a different image. Not exactly what I'd call as easy or practical or even feasible as dropping a file in a folder.

If you have a problem with the image for WebAuthn not being displayed the solution would be to edit the WebAuthn plugin to send the correct **relative** path.

components/com_users/tmpl/login/default_login.php
```diff
@@ -79,7 +79,7 @@
 				<div class="com-users-login__submit control-group">
 					<div class="controls">
 						<button type="button"
-								class="btn btn-secondary <?php echo $button['class'] ?? '' ?>"
+								class="btn btn-secondary w-100 <?php echo $button['class'] ?? '' ?>"
```
nikosdion, line 82: This change is OK with me. If a third party plugin needs to override the width it can do so with CSS.

nikosdion (changes requested):
This PR breaks an important feature and needs to be changed before being considered for merging.

ceford:
I have tested this item :white_check_mark: successfully on 6cd62f7a082adf209958d2e38bd618226dfa7fb1<br /><br />I don't have a secret key but the form looks fine after applying the patch.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32196">issues.joomla.org/tracker/joomla-cms/32196</a>.</sub>

nikosdion:
@ceford Please withdraw your successful test because it's misleading. The PR as is fixes the form but **breaks important functionality** as I've written in my review of the PR in https://github.com/joomla/joomla-cms/pull/32196

Please do not limit your source of information to the Joomla issue tracker site. It doesn't convey the entire picture (and really has no reason of existence, but that's another discussion for another day).

ceford:
I have not tested this item.<br /><br />I looked at the diff and it looked simple! But, untested as requested.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32196">issues.joomla.org/tracker/joomla-cms/32196</a>.</sub>

components/com_users/tmpl/login/default_login.php
infograf768: I was sure there could be an issue. This was why I called you.
The problem lies in a double use of the htmlhelper.
Any proposal to help solving the issue in the plugin itself as it uses an inline SVG?

infograf768:
@N6REJ 
I am afraid my solution here is no good and the plugin has to be corrected.

components/com_users/tmpl/login/default_login.php
nikosdion: The solution is practically screaming at you. The plugin (`plugins/system/webauthn/src/PluginTraits/AdditionalLoginButtons.php`) needs to be corrected in line 165. Instead of `'image' => $image,` you need to do `'svg' => $image,` Then revert the change you made in your PR in the `components/com_users/tmpl/login/default_login.php` file regarding the `image` parameter.

As you can see, if you pass an `svg` parameter it's output verbatim. @brianteeman added that on January 2nd _exactly so_ that you can use inline SVG files. The same code needs to be copied to `modules/mod_login/tmpl/default.php` and `administrator/modules/mod_login/tmpl/default.php`. 

If you look at what's going on, using Git blame, you will see that my original contribution only had a provision for going through HTMLHelper::image. The inline SVG came later and, frankly, the lack of direction or understanding of where that code comes from (nobody heard of Git blame?) led to some code being partially there, some code being modified without ever working (AND MERGED! WTF!) and the three files I had carefully modified to implement additional login buttons are now all out of sync, **each one working in a different and mutually incompatible way**. Little wonder shit's broken all over the place, don't you bloody think?!!!

The PRs which made these changes should have never been merged. If anyone had looked at them they'd have figured out why. Instead, they got two ostensibly successful tests — even when _the first test in this PR_ clearly demonstrated why these tests mean sod all in practice, as I've been saying for years — and someone with commit rights mashed that green GitHub merge button without thinking and **of course** without being held accountable. What are you doing @wilsonge @HLeithner ? Are you still running this project or is your role confined to applying major changes after beta 6? This is a serious question. If you think that Joomla's problem is whether it ships with BS4 or BS5 but not whether it actually, dunno, _bloody works consistently or even at all_ then I think that your priorities are misaligned with the priorities of the people who actually try to use the darned product.

brianteeman: what have I done wrong this time. afaict my only involvement was to fix where a pr had been applied in one place but not another

nikosdion: @brianteeman Your PR was the only thing that wasn't broken. Another commit removed the code that was applied elsewhere. Yet another one broke the plugin to begin with. Your code is the last remaining example of the _correct_ approach (use a different key for inline SVG) and should be copied over in the other two places.

infograf768: I had tested already changing image to svg in the plugin and not modifying the `default_login.php`
In that case. it was working OK for the menu item but I lost the image in the login module.
I did not think about adding the svg conditional in the modules. :)
Modifying PR now.
Thanks for the hint.

infograf768:
@nikosdion @ceford 
Modified PR as discussed above.
(tooltip is still not taken care of and should be the matter of another patch)

Please test again

ceford:
It works as before for me - and I did read the discussion this time. However, being of a nervous disposition, I will wait for @nikosdion approval before pressing the pass button.

richard67:
> It works as before for me - and I did read the discussion this time. However, being of a nervous disposition, I will wait for @nikosdion approval before pressing the pass button.

@ceford No need to be nervous. If your test was successful, press the button. If it later turns out it was a mistake, that can happen. It happens to me, too sometimes. But that should not make us scared from further testing. I'm not sure if Nik will report back when all is ok now.

ceford:
I have tested this item :white_check_mark: successfully on 85c9df053ddb540071b1f29ee6a76e51649b028e<br /><br />Looked at the places changed in the diff - all seem OK.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32196">issues.joomla.org/tracker/joomla-cms/32196</a>.</sub>

nikosdion (approved):
Perfect! The code is consistent, elegant and (mostly) self-documenting. The original intent is preserved. The inline SVG implementation intent is restored. Things work. Well done!

nikosdion:
@richard67 I always comment when I ask someone to fix something and they do. Just not on Sundays, man.

alikon:
I have tested this item :white_check_mark: successfully on 85c9df053ddb540071b1f29ee6a76e51649b028e<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32196">issues.joomla.org/tracker/joomla-cms/32196</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32196">issues.joomla.org/tracker/joomla-cms/32196</a>.</sub>

richard67:
> @richard67 I always comment when I ask someone to fix something and they do. Just not on Sundays, man.

@nikosdion Yes, and thanks for that. I haven't said you won't comment for sure, I just have said that it _might_ not happen. I had the feeling @ceford was discouraged from marking his test result, and I wanted to encourage him again.

richard67:
Thanks!

PR #32204, opened 2021-01-30
[4.0] Protect bootstrap.loadComponent from being redefined

Fedik:
### Summary of Changes

This internal changes. To prevent bootstrap.loadComponent from being redefined.
The bootstrap components should be a list of predefined assets.

For now I suggest just make this method "private", so later one we can improve/trash it safely

### Testing Instructions

Apply patch. All should work as before.

All works

All works

none

bip bip @dgrammatiko @wilsonge

### DIFF
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -61,7 +61,7 @@ abstract class Bootstrap
 		}
 
 		// Include Bootstrap component
-		HTMLHelper::_('bootstrap.loadComponent', 'alert');
+		static::loadComponent('alert');
 
 		$doc           = Factory::getDocument();
 		$scriptOptions = $doc->getScriptOptions('bootstrap.alert');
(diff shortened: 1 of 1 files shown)


dgrammatiko:
I have tested this item :white_check_mark: successfully on 5515b5ea56ca8c7e4cecb348d48400c1f53c18fa<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32204">issues.joomla.org/tracker/joomla-cms/32204</a>.</sub>

wilsonge:
You can't override a single method anymore with the namespaced registry. You can only override the entire class. So not really sure if this solves much?

bembelimen:
I wonder, why do you want to prevent this at all? If someone explicit want to change this method, why not allowing it?

Fedik:
> You can only override the entire class. So not really sure if this solves much?

Well, I not very know how new HTML registry work, so just to be safe,
Someone still can extend the class.

> If someone explicit want to change this method, why not allowing it?

This method need to be improved, and possible will be need to remove in future.
Example, the assets need to be predefined and loaded via `$wa->useScript('bootstrap.blabla')`. 
Currently I do not have a time for full changes.

dgrammatiko:
> This method need to be improved, and possible will be need to remove in future.

Ok, I will do that and drop this stupid method.

PS If I didn't mess up https://github.com/joomla/joomla-cms/pull/32208 should do the trick

Fedik:
That was fast :smile:

PR #32207, opened 2021-01-30
[4.0] Dependency - Bump joomla/filter to 2.0.0-beta5

richard67:
Pull Request for Issue # .

### Summary of Changes

Update the joomla/filter dependency to 2.0.0-beta5.

### Testing Instructions

Check that the CMS continues to function as normal.
In particular, check that:
- Filtering works as usual for fields.
- Saving Global Configuration works.

Check that CI tests succeed.

CMS works. CI tests succeed.

CMS works. CI tests succeed.

None.


richard67:
It seems the input filter fails when saving Global Configuration with this PR.

richard67:
```
PHP Recoverable fatal error:  Object of class stdClass could not be converted to string
in /home/richard/lamp/public_html/joomla-cms-4.0-dev/libraries/vendor/joomla/filter/src/InputFilter.php on line 239,
referer: https://www.joomla-40-dev.vmkubu02.vmnet2.local/administrator/index.php?option=com_config
```

zero-24:
This sounds like that `$source` variable seems to be a stdClass here: https://github.com/joomla-framework/filter/blob/2.0-dev/src/InputFilter.php#L239

richard67:
> This sounds like that `$source` variable seems to be a stdClass here: https://github.com/joomla-framework/filter/blob/2.0-dev/src/InputFilter.php#L239

@zero-24 Yes. And it happens also when updating to 2.0.0-beta3, which doesn't include yet my change for the path filter coming with beta4. So the problem is not related to my recent activities ;-) Currently J4 uses beta2.

richard67:
I've opened an issue in the filter repo: https://github.com/joomla-framework/filter/issues/41 .

richard67:
The problem is caused here

https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/FormField.php#L1171

by `$value` being an object.

The field for which it is happening is the one with `$this->name` being equal to `'jform[filters]'`.

If I add some code like this before that line, saving global configuration works again with the updated filter package:

```
		if ($value && \is_object($value))
		{
			$value = json_decode(json_encode($value), true);
		}

```

But that's just a hack and very likely not a valid fix.

@Fedik As you know much about fields: Do you have an idea what would be the right fix? The filter package is more strict regarding the value being either a string or an array and so doesn't accept objects anymore.

nibra:
Just a side note:

> The filter package is more strict regarding the value being either a string or an array and so doesn't accept objects anymore.

Although the InputFilter accepted objects in the past, it did not treat them in a senseful way.

joomdonation:
Look like a bug in filter method of FormField https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/FormField.php#L1094

I think we should only filter data of form field if filter is set for that field. Right now, we still call InputFilter::getInstance()->clean($value, $filter) even when $filter = '' (see https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/FormField.php#L1171 ) and it is causing the error here.

I think that line of code should be moved inside if condition check and outside that if check, we just return $value directly (no filter)

richard67:
@joomdonation Thanks, I meanwhile came to the same conclusion. The field has explicitly set the filter to '' in the XML file for Global Config, and so it says it doesn't want any validation.

@zero-24 @nibra Does anything speak against it regarding security?

joomdonation:
This should be the fix https://github.com/joomla/joomla-cms/compare/4.0-dev...joomdonation:patch-4 for the issue, I think.

nibra:
>Does anything speak against it regarding security?

Not from my POV. If a field needs filtering, it should state it explicitly in the XML.

richard67:
> This should be the fix [4.0-dev...joomdonation:patch-4](https://github.com/joomla/joomla-cms/compare/4.0-dev...joomdonation:patch-4) for the issue, I think.

@joomdonation Yes.

@nibra @wilsonge @zero-24 Do you think we should do this fix in this PR here? Or would it be better to have 2 separate PRs?

joomdonation:
Ah, No. Might be security issue, I'm afraid of. Right now, when there is no filter set, we are still performing filter (for XSS and other 'bad' code...). Check default case in clean method of libraries/vendor/joomla/filter/src/InputFilter.php on your Joomla 4.0 installation and you will see that. We will have to find a different way to fix this issue.

joomdonation:
Found the error and made a PR to framework filter package https://github.com/joomla-framework/filter/pull/42 . Could you please have a look at it?

nibra:
I'd start the method with something like
```php
if (is_object($value)) {
    return (object) $this->filter((array)$value, $group, $input);
}
```

Then you send an array to the filter, which is fine, and return an object, which might be expected by the caller.
It's not perfect in case you send objects other than stdClass, but that should not happen in this context, anyway.

richard67:
@nibra Before or after this if clause?

https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/FormField.php#L1096-L1100

And should we do it with this PR here so we only can test that it works like before? Or should we do 2 separate PRs so we can clearly test what the other PR fixes?

nibra:
I'd say after that if statement.

> Or should we do 2 separate PRs so we can clearly test what the other PR fixes?

I have no strong oppinion about that. Do whatever makes you more confident.

richard67:
I will update this PR here soon, including the description, but the latter may take a bit. Stay tuned.

richard67:
@nibra Doesn't work, same error as before at the same place (but maybe for a different field?).

richard67:
@nibra If I do it like this, it works:

```
		if (is_object($value))
		{
			return (object) $this->filter(json_decode(json_encode($value), true), $group, $input);
		}
```

Fedik:
Hard to say, need to debug.
But why Filter does not accept objects anymore? that would fix everything :) 

The objects will come when the field has a complex value
https://github.com/joomla/joomla-cms/blob/f980bfd8dca6ce827ff36c6eb52cc4aafafb06c0/libraries/src/Form/Form.php#L1149
Because Registry  transforms multidimensional array to object and $registry->get('blabla');  return an object (or array of objects)

I would allow objects for Filter. But I do not know what the story behind of this strange decision.

richard67:
> I would allow objects for Filter.

I've made a PR for the master branch of the filter package, because I think it should be fixed for J3, too: https://github.com/joomla-framework/filter/pull/43 . The same patch applied to the filter package updated by this PR here makes it work here.

richard67:
@nibra For this one here it needs an upmerge of the master branch to the 2.0-dev branch again in the framework repo, and then a new release 2.0.0-beta5, to which I can update this PR then.

richard67:
PR is ready for tests since long time.

wilsonge:
Merged on review. Thanks!

richard67:
Thanks!

richard67:
Now we are safe for #32076 being merged up via 3.10-dev into 4.0-dev.

PR #32214, opened 2021-01-30
[4.0] Fixing series of Docblocks and mapped class inheritances

Hackwar:
I ran https://psalm.dev/ over our codebase and fixed a few of the reported issues. This replaces the old class names with the new namespaced ones in a few docblocks and also changes the parent classes from a few classes from the old (mapped) names to the new namespaced ones.


wilsonge:
Thanks!

PR #32208, opened 2021-01-30
[4.0] Use Web Asset manager for bootstrap

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Define and properly use the javascript files for Bootstrap

### Testing Instructions

Needs `npm install`. Make sure that all interactive Bootstrap components are still functional 

### Documentation Changes Required

No, not for this one

PR #32217, opened 2021-01-31
[4.0] Fix bootstrap.framework

Fedik:
### Summary of Changes

Additionally to #32208

### Testing Instructions

call `HTMLHelper::_('bootstrap.framework')`

error

works

none

bip bip @wilsonge @dgrammatiko

### DIFF
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -528,11 +528,15 @@ abstract class Bootstrap
 	 */
 	public static function framework($debug = null) :void
 	{
+		$wa = Factory::getApplication()
+			->getDocument()
+			->getWebAssetManager();
+
 		array_map(
-			function ($script) {
-				HTMLHelper::_('bootstrap.' . $script);
+			function ($script) use ($wa) {
+				$wa->useScript('bootstrap.' . $script);
 			},
-			static::$scripts
+			['alert', 'button', 'carousel', 'collapse', 'dropdown', 'modal', 'popover', 'scrollspy', 'tab', 'toast', 'tooltip']
 		);
 	}
 
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -536,7 +536,7 @@ abstract class Bootstrap
 			function ($script) use ($wa) {
 				$wa->useScript('bootstrap.' . $script);
 			},
-			['alert', 'button', 'carousel', 'collapse', 'dropdown', 'modal', 'popover', 'scrollspy', 'tab', 'toast', 'tooltip']
+			['alert', 'button', 'carousel', 'collapse', 'dropdown', 'modal', 'popover', 'scrollspy', 'tab', 'toast']
 		);
 	}
 


dgrammatiko:
Ooops

dgrammatiko:
I have tested this item :white_check_mark: successfully on d6d4cb443bf980144bab0126df524c3dd61d5941<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32217">issues.joomla.org/tracker/joomla-cms/32217</a>.</sub>

libraries/src/HTML/Helpers/Bootstrap.php
```diff
@@ -528,11 +528,15 @@ public static function toast($selector = '.toast', $options = []) :void
...
+			->getWebAssetManager();
+
 		array_map(
-			function ($script) {
-				HTMLHelper::_('bootstrap.' . $script);
+			function ($script) use ($wa) {
+				$wa->useScript('bootstrap.' . $script);
 			},
-			static::$scripts
+			['alert', 'button', 'carousel', 'collapse', 'dropdown', 'modal', 'popover', 'scrollspy', 'tab', 'toast', 'tooltip']
```
dgrammatiko: @Fedik you don't need `tooltip` as `poopover` and `tooltip` are bundled together
```suggestion
			['alert', 'button', 'carousel', 'collapse', 'dropdown', 'modal', 'popover', 'scrollspy', 'tab', 'toast']
```

Fedik: okay, I just copied old scripts :smile:

dgrammatiko:
I have tested this item :white_check_mark: successfully on f79c24106ff1d4a9d06e879ddeb8c1b24747681c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32217">issues.joomla.org/tracker/joomla-cms/32217</a>.</sub>

libraries/src/HTML/Helpers/Bootstrap.php
```diff
@@ -528,11 +528,15 @@ public static function toast($selector = '.toast', $options = []) :void
 	 */
 	public static function framework($debug = null) :void
 	{
+		$wa = Factory::getApplication()
+			->getDocument()
+			->getWebAssetManager();
+
 		array_map(
-			function ($script) {
-				HTMLHelper::_('bootstrap.' . $script);
```
dgrammatiko, line 533: After some coffee I think this change is not doing what people would expect. Calling each method without any selector/options will pass the default selector and options. Now this method is not gonna initialise any components it just load the scripts, so I think for shake of B/C we should keep calling each method instead of loading just the scripts

Fedik: This should just add a script and nothing more, look in old Joomla 3 bootstrap.framework method.

With use HTMLHelper::_('bootstrap.' . $script); it will add bunch of unused stuff in scriptOptions. 
That a bug, even if someone expect it.

dgrammatiko: > look in old Joomla 3 bootstrap.framework method

Yup, this is fine

joeforjoomla: @dgrammatiko loads only the script! I don't want to have the scriptOptions with a fake dummy selector, given that i want to use Bootstrap components in my own javascript files!

wilsonge:
Thanks!

PR #32225, opened 2021-01-31
[4.0] Stop tabs shifting and fix contrast issue

brianteeman:
Apply pr and rebuild css with npm i and test the TWO issues

### Before
before

### After
after

### Before
image

### After
image

### DIFF
--- a/administrator/templates/atum/scss/vendor/joomla-custom-elements/joomla-tab.scss
+++ b/administrator/templates/atum/scss/vendor/joomla-custom-elements/joomla-tab.scss
@@ -136,12 +136,12 @@ joomla-tab {
       position: relative;
       display: block;
       padding: .75em 1em;
+      margin: -1px 0;
       color: var(--atum-special-color);
       text-decoration: none;
       border-top: 1px solid #fff;
       border-bottom: 1px solid $gray-300;
       box-shadow: none;
-      margin: -1px 0;
 
       &[active],
       &:hover {
--- a/administrator/templates/atum/scss/vendor/joomla-custom-elements/joomla-tab.scss
+++ b/administrator/templates/atum/scss/vendor/joomla-custom-elements/joomla-tab.scss
@@ -57,7 +57,7 @@ joomla-tab {
         margin: -1px;
         background-color: $white;
         background-image: none;
-        border: 2px solid var(--bluegray);
+        border: 1px solid var(--bluegray);
         border-bottom: solid 1px #fff;
         border-radius: 0;
         box-shadow: none;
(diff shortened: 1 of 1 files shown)


wilsonge:
I'd made this change on the original BS5 branch because it fixed the vertical tabs spacing. Now when you hover over the vertical tabs (e.g. the user groups on the permissions form field) you get a vertical padding tweak. This PR is probably right as it's the original behaviour - but we need to figure out where the the extra spacing comes from on the vertical tabs before we can merge otherwise it just moves the fix from one place to another.

brianteeman:
Its not as noticable on the vertical ;)

wilsonge:
Adding `margin: -0.5px` to line 147 https://github.com/joomla/joomla-cms/pull/32225/files#diff-d250836f7da8e2204b9426a82d83049c1d94f8a8e359ac5b7efac2603b93fd0cL147 (i.e. above the color definition currently on that line) seems to fix the vertical tab issue. But I'm sure there must be some higher reason why horizontal tabs need -1px margin and vertical tabs only -0.5px from the bs4 move

wilsonge:
> Its not as noticable on the vertical ;)

@Quy pointed it out on the original PR - I can't get anything past him :D

brianteeman:
the margin is needed due to the thickness of the border iirc

wilsonge:
The extra border is common to both elements both. maybe just not noticeable as a 0.5px gap on hovered horizontal tabs as it doesn't shift the content beneath it?

brianteeman:
border-top is 2px
border left and right is 1px

So the correct fix is to change border-top on vertical to 1px

brianteeman:
almost but not quite. I have it working on everything but when it is active AND hover

Will look again tomorrow sometime

brianteeman:
@wilsonge should be ok for both horizontal AND vertical now

administrator/templates/atum/scss/vendor/joomla-custom-elements/joomla-tab.scss
```diff
@@ -138,9 +138,10 @@ joomla-tab {
       padding: .75em 1em;
       color: var(--atum-special-color);
       text-decoration: none;
-      border: 0;
+      border-top: 1px solid #fff;
```
Fedik: if you like, you can use `transparent` color, then it will work with any background :wink:

richard67:
Code style: https://ci.joomla.org/joomla/joomla-cms/39763/1/20

wilsonge:
Tested this locally and can confirm it works. Merging for tomorrow's beta.

brianteeman:
Thanks.

PR #32230, opened 2021-02-01
[4.0] Fixing issues from Psalm quality checks

Hackwar:
This is another batch of fixes to docblocks and method calls from the Psalm quality tool. This can not be tested, but needs to be reviewed.


PhilETaylor:
Careful... you will get dragged down a huge rabbit hole updating the docblocks and the types of each param... literally thousands of them are wrong. I looked to fix them but was overwhelmed! Making joomla type safe is going to be a nightmare in the future.

Hackwar:
Which is why I'm starting small and doing little steps. Not claiming to fix every problem everywhere, but when I run into them, I can fix them. Every improvement helps when later working with an IDE.

PhilETaylor:
Totally agree. phpStorm has a nightmare at the moment  :-(

libraries/src/Cache/Controller/CallbackController.php
```diff
@@ -183,7 +183,7 @@ public function store($data, $id, $group = null, $wrkarounds = true)
 	/**
 	 * Generate a callback cache ID
 	 *
-	 * @param   callback  $callback  Callback to cache
```
wilsonge, line 186: Whilst obviously callback isn't a thing. This new block isn't true either as you can see from the first line of the function it handles objects and array's that aren't direct callable's. I think until we look closer for now this just needs to be `mixed`

libraries/src/Component/Router/RouterViewConfiguration.php
```diff
@@ -147,7 +147,7 @@ public function setKey($key)
 	 *
 	 * @since   3.5
 	 */
-	public function setParent(RouterViewconfiguration $parent, $parentKey = false)
+	public function setParent(RouterViewconfiguration $parent, $parentKey = null)
```
wilsonge, line 150: Why not just an empty string and use `empty()`?

wilsonge:
One fix required and this is good to merge

libraries/src/Cache/Controller/CallbackController.php
Hackwar: fixed

wilsonge:
Thanks!

Issue #32213, opened 2021-01-30
[4.0] dropdown icons


brianteeman:
shouldnt they be fixed width ?

![image](https://user-images.githubusercontent.com/1296369/106368854-dd7fd380-6344-11eb-90fc-46fef6997394.png)

![image](https://user-images.githubusercontent.com/1296369/106368856-df499700-6344-11eb-8cf3-fccbd2e5a947.png)

Quy:
Yes so the labels can be aligned.

brianteeman:
closed see #32243

PR #32243, opened 2021-02-02
[4.0] Actions menu icon width

brianteeman:
Due to the different width of the icons the icon and subsequent text are not aligned vertically

This PR applies a width and align center to the icons in the actions dropdown menu.

Pull Request for Issue #32213

As this is an scss change you need to apply patch and then npm i

### Before
image

image

### After
image

image


wilsonge:
Thanks!

brianteeman:
thanks

Quy:
Buttons are now a little wider.

Before:
32243-before

After:
32243-after

PR #32274, opened 2021-02-03
[4.0] Load the namespace map in execute

laoneo:
Pull Request for pr #31886.

### Summary of Changes
Load the namespace app in execute as the file write operation needs an application object to check if FTP should be used.

### Testing Instructions
- Copy the discover app from here https://github.com/Digital-Peak/DPDocker/blob/php/versions/webserver/scripts/discoverapp.php to the joomla root.
- Remove the lines 25 to 37 in the file discoverapp.php of your joomal root
- Link an extension with a symlink
- Delete the file /administrator/cache/autoload_psr4.php
- Execute `php discoverapp.php` in your Joomla root folder

"Failed to start application" error.

Extensions are discovered.

### DIFF
--- a/libraries/src/Application/CliApplication.php
+++ b/libraries/src/Application/CliApplication.php
@@ -127,10 +127,6 @@ abstract class CliApplication extends AbstractApplication implements DispatcherA
 				->alias(\Joomla\Session\SessionInterface::class, 'session.cli');
 		}
 
-		\JLoader::register('JNamespacePsr4Map', JPATH_LIBRARIES . '/namespacemap.php');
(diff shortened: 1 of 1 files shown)


wilsonge:
Looks fine. But I just realised we should reuse the same trait the other applications are using (`\Joomla\CMS\Application\ExtensionNamespaceMapper`) so can you quickly refactor to that at the same time please?

laoneo:
You are completely right. Now we are inline with the other apps.

wilsonge:
Thanks!

PR #29731, opened 2020-06-21
[4.0] Webauthn gmp warning

PhilETaylor:
Pull Request for Issue #29696 .

### Summary of Changes

A very light touch change to reuse existing code to display a warning and remove the ability to add authenticators if the PHP extension GMP is not loaded.

### Testing Instructions

Install Joomla 4 on a server (or docker container in my case) **without GMP PHP Extension enabled** (On a cPanel server you can toggle this, and its off by default in Easy Apache 4) 

Ensure you access Joomla over a secure & valid SSL connection (learn ngrok if you need this, quick and easy) 

### Before PR 

When adding an authenticator you get an internal server error

### After PR

You get a error and no button allowing you to add a new authenticator (but you can still see the list of existing authenticators - if any - rename them, and delete them.) 

Screenshot 2020-06-21 at 12 24 50

### Who else to ping for visibility ;-) 
// @nikosdion

PR #32289, opened 2021-02-03
[4.0] GMP system info

brianteeman:
As webauthn plugin uses the php GMP extension it made sense to add a check for this in the system info.

There is already code that handles the error message #29731 this is just for the system info list.

Perhaps it should be added to the preinstall/preupdate checks as well as a recommendation (not required) @wilsonge @zero-24  

/administrator/index.php?option=com_admin&view=sysinfo

image


Quy:
I have tested this item :white_check_mark: successfully on fdac7b814eaffa56980ecad7abcb93918a450c06<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32289">issues.joomla.org/tracker/joomla-cms/32289</a>.</sub>

administrator/components/com_admin/src/Model/SysinfoModel.php
```diff
@@ -263,6 +263,7 @@ public function &getPhpSettings()
 			'mbstring'            => extension_loaded('mbstring'),
 			'iconv'               => function_exists('iconv'),
 			'max_input_vars'      => ini_get('max_input_vars'),
+			'gmp'                 => extension_loaded('gmp'),
```
PhilETaylor, line 266: Please review https://github.com/joomla/joomla-cms/pull/29731#issuecomment-647149880 as there are good reasons why doing the lookup with `get_loaded_extensions` or `extension_loaded` is the wrong approach. 

> I'd actually change the detection code to use function_exists() with some GMP function names for two reasons.

> First, I've seen many hosts where get_loaded_extensions itself is disabled, making that kind of code fail immediately with a fatal error. I know, these hosts are crap but they are used by people using Joomla...

> Second, it's possible that the extension is enabled but the host disabled the functions. I have not seen that with GMP but I can't rule out outright stupidity when it comes to cheap hosting.


Also you can have the extension loaded in PHP-FPM, but then the actual functions disabled. I (and Nic reports) seeing this on some web hosts. 

This is why #29731 went for checking `function_exists('gmp_intval')`

The way you have implemented `extension_loaded('gmp')` means that this check COULD pass, but Webauthn could still fail if the functions are disabled by pathetic web hosting configurations.

brianteeman: IF so then all the other tests in sysinfomodel are wrong and all of the uses in the unit tests and all of these
https://github.com/joomla/joomla-cms/search?q=extension_loaded

PhilETaylor: If you want to discard real-world experience of @nikosdion (and others) then that's fine. Thats your choice, that's the Joomla Projects choice. That choice has never ended well though.

PhilETaylor: https://github.com/joomla/joomla-cms/search?q=extension_loaded = 65 Results
https://github.com/joomla/joomla-cms/search?q=function_exists = 81 Results
Also note the search doesn't search the 4.0-dev branch as far as I can tell.

brianteeman: If you are correct then all the 100+ instances that are already in use in both J3 and J4 also need to be corrected. In addition a quick grep of a live site shows that extension_loaded is used by numerous extremely popular extensions. So instead of picking on this one pull request please open an issue for all of the uses of extension_loaded in the enitre codebase of both J3 and J4

toivo:
I have tested this item :white_check_mark: successfully on fdac7b814eaffa56980ecad7abcb93918a450c06<br /><br />Tested successfully in Beta8-dev of 3 February using PHP 8.0.1.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32289">issues.joomla.org/tracker/joomla-cms/32289</a>.</sub>

administrator/components/com_admin/src/Model/SysinfoModel.php
PhilETaylor: Just merge it as is then. I don't really care. Just like you don't care to learn the difference in the functions and what they do. (Hint: BOTH are required and they do different things - replacing all of them one way or the other is NOT the right thing to be doing at all - and no one, apart from you is calling for that!)

Also you don't have a right to dictate and demand what pull requests I open, but I have every right to comment on open pull requests based on my experience and the experience of others, especially when that has already been stated.

brianteeman: I am well aware of what the difference is. I am not dictating anything. I am just stating that IF you are correct then it is a bigger issue than this PR and needs to be addressed as such. Do not put words into my mouth and stop with the attacks.

PhilETaylor: The irony is that https://github.com/joomla/joomla-cms/pull/31735/files is for the exact same type of thing.. where an extension is installed, but the function is not available LMAO... and the author.... YOU!

brianteeman: Thats not irony - thats showing the things that you dont understand. Not wasting my time on you anymore

nikosdion: FWIW this needs to be a project leadership decision and policy, not a random PR with near zero chance of success. The production leadership needs to actually accept that real world servers _do_ disable `extension_loaded` and/or individual functions of an otherwise enabled PHP extension. Once this is taken to heart it's easy to write a helper static method which takes two arguments `string $extension, array $functions` and checks if the extension is loaded (if the `extension_loaded` function is not disabled in itself) _and_ whether individual functions are available. The code is easy and this PR would take around 3-4 hours to write. 

However, no developer in their right mind would spend half a day authoring a PR only to spend another 50-100 hours debating whether it's needed (yes, it is, _obviously_) and being subjected to endless nitpicking from people who can't tell PHP from a dead squirrel. This is a colossal waste of time.

As I've said before, I have no problem authoring this kind of PRs _as long as_ there is direction and protection from the production leadership i.e. they decide that X needs to be done, they assign reviewers who know WTF they are doing and lock the discussion for the peanut gallery. It's been twelve years. Joomla has taken me up on my offer exactly once, for the Events-based plugins in Joomla 4 and that was only because at this point it was a speculative product that wasn't even in pre-alpha. If they'd like to take me up on this offer now I would carve out the necessary time from an already impossible schedule to contribute this code.

brianteeman:
To be clear as I obviously wasnt explaining myself correctly.

If my proposal to use  extension_loaded to determine if gmp is enabled in php is not suitable then dont we also have to correct all the other places where extension_loaded is used for the same thing. Otherwise all the comments about not being able to accurately determine if the extension is available using this method make all the existing tests invalid as well. Or is there something special about gmp that means it can't be tested in the same way as gd, zlib, xlib and mbstring.

Its very confusing because the php documentation says to do it this way https://www.php.net/manual/en/function.extension-loaded.php

SharkyKZ:
This PR is fine for what it is.

administrator/components/com_admin/src/Model/SysinfoModel.php
wilsonge: @nikosdion I took you up on it for the api authentication plugin too. I can also confirm that if you do the PR - I confirm I will make sure it’s merged (subject to the usual code style/unit tests etc passing)

PR #31990, opened 2021-01-10
[4.0][WIP] BS5 JS

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

This PR changes the JS to Bootstrap V5. Although it's a work in progress after https://github.com/joomla/joomla-cms/pull/31990/commits/d96ee241db98e6f943a0c6f931e440590eb15d00 all the components are functional so **help by testing and pointing out what is BROKEN** will be invaluable. Thanks!

### Testing Instructions
Check everything backend/front end

### What's broken:
- [x] [FIXED] Field Modals  (the extra buttons for editing/creating new Item in various forms, Modals are mostly ok)
- [x]  Side by Side Editing com Associations
- [x]  The Toolbar dropdown buttons have wrong html the dropdown needs to be the next sibling element to the button (who did this? )
- [x] [FIXED] Also Multilingual Status broken (empty popup)
- [x]  Frontend Editing
- [x] Please fix Copy Template, Manage Folders and New File modals under System>SiteTemplates>Cassiopeia Details and Files.
- [x]  Other things that I haven't tested yet

### Documentation Changes Required
There is a component that validates what's working at https://github.com/dgrammatiko/bs5/blob/main/BS5Tests.zip This can act as your DEV documentation (also it can act as part of the e2e tests)...

PR #32290, opened 2021-02-03
s/intead/instead

PhilETaylor:
s/intead/instead

Fix typo introduced in https://github.com/joomla/joomla-cms/pull/31990/


Quy:
Thanks.

Issue #32136, opened 2021-01-24
[4.0] Helper.file path convention for modules


webchun:
### Steps to reproduce the issue
I see that all default modules store its helper.php file under Helper folder, for example `modules/mod_articles_archive/src/Helper/ArticlesArchiveHelper.php`
So I'm assuming when I create a custom module, I should place my helper.php file under folder :
`modules/mod_my_module_name/src/Helper/ and name it MyModuleName.php`

But when I try to use com_ajax within my module by using : 
`index.php?option=com_ajax&module=my_module_name&method=getList&format=json `
it returned 
`{"success":false,"message":"The file at mod_my_module_name\/helper.php does not exist.","messages":null,"data":null}
`
Then when I look into com_ajax file : `/components/com_ajax/ajax.php`
the path is indeed hardcoded to :
`$helperFile = JPATH_BASE . '/modules/mod_' . $module . '/helper.php';`

### System information (as much as possible)
Joomla 4 Beta 5

Fedik:
That is namespacing issue.
But can be a problem with com_ajax, I agree.

Either need to fix com_ajax or use dummy helper.php that include MyModuleHelper.php

Fedik:
a better title for the issue: "[4.0] com_ajax do not work with namespaced module"
or something in this way.

dgrammatiko:
This is a Release Blocker

cc @richard67

richard67:
See also #25685 .

joomdonation:
Please test PR https://github.com/joomla/joomla-cms/pull/32291

webchun:
Hi guys, I know this issue is already closed but is this issue really already solved because the issue is still persist in Joomla 4.0.3. Also, it looks like com_ajax still looks for Mod class prefix in helper class name.

dgrammatiko:
@webchun check https://github.com/dgrammatiko/invalidate-cache/tree/main/src/invalidatecache 
There are some changes between J3 and J4

webchun:
> @webchun check https://github.com/dgrammatiko/invalidate-cache/tree/main/src/invalidatecache There are some changes between J3 and J4

I din't see that relate with my question can you explain it?

dgrammatiko:
Follow the changes:
- https://github.com/joomla/joomla-cms/pull/32633
- https://github.com/joomla/joomla-cms/pull/33182

Or just follow the module I just referenced above as it implements the J4 way...

PR #30816, opened 2020-09-30
[4.2] Add namespace support for templates

thednp:
Pull Request for Issue #30272.

### Summary of Changes
As discussed in the #30272 report, Joomla does not support namespaced templates, this commit solves that.

* the `create()` method now checks for namespaced `template` extension types
* the `getNamespaces()` method properly recognize *Administrator* and *Site* templates; in addition, it will look for `templateDetails.xml` file instead of `extensionName.xml` and read the designated namespace (Line 182)

### Testing Instructions
In order to enable namespace for templates, here are the few steps, we'll use the default site template:
* open `site/templates/cassiopeia/templateDetails.xml` and add between `<description>` and `<files>` the following
```markup
<namespace path="src">Joomla\Template\Cassiopeia</namespace>
```
* delete `site/administrator/cache/autoload_psr4.php`
* login to your `site/administrator`
* open the newly created `site/administrator/cache/autoload_psr4.php`, the last element of the return object should be
```
'Joomla\\Template\\Cassiopeia\\Site\\' => [JPATH_SITE . '/templates/cassiopeia/src']
```
* *optional*:  create a custom field in `site/templates/cassiopeia/src/Field` folder and add it to the `templateDetails.xml` along with `addfieldprefix="Joomla\Template\Cassiopeia\Site\Field"` attribute added to the parent `<fieldset>`
* *optional* try and call the newly added custom field from your namespaced template into a module/plugin, test it out

### Actual result BEFORE applying this Pull Request
The `namespacemap.php` class had no support for Site Templates / Administrator Templates. `addfieldpreffix` couldn't be used with template defined custom fields.

### Expected result AFTER applying this Pull Request
Site Templates / Administrator Templates can be developed with namespace features.

### Documentation Changes Required
Yes. (I wish I knew where)

### Questions
* should I commit this change to the `4.1-dev` branch as well?
* should I commit this change to the `staging` branch? If yes, where is the `namespacemap.php` file?

PR #32291, opened 2021-02-03
[4.0] Support namespace module for com_ajax

joomdonation:
Pull Request for Issue #32136 .

### Summary of Changes
Modify com_ajax to support namespace module.

### Testing Instructions
1. Create an instance of module Articles - Categories, make sure it is enabled on all pages. Publish the module
2. Modify the file modules/mod_articles_categories/src/Helper/ArticlesCategoriesHelper.php, add a simple method for testing com_ajax
```php
public static function getDataAjax()
{
	$data = ['test'];

	return $data;
}
```
3. Access to this URL to trigger com_ajax for the module

http://localhost/joomla4/index.php?option=com_ajax&module=articles_categories&method=getData&format=json

4. Before patch, you will get this error
{"success":false,"message":"The file at mod_articles_categories\/helper.php does not exist.","messages":null,"data":null}

5. After patch, you will get this success response:
{"success":true,"message":null,"messages":null,"data":["test"]}

You can install the modified mod_articles_categories.zip which I attached here so that you don't have do the step #2 (modify code)
[mod_articles_categories.zip](https://github.com/joomla/joomla-cms/files/5919797/mod_articles_categories.zip)

### DIFF
--- a/components/com_ajax/ajax.php
+++ b/components/com_ajax/ajax.php
@@ -118,7 +118,6 @@ elseif ($input->get('module'))
 				try
 				{
 					$results = call_user_func($class . '::' . $method . 'Ajax');
-
 				}
 				catch (Exception $e)
 				{
--- a/components/com_ajax/ajax.php
+++ b/components/com_ajax/ajax.php
@@ -63,8 +63,6 @@ elseif ($input->get('module'))
 
 	if ($moduleId && $table->load($moduleId) && $table->enabled)
 	{
-		$helperFile = JPATH_BASE . '/modules/mod_' . $module . '/helper.php';
-
 		if (strpos($module, '_'))
 		{
 			$parts = explode('_', $module);
@@ -73,21 +71,34 @@ elseif ($input->get('module'))
 		{
 			$parts = explode('-', $module);
 		}
-
-		if ($parts)
+		else
 		{
-			$class = 'Mod';
+			$parts = [$module];
+		}
 
-			foreach ($parts as $part)
-			{
-				$class .= ucfirst($part);
-			}
+		$xmlFile = JPATH_BASE . '/modules/mod_' . $module . '/mod_' . $module . '.xml';
 
-			$class .= 'Helper';
+		if (is_file($xmlFile) && $xml = simplexml_load_file($xmlFile))
+		{
+			$namespace = (string) $xml->namespace;
+			$src       = $xml->namespace->attributes('src') ?: 'src';
 		}
 		else
 		{
-			$class = 'Mod' . ucfirst($module) . 'Helper';
+			$namespace = '';
+			$src       = '';
+		}
+
+		if ($namespace)
+		{
+			$helperClassName = implode('', array_map('ucfirst', $parts)) . 'Helper';
+			$helperFile      = JPATH_BASE . '/modules/mod_' . $module . '/' . $src . '/Helper/' . $helperClassName . '.php';
+			$class           = $namespace . '\\' . ucfirst($app->getName()) . '\\Helper\\' . $helperClassName;
+		}
+		else
+		{
+			$helperFile = JPATH_BASE . '/modules/mod_' . $module . '/helper.php';
+			$class      = 'Mod' . implode('', array_map('ucfirst', $parts));
 		}
 
 		$method = $input->get('method') ?: 'get';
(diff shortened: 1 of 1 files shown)


laoneo:
Better to use the Joomla 4 way and boot the module. Then you don't have to manually include the file. How to load the helper can then be done with extending the module interface. But manually construction namespaces and class names should be something of the past.

Fedik:
> Better to use the Joomla 4 way and boot the module. 

@laoneo please show an example, 
it is not obvious how to do that, all I can find is `ModuleHelper::getModule` that does not tell anything about `namespace`

Fedik:
If you mean `$app->bootModule($module->module, $app->getName())`.
It does not help here, because it  does not allow to get namespace even if module have a namespace. 

So there no other way than just parse XML directly.
Same issue with Template #30816

joomdonation:
@laoneo Look at how boot the module works for Joomla 4, I still don't understand how it would help here. Could you please explain more details? Or if you could work on this Release Blocker item, that would be better.

ghazal:
Tested as explained.
Works as expected in the description.

laoneo:
But you can extend the module interface to get the module helper class. SOmething like `$app->bootModule($module->module, $app->getName())->getHelper();` for example. Then getHelper does all the internal loading and returns the helper object. I mean it is just an idea but much better than compiling class names and loading files through require which we not should do anymore in J4.

joomdonation:
So basically, move the logic to load helper class to https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Extension/Module.php ?

laoneo:
You can either move your logic there. Better and cleaner way would be to introduce a helper factory, similar to what we did with the dispatcher factory https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Dispatcher/ModuleDispatcherFactory.php. Guess @wilsonge needs to give here his input what he would accept for 4.0.

Fedik:
> Something like $app->bootModule($module->module, $app->getName())->getHelper(); for example.

@laoneo problem with Helper, they all "static", I mean they is "abstract", 
so in current state the factory cannot build "helper instance", only return the class name of the helper

laoneo:
Perhaps we need to get rid of static then.

joomdonation:
For the time being, how about adding a method to **ModuleHelper** https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Helper/ModuleHelper.php to return the helper class?

Fedik:
hm, what if we try  Anonymous classes?  `->getHelper()` will return :
```php
new class extends $helperClass{}
```

but it a bit dirty :smile: 

Another idea, something like:
```php
$dispatcher = $app->bootModule($module->module, $app->getName())->getDispatcher($module, $app);
$dispatcher->dispatchHelper('onAjaxFooBar');
```
Because a loot of helpers use `Helper::doFooBar($module, $params)` it could be useful when it will be a part of `dispatcher`.
But maybe also not a very good idea.

> For the time being, how about adding a method to ModuleHelper

Would be better if we find solution now, because it will stay forever :wink:

thednp:
Why not add same logic to templates as for modules?

Follow #30816

thednp:
@laoneo perhaps at this point we need to do something about a `getNamespace` utility. Either save it in the manifest JSON part of the extension at install, or [read it](https://github.com/joomla/joomla-cms/pull/30816/files#diff-d2d4036b521b03d9720d3e0a557f8eea32dd70bf3df26d967caf873601ff3583R261-R264) from XML the declaration. Either way I'm fine with any decision and hopeful of things moving forward.

I say let's start with my commit #30816 (RTC now) since it will be required by this commit and child templates features.

laoneo:
@thednp `getNamespace` is the wrong direction. Because we want to get away from manually compiling class names all over Joomla. The right approach would be to use some factories and inject them wherever possible to load the helpers. I know it is a bit more effort but it will payout at the end. Testing will be much more easier on a long term.

thednp:
@laoneo please lay out a plan and lets get to work, I'm ready to take some work.

laoneo:
What do you guys think about #32633?
@thednp if we can bring #32633 somehow to fly, then the same pattern can be applied to templates.

laoneo:
@joomdonation you shouldn't close it that fast. Perhaps I'm on the wrong path too :smirk:

thednp:
@joomdonation I hope you re-commit this PR with updates for #32633

joomdonation:
@thednp https://github.com/joomla/joomla-cms/pull/32633 should be better and is replacement for this PR, so this PR is closed now.

joomdonation:
> @joomdonation you shouldn't close it that fast. Perhaps I'm on the wrong path too 😏

Was waiting for your PR, hehe.

webchun:
Hi guys, I know this issue is already closed but has this issue really been resolved? It's still returning _The file at mod_articles_categories\/helper.php does not exist_ in Joomla 4.0.3. and com_ajax still looks for Mod class prefix in helper class name.

Issue #32296, opened 2021-02-04
[4.0]Beta 7 incorrectly displays content 


jschmi102:
my test-site is multilingual (German and english).
After update from beta 6 to beta 7 content in german language is correctly setup whereas content in english language overlaps right module.  
see attached sample...

![screen shot 2021-02-04 at 08 53 29](https://issues.joomla.org/uploads/1/285eeeaaec101e156c55d21b45542512.jpg)

webchun:
I think it's because of that part of 'code' in your content _({rgraph_chat...)_ that is probably using pre tags and break the template. Please add screenshot for german language too.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32296">issues.joomla.org/tracker/joomla-cms/32296</a>.</sub>

jschmi102:
hi,
you are correct, nothing to do with languages - causer is the pre-tag. But anayway this should not cause overlapping. Rather should set/use max-width of content-column (as it did with Beta 6 and bootstrap 4).  <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32296">issues.joomla.org/tracker/joomla-cms/32296</a>.</sub>

chmst:
@jschmi102
 Are you sure that this was ok in with Beta 6 and bootstrap 4? I think that this is independent of the bootstrap version.

jschmi102:
hi,
did not check problem with beta6. But I think you are right, it is not a bootstrap problem but rather a styling problem of casiopeia template. Using a pre-html tag should not break template formatting. To correct this  e. g I  enclosed article within a div-element like  `<div class="container">` then problem disappeared. (Intead of setting width of element in pixel).

drmenzelit:
@jschmi102 Can you please test https://github.com/joomla/joomla-cms/pull/32322 ?

richard67:
Closing as having a pull request. Please test #32322 . Thanks in advance.

PR #32322, opened 2021-02-05
[4.0] Cassiopeia: Wrap preformatted text blocks

drmenzelit:
Pull Request for Issue #32296  .

### Summary of Changes
Added white-space: pre-wrap for pre + code blocks

### Testing Instructions
Run npm 

Add some text in a pre or pre + code block in an article.

The pre block doesn't fit in the content, it gets a horizontal scroll bar:

grafik

The pre block adapts to the width of the content:

grafik

### DIFF
--- a/templates/cassiopeia/scss/blocks/_global.scss
+++ b/templates/cassiopeia/scss/blocks/_global.scss
@@ -166,6 +166,17 @@ figcaption {
   color: $gray-600;
 }
 
+pre,
+code {
+  white-space: pre-wrap;
+}
+
+pre {
+  code {
+    white-space: pre-wrap;
+  }
+}
+
 .mod-menu {
   flex-direction: column;
 }


brianteeman:
Are you sure this is correct as bootstrap already has classes for this with `word-wrap: break-word;`

jschmi102:
hi,
tested this on my site - works perfectly! Thanks

drmenzelit:
> 
> 
> Are you sure this is correct as bootstrap already has classes for this with `word-wrap: break-word;`

Yes, I saw this, but that would mean the user needs to put the class in each pre block he/she inserts in an article. That is probably not really handy...

richard67:
> hi,
> tested this on my site - works perfectly! Thanks

@jschmi102 You mean you have applied the change and run npm to compile the scss again and tested with success?

If so: Could you mark your test result by going to this PR in the issue tracker here https://issues.joomla.org/tracker/joomla-cms/32322 , then use the "Test this" button, select the right test result and then submit? Thanks in advance, and thanks for testing.

jschmi102:
I have tested this item :white_check_mark: successfully on f9a39a6bba6ccce272f85f90c1a050d3d1192d06<br /><br />tested with some of my sample articles<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32322">issues.joomla.org/tracker/joomla-cms/32322</a>.</sub>
Applied change directly to template.css (template.min.css) of cassiopeia.

Harmageddon:
I don't think that this is the right solution for the problem. There are good reasons for `<pre>` tags not wrapping text. They are used to display code and other things where you want to have exactly the format you specified in the source. Automatic line breaks would be misleading in many cases.
If a `<pre>` tag is wider than the space it may use, it usually gets a vertical scrollbar. This is the case if you look at an article in the "article" view:

Screenshot_2021-02-06 Testing Home

The problem that destroys the layout only seems to occur in blog/featured layouts:
Screenshot_2021-02-06 Home

IMO, `overflow: auto` at the right place might be a better fix. I'm going to have a look where to put that.

Harmageddon:
Proposing #32335 as an alternative.

drmenzelit:
Closing this in favour of #32335

PR #32353, opened 2021-02-08
[4.0][WIP] Add fix to to repeatable fields plugin migration

wilsonge:
Pull Request for Issue #30556 .

### Summary of Changes
WIP on fixing the release blocker above

### Testing Instructions
Create A repeatable article field in 3.10 and run the migration to 4.x following the instructions in https://www.joomla.org/announcements/release-news/5833-joomla-4-0-0-beta7-and-joomla-3-10-alpha5.html. Before patch fatal error. After the patch field migrated

None

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -252,6 +252,9 @@ class JoomlaInstallerScript
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


wilsonge:
OK - This doesn't fix the release blocker - but it does stop the fatal error. Fields now convert (but in an incorrect way - the keys of the json are incorrect). I'm going to merge this to easy people's debugging attempts

PR #32356, opened 2021-02-08
remove deprecated method

rdeutz:
### Summary of Changes
Remove deprecated isAdmin and getCfg calls in core

### Testing Instructions
actions logs should work as before

### DIFF
--- a/libraries/fof/model/model.php
+++ b/libraries/fof/model/model.php
@@ -716,7 +716,7 @@ class FOFModel extends FOFUtilsObject
 
 			if (method_exists($app, 'getCfg'))
 			{
-				$default_limit = $app->getCfg('list_limit');
+				$default_limit = $app->get('list_limit');
 			}
 			else
 			{
--- a/libraries/fof/model/model.php
+++ b/libraries/fof/model/model.php
@@ -712,16 +712,7 @@ class FOFModel extends FOFUtilsObject
 		}
 		else
 		{
-			$app = JFactory::getApplication();
-
-			if (method_exists($app, 'getCfg'))
-			{
-				$default_limit = $app->get('list_limit');
-			}
-			else
-			{
-				$default_limit = 20;
-			}
+			$default_limit = JFactory::getApplication()->get('list_limit', 20);
 
 			$limit = $this->getUserStateFromRequest($component . '.' . $view . '.limit', 'limit', $default_limit, 'int', $this->_savestate);
 			$limitstart = $this->getUserStateFromRequest($component . '.' . $view . '.limitstart', 'limitstart', 0, 'int', $this->_savestate);
--- a/libraries/fof/integration/joomla/platform.php
+++ b/libraries/fof/integration/joomla/platform.php
@@ -150,7 +150,7 @@ class FOFIntegrationJoomlaPlatform extends FOFPlatform implements FOFPlatformInt
 			}
 			else
 			{
-				$isAdmin = !JFactory::$application ? false : JFactory::getApplication()->isAdmin();
+				$isAdmin = !JFactory::$application ? false : JFactory::getApplication()->isClient('administrator');
 			}
 		}
 
@@ -819,7 +819,7 @@ class FOFIntegrationJoomlaPlatform extends FOFPlatform implements FOFPlatformInt
                     $response->email = $user->email;
                     $response->fullname = $user->name;
 
-                    if (JFactory::getApplication()->isAdmin())
+                    if (JFactory::getApplication()->isClient('administrator'))
                     {
                         $response->language = $user->getParam('admin_language');
                     }
--- a/libraries/fof/view/html.php
+++ b/libraries/fof/view/html.php
@@ -135,7 +135,7 @@ class FOFViewHtml extends FOFViewRaw
 
 		// Set the document title
 		$title = $params->get('page_title', '');
-		$sitename = $app->getCfg('sitename');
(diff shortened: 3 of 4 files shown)


libraries/fof/model/model.php
```diff
@@ -716,7 +716,7 @@ public function __construct($config = array())
 
 			if (method_exists($app, 'getCfg'))
```
richard67: Should we change this, too? Or shall we leave it as it is since that method still exists, it's just deprecated?

richard67: But I'm not sure if an `if (method_exists($app, 'get'))` really makes sense.

toivo:
I have tested this item :white_check_mark: successfully on 67c09b2b6343848055986481c1db9aac4470072a<br /><br />Tested successfully in 3.9.25-dev of 11 February in Wampserver 3.2.4 using PHP 8.0.1.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32356">issues.joomla.org/tracker/joomla-cms/32356</a>.</sub>

libraries/fof/model/model.php
richard67: If all possible kinds of applications (`$app`) have a `get` method, we can replace the complete if then else here with this one line:

`$default_limit = $app->get('list_limit', 20);`

If not and we have to keep the `method_exists`, we should change that to `method_exists($app, 'get')`, and also in this case we should use a default of 20 for the get call.

libraries/fof/model/model.php
```diff
@@ -716,7 +716,7 @@ public function __construct($config = array())
...
 			{
-				$default_limit = $app->getCfg('list_limit');
+				$default_limit = $app->get('list_limit');
```
richard67:
```suggestion
				$default_limit = $app->get('list_limit', 20);
```

libraries/fof/model/model.php
richard67: @Fedik What do you think? Do we need an `if (method_exists($app, 'get'))`?

Fedik: does not need, because every joomla application extends Joomla\Application\AbstractApplication that have get()/set() method

libraries/fof/model/model.php
richard67: @SharkyKZ Not good? Why?

libraries/fof/model/model.php
richard67: Seems I'm wrong with my suggestion to use `$app->get('list_limit', 20)`, see reaction on my suggestion below. So forget that.

Fedik: The suggestion is good, but it need to replace whole if/else section, not just 1 line :wink:

richard67: Yes that’s what I had in mind, so I tried to describe it here. Because there was already this code review comment, GitHub did not allow me to add another suggestion here. I‘d have to close this one here before, and then the discussion would be hidden, that’s why I asked first.

libraries/fof/model/model.php
richard67: Sorry for misunderstanding, I tried to  make an additional suggestion because I did not want to hide the discussion, but it should of course replace the complete if then else, like I asked about in the previous comment. Will resolve these now and make a new, correct suggestion.

libraries/fof/model/model.php
richard67: Hmm, I still can't make a code suggestion to replace that if then else, because the end of the else is too far away from code changed by this PR, so I can't mark the code for suggestion so far down.

I think the complete

```
$app = JFactory::getApplication();

if (method_exists($app, 'getCfg'))
{
	$default_limit = $app->get('list_limit');
}
else
{
	$default_limit = 20;
}
```

can be replaced by

```
$default_limit = JFactory::getApplication()->get('list_limit', 20);
```

richard67: I've made a PR to suggest the change: https://github.com/rdeutz/joomla-cms/pull/15

libraries/fof/model/model.php
richard67: https://github.com/rdeutz/joomla-cms/pull/15

bembelimen:
I have tested this item :white_check_mark: successfully on 9b6eabc8d1d20d7f33c8814fb3f1ebc89c799d41<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32356">issues.joomla.org/tracker/joomla-cms/32356</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on 9b6eabc8d1d20d7f33c8814fb3f1ebc89c799d41<br /><br />on Code review of changes, as there were aleady good tests<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32356">issues.joomla.org/tracker/joomla-cms/32356</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32356">issues.joomla.org/tracker/joomla-cms/32356</a>.</sub>

Issue #32259, opened 2021-02-02
[4.0b7][Release Blocker] Frontend Editor has no access to media manager 


PhilETaylor:
### Steps to reproduce the issue

as a super admin create a menu link of type "Create Contact"

On the frontend login as an EDITOR and click menu item "Create Contact"

For the Image field click Select 

<img width="890" alt="Screenshot 2021-02-02 at 22 07 33" src="https://user-images.githubusercontent.com/400092/106669098-10270780-65a3-11eb-92a5-49227163d3b7.png">

### Expected result

If the image field is there, and a select button is there, then the expectation is that it can be used to select media from the media manager.

### Actual result

Because the EDITOR doesn't have "Access Administration Interface" enabled by default in ACL, they cannot access Media Manager.

NOTE JS Console error `TypeError: undefined is not an object (evaluating 'e.filter')`
NOTE the media manager modal opens, but displays no contents
NOTE the Ajax request response is 

```json
{
    "error": true,
    "code": 403,
    "message": "You don't have permission to access this. Please contact a website administrator if this is incorrect."
}
```

### Another path:

Edit an article on the frontend as an EDITOR 
In the WYSIWYG Select CMS Content - > Image, to launch model - then same Actual Result. 

### Additional comments

Similar to https://github.com/joomla/joomla-cms/issues/30005 but probably different root cause

chmst:
Should be resolved now. Could you test #32357 please?

infograf768:
Not resolved by https://github.com/joomla/joomla-cms/pull/32357
`Uncaught TypeError: e is undefined mediamanager.min.js:13:72151`

infograf768:
Would the solution concerning https://github.com/joomla/joomla-cms/issues/32259 be some change in the layout?

Something like
```
diff --git a/layouts/joomla/form/field/media.php b/layouts/joomla/form/field/media.php
index 8dd8cd1..177957a 100644
--- a/layouts/joomla/form/field/media.php
+++ b/layouts/joomla/form/field/media.php
@@ -53,4 +53,12 @@
 $attr .= !empty($onchange) ? ' onchange="' . $onchange . '"' : '';
 
+// Access check
+$user = Factory::getApplication()->getIdentity();
+
+if (!$user->authorise('core.manage', 'com_media'))
+{
+	$disabled = true;
+}
+
 switch ($preview)
 {

```

which would give
[...]

PR #32357, opened 2021-02-08
[4.0] Mitigate some b/c breaks2

dgrammatiko:
Another take of #32146 .

### Summary of Changes

Keep  the functions `endTab, addTab, endTabSet, startTabSet, renderModal, endSlide, addSlide, endAccordion, startAccordion` with the same selector as Joomla 3. No B/C break here!

For the functions `toast, tooltip, tab, scrollspy, popover, modal, dropdown, collapse, carousel, button, alert` there is no default CSS Selector and also the methods can accept any valid CSS Selector. How this works? Here's an overview:

- If a function is called without a selector Joomla **will include the appropriate script BUT WILL NOT TRY TO INITIALISE ANYTHING** (useful if you want to do something programmatically on the client side).
- If a function is called **with** a selector and no options Joomla **will include the appropriate script AND WILL TRY TO INITIALISE the given selector (if it exists in the document) with the Bootstrap Defaults for the component**
-  If a function is called **with** a selector and with provided options Joomla **will include the appropriate script AND WILL TRY TO INITIALISE the given selector (if it exists in the document) with the GIVEN options for the component**

### Testing Instructions
Try to select another component in a menu `administrator/index.php?option=com_menus&view=item&client_id=0&layout=edit&id=101` Accordion still works
Try to edit the tinyMCE options, tabs for the sets still work
Edit an article, Modals for user and images still work
Dropdown like save as (toolbar) and user menu (top bar) still work

### DIFF
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -40,7 +40,7 @@ abstract class Bootstrap
 	 *
 	 * @since   3.0
 	 */
-	public static function alert($selector = '.alert') :void
+	public static function alert($selector = '') :void
 	{
 		// Only load once
 		if (!empty(static::$loaded[__METHOD__][$selector]))
@@ -48,13 +48,19 @@ abstract class Bootstrap
 			return;
 		}
 
-		$doc           = Factory::getDocument();
-		$scriptOptions = $doc->getScriptOptions('bootstrap.alert');
-		$options       = [$selector];
+		$doc = Factory::getDocument();
 
-		if (is_array($scriptOptions))
+		if ($selector !== '')
 		{
-			$options = array_merge($scriptOptions, $options);
+			$scriptOptions = $doc->getScriptOptions('bootstrap.alert');
+			$options       = [$selector];
+
+			if (is_array($scriptOptions))
+			{
+				$options = array_merge($scriptOptions, $options);
+			}
+
+			$doc->addScriptOptions('bootstrap.alert', $options, false);
 		}
 
 		// Include the Bootstrap component
@@ -62,8 +68,6 @@ abstract class Bootstrap
 			->getDocument()
 			->getWebAssetManager()
 			->useScript('bootstrap.alert');
-
-		$doc->addScriptOptions('bootstrap.alert', $options, false);
 
 		static::$loaded[__METHOD__][$selector] = true;
 	}
@@ -79,7 +83,7 @@ abstract class Bootstrap
 	 *
 	 * @since   3.1
 	 */
-	public static function button($selector = '.button') :void
+	public static function button($selector = '') :void
 	{
 		// Only load once
 		if (!empty(static::$loaded[__METHOD__][$selector]))
@@ -88,12 +92,18 @@ abstract class Bootstrap
 		}
 
 		$doc           = Factory::getDocument();
-		$scriptOptions = $doc->getScriptOptions('bootstrap.button');
-		$options       = [$selector];
 
-		if (is_array($scriptOptions))
+		if ($selector !== '')
 		{
-			$options = array_merge($scriptOptions, $options);
+			$scriptOptions = $doc->getScriptOptions('bootstrap.button');
+			$options       = [$selector];
+
+			if (is_array($scriptOptions))
+			{
+				$options = array_merge($scriptOptions, $options);
+			}
+
+			$doc->addScriptOptions('bootstrap.button', $options, false);
 		}
 
 		// Include the Bootstrap component
@@ -101,8 +111,6 @@ abstract class Bootstrap
 			->getDocument()
 			->getWebAssetManager()
 			->useScript('bootstrap.button');
-
-		$doc->addScriptOptions('bootstrap.button', $options, false);
 
 		static::$loaded[__METHOD__][$selector] = true;
 	}
@@ -128,7 +136,7 @@ abstract class Bootstrap
 	 * - slide     string|  false  Autoplays the carousel after the user manually cycles the first item.
 	 *             boolean         If "carousel", autoplays the carousel on load.
 	 */
-	public static function carousel($selector = '.carousel', $params = []) :void
+	public static function carousel($selector = '', $params = []) :void
 	{
 		// Only load once
 		if (!empty(static::$loaded[__METHOD__][$selector]))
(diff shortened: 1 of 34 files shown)


libraries/src/HTML/Helpers/Bootstrap.php
```diff
@@ -357,27 +380,31 @@ public static function popover($selector = '[data-bs-toggle="popover"]', $option
...
+		if ($selector !== '')
+		{
+			// Setup options object
+			$opt['offset']         = isset($options['offset']) ? (int) $options['offset'] : 10;
+			$opt['method']         = isset($options['method']) ? $options['method'] : 'auto';
+			$opt['target']           = isset($options['target']) ? $options['target'] : null;
+
+			Factory::getDocument()->addScriptOptions('bootstrap.scrollspy', [$selector => (object) array_filter((array) $opt)]);
+		}
+
```
richard67, line 401:
```suggestion
```
Fix PHP CS "2 empty lines in a row".

dgrammatiko: thanks

richard67:
@dgrammatiko Are those `preg_replace` still necessary, and are they right?

dgrammatiko:
> Are those preg_replace still necessary, and are they right?

These are fail safe fuses. Lets take `public static function addTab($selector, $id, $title) :string` as an example: 
in J3 it expects an ID (without . or # ) https://github.com/joomla/joomla-cms/blob/e6cbda0560b34f56ecef9213082a9563c1daf540/libraries/cms/html/bootstrap.php#L790-L805 and then passes the selector to the layout https://github.com/joomla/joomla-cms/blob/staging/layouts/libraries/cms/html/bootstrap/addtab.php but if you pass something with `.` or `#` will break badly

In J4 we make sure that `.` or `#` will **not** be passed as a selector to the layout (it will break things if we don't check) https://github.com/joomla/joomla-cms/pull/32357/files#diff-e89418d6aa7e99964d6509d6b9ccb765ec5b9edd3482bd8ad08c29f29d2202e5R817 and https://github.com/joomla/joomla-cms/pull/32357/files#diff-74bf0060b6aa5eed515b6e29b0b4242ab49049d00b711b755f4c96488407fd53R16

Maybe overprotective but better safe than sorry

richard67:
@dgrammatiko But why is that `?` modifier in the pattern only applied to the `\.` but not the `#`? Or isn't that a modifier but a literal `?`?

dgrammatiko:
@richard67 unless I got it totally wrong `/^\.?#/` should be returning either `.` or `#` at the beginning of the string. I think I just copied that from some method in the J3 repo

richard67:
> @richard67 unless I got it totally wrong `/^\.?#/` should be returning either `.` or `#` at the beginning of the string. I think I just copied that from some method in the J3 repo

So a `.` or `#` at the beginning shall be removed? If so, then your pattern seems not to do that, at least not when I've tested with diverse online tools. A `#` was removed, but a `.` not. If it shall be that one `.` or one `#` at the beginning (and only there) shall be removed, then I would think pattern `/^[\.#]/` is the right one.

dgrammatiko:
@richard67 fair enough, I'm not a regex expert but I'll take your word on this one. Changes done in https://github.com/joomla/joomla-cms/pull/32357/commits/67e7c077ee08150881f12bf05d5ebdf765a69d57

richard67:
> @richard67 fair enough, I'm not a regex expert but I'll take your word on this one. Changes done in [67e7c07](https://github.com/joomla/joomla-cms/commit/67e7c077ee08150881f12bf05d5ebdf765a69d57)

Then I only can hope I was not wrong ;-)

laoneo:
Menu items do work, but the code `<?php echo JHtml::_('bootstrap.startTabSet', 'com-dpcalendar-form', ['active' => 'general']); ?>` and `<?php echo JHtml::_('bootstrap.addTab', 'com-dpcalendar-form', 'general, 'The label'); ?>`does not work.Getting:
image

dgrammatiko:
@laoneo should be ok after https://github.com/joomla/joomla-cms/pull/32357/commits/f33d2acb55e9ff19a44ad801120db87f606afe7d

laoneo:
Getting now 
image

laoneo:
You can test it with DPCalendar from here https://joomla.digital-peak.com/download/dpcalendar/dpcalendar-8-0-3?format=html

dgrammatiko:
@laoneo hopefully now it's fine

laoneo:
I have tested this item :white_check_mark: successfully on 8f104334918cf1c88fd0029dbf0e3a3e3dae2576<br /><br />Without the patch the following system tests have been broken:
- Selecting a menu item
- Tabs in back end forms
- User select field in back end forms

All of them do work now with the pr.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32357">issues.joomla.org/tracker/joomla-cms/32357</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on 8f104334918cf1c88fd0029dbf0e3a3e3dae2576<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32357">issues.joomla.org/tracker/joomla-cms/32357</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32357">issues.joomla.org/tracker/joomla-cms/32357</a>.</sub>

infograf768:
See https://github.com/joomla/joomla-cms/issues/32259#issuecomment-777313211

richard67:
@infograf768 So remove RTC?

dgrammatiko:
@infograf768 #32259 is something specific to the media field, nothing touched here could solve that problem

richard67:
> @infograf768 #32259 is something specific to the media field, nothing touched here could solve that problem

@dgrammatiko Do you think you can fix that with another PR?

dgrammatiko:
> Do you think you can fix that with another PR?

It should be a simple conditional for the user access level (if not allowed to edit just display the image or don't display anything at all) it's not even js related.

infograf768:
I agree it's not related to this patch. Was just responding to https://github.com/joomla/joomla-cms/issues/32259#issuecomment-777304512

dgrammatiko:
Posting something like `See #32259 (comment)` without context is very confusing, as both I and @richard67 thought that this was an unwanted side effect

richard67:
No, all fine, it's only me being paranoid sometimes when it's related to software ... I consider everything possible.

wilsonge:
@dgrammatiko fix the conflicts from the BS beta 2 merge - one test to ensure that new beta doesn't change anything and good to merge

richard67:
@laoneo @chmst Could one of you give it a quick test if it still works after Dimitris has fixed the conflicts so the PR uses BS5 Beta 2 now? Thanks in advance.

laoneo:
A first test run failed because the change event is not triggered in the user field. But can also be an issue in our end. Will test over weekend.

dgrammatiko:
@laoneo that sounds like a side effect from https://github.com/joomla/joomla-cms/pull/32379

There's a PR for this issue: https://github.com/joomla/joomla-cms/pull/32397

chmst:
Tested again successfully together with #32397.

laoneo:
I have tested this item :white_check_mark: successfully on 6f50ef3d80a9e41e053b5f75a66e42531e3b3e09<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32357">issues.joomla.org/tracker/joomla-cms/32357</a>.</sub>

laoneo:
So we have two successful tests here.

wilsonge:
Thanks guys 🙏!

Issue #32366, opened 2021-02-09
[4.0] Yes/No buttons on click not persistent


cloudfaction:
### Steps to reproduce the issue

Add a yes/no button to your module

### Expected result

When clicking on yes, it should stay green, when clicking on no it should stay red, even before save.

### Actual result

The color change is on-click, but on release it reverts. Only after save the color is changed

### System information (as much as possible)

J4 B7 / OSX / Google Chrome

### Additional comments

The problem arose when upgrading to BS5, i guess it is a BS5 issue.

richard67:
Closing as having a pull request. Please test #32367 . Thanks in advance.

Issue #32275, opened 2021-02-03
[4.0] btn-group radio doesn't work


OctavianC:
### Steps to reproduce the issue
- Go to `Content > Fields` and press `New`
- Set `Title` to `Radio Test`
- Set `Type` to `Radio (radio)`
- Under radio values, press the plus button twice to add two items: One with `Text` set to `No` and `Value` set to `0` and another one with `Text` set to `Yes` and `Value` set to `1`
- Make sure under the `Options` tab that `Field Class` is set to `btn-group`. It is by default.
- Press `Save`
- Go to `Articles` and press `New`
- Under the `Fields` tab a `Radio Test` will appear.

### Expected result
Clicking on either `No` or `Yes` will toggle between the two buttons.

### Actual result
Clicking on any value doesn't show the button being selected. The input does change, but there's no visual feedback.

chmst:
Confirmed. Also removing the class "btn-group" does not solve the issue.

brianteeman:
Both radio are being given the same class

richard67:
Closing as having a pull request. Please test #32367 . Thanks in advance.

PR #32367, opened 2021-02-09
[4.0] Update buttons.php layout to latest Bootstrap 5 markup, also added LEGEND markup

thednp:
### Summary of Changes
Updated layout file according to Bootstrap 5 new markup
* the `active` class no longer required for the `<label>` element
* the `<input>` is no longer inside the `<label>` element
* the new markup requires no `bootstrap.Button` plugin functionality anymore, everything is pure CSS

Details on the updated markup
https://getbootstrap.com/docs/5.0/components/button-group/#checkbox-and-radio-button-groups

Pull Request for Issues #32366 and #32275.

### Testing Instructions
* Open `site/templates/cassiopeia/templateDetails.xml`
* Create a new field inside the `<fieldset name="advanced">`, here a quick example

```xml
<field
  name="my-radio"
  type="Radio"
  default="0"
  class="btn-group"
  label="Radio Test">
  <option value="0">Disabled</option>
  <option value="1">Enabled</option>
</field>
```
* Login to `site/administrator` 
* go to `System` -> `Site Template Styles` -> `Cassiopeia`
* Check the new `Radio Test`, it should work perfect

The layout won't work with the new Bootstrap 5 CSS and Markup. The JavaScript for radio / checkboxes toggling has been deprecated in favor of a more traditional markup and adjusted CSS.

If instructed as above. or in any case of a `Radio` field used with `class="btn-group"` attribute, the radio input works properly in the sense that showcases the correct chosen option on any user selection.

### Notice
Due to the radio/checkboxes toggle functionality being deprecated in Bootstrap 5 (but replaced with a new CSS + Markup solution) we might need to check for any redundant JavaScript related.

None

### DIFF
--- a/layouts/joomla/form/field/radio/buttons.php
+++ b/layouts/joomla/form/field/radio/buttons.php
@@ -48,6 +48,10 @@ $isBtnYesNo  = strpos(trim($class), 'btn-group-yesno') !== false;
 $classToggle = $isBtnGroup ? ' class="btn-check"' : '';
 $btnClass    = $isBtnGroup ? 'btn btn-outline-secondary' : 'form-check';
 
+// make sure both `btn-group` and `btn-group-yesno` class attribute values
+// both use same `btn-group` class, the .btn-group-yesno class does not exist
+$class = $isBtnYesNo ? str_replace('-yesno','', $class) : $class;
+
 // Add the attributes of the fieldset in an array
 $attribs = ['class="' . trim(
 		$class . ' radio' . ($readonly || $disabled ? ' disabled' : '') . ($readonly ? ' readonly' : '')
--- a/layouts/joomla/form/field/radio/buttons.php
+++ b/layouts/joomla/form/field/radio/buttons.php
@@ -48,10 +48,6 @@ $isBtnYesNo  = strpos(trim($class), 'btn-group-yesno') !== false;
 $classToggle = $isBtnGroup ? ' class="btn-check"' : '';
 $btnClass    = $isBtnGroup ? 'btn btn-outline-secondary' : 'form-check';
 
-// make sure both `btn-group` and `btn-group-yesno` class attribute values
-// both use same `btn-group` class, the .btn-group-yesno class does not exist
-$class = $isBtnYesNo ? str_replace('-yesno', '', $class) : $class;
-
 // Add the attributes of the fieldset in an array
 $attribs = ['class="' . trim(
 		$class . ' radio' . ($readonly || $disabled ? ' disabled' : '') . ($readonly ? ' readonly' : '')
@@ -83,6 +79,9 @@ if ($dataAttribute)
 }
 ?>
 <fieldset id="<?php echo $id; ?>" >
+	<legend class="radio__legend visually-hidden">
+		<?php echo $label; ?>
+	</legend>
 	<div <?php echo implode(' ', $attribs); ?>>
 		<?php foreach ($options as $i => $option) : ?>
 			<?php
--- a/layouts/joomla/form/field/radio/buttons.php
+++ b/layouts/joomla/form/field/radio/buttons.php
@@ -50,7 +50,7 @@ $btnClass    = $isBtnGroup ? 'btn btn-outline-secondary' : 'form-check';
 
 // make sure both `btn-group` and `btn-group-yesno` class attribute values
 // both use same `btn-group` class, the .btn-group-yesno class does not exist
-$class = $isBtnYesNo ? str_replace('-yesno','', $class) : $class;
+$class = $isBtnYesNo ? str_replace('-yesno', '', $class) : $class;
 
 // Add the attributes of the fieldset in an array
 $attribs = ['class="' . trim(
--- a/layouts/joomla/form/field/radio/buttons.php
+++ b/layouts/joomla/form/field/radio/buttons.php
@@ -79,7 +79,7 @@ if ($dataAttribute)
 }
 ?>
 <fieldset id="<?php echo $id; ?>" >
-	<legend class="radio__legend visually-hidden">
+	<legend class="visually-hidden">
 		<?php echo $label; ?>
 	</legend>
 	<div <?php echo implode(' ', $attribs); ?>>
--- a/layouts/joomla/form/field/radio/buttons.php
+++ b/layouts/joomla/form/field/radio/buttons.php
@@ -104,7 +104,6 @@ if ($dataAttribute)
 				$optionClass = trim($optionClass . ' ' . $disabled);
 			}
 			$checked     = ((string) $option->value === $value) ? 'checked="checked"' : '';
-			$optionClass .= $checked ? ' active' : '';
 
 			// Initialize some JavaScript option attributes.
 			$onclick    = !empty($option->onclick) ? 'onclick="' . $option->onclick . '"' : '';
@@ -116,8 +115,8 @@ if ($dataAttribute)
 			<?php if ($required) : ?>
 				<?php $attributes[] = 'required'; ?>
 			<?php endif; ?>
+			<input<?php echo $classToggle; ?> type="radio" id="<?php echo $oid; ?>" name="<?php echo $name; ?>" value="<?php echo $ovalue; ?>" <?php echo implode(' ', $attributes); ?>>
 			<label for="<?php echo $oid; ?>" class="<?php echo trim($optionClass . ' ' . $style); ?>">
-				<input<?php echo $classToggle; ?> type="radio" id="<?php echo $oid; ?>" name="<?php echo $name; ?>" value="<?php echo $ovalue; ?>" <?php echo implode(' ', $attributes); ?>>
 				<?php echo $option->text; ?>
 			</label>
 		<?php endforeach; ?>


Quy:
I have tested this item :white_check_mark: successfully on c53b0c66ac0fd40ed1c43015e53a4c67d7bbbe62<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

OctavianC:
I have tested this item :white_check_mark: successfully on c53b0c66ac0fd40ed1c43015e53a4c67d7bbbe62<br /><br />It fixes #32366 and #32275<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

infograf768:
Agree this solves the original issue but here is the difference between using bootstrap button and switcher:

To compare I modified the Cassiopea Sticky header field to
```
				<field
					name="stickyHeader"
					type="radio"
					label="TPL_CASSIOPEIA_STICKY_LABEL"
					class="btn-group"
					default="0"
					filter="integer"
					>
				</field>
```

Screen Shot 2021-02-10 at 09 58 05

Screen Shot 2021-02-10 at 10 30 35


There is no background color difference between Yes and No.
Also, is this totally a11y compliant?

thednp:
@infograf768 regarding the `Sticky Header` option you have

here's how it should work
yesno

In regards to a11y, I don't think there are any issues when compared to previous version.

thednp:
@infograf768 I will provide a fix for the layout to work with `btn-group-yesno` asap, a small bug I found.

infograf768:
Adding indeed `btn-group-yesno` is much better

Screen Shot 2021-02-10 at 10 39 48

infograf768:
@thednp 
You should also consider adding a legend for a11y, as we have for the switcher.

Screen Shot 2021-02-10 at 10 34 01

thednp:
Roger that, stay close for the updated layout.

thednp:
Alright, I've changed in my dev so that this layout will do the following:
* if you use `class="btn-group"` for any `type="Radio"` field, it should work as expected, which means the container of the input options will use `class="btn-group"`, no change here
* however, if you use `class="btn-group-yesno"` for `type="Radio"`, it will also use `.btn-group` CSS class, because `.btn-group-yesno"` CSS class does not exist
* so you don't have to use something like `class="btn-group btn-group-yesno"` for `type="Radio"` to achieve the same thing as I've showcased in the image attached above

Now about a11y, @infograf768 I don't see why a visible `<label>` need to have a legend for our `buttons` layout here.

thednp:
@infograf768 the switcher layout indeed is a special case, it needs to tell screen readers what's the option about since there is no visible `<label>`, only a hidden `<legend>` which I'm sure readers check.

You think we need to add `<legend>` for `type="Radio"` fields when using this very `buttons` layout?

richard67:
> You think we need to add `<legend>` for `type="Radio"` fields when using this very `buttons` layout?

Maybe @chmst can have a look on that, too.

infograf768:
If our class remains `class="btn-group btn-group-yesno"
Your last changes creates a double btn-group class

Screen Shot 2021-02-10 at 11 11 32

Therefore the class would have to be `class="btn-group-yesno"` which is NOT bc as J3 uses `class="btn-group btn-group-yesno"

Concerning the legend, we should do the same as for switcher which has a visible label

Screen Shot 2021-02-10 at 11 13 09

richard67:
I'm also not happy with the last change, that looks like a hack to me, that str_replace.

chmst:
A legend is necessary (visually hidden).

thednp:
@chmst I will add `<legend>` then, no problem.

@richard67 and @infograf768 you don't have to use `class="btn-group btn-group-yesno"`, just `class="btn-group-yesno"`, but if you REALLY think that's the way to do it, no problem, change incoming.

infograf768:
> but if you REALLY think that's the way to do it, no problem, change incoming.

It would be B/C with J3. As we have already so much non B/C in j4, let's not add one. ;)

richard67:
> @richard67 and @infograf768 you don't have to use `class="btn-group btn-group-yesno"`, just `class="btn-group-yesno"`, but if you REALLY think that's the way to do it, no problem, change incoming.

@thednp Your str_replace would not be safe for the case someone uses a custom class like e.g. `my-fancy-class-yesno` for some reason. You should have done a regex replace to make sure you only replace what you want to replace ;-) But for B/C we should revert that completely, like @infograf768 said.

thednp:
@chmst @infograf768 and @richard67 done.

See commit note.

infograf768:
The `switcher__legend` class exists in css, but not the `radio__legend`

thednp:
Yes, the `.radio__legend` has just been created. But I was referring to the fact that the CSS for `.switcher__legend` is redundant, unless it serves a purpose for J3? I'm not sure.

infograf768:
imho, `switcher_legend` css as well as `radio-legend` are totally useless as the only thing `switcher_legend` does is setting the font-size for a visually-hidden legend.
But let's @drmenzelit comment on that.

ceford:
When I click the button it changes from disabled to enabled. But the two hidden checkboxes don't change checked status. I find this whole thing very confusing!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

thednp:
@ceford it's the same [here](https://getbootstrap.com/docs/5.0/components/button-group/#checkbox-and-radio-button-groups), but it's OK, you can check the `checked` status via JavaScript, here's how:

* open Inspector (assume you use a Chrome based browser)
* right-click on an option, then select `Inspect`, make sure the `<input>` element is selected
* in the browser console, type in `$0.checked`, it should return the current state

It's how the browser engine works now, it won't update the `checked` attribute, but internally it always knows the current state, as it renders your `.btn` elements properly according to the real `checked` state.

infograf768:
Hmm, I now get
Screen Shot 2021-02-10 at 12 29 04

richard67:
> Hmm, I now get ...

@infograf768 I guess you mean that space between the "No" and "Yes" buttons, right?

thednp:
@infograf768 @richard67 that's what my previous fix was about. my implementation wasn't perfect as Rich pointed out, but would have solved that. So yea, you need to use `class="btn-group btn-group-yesno"` to have it working without spacing between.

infograf768:
Forget it
Typo in the xml
I had `class="btn-goup btn-group-yesno"`

instead of
`class="btn-group btn-group-yesno"`

all ok

Screen Shot 2021-02-10 at 12 48 44

sorry

richard67:
I have tested this item :white_check_mark: successfully on 13fe5a0db0890bcb406847493d8fb59d65c65e22<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

OctavianC:
I have tested this item :white_check_mark: successfully on 13fe5a0db0890bcb406847493d8fb59d65c65e22<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

brianteeman:
Not sure why this is RTC as there are outstanding issues. There is no point in adding a class to the legend that doesn't exist in the css. It serves no purpose as the legend is never visually displayed. Or do I miss something?

richard67:
> Not sure why this is RTC as there are outstanding issues. There is no point in adding a class to the legend that doesn't exist in the css. It serves no purpose as the legend is never visually displayed. Or do I miss something?

We have such classes according to BEM elsewhere, too, as once added by @ciar4n , as far as I remember.

brianteeman:
on something that is hidden? That makes zero sense.

richard67:
@thednp Could you remove the `radio__legend ` from the class of the legend? Thanks in advance.

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

thednp:
> 
> 
> @thednp Could you remove the `radio__legend ` from the class of the legend? Thanks in advance.

Done.

richard67:
I have tested this item :white_check_mark: successfully on 9aabd48da16217f6ae3737fd076f5985f12a35ef<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

infograf768:
I have tested this item :white_check_mark: successfully on 9aabd48da16217f6ae3737fd076f5985f12a35ef<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

infograf768:
rtc
let’s get rid of switcher__legend code and css in another pr<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32367">issues.joomla.org/tracker/joomla-cms/32367</a>.</sub>

thednp:
@infograf768 I would do it rn if I knew where's the source.

richard67:
@thednp `find ./ -type f -exec grep -Hn "switcher__legend" {} \;` in a Linux shell or `findstr /s /c:"switcher__legend" *.*` in a Windows Command window will help you to find out when executed in the Joomla root folder.

richard67:
But for SCSS sources in the build folder it needs a git clone of course, that's not included in a normal installation package.

thednp:
Added https://github.com/joomla/joomla-cms/pull/32377 in regards to #issuecomment-776805109

thednp:
Boy I wish we had the same involvement for #30816

infograf768:
Tks

Issue #21354, opened 2018-08-02
[4.0] joomla-field-media element should trigger 'change' event when media is selected


khanhlh:
After a media is selected, the value of input is updated, the joomla-field-media element should trigger 'change' event.

ghost:
@laoneo shouldn't this kind of Issues reported at https://github.com/joomla-projects/media-manager-improvement?

laoneo:
Its fine here

dgrammatiko:
Please test #32379

PR #32379, opened 2021-02-10
[4.0] Add missing change events

dgrammatiko:
Pull Request for Issue #21354 .

### Summary of Changes
Add missing change events for the fields `media`, `user` and `simple color`

### Testing Instructions
in your browser console (assuming Chrome) Selecet the HTML element for the field (eg `joomla-field-simple-color`) and in the console copy paste `$0.addEventListener('change', (ev) => console.log(ev));` Then select a color. Observe the console. Repeat for the other 2 fields

No event triggered

Change Event triggered on each change


wilsonge:
Merged on review

Issue #32412, opened 2021-02-13
[4.0] Pre-Update Check


brianteeman:
### Steps to reproduce the issue
Set the update server as custom with this url https://update.joomla.org/core/nightlies/next_major_list.xml and then check for updates and go to the Pre-U[date check tab and scroll down.

### Expected result
Compatibility check is performed and the list is updated

### Actual result
Nothing happens 
![image](https://user-images.githubusercontent.com/1296369/107864586-940baa00-6e55-11eb-8fcc-42e14c00d065.png)

there is the following js error in the console

```
default.min.js?f53a5302ecc0843c7a67a80d17580c2a:1 Uncaught TypeError: e.data is not a function
    at Object.a.checkCompatibility (default.min.js?f53a5302ecc0843c7a67a80d17580c2a:1)
    at default.min.js?f53a5302ecc0843c7a67a80d17580c2a:1
    at Array.forEach (<anonymous>)
    at HTMLDocument.a.run (default.min.js?f53a5302ecc0843c7a67a80d17580c2a:1)
```

### Additional comments
The path to the ajax-loader is also wrong but I fixed that in its own pr

dgrammatiko:
@brianteeman I have been assigned the task to fix the Joomla update so hopefully, at some point, I will do the PR (my problem is that the markup is totally inaccessible and I'm trying to at least make it somehow accessible. Actually, there are way more problems with the code quality of Joomla update but let's assume that the problem is only accessibility)

brianteeman:
Yes I know the a11y is fubar as I have commented before

dgrammatiko:
Unfortunately for me, there are more problems:
- The UX is unintuitive: if there are incompatible extensions the user shouldn't be able to continue unless they agree to take the risk, right now this thing is just some sparse infos
- The actual Joomla update (although for the front end it's just 4-5 ajax calls) it's rather convoluted and messy code when it comes to json
- The script that is doing the encryption is a pet project (with a clear note to **NOT** be used in production) that shouldn't have been included in a CMS that claims to be the 2-3 most popular (there's some history here: https://github.com/joomla/joomla-cms/pull/20452 )

brianteeman:
At this point in time I am strongly of an opinion its too late to do a complete rewrite

joomdonation:
@brianteeman Look like the issue was fixed elsewhere. I could not see the issue on today nightly build anymore. Could you please check and confirm?

brianteeman:
confirmed

PR #32413, opened 2021-02-13
[4.0] pre-update checker ajax-loader

brianteeman:
This PR corrects the path to the ajax-loader.gif that is used in the pre-update check.

The check itself is broken this PR just fixes the 404. See  #32412 for instructions how to make this view visible


wilsonge:
Thanks!

brianteeman:
Thanks

Issue #32487, opened 2021-02-21
[4.0] New fields layout utility classes don't work


drmenzelit:
The new field layout utility classes from PR #32422 don't work when inside the form-grid (which defines display:grid) a control-group div is present.

### Steps to reproduce the issue
Edit the templateDetails.xml file from Cassiopeia and add e.g. `class="form-select span-1-inline stack"` to the fields. Go to Template Styles and check the layout of the fields.

### Expected result
The fields should display horizontal like described in the PR.

### Actual result
The fields are still one below the other. The classes are there but the grid definition doesn't work here:
"grid-column has no effect on this element because it is not a grid element.
Try adding display:grid or display:inline-grid to the parent element. "

![grafik](https://user-images.githubusercontent.com/9153168/108639375-862cd900-7494-11eb-83d5-07661b64e3c6.png)

@ciar4n can you have a look here? Thanks

ciar4n:
It looks likes I shouldn't have reverted removing the extra attribute (https://github.com/joomla/joomla-cms/pull/32422#issuecomment-779050471). Because the class attribute does not add the class to the containing div, it is not suitable for these utilities. 

I'll create a PR to fix. Originally I call the attribute `display=" ... "`.. I assume this name is ok?

dgrammatiko:
>  Originally I call the attribute display

@ciar4n  will `parentclass` be more appropriate/specific?

ciar4n:
PR created and updated to use `parentclass`.... #32488

PR #32488, opened 2021-02-21
[4.0] Add display field attibute

ciar4n:
Pull Request for Issue  #32487 .

### Summary of Changes
Because the 'class'  XML attribute adds the class to the field itself rather than the outside container, it is not suitable for the classes added to  #32422. This PR adds a `parentclass` attribute which adds a class to the outside `control-group` div.

### Testing Instructions
Ensure https://github.com/joomla/joomla-cms/pull/32422 works as described (updated to match)


drmenzelit:
I have tested this item :white_check_mark: successfully on 7cef8cf9d9e44858f18addad012a4f567dac892c<br /><br />Tested in templateDetails.xml from Cassiopeia.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32488">issues.joomla.org/tracker/joomla-cms/32488</a>.</sub>

layouts/joomla/form/renderfield.php
```diff
@@ -35,6 +35,11 @@
 $id              = $name . '-desc';
 $hide            = empty($options['hiddenLabel']) ? '' : ' visually-hidden';
 $hideDescription = empty($options['hiddenDescription']) ? false : $options['hiddenDescription'];
+
+if($parentclass) {
+	$class .= ' ' . $parentclass;
+}
```
Quy:
```suggestion
if ($parentclass)
{
	$class .= ' ' . $parentclass;
}
```

libraries/src/Form/Field/CalendarField.php
```diff
@@ -81,6 +81,13 @@ class CalendarField extends FormField
 	 */
 	protected $layout = 'joomla.form.field.calendar';
 
+	/**
+	 * The parent class of the field
+	 *
+	 * @var  string
+	 */
```
Quy, line 89: Please add `@since  __DEPLOY_VERSION__`.

ciar4n: Thank you @Quy 👍

Quy: This has to be done with the other instances. 😞 Thanks.

ciar4n: @Quy .. all done!

chmst:
I have tested in com_content, article.xml and there it works fine.

grafik

But in Options (config.xml) there is no effect. 

grafik

ciar4n:
@chmst Thank you for testing. 

Strangely I am unable to replicate this issue. Adding to ` ..\administrator\components\com_config\forms\application.xml` for example `parentclass="span-2"` works without issue.

image

ciar4n:
@chmst My apologies. I see the issue now in component tabs as you said. I'll submit a fix.

ciar4n:
@chmst Should be good now. 👍

chmst:
I have tested this item :white_check_mark: successfully on 780fdc49853d2ea885bdb8e3834a848b3a777de8<br /><br />Tested wit com_content: article edit, com_content config.xml, com_users: user edit
and an own J4 component; 
Documentation will be required. It is a great improvement. Thank you. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32488">issues.joomla.org/tracker/joomla-cms/32488</a>.</sub>

drmenzelit:
I have tested this item :white_check_mark: successfully on 780fdc49853d2ea885bdb8e3834a848b3a777de8<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32488">issues.joomla.org/tracker/joomla-cms/32488</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32488">issues.joomla.org/tracker/joomla-cms/32488</a>.</sub>

korenevskiy:
1. Using the **`display`** attribute is confusing, because after the **`display`** you want to write **`block`**. Yes, in general, and **`parentclass`** is also not clear about which parent we are talking about. I suggest using the **`controlclass`** attribute, we know that in the full sense of the control is just **`<div class= "control-group"/>`**
2. I suggest making the layout multi-column. But with the observance of several rules. the **`note`** field and the **`spacer`** field will be full-width.
Controls located between the **`note`** and **`spacer`** fields, only between them will be distributed in 2-3 columns. This will allow you not to create a garbage can from the fields. And of course the ability to use the **`controlclass`** attribute.
But having the ability to override the default classes and set your own width in the number of columns.
.
In any case, now in Joomla Nightly 1 column mode, by analogy J3,  looks better. 1 column is clearer and the most important thing is available. I propose to center 1 column in the middle.
.
How do I output a control-table without a Label?
img-2021-03-07-23-58-39

brianteeman:
`hiddenLabel="true"`

infograf768:
@ciar4n 

just a note: the code proposed is
```
if ($parentclass)
{
	$class .= ' ' . $parentclass;
}
```

As it already adds a space, we do not need the space in your example in https://github.com/joomla/joomla-cms/pull/32422
```
<field
	name="article_layout"
	type="componentlayout"
	label="JFIELD_ALT_LAYOUT_LABEL"
	class="form-select"
	useglobal="true"
	extension="com_content"
	view="article"
        parentclass=" span-4 stack"
/>
```

also, I do not understand why, in this PR you add stack to the article layout in  `administrator/components/com_content/forms/article.xml` as it does not have any effect.

https://github.com/joomla/joomla-cms/pull/32488/files#diff-48dcb2f2108f7b07952c9c407fd23cab28677e3d2fd437030b2f04a043282320R280-R286

ciar4n:
@infograf768 My bad, I didn't mean to commit https://github.com/joomla/joomla-cms/pull/32488/files#diff-48dcb2f2108f7b07952c9c407fd23cab28677e3d2fd437030b2f04a043282320R280-R286 (reverted with https://github.com/joomla/joomla-cms/pull/32488/commits/cb121a57994836a02a7355b6c2dc1d632a9a6d80). The PR can remain in RTC as this class was ineffective.

And yes, no space is required in the example on #32422. I have edited this accordingly. Thanks for mentioning it!

ciar4n:
@korenevskiy I guess the logic of `parentclass` is that this class is been added to the parent div of the field that the rule is creating. Unlike the `class` attribute, this does not add it to the `control` class div.

korenevskiy:
> @korenevskiy I guess the logic of `parentclass` is that this class is been added to the parent div of the field that the rule is creating. Unlike the `class` attribute, this does not add it to the `control` class div.

The control is `< div class= "control-group" >`. But the field is `< input >`. The field is located inside the control. The field has a parent that is a control. But the control also has a parent. It is intuitive for a layout designer who is unfamiliar with Joomla, but knows Bootstrap. ask which parent is being referred to when the class is specified, the parents of the control, or the parents of the field? But by specifying controlclass, it is clear that this is the class of the tag `< div class="control group" >` .

let the class be added directly to `<div class= "control-group">`
this will allow you to position the `label` at the same time.
Or can use another `fieldclass `class

korenevskiy:
by the way, after all these tricks, the next step is this. 1. Creating an XML element to connect the CSS file to the form. 2. Create an XML tag to include Styles in the form.. Maybe this will be the right decision. This will be bad form for developers, but supporting such a feature is the right solution. I've seen modules where CSS is plugged in inside a custom field. This is worse than connecting directly to XML. but Joomla does not know how to do this.

You can consider a new attribute for `<field>`, such as `<field type="text" css="my_module.css"/>`

korenevskiy:
> @korenevskiy I guess the logic of `parentclass` is that this class is been added to the parent div of the field that the rule is creating. Unlike the `class` attribute, this does not add it to the `control` class div.

img-2021-03-09-13-35-32
it is not clear from the name of the attribute which parent we are talking about
Not intuitive and unclear.
.
But **`<field controlclass="span-4 stack">`** -> **`<div class= "control-group span-4 stack">`**
And now control will have a width across the entire GRID.

ciar4n:
@korenevskiy As this PR is already set to RTC, I'll be unlikely to edit it further. Feel free to create a PR against this code once/if this gets merged 👍

Issue #32490, opened 2021-02-22
[4.0][3.x] (Business) vulnerability


Ruud68:
### Steps to reproduce the issue
This vulnerability applies to both the 3.x and 4.0 versions: I will describe the steps for 4.0

1. install J4 beta 7
2. install blog sample data
3. in global article options set (integration tab) remove IDs from URLs to 'No'
4. goto URL: https://[yourdomain]/blog/4-why-our-business-sucks-big-time

### Expected result
The article 'Why our business sucks big time' doesn't exist and should return a 404 page not found error

### Actual result
the page for article with ID 4 is shown instead: no 404 presented

### Additional comments

Due to the 'feature' where only an ID is needed to display a page, it is possible (and already actively abused) to add fake, but working, links to the Google search results. This is very simple: just add a page to your own website with these bogus links to your competitor's website. Google will index your page, finds the bogus / abusive links and because they resolve okay will add them to the search index.

Now when somebody searches on your competitors company, chances are that these links will display in the search results and (worst of all) when clicking them in the search results will go to your competitors website. This because the URL has a high ranking factor for Google.

This way it is very easy to target a competitor via Google search results.

As a site owner it is almost impossible to remove these bogus links to your website from the search results as that requires adding them (no wildcards allowed) in Google Search Tools.

Last week I had a customer hit by this and actively loosing business because of it!

brianteeman:
In Joomla 3 if you are using the modern router then this doesnt happen
In Joomla 4 the modern router is enabled by default and this doesn't happen

Ruud68:
@brianteeman correct, not everybody will use the modern router as that will require redirects. So the issue is there and will be there for as long as using the 'legacy' router is an option.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32490">issues.joomla.org/tracker/joomla-cms/32490</a>.</sub>

brianteeman:
The fix is to not use the legacy router

Ruud68:
The **workaround** is to set up redirect for every category / article combination to the URL without the numbers and then to set the modern router. Unfortunately that is not an option for every site.

brianteeman:
So you have two working solutions but you dont want to use either of them :(

Ruud68:
[...]

PR #32500, opened 2021-02-23
[4.0] legacy router strict / loose mode fix

Ruud68:
Pull Request for Issue #32490 .

### Summary of Changes
The legacy router used in 3.9 and used in 4.0 allows for Search Engine URL forgeries (see #32490) for the impact of this (business) vulnerability.
This PR introduces a toggle with which the legacy router can be toggles from Loose Mode (current dfault mode) to Strict mode.
Loose Mode will parse an URL based on Article ID only, this maks it possible to add any (discriminating) text to the URL and 'feed' that to e.g. Google and will show up in the search results for your site
Strict Mode requires the ID and Alias match. If no match, a 404 will be produced (same is in the Modern Router)

This PR is a POC and when correct can be ported to other core components with the Legacy Router 'flaw'

### Testing Instructions
Install this PR
In article option tab integration switch Remove IDs from URLs to 'No' > enabling the Legacy Router
Leave the setting Strict Mode in its default value (Loose)

[A] Test:
1. open on the front-end [yourdomain]/index.php/blog/5-your-modules
2. open on the front-end [yourdomain]/index.php/blog/5-dont-do-any-business-with-me-because-i-suck

Set config parameter Strict Mode to 'Strict'
[B]Repeat test above

[A]: 1 > Your Modules page in displayed
[A]: 2 > Your Modules page is displayed (note: the non-existing / fake url will be indexed by Google because it resolves okay and as such will show in your sites search results)

[B]: 1> Your Modules page is displayed
[B]: 2 > a 404 is thrown wich is the correct behavior as this page doesn't exist and should NEVER resolve okay!

Maybe an alert should be shown (in the face of the site maintainer) when switching strict mode to Loose that will inform him/her of the risks involved


brianteeman:
Why is this PR for J4? I would have expected it to be first committed to J3

Ruud68:
@brianteeman because when I do a J3 everybody asks why not J4... #lol
This can easily be 'backported' to 3 (and if accepted should) but since all dev and test attention is currently on 4 I did it for 4 first.
because 4 also has this issue as you can see :)

components/com_content/src/Service/Router.php
```diff
@@ -80,6 +88,7 @@ public function __construct(SiteApplication $app, AbstractMenu $menu, CategoryFa
 
 		$params = ComponentHelper::getParams('com_content');
 		$this->noIDs = (bool) $params->get('sef_ids');
+		$this->legacyStrict = (bool) $params->get('legacy_strict');
 		$categories = new RouterViewConfiguration('categories');
```
HLeithner: code style (all = should be at the same row)

components/com_content/src/Service/Router.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
HLeithner:
```suggestion
```
Why did you add this row? I think we don't have a empty row normally after <?php

components/com_content/src/Service/Router.php
```diff
@@ -274,6 +283,27 @@ public function getArticleId($segment, $query)
 			return (int) $this->db->loadResult();
 		}
 
+		if ($this->legacyStrict)
+		{
+			$id    = (int) $segment;
+			$alias = ltrim(str_replace($id, '', $segment), '-');
```
HLeithner:
```suggestion
			[$id, $alias] = explode('-', $segment, 2);
```

components/com_content/src/Service/Router.php
```diff
@@ -274,6 +283,27 @@ public function getArticleId($segment, $query)
...
+
+			$dbquery = $this->db->getQuery(true);
+			$dbquery->select($this->db->quoteName('id'))
+				->from($this->db->quoteName('#__content'))
+				->where(
+					[
+						$this->db->quoteName('alias') . ' = :alias',
+						$this->db->quoteName('id') . ' = :id',
```
HLeithner:
```suggestion
						$this->db->quoteName('id') . ' = :id',
						$this->db->quoteName('alias') . ' = :alias',
```

Maybe not so important but it's strange if it different to the bind statement

components/com_content/src/Service/Router.php
```diff
@@ -274,6 +283,27 @@ public function getArticleId($segment, $query)
...
+					]
+				)
+				->bind(':id', $id)
```
HLeithner:
```suggestion
				->bind(':id', $id, ParameterType::INTEGER)
```

You need to add `use Joomla\Database\ParameterType` at the top of the file

HLeithner (commented):
I personally see it as a feature to have the id in the url because if you change the alias it link still exists. 

Redirecting to the right url would be make sense.

ceford:
With this url: http://localhost/j4test/index.php/en/blog/5-dont-do-any-business-with-me-because-i-suck in strict mode I get a 404 page but I also get a call stack starting JROOT/libraries/src/Router/Router.php:153 

Also after applying the patch, first page load gave an error page for the good url - but not reproducable.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32500">issues.joomla.org/tracker/joomla-cms/32500</a>.</sub>

components/com_content/src/Service/Router.php
Ruud68: I just added the line, the alignment isn't there in the original file.

Ruud68:
> I personally see it as a feature to have the id in the url because if you change the alias it link still exists.
> 
> Redirecting to the right url would be make sense.

@HLeithner , thanks for your review. as for the comment above: That would be indeed a feature and not a 'fix' what this PR does.
If you do a redirect and not like now (without this PR) just show a page with the same article id, that would be a b/c as the a redirect is not the same as a 404 / 200

Ruud68:
> With this url: http://localhost/j4test/index.php/en/blog/5-dont-do-any-business-with-me-because-i-suck in strict mode I get a 404 page but I also get a call stack starting JROOT/libraries/src/Router/Router.php:153
> 
> Also after applying the patch, first page load gave an error page for the good url - but not reproducable.
> This comment was created with the [J!Tracker Application](https://github.com/joomla/jissues) at [issues.joomla.org/tracker/joomla-cms/32500](https://issues.joomla.org/tracker/joomla-cms/32500).

Hi @ceford  thanks for testing, I cannot reproduce what you reported, also when looking at the code in line 153 (where the exception was thrown to show the 404 page) I do not see how that could produce a callstack on screen

components/com_content/src/Service/Router.php
HLeithner: Doesn't matter if you touch it please fix it if you can ;-)

HLeithner:
> @HLeithner , thanks for your review. as for the comment above: That would be indeed a feature and not a 'fix' what this PR does.
> If you do a redirect and not like now (without this PR) just show a page with the same article id, that would be a b/c as the a redirect is not the same as a 404 / 200

My suggest is to make it a 3 way field, do nothing, make redirect, send 404.

ceford:
I did a new install (clone) a couple of days ago and I have just merged the latest updates. I did install Multi-lingual sample data and have now used the correct article id: http://localhost/j4test/index.php/en/sample-layouts/category-list/9-dont-do-any-business-with-me-because-i-suck but the result is the same. I get a 404 Error page and then the stack trace. I have Error reporting set to Maximum. Maybe that is normal - I just don't remember it that way.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32500">issues.joomla.org/tracker/joomla-cms/32500</a>.</sub>

Ruud68:
> 
> My suggest is to make it a 3 way field, do nothing, make redirect, send 404.

I will not do redirect in this PR. This PR needs to be also back-ported to 3.9 / 3.10 and that is not possible with new functionality.
I suggest making a separate PR for new functionality, also because this requires changes on a complete different level then the B/C 'safe' changes here

components/com_content/src/Service/Router.php
Ruud68: done

administrator/components/com_content/config.xml
```diff
@@ -1048,6 +1048,19 @@
 				<option value="0">JNO</option>
 				<option value="1">JYES</option>
 			</field>
+			<field
+				name="legacy_strict"
+				type="radio"
+				layout="joomla.form.field.radio.switcher"
+				default="0"
+				label="JGLOBAL_SEF_STRICTMODE_LABEL"
+				description="JGLOBAL_SEF_STRICTMODE_DESC"
```
Quy:
```suggestion
				name="legacy_strict"
				type="radio"
				label="JGLOBAL_SEF_STRICTMODE_LABEL"
				description="JGLOBAL_SEF_STRICTMODE_DESC"
				layout="joomla.form.field.radio.switcher"
				default="0"
```

administrator/components/com_content/config.xml
```diff
@@ -1048,6 +1048,19 @@
...
+				filter="integer"
+				showon="sef_ids:0"
+				>
+				<option value="0">Loose</option>
```
Quy: Please use language strings.

components/com_content/src/Service/Router.php
```diff
@@ -38,6 +38,13 @@ class Router extends RouterView
 	 */
 	protected $noIDs = false;
 
+	/**
+	 * Flag to set Legacy Router in Strict or Loose mode
+	 *
+	 * @var    boolean
```
Quy, line 44:
```suggestion
	 * @var    boolean
	 * @since  __DEPLOY_VERSION__
```

administrator/components/com_content/config.xml
Ruud68: ooops :)

components/com_content/src/Service/Router.php
Ruud68: good one

Ruud68:
@HLeithner and @Quy done changes as requested, did you test?
Also in last commit some css files where changed, no idea where these come from and how they ended up in my repo, must have done something wrong. no clue as how to fix that other then start all over again (which I wont) So if somebody could help a hand, that would be appreciated

administrator/components/com_content/config.xml
```diff
@@ -1040,14 +1040,27 @@
...
 				>
 				<option value="0">JNO</option>
 				<option value="1">JYES</option>
 			</field>
+			<field
+				name="legacy_strict"
+				type="radio"
+				layout="joomla.form.field.radio.switcher"
+				label="JGLOBAL_SEF_STRICTMODE_LABEL"
+				description="JGLOBAL_SEF_STRICTMODE_DESC"
```
Quy, line 1056:
```suggestion
				label="JGLOBAL_SEF_STRICTMODE_LABEL"
				description="JGLOBAL_SEF_STRICTMODE_DESC"
				layout="joomla.form.field.radio.switcher"
```

Ruud68:
closing due to there being no  interest in this solution, reopening original issue as the issue is not solved

brianteeman:
Sorry but thats just pathetic closing this. Re-opening the issue when you already have a proposed fix is not very sensible.

No one tested this as you didnt bother to respond to the request for updates.

PR #32505, opened 2021-02-24
Fix en-GB.com_associations.ini

dryabov:
Fix argnum syntax (`%1s` means a string of minimal length of 1 character instead of first argument string `%1$s`)


brianteeman:
@infograf768 This PR does look correct but its been this way for 4+ years since the first commit

infograf768:
Looking now

infograf768:
It does work OK as is with Persian and obviously with this patch.

Better correct though.

EDIT: the reason we never had any problem is that it does not consider the figure, i.e. `%1s` or `%2s` are both considered as `%s`
As we are not in a phrase, but as forced order for 2 words (one is a bit more complex in Persian) separated by a `>` the figure maybe useless in this specific case.

infograf768:
I have tested this item :white_check_mark: successfully on 328f2445f2b70ae9306c5da955a2ad3d3ca2ba8c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32505">issues.joomla.org/tracker/joomla-cms/32505</a>.</sub>

infograf768:
Setting RTC as this is an obvious patch<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32505">issues.joomla.org/tracker/joomla-cms/32505</a>.</sub>

brianteeman:
@infograf768 I did spot at least one other instance while investigating

infograf768:
> I did spot at least one other instance while investigating

I found none more in core. Can you ellaborate?

brianteeman:
Sorry it was a new string in J4 
https://github.com/joomla/joomla-cms/blob/d07424a2e404a8ab6f294daa39a2c55727692331/administrator/language/en-GB/com_installer.ini#L253

infograf768:
That one would need to be patched indeed.

HLeithner:
Thanks

infograf768:
@dryabov 
Could you contact me on glip?
Ref: your message to traduction arobase joomla.fr

Issue #32456, opened 2021-02-18
[4.0] Disable Components


brianteeman:
### Steps to reproduce the issue
In Extensions Manager you can disable a component.
If you disable a component that is displayed in the sidebar components menu then the link is removed from there and from the components dashboard.
You can disable Contacts to see

If you disable a component that is not displayed in the sidebar components menu but is instead displayed on the system dashboard then no changes are made. The link is still displayed in the dashboard and if you click on it you will obviously get an error.

You can disable CSP to confirm this

### Expected result
A disabled component is removed from the admin wherever it is displayed

### Actual result
Only components in the sidebar are removed when disabled

chmst:
Closing as we have a PR

PR #32539, opened 2021-02-27
fix(mod_submenu): hide disabled components in mod_submenu

sawmurai:
Pull Request for Issue #32456 .

### Summary of Changes

The changes introduced hide menu items for disabled components from the cpanel.

### Testing Instructions

1. Disable the CSP extension

The menu item "Content Security Policy" is still available in the cpanel.

The menu item "Content Security Policy" is not available in the cpanel anymore.

### DIFF
--- a/administrator/modules/mod_submenu/src/Menu/Menu.php
+++ b/administrator/modules/mod_submenu/src/Menu/Menu.php
@@ -59,13 +59,14 @@ abstract class Menu
 	/**
 	 * Filter and perform other preparatory tasks for loaded menu items based on access rights and module configurations for display
 	 *
-	 * @param   MenuItem  $parent  A menu item to process
+	 * @param   MenuItem  $parent              A menu item to process
+	 * @param   array     $disabledExtensions  Names of disabled extensions, that shall be removed from the result
 	 *
 	 * @return  void
 	 *
 	 * @since   4.0.0
 	 */
-	public static function preprocess($parent, array $disabledExtensions)
+	public static function preprocess($parent, array $disabledExtensions = [])
 	{
 		$app      = Factory::getApplication();
 		$user     = $app->getIdentity();
--- a/administrator/modules/mod_submenu/src/Menu/Menu.php
+++ b/administrator/modules/mod_submenu/src/Menu/Menu.php
@@ -41,19 +41,21 @@ abstract class Menu
 		$query = $db->getQuery(true);
 
 		// Prepare the query.
-		$query->select($db->quoteName('e.name'))
+		$query->select($db->quoteName('e.element'))
 			->from($db->quoteName('#__extensions', 'e'))
+			->where($db->quoteName('e.type') . ' = ' . $db->quote('component'))
+			->where($db->quoteName('e.client_id') . ' = 1')
 			->where($db->quoteName('e.enabled') . ' = 0');
 		$iterator = $db->setQuery($query)->getIterator();
 
-		$names = [];
+		$extensions = [];
 
 		foreach ($iterator as $item)
 		{
-			$names[] = $item->name;
+			$extensions[] = $item->element;
 		}
 
-		return $names;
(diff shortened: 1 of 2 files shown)


administrator/modules/mod_submenu/src/Menu/Menu.php
```diff
@@ -35,7 +65,7 @@ abstract class Menu
 	 *
 	 * @since   4.0.0
 	 */
-	public static function preprocess($parent)
+	public static function preprocess($parent, array $disabledExtensions)
```
bembelimen: Probably you should set an empty array as default, so it's not braking if some extension using this call...

bembelimen:
@SharkyKZ could you please elaborate why you gave a thumb down? Probably there is room for improvement in this PR based on you feedback. Thanks.

administrator/modules/mod_submenu/src/Menu/Menu.php
sawmurai: Good idea, thank you!

richard67:
I don't know if this is the right approach or not, but it contains a few mistakes:

1. The extension name is not unique in the extensions table, so there may be several extensions of different types having the same name value. What has to be unique is the combination of element, folder and client_id for extensions of type "plugin", and element and client_id for extensions of other types. But the database query here doesn't contain any restrictions ("WHERE ...") for the extension type and client_id:

https://github.com/joomla/joomla-cms/blob/47b2cff4737b0cdd0c23239b97cacaf4ae4cc637/administrator/modules/mod_submenu/src/Menu/Menu.php#L44-L46

As far as I understand the purpose of this PR, only extensions of type "component" should be fetched with that query, and they should be restricted also to the right client_id here (1 for admin).

2. The property "element" is checked to be in the array, which contains the "name" values and not the "element" values, as you could see by the query mentioned above. See here the check:

https://github.com/joomla/joomla-cms/blob/47b2cff4737b0cdd0c23239b97cacaf4ae4cc637/administrator/modules/mod_submenu/src/Menu/Menu.php#L85

For lucky circumstances, the "name" and "element" values are the same for (core) components, but that might not always be granted, so it can happen we compare apples and pears here. Therefore at least from a formal point of view the above check is wrong.

3. Beside the above 2 points my knowledge about the submenu is too little to say if the approach chosen here is right or wrong. Hoping for more detailed feedback e.g. from @SharkyKZ .

richard67:
P.S. to my above comment's items 1 and 2: For the reasons stated there, I would expect the query to be:
```
		$query->select($db->quoteName('e.element'))
			->from($db->quoteName('#__extensions', 'e'))
			->where($db->quoteName('e.type') . ' = ' . $db->quote('component'))
			->where($db->quoteName('e.client_id') . ' = 1')
			->where($db->quoteName('e.enabled') . ' = 0');
```
And later below it should be:
```
$names[] = $item->element;
```

sawmurai:
@richard67 Thank you so much for the comments! I updated the PR. Let's see what @SharkyKZ's feedback is. 

This would be my first contribution to joomla despite having used it for the past couple of years, so I would not be surprised if I misunderstood something and my approach is totally wrong. 

The idea behind my approach is basically: 
Since the menu is rendered from the preset-xml files (in the error case) my only information are the `element` and `type` of the menu item, so I know that its a component of "type" com_csp. To link that to the state of the component (enabled, disabled) I have to query the database beforehand and pass that information to the precheck function, which already appears to be hiding menu items under certain conditions.

alikon:
humm....
too much code imho see #32555 too...

sawmurai:
Agreed! Did not know about the ComponentHelper

richard67:
@sawmurai If you agree that PR #32555 shall replace this PR here: Can we close your PR here? Of course you can do that yourself, too.

sawmurai:
Yep, I will close it :)

PR #32541, opened 2021-02-27
Make fontawesome+icomoon mapping available as Joomla asset

drmenzelit:
First step to make fontawesome available as web asset.
At the moment Cassiopeia delivers an own fontawesome.css file that is the result of importing fontawesome files from the media/vendor folder combined with the icomoon B/C mapping file (stored in media/system). 
Since the fontawesome file can be also useful as a web asset in other parts of Joomla (e.g. frontend editing) or for 3rd party extensions / templates developers, it is better to have it in the media folder.

### Summary of Changes
Removed fontawesome.scss from Cassiopeia.
Changed the joomla.asset.json in Cassiopeia to load fontawesome from media folder.
Added a joomla-fontawesome.scss file in media/system.

### Testing Instructions
Apply the patch, run npm run build:css
Load the website using Cassiopeia as template.

Fontawesome icons are present (e.g. search symbol, chevron in readmore button). Fontawesome is loaded as  templates/cassiopeia/css/vendor/fontawesome-free/fontawesome.min.css

Fontawesome icons are still present. Fontawesome is loaded as media/system/css/joomla-fontawesome.min.css


ChristineWk:
I have tested this item :white_check_mark: successfully on 1b4894dce4e2039cbdad86d84854d4bf64afefc9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32541">issues.joomla.org/tracker/joomla-cms/32541</a>.</sub>

Fedik:
It does no need, there already fontawesome asset, in `media/vendor/joomla.asset.json`
And a fix for font-face files in #31364

Fedik:
hm I have missed "icomoon" changes.
if someone need "icomoon" then ignore my previous comment :wink:

brianteeman:
I would expect the fontawesome css to be the same just in a different location etc. But

image

drmenzelit:
> It does no need, there already fontawesome asset, in `media/vendor/joomla.asset.json`
> And a fix for font-face files in #31364

I didn't see your PR before I did mine... And I was not aware of fontawesome as an asset in the `media/vendor/joomla.asset.json` file :-( 
As I understand the icomoon file contains a mapping old icomoon => new fontawesome. I think it is important to include.

drmenzelit:
@brianteeman the code in your screenshot is included in Autum but not in Cassiopeia
Because Autum loads also an own fontawesome file

Fedik:
I think here some confusion, 
Cassiopeia already have `templates/cassiopeia/scss/vendor/fontawesome-free/fontawesome.scss`  that holds `fontawesome` and `icomoon`.

And `media/vendor/joomla.asset.json` already holds `fontawesome` Asset.

The final file `templates/cassiopeia/css/vendor/fontawesome-free/fontawesome.css`  overrides file from `media/vendor/fontawesome-free/fontawesome.css`, and so `fontawesome` Asset uses `fontawesome.css` provided by Cassiopeia.

All is working good already, you do not need to fix anything :wink:

drmenzelit:
I'm developing a J4 template without Bootstrap and with a parameter to switch fontawesome on / off. It works good. But if I want to use frontend editing, I need fontawesome (or I have to re-define the icons). I don't want to deliver an own version of fontawesome with the template (and Cassiopiea shouldn't do either), I want to load what Joomla already has in core. And I want to be able to load fontawesome only in component.php for the frontend editing. If the version of fontawesome in Joomla core doesn't include the font-face and the icons mapping, it will not work.

brianteeman:
@Fedik The problem being solved is that a template might not use fontawesome but when you open an edit form on the front end then you will need to load fontawesome. So by making it available as an asset it can be added into the edit forms

When I write fontawesome I mean the fontawesome WITH iconmoon conversion file

Fedik:
> So by making it available as an asset it can be added into the edit forms

As I already noted before: the `fontawesome` already provided by `media/vendor/joomla.asset.json`. 
`media/vendor/joomla.asset.json` always loaded, and available for all extensions.

> I'm developing a J4 template without Bootstrap and with a parameter to switch fontawesome on / off.

if you doing it in your personal template then all you need is something like:
```php
if ($needToEnableFont) 
{
  $wa->useStyle('fontawesome');
}
```

If you want this in Cassiopiea, then you have to remove these dependencies from Cassiopiea `joomla.asset.json`:

https://github.com/joomla/joomla-cms/blob/9ca4e5672c82839f0e7933452cd4ea0085f33a4c/templates/cassiopeia/joomla.asset.json#L12-L15

https://github.com/joomla/joomla-cms/blob/9ca4e5672c82839f0e7933452cd4ea0085f33a4c/templates/cassiopeia/joomla.asset.json#L21-L24

And then use in PHP:
```php
if ($needToEnableFont) 
{
  $wa->useStyle('fontawesome');
}
```

Fedik:
> If the version of fontawesome in Joomla core doesn't include the font-face and the icons mapping, it will not work.

The `fontawesome` in Joomla core  contains only original fontawesome style and icons, without mapping to Joomla icons.
These mapping done by the template itself (eg `templates/cassiopeia/scss/vendor/fontawesome-free/fontawesome.scss`).

If people want to have Joomla "icomoon" Independently from the template, then we probably need to re-compile fontawesome in media/vendor. But this is totally different topic.

Fedik:
hm, sorry, I think I am Turtle :smile: 

I finally understood what you tries to do, it a bit in confusing way, but it is okay.

The PR is okay for its purpose.

Sorry again :wink:

templates/cassiopeia/joomla.asset.json
```diff
@@ -96,6 +96,11 @@
       "name": "searchtools",
       "type": "style",
       "uri": "system/searchtools/searchtools.min.css"
+    },
+    {
+      "name": "fontawesome",
+      "type": "style",
+      "uri": "media/system/css/joomla-fontawesome.min.css"
```
dgrammatiko: Please use the relative path here so that the asset could be automatically overridden in the template

```
"uri": "system/joomla-fontawesome.min.css"
```

dgrammatiko:
Also, another comment here: ALL the fields should use SVG icons either as HTML or as css background so that the front end WILL BE font awesome free. Icon fonts are really deprecated and should be considered HARMFUL as their killing performance metrics but also have serious caveats for visually impaired people.

templates/cassiopeia/joomla.asset.json
drmenzelit: Done

templates/cassiopeia/joomla.asset.json
```diff
@@ -100,7 +100,7 @@
     {
       "name": "fontawesome",
       "type": "style",
-      "uri": "media/system/css/joomla-fontawesome.min.css"
+      "uri": "system/css/joomla-fontawesome.min.css"
```
dgrammatiko: Almost, you need to remove also the `/css` part

dgrammatiko (changes requested):

templates/cassiopeia/joomla.asset.json
drmenzelit: I see... thank you for the help

brianteeman:
> also have serious caveats for visually impaired people.

They are no better and no worse than svg. Its all down to how they are used.

dgrammatiko:
> They are no better and no worse than svg. Its all down to how they are used.

I suspect if everyone else dropped already font icons there are obvious drawbacks. Basically abusing Fonts for enhancing style was once an option (no SVG support in IE, etc) but those reasons have been eliminated since IE 11 which basically is like a decade ago. Time to move to something more in line with the actual web devs practices. My 2c anyways...

brianteeman:
> there are obvious drawbacks

I am only commenting on the repeated statement that they are not bad for **accessibility** or that SVG are good for **accessibility**

dgrammatiko:
> I am only commenting on the repeated statement that they are not bad for accessibility or that SVG are good for accessibility

SVG's (depending also on how you're using them) could be better one at least one case: if the network fails to deliver either the css or the actual font then the icons will not be displayed. Inlining (obviously cleverly, without blowing up the document size) eliminates this case. If you have the document you already have the icons 😉. Anyways, there are more cases that fonts fall short for this particular usage but accessibility wise you can end up messing either. I never suggested that SVGs are by default accessible and icons are not but maybe the path to making something accessible could be easier with SVGs (purely because you will end up calling some PHP function that will do the right thing, instead of manually inserting HTML tags/attributes)

Mrs-Jaydot:
May I ask a question here?
I'm just a user, not a coder, but I don't know where else to ask.
I've made a test site with Beta8, and was happy to find FontAwesome installed locally.
However: I cannot use the brand icons (facebook, linkedin). Miscellaneous icons (bicycle, camera) display just fine.
Is there a reason for this?
Thanks.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32541">issues.joomla.org/tracker/joomla-cms/32541</a>.</sub>

ghost:
@Mrs-Jaydot Questions on https://forum.joomla.org/

Mrs-Jaydot:
@sandramay0905 Thanks, I know, but no-one there has any idea. The only answer I got was "I don't think Joomla uses Font Awesome...".

drmenzelit:
> 
> 
> @sandramay0905 Thanks, I know, but no-one there has any idea. The only answer I got was "I don't think Joomla uses Font Awesome...".

Are you using Cassiopeia or an own template?

Mrs-Jaydot:
@drmenzelit Cassiopeia.

brianteeman:
There are multiple font files for fontawesome. Not all are free etc

drmenzelit:
> 
> 
> @drmenzelit Cassiopeia.

`<i class="fab fa-facebook"></i>` should work (fab is for brands)

dgrammatiko:
brands need a `fab` class, eg `<i class="fab fa-facebook-f"></i>` 
`fab` === brands
`far` === regurlar
`fas` === solid

In short there are 3 fonts and if you load all of them your site will be blank for some time for first-time visitors...

Mrs-Jaydot:
@brianteeman I know, but FA4 is free for download, and it contains brand icons.
@drmenzelit Thank you!! Works! I didn't know about the faB.
@Fedik Thank you for responding and explaining, I didn't know it worked that way.
Many thanks to all of you.
I will now move back to forum.org, where I belong :).

RickR2H:
Do we have to test this patch?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32541">issues.joomla.org/tracker/joomla-cms/32541</a>.</sub>

richard67:
> Do we have to test this patch?

@RickR2H In general yes, every PR needs 2 tests, but is currently has a conflict in file `templates/cassiopeia/scss/vendor/fontawesome-free/fontawesome.scss`, so better wait until that has been resolved.

@drmenzelit Is the PR still relevant? If yes, could you solve the conflict? Let me know if you need help with that. Thanks in advance.

drmenzelit:
The discussion was long, I got lost at some point... and it seems my approach is not clear for everyone, so I think it is up to @wilsonge to decide what to do with it

richard67:
@ChristineWk Could you test again with the new packages for this PR? Thanks in advance.

build/media_source/system/scss/joomla-fontawesome.scss
```diff
@@ -0,0 +1,14 @@
...
+$fa-font-path: "../../../media/vendor/fontawesome-free/webfonts" !default;
+$fa-font-display: block !default;
+
+// Font Awesome 5 Free
+@import "../../../../media/vendor/fontawesome-free/scss/fontawesome";
+@import "../../../../media/vendor/fontawesome-free/scss/regular";
+@import "../../../../media/vendor/fontawesome-free/scss/solid";
+
+// Brands must be imported last
+@import "../../../../media/vendor/fontawesome-free/scss/brands";
```
richard67, line 11: Are you sure that these import paths are right? In line 2 above it is one time "../" less to the media folder than it is here.

richard67: And line 2 is right I think.

richard67: Why don't you just use "../../vendor/" instead of "../../../media/vendor"?

richard67: Hmm I just see I might be wrong .. so I assume this here is right and hope there will be careful tests.

brianteeman:
I have tested this item :white_check_mark: successfully on 5b1a08de59d76d51cd2f345aedded15365599474<br /><br />works as described<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32541">issues.joomla.org/tracker/joomla-cms/32541</a>.</sub>

ChristineWk:
I have tested this item :white_check_mark: successfully on 5b1a08de59d76d51cd2f345aedded15365599474<br /><br />Hv this (if it's OK):
/*!
 * Font Awesome Free 5.15.3 by @fontawesome - https://fontawesome.com
 * License - https://fontawesome.com/license/free (Icons: CC BY 4.0, Fonts: SIL OFL 1.1, Code: MIT License)
 */<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32541">issues.joomla.org/tracker/joomla-cms/32541</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32541">issues.joomla.org/tracker/joomla-cms/32541</a>.</sub>

N6REJ:
we created a auto .fa to .iconmoon conversion system long ago.  idky it needs to be done in another way yet again.

richard67:
> we created a auto .fa to .iconmoon conversion system long ago. idky it needs to be done in another way yet again.

@N6REJ This PR here doesn't change anything on the mapping, it just moves scss to another place and makes it available as template-independent asset.

brianteeman:
Its not being done in a different way its just putting the files in the correct place

richard67:
> Its not being done in a different way its just putting the files in the correct place

That's what I've said, too.

brianteeman:
github cache sucks sometimes

N6REJ:
Thanks guys, I was about to say, danggggg after all that work LOL

wilsonge:
Thanks!

PR #32543, opened 2021-02-27
Directory Listing

ErvisTusha:
CWE-548: Exposure of Information Through Directory Listing

https://cwe.mitre.org/data/definitions/548.html

### DIFF
--- a/administrator/components/com_actionlogs/controllers/index.html
+++ b/administrator/components/com_actionlogs/controllers/index.html
@@ -1,0 +1,1 @@
+<!DOCTYPE html><title></title>
--- a/administrator/components/com_actionlogs/helpers/index.html
+++ b/administrator/components/com_actionlogs/helpers/index.html
@@ -1,0 +1,1 @@
+<!DOCTYPE html><title></title>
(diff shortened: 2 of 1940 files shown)


brianteeman:
Thank you for your first (public) pull request to joomla. Good to see that you are still checking joomla. I will let the maintainers explain why the files were removed

Bakual:
Imho the index.html files were removed because this is actually a server confguration issue. The index files just were a bandaid around a misconfigured server. Nowadays servers should have directory listing disabled by default and we should not be needed to ship >1000 empty index.html files.

HLeithner:
As Thomas said, I don't believe that they come back. Anyway thanks for your contribution. 

Releated PR https://github.com/joomla/joomla-cms/pull/3788

ErvisTusha:
intitle:"index of /" inurl:components/com_banners/

brianteeman:
Interestingly the first of those that I came across had an index.html and yet still displayed the folder. I guess you really can't fix bad hosting

https://www.heliciculture.net/components/com_banners/models/
image

PR #32380, opened 2021-02-11
Fix batch menu items

joomdonation:
Pull Request for Issue #32375.

### Summary of Changes
This small PR fixes #32375.

### Testing Instructions
1. Access to Menus -> Main Menu
2. Select some menu items, then press Batch button in the toolbar
3. Before patch: On then popup, you only see unpublished menu item
4. After patch: Both published and unpublished menu items are being displayed

Issue #32375, opened 2021-02-10
Batch processing menu-items unable to select target menu-item


Pinkeltje:
### Steps to reproduce the issue
Select a menu and all its items then try to batch move them to another menu and menu-item

### Expected result
Being able to select target menu-item

### Actual result

Only possible to select menu (no items)

### System information (as much as possible)
Joomla 3.9.24

### Additional comments
Works in 3.9.22 (not tested with 3.9.23)
**3.9.22**
![3_9_22](https://user-images.githubusercontent.com/8941310/107508220-f9834080-6ba0-11eb-9122-c92bcc89cf8d.jpg)
**3.9.24**
![3-9-24](https://user-images.githubusercontent.com/8941310/107508223-fa1bd700-6ba0-11eb-9826-5c725b9defed.jpg)

brianteeman:
I can confirm the problem. For anyone trying to replicate this you will find that only menu items that are unpublished are available to select (so if you have no unpublished items you get the display reported)

joomdonation:
Thanks @brianteeman. That reminds me a similar issue which I fixed before. I made PR https://github.com/joomla/joomla-cms/pull/32380 to fix this small issue.

brianteeman:
It would be a good idea to find the pr that broke this

joomdonation:
The issue causes by this PR https://github.com/joomla/joomla-cms/pull/31536

brianteeman:
A perfect example of why we need automated tests

alikon:
please test #32380

PR #32558, opened 2021-03-02
[4.0] Fix Batch Menu Items in 4.0 - Update default_batch_body.php

particthistle:
Pull Request for Issue #32375, but for Joomla 4 instance.

### Summary of Changes
Issue resolved in "Fix batch menu items" #32380 was applied to staging, but also affects 4.0-dev, as mentioned by @infograf768 in [this comment](https://github.com/joomla/joomla-cms/pull/32380#issuecomment-777255090).

Changing the published option from $published to $this->state->get('filter.published') resolves it.

Problem identified during testing in Joomla 4 Beta 8 nightly build. Problem didn't appear to occur in Beta 6, but does occur in Beta 7 onwards.

### Testing Instructions
- Access to Menus -> Main Menu
- Select some menu items, then press Batch button in the toolbar
- The dropdown for the Copy/Move option to reassign the parent for the batch selection of menu items should show a list of published menu items, but currently does not without the patch applied.

Before patch: On the popup you only see the Menu names, and no options to assign as parent.
image

After patch: Published menu items are displayed to assign as parent items
image

### DIFF
--- a/administrator/components/com_menus/tmpl/items/default_batch_body.php
+++ b/administrator/components/com_menus/tmpl/items/default_batch_body.php
@@ -56,7 +56,7 @@ if ($clientId == 1)
 						<option value=""><?php echo Text::_('JLIB_HTML_BATCH_NO_CATEGORY'); ?></option>
 						<?php
 						$opts     = array(
-							'published' => $published,
+							'published' => $this->state->get('filter.published'),
 							'checkacl'  => (int) $this->state->get('menutypeid'),
 							'clientid'  => (int) $clientId,
 						);


joomdonation:
@particthistle The issue will be fixed when the changes in staging is merged into 4.0-dev, so this PR could be closed (no need for two separate PRs, one for staging and one for 4.0-dev).

infograf768:
Exact. Closing. Thanks.

particthistle:
Apologies - comes with having avoided testing J3 and only concentrating on J4. First time I've encountered an issue where it flows through both code bases.

As it's 3 weeks since #32380 was merged and the issue was still in J4, I mistakenly thought it had been missed vs needing to wait for the fix.

PR #32568, opened 2021-03-02
Remove reference to joomlacode.org

PhilETaylor:
Remove reference to joomlacode.org - as Joomla code is dead (soon) https://developer.joomla.org/news/840-bye-bye-joomlacode.html


PhilETaylor:
https://developer.joomla.org/joomlacode-archive/issue-26026.html

Issue #30874, opened 2020-10-02
[4] "Your YubiKey is already linked to your user account" twice


PhilETaylor:
### Steps to reproduce the issue

Enable Yubikey on an account in admin console of Joomla 4
Edit your account ->Two Factor Authentication Tab

Note "Your YubiKey is already linked to your user account" is said twice.

### Expected result

"Your YubiKey is already linked to your user account" once

### Actual result

"Your YubiKey is already linked to your user account" twice

<img width="1574" alt="Screenshot 2020-10-02 at 18 09 11" src="https://user-images.githubusercontent.com/400092/94950508-a7d3f680-04da-11eb-9d55-f416bb453e87.png">

brianteeman:
please test #30879

PR #30784, opened 2020-09-26
Images loading=lazy, part 2n

dgrammatiko:
This PR solves a lot of bugs and brings feature parity to J3, keep reading

#### Bug fixes:
- Changes the a-tag to button in the media modals
- Restores width and height params in the media manager object (in 2 places)
- Separates the image selection script from the media script (if you had an Utd-button without a media field the button was foobar)

#### New features (parity with J3, basically missing in J4):
- Ability to enter ALT text for any image (used in the editors)
- Ability to set if an image will be lazy loaded in the content area (eg inside an editor)
- Ability to set ALT text for images inserted through drag and drop in tinyMCE
- Ability to set if an image will be lazy loaded for images inserted through drag and drop in tinyMCE
- New helper `HTMLHelper::cleanImageURL`, explainer below

#### Lazyloading
The concept is simple: it just works!
- For any image inserted in an editor (tinyMCE, Codemirror, none, or any 3rd PT) the user **has the ability to select if the image will be lazy loaded**
- For images used though the media field there's a helper to... wait for it, help you. By default the URL that is stored in the database has 2 extra URL parameters `joomla_image_width` and `joomla_image_height` that represent the width and height of the image. Passing the URL in the helper you'll get back an object with the clean url (without the 2 extra params mentioned before, and the attributes (array of width and height). So since we're already in an overridable file (layout) anyone can decide if the position of the image should or not be lazy loaded. Reading the fulltext layout is more explanatory than reading some text out of context: https://github.com/joomla/joomla-cms/blob/3427a9ca0bb265f72cb8da9a61c508a0a5859afa/layouts/joomla/content/full_image.php

### Testing Instructions
Download the package from the bottom part of the PR in Github or do it through git command

Test the tinyMCE dnd
Make sure tinyMCE is your editor
Try to insert an image in an article with dnd, enter some alt text and leave the lazy loaded checked
Repeat but unselect the lasyloaded and leave the alt empty
Toggle the editor and check the produced HTML

Test the tinyMCE media button
Try to insert an image in an article with the media button, enter some alt text and leave the lazy loaded checked
Repeat but unselect the lasyloaded and leave the alt empty
Toggle the editor and check the produced HTML

Repeat the same for Codemirror and none

Custom Fields
setup a media custom field
setup a imagelist custom field
make sure the custom fields work

check the html, should have attributes: loading=lazy, width, height for all the images.

Media field
Insert an full image
Check the html, should have attributes: loading=lazy, width, height

Insert an intro image
[...]

PR #32587, opened 2021-03-04
Removal of an orphaned "alt"

ahotzler:
Pull Request for Issue # .

### Summary of Changes

Removal of an orphaned "alt"

### Testing Instructions

Füge ein "Full Image" zu einem Artikel hinzu und schau Dir den HTML Code oder benutze einen Syntax-Checker

Error: Duplicate attribute alt.
'alt="" alt			 itemprop="'

No Error while HTML Syntax Checking

None

### DIFF
--- a/layouts/joomla/content/full_image.php
+++ b/layouts/joomla/content/full_image.php
@@ -34,7 +34,7 @@ if ((isset($img->attributes['width']) && (int) $img->attributes['width'] > 0)
 ?>
 <figure class="<?php echo htmlspecialchars($imgclass, ENT_COMPAT, 'UTF-8'); ?> item-image">
 	<img src="<?php echo htmlspecialchars($img->url, ENT_COMPAT, 'UTF-8'); ?>"
-			 <?php echo $alt; ?> alt
+			 <?php echo $alt; ?>
 			 itemprop="image"
 			<?php echo $extraAttr; ?>
 	/>


particthistle:
I have tested this item :white_check_mark: successfully on ea3c2a045bbefb1e6622b85a84bb159091aa039a<br /><br />Works correctly. Stray alt introduced in #30784 removed.

Question though for @dgrammatiko regarding the changes in #30784 that led to this.

- https://www.w3.org/WAI/tutorials/images/decorative/ and other sources shows that all images should have an alt tag, and if it's just a decorative image, the alt tag should be alt="" for improved A11y
- so rather than having to specify specifically "No Description" for the image, why doesn't it just default to alt="" if there's no alt text supplied? One less thing the user then has to do when adding an image.

So much looked to go on in #30784 this might have been something overlooked.

Another PR would be needed as a result to modify line 26 to change the generation of $alt and further change the implementation of $images->image_fulltext_alt_empty which then becomes redundant as the fallback will always be alt="" if there's no alt text.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32587">issues.joomla.org/tracker/joomla-cms/32587</a>.</sub>

brianteeman:
> so rather than having to specify specifically "No Description" for the image, why doesn't it just default to alt="" if there's no alt text supplied? One less thing the user then has to do when adding an image.

That is absolutely the opposite of what should be done. The author must make a conscious decision if an image is decorative or not and if it is not then they must enter a suitable description.

You dont make your website accessible by simply ignoring (and therefore hiding) images.

Previously Joomla did exactly what you suggest and defaulted to alt="" 

The problem with that is that its not a conscious decision and perhaps more importantly will pass any accessibility checker so you will never be prompted to fix your site and to make it accessible for all.

Responsible web design is making your content available for everyone and not, as in your suggestion, just to those with vision.

Please read https://magazine.joomla.org/all-issues/october-2020/an-image-is-worth-a-thousand-words

brianteeman:
In summary. If the image is not decorative and you "forget" to add an alt description then no accessibility checker will ever be able to tell you that you "forgot". Hence the bugfix

ahotzler:
Can someone tell me what I need to do so that no height/width information is output? I have not discovered a way to do this in the backend. I then created an override, which was the reason why I looked into the code in the first place.

dgrammatiko:
> regarding the changes in #30784 that led to this.

Actually @brianteeman initiated those changes (if you cared reading the responses on the PR you linked you would have seen it) and I think he's explained it in details also here in the above comments


> I then created an override

Override is always the right answer in joomla (adding options is always the wrong one!!!)

ahotzler:
Does this really mean that height/width is always output by default and this behavior cannot be turned off via backend? I find the size assignment with these attributes directly in the source code under aspects of the image scaling in certain constelations very obstructive, at least if I can't turn it off.

dgrammatiko:
> Does this really mean that height/width is always output by default and this behavior cannot be turned off via backend

YES, the reason is that browsers actually require the exact width and height of any image! It can be disabled by overriding OR when choosing an image deselect the lazyload checkbox (but don't do that, will affect the performance of your site).

> I find the size assignment with these attributes directly in the source code under aspects of the image scaling in certain constelations

Use CSS something like this 

```css
.someClass {
  max-width: 100%;
  height: auto;
}

ahotzler:
> OR when choosing an image deselect the lazyload checkbox (but don't do that, will affect the performance of your site).

I search and search, but can't find it. Can you give me a hint where to find the lazyload **checkbox?**

dgrammatiko:
> I search and search, but can't find it. Can you give me a hint where to find the lazyload checkbox?

Select a new image and check all the options before closing the Modal, it's there

ahotzler:
No, there is no such option!

ahotzler:
And please dont forget: We are talking about "Full Image", The image selection dialog behaves differently here than when inserting an image normally

dgrammatiko:
Screenshot 2021-03-05 at 10 00 54

dgrammatiko:
For the full image and intro image ALL the changes can be done with an override

ahotzler:
Than the comments in the code doesnt make sense

dgrammatiko:
> Than the comments in the code doesnt make sense


??? Context?

ahotzler:
There is a comment:

"// Set lazyloading only for images which have width and height attributes"

It suggests that you as a user have a choice, which you don't have with "Full-Image

dgrammatiko:
> It suggests that you as a user have a choice, which you don't have with "Full-Image

As a site visitor you get whatever was decided by the author of the site.
As an author you have the option to Override the default behaviour and do the 2010 version of an image, eg without width, height and an empty alt. If you're asking for more switches, sorry I'm the wrong guy for this, bye

particthistle:
Thanks for updating my POV on it @dgrammatiko and @brianteeman

#30874 had so much going on the discussion you're referring to I think might have been in amongst clusters of comments that GitHub closed up while I was trying to find it.

I'll go add a note to add concepts of lazyloading and the new alt text features to https://docs.joomla.org/Adding_an_image_to_an_article.

Meanwhile, another test is needed to allow this to be merged and take the excess alt out of the rendered code.

ahotzler:
Can the PR creator, i.e. me, also test or does that violate guidelines?

alikon:
the PR creator is supposed to have already tested it
it needs 2 other testers

joomdonation:
I have tested this item :white_check_mark: successfully on ea3c2a045bbefb1e6622b85a84bb159091aa039a<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32587">issues.joomla.org/tracker/joomla-cms/32587</a>.</sub>

infograf768:
In fact we already have 2 testers.



RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32587">issues.joomla.org/tracker/joomla-cms/32587</a>.</sub>

infograf768:
Tks.

PR #32600, opened 2021-03-06
[4.0] CSP

brianteeman:
Many many months ago we reached the point where no new features can be added to Joomla 4.0. We have to **stop** at some point and release Joomla 4.0 or it really will never be released and many many people's hard work will be a waste of time.

But at the same time there must be a point where we say that code that is not ready for release is removed from 4.0.

14 months ago we had that decision about com_workflows and the decision was to give a period of a few extra months or it would be removed possibly for a 4.1 release. At the time discussing that delay was contentious as it would delay the release. We are now another year on and still no release so we can't  do that again.

This PR removes com_csp.

In its current form it is not usable as already reported and it is not documented as already reported except for some basic this button does this help. Considering the first commit of this component was almost **THREE** years ago it's fair to say that there is no evidence of any documentation any time soon.

I have NOT removed the httpheaders plugin as that might still be usable but might need a few tweaks after the component is removed as they appear to share some configuration options. But I wouldnt see anything wrong with removing that as well for the same reason and its already available as an extension and of course it is currently undocumented.

**_This component can be re-introduced for Joomla 4.1 if the bugs are resolved and documentation provided. Possibly it will need to go through the RFC process but that's not for me to say_**


ceford:
Just noting an interest in this topic. I tried using CSP and was interested in documentation - as time permits. I have no opinion on deferment to 4.1 or later.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32600">issues.joomla.org/tracker/joomla-cms/32600</a>.</sub>

brianteeman:
As I see that the Production Department has made a decision to start rejecting any new feature for 4.0 and to push them to 4.1 I trust that this will be the same here

bembelimen:
Hello Brian,
you have a valid point here. I'm working for a few week already to improve the CSP, but there is a lot to do. But as Joomla! 4 still needs some days, I added the release blocker tag and moved it to draft, so your PR is not merged accidently but if I will not be ready with the code, yours will not be forgotten and used.

brianteeman:
Thanks for the update. I hope this  PR is not needed

Fedik:
I like the idea behind CSP, and would like to see it in, and it a couple years old already.
I do not see much critical issues that would force us to remove it.
Well, I hope they will be fixed.

dgrammatiko:
I will have to object here, CSP is the only tool we have against Spectre. There are even recommendations from the W3: https://w3c.github.io/webappsec-post-spectre-webdev/#tldr that in most cases need some header to be set and setting headers without a way to monitor what is rejected (which if I'm not wrong is the task of this component) is kinda blind operation.

BTW exploits are already publically available: https://twitter.com/mikewest/status/1370394397267869699 https://leaky.page  

Tagging @SniperSister @zero-24 esp for the leaky page and the w3 recommendations

My 2c

PhilETaylor:
> I will have to object here, CSP is the only tool we have against Spectre. 

CSP can be implemented without Joomla's Joomla 4 com_csp component  :-) and decently architected websites would do that. 

In fact, as Joomla 4 com_csp will only output headers when Joomla's Plugins are triggered (When Joomla is running, mainly through its own endpoints like index.php) its not 100% coverage of all files in a website, and there are better ways to set response headers per site, including .htaccess and in web server configurations, rather than in PHP.

com_csp/httpheaders plugins just makes it easier for mass-distributed Joomla admins to implement a CSP. Its not the only way, and is far from perfect. 

A CSP implemented with com_csp/httpheaders within Joomla would NOT output headers when JS/CSS/PHP files are accessed by a direct URL for example, outside of a Joomla page.

dgrammatiko:
> A CSP implemented with com_csp/httpheaders within Joomla would NOT output headers when JS/CSS/PHP files are accessed by a direct URL for example, outside of a Joomla page.

You're totally right, this should be done at the server level (htaccess, nginx conf, etc) but at least this covers the index.php generated content (basically any .HTML document which should be good enough for most of the cases where this would be applied, the document is the entry point for any static asset js, css, image, etc). Could it be improved? Yes but that's not the point of my above comment. I'm just trying to raise awareness **why the CSP is essential** for any 2021 site and to back my objection on the removal of the component...

That's all, nothing personal here (I get @brianteeman 's point as well that there are bugs, no docs, etc) but this is a kinda fundamental piece for security. 

> Remember Joomla 3 doesn't have all this com_csp/httpheaders stuff, so not having it in Joomla 4 doesn't make it any "worse" 

Well that's not an argument, if J3 is bad let's make sure J4 is bad as well (that's what I'm reading)

> Reporting can also be done to third party systems like https://report-uri.com/ until com_csp is ready

I'm fine with that as well, just provide some GUI module/plugin/whatever with sensible defaults and then remove com_csp

PhilETaylor:
> Well that's not an argument, if J3 is bad let's make sure J4 is bad as well (that's what I'm reading)

No my point was that, its currently not Joomla's responsibility with Joomla 3 sites, and with com_csp removed, it would continue not to be Joomla's responsibility and would continue to put the burden on the site developer.... so it would be no different than what we have now - except with HOPE that in the future Joomla might add this new feature 

(I actually think it would be great to see it at a cPanel/WHM/Plesk level, but I digress) 

> I'm just trying to raise awareness why the CSP is essential for any 2021 site

Totally agree 

but 99% of the web still doest have a CSP as they are HARD to get right and HARD to understand and GEEKY and when implemented wrong, break sites functionality. 

(*99% is a made up figure)

brianteeman:
This PR only exists because the current code does not work. If it worked (and I hope @bembelimen can make it work) then great. Unfortunately the current state of the code makes it unusable.

rdeutz:
closing this in favor of #32893

brianteeman:
@rdeutz until #32893 is accepted and merged this PR is still valid

brianteeman:
as #32893 has been closed please re-open this release blocker

Fedik:
@brianteeman @zero-24 I have an offer :)

I will fix  `mod_custom` and `com_csp`  conflict.
Someone should fix com_csp to collect correct "url" and remove "status" column,
and we leave CSP component in "detect mode" for 4.0.

Rest of improvement can be done in future.

brianteeman:
if only it was those items

Fedik:
well, I have made fix anyway #32980 
I think it does not hurt to have :)

brianteeman:
> I think it does not hurt to have :)

agreed and thanks

Fedik:
> if only it was those items

hm, I see that "detect" mode actually different from "custom", and it really does not very play together.

Additionally to my previous suggestion:

- Drop "Mode" selection, it should stay "Custom" always.
- "report-uri" should be added when "Report-Only" enabled (How it was made now does not make sense).

With this changes User will be able to configure own rules, and watch a report when "Report-Only" set On. And adjust rules manually based on the report, therefore we do not need "auto" mode (with `compileAutomaticCspHeaderRules()`) for now, I guess.

I tried look around the issues list, anything else I have missed?

Fedik:
@brianteeman please have a look #32983

but do not close current one

brianteeman:
@fedik sorry I'm not spending any more time on this.

Fedik:
> sorry I'm not spending any more time on this.

I do not ask for full test, enough to enable the CSP and make sure site not exploded, with default config (with empty rules)

brianteeman:
sorry.

zero-24:
FYI a more complet PR to remove com_csp has been made here: https://github.com/joomla/joomla-cms/pull/33550 I will close this PR here for that reason.

Thanks

PR #32605, opened 2021-03-07
[3.x] Make database schema check ignore display widths for all integer types on MySQL or MariaDB databases

richard67:
Pull Request for https://github.com/joomla/joomla-cms/pull/28501#issuecomment-792267368 .

### Summary of Changes

This pull request (PR) extends the fix from #28501 for the database schema check to integer data types other than `int` and `tinyint`.

It only relevant for MySQL and MariaDB databases, ie not PostgreSQL.

As stated in PR #28501 and issue #32542 , the display width modifier has no impact on value range or storage size of a database column of one of the integer data types.

While MySQL deprecated these display widths beginning with version 8.0.17 and doesn't include them anymore in the type returned by a `SHOW COLUMNS` statement, MariaDB still supports and uses display widths modifiers for integer types so it still behaves like MySQL <= 5.7.

With this PR I took the chance for refactoring the parts which have been modified before with #28501 for `int` and `tinyint` types, so that code is more readable and also more precise with checking the types and so doesn't hide anymore typos in update SQL scripts.

### Testing Instructions

**_Requirement:_** Have an installation of current staging or latest 3.9.x nightly build which uses a MySQL or MariaDB database.

Testers please report back which of the following 2 kinds of databases you have used for the test:
- MySQL database servers with versions lower than 8.0.17 or on any server version of MariaDB
- MySQL database servers with version 8.0.17 or later

If you have both scenarios available, please test both.

**_Step 1:_** Add 2 new update SQL scripts as follows to folder `administrator/components/com_admin/sql/updates/mysql`:

If you are not comfortable with editing you can also download them using the link in the script names.

- Script [3.9.26-2021-03-05.sql](https://test5.richard-fath.de/3.9.26-2021-03-05.sql)
```
CREATE TABLE `#__test_table_1` (
  `id`              int NOT NULL AUTO_INCREMENT,
  `col_int`         int,
  `col_int_u`       int unsigned,
  `col_tinyint`     tinyint,
  `col_tinyint_u`   tinyint unsigned,
  `col_smallint`    smallint,
  `col_smallint_u`  smallint unsigned,
  `col_mediumint`   mediumint,
  `col_mediumint_u` mediumint unsigned,
  `col_bigint`      bigint,
  `col_bigint_u`    bigint unsigned,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;

[...]

Issue #32542, opened 2021-02-27
MYSQL 8 deprecation notices


brianteeman:
Debugging another issue in my mysql editor I got a lot of deprecation notices. This is because of 

https://github.com/joomla/joomla-cms/blob/9ca4e5672c82839f0e7933452cd4ea0085f33a4c/installation/sql/mysql/base.sql#L9-L14

Here you can see we are using the deprecated **display width** on all integers

Confusingly when it is a VARCHAR (50) it means a field width of 50 and therefore controls the number of characters in the field but for integers INT(11) the 11 refers to a **display width** and has no impact on the amount of data stored

References 
- https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html
- https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html#mysql-nutshell-deprecations

richard67:
I can confirm all the above.

richard67:
The same applies to `tinyint`.

See related PR #28501 .

To get rid of the deprecation notices we can change our SQL scripts to not use display width values anymore for `int` and `tinyint` columns. But this has to include also all present update SQL scripts, not only the scripts for new installation.

The question is if we want to do this for staging and 3.10-dev and 4.0-dev, or only for 4.0-dev.

The deprecation notices became more urgent since the default value for the server configuration option to show these notices was not to show them by default, but this may have change with a recent 8.0.x version of MySQL.

We have to keep in mind that MariaDB behaves differently and still supports display widths, as far as I know.

But Brian is absolutely right, these values never had any impact on accuracy, value range or storage size of such database columns, so they should never have been used at the beginning, and we safely can remove them after my PR #28501 has made the database checker clever enough to just ignore these display widths.

richard67:
Am working on it.

richard67:
First PR is ready. It doesn't fix the SQL scripts, so it doesn't fix the issue here ~~, but it is a requirement for fixing the SQL scripts in order not to get database problems reported by the database schema checker~~. See PR #32605 .

Update: That PR is not _required_ to fix the issue here as long as we have no update SQL script modifying columns with any of the types "smallint", "mediumint" or "bigint". But it is helpful to be safe for such changes when this issue here has been fixed.

richard67:
@brianteeman The first PR for staging is ready: #32606 . I'd be happy if you could test. Thanks in advance.

richard67:
[...]

PR #28501, opened 2020-03-28
[3.x] Fix errors about int and tinyint columns shown in database schema check on MySQL 8

richard67:
Pull Request for Issue #28367 .

### Summary of Changes

This Pull Request (PR) changes the database schema checker for MySQL so that any display width is ignored for integer (`int` or `tinyint`) columns when checking these columns.

The reason for this is that beginning with MySQL 8.0.19, the `type` in a `SHOW COLUMNS` statement doesn't include anymore the display width if the database has been created in that version, while on previous versions it is included. E.g. on MySQL 5.7 or 8.0.18 the `type` value is "int(11)" or "int(10) unsigned" or "tinyint(3)" while on mySQL 8.0.19 it is just "int" or "int unsigned" or "tinyint".

The display width doesn't have any impact on value range or storage size for an integer column, it only determines how a value would be left-padded with zeros if that would be done, but this padding is deprecated anyway. See [https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html) for details.

#### Requirements

Have a MySQL database with version 8.0.19 and another one with an older version, e.g. 5.7 or 8.0.18 or a MariaDB. If you don't have both but only one of these 2, report back with which version you have tested. At the end it needs each 2 tests for the following cases:
- MySQL 8.0.19: Test that this PR fixes the issue.
- MySQL lower than 8.0.19 or MariaDB: Test that this PR doesn't create a new issue.

#### Instructions

Install a clean staging without the patch of this PR applied on MySQL 8.0.19, using a database which you have created on that database version or creating a new one during installation.

Then go to the database checker.

Result: You see errors about columns of type `int(11)` and `int(10)` and `tinyint(3)` and similar. See section "Actual result" below.

Now apply the patch of this PR and go again to the database checker or reload the page if still there.

Result: No errors are shown, see section "Expected result" below.

Now delete configuration.php and install the staging with the patch of this PR applied on a MySQL version prior to 8.0.19, e.g. 5.7 or 8.0.18.

Then go to the database checker.

Result: With this PR applied it works as before for MySQL prior to 8.0.19, i.e. there are no database errors after a clean install.

### Expected result

All database table structures are up to date.

No problems were found.

### Actual result

j3-error-database-checker-int-display-width

Using the "Fix" button does not fix these false errors.

### Documentation Changes Required

None.

### Additional information

[...]

PR #32606, opened 2021-03-07
[3.x] Remove display widths for integer data types from SQL scripts for MySQL and MariaDB databases

richard67:
Pull Request for Issue #32542 (part).

### Summary of Changes

This pull request (PR) removes display width modifiers for integer data types in [DDL statements](https://en.wikipedia.org/wiki/Data_definition_language) in all SQL scripts for installation and updates and from PHP code (small piece of code in the extension installer's database model to create the utf8 conversion table when updating very old installations) for MySQL and MariaDB databases.

It is only relevant for MySQL and MariaDB databases, i.e. not for PostgreSQL or MS SQL Server.

These **reason for these changes** is that MySQL have deprecated the display width modifier for integer data type columns beginning with server version 8.0.17, see [here in their documentation](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html), which causes warnings on database servers of these versions when running the relevant SQL statements e.g. in phpMyAdmin, and as MySQL don't use semantic versioning (at least not in x.0 versions), they might stop to support it with any future version, and so these statement will cause SQL errors, and Joomla installation or updates will fail.

In opposite to this, MariaDB still continue to use and support the display width modifier for integer data type columns.

That means we have to kinds of database servers regarding these modifiers:
1. MySQL server versions lower than 8.0.17 or any version of MariaDB server which use these modifiers
2. MySQL server versions 8.0.17 or later which don't use these modifiers

As explained in issue #32542, [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html) and [MariaDB docs](https://mariadb.com/kb/en/numeric-data-type-overview/), the display width modifier for integer data type has no impact on storage size or value range. It is only relevant when using the `zerofill` attribute on such a database column, what the CMS core never did.

In my opinion there are only 2 possible reasons why the attribute has been used by the CMS core:
1. The database schema checker and fixer needed it because it is part of the type returned for a database column when checking the data type if not a MySQL server versions 8.0.17 or later, so you had to specify it in the DDL.
This has already been fixed with PRs #28501 and #32605 , so that's not necessary anymore.
2. When looking on how much effort people have put into the display width values, it seems there was a misunderstanding so that many contributors and maintainers thought it an impact on value range or storage size, maybe due to the similar look to e.g. `float(10)`, `real(10)`, `double(10)`, where the `10` is the precision and not a display width.

The solution provided by this PR here in order to be safe from future drop of support for display width modifiers by MySQL is not so use the display width modifier anymore in any SQL statement.

For DDL statements like e.g. `CREATE TABLE` or `ALTER TABLE ... MODIFY ...`, the result is different for the 2 types of databases when not using the display width modifier:
1. On MySQL server versions lower than 8.0.17 or any version of MariaDB server, the columns will have the default display width for the particular data type, which are:
- tinyint(4)
- tinyint(3) unsigned
- smallint(6)
- smallint(5) unsigned
- mediumint(9)
- mediumint(8) unsigned
- int(11)
- int(10) unsigned
- bigint(20)
- bigint(20) unsigned
2. On MySQL server versions 8.0.17 or later the columns will not have any display width.

#### Requirement

You need a MySQL or MariaDB database.

Testers please report back which of the following 2 kinds of databases you have used for the test:
- MySQL database server with versions lower than 8.0.17 or any server version of MariaDB
- MySQL database server with version 8.0.17 or later

If you have both scenarios available, please test both.

This test is optional and can be only executed when having a MySQL database server of version 8.0.17 or later.

If you don't have that kind of MySQL database server, skip this test.

During my tests I wasn't able to configure my MySQL 8.0.20 server in a way that the deprecation warnings for the display widths could be shown in the server log file, and you can believe me I've tried a lot of things.

Furthermore, it appeared that when importing an SQL script using display widths for integers in DDL into phpMyAdmin, there are also no warnings shown.

So if you want to see the issue you have to do as follows:

1. In phpMyadmin, select a database.

2. Copy the first `CREATE TABLE` statements for the `#__assets` table (or any other which contains display widths) from [file installation/sql/mysql/joomla.sql of the staging branch](https://github.com/joomla/joomla-cms/blob/staging/installation/sql/mysql/joomla.sql) (i.e. not from this PR) into the SQL command window.
Hint: You can use also several statements at once, but you should not use the complete content of the SQL script since in this case phpMyAdmin (or the server?) seem to suppress the warnings due to their high count.

3. Execute the SQL in the SQL command window (no need to replace the table name prefix before).

Result: You get warnings `Warning: #1681 Integer display width is deprecated and will be removed in a future release.`.

4. Remove the table created in the previous step.

5. Repeat steps 2 and 3 but using a statement from the [installation/sql/mysql/joomla.sql file of this PR](https://github.com/joomla/joomla-cms/blob/0d017873466be04fbb273d60b0cde6c53e69d4ad/installation/sql/mysql/joomla.sql).

Result: Don't get the warnings from step 3.

#### Test 2: New installation

The automated system tests make already sure that this PR doesn't produce an SQL error in the installation SQL, but it doesn't check if we have database problems in the schema checker after that or not, therefore this test.

1. Make again a new installation with the patch of this PR applied.

2. Login to backend.

3. Go to "Extensions -> Manage -> Database" and verify that no database problems are shown.

Result: No database problems found.

4. Using a tool like e.g. phpMyAdmin, check data type of columns with integer data types.

Result:
- On MySQL server versions lower than 8.0.17 or any version of MariaDB server, the display width values of the integer type columns have the default values for that data type. See above in the summary of changes for the complete list.
- On MySQL server versions 8.0.17 or later, the integer type columns don't have any display width values.

#### Test 3: Update from a Joomla version older than 3.9.0.

Unfortunately the packages built by drone are only for updating previous 3.9.x versions, beginning with 3.9.0, but not older.

This means that the changes on the update SQL scripts can't be tested with such an update package.

Therefore I've build a modified update package for updating older versions.

This can be downloaded from here and then used for the test with "Upload & Update": [https://test5.richard-fath.de/Joomla_3.9.26-dev+pr.32606-Development-Update_Package.zip](https://test5.richard-fath.de/Joomla_3.9.26-dev+pr.32606-Development-Update_Package.zip).

The starting point is a clean Joomla installation with a version older than 3.9.0 (the older, the better, as long as possible on your testing environment).

1. Login to backend.

2. Make sure to see any database errors or warnings by switching on "Debug System" and setting "Error Reporting" to "Maximum" or "Development" in Global Configuration.

3. Update to current staging + the patch of this PR applied by using the update package mentioned above, using the "Upload & Update" method.

4. While the update runs, verify that no SQL errors or warnings are shown in backend or in your MySQL or MariaDB server's log file or in the PHP error log.

5. Go to "Extensions -> Manage -> Database" and verify that no database problems are shown.

Result: No database problems found.

6. Using a tool like e.g. phpMyAdmin, check data type of columns with integer data types.

Result:
1. Display widths of database columns which existed already before the update and which haven't been modified for other changes haven't been changed.
2. New tables have been created with display widths for columns with integer data types depending on the type of database:
- On MySQL server versions lower than 8.0.17 or any version of MariaDB server, the display width values of the new integer type columns have the default values for that data type with this PR applied, see above in the summary of changes for the complete list.
- On MySQL server versions 8.0.17 or later, the new integer type columns don't have any display width values with this PR applied.
3. The same as 2. applies to database columns with integer data types which have been modified for other purpose during the update, e.g. type change or change of default value or null values.

SQL scripts and at one place the installer's database model use the unnecessary and for MySQL versions 8.0.17 or later deprecated display width modifier for integer data types.

No display width modifier is used for integer data types at any place in SQL or other code except of unit tests coming with the framework package.

After a new installation on MySQL database servers with versions lower than 8.0.17 or any version of MariaDB servers, database columns with integer data types have the default display widths for the particular data type. On MySQL database servers with versions 8.0.17 or newer, such database columns have no display widths.

When updating from 3.8.x or older, display widths haven't been changed for database columns which existed already before the update and haven't been modified by the update for other changes than display widths. For newly created or modified columns, the above statement for new installations applies.

None.

### Other information

@zero-24 @wilsonge This PR needs to be merged up into 3.10-dev sooner of later after it has been merged into staging, so that with PR #32607 the 3.10-dev branch will be complete regarding the issue fixed here. But this PR here does **_not_** need to be merged up into 4.0-dev, i.e. any conflicts resulting in an upmerge in 4.0-dev done after my PR #32608 has been merged into the 4.0-dev branch could be solved by accepting the 4.0-dev version of the particular conflict.


richard67:
I've reverted the changes on database unit tests. These were partly made already in the framework.

richard67:
I've also simplified testing instructions. Ready for testing.

chmst:
Tested on xampp, php 7.4, MySQL server versions lower than 8.0.17

Test 1 skipped

Test 2 performed successfully

Test 3 performed with 3.8.0  -

after update there were the following differences in database (should have no consequences)

grafik

richard67:
> after update there were the following differences in database (should have no consequences)
> 
> grafik

@chmst These changes are right, because when updating a 3.9.25, the table columns are not touched and so keep their display widths from before the update, while when updating the 3.8 with my modified package, they are created with the default display widths. They show that the PR works as expected.

chmst:
I have tested this item :white_check_mark: successfully on 76a6fecd245ab4fe9ac671be37f85a59676e0221<br /><br />See comment below concerning my testing environment.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32606">issues.joomla.org/tracker/joomla-cms/32606</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 76a6fecd245ab4fe9ac671be37f85a59676e0221<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32606">issues.joomla.org/tracker/joomla-cms/32606</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32606">issues.joomla.org/tracker/joomla-cms/32606</a>.</sub>

richard67:
Thanks for testing and merge.

Issue #32561, opened 2021-03-02
[4.0] User profile after login via com_users


chmst:
### UX of user profile should be improved
When a user wants to log in via com_users (not via user module) he/she sees the following useless profile. 

<img width="619" alt="profile1-j4" src="https://user-images.githubusercontent.com/1035262/109638192-fbac3f80-7b4d-11eb-8332-6bf94acca699.PNG">

[ ] So why not show at once the profile form if the token must be set? 
[ ] If nothing is to do: write a message like "Welcome, you are logged in -  you can change settings with the button above"
[ ] css for dl-horizontal should be added 
[ ] Instead of the useless "Website default" say what is the default.  
[ ] The hint "save your profile for getting a token" is misleading - the user first has to klick the "edit" button 
[ ] Make the profile-edit  button a real button

bembelimen:
[ ] Going on the login page (by clicking a link) and login should not redirect to the profile but to the last visited page...

chmst:
The screen for menuItem of type login is bad UX.

<img width="1188" alt="login-redirect" src="https://user-images.githubusercontent.com/1035262/109873040-82a40980-7c6d-11eb-8da2-dd9dc2737269.PNG">

As a new user (and in my case as a very old user but with few users in frontend) it seems that the redirect url is set to home, so seems to be correct.  That the SELECT button changes to EDIT after a selection is made is not selfexplaining.

richard67:
Closing as having a pull request. Please test #33603 . Thanks in advance.

PR #32613, opened 2021-03-08
Change placeholder text for login redirect

drmenzelit:
Pull Request for Issue #32561 comment https://github.com/joomla/joomla-cms/issues/32561#issuecomment-790059986 .

### Summary of Changes
~~Changed text in `/administrator/language/en-GB/com_menus.ini` from "Home" to "Select a menu item"~~
Changes in com_menus.ini reverted, because that was wrong.
Changed the value of the fields to "Select menu" (existing string `JOPTION_SELECT_MENU`).

As stated in the comment mentioned above the text "Home" in the redirect field can be misunderstood. The field is disabled, but it seems as the "Home" menu item is already selected as redirect page.
Added descriptions to the fields to explain the behaviour of login and logout redirect.

### Testing Instructions
Create a login menu, set "Choose Login Redirect Type" to "Menu item". 
Create a login modul.

Login menu
Screenshot_2021-03-08 Menus Edit Item - Joomla 4 Beta - Administration

Login module
Screenshot_2021-03-08 Modules Login - Joomla 4 Beta - Administration

Login menu
Screenshot_2021-03-08 Menus Edit Item - Joomla 4 Beta - Administration3
Screenshot_2021-03-08 Menus Edit Item - Joomla 4 Beta - Administration-Logout

Login module
Screenshot_2021-03-08 Modules Login - Joomla 4 Beta - Administration3


chmst:
Much better! But wouldn't it be better to use anoter language key? 
JDEAULT is one of the common language keys as JYES, JNO, JCATEGORY which have the same translation over the whole system. 
In all other language files JDEFAULT is translated wit "Default".

brianteeman:
Does it even need a hint?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32613">issues.joomla.org/tracker/joomla-cms/32613</a>.</sub>

drmenzelit:
"Home" as placeholder is wrong, "Default" as placeholder (if we don't override JDEFAULT) is also not self explaining. So we need something better. Which way is better: override the value of JDEFAULT or use another language string?

chmst:
@brianteeman the hints are very old and inherited fom J3. There the hint is "default", but has added  a description what "default" means.  
grafik

Imho "the same page" is not user profile when I am on the login page - but this is not in scope here. I also always was surprised in J3 installations when the profile appeared after login. 

This PR makes it better for now and blocks nothing. 

Further improvements can be
- Remove the hint (in all xml files)
- Change the default behaviour as the j3 description falsely announces and remain on the login page 
- Change the default behaviour and set "home" as default as the wrong hint falsely promised

chmst:
I have tested this item :white_check_mark: successfully on 17625bfe2c55013f61a0e7e1f424f24ea2510a7f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32613">issues.joomla.org/tracker/joomla-cms/32613</a>.</sub>

joomdonation:
Hmm. Change to Select Menu Item is not right to me because it indicates that we need to select a menu item for these parameters?

# For Login Form menu option
- **Menu Item Login Redirect** should not be **Home**.  If we do not choose a menu item there, users will be redirected to User Profile page. So maybe changes it to **User Profile** will be better
- **Menu Item Logout Redirect** set to **Home** is expected because if we do not choose a menu item there, users will be redirected to Home when logount using that menu option

# For Login module

- **Login Redirection Page** and **Logout Redirection Page** maybe should default to **Current Page** because if we do not select menu item there, users will be redirected to the current page after login/logout.

chmst:
I see this PR as a temporary qick improvement until other changes are made. 

@joomdonation 
I agree with you and see here - in #32561 (https://github.com/joomla/joomla-cms/issues/32561#issuecomment-789053320)  this is already required. 
Maybe you could make the PR?

joomdonation:
@chmst Better if @drmenzelit could update this PR if we agree with that (I'm not confident with my English :D )

brianteeman:
@chmst Just because they are old doesn't mean they can't be improved or changed. If necessary even the description can be put back for this field. And as I am sure you know placeholder text/hints are terrible from an accessibility perspective.

However I just took a look at the actual code and "Home" is not a placeholder text or hint. It is the option which has a value of "". 

Therefore the override for this language string has always been wrong and should never have said Home and the problem is in the XML not the language string.

Instead of 
`<option value="">JDEFAULT</option>`
the xml should have been

`<option value="">JOPTION_SELECT_MENU</option>`

So the language string override should be removed from administrator/language/en-GB/com_menus.ini  and the xml file fixed as above - problem solved without the confusion of incorrect text and without overriding an override. 

It might still need a description and I see that a description is used for the other redirect method

drmenzelit:
I will try to improve this PR. Thank you @joomdonation and @brianteeman for the comments.

drmenzelit:
A new try ;-) I have edited the description of the PR to adapt to the changes made on my last commit.
@brianteeman I will appreciate your help with the new descriptions... you know, English is not my native language ;-)
I think `Select menu` is the correct value for the fields, it shows that an action is needed if you want to change the standard behaviour. And the standard behaviour is explained in the description.

components/com_users/tmpl/login/default.xml
```diff
@@ -42,14 +42,15 @@
 					name="login_redirect_menuitem"
 					type="modal_menu"
 					label="COM_USERS_FIELD_LOGIN_REDIRECTMENU_LABEL"
+					description="JFIELD_LOGIN_REDIRECT_MENU_DESC"
```
brianteeman: Why is this JFIELD and not a COM_USERS string. Is it used outside of this component? joomla.ini should only be for global strings imho

brianteeman:
The texts are fine - my only question is about the language file

drmenzelit:
`JOPTION_SELECT_MENU` was already defined in joomla.ini ...

brianteeman:
> `JOPTION_SELECT_MENU` was already defined in joomla.ini ...

That one is correct as its global.
I am referring to JFIELD_LOGIN_REDIRECT_MENU_DESC

drmenzelit:
Sorry, I didn't read correctly... since the labels `JFIELD_LOGIN_REDIRECT_URL_LABEL` and `JFIELD_LOGOUT_REDIRECT_URL_LABEL` were also in joomla.ini, I thought it is better to put the description also there...

brianteeman:
It makes no sense to introduce a new field into two language files and then only use one of them. That is just creating work for translators that doesn't need to be done. 

Someone else may correct me but my recollection is that joomla.ini is for global strings and as this string is only for com_users it should live in the com_users.ini

There may have been mistakes in the past but that doesnt mean we should repeat them ;)

components/com_users/tmpl/login/default.xml
drmenzelit: Corrected in last commit. The other language strings can be moved in another PR is wished / needed.

brianteeman:
I have tested this item :white_check_mark: successfully on 5adc2ca3bde69dc537c4c15147a94f256d2a3973<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32613">issues.joomla.org/tracker/joomla-cms/32613</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on 5adc2ca3bde69dc537c4c15147a94f256d2a3973<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32613">issues.joomla.org/tracker/joomla-cms/32613</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32613">issues.joomla.org/tracker/joomla-cms/32613</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32613">issues.joomla.org/tracker/joomla-cms/32613</a>.</sub>

wilsonge:
Thanks!

Issue #32571, opened 2021-03-03
J4B8 - API - syntax error, unexpected token "match"


ipv6wiz:
### Steps to reproduce the issue
1. Download latest release or even J4B7.   
1. Setup MySQL DB.  
1. GO through installation - finishes ok and all appears fine
1. Create Category and Article assigned to Category
1. Go to Users and generate Token
1. Go to Postman set token as Bearer Token and base path to point to local installation
1. create GET request to `{{base_path}}/api/index.php/v1/content/article` from the official collection
1. Send request

### Expected result
A list of Articles

### Actual result
syntax error, unexpected token "match"

### System information (as much as possible)
Setting	Value
PHP Built On	Linux ubuntu 5.8.0-44-generic #50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64
Database Type	mysql
Database Version	8.0.23-0ubuntu0.20.04.1
Database Collation	utf8mb4_unicode_ci
Database Connection Collation	utf8mb4_0900_ai_ci
Database Connection Encryption	None
Database Server Supports Connection Encryption	Yes
PHP Version	8.0.2
Web Server	Apache/2.4.41 (Ubuntu)
WebServer to PHP Interface	apache2handler
Joomla! Version	Joomla! 4.0.0-beta8-dev Development [ Mañana ] 2-February-2021 20:02 GMT
User Agent	Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.182 Safari/537.36
PHP Settings
PHP Settings
Setting	Value
Upload Max Filesize	75M
Post Max Size	80M
Memory Limit	128M
Open basedir	None
Display Errors	On
Short Open Tags	Off
File Uploads	On
Output Buffering	On
Session Save Path	/var/lib/php/sessions
Session Auto Start	0
XML Enabled	Yes
Zlib Enabled	Yes
Native ZIP Enabled	No
[...]

PR #32616, opened 2021-03-08
[4] composer update willdurand/negotiation from 2.3 to 3.0

alikon:
Pull Request for Issue #32571 .

### Summary of Changes
willdurand/negotiation from 2.3 to 3.0 needed for php 8

### Testing Instructions
see #32571

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "19772794cc68edcef1d5b83189f520b7",
+    "content-hash": "68cc312f4c0dae0d21a9735a9594c57b",
     "packages": [
         {
             "name": "algo26-matthias/idna-convert",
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -753,16 +753,16 @@
         },
         {
             "name": "joomla/console",
-            "version": "2.0.0-beta3",
+            "version": "2.0.0-beta2",
```
richard67: Downgrade joomla/console from 2.0.0-beta3 to 2.0.0-beta2? Why that?

composer.lock
```diff
@@ -1143,16 +1133,16 @@
         },
         {
             "name": "joomla/filter",
-            "version": "2.0.0-beta5",
+            "version": "2.0.0-beta2",
```
richard67: Downgrade joomla/filter from 2.0.0-beta5 to 2.0.0-beta2? Why that?

composer.lock
```diff
@@ -1163,7 +1153,7 @@
                 "joomla/language": "<2.0"
             },
             "require-dev": {
-                "joomla/coding-standards": "^3.0@dev",
+                "joomla/coding-standards": "^2.0@alpha",
```
richard67: This unwanted changed seems to be caused by one of the previous unwanted changes (downgrades).

composer.lock
alikon: i've started with an old branch 😜

alikon: fixed

composer.lock
alikon: fixed

composer.lock
alikon: fixed

richard67 (approved):

wilsonge:
Thanks!

Issue #32612, opened 2021-03-08
[com_content] 'Category View' loads all items on mysql


trananhmanh89:
### Steps to reproduce the issue
- Site has a category with ~ 30 items
- Site has only one menu item
- Set that menu to any component but not com_content
- Acess category listing by using url `index.php/component/content`
- Open link of category that has ~ 30 items

### Expected result
- As you see, there is no menu with category type, so listing limit should use global config

### Actual result
- As i check, `list.limit` is 0, so joomla will get all items of that category at a time

### System information (as much as possible)
- Joomla 3.9.25
- php 7.2
- mysql 5.7
- os windows 10 - 20H2

### Additional comments
- this issue causes php run out of memory on site has many articles ( in my case is ~ 20,000 items)

chmst:
Your Joomla Version surely is 3.9.25?

trananhmanh89:
> Your Joomla Version surely is 3.9.25?

Oh, sorry.
I correct my info

Quy:
Duplicate #18529?

trananhmanh89:
> Duplicate #18529?

Yeah, same issue, but different way to achieve it

rdeutz:
closing because it is a of duplicate #18529

Issue #18529, opened 2017-11-08
[J5] All articles data is loaded in Category Menu when 0 articles are set to be displayed - performance impact


AndySDH:
**POST UPDATED TO JOOMLA 5**

### Steps to reproduce the issue

- Have a category with many articles (hundreds? the more the better to see the performance impact. for this example I'll use a random number of 432 articles)
- Create a Category Blog Menu Item of that category, but in "Blog Layout" options set 0 articles to display in all four inputs. 
- Visit this menu item in the frontend. 

### Expected result

I would expect the page to load super fast because there is no article data to load whatsoever.

### Actual result

The loading is VERY SLOW, taking dozens of seconds, if not minutes, depending on the amount of articles and content within those articles. Why? Because Joomla is loading in the background the data from ALL those 432 articles as if they were all displayed at the same time in that one page with no pagination, despite the fact you're actually displaying 0!

### Additional comments

Still happening as of Joomla 5

### Additional comments

To top it off, obviously, if in the options you put to only display "1" article, of course it loads super fast because it only loads 1 article, correctly. You put "20" articles? It displays quite fast because it only loads 20 articles, correctly. You put "0" articles? It loads incredibly slow because it loads all 432 articles in the background (and then showing 0 at the end of the loading). This can't be intended behavior.

In an old comment, someone suggested to put "-1" instead of "0" articles in the Menu options to work around the problem, and while it worked before, now in Joomla 5 it's no longer possible because it's impossible to save a menu item with a "-1" as the value. So now there is not even a workaround anymore.

PS: You may ask, why would you want to input 0 articles in the first place? Well, it could be useful for cases when you might want to only display the description of the category, or if you want to display the articles via other means (content plugins or modules) and you don't want to display them natively. Plus, the use case shouldn't even matter, it's just a wrong/flawed behavior regardless.

csthomas:
Because `$limit = 0` usual means no limit there is hard to change it in J3.x.

It can help only in the blog:
```diff
diff --git a/components/com_content/models/category.php b/components/com_content/models/category.php
index 50c66d0efa..a4f4128daa 100644
--- a/components/com_content/models/category.php
+++ b/components/com_content/models/category.php
@@ -190,6 +190,13 @@ class ContentModelCategory extends JModelList
                if (($app->input->get('layout') === 'blog') || $params->get('layout_type') === 'blog')
                {
[...]

PR #32626, opened 2021-03-10
Fix articles limit for category view

joomdonation:
Pull Request for Issue #18529 and #32612 .

### Summary of Changes
This PR proposes a solution to fix the two issues #18529 and #32612. Basically, it now will only load articles if limit > 0  (users want to display articles using the category layout)

See #18529 and #32612 to understand the actual issue. However, this PR has a backward incompatible change, so I'm unsure if it should be accepted (especially for staging).

### Testing Instructions
1. Create two menu items, one links to Category Blog Layout menu item type and one to Category List menu item type.
2.  Apply patch, confirm that two menu items still display articles same as before.

### Backward Incompatible Changes
Please note that there is a backward incompatible changes here then the site has no menu item to link to one of the three menu item types (basically, no suitable menu item could be found to link to a category)

- List All Categories
- Category Blog Layout
- Category List 

For example, when you unpublish all menu items (keep home menu but links to a different component than com_content) and access to a category display in Articles - Categories. Before this PR, it will display all articles from that category, after this PR, only number of articles will be displayed (controlled by Default List Limit parameter in Global Configuration)

### DIFF
--- a/components/com_content/models/category.php
+++ b/components/com_content/models/category.php
@@ -194,7 +194,7 @@ class ContentModelCategory extends JModelList
 		}
 		else
 		{
-			$limit = $app->getUserStateFromRequest('com_content.category.list.' . $itemid . '.limit', 'limit', $params->get('display_num'), 'uint');
+			$limit = $app->getUserStateFromRequest('com_content.category.list.' . $itemid . '.limit', 'limit', $params->get('display_num', $app->get('list_limit')), 'uint');
 		}
 
 		$this->setState('list.limit', $limit);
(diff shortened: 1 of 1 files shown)


richardandersonss:
.[Types Of ESA Letters To Consider AS A Mental Health Patient - 2021 Guide.pdf](https://issues.joomla.org/uploads/1/fafd54f2e47d2d4e34869114a8398ac2.pdf)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32626">issues.joomla.org/tracker/joomla-cms/32626</a>.</sub>

cameronstewart13:
.[Research Paper Outline On Tightening Of Gun Control Laws.pdf](https://issues.joomla.org/uploads/1/3019880abe492da141fa830cf0b2cbe1.pdf)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32626">issues.joomla.org/tracker/joomla-cms/32626</a>.</sub>

cameronstewart13:
.[A Professional Guide To Start A Research Paper.docx](https://issues.joomla.org/uploads/1/7d26674b4dca11d39752734be2600ebf.docx)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32626">issues.joomla.org/tracker/joomla-cms/32626</a>.</sub>

cameronstewart13:
[A Professional Guide To Start A Research Paper.pdf](https://issues.joomla.org/uploads/1/45e3610901f73e46079f3cd90122b097.pdf)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32626">issues.joomla.org/tracker/joomla-cms/32626</a>.</sub>

AndySDH:
Thanks for the effort!

Unfortunately, I tested this unsuccessfully. This results in this error when viewing a category with 0 articles displayed:

0 - Call to a member function getPagination() on null

joomdonation:
@AndySDH Will look at it later.

trananhmanh89:
@joomdonation should we return empty array at category model when `$limit == 0` ?

or return empty array on `ListModel.php`, because this issue can affect other core components.

joomdonation:
> Should we return empty array at category model when `$limit == 0` ?

I have to look at it again but I think it is working like that for category model at the moment

> return empty array on ListModel.php, because this issue can affect other core components

It is a backward-incompatible change, so we are not allowed to do that. Further more, in Joomla!, look like we want all records returned when no limit is set.

trananhmanh89:
I think we only need change line 251 of `components\com_content\models\category.php`

from `if ($limit >= 0)`

to `if ($limit > 0)`

joomdonation:
> I think we only need change line 251 of `components\com_content\models\category.php`
> 
> from `if ($limit >= 0)`
> 
> to `if ($limit > 0)`

I don't think that's enough. It will cause no articles being displayed when access to a category from Articles - Categories module (with no menu items linked to categories or category view). Also, if we just change that, there are still unnecessary code executed ($model->setState commands) when we do not need to query database to get articles.

I will try to review code carefully later.

trananhmanh89:
Correct me if i'm wrong. Your code is trying to resolve unlimited items on two cases:
### 1. Menu com_content, view=category and all settings are set to 0 item
- Your solution: don't call model->getItems() when limit <= 0
- My point: It's ok. And i found that the pagination still load and make query to get total. Is that ok?
### 2. You access com_content, view=category without any menu
- We can use this url to replicate : `index.php?option=com_content&view=category&id={category_id}`
- Your solution: set limit to value of global config `list_limit`
- My point: why don't we use global config of com_content in this case? May be this check no longer needed?

// Set limit for query. If list, use parameter. If blog, add blog parameters for limit.
`if (($app->input->get('layout') === 'blog') || $params->get('layout_type') === 'blog')`

AndySDH:
2. Exactly, Category Blog has default settings, so it should fall back to that:

image

And if this was not the case, I don't think this is a "BC Change", this is a "Bug Fix" :D

trananhmanh89:
> 1. Exactly, Category Blog has default settings, so it should fall back to that:
> 
> image
> 
> And if this was not the case, I don't think this is a "BC Change", this is a "Bug Fix" :D

I've found a mistake at my point. That you can change layout of category itself to `Listing`. In this case, the limit should be global config `list_limit` as @joomdonation did :). 

But if category layout is `blog`, the limit should respect blog global config, otherwise the pagination will be wrong.

PR #32631, opened 2021-03-10
Add use xml castom options.

korenevskiy:
Replace code in field
**`$options = array();`**
to
**`$options = parent::getOptions();`**
This will allow you to add additional tags to the XML field at your own discretion.
Example:
```
<field name="userHeader_tag" type="headertag"
           label="HTML-Tag" 
           default="div" >
        <option value="span">span</option>
 </field>
``` 
.


Quy:
Included in PR #35184

PR #32645, opened 2021-03-11
Renaming the name of the controlClass attribute for form fields

korenevskiy:
I apologize.
I believe that the name of the newly added attribute for the fields is not quite correct.
Calling it the parent class. This is logical, but it is not clear which parent it refers to.
At first glance, you might think that this is the parent of the field. But this is not the case.
After thinking about it, you might think that this is the parent of the control, but this is also not the case. Since the control itself is the target.
img-2021-03-09-13-35-32
https://github.com/joomla/joomla-cms/pull/32488
The author @ciar4n  of the attribute agreed to replace the attribute name.


libraries/src/Form/FormField.php
```diff
@@ -1351,7 +1351,7 @@ protected function getLayoutData()
 			'value'          => $this->value,
 			'dataAttribute'  => $this->renderDataAttributes(),
 			'dataAttributes' => $this->dataAttributes,
-			'parentclass'    => $this->parentclass,
+			'controlclass'    => $this->controlclass,
```
```suggestion
			'controlclass'   => $this->controlclass,
```

ciar4n:
> The author @ciar4n of the attribute agreed to replace the attribute name.

Not exactly true 😄 ...  I just said I wasn't going to apply your suggestion to a PR that had already gone to RTC.

korenevskiy:
> > The author @ciar4n of the attribute agreed to replace the attribute name.
> 
> Not exactly true 😄 ...  I just said I wasn't going to apply your suggestion to a PR that had already gone to RTC. 

but you put a thumbs-up emoji there. 👍

ceford:
I have tested this item :white_check_mark: successfully on 6901dbc973018dabe98c9eaa4deb66f13419552d<br /><br />I looked at the diff, applied the patch and searched for parentclass and controlclass. I think it is OK.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32645">issues.joomla.org/tracker/joomla-cms/32645</a>.</sub>

dgrammatiko:
I think if you dislike the `parentclass` an alternative **logical** value could be `containerclass` and **not** `controlclass`.

korenevskiy:
> I think if you dislike the `parentclass` an alternative **logical** value could be `containerclass` and **not** `controlclass`.

@dgrammatiko 
I agree with you. I think Container class is an equivalent sentence. But there is one thing.
Let's look at the XML configuration.
Field **`<field name= "title" type= "text" label= "Title"/>`**
Please tell me this element is primarily a control, container or field?
Perhaps it also makes sense to consider the name **`fieldclass`**
I think choosing a name for this class is a very important thing, even if Joomla 5 comes out, the name of the attributes will remain for centuries. We must choose the name of this attribute responsibly.
Where to conduct a survey among developers to select
**`controlclass`** VS **`containerclass`** .
**`containerclass`** a good attribute name.
and **`caseclass`** another option.

HLeithner:

PR #32649, opened 2021-03-12
Update default.php

saumyasarkar11:
Pull Request for Issue # .

### Summary of Changes

Check for the emptiness of $urls on lines 98 and 132 has been removed since the variable $urls can never be empty.

### Testing Instructions

Create an article and check whether it is properly displayed or not with its url.

Untitled

Untitled

None


drmenzelit:
This change is not correct. `$urls` can be empty. You didn't tested using the URL fields on an article. In the screenshot you see the fields Link A, Link B and Link C, that are the URL fields:

grafik

saumyasarkar11:
> This change is not correct. `$urls` can be empty. You didn't tested using the URL fields on an article. In the screenshot you see the fields Link A, Link B and Link C, that are the URL fields:
> 
> grafik

Got it. Thanks for the feedback. Will look into it.

PR #32656, opened 2021-03-12
[4.0] Remove unused function parameter

sakiss:
### Summary of Changes
Removal of the unused `$contexts` parameter in the function `batchCopy`

### Testing Instructions
Batch copy a couple of articles

They are copied

They are copied

DKN

### DIFF
--- a/libraries/src/MVC/Model/AdminModel.php
+++ b/libraries/src/MVC/Model/AdminModel.php
@@ -327,7 +327,7 @@ abstract class AdminModel extends FormModel
 
 			if ($cmd === 'c')
 			{
-				$result = $this->batchCopy($commands[$this->batch_copymove], $pks, $contexts);
+				$result = $this->batchCopy($commands[$this->batch_copymove], $pks);
 
 				if (\is_array($result))
 				{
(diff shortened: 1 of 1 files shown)


brianteeman:
What makes you think this is unused?

sakiss:
> What makes you think this is unused?

Check inside the function

brianteeman:
You mean like here where you removed context


```
				$result = $this->batchCopy($commands[$this->batch_copymove], $pks);
				if (\is_array($result))
				{
					foreach ($result as $old => $new)
					{
						$contexts[$new] = $contexts[$old];
					}
					$pks = array_values($result);
```

sakiss:
i mean inside the batchCopy fn

joomdonation:
@sakiss Although method **batchCopy** inside **AdminModel** class does not use that parameter, there are classes (which extends AdminModel) expect to receive this parameter like this https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_categories/src/Model/CategoryModel.php#L940 , so you cannot remove it.

sakiss:
@joomdonation To my knowledge, calling a function with more than it's declared params it's legal php without any effect.
http://sandbox.onlinephpfunctions.com/code/fa82bad0a6f49003f0d651a931be49e2ce104066

joomdonation:
Yes @sakiss. But what if the model class which extends AdminModel need to use that data for some reasons? With your change, they won't receive that parameter anymore (not only core model classes but also third party extensions)

Also, although it is legal, I think that the number of parameters should be the same.

sakiss:
@joomdonation this change does not affect function overrides in child classes whatsoever. 
The only concern is that in php 8, if a function is overwritten and introduces additional parameters, these should have default values. Otherwise an error is generated.
Because changing all the classes that extend that fn, adding default params, is PIITA and also error prone for 3rd party extensions, i am closing the PR.

PR #10090, opened 2016-04-25
Improve searchtools - part 2

andrepereiradasilva:
#### Summary of Changes

In sequence of https://github.com/joomla/joomla-cms/pull/9738, this PR does some improvements to searchtools with the goal of providing more flexibility and to remove all backend searchtools layout overrides.
1. Adds a attribute `noresults` to the search field in the filter fields xml. This attribute contains the language string of the no results message of the searchtools
   
   Example:
   
   ``` xml
   <field
     name="search"
     type="text"
     label="JSEARCH_FILTER"
     description="JSEARCH_FILTER"
     hint="JSEARCH_FILTER"
     noresults="COM_CHECKIN_NO_ITEMS"
   />
   ```
2. Adds a attribute `layout` to the selector field in the filter fields xml. This attribute contains the layout where the field should be rendered (no attribute for all). This is useful for views that have 2 layouts. example modules has default layout and modal layout.
   Adds also a attribute `filtermode` to the selector field in the filter fields xml. This attribute has only one value `selector`. This is used to identify which field is the selector, when exists.
   Example:
   
   ``` xml
   <field
     name="client_id"
     type="list"
     onchange="this.form.submit();"
     filtermode="selector"
     layout="default"
     >
     <option value="0">JSITE</option>
     <option value="1">JADMINISTRATOR</option>
   </field>
   ```
3. Now it can autodetect if the filters button "Search Tools" should exist or not.
4. Removed all the, now not needed, searchtools backend layout overrides.
5. As PoC i implemented, in this PR, the changes in all views that currently have a "selector":
   - com_menus items view: selector menutype
   - com_cache cache view: selector client_id
   - com_modules modules view (default and modal layout): selector client_id
   - com_languages installed view: selector client_id
   
   image
6. Also, as PoC implemented it in (views without selector):
   - com_checkin checkin view
   - com_modules modules view (frontend popup)
   
   image
#### Testing Instructions
1. Test if the searchtools work fine in the several views changed in this PR.
2. Test if the searchtools work fine in some of the unchanged views (ex: articles, categories, etc).
#### Observations

After, and if, this is merged i will change all the other views to use this new searchtools model, by replacing something like this:

``` php
<?php echo JLayoutHelper::render('joomla.searchtools.default', array('view' => $this, 'options' => array('filterButton' => false))); ?>
<div class="clearfix"></div>
<?php if (empty($this->items)) : ?>
    <div class="alert alert-no-items">
[...]

PR #32660, opened 2021-03-12
[4.0] com_fields Search [a11y]

brianteeman:
The Search field on the list of fields and field groups was missing a label.

The field also had a class that doesnt exist and was a remnant from j3

Code review and View generated source to check the search field has a real label and one that can be translated


sksuryan:
So, I verified that the class is being removed but correct me if I wrong, the updated labels should come at the highlighted place right? If so, they aren't being shown here. 

image

**Edit:**
The problem seems to be in `layouts/joomla/searchtools/default/bar.php` at `line 42`.
https://github.com/joomla/joomla-cms/blob/e317fe221b00d67b9e72cc9c4a359e829f20830d/layouts/joomla/searchtools/default/bar.php#L42-L48

This condition seems to cause the issue.

brianteeman:
The labels are deliberately not visible. Just because you dont display a label on the screen doesn't mean that you don't need to have a label on every input. This is very important for accessibility

sksuryan:
I'm not asking why the labels are hidden, I understand that.

image

Take a look at the text of label, the correct text isn't being displayed and I think it's caused due to this particular code.


> **Edit:**
> The problem seems to be in `layouts/joomla/searchtools/default/bar.php` at `line 42`.
> 
> https://github.com/joomla/joomla-cms/blob/e317fe221b00d67b9e72cc9c4a359e829f20830d/layouts/joomla/searchtools/default/bar.php#L42-L48
> 
> This condition seems to cause the issue.

brianteeman:
Sorry - I understand you know. I will submit a patch shortly

sksuryan:
That's alright, my intention was just to help. Could have added a clearer screenshot though, will keep that in mind. 😄

brianteeman:
So this PR is correct. What you have identified is a bigger possible issue effecting many components which I need to look at the reason for that code before proposing a solution in a separate PR

sksuryan:
Yes, @brianteeman would you mind if I create an issue for this?

brianteeman:
No need to create the issue - I will do a full PR when I get to the reasoning

brianteeman:
I have looked back into the history and I cant see why the conditional statement was introduced back in #10090

sksuryan:
I think it's there in case there isn't any label provided. The thinking may have been to check if any label was provided, in case there wasn't any, then to fallback to a standard "Search" label. But since this condition isn't proper, it keeps on falling back to the default. So, We would have to change that particular condition.

brianteeman:
I suspect that at the time the code was introduced there were search fields that didnt have a label so this was a quick and dirty attempt to make sure they all got a label. In which case we can probably just kill the if statement as they all have a valid search label now.

brianteeman:
Or just fix the condition so that it actually does what it was intended to do. Display the search label for the field and if there isnt one to fallback to the global string.

If you want to have a try at creating a PR for that then go for it

sksuryan:
> 
> Or just fix the condition so that it actually does what it was intended to do. Display the search label for the field and if there isnt one to fallback to the global string.
> 
> If you want to have a try at creating a PR for that then go for it

Yes, I would love to! 
I think it's best if we have a fallback just in case a label isn't present.

ceford:
Just noting that the label does not appear in the source and waiting for a bigger fix before testing.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32660">issues.joomla.org/tracker/joomla-cms/32660</a>.</sub>

chmst:
@ceford #32660 and #32670 must be tested together.

sksuryan:
@chmst I think this one would be ready to commit as soon as #32670 is merged which is RTC but haven't been merged yet.

sksuryan:
I have tested this item :white_check_mark: successfully on 6e3a2c2d0804b4db0ab9dc5842a95f525d53b72f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32660">issues.joomla.org/tracker/joomla-cms/32660</a>.</sub>

Quy:
I have tested this item :white_check_mark: successfully on 6e3a2c2d0804b4db0ab9dc5842a95f525d53b72f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32660">issues.joomla.org/tracker/joomla-cms/32660</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32660">issues.joomla.org/tracker/joomla-cms/32660</a>.</sub>

wilsonge:
Thanks!

brianteeman:
thanks

PR #32666, opened 2021-03-13
Removed deprecated iframe attributes from com/mod_wrapper #32598

Megh05:
Pull Request for Issue #32598.

### Summary of Changes

Removed deprecated iframe attributes from 'components/com_wrapper/views/wrapper/tmpl/default.php' & 'modules/mod_wrapper/tmpl/default.php'

### DIFF
--- a/components/com_wrapper/views/wrapper/tmpl/default.php
+++ b/components/com_wrapper/views/wrapper/tmpl/default.php
@@ -30,8 +30,6 @@ JHtml::_('script', 'com_wrapper/iframe-height.min.js', array('version' => 'auto'
 		src="<?php echo $this->escape($this->wrapper->url); ?>"
 		width="<?php echo $this->escape($this->params->get('width')); ?>"
 		height="<?php echo $this->escape($this->params->get('height')); ?>"
-		scrolling="<?php echo $this->escape($this->params->get('scrolling')); ?>"
-		frameborder="<?php echo $this->escape($this->params->get('frameborder', 1)); ?>"
 		<?php if ($this->escape($this->params->get('page_heading'))) : ?>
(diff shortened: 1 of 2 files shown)


brianteeman:
Stop and think about this change for a moment. What are you trying to do and what are you actually doing. Where are the values coming from for the variables you are removing from the output? If this is the only place that those variables are used do you think that you should remove the variables as well.

Megh05:
> Stop and think about this change for a moment. What are you trying to do and what are you actually doing. Where are the values coming from for the variables you are removing from the output? If this is the only place that those variables are used do you think that you should remove the variables as well.

I am extremely sorry for this silly mistake, I'll re-check the code again and implement it wisely.

brianteeman:
Dont be sorry. We all make mistakes - I just did on my own pr.

I didnt want to give you the full answer but I hope I have asked enough questions for you to find it for yourself

Megh05:
> Dont be sorry. We all make mistakes - I just did on my own pr.
> 
> I didnt want to give you the full answer but I hope I have asked enough questions for you to find it for yourself

Sure thing Sir... Thank You!

richard67:
@Megh05 Another thing which is not right with this PR here is that the issue was for Joomla 4, and so the change should also be made for Joomla 4, which belongs to the 4.0-dev branch here, but this PR is for the staging branch, which belongs to Joomla 3.9.

Megh05:
> @Megh05 Another thing which is not right with this PR here is that the issue was for Joomla 4, and so the change should also be made for Joomla 4, which belongs to the 4.0-dev branch here, but this PR is for the staging branch, which belongs to Joomla 3.9.

Totally understood, Thank you sir for your guidance!

alikon:
replaced by #32913
thanks for your contribution

PR #32672, opened 2021-03-13
[4.0] Extensions: Update Sites string

brianteeman:
Adds a translatable label for the Download Key filter
image


wilsonge:
Thanks!

brianteeman:
thanks

Issue #32679, opened 2021-03-14
[4.0] Template menu bar not properly aligned 


Megh05:
### What needs to be fixed

![image](https://user-images.githubusercontent.com/53462448/111058504-099a8280-84b5-11eb-91ed-886e729cbd32.png)

This is some extra white space on the top right corner of the mail template page

### Why this should be fixed

Good UI

### How would you fix it

By applying proper CSS

sanket9787:
Try this
 width=100% 
in css file
if error is still there please provide me the code I will cheque it

brianteeman:
I cannot replicate this

![image](https://user-images.githubusercontent.com/1296369/111062034-60e12880-849e-11eb-9380-dc727ba7ede9.png)

Quy:
Please test PR #32689

PR #32683, opened 2021-03-14
[4.0] fixes table overflow in Mail templates

sksuryan:
Pull Request for Issue #32679 .

### Summary of Changes
This PR fixes the issue where the table in Mailing templates would fail to be responsive because of long `ID`s since they are treated as a single word, causing an overflow. Hence, to fix this problem, `text-break` class for ID column has been added.

Further, this PR tweaks alignment of text in Mail templates table to visually make it better.

### Testing Instructions

- Login to Administrator.
- Go to System > Mail Templates.
- Use Dev tools to resize the window. The overflow could be seen at around 1024px.

image
image

The table should be responsive and be able to adapt to smaller widths.

image

### DIFF
--- a/administrator/components/com_mails/tmpl/templates/default.php
+++ b/administrator/components/com_mails/tmpl/templates/default.php
@@ -38,7 +38,7 @@ $listDirn = $this->escape($this->state->get('list.direction'));
 						</caption>
 						<thead>
 							<tr>
-								<th scope="col" class="w-25" style="min-width:100px">
+								<th scope="col" class="w-20" style="min-width:100px">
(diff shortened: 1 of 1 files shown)


brianteeman:
Text in columns should always be aligned left and never centered

sksuryan:
Okay, will revert this back.

sksuryan:
**EDIT**: I mentioned the wrong issue.

@brianteeman my bad. 😅

sksuryan:
Closing because of latest merged commits in 4.0-dev. Will make a new PR with the rebased branch.

PR #32688, opened 2021-03-14
[4.0] Fixes button margin

srishty-07:
Pull Request for Issue # .

### Summary of Changes
bootstrap margin class was added to the danger button

### Testing Instructions
1) open the cassiopeia page[frontend] on (http://localhost/joomla-cms/index.php?option=com_config&view=modules&id=1&Itemid=101&return=aHR0cDovL2xvY2FsaG9zdC9qb29tbGEtY21zLw%3D%3D)

2)open edit module on the right hand side of the page.

3) set the toggle device toolbar to mobile view.

4)before applying the patch it would show improper margin between the buttons, but after applying the patch proper margin is there.


brianteeman:
Would it not be better to do this in css with a media query by adding a margin-bottom to the .btn class

richard67:
Agree it would be better to do it in (s)css.

@srishty-07 Let us know if you need advise on which scss file to modify.

srishty-07:
@brianteeman @richard67 sir I thought of doing that , but as every button needed margin so just added  bootstrap margin to it, but if you say so I will add media queries in scss file. @richard67 it will be really kind of you if you please tell me about which scss file I have to make changes in.

richard67:
> @richard67 it will be really kind of you if you please tell me about which scss file I have to make changes in.

I have no idea myself right now 😊 Maybe @drmenzelit can advise on that.

drmenzelit:
The definitions for buttons in Cassiopeia are located in `templates/cassiopeia/scss/vendor/bootstrap/_buttons.scss`
Be careful, because changes here will change all buttons in the template, so you have to check the whole frontend to avoid side effects.
After doing changes in the scss file(s) you need to run `npm run build:css`

vyaki1718:
it would better adding margin using css
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32688">issues.joomla.org/tracker/joomla-cms/32688</a>.</sub>

srishty-07:
@vyaki1718 i have made the changes in css only in 2nd pr.

ceford:
In Article Edit view the Versions button needs attention too. Perhaps not in English but in any language where the button labels are longer.

screen shot 2021-03-15 at 09 25 58

So I wonder if the style should be added to the btn class rather than the individual btn-xxx classes.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32688">issues.joomla.org/tracker/joomla-cms/32688</a>.</sub>

brianteeman:
> So I wonder if the style should be added to the btn class rather than the individual btn-xxx classes.

That is exactly how it should have been done as its more robust code as you just discovered.

Also please use rem and not px

PR #32693, opened 2021-03-14
Disable Displays error messages about missing XML attributes addition…

korenevskiy:
…al for fields.

Operation of Joomla in the configuration mode "Error messages: Maximum".
Displays error messages about missing XML attributes for fields.
But the absence of these fields is acceptable for Joomla fields to work.
No error message should be displayed for optional attributes for Joomla in the "Error Messages: Maximum" mode, because the attributes are additional.
On the other hand, these error messages do not inform the administrator about the error, from these messages it is not at all clear what field we are talking about, what attribute we are talking about, and what the error is in general.
Messages appear when there is no **Label** attribute, or there is no **Name** attribute.

img-2021-03-14-23-54-21

### DIFF
--- a/libraries/src/Form/FormField.php
+++ b/libraries/src/Form/FormField.php
@@ -811,7 +811,7 @@ abstract class FormField
 		}
 
 		// Get the label text from the XML element, defaulting to the element name.
-		$title = $this->element['label'] ? (string) $this->element['label'] : (string) $this->element['name'];
+		$title = isset($this->element['label']) && $this->element['label'] ? (string) $this->element['label'] : (isset($this->element['name']) && $this->element['name'] ? (string) $this->element['name'] : '');
 		$title = $this->translateLabel ? Text::_($title) : $title;
 
 		return $title;
@@ -1314,8 +1314,7 @@ abstract class FormField
 	protected function getLayoutData()
 	{
 		// Label preprocess
-		$label = $this->element['label'] ? (string) $this->element['label'] : (string) $this->element['name'];
-		$label = $this->translateLabel ? Text::_($label) : $label;
+		$label = $this->getTitle();
 
 		// Description preprocess
 		$description = !empty($this->description) ? $this->description : null;


brianteeman:
The error messages can always be improved but it is not correct to change the code because a manadatory field is missing and you dont want to see an error message

From the [documentation](https://docs.joomla.org/Email_form_field_type)
- name (**mandatory**) is the unique name of the parameter.
- type (**mandatory**) must be field type.
- label (**mandatory**) (translatable) is the descriptive title of the field.

ceford:
It would be helpful to have a specific example of where this error appears - better test instructions. Just reading the diff leaves too much to do for a tester. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32693">issues.joomla.org/tracker/joomla-cms/32693</a>.</sub>

brianteeman:
@ceford This PR is plain wrong. The label and name are mandatory fields for very good reasons and therefore it is correct that an error is displayed. This PR seeks to ignore that and to remove the display of the error

korenevskiy:
> The error messages can always be improved but it is not correct to change the code because a manadatory field is missing and you dont want to see an error message
> 
> From the [documentation](https://docs.joomla.org/Email_form_field_type)
> 
> * name (**mandatory**) is the unique name of the parameter.
> * type (**mandatory**) must be field type.
> * label (**mandatory**) (translatable) is the descriptive title of the field.


@ceford 
For the **`SPACER`** field type, the **`LABEL`** attribute is **optional**. And the Form Field class in question is the base inherited class. This means that there is no error in my corrections.

rdeutz:
What you are doing here is to make name in some way an optional parameter. I agree that the error message can be better, but your way is wrong. What could be usefull is a checkField function, that checks is all necessary values are set. But this is a big task, because of the flexibility of the form class. Closing this.

PR #32702, opened 2021-03-16
4.0 dev

JajateeNandineeSahoo:
Pull Request for Issue # .


drmenzelit:
@JajateeNandineeSahoo can you please enter some information for this PR? What did you change, why, how to test...

dgrammatiko:
Please remove the javascript tests coverage
Also, the instructions are only for win-dos machines. I use MAC and others use Linux, so...

Edit: not to mention that the WAMP setup is very biased, XAMP when you can do the same with other tools or even docker which is way better than local setups

JajateeNandineeSahoo:
@drmenzelit I have added instructions for Windows user who download the Joomla zip file and are not able to set it up on their local machine. I have mentions few steps how they can extract it and able to install it on their machine without any difficulties. This is basically for Windows users and not for MAC or Linux.

brianteeman:
This should be in the documentation not in the repo

rdeutz:
closing this because close to nobody knows what this PR is about

Issue #32700, opened 2021-03-16
Add Button margins when update


mariantanase:
### Steps to reproduce the issue
When upgrading from Joomla 4 Beta 7 to the current nightly build, you are prompted to enter credentials before continuing.
The two buttons (**Cancel** and **Install**) are attached. No margin or padding between them.

### Actual result

![screen shot 2021-03-16 at 09 26 30](https://issues.joomla.org/uploads/1/5cda17c241681bfa9877125c3204a5ca.png)

### System information (as much as possible)
Joomla_4.0.0-beta8-dev
PHP 7.4.13

anonymous0807:
Sir, let me get a chance to fix this issue.

alikon:
maybe reading this one #32697  may help

richard67:
Whoever wants to fix this issue mind the round corners of the button: When having space between the buttons, it will need all 4 corners of each button to be round.

JajateeNandineeSahoo:
I want to fix the issue but don't know where is the code this issue. Could anyone help me and tell what is the path to get the code?

shitesh2607:
I am working on this issue.

infograf768:
Suggestion:
```
diff --git a/administrator/components/com_joomlaupdate/tmpl/upload/captive.php b/administrator/components/com_joomlaupdate/tmpl/upload/captive.php
index bc7b120..78bc0f9 100644
--- a/administrator/components/com_joomlaupdate/tmpl/upload/captive.php
+++ b/administrator/components/com_joomlaupdate/tmpl/upload/captive.php
@@ -85,12 +85,12 @@
 		<div class="control-group">
 			<div class="controls">
-				<div class="btn-group">
+				<button type="button" class="btn">
 					<a class="btn btn-danger" href="index.php?option=com_joomlaupdate">
 						<span class="icon-times icon-white" aria-hidden="true"></span> <?php echo Text::_('JCANCEL'); ?>
 					</a>
-					<button type="submit" class="btn btn-primary">
-						<span class="icon-play icon-white" aria-hidden="true"></span> <?php echo Text::_('COM_INSTALLER_INSTALL_BUTTON'); ?>
-					</button>
-				</div>
+				</button>
+				<button type="submit" class="btn btn-primary">
+					<span class="icon-play icon-white" aria-hidden="true"></span> <?php echo Text::_('COM_INSTALLER_INSTALL_BUTTON'); ?>
+				</button>
 			</div>
 		</div>
```
[...]

PR #32716, opened 2021-03-17
Update captive.php

mariantanase:
Add some space betweeb two buttons (CANCEL and INSTALL) when doing update.
See Issue [#32700]
Thanks @infograf768 

Pull Request for Issue #32700 .

buttons-attached

Screenshot_2021-03-17 Joomla Update - Joomla 4 Demo - Administration

### DIFF
--- a/administrator/components/com_joomlaupdate/tmpl/upload/captive.php
+++ b/administrator/components/com_joomlaupdate/tmpl/upload/captive.php
@@ -77,14 +77,14 @@ $twofactormethods = \Joomla\CMS\Helper\AuthenticationHelper::getTwoFactorMethods
 		<?php endif; ?>
 		<div class="control-group">
 			<div class="controls">
-				<div class="btn-group">
+				<button type="button" class="btn">
 					<a tabindex="4" class="btn btn-danger" href="index.php?option=com_joomlaupdate">
 						<span class="icon-cancel icon-white" aria-hidden="true"></span> <?php echo JText::_('JCANCEL'); ?>
 					</a>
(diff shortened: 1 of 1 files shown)


infograf768:
Looks like you created the pr based on an older version of 4.0-dev . in fact a beta4.
This needs correcting.

dgrammatiko:
This is totally wrong, you cannot nest a link inside a button element, check https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button permitted parents.

infograf768:
@dgrammatiko 
you are correct.

infograf768:
@mariantanase 
I'm afraid this PR is wrong (my proposal in https://github.com/joomla/joomla-cms/issues/32700 was no good) and as it is also based on an obsolete 4.0-dev, I suggest you close it.
When done we will reopen https://github.com/joomla/joomla-cms/issues/32700

infograf768:
@dgrammatiko 
Would using an onclick be OK?
```
		<div class="control-group">
			<div class="controls">
				<button type="button" class="btn btn-danger" onclick="window.location.href='index.php?option=com_joomlaupdate';">
					<span class="icon-times icon-white" aria-hidden="true"></span> <?php echo Text::_('JCANCEL'); ?>
				</button>
				<button type="submit" class="btn btn-primary">
					<span class="icon-play icon-white" aria-hidden="true"></span> <?php echo Text::_('COM_INSTALLER_INSTALL_BUTTON'); ?>
				</button>
			</div>
		</div>
```
Screen Shot 2021-03-18 at 08 38 47

PR #32740, opened 2021-03-18
[4.0] Introduce an HTML sanitiser

dgrammatiko:
Pull Request for Issue # .

This [when applied to all the scripts using innerHTML] fixes XSS vulnerabilities, so although it could be categorised as a new feature actually it's a feature to strengthen the security, so I guess an exception should NOT be debatable here. (then again I might be wrong, won't be the first time)

### Summary of Changes
- Since the build tools allow us now to utilise properly the `import` of the modern JS we will do exactly that!!!
- Use the Bootstrap sanitiser
- Expose it as `Joomla.sanitizeHtml`
- Patch the `Joomla.renderMessages` as an example case

There is a slight penalty since we're adding code here, but I think it's acceptable, benefits outweigh the performance here:

Without the sanitiser:

Screenshot 2021-03-18 at 22 08 50

With:

Screenshot 2021-03-18 at 22 07 07

 
### Testing Instructions
- Either download the installable zip from the PR's github page or apply the PR using git
- Try to login using wrong username or password
- There should be an alert warning you that username and password were incorrect
- Disable the Smart Search and all the plugins
- Check the components->Smart Search -> Index page (the page should have an alert with a link)
- That's it

The bootstrap documentation: https://getbootstrap.com/docs/5.0/getting-started/javascript/#sanitizer

Documenting ALL the core.js functions should be done at some point

BUT more importantly, if this PR is approved and merged, there are countless instances of `innerHTML` that will have to be patched.

@wilsonge @zero-24 @SniperSister

### DIFF
--- a/build/media_source/system/js/core.es6.js
+++ b/build/media_source/system/js/core.es6.js
@@ -3,6 +3,8 @@
  * @license    GNU General Public License version 2 or later; see LICENSE.txt
  */
 
+import { sanitizeHtml } from 'bootstrap/js/src/util/sanitizer.js';
+
 // Only define the Joomla namespace if not defined.
 window.Joomla = window.Joomla || {};
 
@@ -620,6 +622,17 @@ window.Joomla.Modal = window.Joomla.Modal || {
   };
 
   /**
+   * Sanitize HTML string
+   *
+   * @param {string} unsafeHtml
+   * @param {array} allowList
+   * @param {function} sanitizeFn
+   *
+   * @return string
+   */
+  Joomla.sanitizeHtml = sanitizeHtml;
+
+  /**
    * Render messages send via JSON
    * Used by some javascripts such as validate.js
    *
@@ -685,7 +698,7 @@ window.Joomla.Modal = window.Joomla.Modal || {
       if (typeof title !== 'undefined') {
         titleWrapper = document.createElement('div');
         titleWrapper.className = 'alert-heading';
-        titleWrapper.innerHTML = `<span class="${type}"></span><span class="visually-hidden">${Joomla.Text._(type) ? Joomla.Text._(type) : type}</span>`;
+        titleWrapper.innerHTML = `<span class="${type}"></span><span class="visually-hidden">${Joomla.Text._(type) ? Joomla.Text._(sanitizeHtml(type)) : type}</span>`;
         messagesBox.appendChild(titleWrapper);
       }
 
@@ -693,7 +706,7 @@ window.Joomla.Modal = window.Joomla.Modal || {
       messageWrapper = document.createElement('div');
       messageWrapper.className = 'alert-wrapper';
       typeMessages.forEach((typeMessage) => {
-        messageWrapper.innerHTML += `<div class="alert-message">${typeMessage}</div>`;
+        messageWrapper.innerHTML += `<div class="alert-message">${sanitizeHtml(typeMessage)}</div>`;
       });
       messagesBox.appendChild(messageWrapper);
 
--- a/build/media_source/system/js/core.es6.js
+++ b/build/media_source/system/js/core.es6.js
@@ -5,6 +5,43 @@
 
 import { sanitizeHtml } from 'bootstrap/js/src/util/sanitizer.js';
 
+const ARIA_ATTRIBUTE_PATTERN = /^aria-[\w-]*$/i;
+const DATA_ATTRIBUTE_PATTERN = /^data-[\w-]*$/i;
+
+const DefaultAllowlist = {
+  // Global attributes allowed on any supplied element below.
+  '*': ['class', 'dir', 'id', 'lang', 'role', ARIA_ATTRIBUTE_PATTERN, DATA_ATTRIBUTE_PATTERN],
+  a: ['target', 'href', 'title', 'rel'],
+  area: [],
+  b: [],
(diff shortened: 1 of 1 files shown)


zero-24:
This sounds to me like a very good step into the right direction. :+1:

richard67:
I’m a bit scared of the countless instances to be patched later.

dgrammatiko:
BTW someone needs to patch the Joomla-browser thing used in codeception as we had to inject jQuery just to have the test pass...

richard67:
> 
> 
> BTW someone needs to patch the Joomla-browser thing used in codeception as we had to inject jQuery just to have the test pass...

Is that the reason why system tests are failing for this PR?

dgrammatiko:
> Is that the reason why system tests are failing for this PR?

Nope, this is probably a correct failure, checking rn

richard67:
> > Is that the reason why system tests are failing for this PR?
> 
> Nope, this is probably a correct failure, checking rn

Here the last action in the system test log:

```
 I upload file "com_media/test-image-1.png"
   I see element in dom "input[name='file']"
   I attach file "input[name='file']","com_media/test-im..."
 I see and close system message "Item uploaded."
   I get config "timeout"
   I wait for text "Item uploaded.",90,{"id":"system-mes...}
```

I.e. it uploads an image in media manager and then waits for the message "Item uploaded."

That could be the Joomla.renderMessages failing.

Maybe your sanitizer found an insane html and sanitized it a bit too much? 😉

dgrammatiko:
> Maybe your sanitizer found an insane html and sanitized it a bit too much?

No, it was far simpler: there wasn't anything in the alerts 😂

Should be ok now

dgrammatiko:
@brianteeman @infograf768 can you help here defining the allowed list of elements/attributes? Thanks

brianteeman:
Happy to help but not sure I understand the question. Are the text-filters from global config useful?

brianteeman:
OR perhaps this from tinymce https://www.tiny.cloud/docs/configure/content-filtering/#extended_valid_elements

dgrammatiko:
> Happy to help but not sure I understand the question

So the renderMessage is a JS function in the core.js script and it accepts an object like `{error: ['some HTML text']`

We need to define what we accept for the `some HTML text`, rn I just added `div` and `span` but probably we need `p`, `a`
Basically, the list would be derived from the ini files

dgrammatiko:
@brianteeman so, I opted to use Bootsrap's default rules, if we need another set of rules we can simply compose them per case. Allowed list available here: https://github.com/twbs/bootstrap/blob/df2d5b6a37b54ff967edfa21a630cd1ecd223f7c/js/src/util/sanitizer.js#L58-L90

rdeutz:
I think this should go into 4.0.0, hopefully it is RTC soon :-)

SniperSister:
@dgrammatiko 

> so, I opted to use Bootsrap's default rules

I'm wondering if the default set misses 

`<a href="javascript:[...]"`

as an attack vector?

dgrammatiko:
> I'm wondering if the default set misses

It's covered due to https://github.com/twbs/bootstrap/blob/df2d5b6a37b54ff967edfa21a630cd1ecd223f7c/js/src/util/sanitizer.js#L26 and the rule `  a: ['target', 'href', 'title', 'rel'],`

Screenshot 2021-03-22 at 10 41 02

Edit: it goes without saying that some of our current code will fail to pass sanitization, but that's something that we have to deal on a per case approach

ceford:
So if documentation is required where would we put it? We could link something from the Article Edit Help screen. Or do we need something aimed at coders? Tooltips and Popovers use the built-in sanitizer but right now ordinary users cannot use them (can they?). Something I have been working on today. Just noting interest in documentation!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32740">issues.joomla.org/tracker/joomla-cms/32740</a>.</sub>

dgrammatiko:
> Or do we need something aimed at coders

This affects largely the coders, we need a section for Javascript and then some subsections for the API, Bootstrap, etc...

build/media_source/system/js/core.es6.js
```diff
@@ -619,6 +658,21 @@ window.Joomla.Modal = window.Joomla.Modal || {
     return xhr;
   };
 
+  /**
+   * Sanitize HTML string
+   *
+   * @param {string} unsafeHtml The html for sanitization
+   * @param {object} allowList The list of HTMLElements with an array of allowed attributes
+   * @param {function} sanitizeFn A custom sanitization function
```
brianteeman, line 666:
```suggestion
   * @param {string}     unsafeHtml The html for sanitization
   * @param {object}    allowList      The list of HTML elements with an array of allowed attributes
   * @param {function} sanitizeFn    A custom sanitize function
```

brianteeman:
```suggestion
   * @param {string}   unsafeHtml The html to be sanitized
   * @param {object}   allowList  The list of HTML elements with an array of allowed attributes
   * @param {function} sanitizeFn A custom sanitize function
```

dgrammatiko:
This has been merge with #32747 as the latter was based on this PR.

Issue #30018, opened 2020-07-06
[4.0] Two icons on some buttons


C-Lodder:
### Steps to reproduce the issue

1. Go to `/administrator/index.php?option=com_installer&view=install`
2. Go to the `Install from Web` tab
3. Click any extension

### Actual result

There are 2 icons on the buttons.

![screeny](https://user-images.githubusercontent.com/2019801/86569607-8eb4bb80-bf66-11ea-99a8-57cddf5a0bdc.png)

brianteeman:
Thats because they are links to external websites that open in a new window and we have a css rule that matches that

C-Lodder:
So multiple icons are expected?

brianteeman:
i am just explaining wht it happens. Not commenting on if it is correct

Quy:
Please test PR #30148

Akashrajpoots14:
Hi, Sir, I want to work on this issue

ghost:
@Akashrajpoots14 you can make a pull request without allowing by another.

infograf768:
It is not possible to correct this directly in core.

chmst:
I expect that this rule is not applied on buttons.

brianteeman:
> It is not possible to correct this directly in core.

Yes it is. It is just changing the css rule

ghost:
I want to work on this issue

gaurav24072002:
Hey ! I am interested to solve this issue  but when I go to the required instruction everything looks fine to me . @C-Lodder Please see the screenshot:
![gsoc1](https://user-images.githubusercontent.com/77109758/111348399-2a75fa00-86a6-11eb-8044-3a244b5331ee.JPG)

If you want to say something else please can you define specifically...😅
And I would be more then happy to solve the issue...

himanshu007-creator:
> Hey ! I am interested to solve this issue but when I go to the required instruction everything looks fine to me . @C-Lodder Please see the screenshot:
> ![gsoc1](https://user-images.githubusercontent.com/77109758/111348399-2a75fa00-86a6-11eb-8044-3a244b5331ee.JPG)
> 
> If you want to say something else please can you define specifically...😅
> And I would be more then happy to solve the issue...

looks good @gaurav24072002 , have a look <a href="https://github.com/joomla/joomla-cms/pull/30148">HERE</a>.  Working on this issue is simpler and it would be good to see you work on it😉.
[...]

PR #32743, opened 2021-03-19
Temp#30018 1

NikitaEmberi:
Pull Request for Issue #30018 

### Summary of Changes

modified _icon.scss to move the external link icon after the text of the button.

image

image
(If I try to add padding of more than 2px, it squashes the alignment of buttons like this:
image

so that is why I added a padding of only 2px. )

### DIFF
--- a/administrator/templates/atum/scss/blocks/_icons.scss
+++ b/administrator/templates/atum/scss/blocks/_icons.scss
@@ -63,7 +63,7 @@
     cursor: not-allowed;
     opacity: 1;
   }
-
+  npm run build:css
   &.disabled .#{$jicon-css-prefix}-home,
   &.disabled .#{$fa-css-prefix}-home {
     color: $state-warning-bg;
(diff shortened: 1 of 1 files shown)


administrator/templates/atum/scss/blocks/_icons.scss
```diff
@@ -63,7 +63,7 @@
     cursor: not-allowed;
     opacity: 1;
   }
-
+  npm run build:css
```
drmenzelit, line 66: That doesn't belong in the scss file, please remove

administrator/templates/atum/scss/blocks/_icons.scss
```diff
@@ -85,3 +85,8 @@
 .plg_system_webauthn_login_button svg path {
   fill: var(--white);
 }
+
+a[target=_blank]:before {
+  float: right;
+  padding-left: 2px;
+}
```
drmenzelit, line 92: Add a new line at the end. The code styling test is failing

PR #32754, opened 2021-03-19
[4.0] Fix wrong info for search in fields

chmst:
Pull Request for Issue # .

### Summary of Changes
In fields, there is no fieldname and no author.

### Testing Instructions
See the view com_fields 
grafik

Compare before and after patch.


joomdonation:
Actually, we have Name for each custom field. Author is the user who created that custom field. The text is fine as it is. See this block of code https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_fields/src/Model/FieldsModel.php#L354-L388 if you want to understand how search is performed for custom fields

@SharkyKZ You are usually right with thumb down. Instead of being silent with thumb down, could you please help further by explain what's wrong?

PR #32764, opened 2021-03-20
checkbox placed inline with the label in login model

rjharishabh:
### Summary of Changes

Checkbox placed inline with the label in login model

check_old

check_new

### DIFF
--- a/components/com_users/tmpl/login/default_login.php
+++ b/components/com_users/tmpl/login/default_login.php
@@ -61,9 +61,7 @@ $usersConfig = ComponentHelper::getParams('com_users');
 			<?php if (PluginHelper::isEnabled('system', 'remember')) : ?>
 				<div  class="com-users-login__remember control-group">
 					<div class="control-label">
-						<span class="controls">
 							<input id="remember" type="checkbox" name="remember" class="inputbox" value="yes">
-						</span>
 						<label for="remember">
 							<?php echo Text::_('COM_USERS_LOGIN_REMEMBER_ME'); ?>
 						</label>
--- a/components/com_users/tmpl/login/default_login.php
+++ b/components/com_users/tmpl/login/default_login.php
@@ -60,13 +60,14 @@ $usersConfig = ComponentHelper::getParams('com_users');
 
 			<?php if (PluginHelper::isEnabled('system', 'remember')) : ?>
 				<div  class="com-users-login__remember control-group">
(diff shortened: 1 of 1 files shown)


brianteeman:
This is a valid issue but the proposal is obviously not correct as the classes are not applied correctlty

rjharishabh:
> This is a valid issue but the proposal is obviously not correct as the classes are not applied correctlty

are you talking about 'controls' class

saumyasarkar11:
> > This is a valid issue but the proposal is obviously not correct as the classes are not applied correctlty
> 
> are you talking about 'controls' class

Remove the checkbox from the div with class "contents-label" and create a specific div for it with class ''controls" and then wrap both the divs in one row. Should be alright.

rjharishabh:
> Remove the checkbox from the div with class "contents-label" and create a specific div for it with class ''controls" and then wrap both the divs in one row. Should be alright.

can I use span instead of div

saumyasarkar11:
> > Remove the checkbox from the div with class "contents-label" and create a specific div for it with class ''controls" and then wrap both the divs in one row. Should be alright.
> 
> can I use span instead of div

It's better to keep the code as original as possible. Avoid using span

saumyasarkar11 (changes requested):
I tried to apply this patch but it said : The patch could not be applied because it conflicts with a previously applied patch: components/com_users/tmpl/login/default_login.php

rjharishabh:
> I tried to apply this patch but it said : The patch could not be applied because it conflicts with a previously applied patch: components/com_users/tmpl/login/default_login.php

can you please tell me more about it

drmenzelit:
Keep it simple with correct HTML structure. Probably you don't need a div nor a span element. The `control` classes are remains from Bootstrap 2 used in Joomla 3.

rjharishabh:
> Keep it simple with correct HTML structure. Probably you don't need a div nor a span element. The `control` classes are remains from Bootstrap 2 used in Joomla 3.

Thanks

saumyasarkar11:
I think you should create an issue first regarding this. What say? @rjharishabh

rjharishabh:
> I think you should create an issue first regarding this. What say? @rjharishabh

Yes sure, I will create an issue for this.

drmenzelit:
You don't need to create an issue, you can make a PR directly

PR #32769, opened 2021-03-21
[4.0][webservice][com_installer] add filtering of data

alikon:
### Summary of Changes
add filtering of data

### Testing Instructions

- GET {{base_path}}/api/index.php/v1/installer/manage?core=true/false
- GET {{base_path}}/api/index.php/v1/installer/manage?status=0/1
- GET {{base_path}}/api/index.php/v1/installer/manage?type=component/plugin/module/template

N/A

```json
{
    "links": {
        "self": "https://demo4.loc/api/index.php/v1/installer/manage?core=0"
    },
    "data": [
        {
            "type": "manage",
            "id": "224",
            "attributes": {
                "name": "Action Log - Akeeba Backup",
                "type": "plugin",
                "folder": "actionlog",
                "client_id": "0",
                "status": "0",
                "version": "8.0.1",
                "id": "224"
            }
        },
        {
            "type": "manage",
            "id": "220",
            "attributes": {
                "name": "Akeeba Backup",
                "type": "component",
                "folder": "",
                "client_id": "1",
                "status": "1",
                "version": "8.0.1",
                "id": "220"
            }
        },
        {
            "type": "manage",
            "id": "225",
            "attributes": {
                "name": "Akeeba Backup package",
                "type": "package",
                "folder": "",
                "client_id": "0",
                "status": "1",
                "version": "8.0.1",
                "id": "225"
            }
        },
        {
            "type": "manage",
            "id": "221",
            "attributes": {
                "name": "Console - Akeeba Backup",
                "type": "plugin",
                "folder": "console",
                "client_id": "0",
                "status": "1",
                "version": "8.0.1",
                "id": "221"
            }
        },
        {
            "type": "manage",
            "id": "226",
            "attributes": {
                "name": "file_fef",
                "type": "file",
                "folder": "",
                "client_id": "0",
                "status": "1",
                "version": "2.0.1",
                "id": "226"
            }
        },
        {
            "type": "manage",
            "id": "219",
            "attributes": {
                "name": "file_fof40",
                "type": "file",
                "folder": "",
                "client_id": "0",
                "status": "1",
                "version": "4.0.2",
                "id": "219"
            }
        },
        {
            "type": "manage",
            "id": "218",
            "attributes": {
                "name": "Joomla! Patch Tester",
                "type": "component",
                "folder": "",
                "client_id": "1",
                "status": "1",
                "version": "4.0.1",
                "id": "218"
            }
        },
        {
            "type": "manage",
            "id": "222",
            "attributes": {
                "name": "Quick Icon - Akeeba Backup Notification",
                "type": "plugin",
                "folder": "quickicon",
                "client_id": "0",
                "status": "1",
                "version": "8.0.1",
                "id": "222"
            }
        },
        {
            "type": "manage",
            "id": "217",
            "attributes": {
                "name": "Sample Data - Testing",
                "type": "plugin",
                "folder": "sampledata",
                "client_id": "0",
                "status": "0",
                "version": "4.0.0",
                "id": "217"
            }
        },
        {
            "type": "manage",
            "id": "223",
            "attributes": {
                "name": "System - Backup on update",
                "type": "plugin",
                "folder": "system",
                "client_id": "0",
                "status": "0",
                "version": "8.0.1",
                "id": "223"
            }
        }
    ],
    "meta": {
        "total-pages": 1
    }
}
```

### DIFF
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -46,7 +46,8 @@ class ManageController extends ApiController
 	public function displayList()
 	{
 		$requestBool = $this->input->get('core', $this->input->get->get('core'));
-		if (($requestBool === 'true') || ($requestBool === 'false'))
+
+		if ($requestBool === 'true' || $requestBool === 'false')
 		{
 			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
 		}
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -47,7 +47,7 @@ class ManageController extends ApiController
 	{
 		$this->modelState->set('filter.core', $this->input->get('core', $this->input->get->get('core')), 'INT');
 		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
-		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'INT');
+		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'STRING');
 
 		return parent::displayList();
 	}
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -46,7 +46,6 @@ class ManageController extends ApiController
 	public function displayList()
 	{
 		$requestBool = $this->input->get('core', $this->input->get->get('core'));
-		
 		if (($requestBool === 'true') || ($requestBool === 'false'))
 		{
 			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -45,7 +45,13 @@ class ManageController extends ApiController
 	 */
 	public function displayList()
 	{
-		$this->modelState->set('filter.core', $this->input->get('core', $this->input->get->get('core')), 'INT');
+		$requestBool = $this->input->get('core', $this->input->get->get('core'));
+		
+		if (($requestBool === 'true') || ($requestBool === 'false'))
+		{
+			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
+		}
+
 		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
 		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'STRING');
 
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -11,7 +11,10 @@ namespace Joomla\Component\Installer\Api\Controller;
 
 \defined('_JEXEC') or die;
 
+use Joomla\CMS\Filter\InputFilter;
+use Joomla\CMS\Language\Text;
 use Joomla\CMS\MVC\Controller\ApiController;
+use Tobscure\JsonApi\Exception\InvalidParameterException;
 
 /**
  * The manage controller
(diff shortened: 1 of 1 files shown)


api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -35,4 +35,20 @@ class ManageController extends ApiController
...
+	/**
+	 * Extension list view amended to add filtering of data
+	 *
+	 * @return  static  A BaseController object to support chaining.
+	 *
+	 * @since   4.0.0
+	 */
+	public function displayList()
+	{
+		$this->modelState->set('filter.core', $this->input->get('core', $this->input->get->get('core')), 'INT');
```
wilsonge: Whilst this is an int in the db - this is just a boolean I think? if so the api should expose it as a boolean true/false

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -35,4 +35,20 @@ class ManageController extends ApiController
...
+		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
+		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'INT');
```
wilsonge: This should be a string filter based on your example

api/components/com_installer/src/Controller/ManageController.php
alikon: we already  use  INT for plugin for example `{{base_path}}/api/index.php/v1/plugins?filter[status]=0`
or for content

wilsonge: Yes but that's not boolean because there's multiple different state types (published/unpublished/trashed etc)

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -35,4 +35,26 @@ class ManageController extends ApiController
...
+	 * Extension list view amended to add filtering of data
+	 *
+	 * @return  static  A BaseController object to support chaining.
+	 *
+	 * @since   4.0.0
+	 */
+	public function displayList()
+	{
+		$requestBool = $this->input->get('core', $this->input->get->get('core'));
+
```
richard67:
```suggestion

```

richard67: Remove just the spaces, not the complete line.

Quy: @richard67 To prevent GitHub from removing the line, select lines 48-50, and then remove the spaces in line 49.

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -35,4 +35,25 @@ class ManageController extends ApiController
...
+	 * Extension list view amended to add filtering of data
+	 *
+	 * @return  static  A BaseController object to support chaining.
+	 *
+	 * @since   4.0.0
+	 */
+	public function displayList()
+	{
+		$requestBool = $this->input->get('core', $this->input->get->get('core'));
+		if (($requestBool === 'true') || ($requestBool === 'false'))
```
Quy, line 49:
```suggestion
		$requestBool = $this->input->get('core', $this->input->get->get('core'));

		if (($requestBool === 'true') || ($requestBool === 'false'))
```

Quy:
```suggestion
		$requestBool = $this->input->get('core', $this->input->get->get('core'));

		if ($requestBool === 'true' || $requestBool === 'false')
```

api/components/com_installer/src/Controller/ManageController.php
richard67: ah yes, thanks. silly me. I always forget that.

api/components/com_installer/src/Controller/ManageController.php
alikon: it's ok now i guess

Quy:
For `core`, should all entries be displayed if true/false is not entered?

alikon:
when core is omitted it works like before ie all items are returned

Quy:
`core=0` returns entries. Should it?

alikon:
no
better a 400 

```json
{
    "errors": [
        {
            "title": "Invalid field: core",
            "code": 400
        }
    ]
}
```

Quy:
It is not a consistent to show an error vs. the following.

In this example, a 2 is added at the end.

```
{
    "links": {
        "self": "http://localhost/joomla-cms-4.0-dev/api/index.php/v1/installer/manage?type=component2"
    },
    "data": [],
    "meta": {
        "total-pages": 0
    }
}
```

alikon:
these are different things
it has been requested to manage `core` like boolean  hence the check (invalid field)
but for `type` it's a string  so 0 found it's a correct response

anibalsanchez:
I think that the GET verb is wrong in the REST API controller **Installer**.

**/installer** is not a data entity. It is a controller/helper itself that provides the behaviour, **actions**. 

It looks odd calling an action with a GET. /installer/manage is a method, so for clarity, it is better if it is called with a POST (even if it doesn't persist a change). 

So, if everything that Installer does is managing, then the method manage must return different types of data based on a control parameter. If we follow the Joomla practice, it should have a task parameter to return data or persist changes.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32769">issues.joomla.org/tracker/joomla-cms/32769</a>.</sub>

alikon:
beyond the scope of this pr

anibalsanchez:
Thinking twice about the API for extensions, it is not right to publish the "Installer" /installer. It is not the right semantic.

Check the documentation https://docs.joomla.org/J4.x:Joomla_Core_APIs. The entry points are always entities.

It is better to manage the extensions as **/api/index.php/v1/extensions/...**

For instance, to install an extension you should do a `POST /api/index.php/v1/extensions/install`.

wilsonge:
@anibalsanchez i agree with you - however Nicola is also correct that his pr is just adding filters to an existing endpoint - so can you create a separate issue to track your comment - and I'll try and work on it

alikon:
merged in #33339

PR #32774, opened 2021-03-21
[4] Missing?

astridx:
Pull Request for Issue https://github.com/joomla/joomla-cms/discussions/32760

### DIFF
--- a/plugins/editors/codemirror/codemirror.xml
+++ b/plugins/editors/codemirror/codemirror.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <extension type="plugin" group="editors" method="upgrade">
 	<name>plg_editors_codemirror</name>
-	<version>5.59.4</version>
+	<version>5.60.0</version>
 	<creationDate>28 March 2011</creationDate>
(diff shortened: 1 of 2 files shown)


brianteeman:
This is not correct.
Those files are updated by the build scripts.
So you need to have updated the packages defined in npm in order to have those versions.

Therefore you need to commit the updated package.json and package-lock.json as well

astridx:
package.json and package-lock.json are already updated. For example:

https://github.com/joomla/joomla-cms/blob/c78ef2fb45f68c91d87aa553532dceb4d66478db/package.json#L58

brianteeman:
Look again;)

astridx:
hm, this appeared for me after I ran npm ci. This is probably an error on my part.

dgrammatiko:
@astridx this is a legit PR, can you reopen it?
@richard67 can you merge this on review? (explanation what happed here: we updated the package.json *without* running `npm install` so not all changes were applied)

Ignore me, the current versions are aligned

brianteeman:
@dgrammatiko you are wrong.

As I explained - if you run npm install then it will update these packaged and the xml files. You then have to commit these two xml AND the package.json and package-lock.json

PR #32777, opened 2021-03-21
[4.0] Introduce inlined and lazy attributes to Web Assets

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
Two more options were added to the assets definition: 
- inline: will inline the source of a js or stylesheet
- lazy: will render a `<link href="path/to/file" rel=print class=js-lazy-loaded data-rel=stylesheet />` but will **not** handle the lazyloding js part (eg switch the rel value to whatever is in the `data-rel` )

### Testing Instructions
Apply the PR and check the admin login and dashboard pages (the source and the DOM)

Lazy load a stylesheet
Screenshot 2021-03-21 at 16 56 50

Screenshot 2021-03-21 at 16 57 09

Inlining a js script `"inlined": true` in the joomla.asset.json
Screenshot 2021-03-21 at 16 57 20

Inlining a stylesheet
`$wa->registerAndUseStyle('f-a-1', 'administrator/templates/atum/css/vendor/fontawesome-free/fontawesome.css', ['inlined' => true], [], []);`

Screenshot 2021-03-21 at 16 57 36

# DO NOT MERGE THIS

@Fedik your input here?

### DIFF
--- a/administrator/templates/atum/index.php
+++ b/administrator/templates/atum/index.php
@@ -64,9 +64,9 @@ $wa->usePreset('template.atum.' . ($this->direction === 'rtl' ? 'rtl' : 'ltr'))
 
 // Override 'template.active' asset to set correct ltr/rtl dependency
 $wa->registerStyle('template.active', '', [], [], ['template.atum.' . ($this->direction === 'rtl' ? 'rtl' : 'ltr')]);
-//registerAndUseStyle(WebAssetItem|string $asset, string $uri = '', $options = [], $attributes = [], $dependencies = [])
-$wa->registerAndUseStyle('f-a-1', 'administrator/templates/atum/css/vendor/fontawesome-free/fontawesome.css', ['inlined' => true], [], []);
-$wa->useScript('core-inline');
+
+//$wa->getRegistry()->get('script', 'core')->setOption('inline', true);
+$wa->getRegistry()->get('style', 'fontawesome')->setOption('lazy', true);
 
 // Set some meta data
 $this->setMetaData('viewport', 'width=device-width, initial-scale=1');
@@ -77,7 +77,6 @@ HTMLHelper::getServiceRegistry()->register('atum', 'JHtmlAtum');
 HTMLHelper::_('atum.rootcolors', $this->params);
 
 Text::script('TPL_ATUM_MORE_ELEMENTS');
-
 ?>
 <!DOCTYPE html>
 <html lang="<?php echo $this->language; ?>" dir="<?php echo $this->direction; ?>"<?php echo $a11y_font ? ' class="a11y_font"' : ''; ?>>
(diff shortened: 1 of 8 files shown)


richard67:
@dgrammatiko System test is failing at the same place all the time, doesn't look like the usual unrelated issues. Could you check https://ci.joomla.org/joomla/joomla-cms/40988/1/22 ?

dgrammatiko:
@richard67 this is just a working POC. I'm not expecting all changes here to land as a single PR. Basically what matters here are the changes in the stylesRenderer and scriptsRenderer, everything else is just code that bothers me (eg scripts to bottom) or just code to showcase that this is actually working.

dgrammatiko:
https://github.com/joomla/joomla-cms/pull/32777/commits/4c43cb308c2cecf7603f8be83ba348789cfac569 is a shorter more focus version. The main point here is https://github.com/joomla/joomla-cms/pull/32777/commits/4c43cb308c2cecf7603f8be83ba348789cfac569#diff-2ede9f6c2a9f8106b0c8ccb140ba0c53999c95e84f8f732702ad700589ad7bdf and https://github.com/joomla/joomla-cms/pull/32777/commits/4c43cb308c2cecf7603f8be83ba348789cfac569#diff-16733093e0a3baea0a441e3c42154e63cb2a9e01c4738992d79839e2c3d2e715

dgrammatiko:
@richard67 tests are passing but people seem against this PR 🤣

richard67:
> ... but people seem against this PR

@dgrammatiko Thumb down from Sharky and ... from yourself? The latter confuses me a bit.

dgrammatiko:
> The latter confuses me a bit.

I'm sick and tired of getting a thumb down on every PR I do so I thought I should follow the trend 😂

richard67:
> I'm sick and tired of getting a thumb down on every PR I do so I thought I should follow the trend 😂

Ah, then it's ok. I worried a bit you could develop a dissociative identity disorder 😉

Fedik:
hold on, here is much of everything :smile: 

Please remove `inlined` option, it a questionable topic, it is better to do separately.
Please revert changes for `XHTML`, as long as Document may render XML, it should stay.
And revert "conditional", I know it is old, but as long as it not used, it does not hurt to have, and may be still useful for someone. It is better to do separately.

**Now about the lazy CSS**
I would suggest to do it with help of attributes only, does not need extra option `lazy`.
We only need to fix that `StylesRenderer` to accept a custom `rel` attribute

Check this example:
```html
<link rel="lazy-stylesheet" href="styles.css">

<script type="module">
document.head.querySelectorAll('link[rel="lazy-stylesheet"]').forEach(function($link){
  $link.rel = "stylesheet";
})
</script>
```

So in the Style asset need to add attribute `rel:"lazy-stylesheet"`, and that all.

I thought we can combine it with `data-defer` similar to scripts, but now I do not think that it makes sense.
Custom `rel` attribute should be enough.

dgrammatiko:
> I would suggest to do it with help of attributes only, does not need extra option lazy.

I like it, so if we agree on this then I will add the code for this (only the lazy-stylesheet) in the other PR.

About the other changes: at some point, we have to discuss what we deliver, conditionals are not supported in any of the Joomla 4 supported browsers and XHTML is dead since 2012 so I have no clue why we have to have those conditionals there.
I did those changes just to (hopefully) open the discussion

About the inlining part, this is something that the CMS should provide as it's easy to code and quite helpful in many instances

In short, I'll take the lazy-stylesheet to the other PR but it would be nice  to have some comments on the other issues I raised here

PR #32779, opened 2021-03-21
[4] s/Existant/Existent

PhilETaylor:
s/Existant/Existent

Although a function name, not strictly a b/c issue, as its a Unit Test so the method name is not important and is only descriptive


richard67:
Merging by review with respect of tests passing. Thanks!

PR #32782, opened 2021-03-21
[4] s/Negotation/Negotiation

PhilETaylor:
s/Negotation/Negotiation 

Changed method name but its a unit test which is descriptive and so not a b/c break


richard67:
Merging by review with respect of tests passing. Thanks!

PR #32806, opened 2021-03-22
[4][Tests] Tests without the need of jQuery

astridx:
Pull Request for Issue # https://github.com/joomla/joomla-cms/discussions/32766

### Summary of Changes
With this change, the tests run for me without jQuery.

> There is an error when renaming a folder in the Media Manager. But this has nothing to do with jQuery.


dgrammatiko:
Thank you @astridx for this one. Can you remove the jQuery also here: https://github.com/joomla/joomla-cms/blob/bb709bce4cebd0e10379ca1be3283ef95e533239/administrator/components/com_fields/tmpl/fields/default.php#L23-L24

dgrammatiko:
I have tested this item :white_check_mark: successfully on 94d0bc0f7f3bab4b880881a30f1173e03a437fcf<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32806">issues.joomla.org/tracker/joomla-cms/32806</a>.</sub>

astridx:
When I run all the tests, I get an error. I have attached the HMTL-log here. The error occurs when testing the renaming of a folder in the Media Manager. In my case, the folder is not created before renaming. If I add the creation of the folder before the test, then all tests run without errors.

I am wondering why this error does not occur here in the Git environment. Can someone tell me what is different for me than here in the git environment?

[systeminfo-2021-03-22T19 19 09+00 00.txt](https://github.com/joomla/joomla-cms/files/6184936/systeminfo-2021-03-22T19.19.09%2B00.00.txt)

[report.txt](https://github.com/joomla/joomla-cms/files/6184946/report.txt)

dgrammatiko:
> Can someone tell me what is different for me than here in the git environment?

Most of the times these are permission related

astridx:
> > Can someone tell me what is different for me than here in the git environment?
> 
> Most of the times these are permission related

Thank you very much. You were quick. 
However, I do not think, that it is a permissions problem because it is only one test with this problem and because I can create the folder in the same test in a different way.

I'm sorry. I made a mess. I deleted my comment here because I thought it was more appropriate in the discussion area.

https://github.com/joomla/joomla-cms/discussions/32811

HLeithner:
Thanks @astridx

PR #32826, opened 2021-03-23
[4.1] Expose renderers through JLayout

dgrammatiko:
Pull Request for Issue # .

This is a retake (obviously a lot different) to my first PR ever for Joomla #4052 (wow 7 years...). It seems I did a bunch of PRs but never came around solving my own problem in a way that the core would enable Front End developers to customise the renderers without the need to roll their own plugin. Basically, this PR just completes the work that someone else started way back when introduced the message jLayout. In that sense **is not** a new feature, it just contributes to the goal that everything producing HTML code is overridable.

### Summary of Changes
- Move the code of the renderers to their respective JLayout files

### Why?
Renderers produce fragments of HTML (or XML in case of RSS) but their internals are hard to be overridden (esp for people that are not familiar with the idea of Joomla's plugins). Moving them to jLayouts makes them immediately more approachable as in a template you can override them and plugin your own changes.

### Testing Instructions
Test that all 3 templates working fine (front end, back end, installation)

??? Are the jLayouts documented somewhere?


brianteeman:
It is way too late in the development cycle to introduce something so extensive - my 2c

dgrammatiko:
> It is way too late in the development cycle to introduce something so extensive

Can go to 4.1 or another release could introduce this. I just keep hacking my way around this problem for way too many years so I thought maybe I can fix this in a way that makes Joomla more approachable for front end devs (eg https://github.com/dgrammatiko/sloth/tree/step-6-inherritable/sloth/helper)

brianteeman:
We've more important things to do right now - like making 4.0 usable for people who are not front end devs - which it currently is not

dgrammatiko:
> We've more important things to do right now

Sure, I'm also working on a couple of release blockers. This was a quick one as I was doing some side work on a project and realised that I still have to hack my way around to get the renderers to echo what I want. Also just because I opened a PR doesn't necessarily mean that it's gonna be accepted or merged (you know how things work here)...

brianteeman:
It also distracts from the current issues - perhaps rename it as [4.1] ?

dgrammatiko:
Done

libraries/src/Document/HtmlDocument.php
```diff
@@ -871,6 +873,11 @@ protected function _parseTemplate()
 				}
 			}
 
+			if (count($messages) === 0)
+			{
+				throw new \Exception('A Joomla template needs a messages block, ie: `<jdoc:include type="messages" />`!');
```
ReLater: Why an Exception? Should be the decision of the template developer to have this jdoc tag or not. In some cases I "redirect" to another template file than index.php, e.g. index2.php, where I don't want no messages and I know why.

dgrammatiko: > where I don't want no messages and I know why.

It's not about what **YOU** want but rather what are the minimum interop requirements for proper operation of the CMS. If you remove the messages then how will the user be notified of any of the countless possibilities that could throw an alert?

Basically, if you serve pages without messages you're doing Joomla wrong

dgrammatiko: @ReLater nevermind this wasn't supposed to be part o this PR

dgrammatiko:
I guess plugins are working fine also here, so Frontenders using Joomla should be also PHP experts...

PR #32861, opened 2021-03-25
Adding an attribute aria-valuemin and aria-valuemax in Progress bar

Shubham-com:
Pull Request for Issue # .
aria-valuemin attribute is used to define the minimum value allowed for a range widget such as a slider, spinbutton or progressbar.
aria-valuemax attribute is used to define the maximum value allowed for a range widget such as a slider, spinbutton or progressbar.

### DIFF
--- a/administrator/modules/mod_sampledata/tmpl/default.php
+++ b/administrator/modules/mod_sampledata/tmpl/default.php
@@ -45,7 +45,7 @@ $app->getDocument()->addScriptOptions(
 			<?php // Progress bar ?>
 			<li class="list-group-item sampledata-progress-<?php echo $item->name; ?> d-none">
 				<div class="progress">
-					<div class="progress-bar progress-bar-striped progress-bar-animated" role="progressbar"></div>
+					<div class="progress-bar progress-bar-striped progress-bar-animated" role="progressbar" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" style="width:10%"></div>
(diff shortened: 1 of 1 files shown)


brianteeman:
Did you actually check that this was needed before making the pull request

Shubham-com:
Yes,I think that it needed to define the maximum and minimum value of a Progress bar.

brianteeman:
Can I suggest that you check again

rdeutz:
> aria-valuemin is a required attribute of role slider, scrollbar and spinbutton.

this here has role="progressbar"

I am not an expert, but I am against adding what is not required

brianteeman:
OK - seems that no one is looking at how the code works. - I was trying to be nice and let you see for yourself.

The progress bar should have a min/max etc

The progress bar has a min/max it is set in the javascript

https://github.com/joomla/joomla-cms/blob/4.0-dev/build/media_source/mod_sampledata/js/sampledata-process.es6.js#L81-L87

rdeutz:
> The progress bar has a min/max it is set in the javascript

so then it's fine

Shubham-com:
Ok..thanks now I got it..

Shubham-com:
I am new this codebase..I have not seen that it is already defined in javascript.

brianteeman:
@Shubham-com You would have seen if you had done the most important thing. Tested. Code does not live in an IDE.

PR #29007, opened 2020-05-09
[4.0] Session Manager for destroying sessions

SharkyKZ:
Replacement PR for https://github.com/joomla/joomla-cms/pull/26891.

### Summary of Changes

See https://github.com/joomla/joomla-cms/pull/26891.

### Testing Instructions

See https://github.com/joomla/joomla-cms/pull/26891.

### Documentation Changes Required

Yes, see https://github.com/joomla/joomla-cms/pull/26891.

PR #26891, opened 2019-11-02
[4.0] Session Manager for destroying sessions, fix session destruction in user privacy plugin

mbabker:
Partial Pull Request for Issue #26539

### Summary of Changes

Force logout of a user does not work consistently unless you are using the database as your session handler.  This pull request for 4.0 refactors the system to ensure all functions work with all handlers.

The session handler used internally by `Joomla\CMS\Session\Session` is now exposed as a service, this should never be direct accessed but is needed to make the new manager class work.

The `Joomla\CMS\Session\SessionManager` class is introduced as a means for allowing someone to destroy a session's data using the session handler.

The "User - Joomla!" plugin is adjusted to use the new session manager to handle session destruction when deleting a user or when the force logout option is enabled at logout (contingent on session metadata being available).

The "Privacy - User" plugin is adjusted to do the same when removing data (currently this will fatal out because the plugin tries using the no longer existing `JSessionStorage` class).

### Testing Instructions

- Ensure you can log in correctly
- Ensure you can log out correctly
- With session metadata enabled (default) and the "Force Logout for all Sessions?" option enabled (default) in the "User - Joomla!" plugin, all sessions should be destroyed when logging out regardless of session handler (should be validated using at least database and filesystem, if you have others available then bonus points to you)
- With session metadata disabled and the "Force Logout for all Sessions?" option enabled (default) in the "User - Joomla!" plugin, only the current session should be destroyed when logging out
- With session metadata enabled (default) and the "Force Logout for all Sessions?" option enabled (default) in the "User - Joomla!" plugin, logging a user out from the "Logged-in Users" module should log out all sessions for the given user
- With session metadata enabled (default) and the "Force Logout for all Sessions?" option disabled in the "User - Joomla!" plugin, logging a user out from the "Logged-in Users" module will not work
    - Fixing this is beyond scope of this pull request, the ability to destroy all sessions for a user has always been reliant on this option being enabled and to make it work with the option disabled will require some way of passing a signal from the logout controller to the application and the plugin(s) handling the logout event
- With session metadata disabled, you shouldn't even see a user list in the "Logged-in Users" module to do anything with the account

### Expected result

- Logging a user out with the icon in the "Logged-in Users" module logs a user out regardless of session handler
- Removing a user account with a privacy request does not trigger a fatal error

### Actual result

- Logging a user out with the icon in the "Logged-in Users" module logs a user out only when using the database session handler
- Removing a user account with a privacy request triggers a fatal error due to the use of a class removed from 4.0

[...]

PR #32863, opened 2021-03-25
[4.0] Fix for Button Log out staying when session manager is file

rdeutz:
Follow up  on PR #29007

### Summary of Changes

When the session manager is file and you are trying to log out a user in the frontend, the Log Out Button doesn't go away. The reason is that we still have a row in the session table. 

The user is loged out and when you refresh the frontend the button is going away in the backend. That's kind of confusing because it looks as it hadn't worked.

### Testing Instructions

* Set session mananger to file
* Create a user
* Login with the user in the frontend
* Login with a superuser account in the backend
* Try to log out the user you are logged in in frontend

You should see that the button stays

* Go to the frontend an refresh the page

Now you can see that the Button is not longer there after refreshing the backend

* Apply this PR
* Login with the user in the frontend
* Login with a superuser account in the backend
* Try to log out the user you are logged in in frontend

Button goes away.

grafik

grafik

no

### DIFF
--- a/plugins/user/joomla/joomla.php
+++ b/plugins/user/joomla/joomla.php
@@ -431,6 +431,23 @@ class PlgUserJoomla extends CMSPlugin
 			{
 				return false;
 			}
+
+			// If session manager is not database we need to clean up the internal session table
+			if ('database' !== $this->app->get('session_handler', 'filesystem'))
+			{
+				$query = $this->db->getQuery(true)
+					->delete($this->db->quoteName('#__session'))
+					->whereIn($this->db->quoteName('session_id'),  $sessionIds, ParameterType::STRING);
+
+				try
+				{
+					$this->db->setQuery($query)->execute();
+				}
+				catch (ExecutionFailureException $e)
+				{
+					return false;
+				}
+			}
 		}
 
 		// Delete "user state" cookie used for reverse caching proxies like Varnish, Nginx etc.


plugins/user/joomla/joomla.php
```diff
@@ -405,6 +405,23 @@ public function onUserLogout($user, $options = [])
 			{
 				return false;
 			}
+
+			if ('database' !== $this->app->get('session_handler', 'filesystem'))
+			{
+				// If session manager is not database we need to clean up the internal session table
+				$query = $this->db->getQuery(true)
+					->delete($this->db->quoteName('#__session'))
+					->whereIn($this->db->quoteName('session_id'),  $sessionIds, ParameterType::STRING);
```
wilsonge, line 440: `$sessionIds` doesn't look like it's defined? Might be being blind though.

rdeutz: You are not blind, it's in the other PR, when 29007 is merged we have it. I have to fix a merge conflict then. I will set this one to draft and wait till the other one is merged

joomdonation:
This is not working if **Force Logout for all Sessions?** parameter in **User - Joomla!** plugin set to **No**. Also, maybe you should merge https://github.com/joomla/joomla-cms/pull/29007 into your PR so that it is easier to test? @SharkyKZ deleted his branch, so it is not easy to test his PR (it could not be tested using Patch tester component)

joomdonation:
<strike>Also, we have **MetadataManager** class https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Session/MetadataManager.php and as I understand, it's used to manage session meta data. I wonder if we should move the logic for deleting meta data which you added here into that class and use it?</strike>

Ignore this. It could be done in a separate PR if we want because looking more at the code, we have code to delete session meta data manually in other place, too (not via **MetadataManager** class) - for example, in **ApplicationModel** in com_config

joomdonation:
> This is not working if Force Logout for all Sessions? parameter in User - Joomla! plugin set to No

Never mind about this. Looked at the code and read https://github.com/joomla/joomla-cms/pull/26891 again, I see that it is not possible to fix the issue with this condition.

richard67:
@rdeutz Now as PR #29007 has been merged, could you update your PR to latest 4.0-dev and set it to ready for review? Thanks in advance.

rdeutz:
@richard67 thanks for the reminder, done

joomdonation:
@rdeutz There is still one case which haven't addressed by this PR:

- **Track Session Metadata** in Global Configuration set to Yes.
- **Force Logout for all Sessions?** parameter in the **User - Joomla!** plugin set to No
- Session Handle is not database

In this case, if user logged out himself (from frontend for example), the session meta data for his session is not being deleted, so he is still being displayed in **Logged-in Users** in backend dashboard. That should be fixed, something we will have to process on this block of code https://github.com/joomla/joomla-cms/blob/7c08180da0bc8e8f1b4e3338c62b0aa64b54db56/plugins/user/joomla/joomla.php#L395

plugins/user/joomla/joomla.php
```diff
@@ -431,6 +431,23 @@ public function onUserLogout($user, $options = [])
 			{
 				return false;
 			}
+
+			// If session manager is not database we need to clean up the internal session table
+			if ('database' !== $this->app->get('session_handler', 'filesystem'))
+			{
+				$query = $this->db->getQuery(true)
+					->delete($this->db->quoteName('#__session'))
+					->whereIn($this->db->quoteName('session_id'),  $sessionIds, ParameterType::STRING);
```
joomdonation, line 440: I haven't looked at prepared statement yet but not sure if there is any difference but on other place, we use ParameterType::LARGE_OBJECT instead of ParameterType::STRING https://github.com/joomla/joomla-cms/blob/4.0-dev/plugins/privacy/user/user.php#L178

rdeutz:
> @rdeutz There is still one case which haven't addressed by this PR:

I am sure there are more, but I would like to concentrate here on the issue at hand. This issue comes up when you have default settings. Let's address other situations in other PRs.

joomdonation:
OK, fine. I will try to make a different PR to address that case later.

plugins/user/joomla/joomla.php
richard67: The reason for large object can be found in SharkyKZ's comment in the original PR https://github.com/joomla/joomla-cms/pull/26891#discussion_r341808398 :
```
This is needed for PostgreSQL because session_id column has binary data type.
```
So I think we need `ParameterType::LARGE_OBJECT` here, too.

ghost:
Are no Prebuilt packages available for download because of `continuous-integration/drone/pr Expected — Waiting for status to be reported`?

@rdeutz PR #29007 is merged, can you please update test instructions?

richard67:
@sandramay0905 Drone seems to be hanging for this PR. That can happen in very rare cases. Will be fixed when @rdeutz will update his branch, but @rdeutz please wait with that until the 4.0-dev branch is fixed, see #33028 .

ghost:
I have tested this item :white_check_mark: successfully on 7c08180da0bc8e8f1b4e3338c62b0aa64b54db56<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32863">issues.joomla.org/tracker/joomla-cms/32863</a>.</sub>

wilsonge:
This should be fixed after the upmerge from staging without this PR - we now use https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/User/UserHelper.php#L606 - could someone test please

joomdonation:
@wilsonge I can confirm that the issue is not valid anymore with nightly build. So this PR could be closed.

alikon:
cannot reproduce even me, so closing it

richard67:
Can also not reproduce.

PR #32864, opened 2021-03-25
[4.0] Subforms/subfields

brianteeman:
I was doing some work today with the new subfields and I found it quite difficult to explain. I think that instead of saying

"This plugin lets you create new fields of type 'subfields' in any extension where custom fields are supported."

We change it so that field is called subform and the fields inside it are just fields

"This plugin lets you create new fields of type 'subform' that contains one or more fields in any extension where custom fields are supported."

### DIFF
--- a/administrator/language/en-GB/plg_fields_subfields.ini
+++ b/administrator/language/en-GB/plg_fields_subfields.ini
@@ -3,10 +3,10 @@
 ; License GNU General Public License version 2 or later; see LICENSE.txt
 ; Note : All ini files need to be saved as UTF-8
 
-PLG_FIELDS_SUBFIELDS="Fields - Subfields"
-PLG_FIELDS_SUBFIELDS_LABEL="Subfields (%s)"
-PLG_FIELDS_SUBFIELDS_PARAMS_CUSTOMFIELD_LABEL="Subfield"
-PLG_FIELDS_SUBFIELDS_PARAMS_OPTIONS_LABEL="Subfields"
+PLG_FIELDS_SUBFIELDS="Fields - Subform"
+PLG_FIELDS_SUBFIELDS_LABEL="Subform (%s)"
+PLG_FIELDS_SUBFIELDS_PARAMS_CUSTOMFIELD_LABEL="Field"
+PLG_FIELDS_SUBFIELDS_PARAMS_OPTIONS_LABEL="Fields"
 PLG_FIELDS_SUBFIELDS_PARAMS_RENDER_VALUES_LABEL="Render Values"
 PLG_FIELDS_SUBFIELDS_PARAMS_REPEAT_LABEL="Repeatable"
-PLG_FIELDS_SUBFIELDS_XML_DESCRIPTION="This plugin lets you create new fields of type 'subfields' in any extension where custom fields are supported."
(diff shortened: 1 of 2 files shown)


ChristineWk:
Sorry I'm not sure, because with Patch, I get:

### screen shot 2021-03-25 at 20 53 30<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32864">issues.joomla.org/tracker/joomla-cms/32864</a>.</sub>

brianteeman:
That is because I did not want to change the type of the field at this late stage

dgrammatiko:
Why not renaming also the plugin to `subform`? Seems weird that the filesystem has a `subfields` name and all the other references now are `subform`. Reminds me of the `com_finder`-  `smart search` paradigm and I always found that very unintuitive...

brianteeman:
I can do that but its a bigger change.

Quy:
I have tested this item :white_check_mark: successfully on 22967ae9d8a1873712cd05f43990a455be1b0ab9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32864">issues.joomla.org/tracker/joomla-cms/32864</a>.</sub>

AndySDH:
After thinking about this more, I think 'Subform' may not be the best name either. I'm fine with it, but could be changed.

As 'subform' implies you can create a little form inside a bigger form. But it's the opposite, this field type IS the bigger form, which will contain child fields.

So a few alternative ideas:
- Parent Field
- Nested Fields
- Grouped Fields
- Multi-Fields
- Fields Container
- Fields Collection
- Rabbit Hole (jk :D )
- or something along these lines...

I think Nested Fields may be the one that makes most sense, after all it's exactly what it's doing:
https://en.wikipedia.org/wiki/Nesting_(computing)
> nesting is where information is organized in layers, or where objects contain other similar objects.

Just throwing some ideas.

Or Subform is fine too. Doesn't really matter much, as long as it works :D

AndySDH:
@brianteeman @joomdonation @wilsonge @Fedik @Hackwar @chmst @Quy @richard67 

What do you think? Help us define a final name for this :)

joomdonation:
I have tested this item :white_check_mark: successfully on 22967ae9d8a1873712cd05f43990a455be1b0ab9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32864">issues.joomla.org/tracker/joomla-cms/32864</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32864">issues.joomla.org/tracker/joomla-cms/32864</a>.</sub>

joomdonation:
I think the change in this PR is good, it make it consistent with other field plugins which we are having at the moment. Once this PR is merged, we will try to change the plugin to make it become real Subform.

brianteeman:
Thanks for testing and merging

Issue #31218, opened 2020-10-23
[4.0] Remove obsolete Bootstrap classes in forms and fields


drmenzelit:
Forms (e.g. login form from com_user) and fields (layouts/joomla/forms/renderfield.php) are still using classes from Bootstrap 2: control-group, control-label, controls

That should be changed to classes from Bootstrap 4: form-group without extra div for label and input
https://getbootstrap.com/docs/4.5/components/forms/

hans2103:
oh boy.. that's a nice one to solve

brianteeman:
ITs  not an easy one. There is both markup and css to change AND we would need to keep the existing css for extensions - its too late in the process to change that and force extensions to update

PhilETaylor:
This is one of only 2 outstanding issues flagged with the Joomla 4.0 Milestone. Please can we move the milestone else Joomla 4 Milestone will never reach 100%. Thanks.

drmenzelit:
Done

coolcat-creations:
Can this issue be closed? If not: What's the prefered solution?
Extension Devs could maybe use the @extend Command to apply their classes on the (probably) new markup.

chmst:
@drmenzelit is this still an issue for you?

drmenzelit:
As Brian wrote, it is too late for Joomla 4. We will need to reconsider this for Joomla 5. I'm closing this for the moment.

coolcat-creations:
Why closing and not labeling for Joomla 5 with a release blocker? Otherwise it would be again too late or forgotten.

PR #32865, opened 2021-03-25
removed obsolete bootstrap classes

Vaibhav-afk:
Pull Request for Issue #31218

### Summary of Changes

- I did changes in:

      1. default_login.php.
      2. renderfield.php

### Output:
Capture

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "recursive-readdir": "^2.2.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.42.1",
-    "rollup-plugin-vue": "^5.1.9",
+    "rollup-plugin-vue": "^6.0.0",
     "sass": "^1.32.8",
     "selenium-standalone": "6.23.0",
     "stylelint": "^13.12.0",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@webcomponents/webcomponentsjs": "^2.5.0",
     "accessibility": "^3.0.10",
     "awesomplete": "1.1.5",
-    "bootstrap": "^5.0.0-beta2",
+    "bootstrap": "^5.0.0-beta3",
     "choices.js": "^9.0.1",
     "chosen-js": "^1.8.7",
     "codemirror": "^5.60.0",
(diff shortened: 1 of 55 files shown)


package.json
```diff
@@ -97,7 +97,7 @@
     "recursive-readdir": "^2.2.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.42.1",
-    "rollup-plugin-vue": "^5.1.9",
+    "rollup-plugin-vue": "^6.0.0",
```
dgrammatiko: Please undo this, you're breaking the MEdia manager App here!!!

dgrammatiko (changes requested):

drmenzelit:
Did you checked how the login page look after your changes? You are changing more classes than necessary for the issue mentioned.
And do your changes have side effects on other forms?

Vaibhav-afk:
I did the changes please review.

ceford:
The original issue mentioned Bootstrap 4 but Joomla 4 is now using Bootstrap 5 and form-group has been removed. So this PR is wrong!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32865">issues.joomla.org/tracker/joomla-cms/32865</a>.</sub>

chmst:
https://github.com/joomla/joomla-cms/issues/31218#issuecomment-726579628  

Thank you for contributing. You cannot know this, of course, but as you see in the comment above it is not as easy task. There are 3rd party extension which use these classes for their own js and css - so we cannot remove classes as you did in the default_login.

ceford:
@chmst I think I see: control-group was changed to form-group when Joomla 4 planned to use Bootstrap 4 and the recent decision to use Bootstrap 5 has come too late to change all 170 instances of form-group to mb-3 in Joomla core extensions. I don't see where third party extensions come in. If they have to change control-group anyway why not change to mb-3. I guess there are more complications than I can shake a stick at.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32865">issues.joomla.org/tracker/joomla-cms/32865</a>.</sub>

brianteeman:
This probably should be closed

Quy:
Closing per https://github.com/joomla/joomla-cms/pull/32865#issuecomment-808940864, Thank you for your contribution.

PR #32885, opened 2021-03-27
Hovering effect

Shubham-com:
Pull Request for Issue # .

### Summary of Changes
Code Review..
Menu items are more attractive in light(white) backgroud when you hovering on it.

Screenshot (2683)

image

### DIFF
--- a/administrator/templates/atum/scss/blocks/_header.scss
+++ b/administrator/templates/atum/scss/blocks/_header.scss
@@ -4,7 +4,7 @@
   z-index: $zindex-header;
   box-sizing: border-box;
   padding: 10px;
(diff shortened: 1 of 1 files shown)


online-today:
This is a matter of taste: I like the "before" more and I think it is clearer because the highlighting has not the colour of the background.

richard67:
I agree with @online-today . But as said, it might be a matter of taste.

Shubham-com:
Yes,It might be a colour of Taste...

In this,when I hover the media or any other icon..the lighter colour change into deeper colour...that's
why I think so..that it may the deeper icon should change into lighter colour...
image

PR #32910, opened 2021-03-28
[4] Display relative date instead of date stamp in user action log module

PhilETaylor:
### Summary of Changes

Propose to use a relative date (5 minutes ago) in the Latest Actions module rather than a date stamp. (2020-02-01)

### Testing Instructions

Ensure you have some Actions in your User Action Log 
Check the Home Dashboard -> Latest Actions module - look at it
Apply PR
Look at it again and note the times are now relative

### Actual result BEFORE applying this Pull Request

Screenshot 2021-03-28 at 21 04 36

### Expected result AFTER applying this Pull Request

Screenshot 2021-03-28 at 21 04 18

### Documentation Changes Required

none

PR #32911, opened 2021-03-28
[4] Show relative date instead of date time in logged in module

PhilETaylor:
### Summary of Changes

Propose showing of relative date instead of date time in the Logged in Users module in admin.

### Testing Instructions

Check the Home Dashboard -> Logged In Users module - look at it
Apply PR
Look at it again and note the times are now relative

Screenshot 2021-03-28 at 21 14 12

Screenshot 2021-03-28 at 21 14 02

none

### DIFF
--- a/administrator/modules/mod_logged/tmpl/default.php
+++ b/administrator/modules/mod_logged/tmpl/default.php
@@ -9,6 +9,7 @@
 
 defined('_JEXEC') or die;
 
+use Joomla\CMS\Factory;
 use Joomla\CMS\HTML\HTMLHelper;
 use Joomla\CMS\Language\Text;
 ?>
@@ -54,7 +55,7 @@ use Joomla\CMS\Language\Text;
 					<?php endif; ?>
 				</td>
 				<td>
-					<?php echo HTMLHelper::_('date.relative', JFactory::getDate($user->time)); ?>
+					<?php echo HTMLHelper::_('date.relative', Factory::getDate($user->time)); ?>
 				</td>
 			</tr>
 		<?php endforeach; ?>
--- a/administrator/modules/mod_logged/tmpl/default.php
+++ b/administrator/modules/mod_logged/tmpl/default.php
@@ -54,7 +54,7 @@ use Joomla\CMS\Language\Text;
 					<?php endif; ?>
 				</td>
 				<td>
-					<?php echo HTMLHelper::_('date.relative', date('Y-m-d H:i:s', $user->time)); ?>
+					<?php echo HTMLHelper::_('date.relative', JFactory::getDate($user->time)); ?>
 				</td>
 			</tr>
 		<?php endforeach; ?>
--- a/administrator/modules/mod_logged/tmpl/default.php
+++ b/administrator/modules/mod_logged/tmpl/default.php
@@ -54,7 +54,7 @@ use Joomla\CMS\Language\Text;
 					<?php endif; ?>
 				</td>
 				<td>
-					<?php echo HTMLHelper::_('date', $user->time, Text::_('DATE_FORMAT_LC5')); ?>
+					<?php echo HTMLHelper::_('date.relative', date('Y-m-d H:i:s', $user->time)); ?>
 				</td>
 			</tr>
 		<?php endforeach; ?>


rjharishabh:
It always shows **Less than a minute ago.**

PhilETaylor:
That is correct. It’s actually a last seen date and not a logged in date time

Check without this PR and you will see the same thing - it updates on page load

rjharishabh:
Without this PR, logged in date and time shown,

I think results after this PR should be the same as #32910

PhilETaylor:
> Without this PR, logged in date and time shown

No it doesnt. 

Without this PR, the mod_logged shows the `time` field from the `#__session` table which is updated on every page load. 

It is retrieved from the `#__session` table here: 

https://github.com/joomla/joomla-cms/blob/d4f78beccfea59db8864cad0fa34ff67906809de/administrator/modules/mod_logged/src/Helper/LoggedHelper.php#L43

This module is not telling you the time someone logged in. Its telling you time the session was updated (the last page load normally)

https://github.com/joomla-framework/session/blob/ea7af528c0349af94ed855d3ee5f843cc4986089/src/Handler/DatabaseHandler.php#L290

This PR make no changes to that logic, **all it does is display the date/time in a different way.**

rjharishabh:
Thanks for clarification

chmst (changes requested):
Suggested:  Use Factory::getDate(..)

PhilETaylor:
I can find no examples of `getDate` accepting a unix timestamp as a parameter, and that's what we have here. 

I'll test later tonight.

Harmageddon:
Can we please make this configurable? As I created PR #31675 already some time ago, where it is documented that not everyone prefers the relative display instead of an absolute one. If we add relative display not only for the action logs, but also in other places in core, it might be better to have a user-specific setting for this instead of the parameter I added for com_actionlogs.

PhilETaylor:
If I read the code right, the relative date will only show if the date is less than 4 weeks, else the full date would be shown as normal. The things on the dashboard are recent things, and so a relative date is best for them.

ceford:
I have tested this item :white_check_mark: successfully on 43cd9efa01d6a45e709b1056de3bdfa60ef439f4<br /><br />I was confused by the `Less than a minute ago` too but with the explanation of the time being since the last page load I definitely approve. I use this feature quite a loy in a J3 installation.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32911">issues.joomla.org/tracker/joomla-cms/32911</a>.</sub>

alikon:
i'm not sure if this is the way to go ... like #32910
only my 2eurocent

PhilETaylor:
The home dashboard is an overview of the RECENT happenings on a site therefore it MAKES SENSE that this screen should have relative dates. 

On normal listing screens date stamps are better, and can be ordered etc, but on the Home Dashboard a relative timestamp is better.... unless your site is dead and inactive and you have nothing to show recently of course, then, the relative date AFTER FOUR WEEKS turns back into a date 

IIRC you can also hover over "one hour ago" to see the exact date/time.

PhilETaylor:
> @chmst Suggested: Use Factory::getDate(..)

Done.

alikon:
maybe i've been unexplicative
matter of personal taste
much more imho

PhilETaylor:
Well as Joomla 3.9, 3.10 and 4.0 are no longer accepting "new features" any "preference selection" would have to be a different PR aimed at 4.1 whenever that would be released in 2030+

alikon:
grr 
i always hate when something is on the bordeline
whatever can be your choice....i'm optimistic here 50/50... 2025?
:lollipop:

PhilETaylor:
I'll do the work, but I have no say in when, if ever, it gets merged. At the moment Ive had several rejected.

Harmageddon:
> If I read the code right, the relative date will only show if the date is less than 4 weeks, else the full date would be shown as normal.

That's correct. Still I'm confused by things like "2 weeks ago" or "5 days ago" - was it Wednesday? Thursday? Let me calculate...

Though I see your point here:

> The home dashboard is an overview of the RECENT happenings on a site therefore it MAKES SENSE that this screen should have relative dates.
> 
> On normal listing screens date stamps are better, and can be ordered etc, but on the Home Dashboard a relative timestamp is better.... unless your site is dead and inactive and you have nothing to show recently of course, then, the relative date AFTER FOUR WEEKS turns back into a date

Whether one or the other is better is a question of personal taste, thus, I'd prefer the parameter. As you see in #31675, there are several people saying "just get rid of the parameter and display it in absolute format" and several saying "just get rid of the parameter and display it in relative format".

I agree that the argumentation for relative dates is stronger in this case here than for the whole action logs table. 

> ~IIRC you can also hover over "one hour ago" to see the exact date/time.~ Apparently not, although that could easily be added.

So I'd kindly ask you to at least include this tooltip in your PRs to give users the chance to see the absolute date.

> Well as Joomla 3.9, 3.10 and 4.0 are no longer accepting "new features" any "preference selection" would have to be a different PR aimed at 4.1 whenever that would be released in 2030+

Seems a bit counterintuitive to me that changing the display would be considered a bug fix or cosmetical change (for 4.0), but making it configurable would be a new feature (4.1)?

Okay, you two are writing faster than I am.

> I'll do the work, but I have no say in when, if ever, it gets merged. At the moment Ive had several rejected.

Would be very much appreciated! :-)

administrator/modules/mod_logged/tmpl/default.php
```diff
@@ -54,7 +54,7 @@
 					<?php endif; ?>
 				</td>
 				<td>
-					<?php echo HTMLHelper::_('date', $user->time, Text::_('DATE_FORMAT_LC5')); ?>
+					<?php echo HTMLHelper::_('date.relative', JFactory::getDate($user->time)); ?>
```
Quy:
```suggestion
					<?php echo HTMLHelper::_('date.relative', Factory::getDate($user->time)); ?>
```
Please add `use Joomla\CMS\Factory;`

PhilETaylor: Doh, Im context switching so much at the moment Im forgetting which version of Joomla or symphony Im meant to be working in :)

toivo:
I have tested this item :white_check_mark: successfully on 24796111ce1a1514c3b527513aab21780a659b73<br /><br />Tested successfully in Beta8-dev of 29 March in Wampserver 3.2.4 using PHP 8.0.2.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32911">issues.joomla.org/tracker/joomla-cms/32911</a>.</sub>

chmst (approved):

PhilETaylor:
Im not going to add yet-another-useless-option to toggle between relative/absolute dates on a dashboard of "recent activity"... sorry. Joomla has enough insane options.

So lets get this tested, (hint: it works) and then we can get it merged and move on.

ceford:
I have tested this item :white_check_mark: successfully on 24796111ce1a1514c3b527513aab21780a659b73<br /><br />Works fine - I like it.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32911">issues.joomla.org/tracker/joomla-cms/32911</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32911">issues.joomla.org/tracker/joomla-cms/32911</a>.</sub>

dgrammatiko:
Did anyone tested that filtering by date is still functional after applying this PR?

PhilETaylor:
Its was not possible before or after this PR to sort by date in this module.

dgrammatiko:
Yeah, just realised that this is not a list view although it resembles one

HLeithner:
I'm against this change for multiple reasons but most of them are personal ;-)

Anyway column header is "Date" the value "2 minutes ago" is not a date so it should be "last seen" aside this I prefer real dates especially if the values are bigger then a day. Also there is no fallback like a tooltip with the real date.

Issue #32592, opened 2021-03-05
The extensions updater has several issues


nikosdion:
I have spotted a few issues with the way the Joomla extensions update works.  The Joomla project is already aware of these issues. I had privately discussed them in person with @wilsonge several years ago and emailed @zero-24 about them several months ago. No action has been taken. Just like when I had reported the same issues to your predecessors between 2013 and 2016. No action taken whatsoever.

Since the JED [requires developers to use this subpar updater](https://extensions.joomla.org/support/knowledgebase/submission-requirements/joomla-update-system-requirement/) to be listed in the JED, the Joomla project needs to either fix these issues or allow developers to implement THEIR OWN updater code. It's not reasonable demanding that 3PDs use your updater and bear the temporal, financial and repetitional cost of its inadequacies.

Seeing how certain core contributors abused the discretion expressed in my previous disclosures of these issues in private to make unsubstantiated public claims I feel the urge to report them publicly. You can no longer pretend you have never heard of these issues I have reported to the Joomla project the last 8 years. You can no longer pretend that the JED rule makes perfect sense and somehow serves the 3PDs best interests when all it does is hinder us in delivering updates reliably to our users while bearing the costs of a decision imposed upon us by OpenSourceMatters Inc.

## You cannot update the `extra_query` after an update has been discovered

Joomla is using a weird update process. The update site has the `extra_query` column where the developer (or the user, as of Joomla 4) can enter an authentication query to the server delivering the update. Joomla retrieves the update information, populates the `#__update` table and copies over the `extra_query` column. I know how this part works because I contributed it for Joomla 3.2.

Before you start saying things like that I am reporting a bug I created please be advised that the fact it is implemented the way it is was NOT my choice. It was something discussed with Mark Dexter who explicitly framed my proposed contribution as something that should not change the way Joomla's extensions updates work. A few months after this contribution I had written a white paper on creating a better extensions updater that the Joomla project refused to consider. This was 8 years ago.

In the original version of the code updating the `extra_query` column in `#__update_sites` would simply require to Clear Cache on the Update page and then use Fetch again to retrieve the new updates, copying over the updated `extra_query` column. Further to that, Joomla would display the server response which would be used by the developer to communicate that the user needs to enter the correct download key and refresh the updates. The latter was removed for security reasons but it's largely irrelevant since Joomla 4 does notify the user when an update key is missing.

[...]

PR #32176, opened 2021-01-27
Optimize the code introduced with PR 32006 to rename incorrectly cased files on update

richard67:
Pull Request for Issue # .

### Summary of Changes

This Pull Request (PR) simplifies the code in the new function `fixFilenameCasing` which has been added to `script.php` with PR #32006 by using old and new paths which begin with a slash, so it doesn't need to prepend the slash every time we use it.

This makes it also consistent with the calling function `deleteUnexistingFiles`, where the paths of files and folders to be removed also are paths starting with a slash.

### Testing Instructions

Check that the testing instructions from PR #32006 still work on a current staging or latest 3.9.x nightly build after the patch of this PR here has been applied.

Use the [update package built by drone for this PR](https://ci.joomla.org/artifacts/joomla/joomla-cms/staging/32176/downloads/39683/Joomla_3.9.25-dev+pr.32176-Development-Update_Package.zip) for Update.

### Actual result BEFORE applying this Pull Request

Works as described in the testing instructions of PR #32006 .

### Expected result AFTER applying this Pull Request

Still works as described in the testing instructions of PR #32006 .

### Documentation Changes Required

No.

PR #32915, opened 2021-03-28
[4] Ensure opcache file invalidate on file copy & delete

PhilETaylor:
This is part 2 of https://github.com/joomla/joomla-cms/issues/32592 and directed at Joomla 4

### Summary of Changes

When PHP is configured to use modern Opcache, PHP will attempt to hold the compiled version cached in memory. 

There are many ways to configure opcache invalidations, including never revisiting the source PHP to compile it again ever (if [opcache.validate_timestamps](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.validate-timestamps) is disabled for example) 

To fully ensure that files that Joomla writes to the hard disk are taken into account by a server with a correct configured (or badly configured) opcache, we need to invalidate files from the opcache after writing a new version to the disk. 

**Joomla has previously used `opcache_reset` sporadically to do this, but  this is VERY BAD as causes a WHOLE SERVER cache clear - including other sites, and other sites you might not own! It can also cause spikes in CPU because all those sites need to recompile their opcodes!** 

### Testing Instructions

Hard to test unless you really know what you are doing and can reconfigure all your PHP stack to include opcaching. Also some of the edge cases this fixes will only become clear if you are an extension developer mass distributing extensions. 

Install some extensions - nothing should break. 

EDIT: Some more details on the testing can be found in comment https://github.com/joomla/joomla-cms/pull/32915#issuecomment-812562008

You can install extensions

You can install extensions and opcache for each file is invalidated

none

### Others

WordPress started this conversation 5 years ago and only last year got their changes into core 
https://core.trac.wordpress.org/changeset/48160

// cc @nikosdion


libraries/src/Filesystem/File.php
```diff
@@ -174,10 +176,33 @@ public static function copy($src, $dest, $path = null, $useStreams = false)
...
+	 *
+	 * @return void
+	 */
+	public static function invalidateOpcache($file)
+	{
+		if (function_exists('opcache_invalidate'))
+		{
+			$info = pathinfo($file);
+
+			if ($info['extension'] === 'php')
```
richard67:
```suggestion
			if (isset($info['extension']) && $info['extension'] === 'php')
```
From [PHP docs](https://www.php.net/manual/en/function.pathinfo.php): "Note: If the path does not have an extension, no extension element will be returned (see second example below)."

PhilETaylor: Done.

libraries/src/Filesystem/File.php
```diff
@@ -310,6 +312,7 @@ public static function move($src, $dest, $path = '', $useStreams = false)
 
 			if (!$stream->move($src, $dest))
 			{
+
```
richard67:
```suggestion
```

PhilETaylor:
Screenshot 2021-03-29 at 15 56 09

TBH I never expected anything other than that from @SharkyKZ LMAO... so predictable now...

PhilETaylor:
This is ready for testing now. Thanks.

wilsonge:
Sorry to be a pain but can you rename `invalidateOpcache` to `invalidateFileCache` or something. Just so we have options in the future if we want to call other cache invalidates

PhilETaylor:
s/invalidateOpcache/invalidateFileCache

Done.

wilsonge:
Thankyou!

PhilETaylor:
Done on the Joomla 3 version, and the Joomla-framework version PRs too.

SharkyKZ:
Not sure why this hasn't been closed yet. As maintainers will explain to you, this PR is wrong.

PhilETaylor:
Maybe you could actually tell me?

well its just been merged into Joomla-framework/filesystem so you best go all their maintainers they were wrong too @n9iels  hahaha

PhilETaylor:
> this PR is wrong.

But yet you dont have the decency to actually explain ... Its time you were removed as a maintainer @SharkyKZ

joomdonation:
If @SharkyKZ doesn't want to explain what's wrong, at least one of our maintainers should explain what's wrong in this PR. Closing a PR without explaining will hurt our contributors :(

PhilETaylor:
Well @wilsonge seemed happy enough to ask for changes without rejecting it, and he's a maintainer.

joomdonation:
@PhilETaylor Maybe @SharkyKZ has explained what's wrong with our maintainers. So please wait for an explanation.

PhilETaylor:
Ive waited 4 days since his thumbs down.

Its time his rights were removed.

SharkyKZ:
> Its time his rights were removed.

I agree. In fact, all of my rights should have been revoked in November 2020 when I left JBS. But, clearly, someone royally screwed up with my offboarding. Or perhaps didn't even bother at all. I bet if I hadn't manually removed myself from the organization on Github, I'd still have permissions here too. Unauthorized users managing the issue tracker? Sure, no problem! 👍 And this alone perfectly sums the state of the entire project. Too much ignorance on the part of people who run things here. Frankly, Phil, based on your issue history you should understand this better than anyone else.

PhilETaylor:
and still you have refused to give a reason for your drive by closing of PR's. 

That is disrespectful to contributors WHOEVER they are. 

Hiding behind a alias "SharkyKZ", that even your name is not published in official OSM meeting minutes, is it even legal for OSM to grant administrative permissions to a ghost? 

Are you going to provide feedback on why you have rejected the hard work of contributors or are you just going to personally attack me again?

bembelimen:
Sorry, that this happened.

PhilETaylor:
> Sorry, that this happened.

Im sorry too. I think I deserve and explanation, but I know I won't get one.

richard67:
> I think I deserve and explanation, but I know I won't get one.

I wish I had some, but I know as little as you do about why he does that.

bembelimen:
I have no explaination, it was a missuse of permission (it seems we forgot to revoke all permissions there) and should not have happened.
I really appreciate all of your contributions (wherever I'm not always agreeing, they still should be valued). So I can only say again, I'm sorry, that it escalated and we try to be more careful in the future.

alikon:
i ask excuse to you all
My fault cause i was the JBS team leader at that time
and  i've only moved you @SharkyKZ  in the honor list
Screenshot from 2021-04-01 19-16-05

i've wrongly assumed that was enough

PhilETaylor:
Thanks for the comments.

Does a maintainer actually have a problem with these proposed opcache PRs or not then?

richard67:
> Does a maintainer actually have a problem with these proposed opcache PRs or not then?

@PhilETaylor I don't have a problem with them, but as you maybe know I'm not really the super expert. I have some knowledge on this or that, and I know a bit about opcache and use it for my home page, but I might be wrong or missing something. I don't belong to those people who think about themselves that they know everything. But maybe that's Sharky's problem, that people like me are maintainers ... I don't really know.

richard67:
@PhilETaylor Maybe there is something wrong with this PR, I'm not sure yet. System tests are failing when removing the installation folder after the installation. I've tried a few times. Have started drone again to see if it still fails at that place.

richard67:
Hmm, again same place, when removing installation folder at the end of a new installation:

https://ci.joomla.org/joomla/joomla-cms/41482/1/22

Not sure yet if that could be related to this PR. I have to test later or tomorrow.

PhilETaylor:
I'm out driving my kid to her sleepover. I'll test the tests later tonight.

Might need to test if the file exists before clearing the opcache ? But then that would defeat the changes as we want to force the opcache to notice the file is removed

Thanks for testing

PhilETaylor:
It seems that the Refactoring (renaming) of the static method George requests did not rename all instances (very rare for phpStorm to not do this correctly!) so there were some renames missed in https://github.com/joomla/joomla-cms/pull/32915/commits/f281cf6535d32c09802168374d356e553a4b5ff1 a few hours ago. (Before the drama)

I have corrected those in https://github.com/joomla/joomla-cms/pull/32915/commits/02c03ca63c12f0faa712536170b705fe2e864997 and fingers crossed the tests now pass.

richard67:
Hmm, I should have seen that in code, too, but I was so focused on the drama that I had no eyes for code anymore.

PhilETaylor:
I have also checked if running `opcache_invalidate()` on a non-existing file (for example, when opcache `preloading` is used, or when `opcache.revalidate_timestamps` is not used and you delete a file and need to remove it from the opcache) causes an issue, and it doesn't 

Eg https://3v4l.org/9J3sb

brianteeman:
and thats why we have tests ;)

richard67:
@PhilETaylor I was trying to find out what could be the reasons for Sharky's thumbs down. Not sure if that explains all, but I've found something interesting from another well known product: https://developer.wordpress.org/reference/functions/wp_opcache_invalidate/ . They check not only if the "opcache_invalidate" function exists but also if there is a "opcache.restrict_api" ini parameter set and if the currently handled PHP file is restricted by that parameter. In the comment to their function:
For more details, see:
- https://www.php.net/manual/en/opcache.configuration.php
- https://www.php.net/manual/en/reserved.variables.server.php
- https://core.trac.wordpress.org/ticket/36455

The last link is the most interesting 😄 

I'm not sure if that was the reason for the thumbs down, I don't really think so.

But maybe we should do something like that anyway.

PhilETaylor:
I do not believe there was any code reason for the thumbs down at all. I doubt he even looked at the code. 

So you mean WordPress fixed this 4 years ago and still Joomla refuse to accept it was/is an issue /facepalm :) 

Lets copy WordPress then.. if its good enough for 30% of the internet its good enough for 2.1% of the internet...

richard67:
> So you mean WordPress fixed this 4 years ago and still Joomla refuse to accept it was/is an issue /facepalm :)

@PhilETaylor Who is Joomla? Aren't we all Joomla? As you can see, I don't refuse to accept it is an issue. I just was not aware of it before it was raised again in recent discussions, and then I was not able to provide quickly a fix, and so I am very happy that you do. I'm just one of the maintainers, and I'm only a human being, so my time and energy is limited, and my focus maybe too. The same might apply to other maintainers. So you should not say we still refuse to accept it is an issue. Whatever went wrong in past, let's not stick to that. Let's fix it (what you do, and again thanks for that) and collaborate in the best way so that at the end all will be good.

PhilETaylor:
I was referring to https://github.com/joomla/joomla-cms/issues/32592 - nothing personal. Tongue in cheek :) 

Working on it now.

richard67:
> Working on it now.

@PhilETaylor Maybe you can check their code base first if there have been made any changes on that function after it was implemented? Maybe there were bug fixes on it meanwhile.

PhilETaylor:
I have refactored the static function now to learn the lessons from other projects that have gone before us. 

I have added additional documentation. 

I have also noticed that Joomla Cache doesn't invalidate using the filesystem class, so I have forked that to https://github.com/joomla/joomla-cms/issues/32961 as something else to fix.

PhilETaylor:
Once we are happy here, I will back port these additional changes to the Joomla 3 and Framework PRs.

Fedik:
> Maybe you can check their code base first 

I have looked, their code not much different from what Phil made :wink: 
UPD: Placing "invalidate" in File class much better than WP solution.

In general, to me it like workaround of PHP bug. Maybe there already something on PHP bug tracker? 
I think the cache should be invalidated automaticly by the opcache itself, for every file manipulation with: fwrite, file_put_content, unlink etc. But this is question to opcache developers.

Other thing when the files copied manually by user , via ftp etc. 

> I have also noticed that Joomla Cache doesn't invalidate using the filesystem class, so I have forked that to #32961 as something else to fix.

because the file is removed it does not make sense to "invalidate" the cache or?
there another comment https://www.php.net/manual/en/function.opcache-invalidate.php#119026 :

>> Note that invalidation doesn't actually evict anything from the cache, it just forces a recompile....

so maybe it makes sense to call `invalidate` only for write/copy action. But I have no idea how to test :)

PhilETaylor:
> In general, to me it like workaround of PHP bug.

This is NOT a workaround for a PHP "bug". There is no PHP bug. The OPcache is doing exactly as designed. 

> I think the cache should be invalidated automaticly by the opcache itself, for every file manipulation

You can think that, and for "some" web hosts that is exactly what will happen. For "some" other web hosts this will NOT happen.

OPcache can be run with many different configurations. The most popular is "just enable it and leave it alone" and what you describe is how that will work. 

However some web hosts "optimise" their PHP configuration and attempt to run PHP code exclusively from memory (the OPCache) and try to avoid file system calls. 

Other artisan hosting will use PHP Preload features, and TOTALLY serve all PHP from the cache and NEVER check the file on the hard drive every again after the initial compile. This is how mySites.guru runs, this is how a lot of dockerised PHP apps run. 

There are web hosting services throughout the spectrum above. 

It all depends on the exact [PHP OPcache configuration](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache). Mainly

`opcache.validate_timestamps`
`opcache.revalidate_path`
`opcache.revalidate_freq`
`opcache.enable_file_override`


As Joomla is mass-distributed, we need to ensure that we are compatible with a huge range of configurations. We do already see some web hosts (Siteground) [causing issues](https://regularlabs.com/forum/cachecleaner/44323-purge-opcache-at-siteground-not-cleaned) with over aggressive caching. 

> because the file is removed it does not make sense to "invalidate" the cache 

it totally makes sense. Because under certain circumstances described above, PHP will not even notice the file has been deleted unless you tell it to check by invalidating that files opcache. PHP will just carry on loading the compiled version of the deleted file. This is one of the issues reported by Akeeba backup developers and has caused a lot of support issues. 

> Note that invalidation doesn't actually evict anything from the cache, it just forces a recompile

I know this. However what I propose is still correct.

PhilETaylor:
> But I have no idea how to test :)

Like I said in the testing instructions "Hard to test unless you really know what you are doing" :-) I was specifically vague as to only get tests from people that actually understand the PHP OPcache and how to manually configure PHP. Some PRs just need a little more experience and knowledge to test and simply following a developers provided steps to test something just doesn't cut it. 

However, if you set your PHP configuration with the following settings, your Joomla site will run exclusively from opcache, and thus be at the extreme case end of the test scenario.

```
opcache.validate_timestamps=0
opcache.max_accelerated_files=20000
opcache.interned_strings_buffer=16
opcache.save_comments = 0
opcache.enable_file_override = 1
opcache.fast_shutdown=1
```

Install and use a opcache viewer to view your opcache and its contents and settings. 

Then use Joomla to manipulate files - like installing an extension or removing an extension and ensure no error happen, and that the opcache contents change based on your actions (remembering that with the above settings the PHP is running ONLY from the opcache and its NEVER reading the PHP files on the hard disk, unless it needs to recompile the file for the opcache.

PhilETaylor:
Another good test is to configure as per my last comment, ensure you have the latest copy of this PR, and then set your site offline, and you should see the frontend offline. This means configuration.php has been recompiled into the opcache, which is what [Joomla 4 always used to do](https://github.com/joomla/joomla-cms/blob/3fdb98e66e8f3fa430432e89a6c291fda65c3416/administrator/components/com_config/src/Model/ApplicationModel.php#L969), but now we do it at a Filesystem level. 

Screenshot 2021-04-02 at 16 24 04

richard67:
@PhilETaylor There is one thing which bothers me a bit after I have read the complete https://core.trac.wordpress.org/ticket/36455 : As far as I remember, they said they invalidate the cache for a file **_before_** they modify (write) or delete it, but here and in the corresponding staging PR and in the merged PR in the framework it seems to be done **_after_** having written or deleted the file.

PhilETaylor:
Don't believe everything you read. Trust no one.

richard67:
Yep, seems I was reading something wrong. They do it also after it, like here in this PR.

PhilETaylor:
For writing a file they are wrong. 

You can prove this by saving joomla global configuration with the test config of opcache I provided and you will see it works perfectly.

Once I have finished the ironing I'm going to set up test cases to prove the deletion comments are wrong too.

PhilETaylor:
Looks like passing `force = true` (which I do in this PR) will FORCE a deleted file to be removed from opcache, even if a file doesn't exist (which would make `do_validate_timestamps(persistent_script, &file_handle) == FAILURE` also true, as the `file_handle` would be false as the file would not be able to be opened. 

https://github.com/php/php-src/blob/a6dd92f8201f1de8a1b95f4054c10f7ecc7a4fd8/ext/opcache/ZendAccelerator.c#L1318

Thats the way I read the source code of PHP-SRC. 

But Im no `C` programming expert :)

richard67:
Maybe the not existing file is handled before the place you've mentioned: https://github.com/php/php-src/blob/a6dd92f8201f1de8a1b95f4054c10f7ecc7a4fd8/ext/opcache/ZendAccelerator.c#L1304

PhilETaylor:
It seems you might be right 

accelerator_orig_zend_resolve_path

PhilETaylor:
It seems you might be right:

https://bugs.php.net/bug.php?id=75939

richard67:
Gulp ... "This can be problematic in general and for esoteric in particular." ... nice said there ;-)

richard67:
Do we have a kind of task force for esoteric problems?

PhilETaylor:
> Do we have a kind of task force for esoteric problems?

Thats you and me buddy on easter weekend! 

---- 

Ok I have completed ACTUAL testing now.

I can confirm that **to expire the opcache for a deleted file, you DO need to invalidate it BEFORE deleting it**. The rumours are right, but its always good to CHECK and not blindly follow [comments on the PHP functions documentation](https://www.php.net/manual/en/function.opcache-invalidate.php#116372) :-) 

I have proved this by having two files `deleteme.php` and `deleteit.php`

After setting up [PHP exactly as per the test configuration](https://github.com/joomla/joomla-cms/pull/32915#issuecomment-812562008) I provided (which runs ALL PHP from opcache and NEVER looks back at the PHP files unless the opcache is empty for that file) ... I start PHP, visit `deleteme.php` to create an compiled opcache for that file

TEST A: If I first unlink the file and then invalidate the opcache, the opcache for that file is NOT removed. 

```php
unlink('deleteme.php');
var_dump(opcache_invalidate('deleteme.php', true)); // = FALSE
var_dump(opcache_is_script_cached('deleteme.php')); //  = FALSE
```

TEST B: If I first invalidate opcache and then unlink the file, the opcache for that file is correctly removed. 

```php
var_dump(opcache_invalidate('deleteme.php', true)); // = TRUE
var_dump(opcache_is_script_cached('deleteme.php')); //  = FALSE
unlink('deleteme.php');
```

There is the TINIEST SLITHER of time between invalidating and unlinking where a script COULD POTENTIALLY be recompiled (by another process, another visit to the website by another visitor for example). However there is not much we can do about that as Joomla developers. 

```php
var_dump(opcache_invalidate('deleteme.php', true)); // = TRUE
// new visitor, in a different PHP process, visits deleteme.php here in parallel to this script running, which recompiles deleteme.php into opcache
unlink('deleteme.php');
var_dump(opcache_is_script_cached('deleteme.php')); //  = TRUE
```

richard67:
> Thats you and me buddy on easter weekend!

I was afraid you'd say that.

PhilETaylor:
delete now changed in https://github.com/joomla/joomla-cms/pull/32915/commits/1c893ceb5f75f67de407bec3c8b57d75c051e159 - Thanks.

libraries/src/Filesystem/File.php
```diff
@@ -174,10 +176,48 @@ public static function copy($src, $dest, $path = null, $useStreams = false)
...
+	 * @param   string  $filepath   The path to the file just written to, to flush from opcache
+	 * @param   boolean $force      If set to true, the script will be invalidated regardless of whether invalidation is necessary
+	 *
+	 * @return boolean TRUE if the opcode cache for script was invalidated/nothing to invalidate, or FALSE if the opcode cache is disabled.
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	public static function invalidateFileCache($filepath, $force = true)
+	{
+		if (ini_get('opcache.enable')
```
wilsonge, line 234: Can we micro-optimise a bit here? Set all the OPCache checks into a local function static so for all subsequent files we are only checking it's a PHP File? On larger extensions presumably that will provide a small saving.

richard67:
Hmm, there comes something into my mind. I remember that in past, before my webspace provider has started to enable mod_deflate, I had used an ugly hack for my private website to get js and css compressed: I processed these files with PHP and so could compress the output. Because this was a dangerous hack (there could be something in a js file which is a short tag for PHP), I later changed to statically gzipped files, before I later could use mod_deflate.

Am asking myself now if it can happen that we have opcache working on files which don't have the ".php" extension.

PhilETaylor:
@richard67 opcache is a cache of the compiled source code, not a cache of the output the code generates. So you can still use opcache and php files to output gz compressed js/css with no issues

libraries/src/Filesystem/File.php
PhilETaylor: The smallest saving ever :-)

My code is now highly similar to another cms running 30% of the Internet without that micro optimisation ;-)

richard67:
@PhilETaylor You are right, and I was maybe a bit paranoid. Just wanted to be sure we don't forget anything.

libraries/src/Filesystem/File.php
richard67: Maybe that's the reason why internet is sometimes slow here, because 30% of the websites running slow code? 😉 (joking)

richard67:
@PhilETaylor I should have remembered it earlier, but now I do.

We have a function for renaming files with incorrect case in file names on core updates here:

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_admin/script.php#L7175

This uses `rename` and `unlink` PHP functions directly.

The same applies to J3:

https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_admin/script.php#L2668

That function was once implemented by Sharky and later we took over and finished his work.

What shall we do there? Make it use functions from our File (or on J3 JFile) class? Or add opcache invalidation to that function in script.php?

And maybe we have other places where PHP functions are used instead of using our File class?

Another thing comes into my mind for J3, but I discuss it here because we discussed all the other stuff here, too, and not in the J3 PR: On J3 we might still have older PHP versions which don't have opcache but apc cache, whatever that is. Could that be? Do we have to handle that, too, on J3?

PhilETaylor:
Im far from finished with opcache cases. Searching for `file_put_contents`, `unlink`, `rename` are still on my to do list. 

This PR was just the start, and specifically to address extension installations (which happen more often than Joomla updates) and to address Nicolas's valid concerns that other projects had already addressed. 

The hope is to at least get the Joomla 3 version of this PR into Joomla 3.9.26. 

I dont care about 3rd party cache providers. PHP Opcache is shipped with PHP and is provided and supported by the PHP Team. Im not going to be investigating and invalidating all 3rd party caches, that's a rabbit hole you dont want to go down.

libraries/src/Filesystem/File.php
PhilETaylor: Micro optimisation now done.

libraries/src/Filesystem/File.php
```diff
@@ -174,10 +182,71 @@ public static function copy($src, $dest, $path = null, $useStreams = false)
...
+	 * @return boolean TRUE if the opcode cache for script was invalidated/nothing to invalidate,
+	 *                 or FALSE if the opcode cache is disabled or other conditions returning
+	 *                 FALSE from opcache_invalidate (like file not found).
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	public static function invalidateFileCache($filepath, $force = true)
+	{
+		if (self::canFlushFileCache() && '.php' === strtolower(substr($filepath, -4))
+		)
```
richard67:
```suggestion
		if (self::canFlushFileCache() && '.php' === strtolower(substr($filepath, -4)))
```

PhilETaylor: String. That passes the code style rules in Joomla 4, but breaks it in Joomla 3...

richard67:
@PhilETaylor Maybe it could make sense to do the removal of the `opcache_reset` calls here in this PR, too? These calls might complicate testing a bit if still present.

PhilETaylor:
already working on that, and some risky ones in the update process :scared:

administrator/components/com_joomlaupdate/restore.php
```diff
@@ -1630,7 +1631,19 @@ public function rmdir($directory)
 
 	public function rename($from, $to)
 	{
-		return @rename($from, $to);
+		$ret = @rename($from, $to);
+		$this->clearFileInOPCache($from);
```
richard67, line 1634: Should we maybe move the clear for the "from" file to before the rename, so the "from" file is handled like deleted files regarding order of processing?

PhilETaylor: Yup - good catch

libraries/src/Filesystem/File.php
```diff
@@ -321,6 +404,9 @@ public static function move($src, $dest, $path = '', $useStreams = false)
 				}
 			}
 
+			// Invalidate the compiled OPCache of the old file so its no longer used.
```
brianteeman:
```suggestion
			// Invalidate the compiled OPCache of the old file so it's no longer used.
```

ocd spellcheck

PhilETaylor:
/note to self... TURN OFF AGGRESSIVE OPCACHING WHEN TRYING TO EDIT PHP FILES - DOH! Time lost: 37mins...

richard67:
@PhilETaylor Should we port back all recent changes to the staging PR? Or should we ask release leads for opinions? Me personally I'd like it to be ported back, but I might be wrong.

PhilETaylor:
Im not too fussed with Joomla 3 anymore :) lets make Joomla 4 amazing :) 

Im not sure (because no one answers me) how much time we have for Joomla 3.9.26 to be released, and as that's 2 versions after the advertised "last version" in the Joomla 3.9 series... 

I guess some of the recent changes could be added to the Joomla 3 PR, but I dont want to add anything that will hamper it being merged without a fight.

PhilETaylor:
Also I would want the changes in the Akeeba provided update (restore.php restore_filnilization.php) tested EXTENSIVELY before release :)

libraries/src/Filesystem/File.php
```diff
@@ -134,6 +140,9 @@ public static function copy($src, $dest, $path = null, $useStreams = false)
 				return false;
 			}
 
+			self::invalidateFileCache($src);
```
richard67: Why do we invalidate the source, too, when copying, and not only the destination?

libraries/src/Filesystem/File.php
```diff
@@ -174,10 +183,71 @@ public static function copy($src, $dest, $path = null, $useStreams = false)
 				$ret = true;
 			}
 
+			self::invalidateFileCache($src);
```
richard67: Same as previous comment.

libraries/src/Filesystem/File.php
PhilETaylor: Because the source path would stay in opcache, and PHP would use it if that path is called again. A move can be thought of as a delete and a write.

richard67: Yes, for the move you are right. But here we are in the "copy" function.

PhilETaylor: ah ok - changing now

richard67: Do we even need to invalidate anything in the copy function? We just create a new file for the opcache and not change the old one.

PhilETaylor: Yes. See the big red warning on https://www.php.net/copy that states "If the destination file already exists, it will be overwritten.". So if we are REPLACING an existing file, then its "changed" and needs to be recompiled.

richard67:
@PhilETaylor ~~Do we even need to invalidate anything in the copy function? We just create a new file for the opcache and not change the old one.~~ Ah, ok, got the answer already.

PhilETaylor:
> Ah, ok, got the answer already. 

:-) https://github.com/joomla/joomla-cms/pull/32915#discussion_r606797553

libraries/src/Filesystem/File.php
```diff
@@ -304,12 +388,14 @@ public static function move($src, $dest, $path = '', $useStreams = false)
...
 
 				// Use FTP rename to simulate move
+				// Invalidate the compiled OPCache of the old file so it's no longer used.
 				if (!$ftp->rename($src, $dest))
 				{
 					Log::add(Text::_('JLIB_FILESYSTEM_ERROR_RENAME_FILE'), Log::WARNING, 'jerror');
 
 					return false;
 				}
+				self::invalidateFileCache($dest);
```
richard67: Why this one here? We have another one below in line 411. And the comment in line 391 above tells about the old file. Am confused.

PhilETaylor: Because Im an idiot.

libraries/src/Filesystem/File.php
```diff
@@ -321,6 +406,7 @@ public static function move($src, $dest, $path = '', $useStreams = false)
 				}
 			}
 
+			self::invalidateFileCache($dest);
```
richard67, line 408:
```suggestion
			self::invalidateFileCache($dest);

```

libraries/src/Filesystem/File.php
```diff
@@ -304,6 +388,7 @@ public static function move($src, $dest, $path = '', $useStreams = false)
 				$dest = Path::clean(str_replace(JPATH_ROOT, $FTPOptions['root'], $dest), '/');
 
 				// Use FTP rename to simulate move
+				// Invalidate the compiled OPCache of the old file so it's no longer used.
```
richard67:
```suggestion
```

libraries/src/Filesystem/File.php
```diff
@@ -534,6 +629,7 @@ public static function upload($src, $dest, $useStreams = false, $allowUnsafe = f
 			}
 			else
 			{
+				self::invalidateFileCache($src);
 				if (is_writable($baseDir) && move_uploaded_file($src, $dest))
```
richard67, line 634:
```suggestion

				if (is_writable($baseDir) && move_uploaded_file($src, $dest))
```

richard67: Last one, then CS should be fine.

richard67:
Will test later today or tomorrow or both if it takes that long.

I think we need to test at least following szenarios for the Joomla Update stuff:

- Updating a 4.0 Beta 7 without the patch of this PR applied to the update package built by drone for this PR.
- Updating a 4.0 Beta 7 with the patch of this PR applied to the update package built by drone for this PR.
- Updating from 3.10-dev to the update package built by drone for this PR.

4.0 Beta 7 for the first 2 because there will be some files and folders to be deleted.

It might also make sense to prepare some files to be renamed by that function here so that it should do something:

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_admin/script.php#L7175

And of course we have to test extension updates and saving changes in configuration.php.

We should test all that with realistic and with extreme opcache settings.

PhilETaylor:
Appreciated. Thanks.

richard67:
If we wanna be extremely safe with testing, we do all test on PHP 7.x and PHP 8.

I'll ask my boss in my paid job if she can give me 1 week holiday ;-)

richard67:
Ah and of course to be safe it also needs some test if all continues to work when opcache is not available or switched off.

PhilETaylor:
> when opcache is not available or switched off.

Insane web hosts ;-)

richard67:
Hmm I just see testing with using 4.0 Beta 7 and the patch applied will not work. Have to use current 4.0-dev or latest nightly instead.

PhilETaylor:
define "will not work" ?

richard67:
My 2nd scenario "Updating a 4.0 Beta 7 with the patch of this PR applied to the update package built by drone for this PR.": The patch of this PR can't really be applied to beta 7 because the modified function in script.php did not exist yet.

PhilETaylor:
ah ok, well that's not this PR's fault :)

richard67:
> ah ok, well that's not this PR's fault :)

Right, it was my scenario being wrong.

PhilETaylor:
Phew :)

administrator/components/com_admin/script.php
```diff
@@ -7193,8 +7193,8 @@ protected function fixFilenameCasing()
 			if ($newBasename !== $expectedBasename)
 			{
 				// Rename the file.
```
Quy, line 8162: Update comment?

administrator/components/com_joomlaupdate/restore.php
```diff
@@ -1620,6 +1621,7 @@ public function chmod($file, $perms)
 
 	public function unlink($file)
 	{
+		$this->clearFileInOPCache($file);
```
Quy, line 1622:
```suggestion
		$this->clearFileInOPCache($file);

```

administrator/components/com_joomlaupdate/restore.php
```diff
@@ -1630,7 +1632,19 @@ public function rmdir($directory)
 
 	public function rename($from, $to)
 	{
-		return @rename($from, $to);
+		$this->clearFileInOPCache($from);
+		$ret = @rename($from, $to);
+		$this->clearFileInOPCache($to);
```
Quy, line 1636:
```suggestion
		$this->clearFileInOPCache($to);

```

PhilETaylor: is this tabs/spaces again?! grrr

PhilETaylor: ah no - a space got it.

administrator/components/com_admin/script.php
PhilETaylor: This is renaming the file On Windows or Unix with only the incorrectly cased file... 

not sure what you are asking me to change?

Quy: I see the command `rename` changed to `move` so wondering if the comment should be changed. Now I understand it is not necessary. Thanks for the info.

PhilETaylor: not changed by this PR :)

richard67:
@PhilETaylor Silly me. I've looked at the wrong place. My scenario is valid, i.e. we can use Beta 7 for testing.

richard67:
@PhilETaylor Now after some stuff has been merged up from staging/3.10-dev into 4.0-dev, there is a conflict in script.php. Let me know if I shall help with solving it. It is my PR #32176 coming up from staging which causes this now.

PhilETaylor:
Looks like you took slashes out of the paths, the only bit I changed was the rename/unlink function calls and replaced them with Filesystem API calls. 

Screenshot 2021-04-05 at 20 20 26

richard67:
@PhilETaylor The slashes I have taken out because I have added them before to the array:

https://github.com/joomla/joomla-cms/pull/32176/files#diff-db7eb77540ff419bd7e6557d2779f4cf1e31158c20cb4b63dbbe8d6c426385adR2671-R2673

You can't see that when checking for conflicts only because there was no conflict.

So here where the conflicts are we have to take both changes together, yours and mine.

If you want I can do that for you here.

PhilETaylor:
Im back to the day job tomorrow, and still need to drive the other side of the island in an hour to collect the kid. So I'll not get to this, or any other PR's probably until 24 hours or more - sorry. Easter is over :)

richard67:
> Im back to the day job tomorrow, and still need to drive the other side of the island in an hour to collect the kid. So I'll not get to this, or any other PR's probably until 24 hours or more - sorry. Easter is over :)

@PhilETaylor No problem. I solve the conflict here and you just pull later from remote to local before continuing work.

Independently from the conflict I need to do more tests here, too, stay tuned, it may take some time.

PhilETaylor:
I also need to see how many of the 115 comments here have been, and have not yet been, merged to the Joomla 3 version of this PR. https://github.com/joomla/joomla-cms/pull/32918

richard67:
Current failures in drone are not caused by this PR here but by the 4.0-dev branch not being ok, and I've just updated this PR by that. I will update it again if the 4.0-dev branch is ok.

PhilETaylor:
Drone has been very unstable recently.

richard67:
@PhilETaylor Regarding integrate changes from here back to the staging PR: I discovered some weird problems when testing, of which I am not sure it they come from the opcache_reset calls being removed here, or if I do something wrong with my opcache configuration. Maybe we should wait with that until we know if we are ok here.

richard67:
@PhilETaylor Regarding drone: Right now we have a js cs error in the 4.0-dev branch which will be fixed by #33026 , and a merge conflict here in addition: https://github.com/joomla/joomla-cms/commit/321def11e90b7e9e2ffc9d5e6744c06a58b4a3b5#r49127384 ... so not the usual instability of drone only.

But as said, I will have an eye on it so you don't need to care.

PhilETaylor:
>  some weird problems

I need more than that context haha :) 

You could be like me, and spend an hour wasting time writing new code in my "other project", refreshing and trying to work out why the hell nothing was working on your "other project", and then realise that you still had opcode caching set to very aggressive and never revalidating the PHP files... that was a wasted hour...

richard67:
> You could be like me, and spend an hour wasting time writing new code in my "other project", refreshing and trying to work out why the hell nothing was working on your "other project", and then realise that you still had opcode caching set to very aggressive and never revalidating the PHP files... that was a wasted hour...

No, I was testing this PR here, with and without the one from the framework in addition, and with the very aggressive opcache settings you have proposed above in comments.

I've noticed that when I update a 4.0-dev to the update package for this PR, using the custom URL provided by drone package step, then after the update the usual database error due not not matching CMS version was not shown (like it is expected when using these packages for update). When I then "manually" triggered an opcache_reset, the CMS version was updated, and that database error was shown.

When I started before such an update without the patch of this PR applied, I got after the update to the package with this PR an alert with a warning that the restoration.php file could not be deleted:

failed-delete-restoration-php

But in fact the file has been deleted.

When starting with the patch of this PR applied, that alert was not shown.

In both cases I had the issue with the CMS version not being up to date in backend (but on filesystem it was, so backend was using the old opcached info still.

Another issue I've observed when installing patchtester v 4.0.0, which doesn't show the badges in the pulls list due to missing changes for bs5, and then updating that to v 4.0.1 were this issue was fixed, and then checking again the pulls list, and it was not ok. Only when "manually" resetting opcache, I was the changes.

For manual opcache reset I use a PHP script in my Joomla root which calls `opcache_reset()` and which I've navigated to in my browser.

With opcache lopg severity 4 I could see the corresponding debug messages.

I couldn't drag down yet to where goes what wrong for me.

I think for today I'm done, will continue testing in the next days.

PhilETaylor:
I have a different Mac at the (physical, not home) office so will rebuild from scratch there tomorrow and test like you, as that Mac has not been used in the development of this pr at all. 

It's possible there are more file operations to take into account - remembering that joomla updates are extracted differently (standalone with restore.php) to extension updates (extracted by joomla)

richard67:
@PhilETaylor Now as the 4.0-dev branch has been fixed so CI tests are passing again, it could make sense to update this branch to current 4.0-dev. If you want I can do this for you to save you some work. But I'm asking you for permission first because it may disturb your workflow. As far as I can see, I have the necessary permission on your GitHub repo.

PhilETaylor:
Go for it - I can always pull again.

richard67:
Done. That will help us with testing core updates using the update package or custom URL built by drone because those will be not too old.

richard67:
@PhilETaylor Having your PR #33040 in mind now, I'm asking myself if it could make sense to add come cache clearing to those places where we remove the opcache_reset calls here. I don't think it explains my testing results so far because I had caching switched off in global configuration. But I think it might be necessary since we don't invalidate the cached files, only the "originals", while an opcache reset also will cause a revalidation of the cached files later, right?

PhilETaylor:
The cache files in the cache folder will be invalidated by my other PR .. https://github.com/joomla/joomla-cms/pull/32993

PhilETaylor:
Maybe I should merged the 2 PRs as there have been no other interest from anyone on the other one.

richard67:
> The cache files in the cache folder will be invalidated by my other PR .. #32993

@PhilETaylor Yes, that PR makes sure we invalidate them when we purge the cache. But what I mean is that we have removed calls to opcache_reset here, but we might not have calls to reset cache at all these places. As far as I know we clear cache after core updates, and I guess we do that also after extension updates or deletions or installations, but about the latter I'm not sure. And who knows where else we might need to clear cache now where we didn't need it before.

PhilETaylor:
>  And who knows where else we might need to clear cache now where we didn't need it before. 

that's the whole point. we should NEVER call `opcache_reset` ever. We need to trawl through everything and ensure that its using `opcache_invalidate` - luckily Joomla uses the `Filesystem` API mainly and so we are good in those places.

PhilETaylor:
> . We need to trawl through everything

well, searching for `unlink` `move` `copy` `rename` `file_put_contents` etc...

richard67:
@PhilETaylor I have also seen some places where there are options for stream and not stream, don't remember now where. Does this use anything from the phar stream wrapper which still has the opcache_reset in its belly?

PhilETaylor:
The phar stuff we cannot do anything about, that's a 3pd library.

richard67:
> The phar stuff we cannot do anything about, that's a 3pd library.

Do we use that somewhere, and if so, could we just stop to use it?

PhilETaylor:
IIRC it was a security fix and that's why we have the Phar Wrapper
https://developer.joomla.org/security-centre/781-%2020190502-core-by-passing-protection-of-phar-stream-wrapper-interceptor.html

I honestly dont believe anyone is using phar's with Joomla though :)

richard67:
We support so much stuff which nobody really uses ;-)

PhilETaylor:
This PR seems to have stalled - to be clear for those reading, you can test and provide your test results here, this PR is needed to future proof Joomla on decent web hosts that use decent opcache configurations (and those with crazy opcache configurations)

richard67:
@PhilETaylor Not stalled. I have it on my list for deeper testing. My previous tests were partly not ok, but maybe due to my opcache settings were not right. Stay tuned.

PhilETaylor:
This shit has to stop! 

Screenshot 2021-04-24 at 21 58 03

rdeutz:
> This shit has to stop!
> 
> Screenshot 2021-04-24 at 21 58 03

Agree.

@SharkyKZ I have given you a 7 days break, your negative behaviour is not acceptable. The only thing you contribute are thumbs down without providing more information.

PhilETaylor:
Thank you. To be clear this was continued negative behaviour was reported to Brian Mitchel and Marco Dings weeks ago and both have refused to even reply to my email or acknowledge it, or follow ups but Luca acknowledged the email. 

Joomla Leadership doesn't take contributor safety seriously at all and the CoC process is just a joke! 

Thank you for having the guts to take action.

PhilETaylor:
Merged in the FTP Layer removal stuff that conflicted, resolved conflicts and force pushed again.

nikosdion:
Is there any reason why this is not RTC yet? OPcache not being cleared for .php files being written during an update **DOES** cause issues in any non-trivial extension's post-installation code. See https://www.akeeba.com/news/1744-advice-for-install-probs-with-march-2021-software-releases.html it is under “PHP opcache can cause weird issues”. I can still reproduce this issue after rewriting my extensions using core Joomla 4 MVC and Bootstrap 5 so any notion that the problem was ‘magically’ caused by my frameworks not only flies against common sense but is categorically disproven.

This problem also affects Global Configuration (com_config) writing the configuration.php file. Look in the forums how many times people are told their configuration.php must be unwriteable even though there is no error message telling them this is the case and you will see just how often Joomla not handling OPcache scores an own goal and loses users.

This not only needs to be set RTC and merged a.s.a.p. in Joomla 4 but also back ported to Joomla 3. It's not 2011 anymore when opcode caching was something exotic. It's 2021 and you'd be hard pressed to find a server which doesn't have OPcache turned on by default.

The only reason this has not been a big issue _yet_ is that the default revalidate time is 2 seconds, Joomla 3 is slow, people run it on PHP 5.6 or earlier which is dead slow and stuff it with a million plugins which are make it even slower. This made operations take substantially more than the default 2 seconds of OPcache revalidation frequency. Joomla 4 is much faster, it requires modern PHP versions which are faster and even the plugins (Events) architecture mostly guards against plugins bogging down the site. We will see a heck of a lot more operations going into the few hundreds milliseconds to two seconds which will expose the OPcache issues.

PhilETaylor:
I wrote a Joomla 3 PR at the same time as this one and it got little to no attention and I closed it - it's here on Gh somewhere, probably needs dusting off...

richard67:
@nikosdion I had tested that PR and it failed for me when I had used the aggressive opcache settings which Phil had proposed, see my comment above https://github.com/joomla/joomla-cms/pull/32915#issuecomment-813612685 . With normal opcache settings it worked for me, but I also was not able to reproduce any issue.

If you CAN reproduce an issue: Why don't you test this PR? It would for sure help to come forward with it.

PhilETaylor:
Joomla 3 version https://github.com/joomla/joomla-cms/pull/32918

wilsonge:
I'm interested in merging this. But this is way too scary to just merge on review given the level of things touched. I fully agree it's something that needs testing however :)

PhilETaylor:
The problem is and has been historically getting developers of a calibre that actually understand PHP and how it really works in order to test complex things like this. 

Anyone can test pixel pushing or accessibility... 

But then this is Joomlas whole issue moving forward, attracting and retaining top quality developers to contribute

richard67:
> The problem is and has been historically getting developers of a calibre that actually understand PHP and how it really works in order to test complex things like this.
> 
> Anyone can test pixel pushing or accessibility...
> 
> But then this is Joomlas whole issue moving forward, attracting and retaining top quality developers to contribute

@PhilETaylor I am aware of the fact that I am not a PHP expert. I have knowledge in other things, like SQL. But to say that anyone who isn't a PHP expert just does pixel pushing or accessibility (and so in between the lines also say accessibility is nothing really important) .. that hurts.

PhilETaylor:
Taking everything out of context and believing it's a personal attack is also a trait of this project.

I would hope you knew I was talking about the project in general and not you specifically 

There are very few people at Nicolas's level of the php game, and those that are, tend to be in laravel or symfony and shun joomla for many reasons.

brianteeman:
Subtlety in comments are often lost in translation

richard67:
"Anyone can test pixel pushing or accessibility" .. what is subtile in this? What is lost in translation?

brianteeman:
I was being generous

PhilETaylor:
Serves me right for sitting on the loo and typing... let me finish the sentence


"Anyone can test pixel pushing or accessibility, and can follow a list of instructions on how to test and see if something that was broken, is now changed and is now fixed when it was not fixed before, but what this PR needs is someone who understands the internals of PHP and how opcaches work IN THE REAL WORLD (not just on your local environment with a single tenant test site but on a server with 1000 clients and websites all competing for resources) and that understands what each opcache setting does and how it effects the stack and how each setting interacts with a reset of the opcache - a technology that is ("normally" but there are exceptions) a server wide configuration and can often be badly/wrongly/screwed up by webhosts manipulating the settings and setting up crap servers too. Joomla should could and must be able to run in all kinds of different good bad and screwed up opcache configurations. 

It took WordPress 5 years to perfect a PR, it's not rocket science but it takes understanding, testing and people who have experience in this area.

By your own admission @richard67 this is not you, but I'm grateful for the testing you have already done to get it this far 

We are over 100 comments in and yet no further forward.

nikosdion:
@richard67 Phil wrote this PR after he read my post-mortem which I linked to in my previous comment. Between these two events a member of the Joomla leadership said publicly and in no uncertain terms that the problems I faced where of my own making, blaming Joomla — even with cold, hard facts — is deflecting my responsibility, called for banning my extensions and implementing them in core.  In response, I published [this](https://www.akeeba.com/news/1746-working-around-joomla-s-broken-extensions-updater.html) which gives technical details of all the very real bugs Joomla has, the ones I have been reporting in private for eight years and nobody wants to even acknowledge. I know you were unaware so I won't begrudge you for telling me to test. I am just telling you why I can't and won't.

As long as the environment surrounding contributing to Joomla remains toxic I cannot be involved in reporting issues I encounter or file PRs to fix them _unless_ I can also reproduce them with a third party extension. This is unlikely as I use very little extensions beyond mine. Since I am not aware of an extension other than mine that can be used to reproduce this issue I cannot in good conscience file a test, even though I can test it and find it to be working. Bummer.

@wilsonge I don't buy the “this thing touches too much” excuse. The BS4 to BS5 PR was touching FAR MORE code than this PR. The only difference is that HTML output can be understood by everyone and their dog whereas the OPcache requires people who understand PHP and can think beyond the HTML output.

I can only tell you the following in lieu of a test.

I have reproduced this issue on March 2nd, 2021 on my local server and three separate servers of my clients. I have confirmed that invalidating the OPcache for installed .php files solves this issue on my local server and three separate servers of my clients. FWIW the reproduction instructions for a local server is exactly what Phil told you.

Akeeba Backup 8 has for the last **six (6) months** been doing the same thing as this PR in its post-installation code. I go through all the files installed by Joomla and invalidate OPcache for them. I don't even check for `restricted_api` which is a reasonable check to add. 

Phil's approach to `restore.php` is something I am going to use on Akeeba Kickstart.

As far as I am concerned, this is a valid issue and a valid fix. The issue has been reproduced and the fix confirmed on both local and live servers owned by different people. The PR uses the same technique I have been using in production for 6 months, meaning it's ran on tens of thousands of sites without a single ticket about it. This is extraordinary. With the installed base I have, extremely rare issues are guaranteed to get at least one to two tickets in any given 6 month period. Having none in six months means that, even though I am not checking for opcache.restricted_api, nothing broke on dozens of thousands of sites.

Take this as you like. I just can't and won't file an official test on this PR because politics would kill it. That's all.

wilsonge:
@richard67 are you able to test specifically that the `File::move` in the script.php for the case sensitive file name works on the OS's you worked on please? I'll test OSX this evening too.

richard67:
> @richard67 are you able to test specifically that the `File::move` in the script.php for the case sensitive file name works on the OS's you worked on please? I'll test OSX this evening too.

I can only test if it works with normal opcache settings. Tesing with extreme settings like suggested by @PhilETaylor has failed for me in past.

But I can't promise I have the time today.

wilsonge:
It's fine. I don't need extreme settings. Just some validation that across environments that the case sensitive rename doesn't break - because we had so many iterations of that script.

richard67:
@wilsonge If it will not go into 4.0.0 today I assume it is for 4.0.1, right? At least that's my information here, so I won't hurry with tests.

richard67:
@PhilETaylor I've allowed myself to update the branch of this PR to latest 4.0-dev to force a new package build by drone so people can test using the update package. The previous build for this PR was so long ago that the packages meanwhile have been deleted.

richard67:
@wilsonge Test the files renaming on update on Linux (shared hosting with opcache enabled and working) for different cases (files with right and wrong case or only the ones with wrong case present). It worked. Windows I will test later. For testing exotic OS (Apple) it's your turn. You can use the update package built by drone. I've triggered new builds just before by a branch update.

richard67:
I have tested this item :red_circle: unsuccessfully on 15ae8a275888fe882be6575a73d5665371eac811<br /><br />On Linux all worked fine.

But on Windows, the filename casing fix fails with an internal server error and following in the PHP error log:

```
[21-Aug-2021 17:30:49 Europe/Berlin] PHP Fatal error:  Uncaught Error: Call to undefined method Joomla\CMS\Filesystem\File::move() in C:\xampp\htdocs\joomla-cms\administrator\components\com_admin\script.php:8163
Stack trace:
#0 C:\xampp\htdocs\joomla-cms\administrator\components\com_admin\script.php(7393): JoomlaInstallerScript->fixFilenameCasing()
#1 C:\xampp\htdocs\joomla-cms\administrator\components\com_joomlaupdate\restore_finalisation.php(75): JoomlaInstallerScript->deleteUnexistingFiles()
#2 C:\xampp\htdocs\joomla-cms\administrator\components\com_joomlaupdate\restore.php(5430): finalizeRestore('C:\\xampp\\htdocs...', 'C:/xampp/htdocs...')
#3 {main}
  thrown in C:\xampp\htdocs\joomla-cms\administrator\components\com_admin\script.php on line 8163
```

I have renamed the 2 files in 2 steps before updating current 3.10-dev to the patched package built by drone for this PR just today, so they looked as follows:

```
C:\xampp\htdocs\joomla-cms>dir /b libraries\src\Filesystem\Support\stringcontroller.php libraries\src\Form\Rule\subformrule.php
Stringcontroller.php
SubFormRule.php

C:\xampp\htdocs\joomla-cms>
```

I think (have to check) that it needs to add a function `File::move()` to restore_finalisation.php, too.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32915">issues.joomla.org/tracker/joomla-cms/32915</a>.</sub>

richard67:
@PhilETaylor You have to add a move function to that class here: https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_joomlaupdate/restore_finalisation.php#L87-L128 . If you want I can do that and make a PR for your branch.

richard67:
@PhilETaylor See https://github.com/PhilETaylor/joomla-cms/pull/10 for a fix.

PhilETaylor:
Ta. Back home in 2 hrs

richard67:
Hmm, now with the fix from my PR it works on the Windows host where it failed before, and it still works fine on my Linux shared hosting with and without opcache enabled.

But on my local Ubuntu Linux with PHP 7.4 and opcache disabled, updating from 3.10-dev to the patched update package from this PR fails.

I get a blank page soon after the update has started. In PHP error log I see:

```
[Sun Aug 22 01:07:46.583916 2021] [php7:error] [pid 74204] [client 192.168.98.1:63993] PHP Fatal error:  Uncaught RuntimeException: Library path /libraries/joomla cannot be found. in /home/richard/lamp/public_html/joomla-cms-3.10-dev/libraries/loader.php:344
Stack trace:
#0 /home/richard/lamp/public_html/joomla-cms-3.10-dev/libraries/loader.php(490): JLoader::registerPrefix()
#1 /home/richard/lamp/public_html/joomla-cms-3.10-dev/libraries/import.legacy.php(52): JLoader::setup()
#2 /home/richard/lamp/public_html/joomla-cms-3.10-dev/administrator/includes/framework.php(17): require_once('/home/richard/l...')
#3 /home/richard/lamp/public_html/joomla-cms-3.10-dev/administrator/index.php(40): require_once('/home/richard/l...')
#4 {main}
  thrown in /home/richard/lamp/public_html/joomla-cms-3.10-dev/libraries/loader.php on line 344, referer: https://www.joomla-310-dev.vmubu01.vmnet2.local/administrator/index.php?option=com_joomlaupdate&task=update.install&5bf0ac822cd9ac38a28f72a495a40408=1
```

Using a regular update to 4.0.0 works with the same update method (Live Update) on the same environment.

Will continue to investigate what the reason could be.

**_Update:_** Same environment, same kind of git clone just in another folder which has no dot in the path (i.e. Joomla root is `/home/richard/lamp/public_html/test-1/` instead of `/home/richard/lamp/public_html/joomla-cms-3.10-dev/`: It works.

I wouldn't think it's related to this PR, but without this PR it works on the one with dot, too.

The PR is based on current 4.0-dev, the other test was using a 4.0.0 update package. Maybe something happened between 4.0.0 and now in the 4.0-dev branch which causes that problem? I don't think so, but I will check with an update package based on current 4.0-dev later, or tomorrow with a nightly.

**_Update:_** What I have forgotten to mention: When reloading the white page after that error occurred, the update runs and ends with success. The update log shows that all actions and all SQL statements of the update have been performed.

**_Update:_** The white page is shown after the first log entry in `administrator/logs/joomla_update.php` with message "Starting installation of new version." has been made, i.e. it happens at the very beginning of the update. After page reload the update continues. And as said, it doesn't happen with a regular nightly build update package from tonight, only with the package from this PR.

richard67:
Hmm, right now it worked where I had the problem before ... so maybe my issue on that one environment is related to that environment ... possibly an issue with file or folder permissions or chmod privileges.

richard67:
I have not tested this item.<br /><br />I just see my previous, invalid test result is still counted despite of the later commits which fixed that error.

Setting to "not tested" as long as I'm investigating with the issue I have now with one of my testing environments.

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32915">issues.joomla.org/tracker/joomla-cms/32915</a>.</sub>

PhilETaylor:
Are you sure these issues are related to this PR, and not just problems with the upgrade process? ;-)

richard67:
> Are you sure these issues are related to this PR, and not just problems with the upgrade process? ;-)

@PhilETaylor If I was sure about anything yet, I would not investigate still ;-)

PhilETaylor:
> But on my local Ubuntu Linux with PHP 7.4 and opcache disabled, updating from 3.10-dev to the patched update package from this PR fails.

If opcache is disabled then that pretty much rules out this PR, as everything in this PR is pretty much reliant on opcache being enabled, and if not no harm should be caused :)

richard67:
> as everything in this PR is pretty much reliant on opcache being enabled, and if not no harm should be caused :)

@PhilETaylor Except of the changes where calls to native PHP functions have been replaced by calls to methods of the File and Folder classes of the Filesystem library and the latter include calls to Path::clean. Such changes may have an effect regardless of opcache.

PhilETaylor:
Those methods have been used in Joomla for a long time though and I would be surprised if they had considerable bugs in them :)

PhilETaylor:
> Update: The white page is shown after the first log entry in administrator/logs/joomla_update.php with message "Starting installation of new version."

At this point the code running would be 3.10 right? without any of this PR contained in it? its not until after the restore/extraction that the Joomla 4 code is run containing my changes...

PhilETaylor:
> At this point the code running would be 3.10 right? without any of this PR contained in it? its not until after the restore/extraction that the Joomla 4 code is run containing my changes...

Im correct, this " "Starting installation of new version." log entry is before the restoration file is created, so way before my code is run for the first time.

```
try
		{
			Log::add(Text::_('COM_JOOMLAUPDATE_UPDATE_LOG_INSTALL'), Log::INFO, 'Update');
		}
		catch (\RuntimeException $exception)
		{
			// Informational log only
		}

		/** @var \Joomla\Component\Joomlaupdate\Administrator\Model\UpdateModel $model */
		$model = $this->getModel('Update');

		$file = $this->app->getUserState('com_joomlaupdate.file', null);
		$model->createRestorationFile($file);
```

richard67:
@PhilETaylor Other question: The proxy function I have added with my PR https://github.com/PhilETaylor/joomla-cms/pull/10 here has only 2 parameters: https://github.com/joomla/joomla-cms/pull/32915/files#diff-d0a9b340bfb7e20bbd4f1c8603bc3efc4eeca71c8cd65b078a6ae4c166f3ad01R131

That's ok as long as in script.php it's used with only 2 parameters. 

But the function which is proxied has 2 additional parameters: https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Filesystem/File.php#L264

Shall we add them to the proxy? If so, what to do with the `$useStreams`? Shall we handle that somehow or just silently ignore it?

richard67:
@PhilETaylor Regarding my issue: I think you are right and it's not related to your PR. I just want to solve it and finally check if everything works before I give your PR a good test.

PhilETaylor:
> Shall we add them to the proxy? If so, what to do with the $useStreams? Shall we handle that somehow or just silently ignore it?

Nics code doesnt use streams so its ok to ignore that. The restoration code, being standalone from Joomla on purpose, should be as compact, and featureless as possible. At the end of the day all it needs to be able to do is extract a zip file without erroring, and then return control to Joomla to cleanup.

PhilETaylor:
We all know the `restore.php` is well overdue for a complete rewrite, but none of us are brave enough for that challenge, and maybe end up breaking Joomla update completely  :)

richard67:
> > Shall we add them to the proxy? If so, what to do with the $useStreams? Shall we handle that somehow or just silently ignore it?
> 
> Nics code doesnt use streams so its ok to ignore that. The restoration code, being standalone from Joomla on purpose, should be as compact, and featureless as possible. At the end of the day all it needs to be able to do is extract a zip file without erroring, and then return control to Joomla to cleanup.

@PhilETaylor So what do you finally suggest? Is https://github.com/joomla/joomla-cms/pull/32915/files#diff-d0a9b340bfb7e20bbd4f1c8603bc3efc4eeca71c8cd65b078a6ae4c166f3ad01R131 ok as it is now, or shall I add the other 2 parameters, too, handle the `$path` parameter right and ignore the `$useStreams`, in order to have the same signature and in case if someone in future wants call `File::move` in script.php with such parameters?

PhilETaylor:
I would say leave it. Less we change less we can break. If we add the `path` then we also then need to clean the path and provide `Path::clean`. Its not needed that ever before. 

> if someone in future wants call File::move in script.php with such parameters

Then let them add them then at that point. `File::move` in the update process is not meant to be a fully featured API like the one in the Filesystem namespace. Its mean to be a minimised version that can run standalone to Joomla, for the purposes of extracting the zip file only.

richard67:
@PhilETaylor My last issue was definitely caused by my testing environment. It could have been some sticky bit having been set by mistake on some folders, together with wrong group maybe. No idea how that could happen. Am just finishing a few tests for the installer. Looks good so far with opcache on and also with off.

richard67:
> Then let them add them then at that point. `File::move` in the update process is not meant to be a fully featured API like the one in the Filesystem namespace. Its mean to be a minimised version that can run standalone to Joomla, for the purposes of extracting the zip file only.

Agree.

PhilETaylor:
There are also 19 copies of 

```
/**
 * Akeeba Restore
 * A JSON-powered JPA, JPS and ZIP archive extraction library
 *
 * @copyright   2008-2017 Nicholas K. Dionysopoulos / Akeeba Ltd.
 * @license     GNU GPL v2 or - at your option - any later version
 * @package     akeebabackup
 * @subpackage  kickstart
 */
```

in this file... I think we could also trim that while still being compliant ;-)

richard67:
I have tested this item :white_check_mark: successfully on 15ae8a275888fe882be6575a73d5665371eac811<br /><br />Have tested on diverse environments:
- IONOS shared hosting on Linux with opcache enabled (file cache only)
- Linux without opcache enabled
- Windows without opcache enabled

The renaming of files with wrong case in script.php still works on Linux and Windows.

Installing, updating and uninstalling extensions works.

I'm still worried about certain thumbs down, maybe I've missed something. But as the author of the thumbs down will not explain the reason there is nothing I can do about that.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32915">issues.joomla.org/tracker/joomla-cms/32915</a>.</sub>

richard67:
@wilsonge I've tested the rename stuff on Linux on Windows. Would be cool if you could check on IOS because that was the most weird one.

PhilETaylor:
> I'm still worried about certain thumbs down, maybe I've missed something. But as the author of the thumbs down will not explain the reason there is nothing I can do about that.

The thumbs down from my stalker were applied moments after this PR was first opened, it should be noted that this PR has been drastically improved since then though 44 revisions.

richard67:
> > I'm still worried about certain thumbs down, maybe I've missed something. But as the author of the thumbs down will not explain the reason there is nothing I can do about that.
> 
> The thumbs down from my stalker were applied moments after this PR was first opened, it should be noted that this PR has been drastically improved since then though 44 revisions.

@PhilETaylor Yes, that's my hope, that the reasons for it were fixed on the way.

wilsonge:
@PhilETaylor can we just now change the installation to use the Filesystem method in the PR please https://github.com/joomla/joomla-cms/blob/4.0-dev/installation/src/Model/CleanupModel.php#L33-L46

PhilETaylor:
@wilsonge well ultimately that block of code just needs removing as Folder::delete in turn calls File::delete on all files in the folder, and the File::delete method is now "opcache aware" and invalidates the opcache before deleting the file.

Done.

wilsonge:
Thanks!

PhilETaylor:
> Merged #32915 into 4.0-dev.

/me runs and hides ....

PR #32918, opened 2021-03-28
[3] Ensure opcache file invalidate on file copy & delete

PhilETaylor:
This is part 2 of https://github.com/joomla/joomla-cms/issues/32592 and directed at Joomla 3

### Summary of Changes

When PHP is configured to use modern Opcache, PHP will attempt to hold the compiled version cached in memory. 

There are many ways to configure opcache invalidations, including never revisiting the source PHP to compile it again ever (if [opcache.validate_timestamps](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.validate-timestamps) is disabled for example) 

To fully ensure that files that Joomla writes to the hard disk are taken into account by a server with a correct configured (or badly configured) opcache, we need to invalidate files from the opcache after writing a new version to the disk. 

Joomla has previously used `opcache_reset` to do this, but its clear this is not good enough.

### Testing Instructions

Hard to test unless you really know what you are doing and can reconfigure all your PHP stack to include opcaching. Also some of the edge cases this fixes will only become clear if you are an extension developer mass distributing extensions. 

Install some extensions - nothing should break. 

You can install extensions

You can install extensions and opcache for each file is invalidated

none

// cc @nikosdion

### DIFF
--- a/libraries/src/Filesystem/File.php
+++ b/libraries/src/Filesystem/File.php
@@ -27,6 +27,12 @@ use Joomla\CMS\Client\FtpClient;
 class File
 {
 	/**
+	 * @var    boolean  true if OPCache enabled, and we have permission to invalidate files
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected static $canFlushFileCache;
+
+	/**
 	 * Gets the extension of a file name
 	 *
 	 * @param   string  $file  The file name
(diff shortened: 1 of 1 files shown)


richard67:
@PhilETaylor Unit tests failing:

```
There was 1 error:
1) JFilesystemPatcherTest::testApply with data set "Test delete" ('Index: lao\n==================...ames.\n', 'C:\projects\joomla-cms\tests\...atcher', 0, array('The Way that can be told of i...ames.\n'), array(null), 1, false)
Undefined index: extension
C:\projects\joomla-cms\tests\unit\core\helper.php:55
C:\projects\joomla-cms\libraries\src\Filesystem\File.php:196
C:\projects\joomla-cms\libraries\src\Filesystem\File.php:272
C:\projects\joomla-cms\libraries\src\Filesystem\Patcher.php:187
C:\projects\joomla-cms\tests\unit\suites\libraries\joomla\filesystem\JFilesystemPatcherTest.php:935
FAILURES!
```

See https://ci.appveyor.com/project/release-joomla/joomla-cms/builds/38443204/job/es3oo3tjn6285q8v?fullLog=true .

libraries/src/Filesystem/File.php
```diff
@@ -172,10 +174,33 @@ public static function copy($src, $dest, $path = null, $useStreams = false)
...
+	 *
+	 * @return void
+	 */
+	public static function invalidateOpcache($file)
+	{
+		if (function_exists('opcache_invalidate'))
+		{
+			$info = pathinfo($file);
+
+			if ($info['extension'] === 'php')
```
richard67:
```suggestion
			if (isset($info['extension']) && $info['extension'] === 'php')
```
From [PHP docs](https://www.php.net/manual/en/function.pathinfo.php): "Note: If the path does not have an extension, no extension element will be returned (see second example below)."

richard67: If my suggestion here is correct, it should also be applied on the J4 PR. No idea why tests pass there, maybe they don't have an example without an extension.

PhilETaylor:
Thanks - just got to the office so will update the Joomla 4 and [framework version](https://github.com/joomla-framework/filesystem/pull/37) of this.

PhilETaylor:
Screenshot 2021-03-29 at 15 53 51

Would it even be Monday without another drive by thumbs down from @SharkyKZ without any further comment on why he is upset this time.

richard67:
@PhilETaylor Did his thumb down come before or after the isset check was added? If before, then it maybe was that.

PhilETaylor:
It's just one of many PRs like that ... I'm not the only one complaining about it either, he has been doing it everywhere - it's so unhelpful

PhilETaylor:
This is ready for testing now.

PhilETaylor:
Renamed method as requested by https://github.com/joomla/joomla-cms/pull/32915#issuecomment-811947634

PhilETaylor:
So @SharkyKZ has taken it upon himself to close all my PRs without explaining why...

And you wonder why people dont contribute.

richard67:
For further discussion I've decided to comment in the J4 PR #32915 .

PhilETaylor:
method updated here now too

PhilETaylor:
Abandoning this for Joomla 3. 

The Joomla 4 version of this has received much more attention, refinement and testing and comment and fine tuning. 

Joomla announced that Joomla 3.9.24 was the last version of Joomla 3 to be released (despite 3.9.25 and 3.9.26 already here, and a milestone for 3.9.27 already starting to gather issues), there is no point fixing this in Joomla 3 - but lets get it right for Joomla 4.

PR #32924, opened 2021-03-29
[4.0] Apostrophe in alias

brianteeman:
Add an additional filter to the alias generator to remove any apostrophe instead of replacing it with a dash. This has annoyed me forever and I waste so much time doing this manually for every alias

### Example
For example the alias for an article "Brian's Pull Request"

### before
"brian-s-pull-request"

### after
"brians-pull-request"

**Note** Includes unit test

### DIFF
--- a/libraries/src/Filter/OutputFilter.php
+++ b/libraries/src/Filter/OutputFilter.php
@@ -90,6 +90,9 @@ class OutputFilter extends BaseOutputFilter
 		// Trim white spaces at beginning and end of alias and make lowercase
 		$str = trim(StringHelper::strtolower($str));
 
+		// Remove any apostrophe. We do it here to ensure it is not replaced by a '-'
+		$str = preg_replace('/\'/', '', $str);
+
 		// Remove any duplicate whitespace, and ensure all characters are alphanumeric
 		$str = preg_replace('/(\s|[^A-Za-z0-9\-])+/', '-', $str);
 
--- a/libraries/src/Filter/OutputFilter.php
+++ b/libraries/src/Filter/OutputFilter.php
@@ -91,7 +91,7 @@ class OutputFilter extends BaseOutputFilter
 		$str = trim(StringHelper::strtolower($str));
 
 		// Remove any apostrophe. We do it here to ensure it is not replaced by a '-'
-		$str = preg_replace('/\'/', '', $str);
+		$str = str_replace("'", '', $str);
 
 		// Remove any duplicate whitespace, and ensure all characters are alphanumeric
 		$str = preg_replace('/(\s|[^A-Za-z0-9\-])+/', '-', $str);
--- a/tests/Unit/Libraries/Cms/Filter/OutputFilterTest.php
+++ b/tests/Unit/Libraries/Cms/Filter/OutputFilterTest.php
@@ -68,7 +68,7 @@ class OutputFilterTest extends UnitTestCase
 			'joomlas-version',
 			OutputFilter::stringURLSafe('joomla\'s version'),
 			'Should remove apostrophe from the string'
-		)
+		);
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


richard67:
@brianteeman Wouldn't it be better just to add the apostrophe to the list of allowed characters in the existing `preg_replace` below the one you are adding?

I.e. change

`$str = preg_replace('/(\s|[^A-Za-z0-9\-])+/', '-', $str);`

to

`$str = preg_replace('/(\s|[^A-Za-z0-9\-\'])+/', '-', $str);`

instead of adding the new `preg_replace`.

brianteeman:
because its not an allowed character. it just needs to be removed - not replaced

richard67:
> because its not an allowed character. it just needs to be removed - not replaced

Ah, I forgot the removal, silly me.

PhilETaylor:
as you are not using a regular expression, it would be better PHP to use [str_replace](https://www.php.net/manual/en/function.str-replace.php) (Im guessing more performant too, the tiniest bit)

libraries/src/Filter/OutputFilter.php
```diff
@@ -90,6 +90,9 @@ public static function stringURLSafe($string, $language = '')
 		// Trim white spaces at beginning and end of alias and make lowercase
 		$str = trim(StringHelper::strtolower($str));
 
+		// Remove any apostrophe. We do it here to ensure it is not replaced by a '-'
+		$str = preg_replace('/\'/', '', $str);
```
PhilETaylor: `$str = str_replace("'", '', $str);`

toivo:
I have tested this item :white_check_mark: successfully on d83aacddf0caf3ea1fa3da350b57a4ea6f145136<br /><br />Tested successfully in Beta8-dev of 29 March in Wampserver 3.2.4 using PHP 8.0.2.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32924">issues.joomla.org/tracker/joomla-cms/32924</a>.</sub>

libraries/src/Filter/OutputFilter.php
richard67: Silly me, I should have seen that, too.

alikon:
I have tested this item :white_check_mark: successfully on d83aacddf0caf3ea1fa3da350b57a4ea6f145136<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32924">issues.joomla.org/tracker/joomla-cms/32924</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32924">issues.joomla.org/tracker/joomla-cms/32924</a>.</sub>

simbus82:
But it is valid only for english.
Many languages use the apostrophe with the contraction of definite articles before words (elision), so removing the dash (-) generate some problems.

IT
"L'**acqua** è pulita" -> "**lacqua**-e-pulita" (the word "lacqua" not exists).
Same for: l'uomo, l'aria, l'ora, l'eternità, l'amico, l'onestà, ecc. and all words that start with vowels.

FR
"l'eau" -> leau
"l'homme" -> lhomme
etc. Same as Italian.

Magento, Prestashop, Drupal, Typo3, and many other doesn't remove the dash for the apostrophe. Only Wordpress.

Maybe should we let the language package drive this choiche?

brianteeman:
Then you have a definite article of just the character "l" which is also wrong as the alias would be "l-acqua"

richard67:
The question is what is more wrong to a native speaker of these languages.

infograf768:
Evidently @simbus82 is totally right.
getting `leau` instead of `l-eau` just to please `brians-pull-request` is very much English centered —but we have been used to that behavior for years.

We do have a solution though, @simbus82 , and it is to use unicode aliases.
There, the single quote is still using the dash and we could even make it better by replacing the single quote by a right single quote.

We could add in `OutputFilter` in the method `public static function stringUrlUnicodeSlug($string)`

after line 162
`$str = str_replace('-', ' ', $str);`

the line
`$str = str_replace("'", "\xE2\x80\x99", $str);`

This would give from the original title
`L'acqua è pulita`

the alias
`l’acqua-è-pulita`

Issue #32890, opened 2021-03-27
[4.0] Column 'ordering' cannot be null when restoring original version


Quy:
### Steps to reproduce the issue
Create a new article.
Click `Save` button.
Edit article.
Click `Save`  button.
Edit article.
Click `Versions` button.
Click on a version.
Click `Restore` button.

### Actual result
> Column 'state' cannot be null

brianteeman:
is this with workflow enabled?

Quy:
Yes. However, when workflow is disabled, there is a different message.

> Column 'ordering' cannot be null

brianteeman:
there is a release blocker for versions and workflow

PhilETaylor:
Confirmed. Can you confirm this was tested in PHP 8? 
<img width="511" alt="Screenshot 2021-03-27 at 18 20 39" src="https://user-images.githubusercontent.com/400092/112730498-cf57c900-8f29-11eb-8a80-5b15047f4c6c.png">

Quy:
PHP 7.4.9

PhilETaylor:
Please confirm your CANT restore the original version but your CAN restore subsequent versions

Quy:
I can restore versions except for the last one listed.

PhilETaylor:
The "last one listed" is "the first version created when you first create the article"... when the ordering is 0 in the db...

Quy:
I confirm your assessment.

PhilETaylor:
trying to restore the original version of a Category gives other null warnings

<img width="732" alt="Screenshot 2021-03-27 at 19 01 49" src="https://user-images.githubusercontent.com/400092/112731410-f369d900-8f2e-11eb-9af8-d6684bbf79bc.png">

PhilETaylor:
Ive debugged this but now Im in a loop and I cannot fight my way out, it seems that "ordering" on an Article being inserted to the db gets coverted from a 0 to NULL deep in the process for updating the assets tables.. in libraries/src/Table/Nested.php there is 

```php
$oldDispatcher = clone $this->getDispatcher();
		$blankDispatcher = new Dispatcher;
		$this->setDispatcher($blankDispatcher);

		$result = parent::store($updateNulls);

		// Restore previous callable dispatcher state:
		$this->setDispatcher($oldDispatcher);
```

Before the store, the ordering is 0 and after the store its NULL, which is then passed the the versioning plugin using the onTableAfterStore event...

joomdonation:
[...]

PR #32932, opened 2021-03-30
[4.0] Fix restore category/article from first version in version history

joomdonation:
Pull Request for Issue #32890.

### Summary of Changes
Load data from current version before replacing it with data from history to avoid error if there are some required keys are missing in the history data. See https://github.com/joomla/joomla-cms/issues/32890 to understand more about the issue.

### Testing Instructions
1. See https://github.com/joomla/joomla-cms/issues/32890 , confirm the issue
2. Apply patch, confirm that the issue is sorted . Please test by both restore category and article from it's first version.

### DIFF
--- a/libraries/src/Versioning/VersionableModelTrait.php
+++ b/libraries/src/Versioning/VersionableModelTrait.php
@@ -72,7 +72,7 @@ trait VersionableModelTrait
 
 		/**
 		 * Load data from current version before replacing it with data from history to avoid error
-		 * if there are some required keys are missing in the history data
+		 * if there are some required keys missing in the history data
 		 */
 
 		if (isset($rowArray[$key]))
--- a/libraries/src/Versioning/VersionableModelTrait.php
+++ b/libraries/src/Versioning/VersionableModelTrait.php
@@ -53,11 +53,11 @@ trait VersionableModelTrait
 
 		$rowArray = ArrayHelper::fromObject(json_decode($historyTable->version_data));
 
+		$key = $table->getKeyName();
+
 		if (implode('.', $typeAlias) != $this->typeAlias)
 		{
 			$this->setError(Text::_('JLIB_APPLICATION_ERROR_HISTORY_ID_MISMATCH'));
-
-			$key = $table->getKeyName();
 
 			if (isset($rowArray[$key]))
 			{
@@ -69,6 +69,16 @@ trait VersionableModelTrait
 
 		$this->setState('save_date', $historyTable->save_date);
 		$this->setState('version_note', $historyTable->version_note);
+
+		/**
+		 * Load data from current version before replacing it with data from history to avoid error
+		 * if there are some required keys are missing in the history data
+		 */
+
+		if (isset($rowArray[$key]))
+		{
+			$table->load($rowArray[$key]);
+		}
 
 		return $table->bind($rowArray);
 	}
--- a/libraries/src/Table/Category.php
+++ b/libraries/src/Table/Category.php
@@ -202,8 +202,7 @@ class Category extends Nested implements VersionableTableInterface, TaggableTabl
 		{
 			$this->version = 1;
 		}
-
-
+		
 		return true;
 	}
 
--- a/libraries/src/Table/Category.php
+++ b/libraries/src/Table/Category.php
@@ -187,6 +187,23 @@ class Category extends Nested implements VersionableTableInterface, TaggableTabl
 			$this->alias = Factory::getDate()->format('Y-m-d-H-i-s');
 		}
 
+		// Set default value for the following fields so that it could be restored from version history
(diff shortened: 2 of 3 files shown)


ceford:
For me this works fine for Articles but for Categories I do not get a list of Versions to choose from. Is there some setting I have missed?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32932">issues.joomla.org/tracker/joomla-cms/32932</a>.</sub>

joomdonation:
@ceford I think you will have to edit the category, then change something (so the data changed) to have system save it as new version. Could you please try that?

PhilETaylor:
Im not sure this is the correct fix, as these values do exist at some point, but if it works it works :) 🚀

ceford:
I tried changing a Category option as well as adding lines of text to the description. And I tried on a separate installation I use for other testing. In neither case do I get a list of Versions. I looked in the #__history table - there are entries for the article content but no entries for the category. So there must be some way to enable history versions in categories that I have missed,<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32932">issues.joomla.org/tracker/joomla-cms/32932</a>.</sub>

joomdonation:
> I'm not sure this is the correct fix, as these values do exist at some point, but if it works it works :) 🚀

From what I see, at least for article, when you save a new article, data for ordering is not available and it is set to 0 (0 is default value for ordering field)

I believe the same thing happens for category on first save (first version). Some data is not available on saving process (at least when the system tries to save version history), so it makes sense to set default value like in the propose code.

joomdonation:
@ceford  Are you creating category for article? If not, please do so. Also, do you see Versions button in the toolbar while editing that category?

I use default Joomla 4 installation, and version history is enabled by default, so I don't know why it is not available for your test site.

ceford:
Content -> Categories -> + then follow instructions. No Toolbar Versions label before first save. Same occurs for Contact -> Categories -> New. This is a Multi-Lingual test site! I am using Eclipse with Git to Pull and Merge the latest changes. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32932">issues.joomla.org/tracker/joomla-cms/32932</a>.</sub>

Quy:
> Column 'created_by_alias' cannot be null

Installed `Testing Sample Data`.
Edit `Popular Tags` article.
Click `Versions` button.
Click the only entry.
Click `Restore` button.

> Column 'cid' cannot be null

Edit a banner.

joomdonation:
@Quy Could you please test it again? This should work and not depend on workflow enabled/disabled.

alikon:
I have tested this item :white_check_mark: successfully on 7fce0527625c8b892c46fc8caf954affff4e7dfe<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32932">issues.joomla.org/tracker/joomla-cms/32932</a>.</sub>

libraries/src/Versioning/VersionableModelTrait.php
```diff
@@ -70,6 +70,16 @@ public function loadHistory($versionId, Table &$table)
 		$this->setState('save_date', $historyTable->save_date);
 		$this->setState('version_note', $historyTable->version_note);
 
+		/**
+		 * Load data from current version before replacing it with data from history to avoid error
+		 * if there are some required keys are missing in the history data
```
Quy:
```suggestion
		 * if there are some required keys missing in the history data
```

richard67:
I've restored the previous test result in the issue tracker since the change which has invalidated the test was just a change in a comment.

Quy:
I have tested this item :white_check_mark: successfully on bfc50f5b2f89b06b7de67c5be73ffcfb9029fd71<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32932">issues.joomla.org/tracker/joomla-cms/32932</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32932">issues.joomla.org/tracker/joomla-cms/32932</a>.</sub>

PhilETaylor:
This solution looks much better than the initial one! Great work!

richard67:
Thanks!

PR #32953, opened 2021-04-01
updated README.txt

Piyussshh:
Pull Request for Issue # .

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -25,7 +25,7 @@
 	* Before installing, read the beginners guide: https://docs.joomla.org/Special:MyLanguage/Portal:Beginners
 
 6- What are the benefits of Joomla?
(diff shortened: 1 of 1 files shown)


softforge:
I have tested nothing and nothing worked.
I do feel this is plagiarism though
https://regularlabs.com/whatnothing

Perhaps you could make it do less so its not the same
Just a suggestion

richard67:
@Piyussshh Sorry to say, but this PR is wrong. As the exclamation mark belongs to the trademark "Joomla!", there is no start of a new sentence and, and so the existing text is right. With your change it would become wrong.

If it was an [April Fools' Day joke](https://en.wikipedia.org/wiki/April_Fools%27_Day), then forgive me my lack of humour ;-)

PR #32955, opened 2021-04-01
[4.0] Update vue in media manager to version 3

laoneo:
### Summary of Changes
Update vue to version 3 in media manager. Size of mediamanager JS file is now over 20% smaller.

### Testing Instructions
Open media manager in back end and article image selection.

Media manager works as expected.

Media manager works as expected.

Not sure if it is documented that we are using vue 2. If this is the case it needs to be updated.


wilsonge:
Tests are passing here - and this is the one component with good system tests - so I'm going to merge to get this in and have wider testing.

wilsonge:
Thanks!

PR #31524, opened 2020-11-29
[4.0][CLI] Add a task to discover extensions

astridx:
### Summary of Changes

i added a CLI task to discover extensions. I took https://github.com/joomla/joomla-cms/pull/28666 as an example

### Testing Instructions
1. Apply this PR
2. Copy or symlink an extension to your Joomla installation
3. Run ~~` php ./cli/joomla.php extension:discover --eid=122`~~ ` php ./cli/joomla.php extension:discoverinstall --eid=122` for discovering the extension with the ID 122 
or 
Run the command without a parameter (like this ~~`php ./cli/joomla.php extension:discover`~~ `php ./cli/joomla.php extension:discoverinstall`) for discovering all extensions, that are copied to Joomla but not installed.

4. Run `php ./cli/joomla.php list` or only `php ./cli/joomla.php` and check the entry in the list.
5. Run ~~`php ./cli/joomla.php help extension:discover`~~ `php ./cli/joomla.php help extension:discoverinstall`to check the help information.

### Documentation Changes Required
I would add it here: https://docs.joomla.org/J4.x:CLI_Update

Issue #32967, opened 2021-04-02
[4][cli] conflicting messages with extension:discoverinstall


PhilETaylor:
### Steps to reproduce the issue

php cli/joomla.php extension:discoverinstall

### Expected result

a single message saying that there were no extensions discovered.

### Actual result

Two messages, one is a yellow warning and one is an OK.

<img width="870" alt="Screenshot 2021-04-02 at 16 37 17" src="https://user-images.githubusercontent.com/400092/113430230-b3dd3a00-93d1-11eb-9dd8-5c8710cde533.png">

### System information (as much as possible)

`4.0-dev`

PhilETaylor:
Unlike the web interface this command doesn't "discover" and return a list - it just blindly installs anything it finds which is bad as the expectation is that it would be like the web interface. 

Possibly consider breaking this up into two commands, one for discovering and one for installing, maybe even with a `--installall` param on the discover command to blindly install all it finds.

Thus thoughts

brianteeman:
> Unlike the web interface this command doesn't "discover" and return a list - it just blindly installs anything it finds which is bad as the expectation is that it would be like the web interface.

I thought we had this discussion before?

PhilETaylor:
nothing found https://github.com/joomla/joomla-cms/issues?q=is%3Aissue+CLI+author%3APhilETaylor+is%3Aclosed but I would not be surprised.

brianteeman:
https://github.com/joomla/joomla-cms/pull/31524#issuecomment-735441632

PhilETaylor:
> Does this find the extensions or does it install them?

Well that's the point. It does both, blindly installing anything it finds. My point is that is not really what we (Joomla) should be looking to do and that the command should be broken up or improved,

astridx:
If you want to install an extension specifically, then you can give the id. Then only this one will be installed.

brianteeman:
but how do you find out which ones are available?

PhilETaylor:
Its still a totally different expectation to the admin console. 

There should be a 

"Discover" = "List of discovered extensions"
"install All" = "installs all discovered extensions"
"install a single extension" = "installs a single discovered extension"

At the moment the default is just to discover and install all. I say that's the wrong expectation. Just running the command should give a LIST of discovered extensions, else where am I to get the ID from to install... 

But all this is all mute anyway, because it doesn't EVEN WORK - it doesn't discover anything at the CLI !!! 

[...]

PR #32969, opened 2021-04-02
[4.x] Inject installer via service provider

astridx:
Pull Request for Issue # https://github.com/joomla/joomla-cms/pull/31524#discussion_r532580652.

### Summary of Changes

Inject the installer the same way as the database.

@laoneo Was that what you meant in your comment, and @wilsonge  is that how it's generally intended ? Then I would standardize it in all places for the CLI.

### Testing Instructions

Code review

### Additional Questions

I realized while working on this PR, that the container is not in the Joomla Repo but in the Framework. Should not the framework be unified with the Joomla CMS in version 4? The [roadmap](https://developer.joomla.org/roadmap.html#4) includes this text:
The power of the Joomla Framework merged into the CMS

### DIFF
--- a/libraries/src/Console/ExtensionDiscoverInstallCommand.php
+++ b/libraries/src/Console/ExtensionDiscoverInstallCommand.php
@@ -53,9 +53,17 @@ class ExtensionDiscoverInstallCommand extends AbstractCommand
 	 * Database connector
 	 *
 	 * @var    DatabaseInterface
-	 * @since  4.0.0
+	 * @since  __DEPLOY_VERSION__
 	 */
 	private $db;
+
+	/**
+	 * Installer
+	 *
+	 * @var    	Installer Installer object
+	 * @since  __DEPLOY_VERSION__
+	 */
+	private $installer = null;
 
 	/**
 	 * Exit Code For Discover Failure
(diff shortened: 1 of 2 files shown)


PhilETaylor:
Interesting, I had moments ago raised this one about this cli... https://github.com/joomla/joomla-cms/issues/32967

laoneo:
I would even go one step further and make the installer a service provider. So you can load in the Console service provider the installer from the installer service provider.

wilsonge:
> I would even go one step further and make the installer a service provider. So you can load in the Console service provider the installer from the installer service provider.

I mean I think for Installer that might be a step to far as there's no way to reset all the extension data stored in all it's private/protected properties when installing an extension... (i mean generically speaking I agree)

To me this is in about the right place.

laoneo:
Why did you close? @wilsonge and me were just discussing about this line https://github.com/joomla/joomla-cms/pull/32969/files#diff-7229d3e3025f6259363b5b58ccfc54d9408edd528e0683f18b1fe2f072c24675R171

astridx:
What exactly are you discussing regarding this line?

laoneo:
That this can be also a service provider which offers an installer so you don't have to create an instance of the installer by yourself. The rest is absolutely fine and should come into 4.0.

astridx:
> I mean I think for Installer that might be a step to far as there's no way to reset all the extension data stored in all it's private/protected properties when installing an extension... (i mean generically speaking I agree)
> 
> To me this is in about the right place.

I understand @wilsonge so, that he seems that this might be a step to far and it is fine like it is currently?

laoneo:
In this comment https://github.com/joomla/joomla-cms/pull/32969#issuecomment-814507936 he quoted my comment. @wilsonge can you shed some light into this one?

Issue #32652, opened 2021-03-12
[4.0] LDAP and Composer


brianteeman:
[LDAP support in PHP is not enabled by default. ](https://www.php.net/manual/en/ldap.installation.php)

### Expected result
Composer install runs successfully

### Actual result
Composer fails with 

>   Problem 1
>     - Installation request for joomla/ldap 2.0.0-beta -> satisfiable by joomla/ldap[2.0.0-beta].
>     - joomla/ldap 2.0.0-beta requires ext-ldap * -> the requested PHP extension ldap is missing from your system.
>   Problem 2
>     - Installation request for symfony/ldap v5.2.1 -> satisfiable by symfony/ldap[v5.2.1].
>     - symfony/ldap v5.2.1 requires ext-ldap * -> the requested PHP extension ldap is missing from your system.
> 
> 

### Additional comments
If the host doesn't have LDAP support enabled on their server then a composer install is not possible

alikon:
`composer install --ignore-platform-reqs`
didn't help ?

brianteeman:
That worked - thanks. I did not know about that option

alikon:
https://docs.joomla.org/Special:MyLanguage/J4.x:Setting_Up_Your_Local_Environment

brianteeman:
although it installed with that option (yes I know I should have RTFM) i was getting 500 errors everywhere. 
as soon as i remove those packaged from composer no 500 errors

PR #32970, opened 2021-04-02
[4] add ext-ldap requirement to composer

PhilETaylor:
### Summary of Changes

As `composer.json` is going to install `joomla/ldap` and `symfonty/ldap` it is a requirement of the server to also have the LDAP PHP extension installed and this should be defined as a requirement in the composer.json

Also phpStorm moans if its not defined in the `composer.json` when viewing the LDAP Packages. 

### Testing Instructions

Code review. 

phpStorm moans:

Screenshot 2021-04-02 at 18 04 51

Composer doesn't moan enough about the missing requirements (like it correctly moans about GD)

Screenshot 2021-04-02 at 18 05 08

No Moaning. 

Screenshot 2021-04-02 at 18 06 16

None

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,6 +81,7 @@
         "ext-simplexml": "*",
         "psr/log": "~1.0",
         "ext-gd": "*",
+        "ext-ldap": "*",
         "web-auth/webauthn-lib": "2.1.*"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


brianteeman:
This is not correct. See https://docs.joomla.org/J4.x:Setting_Up_Your_Local_Environment#Possible_Issues

PhilETaylor:
The docs are wrong.

This is literally the point of constraints in composer.json

PhilETaylor:
Drupal example https://www.drupal.org/project/drupal/issues/2713073#comment-12459577

brianteeman:
agree to disagree then

brianteeman:
I am just pointing to the docs that I was pointed to when I reported #32652

PR #32974, opened 2021-04-02
[4] initial concept of table options

PhilETaylor:
### Summary of Changes

**THIS. IS. ALMOST. INCOMPLETE.** 

This is my worse work to date, Im totally at the limit of JS here.. but "it works on my machine", it "looks ugly and needs css"

Happy to merge any PRs made to [the branch](https://github.com/PhilETaylor/joomla-cms/tree/tableoptions) - this needs some help to get it over the line! 

A huge thanks to @dgrammatiko so far. He gave me clean code and I butchered it. 

### Testing Instructions

I have only applied the change to the Users, Articles and Categories pages. 

No way to hide columns 

A new Table options list at the foot of a table, that allows you to deselect columns to HIDE and to select columns to SHOW

Your preferences are STORED in your browser localstorage for that site. When you reload the page or return to the same page, your preferences are reapplied.

### LIMITATIONS & NOTES:
The first 2 columns can never be hidden, these are normally the checkbox and the dragger (or the checkbox and the username on the users list) 
Occasionally Ive seen a CSS issue on the new checkbox list layout, but Ive applied no css at all so far. 
ONLY tested in Safari and Firefox but SHOULD work in all browsers
ONLY shows up in 1024px wide screens and above by design. 
The bottom of the table was chosen for the placement of the Table Options as the page has been "designed" previously and adding them at the top just gets in the way. 
Probably have to target 4.1-dev as this is a new feature.

### Future improvement ideas already suggested:

 - Store the preferences against the user object in the db for browser/computer independentcy
 - Initially hide most things to "clean up" the user interface and allow users to add more clutter
 - Overcome the bootstrap limitation of classes for screens less tan 1024px wide and remove all responsive classes that are currently hiding things, and allow all columns on mobile devices to be toggled with table options.

### Preview:

Screen Recording 2021-04-02 at 09 35 53 pm

Yes.

### DIFF
--- a/build/media_source/system/js/tableoptions.es6.js
+++ b/build/media_source/system/js/tableoptions.es6.js
@@ -77,12 +77,7 @@ if (window.innerWidth > 1024) {
 
   const ul = document.createElement('ul');
   headers.forEach((el, index) => {
-    el.classList.remove('d-none', 'd-md-table-cell', 'd-lg-table-cell');
-
-    // @todo better restrictions on the columns that can never be hidden
-    if (index === 0 /* checkbox */ || index === 1 /* ordering */) {
-      return;
-    }
+    el.classList.remove('d-none', 'd-md-table-cell', 'd-lg-table-cell', 'd-xl-table-cell');
 
     const li = document.createElement('li');
     const label = document.createElement('label');
@@ -111,16 +106,14 @@ if (window.innerWidth > 1024) {
     [].slice.call(col.children)
       .forEach((cc, index) => {
         if (cc.nodeName !== 'TH') {
-          cc.classList.remove('d-none', 'd-md-table-cell', 'd-lg-table-cell');
+          cc.classList.remove('d-none', 'd-md-table-cell', 'd-lg-table-cell', 'd-xl-table-cell');
           if (storage.get(index) === true) {
             toggleHidden(index);
           }
         } else {
           // remove the checkbox for this column as its the "main link" of an item.
           const lis = [].slice.call(document.querySelector('details ul').children);
-          let i = index;
-          i -= 2; // because we skipped the first two columns for now
-          lis[i].remove();
+          lis[index].remove();
         }
       });
   });
--- a/build/media_source/system/js/tableoptions.es6.js
+++ b/build/media_source/system/js/tableoptions.es6.js
@@ -33,7 +33,7 @@ const storage = {
     const obj = this.getData();
 
     if (obj) {
-      return obj[i] === 1;
+      return Object.prototype.hasOwnProperty.call(obj, i) && obj[i] === 1;
     }
 
     return false;
--- a/build/media_source/system/js/tableoptions.es6.js
+++ b/build/media_source/system/js/tableoptions.es6.js
@@ -1,0 +1,113 @@
+if (window.innerWidth > 1024) {
+
+  let storage = {
+    pagens() {
+      return 'joomla-tableoptions-' + document.querySelector('.page-title')
+        .innerText
+        .trim()
+        .replace(/\s\'\//, '-')
+        .toLowerCase();
+    },
+    getData() {
+      let item = window.localStorage.getItem(this.pagens());
+
+      if (null == item) {
+        return {};
+      }
+      try {
+        return JSON.parse(item);
+      } catch (e) {
+        return {};
+      }
+    },
+    has() {
+      return window.localStorage.hasOwnProperty(this.pagens());
+    },
+    set(i, v) {
+      let obj = this.getData();
+
+      if (obj) {
+        obj[i] = v;
+        window.localStorage.setItem(this.pagens(), JSON.stringify(obj));
+      }
+    },
+    get(i) {
+      let obj = this.getData();
+
+      if (obj) {
+        return obj[i] == 1;
+      }
+
+      return false;
+    }
+  };
+
+  const table = document.querySelector('table');
+  const headers = [].slice.call(table.querySelector('thead tr').children);
+  const columns = [].slice.call(table.querySelectorAll('tbody tr'));
+
+  function toggleHidden(index) {
+
+    if (headers[index].classList.contains('d-md-table-cell')) {
+      headers[index].classList.remove('d-md-table-cell');
+    }
+
+    headers[index].classList.toggle('d-none');
+    columns.forEach((col) => {
+      if (col.children[index].classList.contains('d-md-table-cell')) {
+        col.children[index].classList.remove('d-md-table-cell');
+      }
+      col.children[index].classList.toggle('d-none');
+    });
+
+    if (headers[index].classList.contains('d-none')) {
+      storage.set(index, 1);
+    } else {
+      storage.set(index, 0);
+    }
+  }
+
+  headers.forEach((el, index) => {
+    el.classList.remove('d-none', 'd-md-table-cell', 'd-lg-table-cell');
+  });
+  columns.forEach(col => {
+    [].slice.call(col.children)
+      .forEach((cc, index) => {
+        cc.classList.remove('d-none', 'd-md-table-cell', 'd-lg-table-cell');
+        if (storage.get(index) == true) {
+          toggleHidden(index);
+        }
+      });
+  });
+
+  const detailElement = document.createElement('details');
+  const summary = document.createElement('summary');
+  summary.innerText = 'Table options';
+  detailElement.appendChild(summary);
+
+  const ul = document.createElement('ul');
+  const lis = headers.forEach((el, index) => {
+    if (index === 0/*checkbox*/ || index === 1/*ordering*/ || index === headers.length - 2) return;
+    const li = document.createElement('li');
+    const label = document.createElement('label');
+    const input = document.createElement('input');
+    input.type = 'checkbox';
+
+    if (storage.get(index) == 0) {
+      input.setAttribute('checked', '');
+    }
+
+    input.addEventListener('input', () => toggleHidden(index));
+    if (el.querySelector('span')) {
+      label.innerText = el.querySelector('span').innerText;
+    } else {
+      label.innerText = el.innerText;
+    }
+    label.appendChild(input);
+    li.appendChild(label);
+    ul.appendChild(li);
+  });
+  detailElement.appendChild(ul);
+
+  table.insertAdjacentElement('afterend', detailElement);
+}
--- a/build/media_source/system/js/tableoptions.es6.js
+++ b/build/media_source/system/js/tableoptions.es6.js
@@ -46,6 +46,7 @@ const columns = [].slice.call(table.querySelectorAll('tbody tr'));
 
 function toggleHidden(index) {
   headers[index].classList.toggle('d-none');
+  
   columns.forEach((col) => {
     col.children[index].classList.toggle('d-none');
   });
--- a/build/media_source/system/js/tableoptions.es6.js
+++ b/build/media_source/system/js/tableoptions.es6.js
@@ -93,7 +93,7 @@ if (window.innerWidth > 1024) {
     const input = document.createElement('input');
     input.type = 'checkbox';
 
-    if (storage.get(index) === 0) {
+    if (storage.get(index) === false) {
       input.setAttribute('checked', '');
     }
 
--- a/build/media_source/system/js/tableoptions.es6.js
+++ b/build/media_source/system/js/tableoptions.es6.js
@@ -45,15 +45,8 @@ const headers = [].slice.call(table.querySelector('thead tr').children);
 const columns = [].slice.call(table.querySelectorAll('tbody tr'));
 
 function toggleHidden(index) {
-  if (headers[index].classList.contains('d-md-table-cell')) {
-    headers[index].classList.remove('d-md-table-cell');
-  }
-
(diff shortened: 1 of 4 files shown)


brianteeman:
Nice work. For me the main problem with this approach is that its per computer, per browser, per site,
But maybe thats just me

brianteeman:
> ONLY shows up in 1024px wide screens and above by design.
Why add this limitation?

PhilETaylor:
Well the other option was to store your preferences in the database against your user object... not hard to do, but... I have all weekend ...

dgrammatiko:
> For me the main problem with this approach is that its per computer, per browser, per site,

We could add a PHP ajax endpoint to store the state on the user's account options, that's easily doable

PhilETaylor:
> ONLY shows up in 1024px wide screens and above by design.
> Why add this limitation?

Because when you view your site on a mobile, the responsive design of the admin template kicks in and most columns are already hidden. the 1024 was the number @dgrammatiko pulled out of the hat.

brianteeman:
> The first 2 columns can never be hidden,

If you hide the cell that contains the scope then you break the tables accessibility. This is usually (always?) the cell that is the link.
I guess it would be daft for a user to hide that but we all know users can be daft so perhaps prevent this column from being hidden as well.

brianteeman:
> > ONLY shows up in 1024px wide screens and above by design.
> > Why add this limitation?
> 
> Because when you view your site on a mobile, the responsive design of the admin template kicks in and most columns are already hidden. the 1024 was the number @dgrammatiko pulled out of the hat.

But what if I wanted to decide the columns for myself. That wasn't possible before but it would be now

PhilETaylor:
> But what if I wanted to decide the columns for myself. That wasn't possible before but it would be now

Then someone would have to go through each screen removing the responsive classes that hide things

dgrammatiko:
> Why add this limitation?

The limitation comes from Bootstrap's utilities and the forced `!important` everywhere. Removing this means we need an array with all the possible classes that need to be removed... It needs a bit of thinking here

PhilETaylor:
> If you hide the cell that contains the scope then you break the tables accessibility. This is usually (always?) the cell that is the link.

Not always easy to know WHICH link in WHICH column... as there are lots of links in the row... links to toggle state, featured etc...

Its only breaking accessibility if you choose to hide that column and no one would right... ? well never say never :)

dgrammatiko:
> The first 2 columns can never be hidden

@PhilETaylor I would keep also the ID always visible (eg `arr.length - 1`) but that's just personal opinion here

PhilETaylor:
>  but that's just personal opinion here

Which is why its configurable now  - you are now free to keep it and Im free to hide it ;-)

brianteeman:
> Not always easy to know WHICH link in WHICH column... as there are lots of links in the row... links to toggle state, featured etc...

actually it can be determined programmatically because it is a th not a td and has a scope="row"

brianteeman:
@dgrammatiko @PhilETaylor ok I understand the 1024 part now

PhilETaylor:
> actually it can be determined programmatically because it is a th not a td and has a scope="row"

AWESOME.

build/media_source/system/js/tableoptions.es6.js
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  const detailElement = document.createElement('details');
+  const summary = document.createElement('summary');
+  summary.innerText = 'Table options';
+  detailElement.appendChild(summary);
+
+  const ul = document.createElement('ul');
+  const lis = headers.forEach((el, index) => {
+    if (index === 0/*checkbox*/ || index === 1/*ordering*/ || index === headers.length - 2) return;
```
dgrammatiko: This needs a different logic to ensure that these 2 columns are actually the checkbox and the ordering, will post something (probably tomorrow)

PhilETaylor: When I say ordering I actually I mean the dragging column.

The users page has no ordering (draggable)

The ordering columns were also putting some "order by" kind of text from their innerText into the label... quick fix was to exclude them haha

So yeah a few edge cases to play with.

build/media_source/system/js/tableoptions.es6.js
```diff
@@ -0,0 +1,113 @@
...
+    if (obj) {
+      obj[i] = v;
+      window.localStorage.setItem(this.pagens(), JSON.stringify(obj));
+    }
+  },
+  get(i) {
+    const obj = this.getData();
+
+    if (obj) {
+      return obj[i] === 1;
```
dgrammatiko:
```suggestion
      return obj.hasOwnProperty(i);
```

PhilETaylor:
Another bug with this is on the categories 

Screenshot 2021-04-03 at 11 53 39

build/media_source/system/js/tableoptions.es6.js
```diff
@@ -0,0 +1,113 @@
...
+const columns = [].slice.call(table.querySelectorAll('tbody tr'));
+
+function toggleHidden(index) {
+  if (headers[index].classList.contains('d-md-table-cell')) {
+    headers[index].classList.remove('d-md-table-cell');
+  }
+
+  headers[index].classList.toggle('d-none');
+  columns.forEach((col) => {
+    if (col.children[index].classList.contains('d-md-table-cell')) {
```
dgrammatiko, line 54:
```suggestion

```

build/media_source/system/js/tableoptions.es6.js
```diff
@@ -0,0 +1,113 @@
...
+      col.children[index].classList.remove('d-md-table-cell');
```
dgrammatiko, line 55:
```suggestion

```

build/media_source/system/js/tableoptions.es6.js
```diff
@@ -0,0 +1,113 @@
...
+    }
```
dgrammatiko, line 56:
```suggestion

```

build/media_source/system/js/tableoptions.es6.js
PhilETaylor: calling `hasOwnProperty` on an object is against the javascript lint rules, as I just found out. Also we need this to return TRUE only if the property is a 1

dgrammatiko: This is ok for eslint:
```js
Object.prototype.isPrototypeOf.call(obj, i);
```

PhilETaylor: ` return Object.prototype.hasOwnProperty.call(obj, i) && obj[i] === 1;`

This works.

build/media_source/system/js/tableoptions.es6.js
```diff
@@ -0,0 +1,113 @@
...
+
+    return false;
+  },
+};
+
+const table = document.querySelector('table');
+const headers = [].slice.call(table.querySelector('thead tr').children);
+const columns = [].slice.call(table.querySelectorAll('tbody tr'));
+
+function toggleHidden(index) {
```
dgrammatiko: This function is doing some checks that are not necessary:
```js
function toggleHidden(index) {
   headers[index].classList.toggle('d-none');

   columns.forEach((col) => {
     col.children[index].classList.toggle('d-none');
   });

   if (headers[index].classList.contains('d-none')) {
     storage.set(index, 1);
   } else {
     storage.set(index, 0);
   }
 }
```

We already removed the classes at the initialization part

dgrammatiko:
@PhilETaylor can you diff your script with https://gist.github.com/dgrammatiko/4d8121c2cda4096224d4de283a1a433a you need a couple of conditonals for sanity checks

PhilETaylor:
Sure, multi tasking with real life parenting over the weekend so will be delayed sorry.

PhilETaylor:
Todays improvements 
 
 - Now works well on Users, Categories and Articles pages
 - Now correctly handles the non-text headers on the categories page, by using the text in the `visually-hidden` span as text for the checkboxes
 - Now doesn't allow the hiding of the "main thing" 
 - Now allows the id/checkbox/ordering columns to be hidden (useful for screenshots and the like, give the user the power to make their own decisions) 
 - Some sense checks when the page loads to ensure the table contains enough of what we need to provide the options

build/media_source/system/js/tableoptions.es6.js
PhilETaylor: Ive removed the exception for these columns totally. It makes sense to allow the user full control over their own destiny. 

Useful for copy and paste, and screenshots to hide those columns.

ceford:
I must be doing something wring. I applied the patch with Patchtester, ran npm ci, but did not see the dropdown list of columns to select for hiding. I did try the demo by @dgrammatiko. 

Quite separately, I put the table in a div styled overflow-x: scroll; max-width: 1000px; right now I don't know how to get the width of the parent div. I quite like the left-right scroll that leaves the left menu, toolbar and search bar alone while the table scrolls. Please try it. It could be combined with any other column manipulation.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32974">issues.joomla.org/tracker/joomla-cms/32974</a>.</sub>

TLWebdesign:
> I must be doing something wring. I applied the patch with Patchtester, ran npm ci, but did not see the dropdown list of columns to select for hiding. I did try the demo by @dgrammatiko.

I had the same issue. The Patchtester doesn't do the "build" of the JS file of this feature. to Still try it out you could copy the js file contents form the PR to the folder "/media/system/js" and name the file tableoptions.js

At the moment it gave me a JS error on line 124 so i commented both 124 and 125 out so i could see it in action. it looks very promising and i think a very much needed feature to have a "workable" backend.

ceford:
Here is an experiment to try:

Edit your locally installed administrator/components/com_content/articles/tmpl/default.php 
Add these two lines to wrap the table:

```
					<div class="myscroller">
					... table here
					</div>
```
And then this after line 95:
```
$wa->addInlineStyle("
.myscroller {
	max-width: calc(100vw - 20rem);
	overflow-y: scroll;
}
.myscroller.wider {
	max-width: calc(100vw - 5rem);
}
");

$wa->addInlineScript("
	const vw = Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0);
	const test = document.getElementsByClassName('myscroller');

	var ro = new ResizeObserver(entries => {
		for (let entry of entries) {
			const cr = entry.contentRect;
			if (cr.width < 100) {
				test[0].classList.add('wider'); // = vw - cr.width + \"px\";
			} else {
				test[0].classList.remove('wider'); // = vw - cr.width + \"px\";
			}
		}
	});

	ro.observe(document.getElementById('sidebar-wrapper'));
", [], ['type' => 'module']);
```
Click on the table to scroll left-right. Make the screen narrower - try simulating an iPad in landscape and portrait mode. And toggle the side menu.

PhilETaylor:
@ceford this PR relates to a Javascript feature to hide/show columns. If you would like to propose a competing/complimenting solution with horizontal scrolling of tables then please make a new PR. 

Horizontal scrolling on MOBILE devices is a nice feature when correctly implemented. However, as I have shown time and time again, the root issue Im trying to resolve is NOT ON MOBILE DEVICES, but on my HUGE 27" iMac screen, when a Joomla page on half the screen, still has the items squashed into a stupid amount of column space - my screenshot examples of this are here https://github.com/joomla/joomla-cms/discussions/31915

ceford:
```@ceford this PR relates to a Javascript feature to hide/show columns. If you would like to propose a competing/complimenting solution with horizontal scrolling of tables then please make a new PR.```

Sorry, getting this issue mixed up with #31915. Just thinking of the general problem with wide tables.

PhilETaylor:
The problem is not that the table is "wide" the problem is that even in a wide table, the columns are very very narrow as Joomla is trying to put so much into its table grids - out of which no body needs, until they need to sort on that column. 

Having the main title of an article squashed so that you can show the votes (which is disabled by default) and hitscolumn is bad UI IMHO. 

This is out of the box:

112766940-09010080-900c-11eb-8a6e-eb5c5dcb1282

dgrammatiko:
@PhilETaylor may I ask why you've killed this one?

PhilETaylor:
Only so much negativity one person can cope with. I'll be completing my work on Empty States, and now that my work on the FTP Layer is not needed, I'll be spending May trying to cream as much money from consultancy as possible instead of contributing for free... 

Im not sure, in its current state, that this had much chance of being merged anyway.

PhilETaylor:
I think the only bits still needed on this are (from memory) for the next person picking this up:

 - CSS 
 - Posting the preferences back via AJAX to store them in the Users State (or Users Params to persist across sessions), storing against the url/view/component used on. 
 - Reflecting the current preferences when a page loads to hide/show the right columns again
 - Maintainers buy in - no Maintainer has actually agreed this is a feature they want to add
 - As its a new feature, it can no longer be added to Joomla 4.0 and needs to be based on 4.1 (aka some years away probably) 
 - Could/Should be implemented as a Joomla Plugin, so can be enabled/disabled at will. 
 - Initially hide most things to "clean up" the user interface and allow users to add more clutter
 - Disabling completely (or testing perfectly) on mobile devices.

PR #32983, opened 2021-04-03
[4.0] Improve CSP, second attempt

Fedik:
### Summary of Changes

It seems most confusing part of CSP is "mode" selection, especially when it set to "auto".
So I have dropped mode selection, and "auto-header" generation  (we can review it in future).
And turned the CSP component to User configurable.

Also Reporting configuration was changed. Now it possible to enable reporting and "report-only" in separated options.
This allows to collect Reports for live site with enabled "Content-Security-Policy" header.

**New CSP workflow:**
- Enable CSP component, this also will enable reporting.
- Set up CSP rules (in component config).
- Navigate through the site, visit different pages.
- Go to CSP reports page in the backed, and check if here any report.
- Adjust CSP rules  (in component config), or fix blocked content, and clean up the report list.
- (optionally) turn of the "Report"

### Testing Instructions

**Before you begin: make sure your test site domain not `localhost` or IP**

Apply patch.
Enable CSP in component configuration.
Make sure "Client" is Site
Make sure "Report" is Enabled.
Make sure "Report-Only" is Off.

**step 0**
Visit to the site, make sure it is working, nothing broken.
And the CSP report is empty.

**step 1**

Add 2 rules in CSP component configuration.
```
default-src: 'self'
img-src: 'self'
```

**step 2**
Create a custom HTML module, with background, with some text and with an external IMG:
```
<img src="https://picsum.photos/300/300" width="300" height="300" />
```

Check the module on the site.
The module will not show images.

**step 3**

Got to the CSP report page, in backend.
`administrator/index.php?option=com_csp&view=reports`

You should see at least 2 blocked elements: `'unsafe-inline'` and `https://picsum.photos/300/300`

**step 4**

Update CSP rules in the component configuration, to next:
```
default-src: 'self' 'unsafe-inline'
img-src: 'self' picsum.photos *.picsum.photos
```

Visit to the site and check that the module is working.
Clean up the report and make sure there no more blocked elements: `'unsafe-inline'` and `https://picsum.photos/300/300` after further Site navigation.

Yes, the CSP documentation page need to be updated, to reflect the changes.
Additionally need to warn User that the editor may add inline style so he/she may need to add `'unsafe-inline'`. 
And that CSP and  `localhost` not a best friends.

@zero-24  please review


brianteeman:
> be aware: 'unsafe-inline' only for testing here and should be not used on live site

until core is completely reviewed it will have to be used

Fedik:
> until core is completely reviewed it will have to be used

For the site the only issue for it is the mod_custom background,
and it is fixed in separated PR.

But for the Administrator, yes, that true.

UPD:
Default CSP config pointed to "Site".
Probably this should be noted in documentation.

brianteeman:
> For the site the only issue for it is the mod_custom background,
and it is fixed in separated PR.

If only that was true. Navigate the site (with sample data) and watch the errors in the browser console

Fedik:
Ah yea got it.
Right, it can come from the editor, well then it should be really noted in documentation.

upd: I have removed that warning from description.

But note: the default rules is empty (after this PR), so it should not break anything, by just enabling CSP.
We basically gives to User a tool, he/she free to decide what he/she will allow.

brianteeman:
> Right, it can come from the editor, 

and core outputs such as tags

brianteeman:
> Add 2 rules in CSP component configuration.
default-src: 'self'
img-src: 'self'

image

Without doing anything else I visited the frontend

image

And the reports page has 25 items

image

Fedik:
ouch, that new, something "localhost" related,
I do not have such error because I have "virtual domain"

I need to ask google :)

Fedik:
It seems that for localhost testing you have to allow localhost domain,
the rule should be not `self`, but `self localhost` or `self http://localhost` this need to test

upd: this also something that need to add to documentation

brianteeman:
Just another reason why this entire component is not suitable for core

administrator/components/com_csp/config.xml
```diff
@@ -24,28 +24,28 @@
...
+			type="radio"
+			label="COM_CSP_CONTENTSECURITYPOLICY_REPORT"
+			description="COM_CSP_CONTENTSECURITYPOLICY_REPORT_DESC"
+			layout="joomla.form.field.radio.switcher"
+			default="1"
 			showon="contentsecuritypolicy:1"
-			>
-			<option value="custom">COM_CSP_CONTENTSECURITYPOLICY_MODE_CUSTOM</option>
-			<option value="detect">COM_CSP_CONTENTSECURITYPOLICY_MODE_DETECT</option>
-			<option value="auto">COM_CSP_CONTENTSECURITYPOLICY_MODE_AUTO</option>
```
Quy, line 36: Please delete language strings.

administrator/components/com_csp/src/View/Reports/HtmlView.php
```diff
@@ -103,27 +103,6 @@ public function display($tpl = null)
 			$this->httpHeadersId = ReporterHelper::getHttpHeadersPluginId();
 		}
 
-		if (ComponentHelper::getParams('com_csp')->get('contentsecuritypolicy_mode', 'detect') === 'detect'
-			&& ComponentHelper::getParams('com_csp')->get('contentsecuritypolicy', 0)
-			&& ReporterHelper::getCspTrashStatus())
-		{
-			$this->trashWarningMessage = Text::_('COM_CSP_COLLECTING_TRASH_WARNING');
```
Quy, line 123: Please delete `use Joomla\CMS\Component\ComponentHelper;`.

components/com_csp/src/Controller/ReportController.php
```diff
@@ -69,8 +70,9 @@ class ReportController extends BaseController
 	public function log()
 	{
 		// Make sure we we are in detect mode and csp is active
```
Quy:
```suggestion
```

Fedik:
It seems that `localhost` untrusted in CSP by default. (https://stackoverflow.com/a/50472608/1000711)
If all your previous test also was on localhost, that may explain everything ;) 

btw, I would say that "Step 1" of your test was successful,
because now you can see why and where it filed, and what to do (compared to "before patch")

If you familiar with "hosts" file ( /etc/hosts ) you can try add virtual domain :
```
127.0.0.1 localhost brianteeman.local
```
then open your site with `brianteeman.local/joomla-path/` url

Fedik:
I have updated description.

brianteeman:
>If all your previous test also was on localhost, that may explain everything ;)

Never got as far as setting up any directives :)

administrator/components/com_csp/config.xml
Fedik: done

Fedik:
note: test fails because JS code style, not related to this PR

Quy:
Add/include `contentsecuritypolicy:1` to the `showon` of `Report-Only`?

Fedik:
good point, I will update

plugins/system/httpheaders/httpheaders.php
```diff
@@ -409,142 +390,6 @@ private function setCspHeader(): void
...
-
-		$this->db->setQuery($query);
-
-		try
-		{
-			$rows = (array) $this->db->loadObjectList();
-		}
-		catch (\RuntimeException $e)
-		{
-			$this->app->enqueueMessage(Text::_('JERROR_AN_ERROR_HAS_OCCURRED'), 'error');
```
Quy, line 434: Please delete `use Joomla\CMS\Language\Text;`.

administrator/components/com_csp/tmpl/reports/default.php
```diff
@@ -80,9 +72,6 @@
 								<td class="text-center">
 									<?php echo HTMLHelper::_('grid.checkall'); ?>
 								</td>
-								<th scope="col" class="text-center">
-									<?php echo HTMLHelper::_('searchtools.sort', 'JSTATUS', 'a.published', $listDirn, $listOrder); ?>
```
Quy, line 84: Remove the following from the `Sort by` dropdown:

```
<option value="a.published ASC">JSTATUS_ASC</option>
<option value="a.published DESC">JSTATUS_DESC</option>
```

plugins/system/httpheaders/httpheaders.php
Fedik: done

administrator/components/com_csp/config.xml
```diff
@@ -24,28 +24,28 @@
...
+		>
```
Quy:
```suggestion
			>
```

Quy:
@Fedik Please fix file conflicts. Thanks.

astridx:
If tested with a clean Cassiopeia local - not on locahost but on a virtual Domain. 

I find this CSP-function very important for Joomla 4 and have also been able to implement everything I wanted. However, I also think that not everything is crystal clear for users. Perhaps you agree with me on the points I mention below. I find that everything can be explained with a hint.

But first:

1. After install, the plugin was active and the component was inactive as default.
 
The headers after install:
header_afterinstall

To make it short: The test scenarios you described in the intro comment worked for me fine.




Moreover I noticed the following.

1. I was confused by the fact that the setting of some headers are possible in two places in the plug-in. The upper one (in the picture) overwrites the lower one. It is still `x-frame-options SAMEORIGIN.` I had expected the opposite.

headermist1



2. It is in my eyes not self-explanatory which setting is taken when I activate  the component and set  `content-security-policy default-src 'self' img-src 'self'` in the plugin.

For example: I set this in the plugin

1a

and this in the component

1b

I expected, that the picture from picsum.photos is not loaded. But it was loaded. This are the headers:

1c

If I deactivate the component, the picture is hidden and the headers are from the plugin:

1d



3. There is also no hint in the component that the plugin must be activated. In my opinion this would be good. Especially in connection with point 2 . The result of point 2 led me to the thought that the plugin might not be necessary for the component. But if I activate the component and deactivate the plugin, I only see the header from the .htaccess. 




4. We need documentation, that explains, what I should enter into the field of the Plugin, if I use `Force Http Headers`. I have to look into the code, to know, what is the value for `Report-to`

zero-24:
FYI an PR to drop this feature from 4.0 has been made here: https://github.com/joomla/joomla-cms/pull/33550

Fedik:
@zero-24  I can close mine pr or?

Fedik:
okay, I see, then I close this one

zero-24:
Yes thanks for all the support and effort you put into this here.

Issue #32961, opened 2021-04-02
[4] Clearing cache doesn't flush PHP files from opcache


PhilETaylor:
### Steps to reproduce the issue

File Cache Storage class uses a lot of `unlink` instead of using the Filesystem layer and so removed PHP files from the cache are not removed from the opcache. 

https://github.com/joomla/joomla-cms/blob/661407c91b0b3230c1a354a8c7c340e79f8fa8cb/libraries/src/Cache/Storage/FileStorage.php#L525

https://github.com/joomla/joomla-cms/blob/661407c91b0b3230c1a354a8c7c340e79f8fa8cb/libraries/src/Cache/Storage/FileStorage.php#L537

https://github.com/joomla/joomla-cms/blob/661407c91b0b3230c1a354a8c7c340e79f8fa8cb/libraries/src/Cache/Storage/FileStorage.php#L20

Soooooo anyone know why the Filesystem package has such a performance issue that Joomla itself refuses to use it? Without using the Filesystem class its not invalidating Opcache

PhilETaylor:
https://github.com/joomla/joomla-cms/blob/661407c91b0b3230c1a354a8c7c340e79f8fa8cb/libraries/src/Cache/Storage/FileStorage.php#L20

Soooooo anyone know what the Filesystem package has such a performance issue that Joomla itself refuses to use it?

PhilETaylor:
PR #32993

PR #32993, opened 2021-04-04
[4] Invalidate opcache on cache file deletion without using Filesystem API for 'performance reasons'

PhilETaylor:
Pull Request for Issue #32961

### Summary of Changes

When purging your cache, we need to clear the PHP Opcache of these files you are deleting. 

### Testing Instructions

THIS PR RELIES ON THE NEW METHODS IN https://github.com/joomla/joomla-cms/pull/32915 and so you first need to apply that PR, or wait until its merged before testing this. 

Set up your PHP to aggressively use the PHP Opcache and never use the files on the disk after compiling. Instructions for that are in https://github.com/joomla/joomla-cms/pull/32915#issuecomment-812562008

Turn on Joomla Global Configuration Caching to file system. Generate a load of cache. Purge your cache in the cache manner in admin. 

Files are removed from the hard disk, but not from the PHP OPcache

Files are removed from the hard disk, and expired in the PHP OPcache

None.

### DIFF
--- a/libraries/src/Cache/Storage/FileStorage.php
+++ b/libraries/src/Cache/Storage/FileStorage.php
@@ -11,6 +11,7 @@ namespace Joomla\CMS\Cache\Storage;
 \defined('JPATH_PLATFORM') or die;
 
 use Joomla\CMS\Cache\CacheStorage;
+use Joomla\CMS\Filesystem\File;
 use Joomla\CMS\Language\Text;
 use Joomla\CMS\Log\Log;
 
@@ -68,6 +69,7 @@ class FileStorage extends CacheStorage
 				// Delete only the existing file if it is empty.
 				if (@filesize($path) === 0)
 				{
+					File::invalidateFileCache($path);
 					@unlink($path);
 				}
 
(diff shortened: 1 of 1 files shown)


libraries/src/Cache/Storage/FileStorage.php
```diff
@@ -534,7 +540,8 @@ protected function _deleteFolder($path)
 			{
 				$file = $this->_cleanPath($file);
 
-				// In case of restricted permissions we zap it one way or the other as long as the owner is either the webserver or the ftp
+				// In case of restricted permissions we delete it one way or the other as long as the owner is either the webserver or the ftp
+				File::invalidateFileCache($file);
```
Quy, line 544:
```suggestion
				File::invalidateFileCache($file);

```

PhilETaylor: Mobile GitHub won't let me merge that sorry 

image

richard67:
@PhilETaylor I was able to commit Quy's suggested CS change here on GitHub.

PhilETaylor:
Thanks - I think it was the GitHub iPhone App just being silly. I have reported it this morning to their developers.

PhilETaylor:
Merged into https://github.com/joomla/joomla-cms/pull/32915

PR #32994, opened 2021-04-04
[4] Disable auto complete on url field in extension installer

PhilETaylor:
### Summary of Changes

When installing from url, the input field allows autocomplete, and browsers then store this for use later.

This has lead to urls being stored in the browser, and `extra_query` values also being stored

As the chances of having to install the same url to the same site multiple times is very very low, there is no reason for the browser to be storing the URL values for future use, and there's no point in autofill on this field.

Also another bug, is if your autofill cache has a longer url of the same one you are trying to install (for example the same url with an appended `extra_query`) then after pasting in your URL (without  `extra_query`), the browser sometimes auto-appends with the  `extra_query` on blur, or just before blur, just when you are clicking the button, at the worst time! 

Most browsers the autofill cache is PER SITE, so you lose nothing by disabling this. 

### Testing Instructions

Apply PR, check the field doesn't present any autofill suggestions. 

Screen Recording 2021-04-04 at 01 30 56 pm

Screen Recording 2021-04-04 at 01 24 29 pm

Browser doesn't store, or present a saved listed of urls for autocompletion. 

None

### DIFF
--- a/plugins/installer/urlinstaller/tmpl/default.php
+++ b/plugins/installer/urlinstaller/tmpl/default.php
@@ -26,7 +26,7 @@ $this->app->getDocument()->getWebAssetManager()
 		<?php echo Text::_('PLG_INSTALLER_URLINSTALLER_TEXT'); ?>
 	</label>
 	<div class="controls">
-		<input type="text" id="install_url" name="install_url" class="form-control" placeholder="https://">
+		<input type="text" id="install_url" name="install_url" class="form-control" placeholder="https://" autocomplete="off">
 	</div>
 </div>
 <hr>


brianteeman:
Please dont. I use this ability everyday several times to install patchtester

PhilETaylor:
Install a text snippets expander :) then you can just map a few chars to the url and then just type a few chars like "ipttt" and have that expand to the url.  or even use Mac keyboard shortcuts 

Screenshot 2021-04-04 at 14 19 11

brianteeman:
No thanks I would rather use the native functionality of a url field - and I dont use a mac ;)

PS lol how you have a correction for Lela's name

PhilETaylor:
Because Lela is always auto corrected to Lea on iPhone (Mac syncs these to iPhone)

PhilETaylor:
I only use Macs BECAUSE OF YOU! Else I would still be on Ubuntu.

PR #33004, opened 2021-04-04
[4] s/int/mixed 

PhilETaylor:
Code review

The `min()` function in PHP returns a `mixed` and not `int`

https://www.php.net/manual/en/function.min.php

### DIFF
--- a/libraries/src/Utility/Utility.php
+++ b/libraries/src/Utility/Utility.php
@@ -54,7 +54,7 @@ class Utility
 	 *
 	 * @param   mixed  $custom  A custom upper limit, if the PHP settings are all above this then this will be used
 	 *
-	 * @return  integer  Size in number of bytes
+	 * @return  mixed  Size in number of bytes
 	 *
 	 * @since   3.7.0
 	 */


HLeithner:
Not sure about this on, I mean it's a b/c break changing this, so maybe cast to int would be better and possible?

PhilETaylor:
well that your call. I don't see how following thePHP documentation is a b.c break though.

ceford:
I have tested this item :white_check_mark: successfully on 68cedf8875ebfb6317ecb097f60db0faeb639674<br /><br />I looked at the code and noted that getMaxUploadSize is used in three places and as this is just a correction to the description of what is returned I think it is OK. If it were to be cast as an int that would be a change that might have b/c consequences. I declare myself a php amateur here so agree it is for the experts to call.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33004">issues.joomla.org/tracker/joomla-cms/33004</a>.</sub>

PhilETaylor:
> I looked at the code and noted that getMaxUploadSize is used in three places

In Joomla core. You do not know how many calls this method gets (if any) from 3rd Party code in plugins/extensions (probably none, but that's not the point.

Before this PR this method ALWAYS returned with a type of `mixed` as that is literally the PHP documented type returned from `min()` according to https://www.php.net/manual/en/function.min.php

```
min() returns the parameter value considered "lowest" according to standard comparisons. If multiple values of different types evaluate as equal (e.g. 0 and 'abc') the first provided to the function will be returned.

If an empty array is passed, then false will be returned and an E_WARNING error will be emitted.
```

Therefore, to cast the return to an `int` is (theoretically) a backward incompatible change. 

Changing the docblock to factually state that the return type from the min is `mixed` is fully backward compatible.

ceford:
@PhilETaylor on your last comment I think that is all what I said or meant. You are working too hard. But keep going!

PhilETaylor:
My last part was disagreeing with @HLeithner

HLeithner:
In the end it's like this several years so better documentation follows code then the way around (at least in this case)

PhilETaylor:
Don't get me started with how bad some of the docblocks are in joomla 3 haha

PR #33007, opened 2021-04-04
[4] Cast data object to string

PhilETaylor:
Code review

When `$data` is passed into this method it has type `object` according to the docblock for this method.

There is already a case on line 447 where the object is cast to string

`$source->url` has a docblock type (in DownloadSource.php) of string.

Therefore assign an object to an object property of type string is incorrect and we first need to cast to a string before assigning.

### DIFF
--- a/libraries/src/Updater/Update.php
+++ b/libraries/src/Updater/Update.php
@@ -453,7 +453,7 @@ class Update extends CMSObject
 		{
 			// Grab the last source so we can append the URL
 			$source = end($this->downloadSources);
-			$source->url = $data;
+			$source->url = (string) $data;
 
 			return;
 		}


PhilETaylor:
In fact this might not be right and needs further investigation of exactly what `$data` type and contains as Psalm is now contradicting the other cast saying an object cannot be cast to string! 

```
ERROR: InvalidCast - libraries/src/Updater/Update.php:447:76 - object cannot be cast to string (see https://psalm.dev/103)
			$this->currentUpdate->stability = $this->stabilityTagToInteger((string) $data);


ERROR: InvalidPropertyAssignmentValue - libraries/src/Updater/Update.php:456:19 - $source->url with declared type 'string' cannot be assigned type 'object' (see https://psalm.dev/145)
			$source->url = $data;
```

PR #33011, opened 2021-04-04
[4] append additional type to type hint

PhilETaylor:
Code review.

This method can and is called many times with a string as the value for `$button`

E.g:

```php
$bar->appendButton('Custom', $layout->render($options), 'versions');

$bar->appendButton('Standard', 'cancel', $alt, $task, false);
```

### DIFF
--- a/libraries/src/Toolbar/Toolbar.php
+++ b/libraries/src/Toolbar/Toolbar.php
@@ -159,8 +159,8 @@ class Toolbar
 	/**
 	 * Append a button to toolbar.
 	 *
-	 * @param   ToolbarButton  $button  The button instance.
-	 * @param   array          $args    The more arguments.
+	 * @param   ToolbarButton|string  $button  The button instance.
+	 * @param   array                 $args    The more arguments.
 	 *
 	 * @return  ToolbarButton|boolean  Return button instance to help chaining configure. If using legacy arguments
 	 *                                 returns true
--- a/libraries/src/Toolbar/Toolbar.php
+++ b/libraries/src/Toolbar/Toolbar.php
@@ -158,8 +158,9 @@ class Toolbar
 
 	/**
 	 * Append a button to toolbar.
+	 * @deprecated 5.0  Note that the passing of $button as a (string) is deprecated and will be removed with Joomla 5.0.
 	 *
-	 * @param   ToolbarButton|string  $button  The button instance. String type is deprecated and will be removed with Joomla 5.0
+	 * @param   ToolbarButton|string  $button  The button instance.
 	 * @param   array                 $args    The more arguments.
 	 *
 	 * @return  ToolbarButton|boolean  Return button instance to help chaining configure. If using legacy arguments
--- a/libraries/src/Toolbar/Toolbar.php
+++ b/libraries/src/Toolbar/Toolbar.php
@@ -159,7 +159,7 @@ class Toolbar
 	/**
 	 * Append a button to toolbar.
 	 *
-	 * @param   ToolbarButton|string  $button  The button instance.
+	 * @param   ToolbarButton|string  $button  The button instance. String type is deprecated and will be removed with Joomla 5.0
 	 * @param   array                 $args    The more arguments.
 	 *
 	 * @return  ToolbarButton|boolean  Return button instance to help chaining configure. If using legacy arguments


HLeithner:
You can but you shouldn't because it's deprecated maybe comment this in the parameter description?

PhilETaylor:
If the docblocks don't match real life then IDEs like phpStorm and static analysis like psalm will continue to error and moan.

libraries/src/Toolbar/Toolbar.php
```diff
@@ -159,8 +159,8 @@ public function setFactory(ToolbarFactoryInterface $factory): self
 	/**
 	 * Append a button to toolbar.
 	 *
-	 * @param   ToolbarButton  $button  The button instance.
-	 * @param   array          $args    The more arguments.
+	 * @param   ToolbarButton|string  $button  The button instance.
```
HLeithner, line 163:
```suggestion
	 * @param   ToolbarButton|string  $button  The button instance. String type is deprecated and will be removed with Joomla 5.0
```

I thought about something like this.

brianteeman: then it should be a proper deprecated note

ceford:
I have tested this item :white_check_mark: successfully on dec2be7f143ee9f4857e3408539259f323b2c1e6<br /><br />I have looked up the toolbar button code quite a bit so approve the docblock clarification that the parameter can be a ToolbarButton or a string, and the deprecation comment.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33011">issues.joomla.org/tracker/joomla-cms/33011</a>.</sub>

richard67:
@brianteeman I don't think we have a proper way for deprecation notes if the deprecation only concerns one of the possible data types of a mixed parameter. But I'm not sure with that. To me it seems ok like it is now in this PR.

brianteeman:
It will be ignored by any tooling that reports on code being deprecated

richard67:
> It will be ignored by any tooling that reports on code being deprecated

Yes, I saw. But do you know a way to make it ok for the tools?

PhilETaylor:
It needs a `@deprecated` tag

richard67:
> It needs a `@deprecated` annotation

Yes, but how is it made for a function parameter with mixed type where only one of the possible types is deprecated? I can't remember having ever seen an example for that.

wilsonge:
I don't believe there's any deprecated annotations for parameters. You just write it in the method description (and can choose to use the @note annotation as well) - at least to the best of my knowledge. Not sure where the latest developments on the PSR-5 saga have got to...<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33011">issues.joomla.org/tracker/joomla-cms/33011</a>.</sub>

richard67:
I have tested this item :white_check_mark: successfully on 35f71bfed06f8448366eaa78567c22941b0d42d8<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33011">issues.joomla.org/tracker/joomla-cms/33011</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33011">issues.joomla.org/tracker/joomla-cms/33011</a>.</sub>

richard67:
Previous test by @ceford was still valid since last change after that was just a change on the deprecation comment. It has a correct annotation now so can be found by tools.

HLeithner:
if you use string now you get a deprecation, if you read the documentation you don't see any deprecation. So I would simple add the note to the parameter or what george suggested use the @note annotation please?

rdeutz:
@PhilETaylor could you add a note and we are good to merge this, thanks.

PhilETaylor:
Using `@note` will not achieve the aim of this PR which is to have the deprecation shown in IDE and static analysis tools. Those tools need the `@deprecated` tag.

SharkyKZ:
I have tested this item :red_circle: unsuccessfully on 35f71bfed06f8448366eaa78567c22941b0d42d8<br /><br />This method is not deprecated.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33011">issues.joomla.org/tracker/joomla-cms/33011</a>.</sub>

PR #33023, opened 2021-04-05
[4.x] More flexible display of relative dates.

Harmageddon:
### Summary of Changes

This is an alternative solution to #31675 and fixes the concerns discussed in https://github.com/joomla/joomla-cms/pull/32911#issuecomment-809301133 and the following comments.
This PR introduces a new HTMLHelper method called `date.relativeFormatted` which can be used to display dates in a relative format without losing important information. 

Key features:
- Relative or absolute format is chosen based on user preference.
- Using the flag `$forceRelative`, relative display can be forced, ignoring the user's preference (use carefully!).
- In addition to the relative date display, the full absolute date is shown either in a tooltip or as small text below (like it is currently in the actionlogs / privacy components). This can be deactivated via method parameter.
- If we use the same format for an absolute display as for the full date (DATE_FORMAT_LC6), it is not displayed twice.

I also added unit tests for the new method.

### Testing Instructions

1. Log in to backend.
2. In the user menu (top right), click "Edit Account". In the Tab "Basic Settings", you should find a new parameter "Show Relative Dates in Lists" that defaults to true.
3. Check the following views with this parameter once enabled and once disabled:
   - Module "Latest Actions" on the administrator home dashboard.
   - Module "Logged-in users" on the administrator home dashboard (**attention:** not yet implemented, only after #32911 is merged and I updated this PR here accordingly)
   - Users - User Actions Log
   - Users - Privacy - Requests: You have to create a new request here if there are none yet.
   - Users - Privacy - Consents: You have to give consent with at least one user beforehand. To do this, enable the plugin "System - Privacy Consent" and log in to frontend with a user account.

joomla-pr-date-relative-below

The dates in the mentioned components and modules are relative for dates which are more recent than a month. For user actions log and the privacy views, the full date is displayed below the relative date.

joomla-pr-date-relative

For the dashboard modules, the full date is not shown anywhere.

#### With "Show Relative Dates in Lists" enabled

joomla-pr-date-relative-tooltip

In the dashboard modules, the relative dates have a tooltip now showing the full exact date of the action.

In the actions log and privacy views, the display is the same as before.

#### With "Show Relative Dates in Lists" disabled

joomla-pr-date-absolute-below

Dates in the mentioned locations are displayed in an absolute format. The full date and time is shown below or in the tooltip, respectively.

Maybe we need to document this user parameter somewhere?

### Open Discussion Points

1. Should we include a default setting for users who have not set this parameter by themselves? Or just default to relative display (the way this PR does it now).
2. If date display is set to relative, the date format is DATE_FORMAT_LC6, and the date is more than one month in the past, we get the same date twice. The full date tooltip or div does not provide any additional information in this case. Should we try to detect this case or just leave it that way?
3. Which date format to use for the actions log / privacy components? Currently, it defaults to DATE_FORMAT_LC1, which is "Monday, 05 April 2021". In these list formats, something like DATE_FORMAT_LC5 (2021-05-04 19:00) or DATE_FORMAT_LC6 (2021-05-04 19:00:00) might be better?
4. Is there a chance to get this in 4.0 or should I directly re-target it to 4.1? In my view, #32910 and #32911 (one already merged into 4.0, one RTC) take away some information from these modules. This PR would "fix" this. But I also see that this might count as a new feature, so 4.1 would be okay as well I guess.


administrator/components/com_users/forms/user.xml
```diff
@@ -197,6 +197,17 @@
 				<option value="">JOPTION_USE_DEFAULT</option>
 			</field>
 
+			<field
+				name="use_relative_dates"
+				type="radio"
+				label="COM_USERS_USER_FIELD_USE_RELATIVE_DATES_LABEL"
+				layout="joomla.form.field.radio.switcher"
+				default="1"
+			>
```
brianteeman:
```suggestion
				>
```

brianteeman:
Please dont use bootstrap tooltips for this - they are horrible for accessibility.

Please look at the way the code is done for the search box and use the pseudo tip code for that

PhilETaylor:
image

Is this screenshot wrong? Whats the point of a tool tip that shows the same as the text hovered over?

brianteeman:
BTW I think we used to have this functionality and it was removed

PhilETaylor:
Not to kill contributions, but personally, as I have already commented elsewhere:

> The home dashboard is an overview of the RECENT happenings on a site therefore it MAKES SENSE that this screen should have relative dates.

> On normal listing screens date stamps are better, and can be ordered etc,

And as you have said... "do we really need another option"....

PhilETaylor:
Although I do agree a RELATIVE date should have a popover with the full timestamp - that makes sense.

brianteeman:
> Although I do agree a RELATIVE date should have a popover with the full timestamp - that makes sense.

I think if my memory is correct the problem was that the relative date was only for the first month and after that you got the real date and then the tooltip was a duplicate. But my memory might be tricking me

ChristineWk:
> Although I do agree a RELATIVE date should have a popover with the full timestamp - that makes sense.

Do you mean in this way: Patch: ON, View enabled:

sorry, hv trbls to add screenshot ....

PhilETaylor:
I mean like [timeago for jQuery](https://timeago.yarp.com) has been doing for a decade. 

Screenshot 2021-04-06 at 17 14 30

Harmageddon:
> Please dont use bootstrap tooltips for this - they are horrible for accessibility.
> 
> Please look at the way the code is done for the search box and use the pseudo tip code for that

Done, thank you!


> image
> 
> Is this screenshot wrong? Whats the point of a tool tip that shows the same as the text hovered over?

That's a good point. Technically, it's not the same (in case of absolute formatting and the same date format, the tooltip would not be shown), but it's very close. I added the tooltip for situations where we use a less precise format like the second one you showed here: https://github.com/joomla/joomla-cms/pull/33023#issuecomment-814248799

> I think if my memory is correct the problem was that the relative date was only for the first month and after that you got the real date and then the tooltip was a duplicate. But my memory might be tricking me

I don't know if that was the reason back then, but it is an issue, yes (see discussion point 2 in the description). I could try to detect and avoid this case, but then we would have an inconsistency between "two days ago (tooltip: 2021-04-04 19:05:00)" and "2021-02-02 19:00:00 (without tooltip)".

> Not to kill contributions, but personally, as I have already commented elsewhere:
> 
> > The home dashboard is an overview of the RECENT happenings on a site therefore it MAKES SENSE that this screen should have relative dates.
> 
> > On normal listing screens date stamps are better, and can be ordered etc,
> 
> And as you have said... "do we really need another option"....

As discussed in #31675, we do need another option, at least for the list case, because different users have different preferences there. Why not use one option for all instead of adding such an option everywhere?

ChristineWk:
Although I do agree a RELATIVE date should have a popover with the full timestamp - that makes sense.

Do you mean in this way: Patch: ON, View enabled:

screen shot 2021-04-06 at 18 02 44<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33023">issues.joomla.org/tracker/joomla-cms/33023</a>.</sub>

PhilETaylor:
that's what I would expect - yes. 

Whenever a relative date is displayed, one should always be able to see the actual date in a popover 

I know what 8 mins ago was - that is easy, but when it says 3 weeks ago, I need to know what date that was... that's not easy to quickly do in ones head.

ChristineWk:
The other point:



    I think if my memory is correct the problem was that the relative date was only for the first month and after that you got the real date and then the tooltip was a duplicate. But my memory might be tricking me

@ Harmageddon wrote: I don't know if that was the reason back then, but it is an issue, yes (see discussion point 2 in the description). I could try to detect and avoid this case, but then we would have an inconsistency between "two days ago (tooltip: 2021-04-04 19:05:00)" and "2021-02-02 19:00:00 (without tooltip)".

sorry, I can't send screenshot via Github, therefore here:

screen shot 2021-04-06 at 18 16 44<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33023">issues.joomla.org/tracker/joomla-cms/33023</a>.</sub>

libraries/src/HTML/Helpers/Date.php
```diff
@@ -90,4 +92,62 @@ public static function relative($date, $unit = null, $time = null, $format = nul
...
+	 *                                            - 'below': Display the full date below the relative date.
+	 *                                            - 'hide': Don't display the full date.
+	 *
+	 * @return  string  The relative or absolute date, plus the full date if applicable.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function relativeFormatted(
+		$date, $unit = null, $time = null, $format = null, $forceRelative = false, $showAbsoluteDate = 'tooltip'
+	)
```
Quy:
```suggestion
		$date, $unit = null, $time = null, $format = null, $forceRelative = false, $showAbsoluteDate = 'tooltip')
```

libraries/src/HTML/Helpers/Date.php
```diff
@@ -90,4 +92,62 @@ public static function relative($date, $unit = null, $time = null, $format = nul
...
+		$dateAbsolute = HTMLHelper::_('date', $date, $formatAbsolute);
+
+		if ($showAbsoluteDate === 'tooltip')
+		{
+			return '<span>' . $dateMain . '</span><div role="tooltip">' . $dateAbsolute . '</div>';
+		}
+
+		if ($showAbsoluteDate === 'below')
+		{
+			return $dateMain . '<div class="small">' . $dateAbsolute . '</div>';
```
dgrammatiko: Use `<date>`

libraries/src/HTML/Helpers/Date.php
```diff
@@ -90,4 +92,62 @@ public static function relative($date, $unit = null, $time = null, $format = nul
...
+		{
+			return HTMLHelper::_('date', $date, $format);
+		}
+
+		$dateMain = $useRelative ? HTMLHelper::_('date.relative', $date, $unit, $time, $format) : HTMLHelper::_('date', $date, $format);
+		$dateAbsolute = HTMLHelper::_('date', $date, $formatAbsolute);
+
+		if ($showAbsoluteDate === 'tooltip')
+		{
+			return '<span>' . $dateMain . '</span><div role="tooltip">' . $dateAbsolute . '</div>';
```
dgrammatiko: DO NOT use tooltip...

libraries/src/HTML/Helpers/Date.php
```diff
@@ -90,4 +92,62 @@ public static function relative($date, $unit = null, $time = null, $format = nul
...
+	 *                                            - 'tooltip': Display the full date in a tooltip.
+	 *                                            - 'below': Display the full date below the relative date.
+	 *                                            - 'hide': Don't display the full date.
+	 *
+	 * @return  string  The relative or absolute date, plus the full date if applicable.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function relativeFormatted(
+		$date, $unit = null, $time = null, $format = null, $forceRelative = false, $showAbsoluteDate = 'tooltip'
```
dgrammatiko: Any function with more than 4 arguments is realistically unusable (nobody will ever remember 6 params in a row, never).
Also if you want to have Bootstrap mixed here I would suggest creating a JLayout instead of an HTMLHelper. HTMLHelpers need a plugin for an override, JLayouts not and because not everybody is happy with BS easy overriding is better for the future of the project

PhilETaylor: > nobody will ever remember 6 params in a row, never

Thats why we use IDE's like phpStorm hahahahahahah

dgrammatiko: Dude, it's not about IDE's it's about **WRONG** patterns, according to everybody in the business before we were even born (ref: https://softwareengineering.stackexchange.com/a/145073 )

PhilETaylor: just pulling your leg :)

BertaOctech:
I have tested this item :white_check_mark: successfully on 9bbe5e67a51e9b63a4139aee28d2632617fa2b02<br /><br />This patch works as described.

After applying the patch a new option appears in the User Profile. When "Show Relative Dates in Lists = NO" is selected, the dates in the action logs appear as absolute dates. Otherwise they appear as relative dates.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33023">issues.joomla.org/tracker/joomla-cms/33023</a>.</sub>

libraries/src/HTML/Helpers/Date.php
Harmageddon: Do you have an alternative suggestion? Always displaying the full date below the relative one might lead to crowded layouts in some cases, and confuse rather than help users. And not displaying this additional information at all seem worse to me than a tooltip.

libraries/src/HTML/Helpers/Date.php
Harmageddon: Is this a custom element? Can't find it anywhere in core or the HTML5 reference. Or are you referring to `<time>`?

libraries/src/HTML/Helpers/Date.php
Harmageddon: > Any function with more than 4 arguments is realistically unusable (nobody will ever remember 6 params in a row, never).

True. But my goal was to create a rather generic function that can be used in various places throughout the core for different displays of dates. I could rearrange the parameters, so the (expectedly) less used parameters can be omitted and use default values. 

> Also if you want to have Bootstrap mixed here I would suggest creating a JLayout instead of an HTMLHelper. HTMLHelpers need a plugin for an override, JLayouts not and because not everybody is happy with BS easy overriding is better for the future of the project

Following the suggestion of Brian above, there is no Bootstrap anymore in this PR.

libraries/src/HTML/Helpers/Date.php
Harmageddon: @Quy Hmm, PHPCS has a different opinion about that: https://ci.joomla.org/joomla/joomla-cms/41800/1/6 What is right here?

libraries/src/HTML/Helpers/Date.php
dgrammatiko: Yes: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time
Also FYI there's a `<small>` element also so div and span should be really avoided.

The point of using the proper element is that this text doesn't convey any message for disabled people but `time` does.

Harmageddon: Better like this?

dgrammatiko:
@Harmageddon sorry but this is not quite ok. Do **NOT** use HTMLHelper here, use JLayout, eg create a file in `layouts/joomla/html/relative-date.php` with content:

```php
<?php
/**
 * @package     Joomla.Site
 * @subpackage  Layout
 *
 * @copyright   (C) 2016 Open Source Matters, Inc. <https://www.joomla.org>
 */

use Joomla\CMS\Language\Text;

defined('_JEXEC') or die;

extract($displayData);

/**
 * Layout variables
 * -----------------
 * @var   boolean  $forceRelative
 * @var   string  $format
 * @var   string  $date
 * @var   string  $showAbsoluteDate
 * @var   string  $unit
 * @var   string  $time
 */

$user = Factory::getApplication()->getIdentity();

if ($user === null)
{
	$useRelative = true;
}
else
{
	$useRelative = $forceRelative || $user->getParam('use_relative_dates', true);
}

// Format for the full / absolute date display.
$formatAbsolute = Text::_('DATE_FORMAT_LC6');

if (!$useRelative && $format === $formatAbsolute)
{
	return HTMLHelper::_('date', $date, $format);
}

$dateMain = $useRelative ? HTMLHelper::_('date.relative', $date, $unit, $time, $format) : HTMLHelper::_('date', $date, $format);
$dateAbsolute = HTMLHelper::_('date', $date, $formatAbsolute);

if ($showAbsoluteDate === 'tooltip')
{
	echo '<span>' . $dateMain . '</span><time role="tooltip">' . $dateAbsolute . '</time>';
}

if ($showAbsoluteDate === 'below')
{
	echo $dateMain . '<time class="small d-block">' . $dateAbsolute . '</time>';
}

echo $dateMain;

````

and then call it like:
```php

<?php echo \Joomla\CMS\Layout\LayoutHelper::render(
	'joomla.html.relative-date',
	[
		'date'   => $item->log_date,
		'format' => Text::_('DATE_FORMAT_LC5'),
		$showAbsoluteDate = 'tooltip',
		$time = $now,
		$forceRelative = false,
		$unit = null
	]
); ?>
````

The reason is simple: using JLayout is easily overridable for front end devs. HTMLHelper need a plugin in order to be overriden

**Comment by dgrammatiko** [comment]
Actually, reading https://github.com/joomla/joomla-cms/pull/33023#issuecomment-814248799 I think the idea of processing this data in the server is wrong, should be done on the client-side using Temporal: https://github.com/tc39/proposal-temporal and a simple web component.

Sorry, the JS is not temporal but https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

Actually @PhilETaylor I think your PR that introduced the relative dates is wrong, we can do this using the browser native power... ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat
)

File: libraries/src/HTML/Helpers/Date.php (hunk 2 above)
Comment (Quy) [reply libraries/src/HTML/Helpers/Date.php]: Per [Joomla! Coding Standards](https://developer.joomla.org/coding-standards/docblocks.html):

> If a function definition goes over multiple lines, all lines must be indented with one tab and the closing bracket must go on the same line as the last parameter.

File: administrator/components/com_privacy/tmpl/consents/default.php [hunk 6]
```diff
@@ -104,10 +104,7 @@
 								<?php echo $item->body; ?>
 							</td>
 							<td class="break-word">
-								<?php echo HTMLHelper::_('date.relative', new Date($item->created), null, $now); ?>
```
Quy, line 107: Please delete `use Joomla\CMS\Date\Date;`

michwebdev:
I have tested this item :white_check_mark: successfully on 9ccc859f9885e2990a6848b9ec3f562d5c0b9123<br /><br />This worked.  I had to learn about setting up a menu item for Privacy --> Create Request.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33023">issues.joomla.org/tracker/joomla-cms/33023</a>.</sub>

obuisard:
I have tested this item :white_check_mark: successfully on 9ccc859f9885e2990a6848b9ec3f562d5c0b9123<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33023">issues.joomla.org/tracker/joomla-cms/33023</a>.</sub>

richard67:
It's still draft and so tests don't count.

richard67:
@Harmageddon @dgrammatiko Is there anything left to be done or clarified regarding the above discussions? Shall I set the "Updates Requested" label?

dgrammatiko:
>  Is there anything left to be done or clarified regarding the above discussions?

I'll stick to my opinion that this should **NOT** be an HTMLHelper thing but a JLayout, HTMLHelper should be used only for PHP functionality, if anything is echoed then that should be done directly with Layouts

Harmageddon:
I see the point in making it a layout, but haven't found the time yet to convert and test it. I'm not sure about the web component proposal though, because apart from shifting the translation from our own translation files to the browser, I don't see great advantages, as we need to do the same calculations as far as I understand. 

Thank you for testing this already, but as @richard67 said, this is still in draft state because of the discussions and open points listed above. As soon as it is ready for testing, I'm going to mark it as such.

> Shall I set the "Updates Requested" label?

Does that make any difference while the PR is in draft state? If yes, feel free to do so.

richard67:
> Does that make any difference while the PR is in draft state?

Right, draft state is sufficient.

richard67:
@Harmageddon @wilsonge @bembelimen Would it make sense to move this PR to 4.1?

PhilETaylor:
I never thought something as simple as relative dates could be as controversial as it has been.

Thanks for the link for Intl.RelativeTimeFormat - I have certainly never seen this before, and Regli use JavaScript libraries (timeago.js and jQuery) ones to do this.

Harmageddon:
> @Harmageddon @wilsonge @bembelimen Would it make sense to move this PR to 4.1?

As I sadly don't have the time right now to proceed with this, I assume it's not going to make it into 4.0. So yes, I'd re-target this for 4.1 and come up with a more consistent and overridable solution when I find the time.

HLeithner:

drmenzelit:
The feature is interesting, but due to inactivity we are closing this draft. You can open it again, when you have the time to rework it.

brianteeman:
@rdeutz 

> #PROD2020/024 - There shall be no indiscriminate closing of Issues or Pull Requests
Closing of Issues and Pull requests PR's shall NOT be done without underlying rationale and individual reasoning in the Issue or PR. Elapsing of a time period, inactivity of the submitter are examples that do NOT constitute valid reasons in and by themselves.

rdeutz:
This motion is undicided and the maintainers have decided how we will do the Handling of abandoned PR’s.

> The maintainers team will check older (older means, no relevant activity for a certain time frame, 18 months) Pull Requests and discuss if a Pull Request should be closed.

brianteeman:
> This motion is undicided and the maintainers have decided how we will do the Handling of abandoned PR’s.

My apologies - I missed thaat part

PR #33031, opened 2021-04-06
proper html syntax

cric2000:
Pull Request for Issue # .

### DIFF
--- a/tmp/index.html
+++ b/tmp/index.html
@@ -1,1 +1,5 @@
-<!DOCTYPE html><title></title>
+<!DOCTYPE html>
+<head>
+<title></title>
+</head>
+</html>


brianteeman:
This is just as invalid as the original markup. The original file is adequate for the purpose it serves. Plus if you look more closely you will find that there are about 50 files that would need to be updated not just this one. Finally when you submitted the pull request you were asked 5 questions that you need to answer.

Fedik:
The existing html is pretty valid (if not count empty title tag).
Nothing to fix here.

richard67:
@cric2000 First of all thank you for your first pull request. Unfortunately it has a few mistakes:

1. When you create a new pull request (PR) you can see there are headings like "Summary of Changes", "Testing Instructions". They are not there for nothing. We assume that the creator of the PR adds the corresponding information below these headings. For testing instructions in such a case like here a "Code review" would be sufficient.
2. The index.html files in some of Joomla's folders are not meant to be shown in a browser. They are dummy files in otherwise empty folders which are needed to create the zip packages, because zip by default doesn't pack empty folders. In past we had these files in all folders in order to prevent directory listing in the browser. But this is not necessary anymore since directory listing is blocked in every reasonable webserver configuration. So these files don't require things like "normal" html pages should contain.
3. As @Fedik pointed out, the html syntax of these files is already valid. Maybe you have adjusted the validator for the wrong kind of html version so it has complained.

I hope you understand now why I close this PR.

Please don't be disappointed. It can happen even to experienced contributors that we make wrong PR's.

The next one will be better :smile:

Issue #33042, opened 2021-04-06
[4.0] Collapse in categories view doesn't work if navbar is not present


drmenzelit:
### Steps to reproduce the issue
Create a menu in position `menu`
Create some categories with subcategories and some articles.
Create a menu item of type `List All Categories in an Article Category Tree` 

Go to the categories page. You will see a list of categories and a + symbol in the category with a subcategory (Blog)
![grafik](https://user-images.githubusercontent.com/9153168/113733629-ef8e4180-96fa-11eb-9324-3175c3f907be.png)

When you click on the + symbol the category expands and you see the subcategory (General)
![grafik](https://user-images.githubusercontent.com/9153168/113733778-12b8f100-96fb-11eb-80eb-28a7dc937d6a.png)

**Now unpublish the menu**

### Expected result
The expand / collapse function works

### Actual result
The expand / collapse function doesn't work

### Explanation
The `com_content/tmpl/categories/default.php` file loads a script for accordion `$wa->useScript('com_categories.shared-categories-accordion');` that only changes the `aria-label` and the `icon` depending on the status expanded or collapsed. 
The collapse function itself is on a Javascript from Boostrap (`vendor/bootstrap/collapse.min.js`), that is not being loaded in the view.
In Cassiopeia the collapse Javascript is being loaded when a menu is published in the position `menu`. 
`<nav class="navbar navbar-expand-md">
	<?php HTMLHelper::_('bootstrap.collapse', '.navbar-toggler'); ?>
...
`
That is why the collapse in the categories list only works if the menu is present.

### Solution

- Load the collapse script from Bootstrap in the view or
- Create an own collapse / accordion function (e.g. extend the functionality of the `com_categories.shared-categories-accordion` script). A good example can be found here: https://inclusive-components.design/collapsible-sections/

chmst:
Related to #33037, the buttons needs a complete redesign on layout and function

drmenzelit:
I'm working on a solution. PR coming soon

drmenzelit:
Closing as we have a PR #33052

Issue #33037, opened 2021-04-06
[4.0] [a11y] View categories needs improvement


chmst:
### Steps to reproduce the issue
Make a menuItem to a category List.

The result looks like this: the Category Help has a subcategory but the button for expanding is hardly visible. 
All items (including the badges) are tagged as h3, which is wrong in most cases not only accessibility
The tree sstructure is not visible at all. 

### Expected result
A propery looking tree structure with buttons for expand/collapse near to the respective category.

### Actual result
![grafik](https://user-images.githubusercontent.com/1035262/113713095-36723c00-96e7-11eb-8ae4-57c9a1587dcf.png)

brianteeman:
Everything is inside a single h3 🤦

drmenzelit:
Closing as we have a PR #33052

PR #33052, opened 2021-04-07
[4.0] Collapse / accordion function for categories list

drmenzelit:
Pull Request for Issue #33042 and #33037  .

### Summary of Changes
Extended existent Javascript (`shared-categories-accordion.js`) to a collapse function for categories with subcategories.
Removed Bootstrap collapse attributes
Added a css for the view

### Testing Instructions
See issues

See issues

The collapse function works with and without published menu (see explanation in issue #33042 

grafik


chmst:
What do you think about removing line 29 in  components/com_content/tmpl/categories/default_items.php? Or at least hide it? 
We have nowhere else a text for article counts.

drmenzelit:
@Fedik it is possible to solve this with Joomla custom elements? I saw a collapse.js, but I'm not good in Javascript and I'm not sure if it can be used in replacement of Bootstrap collapse.

Fedik:
What you did is also okay.

If you mean bootstrap/collapse.js, it part of Bootstrap 5 changes 
https://getbootstrap.com/docs/5.0/components/collapse/

dgrammatiko:
>  it is possible to solve this with Joomla custom elements?

Yes and No. Yes as the collapse.js will do the accordion thing but No as the repo is basically abandoned. Even the existing elements that Joomla is using have several defects...

@Fedik she means https://github.com/joomla-projects/custom-elements/blob/master/src/js/collapse/collapse.js which in theory should be CSS framework agnostic

drmenzelit:
@Fedik , @dgrammatiko thank you guys for the answers

hans2103:
@drmenzelit what if... we change it into `<detail>` and `<summary>`? 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

It will collapse out of the box.

dgrammatiko:
> what if... we change it into <detail> and <summary>

We are already doing it in https://github.com/joomla/joomla-cms/blob/f4ac18829fad6eadd7260ac2f951aca0c7da459b/build/media_source/system/js/fields/joomla-image-select.w-c.es6.js#L254-L295

drmenzelit:
> @drmenzelit what if... we change it into `<detail>` and `<summary>`?
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
> 
> It will collapse out of the box.

That is a good idea! I will look into it

drmenzelit:
summary / detail will not fit in this case, because we need a link to the category too.

joomdonation:
I tested this and it worked. So if one of our frontend dev looks at this and confirm that the solution is OK, I will post my test result. Hope it will be merged before next RC.

Fedik:
It looks good to me, 
I only think that the CSS added here need to be part of the template style

drmenzelit:
I think a little bit of CSS is important to make the categories list looks good independently of the template used (we have CSS also on some core modules). We can improve things in Cassiopeia and other templates can use or override the CSS here.

dgrammatiko:
> I only think that the CSS added here need to be part of the template style

No, the css should not be monolithic. Joomla is modular and the css/js should be modular

joomdonation:
@dgrammatiko So this PR is fine as how it is? Or should the css/js code be moved to system assets so that we can use accordion anywhere we want? If it is fine, I will submit my test result as I tested the feature and it is working well.

dgrammatiko:
> Or should the css/js code be moved to system assets so that we can use accordion anywhere we want?

Good question. Turns out that this is the 3rd accordion that Joomla 4 will ship:
- Bootstrap accordion
- Joomla Custom Elements tabs (which also renders accordions)
- this one

But then again if this is working as expected and passes the a11y reqs it's fine

drmenzelit:
>  
> > it is possible to solve this with Joomla custom elements?
> 
> Yes and No. Yes as the collapse.js will do the accordion thing but No as the repo is basically abandoned. Even the existing elements that Joomla is using have several defects...
> 
> @Fedik she means https://github.com/joomla-projects/custom-elements/blob/master/src/js/collapse/collapse.js which in theory should be CSS framework agnostic

@dgrammatiko I asked for Joomla Custom Elements before... 
I don't mind, if this PR get rejected, if there is another thing that do the job. I wanted to have a framework agnostic solution....

dgrammatiko:
> @dgrammatiko I asked for Joomla Custom Elements before...

I know and I answered that you shouldn't use that code (collapse) which is still true today. What changed is that I rewrote from scratch the Tabs but for this particular instance a plain (ie non CE) implementation is better (ie less code)
The PR is totally fine as is

dgrammatiko:
@drmenzelit since the js file is loaded as a module the code shouldn't use the iife, should be like:

```js
/**
 * @copyright  (C) 2018 Open Source Matters, Inc. <https://www.joomla.org>
 */

if (!Joomla || !Joomla.Text) {
  throw new Error('core.js was not properly initialised');
}

// Selectors used by this script
const buttonsSelector = '[id^=category-btn-]';

/**
 * Handle the category toggle button click event
 * @param event
 */
const handleCategoryToggleButtonClick = ({ currentTarget }) => {
  const button = currentTarget;
  const icon = button.querySelector('span');

  // Toggle icon class
  icon.classList.toggle('icon-plus');
  icon.classList.toggle('icon-minus');

  // Toggle aria label, aria-expanded
  const ariaLabel = button.getAttribute('aria-label');
  const ariaExpanded = button.getAttribute('aria-expanded');
  button.setAttribute(
    'aria-label',
    (
      ariaLabel === Joomla.Text._('JGLOBAL_EXPAND_CATEGORIES') ? Joomla.Text._('JGLOBAL_COLLAPSE_CATEGORIES') : Joomla.Text._('JGLOBAL_EXPAND_CATEGORIES')
    ),
  );
  button.setAttribute(
    'aria-expanded',
    (
      ariaExpanded === 'false' ? 'true' : 'false'
    ),
  );

  const target = button.nextElementSibling;
  target.toggleAttribute('hidden');
};

Array.from(document.querySelectorAll(buttonsSelector)).forEach((button) => {
  button.addEventListener('click', handleCategoryToggleButtonClick);
});
```

Fedik:
> No, the css should not be monolithic. Joomla is modular and the css/js should be modular

If you look in CSS it doing nothing "category specific", it just a general styling like `display:flex` and stuff, which I suspect can be done with `.d-flex` and other bootstrap. classes.

It not much important to me, I only saying that it can be done without this file

drmenzelit:
I have modified the JS file as suggested by @dgrammatiko

dgrammatiko:
@drmenzelit you have to push the changes to Github...

drmenzelit:
> @drmenzelit you have to push the changes to Github...

My notebook is driving me crazy today... push is done now

joomdonation:
I have tested this item :white_check_mark: successfully on acca0c35e1409885a090f7bf05104a94fc44cb11<br /><br />I don't have good enough frontend skill to confirm the solution is valid or not. But it is working as expected, tested it again today.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33052">issues.joomla.org/tracker/joomla-cms/33052</a>.</sub>

hans2103 (approved):

hans2103:
I have tested this item :white_check_mark: successfully on acca0c35e1409885a090f7bf05104a94fc44cb11<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33052">issues.joomla.org/tracker/joomla-cms/33052</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33052">issues.joomla.org/tracker/joomla-cms/33052</a>.</sub>

wilsonge:
One question about presets otherwise fine to merge for 4.0.0 as this is core functionality. Thanks for reviewing the JS @dgrammatiko !

wilsonge:
Thanks!

PR #33065, opened 2021-04-08
[4.0] Add missing use statement

joomdonation:
Pull Request for Issue # .

### Summary of Changes
The **UpdateModel** class use **Folder::create** method, so we need to add this missing statement so that the class is available for using. I found this issue while testing opcache PR from Phil.

### Testing Instructions
Code review


wilsonge:
Merging on review - thanks!

Issue #31242, opened 2020-10-26
[4] Atum. Sidebar menu. Only 1 expand/collapse action possible when animation disabled in Windows 10 OS.


ReLater:
### Steps to reproduce the issue

Supplement after comment https://github.com/joomla/joomla-cms/issues/31242#issuecomment-716818432:
DIsable animations in your OS. Windows 10 in my case.

- I see this issue on XAMPP 7.3 and on an online installation.
- First one uses a custom sidebar menu (preset menu). Latter one the preinstalled sidebar menu.
- Both are updated nightly 4 builds.

- When I click on a first level menu item (e.g. "Content"), area with arrow-right icon, to expand the submenu items I can't close the expanded part with an additional click afterwards.
- And I can't open another first level menu item (e.g. "Menus") with click.
- Menu actions (expand/collapse) are frozen.
- No JavaScript errors displayed.  
- As far as I see all relevant JS libraries/files are loaded.

- I have already overwritten all core files on the XAMPP version. No change.
- Browser cache. No change.
- Browsers: Current edge/chromium, current Firefox.

Can somebody confirm the issue?

richard67:
@ReLater What do you mean with "updated nightly 4 builds"? Updated from 3.10-dev or alpha to a 4.0 nightly? Or updated a previous 4.0 nightly to a later 4.0 nightly? You know that the latter is not supported because we might have to (and have) changed existing 4.0 update SQL scripts which have not been released yet.

You can see that in the release announcements of the 4.0 Beta versions, e.g. the last one [https://www.joomla.org/announcements/release-news/5822-joomla-4-0-0-beta4-and-joomla-3-10-alpha2.html](https://www.joomla.org/announcements/release-news/5822-joomla-4-0-0-beta4-and-joomla-3-10-alpha2.html), in section "How can I help Joomla 4 development?":

> To make the life of testers (and early adopters) easier, updating is possible from the Beta version to the next Beta, RC or stable version (we will remain strict in updating the deleted files and modifying the SQL changes).

As you can see, this doesn't include updates from nightly to nightly, because if we would do that we would have hundreds of 4.0 update SQL script meanwhile.

richard67:
P.S. And no, I neither can reproduce that nor have I ever observed that. It really seems your database has missed updates due to the reaon I have stated in my previous comment.

SharkyKZ:
Sounds like https://github.com/joomla/joomla-cms/issues/31106 if you have animations disabled.

[...]

PR #33069, opened 2021-04-08
[4.0] admin menu reduced motion

brianteeman:
pr for #31242

This is a temp fix until upstream addresses this https://github.com/onokumus/metismenujs/issues/9

The fix applied is the exact same one used by webaim.org so should be acceptable from an a11y perspective.

There will be no change for most users. It is only if they have reduced motion set in their operating system that they encounter the problem this fixes.

### DIFF
--- a/administrator/templates/atum/scss/template.scss
+++ b/administrator/templates/atum/scss/template.scss
@@ -76,7 +76,7 @@
     background-attachment: initial !important;
     transition-delay: 0s !important;
     // temp fix until https://github.com/onokumus/metismenujs/issues/9 is addressed
-    transition-duration: 0.001ms !important;
+    transition-duration: .001ms !important;
     animation-duration: 1ms !important;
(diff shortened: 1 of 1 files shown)


HLeithner:
Maybe a PR upstream would be more useful?

brianteeman:
@HLeithner it would be, but upstream hasnt responded yet at all to the issue report filed by @ciar4n

HLeithner:
Yes I have seen this, but also see that this repo never got a pull request^^

brianteeman:
@HLeithner one is javascript and one is css. I wouldnt dream of submitting a js pull request.

Not sure what you are trying to prove by blocking this one line pull request

HLeithner:
I'm not blocking, just try to prevent to add a `TODO` to the code base which maybe is never found again ;-)

But yeah doesn't look good for metismenujs maybe we should remove/replace it because it's unmaintained?

PR #33077, opened 2021-04-09
[4.0] com_installer FTP

brianteeman:
This PR corrects the error with the id of the username field
This PR changes the labels to FTP Username and FTP Password to be consistent with other instances of ftp login details eg com_joomlaupdate and coom_config
This PR changes the name of the fields to ftp_user and ftp_password  to be consistent with other instances of ftp login details eg com_joomlaupdate and coom_config

NOTE: I am unable to test if any of the ftp functionality even works (I suspect it doesnt)


administrator/language/en-GB/com_installer.ini
```diff
@@ -42,6 +42,8 @@ COM_INSTALLER_FIELD_EXTRA_QUERY_LABEL="Download Key"
 COM_INSTALLER_FIELD_NAME_LABEL="Name"
 COM_INSTALLER_FIELD_TYPE_LABEL="Type"
 COM_INSTALLER_FIELD_TYPE_LOCATION="Location"
+COM_INSTALLER_FIELD_FTP_PASSWORD_LABEL="FTP Password"
```
Quy, line 45: Please alpha sort.

joomdonation:
You cannot change name of the inputs like this. The reason is because the extension expects to receive data from username and password for ftp credentials (if provided)

Related code https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_installer/src/Model/InstallModel.php#L84

And https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Client/ClientHelper.php#L213

And I wonder why we allow entering ftp credentials in multiple places like this? Anyone really uses it? If FTP is really needed, isn't enable it in Global Configuration enough?

joomdonation:
> And I wonder why we allow entering ftp credentials in multiple places like this? Anyone really uses it? If FTP is really needed, isn't enable it in Global Configuration enough

Ah, maybe for security purpose because the credentials if entered in Global Configuration will be stored as plain text in configuration.php file

brianteeman:
@joomdonation @Quy I'm not going to do anything here until we get an answer about using ftp.

joomdonation:
Agree. We will wait for maintainer decision before doing anything further.

PhilETaylor:
Loving the addition of a new RLDQ label... I wonder how effective that will be at getting decisions from the projects leadership. One could argue that decisions should not be made by a single person (the release lead) but by a project architect team, who has vision, purpose and direction for the future of the product, and not just a single release...

PhilETaylor:
Following [the projects decision to remove the FTP Layer](https://github.com/joomla/joomla-cms/pull/33390#issuecomment-830000434) from Joomla 4 - this can now be closed.

brianteeman:
closed

PR #32315, opened 2021-02-05
[4.0] Deliver ES2015+ javascript

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Use `type=module` for ES2017 and `nomodule` for ES5 scripts

### Testing Instructions

Test every url Front End  and Back End to confirm that there's no console errors and all the interactive elements continue to work as before
Especially check the Admin login by insering a username but ignoring the password should raise a message (formbvalidation)
Set the editor to Codemirror and create an article and also check that editing a template file is working as before

### Documentation Changes Required

Just as a comment here: all `.es6.js` and `.w-c.es6.js` are handled by Rollup, meaning that any `import ` will be correctly bundled. Also the compiled `-es5.js` is again created by Rollup using Babel. This adds a delay on the build command BUT now all the ES6 files are ES2017 (meaning there is translation to a minimum version) against the previous state (ES6 files 
could have eg 2021 code that would be problematic for many browsers as the coverage is way bellow the 95% threshold, which is assumed the industry standard for the compatibility lower boundary. Worth watching https://www.youtube.com/watch?v=cLxNdLK--yI by @developit and @housseindjirdeh as this PR basically implements exactly that)

### Performance

The impact could be from 1s to 20s depending on the actual hardware

#### Before

The CI server
Screenshot 2021-02-25 at 13 26 35

My mac
Screenshot 2021-02-25 at 22 22 32

#### After

The CI server
Screenshot 2021-02-25 at 13 28 16
 
My mac
Screenshot 2021-02-25 at 22 20 23

Calling for testing @richard67 @Fedik @anibalsanchez @alikon @laoneo 

> how Joomla is doing the transition

- Started with the bootstrap PR: https://github.com/joomla/joomla-cms/pull/31990
- Then transitioned the Web Components/Custom Elements https://github.com/joomla/joomla-cms/pull/32171
- Finally this PR transitions all the remaining scripts

> what an extension developer should do to follow the steps

- If the dev is using Web Assets then the change is a simple switch of the definitions of the scripts (assuming that the source is modern JS and there is a bundler capable of producing ES2017 and ES5 scripts, continue reading).
Eg legacy script is, let's say some jQuery, so:

```json
    {
       "name": "com_example.awesome-script",
       "type": "script",
       "uri": "com_example/awesome-script.min.js",
       "dependencies": [
         "core"
       ],
       "attributes": {
         "defer": true
       }
``` 

[...]

Issue #25639, opened 2019-07-18
[4.0] Improve ES5 Generation with polyfills for IE


wilsonge:
Despite core is dropping formal support for IE in our template neither do we want to make it too hard to build frontend templates on IE. Basically in core we can use the ES6 files natively now and we also don't want to have to add IE code in this.

I believe we can use this https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#web-standards for the es5 files which will hopefully still be ie11 compat (but not used by core but provided as a service for templates who require it). And that will polyfill nodeList.foreach removing our requirement for example to use `[].slice.call()` from JS as an example

This came out of this closed issue https://github.com/joomla/joomla-cms/issues/25566

brianteeman:
@Fedik @dgrammatiko @wilsonge Is this still an issue?

dgrammatiko:
Partially. The core-js that @wilsonge mentioned here is already in our dependencies https://github.com/joomla/joomla-cms/blob/f4ac18829fad6eadd7260ac2f951aca0c7da459b/package.json#L77 **BUT** the ES5 rollup config needs some attention. I mean we can inflate every script by injecting the polyfills from core-js (easy fix I already have the code muted in place) or serve the polyfills as separate scripts (with a `nomodule` attribute) so we don't repeat the polyfills per script. Reference: https://github.com/joomla/joomla-cms/pull/32315#issuecomment-787427272

dgrammatiko:
@brianteeman this can be closed as there's a PR: https://github.com/joomla/joomla-cms/pull/33084

dgrammatiko:
This can be closed but there is still https://github.com/joomla/joomla-cms/issues/22164 which needs to be addressed (eg polyfills not provided by babel)

@Quy @alikon

joomdonation:
Closing as requested. Thanks @dgrammatiko

PR #33084, opened 2021-04-10
[4.0] ES5 scripts with included polyfills

dgrammatiko:
Pull Request for Issue #25639 .

### Summary of Changes

When I refactored the build tools I left the ES5 babel config basically unconfigured. The reason was I was expecting that we could come up with an action plan (eg a broader talk about polyfills, https://github.com/joomla/joomla-cms/pull/32315#issuecomment-787427272). This obviously didn't happen and so here's the missing configuration.
 

### Testing Instructions
Apply the branch
Run `npm i`
CHECK WITH IE and report any console errors.

**Attention** there are known missing polyfills that this PR **IS NOT** addressing. This PR just ensures modern code is transpilled so IE can run it (this includes also a gazillion of polyfills but there are many DOM related parts that are not polyfillable from babel)


dgrammatiko:
@Fedik can you test this one with IE11 and report back the errors logged?

Fedik:
well, let's forget about IE 11 :smile: 
I tested, but there bunch of syntax error, because arrow function and stuff, because of es6

I not sure it worth the time that much

dgrammatiko:
> well, let's forget about IE 11

Actually, you're right, IE will be dead in 4 months so there's not point https://death-to-ie11.com

Issue #32997, opened 2021-04-04
[4] Highlight CSS missing.


PhilETaylor:
### Steps to reproduce the issue

Joomla `4.0-dev`
Create an article with the word "create" and the word "admin" in the body text.
Ensure the `System - Highlight plugin` is enabled
View the article on the frontend
Then append the url you are on with `?highlight=WyJjcmVhdGUiLCJhZG1pbiJd` 

`WyJjcmVhdGUiLCJhZG1pbiJd` = base64encode of the array `["create","admin"]`

View the source code. 

### Expected result

The words `admin` and `create` to be highlighted. 

### Actual result

The HTML source is there, but no css is applied to "highlight" the terms.

```html
<span class="highlight" rel="create">create</span>

<span class="highlight" rel="admin">admin</span>
```

### System information (as much as possible)

Safari on mac

RahulJha393:
Hey !! I would love to contribute to this. How can I share my code to help you highlighting the terms (i.e. "create" and "admin") . Kindly guide me

prajjwal1231:
Hey @PhilETaylor 
I would like to contribute in this issue Pls assign it to me
& can you brief me in which color you want it to be highlighted.

infograf768:
@PhilETaylor 
the css exists but there is an error in it. Error comes from finder.css

we have
```
.result__item .highlight::before {
    content: '';
    position: absolute;
    z-index: -1;
    top: 50%;
    left: 50%;
    -webkit-transform: translate(-50%, -50%);
    -ms-transform: translate(-50%, -50%);
    transform: translate(-50%, -50%);
    width: calc(100% + 6px);
    height: calc(100% + 2px);
    background: hsla(60, 100%, 50%, .5);
    pointer-events: none;
    border-radius: 3px;
}
```
Looks like taking off `z-index: -1;` solves the issue. z-index can't work here as the body background is set to white.

![highlight](https://user-images.githubusercontent.com/869724/113545953-63d9bf80-95eb-11eb-9945-5843351dd89f.gif)

Change has to be done in
https://github.com/joomla/joomla-cms/blob/3ff6d0ccefb8f936349a9627db5d28ecab137b81/build/media_source/com_finder/css/finder.css#L116

RahulJha393:
I have done the changes https://github.com/RahulJha393/joomla-cms/commit/2f951b832cee31baeaa85cf855b92a97842ef4cf 
hope this solves the issue

infograf768:
[...]

PR #16357, opened 2017-05-30
[4.0] Replace highlighter javascript with mark.js

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
Replace Joomla's own implementation of highlighting with mark.js

### Testing Instructions
Check the smart search in the front end, you should get:
screen shot 2017-05-30 at 12 52 47

PR #33089, opened 2021-04-10
[4.0] highlight js

dgrammatiko:
Pull Request for Issue #32997 redo of #16357 .

### Summary of Changes
Use abetter script for marking HTML: https://markjs.io https://bundlephobia.com/result?p=mark.js@8.11.1

### Testing Instructions

After pulling this PR and running `npm ci`:

Create an article with the word "create" and the word "admin" in the body text.
Ensure the System - Highlight plugin is enabled
View the article on the frontend
Then append the url you are on with ?highlight=WyJjcmVhdGUiLCJhZG1pbiJd

WyJjcmVhdGUiLCJhZG1pbiJd = base64encode of the array ["create","admin"]

No highlight

Screenshot 2021-04-10 at 15 08 19

This should be 100% B/C but the idea is to deprecate as many HTMLHelpers as possible as they are really unfriendly for front end devs (nobody will ever create a Plugin just to adjust some HTML/JS/CSS and then have to maintain it)

@PhilETaylor can you give this a test?


infograf768:
Works OK but imho the default bootstrap `$mark-bg:                     #fcf8e3 !default;` is too weak.
we have
```
mark, .mark {
    padding: 0.2em;
    background-color: #fcf8e3;
}
```
gives
Screen Shot 2021-04-11 at 09 46 04

I suggest overriding it with a stronger hue in `.../templates/cassiopeia/scss/tools/variables/_variables.scss`

for example: `#fbeea8`

which would give

Screen Shot 2021-04-11 at 09 54 58

Fedik:
It looks good,
but I have a couple notes :)

build/media_source/system/joomla.asset.json
```diff
@@ -112,6 +112,29 @@
...
+      "attributes": {
+        "defer": true,
+        "nomodule": true
+      },
+      "dependencies": [
+        "core"
+      ]
+    },
+    {
+      "name": "js-highlight",
```
Fedik: please call it `highlighter` (similar to old js), or `highlight` if you like,
but not `js-potato` :wink:

layouts/joomla/highlight/highlight.php
```diff
@@ -0,0 +1,49 @@
...
+if (empty($terms))
+{
+	return;
+}
+
+$doc = Factory::getDocument();
+
+$doc->getWebAssetManager()->useScript('js-highlight');
+$doc->addScriptOptions(
+	'js-highlight',
```
Fedik: the same here, please call it `highlighter` or `highlight`

build/media_source/system/joomla.asset.json
dgrammatiko: done

layouts/joomla/highlight/highlight.php
```diff
@@ -0,0 +1,49 @@
...
+		'class'          => !empty($options['class']) ? $options['class'] : 'js-highlight',
+		'iframes'        => !empty($options['iframes']) ? $options['iframes'] : false,
+		'iframesTimeout' => !empty($options['iframesTimeout']) ? $options['iframesTimeout'] : 5000,
+		'debug'          => !empty($options['debug']) ? $options['debug'] : false,
+		'highLight'      => $terms,
+		'accuracy'       => 'partially',
+		'diacritics'     => true,
+		'exclude'        => !empty($options['exclude']) ? $options['exclude'] : [],
+
+		// For B/C with the old code!!!! will be remove @5.0
```
Fedik: Just drop it, `start/end` was need for old script, it makes no sees for `mark.js`, or I something missed?

layouts/joomla/highlight/highlight.php
dgrammatiko: and done, that was easy 😄

Fedik: It also about the options name :wink:

layouts/joomla/highlight/highlight.php
dgrammatiko: The `HTMLHelper::_('behavior.highlighter', ...)` still works thus the B/C part here. If people think that nuking it in J4 is better I can remove both the PHP and this part. I was just playing safe here

layouts/joomla/highlight/highlight.php
dgrammatiko: oops, coffee needed

dgrammatiko: Should be ok now

Fedik:
Now something more complicated :smile: 

**1)** We do not need layout `joomla.highlight.highlight`.

You should not define ALL possible options in the layout.
The JS script should have a default options, and merge with User options, in this way it will be more efficient.
So, example, if your code have `'iframesTimeout'=> 5000, 'class' => 'js-highlight'` then this should be in JS default, not PHP.
Then in PHP set only modified options:
```php
$doc->addScriptOptions('highlight', [
 'highLight' => $terms,
])
```
Developer should set them only if he want diferent `class`, `timeout`:
```php
$doc->addScriptOptions('highlight', [
 'highLight' => $terms,
 'class' => 'foo-bar',
])
```
In this way we save a loot of KB, and we do not need `joomla.highlight.highlight`.

**2)** For this script it better to allow to have a couple highlight on the page.

In `addScriptOptions` set array of options, instead of just options, 
in this way we can have a couple highlight, for different terms:
```php
$doc->addScriptOptions('highlight', [[ 'highLight' => $terms]], false);
$doc->addScriptOptions('highlight', [[ 'highLight' => $terms2]], false);
```
The JS could do some caching/checking to avoid the terms duplication, so we safe if someone have multiple call of same terms.

layouts/joomla/highlight/highlight.php
Fedik: I see, then keep it `HTMLHelper::_('behavior.highlighter' ...`
But it does not need to add them in JS options, it will be just a trash there :wink:

dgrammatiko:
> Now something more complicated

Should be ok now, fwiw I was copying fixing the old PR here without paying too much attention on my previous assumptions

dgrammatiko:
@infograf768 as this PR is just replacing the highlighting script can you make a PR with the colour change once this is merged?

Edit: nevermind I added it here

layouts/joomla/highlight/highlight.php
dgrammatiko: @wilsonge should I nuke the `HTMLHelper::_('behavior.highlighter', ...)` ?

Fedik:
I still would drop from php `'accuracy'   => 'partially', 'diacritics' => true,`
But it already better than before :wink: 

Later I will try to test

build/media_source/system/js/highlight.es6.js
```diff
@@ -0,0 +1,70 @@
...
+  ignoreJoiners: false,
+  wildcards: 'disabled',
+  compatibility: false,
+};
+
+if (Joomla.getOptions && typeof Joomla.getOptions === 'function' && Joomla.getOptions('highlight')) {
+  const scriptOptions = Joomla.getOptions('highlight');
+  scriptOptions.forEach((currentOpts) => {
+    const options = { ...defaultOptions, ...currentOpts };
+    console.log(options);
```
Fedik: I seee it :smile:

plugins/system/highlight/highlight.php
```diff
@@ -86,12 +86,24 @@ public function onAfterDispatch()
...
+					'highLight'  => $cleanTerms,
+				]]
+			);
+		}
 
 		// Adjust the component buffer.
 		$doc = $this->app->getDocument();
 		$buf = $doc->getBuffer('component');
-		$buf = '<br id="highlighter-start" />' . $buf . '<br id="highlighter-end" />';
+		$buf = '<div class="js-highlight" />' . $buf . '</div>';
```
ReLater:
```suggestion
		$buf = '<div class="js-highlight">' . $buf . '</div>';
```

components/com_finder/tmpl/search/default_results.php
```diff
@@ -48,11 +47,24 @@
 <?php endif; ?>
 <?php // Activate the highlighter if enabled. ?>
 <?php if (!empty($this->query->highlight) && $this->params->get('highlight_terms', 1)) : ?>
-	<?php HTMLHelper::_('behavior.highlighter', $this->query->highlight); ?>
+	<?php
+		if (!empty($terms))
+		{
+			$doc = Factory::getDocument();
```
Fedik: btw, in the component layout there already `document` injected, and available with `$this->document`:
```php
$this->document->getWebAssetManager();
```

plugins/system/highlight/highlight.php
dgrammatiko: thanks

build/media_source/system/js/highlight.es6.js
dgrammatiko: pfff

Fedik:
okay, I have tested, and it does not work :smile: 
well, it partly work, only for ?highlight=WyJjcmVhdGUiLCJhZG1pbiJd
but not work for search result view

components/com_finder/tmpl/search/default_results.php
```diff
@@ -48,11 +47,24 @@
 <?php endif; ?>
 <?php // Activate the highlighter if enabled. ?>
 <?php if (!empty($this->query->highlight) && $this->params->get('highlight_terms', 1)) : ?>
-	<?php HTMLHelper::_('behavior.highlighter', $this->query->highlight); ?>
+	<?php
+		if (!empty($terms))
```
Fedik: you mixed up with this `if()` section, `$terms` never exits, where you copied it from? :smile: 

Remove this `if()` and do:
```
$this->document->getWebAssetManager()->useScript('highlight');
$this->document->addScriptOptions(
  'highlight',
  [[
    'class'      => 'js-highlight',
    'highLight'  => $this->query->highlight,
  ]]
);
```

dgrammatiko: Done

Fedik:
I have tested this item :white_check_mark: successfully on 86aaaf7d1738ded8e5965c02acfd39ec2f7c5f6d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33089">issues.joomla.org/tracker/joomla-cms/33089</a>.</sub>

particthistle:
I have tested this item :white_check_mark: successfully on 86aaaf7d1738ded8e5965c02acfd39ec2f7c5f6d<br /><br />Created a highlight string to match content I had on my J4 installation and highlighting appeared as per expected result.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33089">issues.joomla.org/tracker/joomla-cms/33089</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33089">issues.joomla.org/tracker/joomla-cms/33089</a>.</sub>

PR #32611, opened 2021-03-08
[4.0] Migrate data for repeatable fields

joomdonation:
Pull Request for Issue #30556.

### Summary of Changes
This PR solves the Release Blocker issue #30556. It convert data of repeatable fields to the format accepted by subfield.

### Testing Instructions
1. Install Joomla 3.10.0. You can download the nightly build package here https://developer.joomla.org/nightlies/Joomla_3.10.0-alpha6-dev-Development-Full_Package.zip to install it.
2. Create one or two repeatable field for articles.
3. Create one or two articles, enter data for the created custom fields.
4. Upgrade your site to latest Joomla 4.0-dev package (generate for this PR). Just download the updated package here https://ci.joomla.org/artifacts/joomla/joomla-cms/4.0-dev/32611/downloads/41496/Joomla_4.0.0-beta8-dev+pr.32611-Development-Update_Package.zip , then go to **Components** -> **Joomla ! Update**, look **Upload & Update** tab and install that update package.
5. Check the articles on upgraded site, edit these article, check and make sure data in Fields tab is there.

### Actual result BEFORE applying this Pull Request
Data for repeatable fields not being migrated

### Expected result AFTER applying this Pull Request
Data for repeatable fields are migrated properly

PR #33094, opened 2021-04-11
[4.0] Fix media field value on Subfield

joomdonation:
Pull Request for Issue # .

### Summary of Changes
When data for a Repeatable field from Joomla 3 migrated to a SubField on Joomla 4, the value for media field type is stored in String, not in JSON format as expected, thus it throws some warnings when the field is being displayed (see https://github.com/joomla/joomla-cms/pull/32611#issuecomment-816985442 )

This PR solves that error by overriding method onCustomFieldsPrepareField for Media field to make sure $field->value is an array (for Media field not belong to a Subfield, that's prepared on onCustomFieldsBeforePrepareField already)

### Testing Instructions
This only affect Repetable field migrated to Subfield, so @HLeithner can take a look and if it solves the issue, he can get it merged. I don't want to waste time of tester by having to install Joomla 3.10.0, upgrade to Joomla 4, then test this small fix.

### DIFF
--- a/plugins/fields/media/media.php
+++ b/plugins/fields/media/media.php
@@ -67,6 +67,31 @@ class PlgFieldsMedia extends \Joomla\Component\Fields\Administrator\Plugin\Field
 	}
 
 	/**
+	 * Prepares the field
+	 *
+	 * @param   string    $context  The context.
+	 * @param   stdclass  $item     The item.
+	 * @param   stdclass  $field    The field.
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function onCustomFieldsPrepareField($context, $item, $field)
+	{
+		/**
+		 * At this stage, $field->value supposes to be an array, however, in case it is part of a subfield with data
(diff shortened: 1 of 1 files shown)


HLeithner:
Wouldn't it be better to convert it? Else we need this code till joomla 10.

joomdonation:
I'm unsure. The **onCustomFieldsPrepareField** trigger is available for us to use, so I think we can use it to handle this case, and it is simpler than converting the data. We already do similar check/conversion for media field in **onCustomFieldsBeforePrepareField**, so It is not hurt to do the same on onCustomFieldsPrepareField

@laoneo Could you give some suggestions here?

If you both agree that we should do data conversion, I will try to do so.

laoneo:
@joomdonation looks fine for me...

joomdonation:
Thanks @laoneo.  @HLeithner Are you OK with this approach? If not, I will try to find sometime to find a way to convert data to expected format in migration script.

laoneo:
Migration script would be much better tbh...

joomdonation:
OK. Thanks for the feedback. I will try to find sometime to do that with migration script, then.

Issue #32892, opened 2021-03-27
Joomla 4 website's backend not working after renaming htaccess.txt to .htaccess


Mohamed1308:
Hello.
I got a issue after renaming htaccess.txt to .htaccess in the public_html folder of my Joomla 4 website (https://www.freepetsitters.com). I renamed it in order to get rid of "index.php" in the URLs.
### Steps to reproduce the issue
1. Download Helix Ultimate 2 Beta 1 Quickstart Package (https://github.com/JoomShaper/helix-ultimate/releases/tag/v2.0.0-beta.1), upload it to the public_html folder of your website, unzip it and then follow the standard Joomla installation procedure.
2. Upgrade your website to Joomla 4 Beta 7.
3. Set "Use URL Rewriting" to "Yes" (In the Global Configuration tab of the Joomla administrator panel).
4. Rename htaccess.txt to .htaccess in the public_html folder of your website.
5. Delete the cookies and website data, then refresh the Joomla administrator panel's tab.
### Expected result
The website's backend looking like it was (before renaming htaccess.txt to .htaccess) and working fine.
### Actual result
The administrator panel looking weird and not working (There is no interactivity when clicking on any button).
### System information (as much as possible)
**LiteSpeed-based server**
PHP: 7.3 lsphp
Joomla: 4 Beta 7
Helix Ultimate: 2 Beta 1
### Additional comments
Here is what I got in the logs file:
`WARNING ...	assets	No asset found for com_modules.module.92, falling back to com_modules`

I also don't think that the problem is related to the website's template (Helix Ultimate 2 Beta 1) since the issue only occurs for the backend and not the frontend.

AMurray2016:
I've left a comment on the Joomla forum about this: https://forum.joomla.org/viewtopic.php?f=803&t=985622&p=3627757#p3627757.  In essence: Quick-starts are not the recommended way to install Joomla. 

For your testing, I would use the J4 nightly build, then install Helix Ultimate 2.0 Beta 1 through Extension Manager, and avoid the quick-start package altogether; Joomshaper should provide this Beta 1 release as a stand-alone template install package on their GITHUB site.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/32892">issues.joomla.org/tracker/joomla-cms/32892</a>.</sub>

Mohamed1308:
Hello @AMurray2016.
[...]

PR #33097, opened 2021-04-11
[4] ERR_CONTENT_DECODING_FAILED help

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/32892

### Summary of Changes

added help text to .htaccess to explain `ERR_CONTENT_DECODING_FAILED` so that users are not confused like https://github.com/joomla/joomla-cms/issues/32892

### Testing Instructions

None

Users confused and sites look broken if web hosts are gzipping at the server level, when a user enables .htaccess

Hopefully people will read :-) and understand. 

None

### DIFF
--- a/htaccess.txt
+++ b/htaccess.txt
@@ -127,6 +127,10 @@ Options -Indexes
 ## These directives are only enabled if the Apache mod_headers module is enabled.
 ## This section will check if a .gz file exists and if so will stream it
 ##     directly or fallback to gzip any asset on the fly
+## If your site starts to look strange after enabling this, and your see
+##     ERR_CONTENT_DECODING_FAILED in your browser console network tab,
+##     then your server is already gzipping js files and you dont need this 
+##     enabled in your .htaccess
 <IfModule mod_headers.c>
 	# Serve gzip compressed CSS files if they exist
 	# and the client accepts gzip.
--- a/htaccess.txt
+++ b/htaccess.txt
@@ -127,10 +127,10 @@ Options -Indexes
 ## These directives are only enabled if the Apache mod_headers module is enabled.
 ## This section will check if a .gz file exists and if so will stream it
 ##     directly or fallback to gzip any asset on the fly
-## If your site starts to look strange after enabling this, and your see
+## If your site starts to look strange after enabling this, and you see
 ##     ERR_CONTENT_DECODING_FAILED in your browser console network tab,
-##     then your server is already gzipping js files and you dont need this 
-##     enabled in your .htaccess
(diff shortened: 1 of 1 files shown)


htaccess.txt
```diff
@@ -127,6 +127,10 @@ Options -Indexes
 ## These directives are only enabled if the Apache mod_headers module is enabled.
 ## This section will check if a .gz file exists and if so will stream it
 ##     directly or fallback to gzip any asset on the fly
+## If your site starts to look strange after enabling this, and you see
+##     ERR_CONTENT_DECODING_FAILED in your browser console network tab,
+##     then your server is already gzipping js files and you don't need this
```
richard67:
```suggestion
##     then your server is already gzipping css and js files and you don't need this
```

richard67: Not sure though if to use "and" or "or" or "and/or", because the problem may happen when only one of these kinds of files is zipped, but that would also mean to disable only a part of that section, and the explanation could become too complicated then.

richard67: I'd think in most if not all cases they zip both kinds of files, so the "and" suggested above would be ok.

PhilETaylor: Ive seen web hosts blanket gzip everything, including the HTML Response, which means if you have GZIP enabled in Joomla Global Config then you get ERR_CONTENT_DECODING_FAILED when viewing the actual page too. Its always best to gzip at the server level than making PHP run `gz*` functions.

Fedik:
I have tested this item :white_check_mark: successfully on 6d891505a3d4e60060570e4a342d9889ba05529b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33097">issues.joomla.org/tracker/joomla-cms/33097</a>.</sub>

richard67:
I have tested this item :white_check_mark: successfully on 6d891505a3d4e60060570e4a342d9889ba05529b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33097">issues.joomla.org/tracker/joomla-cms/33097</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33097">issues.joomla.org/tracker/joomla-cms/33097</a>.</sub>

PR #31092, opened 2020-10-14
Added height control and v-scroll to com_admin sidebars

ceford:
Pull Request for Issue #30507.

### Summary of Changes

Added css to make com_admin index divs limited height vertical scroll
Added code to toggle index display on small screens

### Testing Instructions

Go to Admin Home Dashboard / Help / Joomla Help
Scroll to the bottom of the list - a long way - and select the last item
View in responsive mode
Repeat for the Global Configuration page

Apply the patch and then run an npm build

### Actual result BEFORE applying this Pull Request

Nothing appears to happen because the iframe to the right is out of sight.
In the small screen view the menu cannot be collapsed.

### Expected result AFTER applying this Pull Request

The index div is scrollable and the iframe is always visible - so a selection change is obvious
On small screens the index can be collapsed.

### Documentation Changes Required

None

Issue #31928, opened 2021-01-08
[4b6] UX on Global Configuration Menu, hides items with no visible indication


PhilETaylor:
### Steps to reproduce the issue

On a "normal sized" screen, when you visit Joomla Global Configuration, the sub menu on the left "looks" to be rendering at the height of the viewport only. 

This truncates the menu with **NO INDICATION that there are more menu items.** 

If you did not know Joomla, or did not look hard, you would miss that the lower menu items are not visible and cut off.

When you scroll the page, the size of the left menu doesnt change, so the items remain in a small block, with the overflow menu items hidden. 

You can hover and scroll (on a Mac) and then you get a scroll bar indication that there are "more" to look at below the virtual fold of the outlying block.

Easier to explain with a video:

https://user-images.githubusercontent.com/400092/104076217-52ac2d00-520d-11eb-96d3-fff945dca836.mp4

### Expected result

All menu items should be visible at all times
OR 
If you need to hide some, make some way to show they are hidden and can be scrolled to

### Actual result

Described above. 

### System information (as much as possible)

Safari on macOS

### Additional comments

The full menu is this: 

<img width="325" alt="Screenshot 2021-01-09 at 00 00 09" src="https://user-images.githubusercontent.com/400092/104076317-ac145c00-520d-11eb-8c19-6df923b3ce60.png">

infograf768:
Can't reproduce here Mac 10.12.6 (16G2136) Safari Version 12.1.2 (12607.3.10)
I always have a scroll bar.

brianteeman:
Not a bug. OSx insists on defaulting to not showing scrollbars until hover which is what your video shows.

@infograf768 has probably changed his system preferences
https://www.webnots.com/how-to-customize-scroll-bar-in-mac/

PhilETaylor:
Just because you say somethings not a bug doesn't mean it's not a bug

The bug here is the terrible user experience where content is hidden - not the lack of scroll bars

The lack of scroll bars by default just emphasises the problem greater making it a bigger bug

Content is hidden for no reason and a new person to joomla would have no idea it's there

Scrolling within a page element is new to most people, I can't think of any other app or website a "normal" person would come across that 

There is no reason for this block to autosize to the height of the view port is there? It could be just left open at full height to display the content

It is hiding for the sake of hiding

PhilETaylor:
[...]

PR #33108, opened 2021-04-11
[J4] Removes fixed max-height on sidebar-nav on wide screen

regularlabs:
In PR #31092 a fixed height was introduced to the sidebar-nav to fix things on mobile screens (smaller than md).
But this causes a useless scrollbar and height limit on the sidebar nav on wider screens.

So this PR removes it.

Check the styling of the sidebar in the global configuration before and after this PR.

Before:
image

After:
image

### DIFF
--- a/administrator/templates/atum/scss/blocks/_sidebar-nav.scss
+++ b/administrator/templates/atum/scss/blocks/_sidebar-nav.scss
@@ -1,10 +1,11 @@
 // Sidebar
 
 .sidebar-nav {
-
+  @include media-breakpoint-down(md) {
+    max-height: 75vh;
+    overflow-y: scroll;
+  }
   // override bootstrap important
-  max-height: 75vh;
-  overflow-y: scroll;
   background: var(--atum-sidebar-bg) !important;
   box-shadow: $atum-box-shadow;
 


regularlabs:
Also fixes this issue: #31928

brianteeman:
(it wasn't just for mobile screens @ceford do you want to explain what the problem was you were solving)

ceford:
31092 had nothing to do with mobile screens or screen width in general. It was about height. The list of Help screens in particular is very long - you had to scroll down a long way for the later ones, and when clicked on nothing appeared to happen because the help screen is loaded into a frame out of site at the top. Frankly, the Options list works better this way too. So I object to this PR. It is a step backwards and does not seem to solve any particular problem. The scrollbar is not useless! The lack of indication that there are more items further down is a more general problem that needs its own solution.

regularlabs:
Well, I think it is very weird to have the sidenav as an extra scrollable section. It would only make sense if that part was fixed to the same height of the browser window. But now, on many screen sizes, you have to scroll the main window down to see the bottom of that nav section... and then scroll that section down even more.

In my opinion the 'fix' applied in #31092 is a crude way to solve something that causes more issues and frustration than it is worth.

Anyway... my 2 cents. Do with it what you want...

regularlabs:
PRs that cause discussions are not worth my time. So closing this.

PR #33114, opened 2021-04-12
[4] s/ordern/order

PhilETaylor:
Correct spelling mistake introduced as part of a security fix

### DIFF
--- a/libraries/src/Form/Rule/FilePathRule.php
+++ b/libraries/src/Form/Rule/FilePathRule.php
@@ -49,7 +49,7 @@ class FilePathRule extends FormRule
 			return true;
 		}
 
-		// Make sure $value starts with an a-z/A-Z character in ordern to not allow to break out of the current path
+		// Make sure $value starts with an a-z/A-Z character in order to not allow to break out of the current path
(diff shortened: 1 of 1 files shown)


libraries/src/Form/Rule/FilePathRule.php
```diff
@@ -49,7 +49,7 @@ public function test(\SimpleXMLElement $element, $value, $group = null, Registry
 			return true;
 		}
 
-		// Make sure $value starts with an a-z/A-Z character in ordern to not allow to break out of the current path
+		// Make sure $value starts with an a-z/A-Z character in order to not allow to break out of the current path
```
brianteeman, line 52: Lets make it real english at the same time

```suggestion
		// Make sure $value begins with an a-z/A-Z character to prevent breaking out of the current path
```

PhilETaylor: Actually this PR should just be closed, because this has already been removed in Joomla 3 - as it was part of that dodgy security fix, and should eventually be merged from three into four when George finally gets around to merging up￼￼

brianteeman: fair enough - just close it then

PR #33130, opened 2021-04-14
[4.0] Fix tinymce template

brianteeman:
TinyMCE has a really powerful feature that almost knows about - the ability to have predefined snippets of text and html.

This is not a new feature of TinyMCE it has been present for at least 10 years. The reason no one knows about this or talks about it is that its very hard to use because of a bug in the path that Joomla uses to store the snippets.

I am only aware of a blog post by @coolcat-creations [blog post](https://brian.teeman.net/joomla/883-multicolumn-editing-for-joomla-authors)  explaining how to use it and you will see it is both very powerful but also a pain to setup

## The Problem
The folder that tinymce looks in for the stored "templates" is outside of the folder structure that can be accessed within the Joomla administrator ui. In Joomla 4 it is `media/vendor/tinymce/templates` so it can only be accessed by ftp or cli.

## The Solution
Change the folder where the "templates" are stored to a user defined folder in the site/templates folder structure. This enables the user to use the template manager to create and edit the "templates".

## Backwards Compatibility
No problem here as if a custom folder is not set then the default is the current `media/vendor/tinymce/templates`. The one difference is that the current (and therefore default) behaviour is for the files to be *.html. Because that filetype is by default prohibited in the template manager I have changed it to additionally support .txt filetype __only__ when using the custom folder/

### Bonus
A template developer can include predefined snippets of text and html with their template

## To Test
Firstly familiarise yourself with this TinyMCE feature. You will find it on the Insert dropdown of the editor toolbar. TinyMCE ships with two examples you can play with.

### Apply the PR

- Go to the Editor - TinyMCE plugin and there is a new field where you can select the folder to use.
- **Make sure you are in a configuration set that applies to your user level (it is probably Set 0) and then select a folder and save.**
- Go to an article and try to insert a template (as you did before)
- You should get a template not found message
- Go to the template manager and create two text files in the folder you previously selected

- - with just plain text
- - with basic html

- Now return to the article and you should be able to insert both these templates successfully.

~## Issues to solve? - need help~
~The list of folders presented to the user to select from was intended to be limited to templates\[template_name]\html\ and all its subfolders but despite my best efforts I was not able to get the folderFilter to includes those subfolders~

# Bonus!!!!
This is an incredibly marketable feature. (There aren't that many in J4 that the regular user will understand)
Its uses are endless not least the which the ability to build up a complex web page without the bloat and performance issues from a pagebuilder.

## Video
As I recorded the talk at joomladay usa which mentions this feature and how to use it I have created a 3 minute video to explain it.

[](http://www.youtube.com/watch?v=OyZQ63Ahnt4 "content template")


dgrammatiko:
@brianteeman you need the pieces for child templates from https://github.com/joomla/joomla-cms/pull/31785 as well here

brianteeman:
@dgrammatiko if that gets accepted we can look at that then.

Fedik:
Looks like a good idea

administrator/language/en-GB/plg_editors_tinymce.ini
```diff
@@ -23,6 +23,8 @@ PLG_TINY_FIELD_CSS_LABEL="Template CSS Classes"
 PLG_TINY_FIELD_CUSTOM_CSS_DESC="Optional CSS file that will override the standard editor.css file. Enter a file name to point to a file in the CSS folder of the default template (for example, templates/cassiopeia/css/). Or enter a full URL path to the custom CSS file. If you enter a value in this field, this file will be used instead of the editor.css file."
 PLG_TINY_FIELD_CUSTOM_CSS_LABEL="Custom CSS Classes"
 PLG_TINY_FIELD_CUSTOM_PATH_LABEL="Images Directory"
+PLG_TINY_FIELD_CUSTOM_CONTENT_TEMPLATE_PATH_DESC="The directory that TinyMCE will look in for any custom templates or snippets."
```
Quy: Please alpha sort.

plugins/editors/tinymce/tinymce.php
```diff
@@ -470,31 +470,66 @@ public function onDisplay(
...
-					$description = ' ';
+					$filename = basename($filename, '.txt');
 
-					if ($lang->hasKey('PLG_TINY_TEMPLATE_' . strtoupper($filename) . '_TITLE'))
+					if ($filename !== 'index')
 					{
-						$title = Text::_('PLG_TINY_TEMPLATE_' . strtoupper($filename) . '_TITLE');
+						$lang        = Factory::getLanguage();
+						$title       = $filename;
+						$description = ' ';
```
Quy: Is the space intentional?

Fedik:
the `if/else` block can be eliminated :

```php
$templates_dir = $levelParams->get('content_template_path');
$templates_dir = $templates_dir ? '/templates/' . $templates_dir : '/media/vendor/tinymce/templates';

foreach (glob(JPATH_ROOT . $templates_dir . '/*.{html,txt}', GLOB_BRACE) as $filename)
{
  ...
}
```

> I have changed it to the supported .txt filetype

What about `.tmpl`? some editors have a code highlight for this file type. With `.txt` can be hard. 
but this need to test

plugins/editors/tinymce/tinymce.php
brianteeman: I assume so - this code block is just a partial duplicate of the original code

brianteeman:
Thanks @fedik I will test that code

I chose txt because its non-executable and the template manager already supports txt files

brianteeman:
@fedik just looking at the if else. How would you deal with
`$filename = basename($filename, '.txt');`
`'url' => Uri::root(true) . $content_template . $filename . '.txt',`

Fedik:
I think with `pathinfo` https://www.php.net/manual/en/function.pathinfo.php

Maybe something like:
```php
// Use $name for Template name, and translation constant. Keep original $filename unchanged
$name = pathinfo($filename, PATHINFO_FILENAME); 

 // Use original $filename
'url' => Uri::root(true) . $content_template . $filename;
```

Fedik:
Hm, can stay `basename()` with same logic:
```php
// Use $name for Template name, and translation constant. Keep original $filename unchanged
$name = basename($filename); 

 // Use original $filename
'url' => Uri::root(true) . $content_template . $filename;
```

brianteeman:
@fedik I am reluctant to make these changes. The current code works - it can always be refactored at a later date.

Fedik:
hm, I have tested a bit and I noticed that cannot select any other folder than `cassiopeia/html`. 
It does not looks right.

> I am reluctant to make these changes

I have made PR with that changes

Fedik:
> The list of folders presented to the user to select from was intended to be limited to templates[template_name]\html\ and all its subfolders but despite my best efforts I was not able to get the folderFilter to includes those subfolders

okay, the problem here is that `folderFilter="^html"` looking for single folder name, not for full path
https://github.com/joomla/joomla-cms/blob/f2c275006b3e99256a0c31fab5382f1a0aa9ce2f/libraries/src/Filesystem/Folder.php#L622-L624

that means it will work for:
```
html/
html_foo/
htmlbla/
```
but not for:
```
html/foo/
html/bla/
```

that complicated :)

Fedik:
As idea, maybe instead of `folderFilter="^html"` make `folderFilter="^tinymce"`,
and create an empty folder in `cassiopeia/html/tinymce`

Can be need update description/documentation, so User know where to look.

brianteeman:
Thanks for your suggestions they sound good (as does the patch you contributed) I will check this tomorrow as it goes a bit further than I had intended originally

brianteeman:
@fedik Regarding the path/filter changes
with the option `recursive=true` and the folderFilter I had assumed that they would work together for the desired result but clearly they don't

brianteeman:
Resolved conflicts.

Would be great to get some tests on this. It got really good feedback at joomladayusa

plugins/editors/tinymce/forms/setoptions.xml
```diff
@@ -61,14 +61,27 @@
...
-			<field
+      <field
 				name="path"
 				type="uploaddirs"
 				label="PLG_TINY_FIELD_CUSTOM_PATH_LABEL"
 				showon="drag_drop:1"
 			/>
 
-			<field
+      <field
```
Quy: Please fix indentation.

Fedik:
I wanted to test
But found that the man with name @brianteeman broke this PR with #33177 :smile:

Fedik:
@brianteeman please check the conflict, 
the options is fixed in #33464, but it made a conflict with your changes

brianteeman:
the conflict looked fixed to me locally so I am at a loss as to why it says conflicts now. Any help appreciated

Fedik:
it an indentations,
try like next: copy setoptions.xml from 4.0-dev branch, and re-add "content_template_path" field.
this will be easiest way

Fedik:
I have tested this item :white_check_mark: successfully on 7652e3105b2976a8da7a7e862959641854ff8989<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33130">issues.joomla.org/tracker/joomla-cms/33130</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 7652e3105b2976a8da7a7e862959641854ff8989<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33130">issues.joomla.org/tracker/joomla-cms/33130</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33130">issues.joomla.org/tracker/joomla-cms/33130</a>.</sub>

brianteeman:
yay - thank you all!

brianteeman:
care to be more constructive and informative

PhilETaylor:
Care to be more patient while I actually type?? I was copy and pasting. 

PR to revert is here #33804 and the issue is  #33783

PR #33131, opened 2021-04-14
Merge two if blocks to reduce code

eopws:
### Summary of Changes

Merged two if blocks with the same condition

### Testing Instructions

Go to libraries/src/Document/HtmlDocument.php, see lines 707 and 712, if blocks on these lines have the same conditions and can be merged in order to reduce code.

### DIFF
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -707,10 +707,6 @@ class HtmlDocument extends Document
 		if (!file_exists($directory . '/' . $template . '/' . $file))
 		{
 			$template = 'system';
-		}
-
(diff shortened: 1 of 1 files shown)


richard67:
@eopws Sorry but this pull request is wrong. You are right, the "if" conditions are identical. But inside the first "if" block the "$template" variable is modified, and this is checked again in the "2nd" if condition. And so this sequence of if conditions has to be like it is. In general it is not enough to read just 1 or 2 lines of code, one should also read what is before and after these 2 lines of code and is relevant.

eopws:
Sorry, I didn't notice that. I have to be more careful next time.

Issue #33063, opened 2021-04-07
[4] media manager destroys images when cropping


PhilETaylor:
### Steps to reproduce the issue

I was convinced this was just me - but can you replicate it?

Go to Media Manager in Joomla 4 admin 

Click to Edit the joomla_black.png file

When on the CROP tab, just keep pressing save, save save save until the 4 fields show NaN

Takes 3-6 clicks. 

Then you have destroyed the image. and if you have error reporting to maximum you get in your Ajax from the save, and on the next refresh of the page 

```
Notice: exif_imagetype(): Error reading from /application/images/joomla_black.png! in /application/libraries/src/Helper/MediaHelper.php on line 74

Notice: getimagesize(): Error reading from /application/images/joomla_black.png! in /application/libraries/src/Image/Image.php on line 185
```

Is it me?

himanshu007-creator:
@PhilETaylor , ✅ confirmed same behaviour ( chrome on windows )

![proof](https://user-images.githubusercontent.com/65963997/114066518-70933780-98b9-11eb-8c52-bfb50d85cef6.gif)

PhilETaylor:
The more I play with the crop/resze/rotate the more times I notice NaN javascript errors in the text fields.

Magnytu2:
@PhilETaylor This is an error that has been pointed out several times from memory. #32326

PhilETaylor:
> @PhilETaylor This is an error that has been pointed out several times from memory. #32326

And yet still unresolved :)

Magnytu2:
@PhilETaylor  Yes and as soon as you use the media manager with more than 12 images it's the same. The media manager is not solid. The display of fields over the entire width is too large. And I find that its use from an article is even more fragile and I very often find myself without illustrations.

ceford:
After the joomla_black.png file gets corrupted the Media screen stops working completely. Three clicks to kill a major component of Joomla 4! This needs to be marked a **release blocker.** Ordinary users will not know they have to deleted the corrupted file using something other than com_media to get it to work again.

PhilETaylor:
> After the joomla_black.png file gets corrupted the Media screen stops working completely. 

I tried to fix that and it was rejected #33141

PhilETaylor:
IIRC its only broken if you have error reporting set to anything other than none.

brianteeman:
I am unable to replicate this now. There have been a lot of changes in the js since this issue was opened - probably solved there

Issue #33062, opened 2021-04-07
[4] media manager cannot cope with broken images


PhilETaylor:
### Steps to reproduce the issue

https://user-images.githubusercontent.com/400092/113942022-17170400-97f8-11eb-88a3-f4164ed1eccc.png

Download this broken image and place in your /images folder as broken.png

Set error reporting to maximum in Joomla global config 

Visit media manager in the backend.

### Expected result

Media Manager works

### Actual result

Ajax call has warning:

```
<br/>
<b>Notice</b>
:  exif_imagetype(): Error reading from /application/images/broken.png! in 
<b>/application/libraries/src/Helper/MediaHelper.php</b>
 on line 
<b>74</b>
<br/>
<br/>
<b>Notice</b>
:  getimagesize(): Error reading from /application/images/broken.png! in 
<b>/application/libraries/src/Image/Image.php</b>
 on line 
<b>185</b>
<br/>
```

ghost:
Confirm that after upload of broken.png media manager don't work. I don't know how to see a ajax call but in console got:

![grafik](https://user-images.githubusercontent.com/78906980/114133929-2173f380-9907-11eb-9d70-b8ee78a78d26.png)

After removing broken.ong media manager works again, so i guess the above message is caused by broken.png.

PhilETaylor:
@richard67 please reopen. Seems I no longer have rights to reopen my own closed issues.

PhilETaylor:
My two attempts at fixing this have been rejected by the maintainers. Therefore I bow out and someone else can fix it. Im not wasting anymore time on it.

richard67:
I haven’t rejected anything.

PhilETaylor:
Apologies for the plural use of the word "maintainers" :) ...

PR #33141, opened 2021-04-15
[4] Suppress errors on broken images

PhilETaylor:
Pull Request for Issue #33062

### Summary of Changes

**After considering all the alternatives, reluctantly** use the `@` error suppressor to avoid outputting PHP Notices when you have any broken (unparseable) images in your folder.

yes yes yes I know use of the  `@` error suppressor is highly frowned upon, but there honestly seems to be no alternative here. 

### Testing Instructions

https://user-images.githubusercontent.com/400092/113942022-17170400-97f8-11eb-88a3-f4164ed1eccc.png

Download this broken image (which was actually [generated by Joomla's broken media manager feature](https://github.com/joomla/joomla-cms/issues/33063)s) and place in your /images folder as broken.png

Set error reporting to maximum in Joomla global config

Visit media manager in the backend.

Ajax call has 2 warnings:

```
<br/>
<b>Notice</b>
:  exif_imagetype(): Error reading from /application/images/broken.png! in 
<b>/application/libraries/src/Helper/MediaHelper.php</b>
 on line 
<b>74</b>
<br/>
<br/>
<b>Notice</b>
:  getimagesize(): Error reading from /application/images/broken.png! in 
<b>/application/libraries/src/Image/Image.php</b>
 on line 
<b>185</b>
<br/>
```

Media manager loads correctly. 

None.

### DIFF
--- a/libraries/src/Helper/MediaHelper.php
+++ b/libraries/src/Helper/MediaHelper.php
@@ -71,7 +71,8 @@ class MediaHelper
 		{
 			if ($isImage && \function_exists('exif_imagetype'))
 			{
-				$mime = image_type_to_mime_type(exif_imagetype($file));
+				// Must use @ Error Suppressor here to avoid "Notice: exif_imagetype(): Error reading from" message on invalid image.
+				$mime = image_type_to_mime_type(@exif_imagetype($file));
 			}
 			elseif ($isImage && \function_exists('getimagesize'))
 			{
--- a/libraries/src/Image/Image.php
+++ b/libraries/src/Image/Image.php
@@ -181,8 +181,12 @@ class Image
 			throw new \InvalidArgumentException('The image file does not exist.');
 		}
 
-		// Get the image file information.
-		$info = getimagesize($path);
+		//
+		/**
+		 * Get the image file information.
+		 * Must use @ Error Suppressor here to avoid "Notice: getimagesize(): Error reading from" message on invalid image.
+		 */
+		$info = @getimagesize($path);
 
 		if (!$info)
 		{
--- a/libraries/src/Image/Image.php
+++ b/libraries/src/Image/Image.php
@@ -181,7 +181,6 @@ class Image
 			throw new \InvalidArgumentException('The image file does not exist.');
 		}
 
-		//
 		/**
 		 * Get the image file information.
 		 * Must use @ Error Suppressor here to avoid "Notice: getimagesize(): Error reading from" message on invalid image.


toivo:
I have tested this item :white_check_mark: successfully on 3740f9851b8b49fdbc2c9f9fbe38469a145492b2<br /><br />Tested successfully in Beta8-dev of 17 April in Wampserver 3.2.4 using PHp 8.0.2.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33141">issues.joomla.org/tracker/joomla-cms/33141</a>.</sub>

particthistle:
I have tested this item :white_check_mark: successfully on 3740f9851b8b49fdbc2c9f9fbe38469a145492b2<br /><br />Before: Media Manager loads a white screen due to the error.

After: Media Manager loads correctly.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33141">issues.joomla.org/tracker/joomla-cms/33141</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33141">issues.joomla.org/tracker/joomla-cms/33141</a>.</sub>

richard67:
Not a nice fix maybe, but I'd say for now better than no fix.

richard67:
Maybe @wilsonge or @rdeutz should have a final look on it.

PhilETaylor:
Just merge this already - at the end of the day there are some PHP functions that still output notices and warnings instead of throwing exceptions, that's life and we work around it in many places. 

Other than using different functions to validate if the image is valid, before calling these methods, there is no way to suppress these errors.

richard67:
You did comment it well, so for me it's much better than what we have elsewhere uncommented.

But I'll not merge since I was the one who has set RTC. Always good to have a final instance.

HLeithner:
I hate using @ to suppress warnings. Can we be sure that the function exists when called? I mean we get a fatal error with no error if I'm not wrong.

Wouldn't it be better to disable only the "expected" error type (E_NOTICE or E_WARNING) depending on the function call?

PhilETaylor:
/facepalm this has NOTHING to do with if the function exists or not!

PhilETaylor:
> I hate using @ to suppress warnings.  

We all do. 328 times Joomla 4 does it with 80 of those being in the restore.php used by upgrades.  At this least this PR has a reason to use it unlike the amazing overuse of `@` in https://github.com/joomla/joomla-cms/blob/2c302641aaad823ac91421a0b11f14e3f8e2fd27/components/com_finder/src/Helper/RouteHelper.php#L115

PhilETaylor:
> Wouldn't it be better to disable only the "expected" error type (E_NOTICE or E_WARNING) depending on the function call?

What you are saying is to mess with error_reporting value, before and after making these function calls... so adding up to 5-10 lines of code per function, when the @ suppressor will do all that in a single char.

That would achieve the same thing, in more code, with more complexity.

richard67:
>. .. the amazing overuse of `@` in
> 
> https://github.com/joomla/joomla-cms/blob/2c302641aaad823ac91421a0b11f14e3f8e2fd27/components/com_finder/src/Helper/RouteHelper.php#L115

Maybe we should change the `return null;` at the end to `@return null;` ... just to be safe 😄

PhilETaylor:
Maybe something like this which we use for deprecated errors in libraries/bootstrap.php might work for Notice/Warnings - no idea - not tested - looks like a nicerway to handle warnings and errors and would allow us to remove all `@` suppression in Joomla 4?

`
set_error_handler(['Joomla\CMS\Exception\ExceptionHandler', 'handleUserDeprecatedErrors'], E_USER_DEPRECATED);
`

HLeithner:
> > Can we be sure that the function exists when called?
> 
> /facepalm this has NOTHING to do with if the function exists or not!

maybe you could be less insulting?

The non existing function thing was fixed by php with php 7.0 because fatal errors are exceptions starting with this version. So no problem at least for non existing functions since joomla 4 only support php 7.2.5+

> We all do. 328 times Joomla 4 does it with 80 of those being in the restore.php used by upgrades. At this least this PR has a reason to use it unlike the amazing overuse of `@` in

Only because we use it doesn't mean it's the right thing to do and I'm not responsible what other people do.
The @ operator is a performance nightmare that's one of the important reason not to use it if possible.

Cour using the Joomla Error handler could be a good way to suppress notices/warnings for broken functions like `exif_imagetype` and (without checking the code) it should be easy to backtrace the notice to the function call and if it's `exif_imagetype` for example only write it to the log or show it if debugging is active or something smarter.

PhilETaylor:
> The non existing function thing was fixed by php with php 7.0 because fatal errors are exceptions starting with this version. So no problem at least for non existing functions since joomla 4 only support php 7.2.5+

You continue to make no sense at all. This PR is nothing to do with "non existing functions" it is about documented notices and warnings generated by calling existing functions. 

I think you actually mean this, which is actually from 8.0.0: 

> Warning: Prior to PHP 8.0.0, it was possible for the @ operator to disable critical errors that will terminate script execution. For example, prepending @ to a call of a function which did not exist, by being unavailable or mistyped, would cause the script to terminate with no indication as to why.

**Which, still, has, nothing, to, do, with, this PR!** 

> Only because we use it doesn't mean it's the right thing to do and I'm not responsible what other people do.

Im sorry that you do not believe you are not responsible - but that is literally the job of a maintainer - being responsible for the code other people write that then gets merged. 

If you are not responsible then why are you blocking this PR?.... 

> The @ operator is a performance nightmare

But yet we have 328 instances.  

> be a good way to suppress notices/warnings for broken functions

The functions are not broken. The functions are doing exactly as documented.

HLeithner:
> > The non existing function thing was fixed by php with php 7.0 because fatal errors are exceptions starting with this version. So no problem at least for non existing functions since joomla 4 only support php 7.2.5+
> 
> You continue to make no sense at all. This PR is nothing to do with "non existing functions" it is about documented notices and warnings generated by calling existing functions.
> 
> I think you actually mean this, which is actually from 8.0.0:
> 
> > Warning: Prior to PHP 8.0.0, it was possible for the @ operator to disable critical errors that will terminate script execution. For example, prepending @ to a call of a function which did not exist, by being unavailable or mistyped, would cause the script to terminate with no indication as to why.
> 
> **Which, still, has, nothing, to, do, with, this PR!**
> 
> > Only because we use it doesn't mean it's the right thing to do and I'm not responsible what other people do.
> 
> Im sorry that you do not believe you are not responsible - but that is literally the job of a maintainer - being responsible for the code other people write that then gets merged.
> 
> If you are not responsible then why are you blocking this PR?....
> 
> > The @ operator is a performance nightmare
> 
> But yet we have 328 instances.
> 
> > be a good way to suppress notices/warnings for broken functions
> 
> The functions are not broken. The functions are doing exactly as documented.


Ok I try it again maybe it's missing in translation...

The non existing function had been suppressed by the @ in versions before PHP 7.0 since 7.0 it's an throwable exception and will not be suppressed except you catch it. What I said with that is that you are right it has nothing to do with this PR:

You make me responsible for code that had been merged by other a long time ago at least I can't remember to merged anything with an @ suppression and no I'm not responsible to rewrite joomla to not use this. No idea why you think that this is my job.

And exactly the 328 instanced should be reconsidered but I except the fact that we don't have the man/women power to do this or at least is not a high priority. But not introducing new once is a start in my opinion.

And no I don't block this PR I just wanted to talk about it, better to solve it with this PR then throwing a notice/warning in production (or development where it maybe unrelated).

The function is broken (even if it's documented) because it throws a notice instead of returning a proper error code or throw an exception or do the same as json_last_error if there is no better way to tell the developer what happens.

PR #33150, opened 2021-04-15
[4] unpublish stats plugin on selecting once or never

PhilETaylor:

Joomla 4 asks for stats collection after installation.

A user can answer with a click on "NEVER" or "once" button

However this just sets a flag in the params of the plugin, it doesn't actually unpublish the plugin

### Expected result

The plugin is then disabled on clicking NEVER/ONCE, so that it doesn't take up any processing time/power/cpu/ms on page load ever again to ensure Joomla remains running a peak performance but the user has the option to reenable the plugin at any time

### To test:
install joomla 4
login to admin and ignore the stats message
navigate the plugins screen and search for stats
you can now see the stats message and the status of the stats plugin
click NEVER or once, the page reloads and the message has gone away and you can see instantly that the plugin has been disabled also 

### Actual result

The PlgSystemStats plugin loads on every page load, runs some checks, takes some time, and in the end doesn't even send stats because we have said never. 

There are 7 different `if` statements run in onAfterInitialise/onAfterDispatch, several of these have 2 conditions to check, objects to retrieve etc...

Retrieving the params from the database is an extra db query that's not needed if NEVER is selected.

### DIFF
--- a/plugins/system/stats/stats.php
+++ b/plugins/system/stats/stats.php
@@ -199,6 +199,11 @@ class PlgSystemStats extends CMSPlugin
 			throw new RuntimeException('Unable to save plugin settings', 500);
 		}
 
+		if (!$this->unpublishPlugin())
+		{
+			throw new RuntimeException('Unable to unpublish stats plugin', 500);
+		}
+
 		echo json_encode(['sent' => 0]);
 	}
 
@@ -210,7 +215,7 @@ class PlgSystemStats extends CMSPlugin
 	 * @since   3.5
 	 *
 	 * @throws  Exception         If user is not allowed.
-	 * @throws  RuntimeException  If there is an error saving the params or sending the data.
+	 * @throws  RuntimeException  If there is an error saving the params, unpublishing the plugin or sending the data.
 	 */
 	public function onAjaxSendOnce()
 	{
@@ -228,6 +233,11 @@ class PlgSystemStats extends CMSPlugin
 
 		$this->sendStats();
 
+		if (!$this->unpublishPlugin())
+		{
+			throw new RuntimeException('Unable to unpublish stats plugin', 500);
+		}
+
 		echo json_encode(['sent' => 1]);
 	}
 
@@ -240,7 +250,7 @@ class PlgSystemStats extends CMSPlugin
 	 * @since   3.5
 	 *
 	 * @throws  Exception         If user is not allowed.
-	 * @throws  RuntimeException  If there is an error saving the params or sending the data.
+	 * @throws  RuntimeException  If there is an error saving the params, unpublishing the plugin or sending the data.
 	 */
 	public function onAjaxSendStats()
 	{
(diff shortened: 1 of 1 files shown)


dgrammatiko:
@PhilETaylor this is a nice one. Could we abstract the unpublish logic to a trait/helper/something.else? The reason I'm asking this is that the same exact pattern should be applied to the sampledata plugins (once the data is successfully installed the ~~plugin~~ module should be unpublished/disabled automatically, it throws an error message if you try to run it again)

PhilETaylor:
Cant really be abstracted much further, but the concept to speed up Joomla could be used for sure.

richard67:
> The reason I'm asking this is that the same exact pattern should be applied to the sampledata plugins (once the data is successfully installed the ~plugin~ module should be unpublished/disabled automatically, it throws an error message if you try to run it again)

@dgrammatiko That's not 100% true. The Blog Sample Data can be installed multiple times when having multiple languages in backend, one time for each language. So at least that one should not be unpublished or disabled after installation.

@PhilETaylor Sorry for that off-topic comment, but I had to correct the above statement.

PhilETaylor:
> @dgrammatiko That's not 100% true. The Blog Sample Data can be installed multiple times when having multiple languages in backend, one time for each language. So at least that one should not be unpublished or disabled after installation.

You could just republish the module/plugins and run them again :)

dgrammatiko:
> The Blog Sample Data can be installed multiple times when having multiple languages in backend

If you need to run multiple times (one per language) the sampledata then this is a huge UX fail. The plugin should check the installed languages and do that automatically...

PhilETaylor:
hold the comments :) I have a PR about to push and you can chat there, instead of off railing this PR :) :)

dgrammatiko:
> instead of off railing this PR

oops sorry

PhilETaylor:
Please see: https://github.com/joomla/joomla-cms/pull/33165

PhilETaylor:
For those reading, this PR is still on track and should be tested as normal please :)

Quy:
I have tested this item :white_check_mark: successfully on 81893f401d6f61cd59e0601e2a676dc62082a47e<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33150">issues.joomla.org/tracker/joomla-cms/33150</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 81893f401d6f61cd59e0601e2a676dc62082a47e<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33150">issues.joomla.org/tracker/joomla-cms/33150</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33150">issues.joomla.org/tracker/joomla-cms/33150</a>.</sub>

PhilETaylor:
Thanks for testing

PR #33151, opened 2021-04-15
Folder or file could start with dot

n3t:
Files and folders starting with dot didn't pass the regexpr test. So for example \var\www\.tmp path didn't pass the test, which causes error during Joomla update.

### Summary of Changes
Modified Reg Expressions to include files and / or folders starting with dot. The check was brought by this PR https://github.com/joomla/joomla-cms/pull/32076 which adds InputFilter to temp path settings, however ignoring folders with dot at the beginning.

### Testing Instructions
Go to Joomla settings and change your Temp path to some folder with dot at the beginning, for example \path_to_joomla\.tmp (and create that folder of course). Go to Joomla update, and try to install the update.

Error is displayed, update is not installed.

No errror, update is installed.

None

### DIFF
--- a/libraries/vendor/joomla/filter/src/InputFilter.php
+++ b/libraries/vendor/joomla/filter/src/InputFilter.php
@@ -1016,14 +1016,14 @@ class InputFilter
 	 */
 	private function cleanPath($source)
 	{
-		$linuxPattern = '/^[A-Za-z0-9_\/-]+[A-Za-z0-9_\.-]*([\\\\\/]+[A-Za-z0-9_-]+[A-Za-z0-9_\.-]*)*$/';
+		$linuxPattern = '/^[A-Za-z0-9_\/-]+[A-Za-z0-9_\.-]*([\\\\\/]+[A-Za-z0-9_\.-]+)*$/';
 
 		if (preg_match($linuxPattern, $source))
 		{
 			return preg_replace('~/+~', '/', $source);
 		}
 
-		$windowsPattern = '/^([A-Za-z]:(\\\\|\/))?[A-Za-z0-9_-]+[A-Za-z0-9_\.-]*((\\\\|\/)+[A-Za-z0-9_-]+[A-Za-z0-9_\.-]*)*$/';
+		$windowsPattern = '/^([A-Za-z]:(\\\\|\/))?[A-Za-z0-9_-]+[A-Za-z0-9_\.-]*((\\\\|\/)+[A-Za-z0-9_\.-]+)*$/';
 
 		if (preg_match($windowsPattern, $source))
 		{


brianteeman:
>Files and folders starting with dot didn't pass the regexpr test.

thats good as they are not supposed to

PhilETaylor:
Paging @joomla/security as this was a security fix that is involved here @zero-24 #32076 

This should not be merged until reviewed by the JSST.

n3t:
@brianteeman There is no reason that those files and folders should be ommited. Especially folders. For example ".tmp" makes that folder hidden on Linux, not visible for example through FTP access.

PhilETaylor:
>  not visible for example through FTP access.

LMAO!!! Yes. They. Are. :-) 

Screenshot 2021-04-15 at 22 01 35

n3t:
@PhilETaylor I know you can switch on to show it, but if you just want to keep your view clean, you can easily hide it with dot.

PhilETaylor:
Also this repo is not the place for changes to `libraries/vendor/joomla/filter/src/InputFilter.php` 

Those should be made in the framework repo at https://github.com/joomla-framework/input

This PR should be closed, and if you really want to propose it, and have it commented on by the @joomla/security team, then https://github.com/joomla-framework/input is the correct place.

PhilETaylor:
or maybe you meant to update this file? `libraries/src/Filter/InputFilter.php`

n3t:
`libraries/src/Filter/InputFilter.php` only calll parent::clean function. To modify it would mean to copy all the code from `libraries/vendor/joomla/filter/src/InputFilter.php`...

How to get comment from @joomla/security team? Here, or need to be contacted other way?

PhilETaylor:
https://developer.joomla.org/security/contact-the-team.html

n3t:
Ok, thanks. The PR should be propably made to https://github.com/joomla-framework/filter package, or as you proposed, to 'libraries/src/Filter/InputFilter.php', or maybe directly to Model of Joomla Update component. I contacted security team to review this and give advice.

Generally this is related to issue you mentioned here https://github.com/joomla/joomla-cms/issues/32567 but on other place of Joomla.

richard67:
@n3t Your pull request is wrong for following reasons:
1. Your regular expressions are wrong. They will, in opposite to those in current Joomla 3 code, allow paths like e.g. `/some/folder/../../../evil.php` on Linux, which means breaking out of the Joomla root. This open a security hole and so cannot be accepted. You can easily very that with one of the many available sites for online execution of preg_match. Just Google for it.
2. Any change has to be made in the https://github.com/joomla-framework/filter repository .

I suggest you open an issue with a feature request in the https://github.com/joomla-framework/filter repository. The feature request should clearly describe your requirement (to be able to use folder and file names starting with a dot).

And please don't contact anymore the SST for discussing this PR via their email for reporting security issues. The form which you have used is ONLY for reporting security issues. @PhilETaylor had already notified the right person in his comment here https://github.com/joomla/joomla-cms/pull/33151#issuecomment-820724277 .

n3t:
@richard67 Thanks for your review, sorry for using contact form, I did it just because asked to. I will raise an issue on framework. Anyway, checking double dots in tmp folder set in configuration.php doesn't bring any real security in Joomla update process, as I can easily go to configuration, and set directly `/other/folder/evil.php`, which will pass the test as correct path...

richard67:
@n3t We don't protect a super user from himself (or herself). If you enter such an absolute path in backend you should know what you are doing. There are other ways to manipulate paths than changing in backend.

PR #33156, opened 2021-04-15
[4] Move class constants from DateTime to DateTimeInterface

PhilETaylor:
### Summary of Changes

In PHP 7.2 the class constants from `DateTime` were moved to `DateTimeInterface`

https://www.php.net/manual/en/class.datetimeinterface.php

Code review.


wilsonge:
LGTM. Thanks!

Issue #32878, opened 2021-03-26
 Restructure plugins for sample data


chmst:
### Is your feature request related to a problem? Please describe.
Sample data plugins were made in J3 but now we have much changes in views / database / features.

Blog sample data have been updated, but this is not a valid 
Now in J4 these plugins (blog sample data, multilanguage, testing) are not up to date and very hard to maintain.
Hard to maintain 

- manily because data and control structure are not separated properly
- Texts are not all matching features and naming in J4
- Coding standads have changed
- and others

### Describe the solution you'd like
A new concept is needed für plugins 
Data must be checked and updated, for **all views** of the whole system
Restructuring should also contain the installation process - the menuItem "home", which then is in relation to mutilanguage data.

A good example of how plugins could be made is here https://github.com/joomla/joomla-cms/discussions/32744#discussioncomment-533950

ghost:
> Now in J4 these plugins (blog sample data, multilanguage, testing)

@chmst can i please ask you where i can get this `testing sample data`? 
I installed J4 using launch.joomla.org having blog sample data, multilanguage but no testing, also don't get `testing` by `Extensions: Discover`.

infograf768:
@Bakual

chmst:
@sandramay0905 if you clone the 4.0 dev branch of Joomla-cms, you get the plugin for testing sample data.

ghost:
> if you clone the 4.0 dev branch 

thanks for the answer, no idea what it mean. I was only looking if i haven't found a plugin.

Bakual:
Texts should actually be simple to adjust. Those are regular language strings.
Coding standards should also be simple.

The control/data thing is actually hard to do in a plugin. Back then I thought about defining the data in an XML file and then read by the plugin. But since some parts have to be defined dynamically (eg the category id), static files don't work well.

As for copying files like suggested in the other thread. Imho it's a bit pointles, but could of course be done easily.

PR #20749, opened 2018-06-14
[4.0] Progressive Disclosure for the multilingual sampledata

Bakual:
As promised in https://github.com/joomla/joomla-cms/pull/20711#issuecomment-397023701

As I wrote in the referenced PR, I'm not convinced this is actually needed, but I wrote it as promised :smile: 
#### Pros:
* Sampledata installation is only proposed if it actually works.
* We save a bit of space.
* PROGRESSIVE DISCLOSURE (buzzword are always good)
#### Cons:
* Unlike other places (eg "showon" feature or toolbar buttons), the progressive disclosure doesn't work as nicely here since the user doesn't see the immediate change to his action (installing a language is a complete other view).
* The sampledata currently only features one sampledata, it would be nice to have a second one so people actually see it is supposed to be a list :smile: Also the space used by that second option isn't really an issue here.
* The user is aware that this feature exists to help set up a multilingual site. I can see the arguments for both ways.

I don't have any hard feelings either way. So you guys test and decide if it's merged or not. And please only comment after having at least tried to understand the other side (should be natural)

### Summary of Changes
Only shows the multilingual sample data option on sites that have more than one language installed.

### Testing Instructions
* Make sure the multilingual sampledata plugin is installed (discover it otherwise first) and enabled
* Check with one language installed -> Multilingual Sampledata doesn't appear in the cPanel Sample Data module
* Check with more than one language installed -> Multilingual Sampledata option appears in the module.

### Expected result
Option only shows when more than one language is installed

### Actual result
Option is always there as long as plugin is enabled.

### Documentation Changes Required
None

PR #33165, opened 2021-04-16
[4] When we have finished with sample data, unpublish its module and plugins

PhilETaylor:
As requested in https://github.com/joomla/joomla-cms/pull/33150

### Summary of Changes

When we have finished with sample data, unpublish its module and plugins - therefore gaining a tiny bit of performance.

### Testing Instructions

Install Joomla 4
Install any of the sample data
refresh the Home Dashboard - note the module is no longer there
visit the Plugins and see they are unpublished
No side effects. 

After using the sample data plugins, the module and plugins remain published, which is more db calls, more code running on each page load

module and plugins are uninstalled for the life of a live site - gaining a tiny bit of performance as code is not run. 

none

### DIFF
--- a/administrator/modules/mod_sampledata/src/Helper/SampledataHelper.php
+++ b/administrator/modules/mod_sampledata/src/Helper/SampledataHelper.php
@@ -34,4 +34,30 @@ abstract class SampledataHelper
 
 		return Factory::getApplication()->triggerEvent('onSampledataGetOverview', array('test', 'foo'));
 	}
+
+	/**
+	 * When we have finished with sample data, unpublish its module and plugins
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	public static function unpublish()
+	{
+		$db = Factory::getContainer()->get('DatabaseDriver');
+
+		$query = $db->getQuery(true)
+			->update('#__modules')
+			->set($db->quoteName('published') . ' = 0')
+			->where($db->quoteName('module') . ' = ' . $db->quote('mod_sampledata'));
+
+		$db->setQuery($query);
+		$db->execute();
+
+		$query = $db->getQuery(true)
+			->update('#__extensions')
+			->set($db->quoteName('enabled') . ' = 0')
+			->where($db->quoteName('folder') . ' = ' . $db->quote('sampledata'));
+
+		$db->setQuery($query);
+		$db->execute();
+	}
 }
--- a/plugins/sampledata/blog/blog.php
+++ b/plugins/sampledata/blog/blog.php
@@ -19,6 +19,7 @@ use Joomla\CMS\Language\Text;
 use Joomla\CMS\Plugin\CMSPlugin;
 use Joomla\CMS\Session\Session;
 use Joomla\Database\ParameterType;
+use Joomla\Module\Sampledata\Administrator\Helper\SampledataHelper;
 
 /**
  * Sampledata - Blog Plugin
@@ -1790,6 +1791,8 @@ class PlgSampledataBlog extends CMSPlugin
 
 		$response['success'] = true;
 		$response['message'] = Text::_('PLG_SAMPLEDATA_BLOG_STEP4_SUCCESS');
+
+		SampledataHelper::unpublish();
 
 		return $response;
 	}
--- a/plugins/sampledata/multilang/multilang.php
+++ b/plugins/sampledata/multilang/multilang.php
@@ -23,6 +23,7 @@ use Joomla\CMS\Table\Table;
 use Joomla\CMS\Workflow\Workflow;
 use Joomla\Database\Exception\ExecutionFailureException;
 use Joomla\Database\ParameterType;
+use Joomla\Module\Sampledata\Administrator\Helper\SampledataHelper;
 
 /**
  * Sampledata - Multilang Plugin
@@ -486,6 +487,8 @@ class PlgSampledataMultilang extends CMSPlugin
 
 		$response['success'] = true;
 		$response['message'] = Text::_('PLG_SAMPLEDATA_MULTILANG_STEP8_SUCCESS');
+
+		SampledataHelper::unpublish();
 
 		return $response;
 	}
--- a/plugins/sampledata/multilang/multilang.php
+++ b/plugins/sampledata/multilang/multilang.php
@@ -488,7 +488,11 @@ class PlgSampledataMultilang extends CMSPlugin
 		$response['success'] = true;
 		$response['message'] = Text::_('PLG_SAMPLEDATA_MULTILANG_STEP8_SUCCESS');
 
-		SampledataHelper::unpublish();
+		// If only one language then unpublish the module/plugins
+		if (count(LanguageHelper::getContentLanguages(array(0, 1))) === 1)
+		{
+			SampledataHelper::unpublish();
+		}
 
 		return $response;
 	}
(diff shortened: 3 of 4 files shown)


PhilETaylor:
@dgrammatiko @richard67 cc

richard67:
This will make it harder to install Blog sample data multiple times when having a multilanguage site, one time for each backend language. It will need to enable the plugin again each time.

Currently we can change language in backend and install blog sample data for each language, and when we later install another backend language, we can do that again for that language.

This will not be possible anymore in this way.

Ping @infograf768 .

PhilETaylor:
> This will not be possible anymore in this way.

Not impossible, just more steps :) you would need to republish the module and plugin each time.

Lets keep this in context. for 99.9% of the life of a REAL Joomla site, the Sample Data plugin/Module are not needed to be running on the site.

The sample data plugin is for developers in development mode, and for those testing Joomla.

This is not a flagship feature of Joomla that is used day in and day out by 100% of those using Joomla. Its used by a select few people for a select reason.

dgrammatiko:
> Currently we can change language in backend and install blog sample data for each language

This is a flaw in the current architecture of the sampledata plugin/modules. The CMS has all the data (installed languages and content languages) to decide (or even ask the user) if the data should be installed in one or more languages. AFAIK there's already an open issue about sampledata: https://github.com/joomla/joomla-cms/issues/32878

richard67:
> AFAIK there's already an open issue about sampledata: #32878

That issue doesn't cover the multilanguage aspect.

Regarding this PR here it's not on me to decide if it's ok or not, I just wanted to mention that aspect here so it is not forgotten.

infograf768:
I agree with @richard67 
We absolutely need to be able to install multingual sample data and thereafter blog sample data for each content language.

PhilETaylor:
> We absolutely need to be able to install multingual sample data and thereafter blog sample data for each content language.

This PR doesn't stop you doing that at all.

It just means you need to REPUBLISH the module/plugin AFTER each use.

brianteeman:
As I have stated before ad nauseum there is zero need to have the multilingual sample data plugin enabled if the site is not multilingual

dgrammatiko:
> We absolutely need to be able to install multingual sample data and thereafter blog sample data for each content language.

If the only way to do that is by doing it manually we should pack it here and start farming...

PhilETaylor:
> As I have stated before ad nauseum there is zero need to have the multilingual sample data plugin enabled if the site is not multilingual

tell me a check I can make to find this out. I have never used Joomla and multiple languages. Im white british, the education system only taught me one language (badly)

dgrammatiko:
> tell me a check I can make to find this out

Count the languages in `__extensions select * where 'type' = 'language'`and the content languages in `__languages`

infograf768:
It is very simple, once one has tested the feature at will, to start from scratch _as was always done_ and then unpublish the module.

but, as usual, do whatever you like.

brianteeman:
Nothing in that sentence makes any sense at all

PhilETaylor:
> then unpublish the module.

I bet 10-99% of Joomla admins will never unpublish this module manually ;-)

richard67:
In Blog Sample Data we use this to check if we have multilanguage on:

https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Language/Multilanguage.php#L43

dgrammatiko:
> I bet 10-99% of Joomla admins will never unpublish this module manually ;-)

Anything less than 98% would be a surprise for me

brianteeman:
@PhilETaylor see #20749

PhilETaylor:
@brianteeman thanks - that's looks simple and that logic can be added to this PR... basically if there are more than 1 lang then we will not unpublish if you just installed the multilane sample data... simple. Keeps everyone happy.

brianteeman:
It is simple but one person objects to it as they believe no one will know you can do it unless you shove it in front of their face until they get so annoyed with it they uninstall joomla

PhilETaylor:
If I were (as I wanted to do) uninstalling (which means deleting the db rows and the module/plugin files from the hard disk) then I could understand... however all we are doing here is unpublishing here :) :

PhilETaylor:
https://github.com/joomla/joomla-cms/pull/33165/commits/081da589c02bf905edd99267d1b1f4904510f206 just pushed will only unpublish if the Multilane data is installed on a site with just ONE language 

Please test.

Bakual:
The module itself should not be unpublished. Keep in mind that the module isn't exclusively used for core sample data. Each 3rd party extension can provide its own sample data as well. And I know for example DP Calendar does that.

I'm not a big fan of extensions who publish/unpublish themself. I would offer a button to directly unpublish the plugin instead. So it can be hidden with a single click.

brianteeman:
We already have that button ;)

Disable the plugin automatically THEN
If there are no plugins to display the module is empty and as there is no point in having an empty module just hide the module. That way if an extension is installed later that has a sample data plugin it will still be displayed

PhilETaylor:
> Each 3rd party extension can provide its own sample data as well. And I know for example DP Calendar does that.

I bet you cannot name another 5 extensions that use sample data like this ;-) ;-) ;-)

dgrammatiko:
> I bet you cannot name another 5 extensions that use sample data like this ;-) ;-) ;-)

I can't name 5 but I can point you at https://github.com/dgrammatiko/sloth-pkg/tree/main/plg_sampledata which is way different from the core monstrosity plugins... ;)

PhilETaylor:
LMAO - typical JS developer :) hahahahah

```php
 defined('_JEXEC') || die('<html><head><script>location.href = location.origin</script></head></html>');
```

Bakual:
> > Each 3rd party extension can provide its own sample data as well. And I know for example DP Calendar does that.
> 
> I bet you cannot name another 5 extensions that use sample data like this ;-) ;-) ;-)

I dunno, but that was one of the initial ideas of using this approach. So instead that the extensions put their sample data hardcoded into their SQLs (like core did before), they can offer a plugin as well (like core does now).

Obviously, the module could return empty if no sample data plugin is active (don't know if it does that today already).

Bakual:
>>  I would offer a button to directly unpublish the plugin instead. So it can be hidden with a single click.

> We already have that button ;)

@brianteeman I meant for the plugins. I know that the modules all have such a button, that's where the idea came from 😄

infograf768:
https://github.com/joomla/joomla-cms/pull/20749#issuecomment-397266453

PhilETaylor:
> As I see it, if merged, this feature would become a hidden secret which would only be eventually used by devs and bugsquad.

Well it certainly doesn't need to command prize position on the home page of the admin console does it? it gets used less times in a sites lifetime than the number of fingers on a hand...

Bakual:
Imho, it would be nice to be in that position if there are sample data available. Otherwise people don't see them.

But it would also nice to be able to unpublish the plugins directly from the module, as it is not very intuitive to know where they can be hidden. So a button either on each sample data set or on the module ("Manage Sample Data" for the lack of better wording) would be great.

It would also be nice if the module hides completely if no plugin is active.
Currently it shows like this:
image

I certainly wouldn't automatically unpublish anything.

Bakual:
I've created a simple PR to hide the module if no sample data plugin is published. See https://github.com/joomla/joomla-cms/pull/33168

Bakual:
Button to go directly to the pluginmanager to disable plugins:
https://github.com/joomla/joomla-cms/pull/33169

PR #33155, opened 2021-04-15
[4] Correct return type as CLI Application has no menu.

PhilETaylor:
code review

PR #33184, opened 2021-04-18
[4] Correct docblock in stubGenerator

PhilETaylor:
replacement for https://github.com/joomla/joomla-cms/pull/33155

Code review.

### DIFF
--- a/build/stubGenerator.php
+++ b/build/stubGenerator.php
@@ -116,7 +116,7 @@ PHP;
 	 * @param string $name    The application name for the menu
 	 * @param array  $options An array of options to initialise the menu with
 	 *
-	 * @return  \Joomla\CMS\Menu\AbstractMenu|null  An AbstractMenu object or null if not set.
+	 * @throws   \BadMethodCallException  Exception thrown as CLI Application has no menu.
 	 *
 	 * @since   4.0.0
 	 */


richard67:
Would a `@return  void` in addition make sense? Or would that be wrong?

PhilETaylor:
I wrote this up and deleted it before I submitted... chose not to include the commentary - If you or anyone else can prove that this method actually ever returns anything, then we can documentation for that return, but https://3v4l.org/gPnHd leads me to believe that nothing is ever "returned".

richard67:
> I wrote this up and deleted it before I submitted... chose not to include the commentary - If you or anyone else can prove that this method actually ever returns anything, then we can documentation for that return, but https://3v4l.org/gPnHd leads me to believe that nothing is ever "returned".

You are right, it never will return anything.

richard67:
I have tested this item :white_check_mark: successfully on 94b39ddadf17ac93525669fcb468317f2dd5e906<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33184">issues.joomla.org/tracker/joomla-cms/33184</a>.</sub>

PR #33186, opened 2021-04-18
[4.0] Categories don't contain only articles

Bakual:
Batch copying a category allows to reverse order all articles in the selected categories. But categories are not only for articles. It can be anything. So "items" is a better term.

### Summary of Changes
Changing the string to order items, not articles.

### Testing Instructions
Go to any categories manager (eg articles categories, banner categories, contact categories), select at least one category and then select "Batch" from the "Actions" dropdown. 

image

image

None


brianteeman:
Are you sure that the bug is not the text but that it should only be shown when the context is com_content? In J3 it only displays in com_content

Bakual:
Oops, that could be the case. I just assumed it works, but didn't test it. I'll check that and close this PR in the meantime.

Bakual:
@brianteeman You're right, it only affects articles. But the bug is also in J3, as it shows in all category types, not only com_content. It probably shouldn't have been implemented in the com_category code 😄

Bakual:
New PR: https://github.com/joomla/joomla-cms/pull/33190

PR #33203, opened 2021-04-20
[4.0] added description in reCaptcha

rjharishabh:
Pull Request for Issue #.

### Summary of Changes
Added description for form fields
* ```plugins\captcha\recaptcha\recaptcha.xml```

### Testing Instructions

* Open plugin reCaptcha in the backend

and see the changes

recaptcha-before

recaptcha-after

No


YatharthVyas:
The descriptions work fine.
However, you've added a new `version` field in reCaptcha invisible but this field hasn't been used anywhere in [recaptcha_invisible.php](https://github.com/joomla/joomla-cms/blob/staging/plugins/captcha/recaptcha_invisible/recaptcha_invisible.php)

Here, the version is unconditionally set as `auto` which makes this new field redundant.
https://github.com/joomla/joomla-cms/blob/697a331a94ae94cb8d9db4066701667e19901df8/plugins/captcha/recaptcha_invisible/recaptcha_invisible.php#L71

Please correct me if I'm wrong here but I couldn't find a single occurrence of `$this->params->get('version'` in recaptcha_invisible.php which means that the new field isn't getting used anywhere 

I believe your motivation behind adding this field was because it was present in reCaptcha but here it was also being used to render the captcha and load the script as per the version selected
https://github.com/joomla/joomla-cms/blob/697a331a94ae94cb8d9db4066701667e19901df8/plugins/captcha/recaptcha/recaptcha.php#L68

https://github.com/joomla/joomla-cms/blob/697a331a94ae94cb8d9db4066701667e19901df8/plugins/captcha/recaptcha/recaptcha.php#L145

richard67:
What shall the new field be good for if it has only one option value "2.0"? If it shall be for information only, it doesn't need a dropdown.

brianteeman:
This serves no purpose

rjharishabh:
There are two variants in version 2, invisible and tickbox.
Added description

richard67:
yes but always version 2, not 3.

brianteeman:
> There are two variants in version 2, invisible and tickbox.

Thats why there are two different plugins

dgrammatiko:
> Here, the version is unconditionally set as auto which makes this new field redundant.

This is irrelevant to the google version, it is part of the Joomla API and for local hosted assets it generates a `?some-hash` so the assets could be correctly invalidated on updates.

richard67:
> > There are two variants in version 2, invisible and tickbox.
> 
> Thats why there are two different plugins

Exactly, and that's good because we can use the one here and the other one there. If we had only one plugin and made all adjustments in that one plugin, it would be the same everywhere, and we could not chose to show e.g. an invisible captcha on this page and a visible on on another page.

brianteeman:
The only thing that could possibly be changed would be the plugin description.

rjharishabh:
@brianteeman shall I remove the version label in Invisible reCaptcha

brianteeman:
As stated before this is simply wrong.

PR #33207, opened 2021-04-20
[4.0] Adapt unit tests to PR 32718 having been merged up from 3.10

richard67:
Pull Request for Issue # .

### Summary of Changes

Adapt the unit tests for the FilePathRule validation rule to PR #32718 , which meanwhile has been merged up from 3.10-dev.

### Testing Instructions

Check if unit tests are passing in Appveyor and Drone.

### Actual result BEFORE applying this Pull Request

Unit tests failing on the 4.0-dev branch.

### Expected result AFTER applying this Pull Request

Unit tests passing for this PR here.

### Documentation Changes Required

None.

PR #33208, opened 2021-04-20
[4.3] Moving menu lookup for tags from Router helper to actual router

Hackwar:
### Summary of Changes
This moves the menu lookup for tags from the router helper to the actual router. This improves the tag routing a little bit, especially allowing for menu items with more than one tag. However lookup of multi-level tags still doesn't work and would require something like the CategoryNode class for tags, which we don't have and wont have for 4.0. 

### Testing Instructions
After applying this PR, the generated URLs shouldn't have changed.


richard67:
Failing unit tests in drone and appveyor are not related to this PR but to the 4.0-dev branch at the time when this PR was created based on it. It will be fixed in the 40-dev branch with #33207 and if that is merged, this PR should be updated to that so we can see if unit tests pass here.

joomdonation:
I have tested this item :red_circle: unsuccessfully on cb9b7604aae4610dc58102fa050fd45134e06796<br /><br />I tested this using sample blog data. Link to a tag has this format http://localhost/joomla4/index.php/component/tags/millions?Itemid=101 and it causes the error (something like NomenuRules which you fixed sometime ago)

I wonder why we could not use same Router structure with other extensions?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33208">issues.joomla.org/tracker/joomla-cms/33208</a>.</sub>

Hackwar:
The problem is, that the tags don't use simple IDs, but also arrays of IDs. There is actually a chunk of issues here, but unfortunately out of scope for 4.0.

joomdonation:
@Hackwar So what we can do with this PR ? Close it now and re-visit it later? Or leave it open and you will find a solution for this later?

Hackwar:
I'll have another look at this. If it is not merged into J4, we will have to look if we can do this in a b/c way in 4.1 or similar.

components/com_tags/src/Service/Router.php
```diff
@@ -219,4 +356,83 @@ protected function fixSegment($segment)
...
+	 * @since   4.0
+	 */
+	protected function buildLookup($language = '*')
+	{
+		// Prepare the reverse lookup array.
+		if (!isset($this->lookup[$language]))
+		{
+			$this->lookup[$language] = array();
+
+			$component  = ComponentHelper::getComponent('com_tags');
```
Babofett, line 376: You forgot to refer to the ComponentHelper.
e.g. Line 21 `use Joomla\CMS\Component\ComponentHelper;`

Babofett (changes requested):
You need to refer to the ComponentHelper in Router.php
e.g. Line 21 `use Joomla\CMS\Component\ComponentHelper;`

components/com_tags/src/Service/Router.php
```diff
@@ -26,11 +27,19 @@
 class Router extends RouterBase
 {
 	/**
+	 * Lookup array of the menu items
+	 *
+	 * @var   array
+	 * @since 4.1
```
PhilETaylor:
```suggestion
	 * @since __DEPLOY_VERSION__
```

components/com_tags/src/Service/Router.php
```diff
@@ -26,11 +27,19 @@
...
+	 */
+	protected $lookup = array();
+
+  /**
 	 * The db
 	 *
 	 * @var DatabaseInterface
 	 *
-	 * @since  4.0.0
+	 * @since  4.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

components/com_tags/src/Service/Router.php
```diff
@@ -46,9 +55,137 @@ class Router extends RouterBase
...
+	}
+
+	/**
+	 * Method to preprocess a URL
+	 *
+	 * @param   array  $query  An associative array of URL arguments
+	 *
+	 * @return  array  The URL arguments to use to assemble the subsequent URL.
+	 *
+	 * @since   4.0
```
PhilETaylor:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

components/com_tags/src/Service/Router.php
```diff
@@ -219,4 +356,83 @@ protected function fixSegment($segment)
...
 	}
+
+	/**
+	 * Method to build the lookup array
+	 *
+	 * @param   string  $language  The language that the lookup should be built up for
+	 *
+	 * @return  void
+	 *
+	 * @since   4.0
```
PhilETaylor:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

joomdonation:
I just tested it quickly to see if it solves the issue https://github.com/joomla/joomla-cms/issues/35382 but it is still not working yet.

DonPedro34:
Any hope that this will be fixed soon?
https://issues.joomla.org/tracker/joomla-cms/35382

HLeithner:

Hackwar:
This isn't going anywhere. I'm closing this one and will maybe start over sometime soon... or something like that...

PR #33212, opened 2021-04-20
[3.9][Privacy] Block usage of FLoC by default

Hackwar:
### Issue
As a replacement for third party cookies, Google has introduced the Federated Learning of Cohorts into the Chrome browser lately. This feature is supposed to allow better tracking of users while also keeping data privacy. Details can be read [here](https://github.com/WICG/floc). The Joomla! project disagrees that this feature is in the interests of the owners of Joomla!-powered websites as well as their visitors. We share the [assessment](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea) of the EFF as well as many other organisations ([Mozilla, Microsoft, Opera](https://www.theverge.com/2021/4/16/22387492/google-floc-ad-tech-privacy-browsers-brave-vivaldi-edge-mozilla-chrome-safari), [WordPress]( https://make.wordpress.org/core/2021/04/18/proposal-treat-floc-as-a-security-concern/)).

We consider this a security issue and thus will roll out countermeasures with the next bugfix release of the Joomla 3.9 series.

### Summary of Changes
This PR introduces a header to disable Federated Learning of Cohorts, which is sent with every request that is handled by the Joomla framework. The header specifically looks like this:
`
Permissions-Policy: interest-cohort=()
`
and is the equivalent of opting out of this feature. Please be aware that this only adds this header to requests which go through the Joomla! application. Assets like CSS files, images or javascript are handled by your webserver directly and we would advise to modify the webserver to add this header to every request directly.

If you really want to disable the blocking of this feature, we added a switch in the global configuration to remove this header.

A postinstall message has been added to the database to inform administrators of this decision.

### Testing Instructions
Apply this patch and load a random page of your testing site. The answer to your request should contain the above noted header.

None.

### Handling of this in the future
This PR fixes this issue for the 3.9 and 3.10 releases. We will decide on how to implement this countermeasure in 4.0 in the time before the final release.


administrator/language/en-GB/en-GB.com_admin.ini
```diff
@@ -139,6 +139,8 @@ COM_ADMIN_PLATFORM_VERSION="Joomla! Platform Version"
...
 COM_ADMIN_POSTINSTALL_MSG_BEHIND_LOAD_BALANCER_TITLE="New Server Setting \"Behind Load Balancer\""
+COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="Google is introducing a replacement for third party tracking cookies, a feature named Federated Learning of Cohorts. You can read more about it <a href="https://github.com/WICG/floc">here</a>. The Joomla! project believes this feature to be problematic and against the interests of our users, both the owners of Joomla!-powered sites, as well as their visitors. We do follow the EFF in their <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">assessment</a> and have added a feature to block this technology on all Joomla sites. This feature has been enabled by default. If you disagree with this decision and want to allow FLoC for your website, you can disable it from the Global Configuration."
+COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_TITLE="Block Federated Learning of Cohorts"
```
brianteeman:
```suggestion
COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_TITLE="Block Federated Learning of Cohorts (FLoC)"
```

As this is being reported in most places just by the acronym its important that we use it.

administrator/language/en-GB/en-GB.com_admin.ini
```diff
@@ -139,6 +139,8 @@ COM_ADMIN_PLATFORM_VERSION="Joomla! Platform Version"
...
 COM_ADMIN_POSTINSTALL_MSG_BEHIND_LOAD_BALANCER_TITLE="New Server Setting \"Behind Load Balancer\""
+COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="Google is introducing a replacement for third party tracking cookies, a feature named Federated Learning of Cohorts. You can read more about it <a href="https://github.com/WICG/floc">here</a>. The Joomla! project believes this feature to be problematic and against the interests of our users, both the owners of Joomla!-powered sites, as well as their visitors. We do follow the EFF in their <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">assessment</a> and have added a feature to block this technology on all Joomla sites. This feature has been enabled by default. If you disagree with this decision and want to allow FLoC for your website, you can disable it from the Global Configuration."
```
brianteeman:
```suggestion
COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="Google is introducing a replacement for third party tracking cookies, a feature named Federated Learning of Cohorts (FLoC). You can read more about it <a href="https://github.com/WICG/floc">here</a>. The Joomla! project believes this feature to be problematic and against the interests of our users, both the owners of Joomla!-powered sites, as well as their visitors. We do follow the EFF in their <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">assessment</a> and have added a feature to block this technology on all Joomla sites. This feature has been enabled by default. If you disagree with this decision and want to allow FLoC for your website, you can disable it from the Global Configuration."
```

administrator/language/en-GB/en-GB.com_admin.ini
brianteeman: It is also standard practice to explain what an acronym stands for in this way.

administrator/language/en-GB/en-GB.com_config.ini
```diff
@@ -93,6 +93,8 @@ COM_CONFIG_FIELD_FILTERS_CUSTOM_BLACK_LIST="Custom Blacklist"
 COM_CONFIG_FIELD_FILTERS_NO_HTML="No HTML"
 COM_CONFIG_FIELD_FILTERS_NO_FILTER="No Filtering"
 COM_CONFIG_FIELD_FILTERS_WHITE_LIST="Whitelist"
+COM_CONFIG_FIELD_FLOC_BLOCKER_DESC="Send a header to not support the tracking through the FLoC method proposed by Google."
```
brianteeman: is it "not support" or is it "to prevent" ?

PhilETaylor:
> We will decide on how to implement this countermeasure in 4.0 in the time before the final release.

In Joomla 4 this should just be part of https://docs.joomla.org/J4.x:Http_Header_Management

PhilETaylor:
Should we also be putting this in htaccess.txt like the other headers like no-sniff?

PhilETaylor:
> "Google is introducing a replacement..."

Should the language be written time independently, so that in a decade it is still true? Maybe something like "In 2021, Google introduced..."

administrator/language/en-GB/en-GB.com_config.ini
Hackwar: Unfortunately it is to "not support", at least from my PoV. The header only announces that we don't want to be tracked, but if the browser honours that we don't know.

Hackwar:
> > We will decide on how to implement this countermeasure in 4.0 in the time before the final release.
> 
> In Joomla 4 this should just be part of [docs.joomla.org/J4.x:Http_Header_Management](https://docs.joomla.org/J4.x:Http_Header_Management)

Agreed, we are just still discussing about that. We definitely don't want the above special case to live on in J4.

Hackwar:
> > Assets like CSS files, images or javascript are handled by your webserver directly and we would advise to modify the webserver to add this header to every request directly.
> 
> Should we also be putting this in htaccess.txt like the other headers like no-sniff? This would then apply the header globally and not only on PHP requests through the app.

Maybe, however since most people don't update their htaccess on a Joomla update, we can't rely on this alone. Be my guest to add this header to the htaccess.

Hackwar:
> > "Google is introducing a replacement..."
> 
> Should the language be written time independently, so that in a decade it is still true? Maybe something like "In 2021, Google introduced..."

I'm happy to accept a changed text PR to this branch.

PhilETaylor:
>  we can't rely on this alone

Agreed, but the htaccess is still the right place for this also to be, over and above this PR. Just thought you might like to add it here for completeness.

PhilETaylor:
> I'm happy to accept a changed text PR to this branch.

haha that's @brianteeman domain :) I only speak crap.

PhilETaylor:
> Maybe, however since most people don't update their htaccess on a Joomla update, we can't rely on this alone. Be my guest to add this header to the htaccess.

Also need to check if .htaccess has this header, and Joomla outputs this header, do they resolve to a single header, or to two separate duplicated headers in the response - need to test that first.

brianteeman:
Such a shame that the project doesn't take the use of cookies as seriously

Hackwar:
> > Maybe, however since most people don't update their htaccess on a Joomla update, we can't rely on this alone. Be my guest to add this header to the htaccess.
> 
> Also need to check if .htaccess has this header, and Joomla outputs this header, do they resolve to a single header, or to two separate duplicated headers in the response - need to test that first.

Header has been set in htaccess.txt as well. Testing on my system only showed one header to be sent.

richard67:
Update SQL scripts for PostgreSQL and for MS SQL Server/SQL Azure are missing.

Hackwar:
Have been added. Sorry, it was late yesterday...

htaccess.txt
```diff
@@ -24,6 +24,7 @@
 ## Suppress mime type detection in browsers for unknown types
 <IfModule mod_headers.c>
 Header always set X-Content-Type-Options "nosniff"
+Header always set Permissions-Policy: interest-cohort=()
```
richard67: Your new line has a colon and is not quoting the value, the existing line above has no colon and quotes the value with double quotes. This looks at least weird, inconsistent to me, and makes me think that the colon might be wrong for this directive.

bembelimen:
> 
> 
> Such a shame that the project doesn't take the use of cookies as seriously

We try to improve: https://docs.joomla.org/GSoC_2021_Project_Ideas#Cookies_management

htaccess.txt
richard67:
```suggestion
Header always set Permissions-Policy "interest-cohort=()"
```
See examples on https://httpd.apache.org/docs/current/mod/mod_headers.html .

htaccess.txt
```diff
@@ -24,6 +24,7 @@
 ## Suppress mime type detection in browsers for unknown types
```
PhilETaylor, line 24: An additional comment about the Permissions Policy is needed, especially to state that if you have enabled the feature in Joomla, or if you need a custom Permissions Policy then its best not to enable this blindly in .htaccess

dgrammatiko:
Since J4 has already the system_httpheader and @zero-24 already did the backport for J3 https://github.com/zero-24/plg_system_httpheader why don't you merge that to J3 (assuming that @zero-24 has no objections)?

PhilETaylor:
Was consideration given to implement this as a standalone Plugin instead of "yet another" configuration option? 

Was consideration given to the Post Installation message, to have "condition" for showing only if not enabled in global config and "a one click action" to implement the change like the 2FA and the Load Balancer change recently?

Hackwar:
> Since J4 has already the system_httpheader and @zero-24 already did the backport for J3 [zero-24/plg_system_httpheader](https://github.com/zero-24/plg_system_httpheader) why don't you merge that to J3 (assuming that @zero-24 has no objections)?

> Was consideration given to implement this as a standalone Plugin instead of "yet another" configuration option?

It was decided to keep the implementation minimal and to not use a separate plugin.

> Was consideration given to the Post Installation message, to have "condition" for showing only if not enabled in global config and "a one click action" to implement the change like the 2FA and the Load Balancer change recently?

This feature is enabled by default, so people would need to disable it actively. In that case they can also click away the post install message.

rodlie:
> We consider this a security issue and thus will roll out countermeasures

This has nothing to do with security, add it in J4 (default off), but don't force this in J3 in the name of security. At best this is a privacy concern (in the browser).

nikosdion:
I see that Joomla is trying to not fall behind WP but please note that WP is NOT including this feature in its code, despite what the tech press misguidedly reported...

Moreover, this is not a definite security or privacy issue. Please do read https://twitter.com/Log3overLog2/status/1384337637763387394 for the whole story. In short, FLoC is only a concern if you include JavaScript which explicitly requests the FLoC cohort. Moreover, FLoC is currently an experiment, not a final implementation. In the few Chrome browsers this experiment is enabled — notably none in the EU, where the author of this PR lives — this is currently an opt-out feature just because it's the only way to run an experiment of a technology nobody has heard about before. Google seems to indicate that the final implementation is very likely to be opt-in, not opt-out, or at the very least provide a user warning the same way that asking for location data does. It is, in fact, extremely likely to be so because of GDRP and that's probably why the opt-out format experiment does not run in browsers of EU residents.

Further to that, this has NOTHING AT ALL to do with security. Your site can't be hacked with FLoC. Maybe — **MAYBE** — it will have to do with privacy **IF AND ONLY IF** the final implementation of FLoC is opt-out instead of opt-in.  Marking it as a _security_ concern dilutes the importance of fixing _real security issues_ and projects the wrong message to users. If we label speculative, non-security issues as "SECURITY" the users are trained to ignore any security fix as superficial and unimportant, undermining their safety in the long term.

On top of that, sending a Permissions-Policy to disable FLoC doesn't mean that the user will magically no longer be tracked. In fact, it's possible that instead of being assigned into a cohort of at least 1000 users he will be otherwise tracked and placed in a finer grained cohort using a different set of tracking technologies including but not limited to ISP detection, IP geolocation, per-user subdomain pinging and so on. FLoC was actually designed to strengthen the privacy by making cohorts large enough so as not to allow individual tracking and deanonymisation. Whether we trust Google to do that is a largely philosophical question, not a technical one. From a technical perspective using the Permissions-Policy might actually prove counter-productive to the stated goal of protecting the user's privacy.

It is also worth noting that FLoC is only ever enabled if you include JavaScript in your page which uses FLoC (either directly on the page or via an IFRAME). This means that unless you are using analytics, ads or similar technology which uses FLoC-enabled JavaScript there is no placement in a cohort. Pretending that merely using Chrome on any site is magically destroying your privacy is misguided at best.

Based on this, there is no real benefit in including this in Joomla, let alone as a feature enabled by default. We are not protecting anyone from anything, especially the site owner i.e. it's not a "security" feature. What this feature does is probably disable FLoC if you are using third party ads, analytics etc on your site. Whether it will succeed depends entirely on the implementation. If a third party ad is included in your site via an IFRAME it's easy for it to add the `allow="interest-cohort"` in the attribute, nullifying this "protection". It would only work if the FLoC-enabled third-party JS only ever runs in the main page context on your site. Even then, whether it would be something that actually protects your users instead of enabling more tracking for them is something to be determined. Not to mention that we still _do not know_ if Chrome will make it opt-in or display a warning before interest cohorts are used, especially in jurisdictions with strict personally identifiable information legislation. 

In short, this PR jumps the shark by a freaking mile.

Finally, the implementation as it stands right now is **completely wrong** and breaks sites. It replaces the Permissions-Policy header instead of amending it. People who use plugins to set the Permissions-Policy for legal compliance, privacy or security reasons will be very surprised to see that their Permissions-Policy is overwritten and replaced with a BS non-fix of a speculative issue. So, while nothing (or at the very least nothing of importance) is fixed something of _actual_ importance is broken for them! A better way to do it is more like along these lines (from a PR in Admin Tools Professional I have not merged yet for the reasons stated above):

```php
	public function onAfterRender_disableFLoC()
	{
		// Get all headers
		$headers = $this->app->getHeaders();

		// Find the Permissions-Policy header's contents
		$headerContent = array_reduce($headers, function (?string $carry, array $headerDefinition) {
			if (strtolower($headerDefinition['name'] ?? '') === 'permissions-policy')
			{
				$carry = ($headerDefinition['value'] ?? '') ?: $carry;
			}
		}, '');

		$headerContent = trim($headerContent ?? '');

		// If the content is empty, append our special interest-cohort permissions
		if (empty($headerContent))
		{
			$headerContent = 'interest-cohort=()';
		}
		// If it's not empty and doesn't have an interest-cohort permissions add one
		elseif (stripos($headerContent, 'interest-cohort') === false)
		{
			$headerContent .= ',interest-cohort=()';
		}

		// Separate into separate permissions and replace any existing interest-cohort permissions with our special one.
		$permissions = array_map(function ($permission) {
			$permission = trim($permission ?? '');

			if (stripos($permission, 'interest-cohort') !== 0) {
				return $permission;
			}

			return 'interest-cohort=()';
		}, explode(',', $headerContent));

		// Remove empty permissions which would make our header invalid
		$permissions = array_filter($permissions, function ($x) {
			return !empty($x);
		});

		// Recombine everything into one big string separated by semicolons
		$headerContent = implode(';', $permissions);

		// Set the header in the application's response.
		$this->app->setHeader('Permissions-Policy', $headerContent, true);
	}
```

While this won't overwrite a permissions policy set by a plugin — including Joomla 4's core feature — it's still mostly window dressing because it can and will be overwritten by .htaccess or equivalent server-level code if it sets the Permissions-Policy header and can and will be overwritten by IFRAME attributes. As we say in Greece, this is very much like drilling holes in the water...

It's important to get all the facts straight before jumping the gun and implementing something under a false security banner, even more so when the implementation is plain bad code that breaks another core feature.

brianteeman:
I believe (but may be wrong) that the only reason it is tagged as security is because thats what wordpress did. However they only did that so that they could apply the change outside of their normal release procedure.

ReLater:
For me, Google and others leveraging my privacy through the back door is a **security issue**. Preparing for such an issue early on is the right way to go and a good signal.  The counter-reactions aren't just coming from WordPress. Google statements and maybe-promises as arguments against such a preparation are not to be considered serious.

brianteeman:
@relater so dont enable anything on your site that implements FLoC or similar. This doesnt just appear on your site without you doing something

Hackwar:
https://github.blog/changelog/2021-04-27-github-pages-permissions-policy-interest-cohort-header-added-to-all-pages-sites/

PhilETaylor:
Interesting that GitHub would force this on its users pages - but not on its own site, or the GitHub.com service entirely. :)

brianteeman:
Since github enabled this "feature" chrome browser 90.0.4430.93 (Official Build) (64-bit) issues the following error

Error with Permissions-Policy header: Unrecognized feature: 'interest-cohort'.

image

PhilETaylor:
Symfony just implemented it https://twitter.com/fabpot/status/1387777376898228232

> Fabien Potencier @fabpot #Google FLoC is now out on symfony.com. All other #Symfony websites will follow soon.

libraries/src/Application/CMSApplication.php
```diff
@@ -199,6 +199,11 @@ public function enqueueMessage($msg, $type = 'message')
 	 */
 	public function execute()
 	{
+		if ($this->get('block_floc', 1))
```
PhilETaylor: I would say this is the wrong place for setting this header. 

It should be done in the `toString` method (at the end of the file near where the cache headers are set and before the headers are sent to the browser https://github.com/joomla/joomla-cms/blob/21f71d0d1971d8dad5d1ff02da73575151d3d248/libraries/src/Application/CMSApplication.php#L1216) and it should first check to see if anything has already set a header for permissions-policy, and then if not then it can add this header. This will allow developers to make a choice, and override this by setting their own values before Joomla does, regardless of the setting in global config.

dgrammatiko: > and it should first check to see if anything has already set a header for permissions-policy,

You don't have to do this. This is a **NEW** key-val and you **do** want the core settings to take over since there is a GUI exposing the setting! In short, the code is safe as is... My 2c

dgrammatiko: https://caniuse.com/permissions-policy

PhilETaylor: A gui setting that will be deprecated as soon as its implmented as its not going to be implemented this way in Joomla 4 apparently. 

Symfony make a check to see if the header is set before blindly overwriting it - https://github.com/symfony/symfony/pull/40989/files 

The discussion from WordPress has code that shows they will take into account any previous setting also. https://make.wordpress.org/core/2021/04/18/proposal-treat-floc-as-a-security-concern/

remember that FLOC, is only one of many values for PERMISSIONS POLICY - many sites ALREADY output a permissions policy!!! What this guy setting will blindly do is override that (depending on how the site currently implements it)

PhilETaylor: but remember that permissions-policy is just a renamed feature-policy header - and that had a lot of support and we can expect permissions-policy also to gain a lot of support https://caniuse.com/feature-policy

dgrammatiko: Also please remember that this PR is for J3 which is a kinda dead fish...

PhilETaylor: ;-) A series that has a policy of "No new features will be added" haha :-)

PhilETaylor:
So I have been testing this today. 

Although @Hackwar [states he tested](https://github.com/joomla/joomla-cms/pull/33212#issuecomment-823843179) .htaccess headers - my findings directly contradict his. 

IF you have .htaccess line (which is enabled by default if you use the Joomla provided htaccess) and you have `block_floc` enabled (which is enabled by default), and IF you are running apache with Headers module enabled, then you get a double header value which is not valid

```
Permissions-Policy: interest-cohort=(), interest-cohort=()
```

If my .htaccess already contains a "Permissions-Policy" header (Which is very possible for sites that have been using Feature Policy headers promoted for years by securityheaders.com) - for example this header in .htaccess:

```
Header always set Permissions-Policy "accelerometer=(), ambient-light-sensor=(), autoplay=(self), camera=(), encrypted-media=(), fullscreen=(self), geolocation=(), gyroscope=(), magnetometer=(), microphone=(), midi=(), payment=(self), picture-in-picture=(), speaker=(), sync-xhr=(), usb=(), vr=()"
```

And then I upgrade to Joomla 3.x that contains FLoC enabled by default, I then get a response with both my existing settings, and the Joomla enforced setting as separate headings:

```
Permissions-Policy: accelerometer=(), ambient-light-sensor=(), autoplay=(self), camera=(), encrypted-media=(), fullscreen=(self), geolocation=(), gyroscope=(), magnetometer=(), microphone=(), midi=(), payment=(self), picture-in-picture=(), speaker=(), sync-xhr=(), usb=(), vr=()
Permissions-Policy: interest-cohort=()
```

Therefore it will be REALLY REALLY REALLY Important to ACCURATELY COMMUNICATE this b/c incompatible change if it is merged into Joomla 3.

Hackwar:
I've updated the implementation. As @PhilETaylor noticed, my testing of the duplicate headers has been false. I handled the tool to display the headers wrongly, which resulted in this error. I apologise for this mistake. The goal is to send this header with every request from our server, so we would have to include it in our .htaccess. Since I couldn't find a reliable way to detect if this header has already been set when in the .htaccess, I commented the rule and added a pointer to our postinstall message.

The postinstall message has now also been correctly added to the install SQLs.

Regarding duplicate or overwritten headers: Again, I wasn't able to find a way to detect headers set in the .htaccess inside of PHP and my tests didn't allow me to read headers set by .htaccess. So if someone has set the Permissions-Policy in his .htaccess, our code would indeed add a duplicate header. If the header was set in PHP, I removed the $replace flag in order to not overwrite previously set headers. The code was initially inserted before the application is executed in order to catch as many requests as possible, but as with the .htaccess directive, the new position would be the safer solution while not being sent with every request.

I'm open for improvements to the postinstall message.

htaccess.txt
```diff
@@ -24,6 +24,13 @@
...
 <IfModule mod_headers.c>
 Header always set X-Content-Type-Options "nosniff"
+##
+# Joomla considers the feature of Federated Learning of Cohorts (FLoC) to be against the
+# interest of users and site owners. We encourage you to block this feature by
+# uncomment the below directive. Read more about this in the Post-Installation message
```
PhilETaylor:
```suggestion
# uncommenting the below directive. Read more about this in the Post-Installation message
```

PhilETaylor:
The OSM, Inc. - F[ull Board Meeting - April 22, 2021 minutes](https://volunteers.joomla.org/board-of-directors/reports/1495-osm-inc-full-board-meeting-april-22-2021) state:

>  The Production Department is looking into a way to support this feature but have it disabled by default for a release of Joomla! 3.9.27

But this PR makes it enabled by default. 

Was there a design to make this enabled by default or are the minutes incorrect? which is it?

brianteeman:
I think that meant FLoC would be disabled by default

rachellawson:
FYI now committed to Drupal core: https://www.drupal.org/project/drupal/issues/3209628

brianteeman:
Thanks for letting us know @rachellawson much appreciated

richard67:
@Hackwar There is still missing what @PhilETaylor wrote in [his comment above](https://github.com/joomla/joomla-cms/pull/33212#issuecomment-829402014) as last:

> (I.e that if you enable it in .htaccess then you must disable it in joomla global config!)

Even if in brackets, it is the most important information in order to avoid the duplicate and so invalid header value, and as far as I can see neither the comment in htaccess.txt nor the language strings tell that as clearly as Phil's comment.

But I think it should be done somewhere.

Update 2021-05-18Z14:21: Solved with https://github.com/joomla/joomla-cms/pull/33212/commits/d4cc16b2932c79aa94b8866f709bd214301a6be2 .

htaccess.txt
```diff
@@ -24,6 +24,13 @@
 ## Suppress mime type detection in browsers for unknown types
 <IfModule mod_headers.c>
 Header always set X-Content-Type-Options "nosniff"
+##
+# Joomla considers the feature of Federated Learning of Cohorts (FLoC) to be against the
+# interest of users and site owners. We encourage you to block this feature by
+# uncommenting the below directive. Read more about this in the Post-Installation message
+# in the backend.
```
richard67:
```suggestion
# Joomla considers the feature of Federated Learning of Cohorts (FLoC) to be against the
# interest of users and site owners. If the option "Block Google FLoC" is switched on in
# Global Configuration, the Joomla application adds an http header to each http response
# to block FLoC. But this applies only to request going to the application.
# Therefore we encourage you to uncomment the below directive to apply this header to ALL
# http responses. If you do that, switch off the "Block Google FLoC" option in Global
# Configuration, too.
# Read more about this in the Post-Installation message in the backend.
```
@Hackwar We should write something like that to inform people to switch off the backend option when using the htaccess directive in order not to get duplicate header values.

@brianteeman Maybe you can find a shorter way, my explanations are sometimes a bit long.

libraries/src/Application/CMSApplication.php
```diff
@@ -211,7 +211,30 @@ public function execute()
...
 		if ($this->get('block_floc', 1))
 		{
-			$this->setHeader('Permissions-Policy', 'interest-cohort=()');
+			$headers = $this->getHeaders();
+
+			$notPresent = true;
+
+			foreach($headers as $header)
+			{
+				if ($header['name'] === 'Permissions-Policy')
```
PhilETaylor: Header names are case insensitive. 

From RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1", Section 4.2, "Message Headers":

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

The updating RFC 7230 does not list any changes from RFC 2616 at this part.

administrator/language/en-GB/en-GB.com_admin.ini
```diff
@@ -139,7 +139,7 @@ COM_ADMIN_PLATFORM_VERSION="Joomla! Platform Version"
...
+COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="Google is introducing a replacement for third party tracking cookies, a feature named Federated Learning of Cohorts (FLoC). You can read more about it <a href="https://github.com/WICG/floc">here</a>. The Joomla! project believes this feature to be problematic and against the interests of our users, both the owners of Joomla!-powered sites, as well as their visitors. We do follow the EFF in their <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">assessment</a> and have added a feature to block this technology on all Joomla sites. This feature has been enabled by default. If you disagree with this decision and want to allow FLoC for your website, you can disable it from the Global Configuration. To additionally enable this feature for all other requests to your server, enable it in your .htaccess. In that case you have to disable the block in the Global Configuration."
```
PhilETaylor:
```suggestion
COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="Google is introducing a replacement for third party tracking cookies, a feature named Federated Learning of Cohorts (FLoC). You can read more about it <a href="https://github.com/WICG/floc">here</a>. The Joomla! project believes this feature to be problematic and against the interests of our users, both the owners of Joomla!-powered sites, as well as their visitors. We do follow the EFF in their <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">assessment</a> and have added a feature to block this technology on all Joomla sites. This feature has been enabled by default. If you disagree with this decision and want to allow FLoC for your website, you can disable it from the Global Configuration. To additionally enable this feature for all other requests to your server, enable it in your .htaccess. In that case you have to disable this feature in the Global Configuration."
```

htaccess.txt
```diff
@@ -24,6 +24,14 @@
 ## Suppress mime type detection in browsers for unknown types
 <IfModule mod_headers.c>
 Header always set X-Content-Type-Options "nosniff"
+##
+# Joomla considers the feature of Federated Learning of Cohorts (FLoC) to be against the
```
PhilETaylor:
```suggestion
# The Joomla! Project considers the feature of Federated Learning of Cohorts (FLoC) to be against the
```

As per the text in COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION, differentiate between The Joomla Project (that is political) and the Joomla CMS code (which is not political)

htaccess.txt
```diff
@@ -24,6 +24,14 @@
...
+# interest of users and site owners. Joomla blocks this feature by default through its
+# code. To block it for all requrests, uncomment the below directive and then disable the
```
PhilETaylor:
```suggestion
# code. To block it for all requests, uncomment the below directive and then disable the
```

htaccess.txt
```diff
@@ -24,6 +24,14 @@
...
+# code in the global configuration of Joomla. Read more about this in the Post-Installation
```
PhilETaylor:
```suggestion
# this feature in the global configuration of Joomla. Read more about this in the Post-Installation
```

libraries/src/Application/CMSApplication.php
```diff
@@ -209,6 +209,34 @@ public function execute()
 			$this->render();
 		}
 
+		if ($this->get('block_floc', 1))
+		{
+			$headers = $this->getHeaders();
+
+			$notPresent = true;
+
+			foreach($headers as $header)
```
rachellawson: (My Joomla knowledge is slightly lacking, I admit, but...) Is iterating across the array the only way to find this header? Seems a lot of processing

PhilETaylor: unfortunately its a non keyed array and therefore it has to be done this way :-(

htaccess.txt
PhilETaylor:
```suggestion
# code in the Global Configuration of Joomla. Read more about this in the Post-Installation
```

libraries/src/Application/CMSApplication.php
PhilETaylor: My point is, that you cannot do a `===` compare here, you have to assume the `Permissions-Policy` could be absolutely any cased string including `perMISSIONS-PoLiCy` which is still perfectly valid.

libraries/src/Application/CMSApplication.php
rachellawson: Ah! Fair enough.

libraries/src/Application/CMSApplication.php
richard67: Sure you can do a `===` 😛 , e.g. `if (strtolower($header['name']) === 'permissions-policy')`.

libraries/src/Application/CMSApplication.php
richard67:
```suggestion
			foreach ($headers as $header)
```
Fix PHPCS.

administrator/language/en-GB/en-GB.com_config.ini
```diff
@@ -93,6 +93,8 @@ COM_CONFIG_FIELD_FILTERS_CUSTOM_BLACK_LIST="Custom Blacklist"
...
+COM_CONFIG_FIELD_FLOC_BLOCKER_LABEL="Block Google FLoC"
```
PhilETaylor:
```suggestion
COM_CONFIG_FIELD_FLOC_BLOCKER_LABEL="Block FLoC"
```
The FloC specification was published by the Web Platform Incubator Community Group. It is not a W3C Standard nor is it on the W3C Standards Track

It is not "Google FloC" - although its authors are Google.

libraries/src/Application/CMSApplication.php
PhilETaylor: Correct, but the code, as proposed in the PR currently, is wrong and the comparison, currently, is not good enough. Maybe my emphasis was lost in communication :)

richard67: Nothing lost, only wanted to tease you a bit with my nit picking.

PhilETaylor:
I propose removing all references to Google. 

The FloC specification was published by the Web Platform Incubator Community Group. It is not a W3C Standard nor is it on the W3C Standards Track - apparently, whatever that means. 

It is not "Google FloC" - although its authors are Google and Google are the first to use it, there is nothing stopping others from using the same tech

libraries/src/Application/CMSApplication.php
HLeithner:
```suggestion
				if (strtolower($header['name']) === 'permissions-policy')
```

brianteeman:
We have a **legal problem** with this entire pull request and its accompanying text.

As has been pointed out above in the comments and the descriptions that accompany the code this is a political decision.

> The Joomla! project disagrees that this feature is in the interests of the owners of Joomla!-powered websites as well as their visitors. 

It doesn't improve Joomla in any way but purely exists to send a statement to Google that the Joomla project objects to this **Federated Learning of Cohorts**.

This creates a problem because OpenSourceMatters, the legal entity responsible for Joomla, passed a motion unanimously that was proposed by the Production Department Co-ordinator, a director of OpenSourceMatters, which specifically prohibits Joomla making a political statement on anything.

> #2020/136 - Joomla is a non-political organisation and inclusive of everybody. In that context any public facing Joomla oriented communication should be void of political or otherwise offensive comments or references.
**Proposed by Marco Dings, seconded by Luca Marzo.**
_The motion passed with unanimous consent._

This code and its accompanying text is clearly a political statement and as such is contrary to the above policy. The original intent may have been for the policy to address something completely different, we don't know as there is no record in the minutes of anything further. But legally it doesn't matter. The only thing that matters is the text of the motion that was proposed by Marco Dings and was approved.

The President of OpenSourceMatters was made aware, at the time that the motion was published in the minutes, that it would cause problems due to its wording and it was suggested that the policy should be either rejected as is or replaced with one that actually covered the issue it intended to address.. Additionally the President of OpenSourceMatters was made aware of this Pull Request when it was submitted and the legal issue was highlighted. 

Unfortunately as neither of those have happened the rules have not changed and it is my belief (i am not a lawyer) that until OpenSourceMatters formally rescinds the above motion this Pull Request cannot be merged as it would be acting against its own rules.

_My personal views about FLoC are irrelevant to the comments above._

htaccess.txt
brianteeman: This should point to documentation somewhere other than the site itself. If the site is not working you cannot read the notes

rachellawson:
> As has been pointed out above in the comments and the descriptions that accompany the code this is a political decision.
> 
> > The Joomla! project disagrees that this feature is in the interests of the owners of Joomla!-powered websites as well as their visitors.

Is it necessary to make a statement of belief at all? The text used in Drupal core is as follows:

```
/** 
 *
 * By default, Drupal sends the 'Permissions-Policy: interest-cohort=()' header
 * to disable Google's Federated Learning of Cohorts feature, introduced in
 * Chrome 89.
 *
 * See https://plausible.io/blog/google-floc for more information about FLoC.
 *
 * If you don't wish to disable FLoC in Chrome, you can set this value
 * to FALSE.
 */
```

No statement of belief, just what the setting does.

PhilETaylor:
> The text used in Drupal core is as follows:

Which links to an article https://plausible.io/blog/google-flo entitled "How to fight back against Google FLoC"... so hardly neutral on the point. :-)

brianteeman:
@rachellawson  The link in the comment is the de facto drupal statement. Drupal hopefully doesn't have such a badly worded policy that we are governed by

administrator/language/en-GB/en-GB.com_admin.ini
```diff
@@ -139,6 +139,8 @@ COM_ADMIN_PLATFORM_VERSION="Joomla! Platform Version"
...
 COM_ADMIN_POSTINSTALL_MSG_BEHIND_LOAD_BALANCER_TITLE="New Server Setting \"Behind Load Balancer\""
+COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="A new feature is currently being rolled out to browsers to replace third party tracking cookies. This features is named Federated Learning of Cohorts (FLoC) and you can read more about it <a href="https://github.com/WICG/floc">here</a>. The Joomla! project believes this feature to be a security issue and against the interests of our users, both the owners of Joomla!-powered sites, as well as their visitors. We do follow the EFF in their <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">assessment</a> and have added a feature to block this technology on all Joomla sites. This feature has been enabled by default. If you disagree with this decision and want to allow FLoC for your website, you can disable it from the Global Configuration. To additionally enable this feature for all other requests to your server, enable it in your .htaccess. In that case you have to disable this feature in the Global Configuration."
```
HLeithner:
```suggestion
COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="A new technology is currently being rolled out to browsers to replace third party tracking cookies. This technology is named Federated Learning of Cohorts (FLoC) and you can read more about it <a href="https://github.com/WICG/floc">here</a> and <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">here</a>. Starting with Joomla! 3.9.27 your website blocks this technology, you can re-allow it from the Global Configuration. Additionally to disable this technology for all requests to your server, you have to update your .htaccess."
```
Simplify the post installation messages.

htaccess.txt
```diff
@@ -24,6 +24,15 @@
 ## Suppress mime type detection in browsers for unknown types
 <IfModule mod_headers.c>
 Header always set X-Content-Type-Options "nosniff"
+##
+# The Joomla! Project considers the feature of Federated Learning of Cohorts (FLoC)
+# to be a security issue and to be against the interest of users and site owners.
+# Joomla blocks this feature by default through its code. To block it for all requests,
+# uncomment the below directive and then disable the this feature in the global
+# configuration of Joomla. Read more about this in the Post-Installation message
+# in the backend.
```
HLeithner:
```suggestion
# Disable Federated Learning of Cohorts (FLoC)
# If you uncomment the below directive you have to allow this technology in the 
# Global Configuration of Joomla. Read more about this in the Post-Installation 
# message in the backend.
```

HLeithner:
This PR is not political motivated we consider this technology as a risk for our user base not more not less. We just disable it as default but still let informed users the possibility to change the setting at anytime.

administrator/language/en-GB/en-GB.com_admin.ini
```diff
@@ -139,6 +139,8 @@ COM_ADMIN_PLATFORM_VERSION="Joomla! Platform Version"
...
 COM_ADMIN_POSTINSTALL_MSG_BEHIND_LOAD_BALANCER_TITLE="New Server Setting \"Behind Load Balancer\""
+COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="A new technology is currently being rolled out to browsers to replace third party tracking cookies. This technology is named Federated Learning of Cohorts (FLoC) and you can read more about it <a href="https://github.com/WICG/floc">here</a> and <a href="https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea">here</a>. Starting with Joomla! 3.9.27 your website blocks this technology, you can re-allow it from the Global Configuration. Additionally to disable this technology for all requests to your server, you have to update your .htaccess."
```
Quy:
```suggestion
COM_ADMIN_POSTINSTALL_MSG_GOOGLE_FLOC_BLOCKER_DESCRIPTION="A new technology is currently being rolled out to browsers to replace third party tracking cookies. This technology is named Federated Learning of Cohorts (FLoC) and you can read more about it <a href=\"https://github.com/WICG/floc\">here</a> and <a href=\"https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea\">here</a>. Starting with Joomla! 3.9.27 your website blocks this technology, you can re-allow it from the Global Configuration. Additionally to disable this technology for all requests to your server, you have to update your .htaccess."
```

richard67:
I have tested this item :white_check_mark: successfully on 1c30880ae5ae446412e71e42b3569d6e80adfae9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 1c30880ae5ae446412e71e42b3569d6e80adfae9<br /><br />besides the wording...<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

richard67:
@alikon Thanks for fixing the milestone. I see I've set the wrong one.

HLeithner:
Thanks

alikon:
i think this needs more fanfare....docs, tweet, marketing ...etc

richard67:
I've discovered an issue. When I've tested this at home on my local Apache server, it worked like described in the comment in htaccess.txt that when I enabled the directive in the .htaccess file, I had to disable the "Block FLoC" in Global configuration in order not to get duplicate headers.

But now I've noticed that on my webspace, which is shared hosting by IONOS (formerly 1and1), it did not work like that. I had to use both, the backend block option and the htaccess directive, to get the header set for the page (i.e. the index.php file) and for real files like css, js, ...

The reason for this is following: On IONOS shared hosting, PHP runs as CGI, and in Apache configuration the "AllowOverride" parameter is set to false. That means that such htaccess directives like the new header or the nosniff header are not applied to PHP files.

So for this special case the comment in the htaccess.txt telling to switch off the blocker option in backend when enabling the htacess directive is wrong.

But to explain that in the comment in htaccess.txt might result in the comment being a bit long, especially if I have to write that explanation.

What shall we do? Make a PR to extend the comment by an explanation of that? Does someone have a proposal for a short enough text for the comment?

PhilETaylor:
This fascination with .htaccess when joomla officially supports IIS and nginx too haha....

richard67:
> This fascination with .htaccess when joomla officially supports IIS and nginx too haha....

@PhilETaylor Is this your proposal for the helping comment in htaccess.txt? 😉

richard67:
Yes, I know, "stay away from shared hosting".

PhilETaylor:
> HLeithner changed the title [3.9][SECURITY] Block usage of FLoC by default [3.9][Privacy] Block usage of FLoC by default yesterday

LMAO - So yesterday it was a SECURITY issue that needed to be flagged as such to break all the rules about new features in Joomla 3, but now its relabelled as a [Privacy] issue after its been added to Joomla 3... thus making this even more controversial.

dmleeman:
Joomla 3.9.26 -> 27
PHP 7.4.16
Litespeed UK Servers


Just installed 3.9.27 with the FLoC Block set on as default and it gnarled an animation that starts on the open of my page. The animation was fine yesterday, but glitched on the update.

It also knocked about 3 to 4% from my GT Metrix performance on the same site.

Rebuilt the cache and recompiled the SCSS, but still misfired the anim, so turned of FLoC and bingo! It's back to normal.

So, no problem with the idea but is there another way of implementing this?

I see a suggestion of using .htaccess, is this a possibility?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

dmleeman:
With FLoC turned on I'm eyeballing a second or so delay on larger sites too.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

richard67:
> I see a suggestion of using .htaccess, is this a possibility?

@dmleeman Yes. You can add (uncommented) section from the htaccess.txt to your .htaccess file:

```
## Suppress mime type detection in browsers for unknown types
<IfModule mod_headers.c>
Header always set X-Content-Type-Options "nosniff"
##
# Disable Federated Learning of Cohorts (FLoC)
# If you uncomment the below directive you have to allow this technology in the 
# Global Configuration of Joomla. Read more about this in the Post-Installation 
# message in the backend.
##
Header always set Permissions-Policy "interest-cohort=()"
</IfModule>
```
(The "nosniff" thing is not related to the FLoC but was added in past by another recommendation, see the postinstall messages in past.)

Then you should switch off the FLoC block in Global Configuration in order not to get that "Permissions-Policy" header duplicated.

There is one exception I've noticed on shared hosting of my provider (IONOS):

They run PHP as CGI and have the "AllowOverride" parameter set to false in the Apache config, so the header directive from .htaccess is not applied to PHP files, e.g. index.php. In such case you need both, the .htacess directive and the Global Configuration option.

brianteeman:
it is concerning that this is having a performance and functionality impact!!

PhilETaylor:
> it is concerning that this is having a performance and functionality impact!!

"Allegedly" ... at this stage. 

`Permissions-Policy` is not a new header (its a renamed Feature Policy header) and so I find it incredible to believe that enabling such a thing would cause "3 to 4% from my GT Metrix performance"

@dmleeman please provide the before and after GTMetrix share urls with us.

dmleeman:
Yup, not too proud of myself on this.

Just realised I had not set my gantry template from Dev to Production on the FLoC enabled test. Also had 3 other site admins open so may have checked the wrong ones. 

Looks ok now I've compared them side by side properly.

Schoolboy error, must be due a day off

PhilETaylor:
Thanks for letting us know. Pleased it was not the Joomla change :-)

Sophist-UK:
I am also an IONOS user. Does anyone know of a Joomla Plugin that allows you to define arbitrary headers and have them included in the HTTP response headers? Ideally allowing you to both set and unset headers. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

molhokwai:
Hello all,
Joomla lover and user since... 2010?
This is **constructive criticism**, so please understand it as such.

After reading the [FLoC explainer](https://github.com/WICG/floc "FloC explainer"), it seems that it increases privacy, by tracking only data relevant for ads through cohorts instead of individuals.
So it seems that it will be better than cookies in terms of privacy and more, cookies which are (obviously) accepted by _everyone_.

Would you care to explain why the the Joomla! project believes that this feature is not in the interests of owners of Joomla!-powered websites and their visitors?

**Thank you**<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

Sophist-UK:
@molhokwai A simple google search for "floc bad" will bring up results from respected sites such as the Electronic Frontier Foundation (eff.com) and Wired.com explaining why FLoC is a bad idea. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33212">issues.joomla.org/tracker/joomla-cms/33212</a>.</sub>

PhilETaylor:
Joomla gives you the option to enable or disable it... regardless of the politics surrounding it.

Sophist-UK:
@PhilETaylor That is true, but the default value is a statement about whether the Joomla team believes that FLoC is a good thing or a bad thing. (And the consensus of privacy organisations worldwide is that it is a bad thing - and even Google has backtracked on it.)

PhilETaylor:
This was debated considerably at the time, many agreed with you, however this is not the place to continue that discussion on the bottom of. a merged PR. 

You have two choices, provide a PR to revert the inclusion of this feature for Joomla 5, or continue talking to the void.

Sophist-UK:
@PhilETaylor I am in agreement that FLoC is a bad thing - so you will not be seeing a PR from me. I was just saying that whilst having an option is non-partisan, the choice of default value is a consensus (i.e. political in the definition you used) statement about the community view of FLoC.

PhilETaylor:
>  the choice of default value is a consensus...statement about the community view of FLoC

I can assure you the "Joomla Community" had no say in the default chosen, and many people actually spoke out about this at the time. The decision was made in a private meeting, and implemented by someone with write access to the repo, and they were hell bent on pursuing it regardless of the opinions of the "Joomla Community".

Sophist-UK:
@PhilETaylor Ahhh - the anarchy of open source where an individual with write access can do what they like and consensus is optional.

I have contributed to many open source projects, and they vary considerably in these areas. I have certainly contributed to projects with good ethics, and also one (a Joomla extension) where the project has effectively been destroyed by being taken over by a company who then decided that open source meant make it available for free but do limited maintenance and refuse all contribution from external users. 

On the whole, my impression of the Joomla team/community is that there is a lot of collaboration and consensus building, but I guess there will always be exceptions.

PhilETaylor:
Its documented above. There was a decision made and a person implemented it. There is no point discussing this further.

The decision was taken by OSM, Inc. - Full Board Meeting - April 22, 2021 https://volunteers.joomla.org/board-of-directors/reports/1495-osm-inc-full-board-meeting-april-22-2021

Hackwar:
I can guarantee you, that I don't have write access to the Joomla repo (and for these exact reasons, I also don't want to have that access.) I brought the issue forward in a meeting of the production team and that escalated that to OSM. I've been vocal about this change and will defend it further, but I did not make the final decision. In the end, every decision has to be made by someone "in power", even if it is just through taking action. However, the teams involved made the decisions based on solid infos and based on feedback from the community.

PhilETaylor:
> However, the teams involved made the decisions based on solid infos and based on feedback from the community.

In private and against the wishes of "the community" actually but other than that... pushed into an existing series as a "security" issue when it has nothing to do with security, breaking Joomla's own rules on semver...

PhilETaylor:
Google still has not even started its testing of FLOC/FLEDGE! So implementing this feature was way premature and could have waited for Joomla 4 after all...

https://www.privacysandbox.com/timeline/
https://searchengineland.com/google-pushes-back-floc-testing-to-q1-2022-374905

PhilETaylor:
> Google is considering a new algorithm and API for improvements from OT feedbacks. See https://datatracker.ietf.org/meeting/111/materials/slides-111-pearg-lets-talk-about-floc-00.
That could be incompatible with the current FLoC so that it is no longer needed to be public now. But it is their choice. 

https://github.com/WICG/floc/issues/133#issuecomment-906001526

PhilETaylor:
Todau FLOC is not even a thing, the trials ended 14 Jul 2021 - the feature is redundant and a waste of bytes anyway

https://developer.chrome.com/origintrials/#/view_trial/213920982300098561

Sophist-UK:
It was not my intent to provoke a flame war. As @PhilETaylor points out, FLoC is already dead in the water, so this feature is nugatory and should be reverted in due course.

PR #33218, opened 2021-04-21
Added plugin enable/disable with AJAX

eopws:
### Summary of Changes

Added new feature to plugins component in administrator panel. Now if user change status of a plugin (enabled/disabled) he doesn't need to press "Save" button, everything is done live with AJAX technology.

### Testing Instructions

In administrator panel go to Extensions > Plugins > choose any plugin and then switch its status. Befor applying these changes you have to press "Save" button, but after applying these changes you are able to see "Plugin disabled" or "Plugin enabled" notification and plugin is really enabled/disabled.

Nothing happens after switching plugin status on Extensions > Plugins > choose any plugin page

A plugin gets enabled/disabled after switching plugin status on Extensions > Plugins > choose any plugin page

### DIFF
--- a/administrator/components/com_plugins/views/plugin/tmpl/edit.php
+++ b/administrator/components/com_plugins/views/plugin/tmpl/edit.php
@@ -38,6 +38,50 @@ JFactory::getDocument()->addScriptDeclaration("
 		}
 	};
 ");
+
+// Add plugin enable/disable AJAX
+JFactory::getDocument()->addScriptDeclaration("
+	jQuery(document).ready(function() {
+		jQuery('#jform_enabled').change(function(e) {
+			let pluginAction = e.target.value;
+			let coverDiv = document.createElement('div');
+
+			jQuery.ajax(
+				'index.php?option=com_plugins&extension_id=" . (int) $this->item->extension_id . "&".JSession::getFormToken()."=1',
+				{
+					cache: false,
+					data: { task: 'ajaxControlPlugin', pluginAction },
+					beforeSend: function() {
+						coverDiv.style.background = 'rgba(13, 13, 13, 0.5)';
+						coverDiv.style.zIndex = '9999';
+						coverDiv.style.height = '100vh';
+						coverDiv.style.width = '100vw';
+						coverDiv.style.position = 'fixed';
+						coverDiv.style.top = '0';
+						coverDiv.style.left = '0';
+
+						document.body.append(coverDiv);
+					},
+					complete: onComplete,
+					success: function(result, status, xhr) {
+						if (pluginAction == 1) {
+							alert('".JText::_( 'COM_PLUGINS_N_ITEMS_PUBLISHED_1')."');
+						} else {
+							alert('".JText::_( 'COM_PLUGINS_N_ITEMS_UNPUBLISHED_1')."');
+						}
+					},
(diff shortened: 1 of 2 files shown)


administrator/components/com_plugins/views/plugin/tmpl/edit.php
```diff
@@ -38,6 +38,50 @@
...
+						coverDiv.style.width = '100vw';
+						coverDiv.style.position = 'fixed';
+						coverDiv.style.top = '0';
+						coverDiv.style.left = '0';
+
+						document.body.append(coverDiv);
+					},
+					complete: onComplete,
+					success: function(result, status, xhr) {
+						if (pluginAction == 1) {
```
dgrammatiko, line 67: No alerts,  pelase use the API `Joomla.renderMessages()`

dgrammatiko (changes requested):

administrator/components/com_plugins/views/plugin/tmpl/edit.php
```diff
@@ -38,6 +38,50 @@
...
+							alert('".JText::_( 'COM_PLUGINS_N_ITEMS_PUBLISHED_1')."');
+						} else {
+							alert('".JText::_( 'COM_PLUGINS_N_ITEMS_UNPUBLISHED_1')."');
```
dgrammatiko, line 70: No alerts, pelase use the API Joomla.renderMessages()

dgrammatiko (changes requested):

administrator/components/com_plugins/views/plugin/tmpl/edit.php
```diff
@@ -38,6 +38,50 @@
 		}
 	};
 ");
+
+// Add plugin enable/disable AJAX
+JFactory::getDocument()->addScriptDeclaration("
+	jQuery(document).ready(function() {
```
dgrammatiko, line 44: Move all this inline JS to it's own file. Pass the only 2 variables that you need from the PHP world using the API: `JFactory::getDocument()->addScriptOptions()`

dgrammatiko (changes requested):

dgrammatiko (changes requested):
Also don't use jQuery.Ajax, use the API: `Joomla.request()`
And if you drop that then you shouldn't need jQuery at all here

dgrammatiko:
Also, Joomla 3 is not accepting new features so I guess you should try this for 4.1 (J4 is also feature freeze atm)

administrator/components/com_plugins/views/plugin/tmpl/edit.php
eopws: Where the file should be stored?

dgrammatiko: `media/com_plugin/js/admin_edit_default.js`

eopws: Thank you

eopws:
@dgrammatiko Thank you for pointing on errors

dgrammatiko:
> Thank you for pointing on errors

These are not errors. I'm just pointing out that you could use Joomla's API and not use jQuery as it's not needed.

PR #33236, opened 2021-04-22
Use null coalescing operator instead of if

eopws:
### Summary of Changes

`if` block replaced with null coalescing operator to reduce code and make it more readable.

### Testing Instructions

Go to administrator/components/com_plugins/src/Model/PluginsModel.php

### DIFF
--- a/administrator/components/com_plugins/src/Model/PluginsModel.php
+++ b/administrator/components/com_plugins/src/Model/PluginsModel.php
@@ -116,10 +116,7 @@ class PluginsModel extends ListModel
 		$ordering = $this->getState('list.ordering', 'ordering');
 
 		// If "Sort Table By:" is not set, set ordering to name
-		if ($ordering == '')
-		{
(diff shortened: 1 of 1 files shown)


richard67:
I don't think this is right. The null coalesce operator returns the first value if it exists and is not null, otherwise the second operator. An empty string is not null, i.e. with this change here, the empty string will be used.

richard67:
Another question is if this PR is right, then why is it made for the 4.1-dev branch? The same code exists also in the 4.0-dev branch, where we also could use null coalesce. Only in J3 we can't use it due to minimum version requirement for PHP being not high enough.

eopws:
> I don't think this is right. The null coalesce operator returns the first value if it exists and is not null, otherwise the second operator. An empty string is not null, i.e. with this change here, the empty string will be used.

I think that's right. Sorry for disturbing. I will close the pull request

Issue #33253, opened 2021-04-22
[4] Joomla stats plugin has no preview of what will be sent


PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/33253

### Summary of Changes

Show preview of stats even if the module is unpublished 

### Testing Instructions

Install Joomla 4 
edit "System - Joomla! Statistics" plugin
Disable the plugin and save it
edit "System - Joomla! Statistics" plugin
Click "Select here to see the information that will be sent."

### Actual result BEFORE applying this Pull Request

When the plugin is disabled there is no data preview

https://user-images.githubusercontent.com/400092/115774707-d3043180-a3a9-11eb-917e-a2d462a84f4e.mp4

### Expected result AFTER applying this Pull Request

a list of system info that will be sent to the stats server, to allow me to make an informed decision to enable this plugin

https://user-images.githubusercontent.com/400092/115774860-02b33980-a3aa-11eb-8c9a-f9596b43abf6.mp4

### Documentation Changes Required

none

PhilETaylor:
https://github.com/joomla/joomla-cms/issues/33253

PR #33255, opened 2021-04-22
[4] Show stats preview when unpublished

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/33253

### Summary of Changes

Show preview of stats even if the module is unpublished 

### Testing Instructions

Install Joomla 4 
edit "System - Joomla! Statistics" plugin
Disable the plugin and save it
edit "System - Joomla! Statistics" plugin
Click "Select here to see the information that will be sent."

When the plugin is disabled there is no data preview

https://user-images.githubusercontent.com/400092/115774707-d3043180-a3a9-11eb-917e-a2d462a84f4e.mp4

a list of system info that will be sent to the stats server, to allow me to make an informed decision to enable this plugin

https://user-images.githubusercontent.com/400092/115774860-02b33980-a3aa-11eb-8c9a-f9596b43abf6.mp4

none

### DIFF
--- a/plugins/system/stats/src/Field/DataField.php
+++ b/plugins/system/stats/src/Field/DataField.php
@@ -13,6 +13,7 @@ namespace Joomla\Plugin\System\Stats\Field;
 
 use Joomla\CMS\Factory;
 use Joomla\CMS\Plugin\PluginHelper;
+use Joomla\Registry\Registry;
 
 /**
  * Unique ID Field class for the Stats Plugin.
@@ -52,8 +53,35 @@ class DataField extends AbstractStatsField
 
 		$result = Factory::getApplication()->triggerEvent('onGetStatsData', array('stats.field.data'));
 
-		$data['statsData'] = $result ? reset($result) : array();
+		// If the plugin is not published, we need to compile the stats manually
+		if (!$result)
+		{
+			$stats         = Factory::getApplication()->bootPlugin('stats', 'system');
+			$stats->params = new Registry(\json_decode($this->getParams()));
+			$result = [$stats->onGetStatsData('')];
+		}
+
+		$data['statsData'] = reset($result);
 
 		return $data;
+	}
+
+	/**
+	 * If the plugin is not published, we need to compile the stats manually, for that we need the params from the db.
+	 *
+	 * @return  string   The JSON of the params of this plugin.
+	 */
+	private function getParams()
+	{
+		$db = Factory::getContainer()->get('DatabaseDriver');
+
+		$query = $db->getQuery(true)
+			->select('params')
+			->from('#__extensions')
+			->where($db->quoteName('name') . ' = ' . $db->quote('plg_system_stats'));
+
+		$db->setQuery($query);
+
+		return $db->loadResult();
 	}
 }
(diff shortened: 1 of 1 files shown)


plugins/system/stats/src/Field/DataField.php
```diff
@@ -52,8 +53,35 @@ protected function getLayoutData()
 
 		$result = Factory::getApplication()->triggerEvent('onGetStatsData', array('stats.field.data'));
 
-		$data['statsData'] = $result ? reset($result) : array();
+		// If the plugin is not published, we need to compile the stats manually
```
brianteeman:
```suggestion
		// If the plugin is disabled, we need to compile the stats manually
```

plugins are enabled/disabled not published

plugins/system/stats/src/Field/DataField.php
```diff
@@ -52,8 +53,35 @@ protected function getLayoutData()
...
+			$result = [$stats->onGetStatsData('')];
+		}
+
+		$data['statsData'] = reset($result);
 
 		return $data;
 	}
+
+	/**
+	 * If the plugin is not published, we need to compile the stats manually, for that we need the params from the db.
```
brianteeman:
```suggestion
	 * If the plugin is disabled, we need to compile the stats manually, for that we need the params from the db.
```

ghost:
I have tested this item :white_check_mark: successfully on 1670a0025a5e7db18ac36fb4e1d7f31944539237<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33255">issues.joomla.org/tracker/joomla-cms/33255</a>.</sub>

Fedik:
Sorry, the fix is to hacky, and will show a bad example for other developers. 

I think, here would be better to show the message that explain to User what is going on, when plugin is disabled:
`The preview not available for unpublished plugin, please do "this" and "this"...`.

Or just hide whole "preview section" for unpublished plugin.

PhilETaylor:
> Sorry, the fix is to hacky, and will show a bad example for other developers.

You mean like much of the code in Joomla...

PhilETaylor:
The problem is that the dataField has no knowledge of whether the plugin is disabled or not... so another hack needed to find that information...

Fedik:
> You mean like much of the code in Joomla...

Sorry, I did not meant to offend someone.

> The problem is that the dataField has no knowledge of whether the plugin is disabled or not

There `PluginHelper::isEnabled()`.

Or maybe this does not need, just improve the message, around:
`The data not available, there is no data or the plugin is disabled and so data collection also disabled. Please check if plugin enabled`
Or something like that, I am not strong in English.

PhilETaylor:
`PluginHelper::isEnabled()` doesnt do what it says it does. I already wasted time on that. I'll just hide everything if there is no stats returned (no stats returned = the plugin is disabled so the triggerEvent doesnt return stats)

Fedik:
> PluginHelper::isEnabled() doesnt do what it says it does. 

I just checked on j3 `var_dump(JPluginHelper::isEnabled('system', 'stats'))` return correct `true`/`false`, when `enabled`/`disable`
Cannot check on j4 currently, but the code is the same there.

PhilETaylor:
checking if its disabled was not the problem anyway... 

replacement https://github.com/joomla/joomla-cms/pull/33261

PR #33256, opened 2021-04-22
[4.0] On updating Joomla, "Overridden files have changed" message should be linked to allow resolution.

brianteeman:
Pull Request for Issue ##33173.

### Summary of Changes
Adds an additional message which contains the link to the templates component

### Testing Instructions
I honestly have no idea


particthistle:
Yes, not an easy core one to test. Need to create a test component to install with an override to trigger something... or find a need to modify a template item in Cassiopeia.

@Anu1601CS do you have any test components that you had floating around from when you created the override check and comparison functionality?

plugins/installer/override/override.php
```diff
@@ -188,11 +189,14 @@ public function getOverrideCoreList()
 	 */
 	public function finalize($result)
 	{
-		$num = count($result);
+		$num  = count($result);
+		$url  = Uri::base() . 'index.php?option=com_templates&view=templates';
```
PhilETaylor: If $url and $link are only needed if the $num is not 0 then I would move them into the if statement, so they only run and take up the fraction of time, if ever needed.

PhilETaylor:
Might want to update the PR title as "first commit" doesnt really explain this :)

plugins/installer/override/override.php
```diff
@@ -188,11 +189,13 @@ public function getOverrideCoreList()
...
 	public function finalize($result)
 	{
-		$num = count($result);
+		$num  = count($result);
 
 		if ($num != 0)
 		{
-			$this->app->enqueueMessage(Text::plural('PLG_INSTALLER_OVERRIDE_N_FILE_UPDATED', $num), 'notice');
+			$url  = Uri::base() . 'index.php?option=com_templates&view=templates';
+			$link = HTMLHelper::_('link', $url, Text::_('COM_TEMPLATES'), 'class="alert-link"');			$this->app->enqueueMessage(Text::plural('PLG_INSTALLER_OVERRIDE_N_FILE_UPDATED', $num), 'notice');
```
richard67:
```suggestion
			$link = HTMLHelper::_('link', $url, Text::_('COM_TEMPLATES'), 'class="alert-link"');
			$this->app->enqueueMessage(Text::plural('PLG_INSTALLER_OVERRIDE_N_FILE_UPDATED', $num), 'notice');
```

richard67:
Regarding testing: How about using patchtester? Install version 4.0.0, make some override for it, and then update to 4.0.1?

brianteeman:
@richard67 Might work. I was trying to test it with core last night but you cant because the file was replaced in the core update. Will test again myself later today

richard67:
> Regarding testing: How about using patchtester? Install version 4.0.0, make some override for it, and then update to 4.0.1?

Hmm, no, doesn't work. No overrides can be made for the patchtester. Have to find another one.

plugins/installer/override/override.php
```diff
@@ -13,6 +13,7 @@
 use Joomla\CMS\Date\Date;
 use Joomla\CMS\Language\Text;
```
richard67, line 15:
```suggestion
use Joomla\CMS\HTML\HTMLHelper;
use Joomla\CMS\Language\Text;
```

richard67:
My review suggestion above fixes that.

richard67:
@brianteeman I've just tested with a 3rd party admin module. Both the link added with this PR as well as the Quickicon for the overrides check point to site templates, but they should point to admin templates in case of an admin module.

Do you think this can be fixed for your link with this PR here, and then for the Quickicon with another PR?

Or should I ignore that and give this PR a good test result?

richard67:
For testing I have used https://www.phoca.cz/download/category/123-phoca-top-menu-module . There is a glitch with the Override Check not related to this PR: I did not get the alert modified by this PR when updating that module from 4.0.4. to 4.0.5. So I have uninstalled that extension after I had created the override, and then installed it again, and then I could see that alert.

brianteeman:
I would say thats unrelated. but a genuine issue with the override check

richard67:
> I would say thats unrelated. but a genuine issue with the override check

@brianteeman Does that refer to my last comment about the glitch? Then I agree with you. Or does it refer to my comment before that one about link always going to site templates, also for admin modules? In this case I have to think about it.

richard67:
In the `#__template_overrides` table in database is information about the template, so it should be possible to check if it's a site or an admin template. But maybe that should be done with another, future PR, at all places, including the one being added here.

brianteeman:
The one about the site templates. If the quickicon also goes there (and I suspect it does) then its a bigger bug than addressed here

brianteeman:
If this PR goes to the same url as the quickicon then this PR is correct and anything else should be its own issue

richard67:
I have tested this item :white_check_mark: successfully on 320fa885f0c53ee2f1ce521e512cf69edda6a766<br /><br />Tested with use of https://www.phoca.cz/download/category/123-phoca-top-menu-module .

There is a glitch with the Override Check not related to this PR: I did not get the alert modified by this PR when updating that module from 4.0.4. to 4.0.5. So I have uninstalled that extension after I had created the override, and then installed it again, and then I could see that alert.

The link in the alert added by this PR has the same target as the Quickicon for the Override Checks has. Both have the problem pointing to site templates in any case, also if the override was made in an admin template e.g. for an admin module.

So this PR is consistent with the Quickicon.

The other issues mentioned above need to be fixed separately.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33256">issues.joomla.org/tracker/joomla-cms/33256</a>.</sub>

richard67:
I have not tested this item.<br /><br />Sorry, my previous test was wrong.

The quickicon correctly points to admin templates if the override was made in an admin template, and so should do the link added by this PR.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33256">issues.joomla.org/tracker/joomla-cms/33256</a>.</sub>

richard67:
Hmm, now after having logged out and in again, the Quickicon points to site templates, which is wrong again.

richard67:
The more I think about it, the less I know what's the right way. What if one has overrides in site and in admin templates, where should that link go to?

I think this PR is ok so far as it solve the issue it refers to and is an improvement, and other glitches have to be solved separately.

richard67:
I have tested this item :white_check_mark: successfully on 320fa885f0c53ee2f1ce521e512cf69edda6a766<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33256">issues.joomla.org/tracker/joomla-cms/33256</a>.</sub>

richard67:
It seems that both the link added here and the quickicon go to either admin or site templates depending on which of these you have visited the last time before in your session, i.e. if you once have been in the admin templates view and now use the link or the quickicon, you get to the admin templates, but after logout and login again it goes to site templates. I have some vague memory that this was subject of some issue or some comment in some issue or PR, but I am not sure. Maybe someone else remembers better if that is a known issue?

brianteeman:
It was for editing modules

plugins/installer/override/override.php
```diff
@@ -188,11 +190,14 @@ public function getOverrideCoreList()
 	 */
 	public function finalize($result)
 	{
-		$num = count($result);
+		$num  = count($result);
```
Quy:
```suggestion
		$num = count($result);
```

PhilETaylor: This space should be removed as suggested

brianteeman:
What is the output when language debug is NOT enabled?

Quy:
33256-no-debug

brianteeman:
Closing as the more I look at it i dont think this is the correct approach

Issue #33019, opened 2021-04-05
[4.0] autoload_psr and ftp


brianteeman:
I dont have the ability to test this but I am fairly certain that the recent change to not store the ftp password will mean that on a site that needs to use ftp to write files it will never be able to update the autoload_psr file

PhilETaylor:
I vote for the FTP layer to die.

brianteeman:
so do I but Nik and a few others insisted it was needed

PhilETaylor:
[4] Allow in-memory autoload_psr4.php #33263

brianteeman:
closing this as its documented elsewhere now

Issue #31982, opened 2021-01-10
[4b6] Warnings & subsequent site inconsistency when autoload_psr4.php unwritable


PhilETaylor:
### Steps to reproduce the issue

The concept of a autoload_psr4.php has been introduced in Joomla 4. 

This file is stored at /administrator/cache/autoload_psr4.php

This file, if not-exists, is created on first interaction with Joomla, and regenerated at times, like on installing a Joomla extension/template/plugin.

If this file is not writable, or the folder not writable, by Joomla then a warning is generated, and no helpful text explaining that Joomla is now in an odd an inconsistent state as any newly added classes are not in the autoloader. Depending on the extension installed, this could have knock on effects. Joomla doesn't (probably correctly) automatically keep trying to update the autoload_psr4.php on subsequent page loads, as it just loads it and uses it, regardless of the fact it could be missing classes.

Why would autoload_psr4.php not be writable you may be asking? Well. Because. Internet. Users. Webhosts. [Etc...](https://github.com/joomla/joomla-cms/issues/29424) See Additional Comments below for my case today, however this is not as uncommon as I make out. Many web servers run different users for FTP and for the web server. Many web hosts have permissions issues generally. Just look at the forum and you will see that. In a perfect world Joomla would always have access to write to its files. but this is an imperfect world. 

How you can replicate this issue manually: 

 - Interact with your newly installed Joomla 4 site
 - Now that you have a /administrator/cache/autoload_psr4.php file, make it unwritable by changing permissions and/or ownership manually
 - Attempt to install Akeeba Backup (A rock solid Joomla 4 compatible extension at this time). 
 - Note the warning.
 - Note that the /administrator/cache/autoload_psr4.php has not been updated

### Expected result

I don't have all the answers. 

At the very least we should not be showing PHP warnings to the user.

At the best, we should be checking for the ability to write to `administrator/cache/autoload_psr4.php` BEFORE attempting to undertake a task (such as installing an extension) that will require the regeneration of the autoloader and writing of the file. 

At the extreme we could check the consistency of the content of autoload_psr4.php periodically. 

### Actual result

`
Warning: file_put_contents(/pathto/public_html/administrator/cache/autoload_psr4.php): failed to open stream: Permission denied in /pathto/public_html/libraries/src/Filesystem/File.php on line 384
`

[...]

PR #31857, opened 2021-01-05
[4.0] Re-Implementing FTP-layer during CMS installation

Hackwar:
Pull Request for Issue  #20774 .

### Summary of Changes
Joomla 3.x (and below) supports installing on a system without write permissions for the apache user. This means, that we either provide FTP credentials during installation to write the configuration.php or you get the configuration.php displayed at one point and need to create that file manually. This feature is missing in Joomla 4 right now.

This PR adds that back in. When the installation can't write to the filesystem (specifically the root folder of the installation or an existing configuration.php), it redirects to a preinstall screen, where you can provide the FTP credentials. Below are 2 buttons, one to check the credentials and one to skip this step. There is also a button next to the path field to auto-discover the correct path of the installation.

Clicking the auto-discover button triggers an AJAX request which takes the FTP credentials and tries to connect with those. It then will fill that field. Clicking the "Verify FTP Settings" button checks if all given info is correct and we are now able to write to that folder. "Manually create file" skips this step and later displays the content for the configuration.php.

When displaying the content for the configuration.php, there is a button "Continue", which will check if the configuration.php has been created and then redirects to the "remove" page.

### Testing Instructions
1. Get a setup (preferably on Unix), where only your FTP user has write access to the folder, none else.
2. Clone this repo and try the installation. Before this PR, you can only progress with the installation when fixing the write permissions. Afterwards you can also install with FTP credentials.

### Notes
I'm more than welcome for any PRs to improve the language strings here.

PR #33263, opened 2021-04-23
[4] Allow in-memory autoload_psr4.php

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/31982 and https://github.com/joomla/joomla-cms/issues/33019 and https://github.com/joomla/joomla-cms/pull/31857#issuecomment-824986563

### Summary of Changes

On page load, and on demand, Joomla 4 will attempt to write autoload_psr4.php to the JPATH_CACHE folder.

If that folder is unwritable for any reason (many reasons, incorrect permissions, doesnt exist, or needs a FTP Layer to be writable) then Joomla will crash and cannot be used, doesnt display any helpful error, shows a 500 Internal Server Error page. Its a dead end.

This PR, catches the failure of the creation of a physical file, and caches the mapping of the namespace to the path in memory and logs (if enabled) a note to the Joomla log to report the failure of file creation.

Of course it would be more performant to not then have to regenerate the map each page load, but equally, we dont want Joomla to "crash" just because it cannot write an important file. 

If the JPATH_CACHE folder exists and is writable, then the code in this PR is never used and full performance is kept, thus making this backward compatible.

### Testing Instructions

Many different ways, but quickest is to install 
 - install Joomla 4
 - Login to admin
 - DELETE /administrator/cache/autoload_psr4.php
 - chmod 000 /administrator/cache folder
 - Reload the page - See error messages
 - repeat this with debug/error reporting on in Joomla Global Config to see the other messages
 - Apply this PR
 - Repeat the process, note that Joomla doesnt "crash". With error reporting high then you might see a warning about file not being written which is not suppressed (outside the scope of this PR) 

totally unusable Joomla if JPATH_CACHE path is unwritable. 

totally stable Joomla if JPATH_CACHE path is unwritable. 

none.

### DIFF
--- a/libraries/namespacemap.php
+++ b/libraries/namespacemap.php
@@ -131,6 +131,16 @@ class JNamespacePsr4Map
 		}
 
 		$content[] = '];';
+		
+		/**
+		 * Backup the current error_reporting level and set a new level
+		 *
+		 * We do this because file_put_contents can raise a Warning if it cannot write the fautoload_psr4.php file
+		 * and this will output to the response BEFORE the session has started, causing the session start to fail
+		 * and ultimately leading us to a 500 Internal Server Error page just because of the output warning, which
+		 * we can safely ignore as we can use an in-memory autoload_psr4 map temporarily, and display real errors later.
+		 */
+		$error_reporting  = error_reporting(0);
 
 		if (!File::write($this->file, implode("\n", $content)))
 		{
@@ -154,6 +164,9 @@ class JNamespacePsr4Map
 
 			$this->cachedMap = $map;
 		}
+		
+		// Restore previous value of error_reporting
+		error_reporting($error_reporting);
 	}
 
 	/**
--- a/libraries/namespacemap.php
+++ b/libraries/namespacemap.php
@@ -135,7 +135,7 @@ class JNamespacePsr4Map
 		/**
 		 * Backup the current error_reporting level and set a new level
 		 *
-		 * We do this because file_put_contents can raise a Warning if it cannot write the fautoload_psr4.php file
+		 * We do this because file_put_contents can raise a Warning if it cannot write the autoload_psr4.php file
 		 * and this will output to the response BEFORE the session has started, causing the session start to fail
 		 * and ultimately leading us to a 500 Internal Server Error page just because of the output warning, which
 		 * we can safely ignore as we can use an in-memory autoload_psr4 map temporarily, and display real errors later.
--- a/libraries/namespacemap.php
+++ b/libraries/namespacemap.php
@@ -131,7 +131,7 @@ class JNamespacePsr4Map
 		}
 
 		$content[] = '];';
-		
+
 		/**
 		 * Backup the current error_reporting level and set a new level
 		 *
@@ -164,7 +164,7 @@ class JNamespacePsr4Map
 
 			$this->cachedMap = $map;
 		}
-		
+
 		// Restore previous value of error_reporting
 		error_reporting($error_reporting);
 	}
--- a/libraries/namespacemap.php
+++ b/libraries/namespacemap.php
@@ -140,7 +140,7 @@ class JNamespacePsr4Map
 		 * and ultimately leading us to a 500 Internal Server Error page just because of the output warning, which
 		 * we can safely ignore as we can use an in-memory autoload_psr4 map temporarily, and display real errors later.
 		 */
-		$error_reporting  = error_reporting(0);
+		$error_reporting = error_reporting(0);
 
 		if (!File::write($this->file, implode("\n", $content)))
 		{
--- a/libraries/namespacemap.php
+++ b/libraries/namespacemap.php
@@ -10,6 +10,7 @@ defined('_JEXEC') or die;
 
 use Joomla\CMS\Filesystem\File;
 use Joomla\CMS\Filesystem\Folder;
+use Joomla\CMS\Log\Log;
 
 /**
  * Class JNamespacePsr4Map
@@ -25,6 +26,12 @@ class JNamespacePsr4Map
 	 * @since  4.0.0
 	 */
 	protected $file = JPATH_CACHE . '/autoload_psr4.php';
+
+	/**
+	 * @var array|null
+	 * @since __DEPLOY_VERSION__
+	 */
+	private $cachedMap = null;
 
 	/**
 	 * Check if the file exists
@@ -90,7 +97,7 @@ class JNamespacePsr4Map
 			$this->create();
 		}
 
-		$map = require $this->file;
+		$map = $this->cachedMap ?: require $this->file;
 
 		$loader = include JPATH_LIBRARIES . '/vendor/autoload.php';
 
@@ -125,7 +132,28 @@ class JNamespacePsr4Map
 
 		$content[] = '];';
 
-		File::write($this->file, implode("\n", $content));
+		if (!File::write($this->file, implode("\n", $content)))
+		{
+			Log::add('Could not save ' . $this->file, Log::WARNING);
+
+			$map = [];
+			$constants = ['JPATH_ADMINISTRATOR', 'JPATH_API', 'JPATH_SITE', 'JPATH_PLUGINS'];
+
+			foreach ($elements as $namespace => $path)
+			{
+				foreach ($constants as $constant)
+				{
+					$path = str_replace($constant,  constant($constant) , $path);
+				}
+
+				$path = str_replace(["'", " ", "."], "", $path);
+				$namespace = str_replace('\\\\', '\\', $namespace);
+
+				$map[$namespace] = [ $path ];
+			}
+
+			$this->cachedMap = $map;
+		}
 	}
 
 	/**
--- a/libraries/namespacemap.php
+++ b/libraries/namespacemap.php
@@ -143,7 +143,7 @@ class JNamespacePsr4Map
 			{
 				foreach ($constants as $constant)
 				{
-					$path = str_replace($constant,  constant($constant) , $path);
+					$path = str_replace($constant, constant($constant), $path);
 				}
 
 				$path = str_replace(["'", " ", "."], "", $path);


anibalsanchez:
I have tested this item :white_check_mark: successfully on 417d0f896e224048fff425a44253096f986ddcf8<br /><br />Test OK<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33263">issues.joomla.org/tracker/joomla-cms/33263</a>.</sub>

pabloarias:
I have tested this item :red_circle: unsuccessfully on 417d0f896e224048fff425a44253096f986ddcf8<br /><br />After chmod 000 /administrator/cache and applying the path, Joomla is still unusable.

I see in frontend Casiopeia header but a message "The requested page can't be found".

Backend is unusable too: error 500.

Maybe Am I doing something wrong?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33263">issues.joomla.org/tracker/joomla-cms/33263</a>.</sub>

PhilETaylor:
@pabloarias Thanks for testing. 

You need to test with Debug = Off and Error Reporting = None in Joomla Global Config. This is the "normal user live site" configuration.

If you have debug on, or error reporting set high (development) then you will correctly get the 500 Internal Server Error message and the  (new) red screen of death.

pabloarias:
I'm testing now with Debug = Off and Error Reporting = None.

I obtain in frontend "The requested page can't be found".

Backend is still unusable with 500 error.

Sorry, something else wrong? Thank you very much.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33263">issues.joomla.org/tracker/joomla-cms/33263</a>.</sub>

PhilETaylor:
ok that is because you HAVE a /administrator/cache/autoload_psr4.php file but it cannot be read. That is out of scope of this PR which seeks to prevent if Joomla cannot write to the folder, not read from it (and this PR is part of the FTP Layer fixes later on)

So do this:

```
chmod 777 /administrator/cache/
rm /administrator/cache/autoload_psr4.php 
chmod 000 /administrator/cache/
```

and then you will be able to test this PR and the admin console will work

PhilETaylor:
Screenshot 2021-04-23 at 20 44 18

libraries/namespacemap.php
```diff
@@ -131,6 +131,16 @@ protected function writeNamespaceFile($elements)
...
+		
+		/**
+		 * Backup the current error_reporting level and set a new level
+		 *
+		 * We do this because file_put_contents can raise a Warning if it cannot write the fautoload_psr4.php file
+		 * and this will output to the response BEFORE the session has started, causing the session start to fail
+		 * and ultimately leading us to a 500 Internal Server Error page just because of the output warning, which
+		 * we can safely ignore as we can use an in-memory autoload_psr4 map temporarily, and display real errors later.
+		 */
+		$error_reporting  = error_reporting(0);
```
richard67:
```suggestion
		$error_reporting = error_reporting(0);
```

PhilETaylor: damn you are quick :) that was caused by copy and paste into GitHub... more to come.

joomdonation:
As you mentioned, it would cause performance issue because each page load, the system will have to read, parse xml files from all our extensions to create namespace map and users won't know about this.

To me, I would let the error happens so that users will know and fix this error instead. Just my personal option.

PhilETaylor:
Yes it really is suboptimal, however without it Joomla! cannot be installed using FTP layer because it cannot write files until after it has asked for the credentials for FTP, and in order to do that it needs to boot which attempts to create this file.

I don't think it is optimal that Joomla! should crash just because it cannot write one file to the hard disk - this PR provides a workaround 

99.999% of sites will never use this code, will never use in memory version of this file, because modern professional web hosting will allow the file to be written￼￼￼￼￼￼ - but this PR or handle the edge case when it cannot be written￼

joomdonation:
@wilsonge will make decision, I just express my concern :). Owners of these website could complain why my sites are slow, bla bla... :D.

PhilETaylor:
Well if the decision to keep the FTP Layer doesn't change - then this is the only way, because without this file Joomla cannot boot, and without booting Joomla cannot install over FTP because it cannot ask for the FTP Credentials. 

Their sites would ONLY be slow if they had a READ ONLY file system that did not allow the file writes - for example if they [did not store their FTP Credentials in Joomla Global Configuration](https://github.com/joomla/joomla-cms/issues/33304) and then used (other currently broken) features of Joomla (Like extension install over FTP) that made a change that needed the namespacemap to change, and Joomla could not write the file. 

Its an edge case. No one in the real world would ever experience a slow site due to this PR handling the edge case.

wilsonge:
Ok. So this is a fallback to cache not being writeable. So I'm ok with this PR from that perspective.

To address your concerns @joomdonation :

I think we need to catch the exception from adding the log entry (if the cache isn't writeable there's a decent chance the logs aren't writeable either!)

I think we need to get better at having logs for admins for the site maintainers - not adding the messages into the UI. If an end user sees this it's useless to them... even if that comes with us needing to add a way to view site logs in the admin ui

wilsonge:
I didn't mean to merge - was on complete autopilot *doh*. @PhilETaylor can you do an additional PR catching a potential exception from Log::add please

PhilETaylor:
As logging should be viewed as a nice to have I would say that inability to write a log file should not hinder a user with an exception (unless debug mode is on)

The reason for this PR, in the FTP layer section of fixes, would probably mean that logs could not be written either, and also we do not yet have a configuration so we could not have debug mode on, so we want to gracefully allow the loading of the installation page so that we can finally ask for the credentials for the FTP layer.

If we go about throwing an exception because we can't write the log because we can't write the name space file then we are no further along in our quest to install joomla￼￼￼￼

PhilETaylor:
Sorry, that was a long way to say yes I will look at it further tomorrow￼

PhilETaylor:
Following [the projects decision to remove the FTP Layer](https://github.com/joomla/joomla-cms/pull/33390#issuecomment-830000434) from Joomla 4 - this could possibly be reverted.

richard67:
> Following [the projects decision to remove the FTP Layer](https://github.com/joomla/joomla-cms/pull/33390#issuecomment-830000434) from Joomla 4 - this could possibly be reverted.

@wilsonge What do you think?

wilsonge:
This seems like a fairly safe fallback. In this case we would allow things like the error pages to render I suspect - whereas without this we'll spectacularly crash and burn. So I'd leave this one as it is only because it's so early in the execution stack

SharkyKZ:
This does not work in its current state.

PhilETaylor:
> This does not work in its current state.

Care to actually elaborate? or just trolling?

PhilETaylor:
- git checkout 4.0-dev
- rm administrator/cache/autoload_psr4.php
- chmod 000 administrator/cache
- Load Joomla - it works!

SharkyKZ:
Try on a real hosting environment. And stop with the personal attacks! I don't troll, ever.

PhilETaylor:
You are a troll. Period. 

I have tested this PR (Actually the Nightly Zip from https://developer.joomla.org/nightly-builds.html which has this PR already merged into it on the two largest hosting platforms, Plesk and cPanel and it works perfectly with the /administrator/cache folder set to 000

 - wget https://developer.joomla.org/nightlies/Joomla_4.0.0-beta8-dev-Development-Full_Package.zip
 - unzip -o https://developer.joomla.org/nightlies/Joomla_4.0.0-beta8-dev-Development-Full_Package.zip
 - rm administrator/cache/autoload_psr4.php
 - chmod 000 administrator/cache
 - Load Joomla Installer - it works!
 - Install Joomla - it works! 
 - Login to Joomla Admin - It works!

And to prove I actually did this - here is the live site, on cPanel hosting, http://sharkykz.phil-taylor.com/ running with /administrator/cache with 000 permissions and unwritable.

Screenshot 2021-05-02 at 16 50 42


So you state:

> This does not work in its current state.

And I ask you again:

> Care to actually elaborate? or just trolling?

Would you care to actually tell me why you believe this PR doesn't work in hosting environments, when all the testing proves it does?

WHAT EXACTLY are you trying to report is wrong with it?

SharkyKZ:
Going by code review, my guess this is the culprit https://github.com/joomla/joomla-cms/pull/33263/files#diff-34e6fec5937ad1928863d81b93501ed9d0e5ddb3a607da8bc0d91f3611ca18d7R159. Why strip out random characters? Since when are these not valid in paths?

PhilETaylor:
> Why strip out random characters? Since when are these not valid in paths?

`$path` incoming to my changed code code is a string of

`"JPATH_ADMINISTRATOR . '/components/com_actionlogs/src'"`

We were then replacing the constant with the correct folder structure:

This would give a string of something like 

`"/application/public_html/administrator . '/com_actionlogs/src'"`

The reason for the `str_replace` to remove spaces, the concated period, and a single quote at the end of the string to give 

`/application/public_html/administrator/com_actionlogs/src`

This obviously works, and so simply saying "This does not work in its current state" is simply untrue and unhelpful. 

> Since when are these not valid in paths?

Of course `[spaces, periods, single quotes]` are valid (albeit stupid) in folder structures. 

My mistake was to remove the `[spaces, periods, single quotes]` from the WHOLE absolute path, and not just the Joomla folder structure, which as we know, has zero folders with `[spaces, periods, single quotes]` in their paths. 

> Try on a real hosting environment

I did. Unlike you who by your own admission just did a "code review" and provided no helpful feedback until repeatedly requested from you. 

I have access to the absolute paths of 67,348 Joomla sites, from Joomla 1.5.0 to Joomla 4.0-beta7

I have checked the absolute paths and found only two sites that has either a period or a space in their JPATH_BASE. 

One site is in a folder called /home/username/restored from Dec.2014/ 
One site is in a folder called /home/username/My Documents/Sites/ 

Therefore, your assumption that "This does not work in its current state" currently effects 2 out of a random sample of 67,348 sites. 0.0000297%.

However, of course, you are right, 0.0000297% right, and as promosed this has been fixed and improved upon and, in absense of a PR from you, I have proposed PR https://github.com/joomla/joomla-cms/pull/33504

That PR was tested with a JPATH_BASE of `/application/My Folder.Old/` and `chmod 000 /administrator/cache` and it works perfectly by `preg_replace` to replace the constant, the contcatenation, and the trailing single quote in one go. 

Its not pretty or performant, but I fully expect this code never to be executed in the real world, it was meant to allow us to get to the first page of the Installation process on a server that required the FTP Layer to write tot he database (as, at the first page load we yet dont have FTP Credentials) - since then the Joomla PLT have decided in favour of removing the FTP Layer. I proposed the immediate reverting of this PR, and George explained he was happy to keep it merged. Then you showed up days later.

Thank you for (eventually) verbalising your objection instead of just dismissing others hard work with thumbs down and saying "it doesnt work" - that is unhelpful. As you can see, if you take the time to explain your thinking then I, and others, can improve the code we have proposed.

Issue #33251, opened 2021-04-22
[4] FTP Layer Release Blocker Issues Summary


PhilETaylor:
### Issues

- [ ] [4][FTP Layer] Optionally saving FTP credentials creates impossible file operation scenarios #33304
- [ ] [4][FTP Layer] When FTP Layer enabled, Joomla NEVER caches pages #33249 @PhilETaylor 
- [ ] [4][FTP Layer] Folder Permissions screen incorrect states folders are unwritable. #33245 @PhilETaylor 
- [ ] [4][FTP Layer] Cannot create new template files #33246 @PhilETaylor 
- [ ] [4][FTP Layer] Cannot upload images in Media Manager #33244 @PhilETaylor 
- [ ] [4][FTP Layer][SECURITY] FTP Credentials displayed to the screen in plain text on Upload & update page #33242 @PhilETaylor 
- [ ] [4][FTP Layer] Cannot install extensions at all #33241 @PhilETaylor 
- [ ] [4][FTP Layer] Cannot save template file modifications #33239 @PhilETaylor 
- [ ] [4][FTP Layer] use of is_writable when FTP Layer needed causes false error messages #33238 @PhilETaylor 
- [ ] [4][FTP Layer] Filesystem based Sessions do not work. #33237 @PhilETaylor 
- [ ] FTP doesnt work #33079 @brianteeman
- [x] [4.0] autoload_psr and ftp #33019  @brianteeman
- [ ] [3] ERROR: Wrong FTP host or port #28199 @twister65

### Pull Requests to fix things in FTP Layer

- [x] [4][FTP Layer] JLIB_CLIENT_ERROR_FTP_BAD_RESPONSE #33284 @PhilETaylor 
- [x] [4] Use CMS library which is FTP Layer aware #33283 @PhilETaylor 
- [ ] [4.0] Fix Update Joomla Using FTP #33281 @joomdonation
- [ ] [4.0] FTP error message #33123 @brianteeman
- [ ] [4.0] Re-Implementing FTP-layer during CMS installation #31857
- [ ] [4.0] com_installer FTP #33077 @Hackwar
- [ ] [4] Ensure opcache file invalidate on file copy & delete #32915 (If we change files by FTP, we also need to clear PHP Opcache for those files! ) @PhilETaylor 
- [x] [4] Allow in-memory autoload_psr4.php #33263 @PhilETaylor 

### Pull Requests proposed to REMOVE the FTP layer

[...]

Issue #28199, opened 2020-03-03
ERROR: Wrong FTP host or port


twister65:
### Steps to reproduce the issue

1. Joomla Update -> Upload & Update

2. Choose an update package file.

3. Installation method -> Write files using FTP

4. Set the FTP options.

### Expected result
Update your Joomla files via FTP.

### Actual result
![wrong_ftp](https://user-images.githubusercontent.com/26852967/75794139-ab4ea600-5d70-11ea-8649-d424399e08e3.png)

### System information (as much as possible)
Joomla! Version 3.9.15
PHP Version 7.4.3
Web Server Apache/2.4.41 with APACHE_RUN_USER corresponding to my ftp_login in  /etc/apache2/envvars

### Additional comments
This issue is solved in PR #28029 (Joomla4) using FTP options in the global configuration.
This PR also fixes another issue with the FTP connection function.

brianteeman:
If you are saying that it does NOT currently work in joomla 3 at all and we have had zero reports of this then it would suggest that no one is actually using it.

Or am I missing something

ReLater:
I can confirm the issue with Joomla 3 nightly build of today.

twister65:
@brianteeman , like PostgreSQL 🤔

navonwolf:
A report to recovery the error [io games](https://iogames.onl)[skribbl io](https://iogames.onl/skribbl-io-games)
  
  <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/28199">issues.joomla.org/tracker/joomla-cms/28199</a>.</sub>

PhilETaylor:
Are we going to fix FTP issues in Joomla 3 Following [the projects decision to remove the FTP Layer](https://github.com/joomla/joomla-cms/pull/33390#issuecomment-830000434) from Joomla 4?

brianteeman:
Thank you for raising this issue.

Joomla 3 is now in security only mode with no further bug fixes or new features.

As this issue doesn't relate to Joomla 4 it will now been closed.

If we are mistaken and this does apply to Joomla 4 please open a new issue (and reference this one if you wish) with updated details for testing in Joomla 4.
cc [@zero-24](https://github.com/zero-24)

PR #33281, opened 2021-04-24
[4.0] Fix Update Joomla Using FTP

joomdonation:
Pull Request for Issue # .

### Summary of Changes
During update process (using FTP or Hybrid installation method), Joomla! needs to write FTP credentials into  restoration.php file so that these data can be used by restore.php on restoration process.

Due to a bug in getting FTP credentials (trying to get directly from input but these fields are not available in the input at all), update Joomla using these two methods never works (or at least it never works in the recent version). Also, the code in restore.php file does not like the path with window style, so I added some code to change it to Linux path style

Tested on my local Joomla 4 using Filezilla server and it worked. Not sure how to give testing instructions because now it is hard to find testers for FTP stuffs.

### Testing Instructions

### DIFF
--- a/administrator/components/com_joomlaupdate/restore.php
+++ b/administrator/components/com_joomlaupdate/restore.php
@@ -1824,7 +1824,7 @@ class AKPostprocFTP extends AKAbstractPostproc
 		}
 
 		// Try to download ourselves
-		$testFilename = defined('KSSELFNAME') ? KSSELFNAME : basename(__FILE__);
+		$testFilename = defined('KSSELFNAME') ? KSSELFNAME : 'index.php';
 		$tempHandle   = fopen('php://temp', 'r+');
 		if (@ftp_fget($this->handle, $tempHandle, $testFilename, FTP_ASCII, 0) === false)
 		{
--- a/administrator/components/com_joomlaupdate/src/Controller/UpdateController.php
+++ b/administrator/components/com_joomlaupdate/src/Controller/UpdateController.php
@@ -278,6 +278,18 @@ class UpdateController extends BaseController
 		// Did a non Super User tried to upload something (a.k.a. pathetic hacking attempt)?
 		Factory::getUser()->authorise('core.admin') or jexit(Text::_('JLIB_APPLICATION_ERROR_ACCESS_FORBIDDEN'));
 
+		$method = $this->app->getUserStateFromRequest('com_joomlaupdate.method', 'method', 'direct', 'cmd');
+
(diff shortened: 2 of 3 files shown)


brianteeman:
Doesn't the same problem exist in Joomla 3? At least from your description of the reason it would appear so.

joomdonation:
Yes, same happens with Joomla 3. I just made the PR here quickly to prove that updates Joomla never works properly (at least for recent release, maybe from the time we added Upload and Update option back, I don't remember the time)

So the point here is if update is not working and no-one reports it, can we assume that no-one is using it? We can try to spend time to fix FTP issues, but if no one is using it, it will be a waste of time

brianteeman:
> So the point here is if update is not working and no-one reports it, can we assume that no-one is using it? We can try to spend time to fix FTP issues, but if no one is using it, it will be a waste of time

That's what I've been trying to say but people keep sidetracking with conversations about code purity

joomdonation:
Hope this PR shows something. If update Joomla not working, if install extensions not working (as I saw somewhere from list of issues which Phil mentioned) and no-one complains, we should not have to spend time to fix it.

PhilETaylor:
> we should not have to spend time to fix it.
>  it will be a waste of time

I think Im going to log off now :) :) 

I'll add this PR to the list at https://github.com/joomla/joomla-cms/issues/33251

>  because now it is hard to find testers for FTP stuffs.

Because very few people even understand the root problem, and others view it as a server "misconfiguration", a "botched installation" of a server instead of a genuine type of server configuration (from 15 years ago)...

PhilETaylor:
Is this related to this Joomla 3 issue? https://github.com/joomla/joomla-cms/issues/28199

joomdonation:
> Is this related to this Joomla 3 issue? #28199

Yes, I think so.

PhilETaylor:
Drone died but unrelated

richard67:
Drone resuscitated 😄 .

joomdonation:
Closing because FTP layer will be dropped.

richard67:
> Closing because FTP layer will be dropped.

@joomdonation Who says that?

PhilETaylor:
Following [the projects decision to remove the FTP Layer](https://github.com/joomla/joomla-cms/pull/33390#issuecomment-830000434) from Joomla 4

joomdonation:
@richard67 See https://github.com/joomla/joomla-cms/pull/33390#issuecomment-830000434

PR #33282, opened 2021-04-24
[5.0] Enforce Strict parsing for both article and category alias …

Ruud68:
Pull Request for Issue #32490 .

### Summary of Changes
The router used in 3.9 and used in 4.0 allows for Search Engine URL forgeries (see #32490) for the impact of this (business) vulnerability.
This PR enforces the ID and Alias in the URL to match. If no match, a 404 will be produced (same as in the Modern Router)

This PR is a redo of #32500 (it now also checks the category alias), after several attempted PR's by @PhilETaylor and discussions on these PRs (#32880 #33270 #32879 #32490 and probably missing some) .

In the discussions the preference was shown to fix this in the router, which this PR does.

### Testing Instructions
In article option tab integration switch Remove IDs from URLs to 'No'
_Testing instructions are for com_content, but should be repeated with com_contact and com_newsfeeds_

[A] Test:
1. open on the front-end [yourdomain]/index.php/blog/5-your-modules
2. open on the front-end [yourdomain]/index.php/blog/5-dont-do-any-business-with-me-because-i-suck

Install this PR
[B]Repeat test above

[A]: 1 > Your Modules page in displayed
[A]: 2 > Your Modules page is displayed (note: the non-existing / fake URL will be indexed by Google because it resolves okay and as such will show in your sites search results)

[B]: 1> Your Modules page is displayed
[B]: 2 > a 404 is thrown which is the correct behavior as this page doesn't exist and should NEVER resolve okay!

cc: @PhilETaylor @Hackwar


Hackwar:
There is no legacy router in 4.0 anymore. And as I said, please remove the config option for this. Either we enforce this all the time or we don't add this at all. We also need this for every core component, not just com_content.

Ruud68:
> There is no legacy router in 4.0 anymore. And as I said, please remove the config option for this. Either we enforce this all the time or we don't add this at all. We also need this for every core component, not just com_content.

There is a reason I made the toggle as when you come from 3.9 > 3.10 > 4.0 there is a incompatibility in the build URL's (switching to with ID in J4 will not fix that). This toggle gives the user a way out of not being able to migrate. current default is 'Loose', but that could of course also be 'Strict'

As for the other components: yes of course, first 'proof' this 'concept' before putting a lot of time in redoing etc.

Hackwar:
Again, this change here has no influence on that build part. People will have to come up with a migration strategy for their URLs anyway, regardless of your change here. The alias should be identical in both 3.x and 4.0.

Ruud68:
> Again, this change here has no influence on that build part. People will have to come up with a migration strategy for their URLs anyway, regardless of your change here. The alias should be identical in both 3.x and 4.0.

The alias is identical, the rest isn't always identical. I am trying to make the migration to 4 as easy as possible, I believe this switch helps, but I agree that when moving forward people should switch their sites to the 'modern' with NO id's URLs. IMO They should have done that already when the modern router was introduced.

That said: I will remove it (will make the code even simpler :)), but know that without this toggle certain URL's that resolved okay before this PR will then give a 404

Hackwar:
For the category part, you can simply replace that one check from `$child->id == (int) $segment` to `$child->id . ':' . $child->alias == $segment`

Ruud68:
> For the category part, you can simply replace that one check from `$child->id == (int) $segment` to `$child->id . ':' . $child->alias == $segment`

PR is updated now

components/com_content/src/Service/Router.php
```diff
@@ -273,6 +273,25 @@ public function getArticleId($segment, $query)
 
 			return (int) $this->db->loadResult();
 		}
+		else
```
Hackwar: No else necessary here.

components/com_content/src/Service/Router.php
```diff
@@ -273,6 +273,25 @@ public function getArticleId($segment, $query)
...
+					]
+				)
+				->bind(':id', $id, ParameterType::INTEGER)
+				->bind(':alias', $alias);
+			$this->db->setQuery($dbquery);
+
+			return (int) $this->db->loadResult();
+		}
 
 		return (int) $segment;
```
Hackwar: This has to go then.

Hackwar (changes requested):

Hackwar:
Please also add the other components here as well.

Ruud68:
> Please also add the other components here as well.

done

HLeithner:
I'm not sure with this hardcore version. IIRC the last version had an option to redirect wrong alias to the correct uri.
I would really like to see this version as an option. I'm not sure how many urls get changed with the "new" router. So it would better to have the possibility to redirect a wrong alias to the correct right alias.

Ruud68:
> I'm not sure with this hardcore version. IIRC the last version had an option to redirect wrong alias to the correct uri.
> I would really like to see this version as an option. I'm not sure how many urls get changed with the "new" router. So it would better to have the possibility to redirect a wrong alias to the correct right alias.

I have removed it as requested by @Hackwar
as for the url changing, I have tested upgrading huge site (90000 articles, 1300 categories) from 3.9 (with legacy router) to 3.10 to 4.0. Even when 4.0 is set with ID in the URl, the URLs change. So either way, this is going to be potentially 'dramatic' to migrate.
When doing the redirects to correct this when upgrading to 4.0 then people are best of to do the 'extra' step to disable the ID in the URL altogether.

@Hackwar can you advise?

HLeithner:

Hackwar (approved):

Hackwar:
I looked through this again and I fear we are to late to merge this into 4.x. This is going to be a change which will still break some peoples sites and thus we can't merge it into 4.2 or 4.3. So unfortunately 5.0 it is.

PR #33283, opened 2021-04-24
[4] Use CMS library which is FTP Layer aware

PhilETaylor:
### Summary of Changes

Logrotation currently uses the Joomla framework Filesystem package which is not FTP Layer aware. 

This leads to issues when log rotation attempts to move a file when it has no permissions to do so, when FTP is needed. 

This change starts using the Folder and File packages from the CMS and not the framework, although `Folder` doesn't yet need to be FTP aware, Im changing that so that if we ever did make Folder FTP Aware (hint: probably soon coming!) then it will "just work" in future. 

### Testing Instructions

Code review, unless you have a spare hour to follow along the whole FTP Layer set up.

After installation, and clocking on the Open Site button on the last page of the installation wizard

Screenshot 2021-04-24 at 17 42 51

Use the CMS Filesystem lib, which is FTP aware, and can move the file with FTP. 

none

### DIFF
--- a/plugins/system/logrotation/logrotation.php
+++ b/plugins/system/logrotation/logrotation.php
@@ -12,8 +12,8 @@ defined('_JEXEC') or die;
 use Joomla\CMS\Cache\Cache;
 use Joomla\CMS\Factory;
 use Joomla\CMS\Plugin\CMSPlugin;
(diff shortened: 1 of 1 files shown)


wilsonge:
Thanks!

Issue #33368, opened 2021-04-27
[4.0] Fields, Field Groups: Hide "Users"-Dropdown as only one value is selectable


ghost:
### What needs to be fixed
As title say, in J3 was no Dropdown.

Joomla 4 | Joomla 3
---- | ----
![Bildschirmfoto 2021-04-27 um 10 32 53](https://user-images.githubusercontent.com/78906980/116212746-74420d80-a745-11eb-8ef9-c7de76f2e335.png) | ![Screenshot_2021-04-27 Users Fields - sandramay0905 - Administration](https://user-images.githubusercontent.com/78906980/116212783-7e640c00-a745-11eb-99b8-e5870e010279.png)

### Why this should be fixed
A dropdown isn't needed if user cannot select another option.

### How would you fix it
Hide the dropdown.

### Side Effects expected
Don't know.

ChristineWk:
Similar to this? #33300<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33368">issues.joomla.org/tracker/joomla-cms/33368</a>.</sub>

ghost:
@ChristineWk i guess not, it can be like in joomla 3 as there are fields only for users.

rjharishabh:
PR #33371

ghost:
Close as having pull request.

PR #33373, opened 2021-04-27
[4.0] Field context selector only if there are more than one contexts

chmst:
Pull Request for Issue #33368 .

### Summary of Changes
As title says.

### Testing Instructions
Apply the patch and check that in all components the field context selector appears only if there are at least 2 options.

Fields overview in com_users has a context selector with only one option.

Fields overview in com_users has no context_selector


joomdonation:
@chmst Guess we should close this and merge the other one to make our contributors happy :).

PR #33374, opened 2021-04-27
[4] Remove unused file

PhilETaylor:
Removing an unused file, as this project doesn't want to take code quality seriously.


brianteeman:
this should be closed. its just a petulant  pr

PhilETaylor:
This file is unused. It is not used by the Joomla project or the the drone/appveyour. It serves no purpose in this project.

joomdonation:
Came down @PhilETaylor . You know that we care about code quality, we just want to do a bit later to avoid merge conflicts as I mentioned somewhere.

PhilETaylor:
> we just want to do a bit later to avoid merge conflicts as I mentioned somewhere.

Same excuse as 5 years ago.

PhilETaylor:
I dont see why this was closed. This file is unused. It is not used by the Joomla project or the the drone/appveyour. It serves no purpose in this project's repository.

Please tell me why you are keeping this file?

joomdonation:
It was created, so it was used. Maybe it is not used recently, but that doesn't mean we won't use it in the future, maybe when after Joomla 4 releases. There is no reason to delete it now, then create it again when we need to use it.

Issue #33396, opened 2021-04-28
Joomla 4 - startTabSet() in some circumstances can return void


BrainforgeUK:
### Steps to reproduce the issue

Component view html making use of Bootstrap tabs.

### Expected result

startTabSet() to return a string or a raise a sensible exception.

### Actual result

Get this message on screen:
Return value of Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet() must be of the type string, none returned

### System information (as much as possible)

PHP 7.4

### Additional comments

Before closing braces should either return string or raise a sensible exception.

PhilETaylor:
What is the full content of your call including the selector and params? Are you calling the method twice with the same selector and params?

PhilETaylor:
Fixed in https://github.com/joomla/joomla-cms/pull/33397

BrainforgeUK:
Not tried fix, but problem in component was due to _sometimes (!)_ startTabSet was being called twice with the same selector and params.

PhilETaylor:
That was the problem I identified and debugged, my notes are in the PR. Thanks for confirming.

PhilETaylor:
Thanks for reopening 

The agreement made in #33397 is that this is probably a "Wont fix" and that developers should know not to use the same id multiple times.

However the thing we will fix is to die gracefully with either an exception thrown, or just to ignore the fact that you asked for a duplicate element with the same id - instead of a PHP error message which is not good enough really for Joomla.

PhilETaylor:
Please test PR https://github.com/joomla/joomla-cms/pull/33409

alikon:
please test #33409

PR #33397, opened 2021-04-28
[4] Fix calling startTabSet twice causes PHP Error

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/33396

### Summary of Changes

 Fix calling startTabSet twice causes PHP Error

### Testing Instructions

Run startTabSet twice with the same input params

```
\Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet('asd', []);
\Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet('asd', []);

```

Screenshot 2021-04-28 at 21 22 35

One set of output:

```html
<ul class="joomla-tabs nav nav-tabs" id="asdTabs" role="tablist"></ul>
<div class="tab-content" id="asdContent">
```

none

### DIFF
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -809,25 +809,24 @@ HTMLSTR;
 	 */
 	public static function startTabSet($selector = 'myTab', $params = []) :string
 	{
-		$sig = md5(serialize([$selector, $params]));
-
-		if (!isset(static::$loaded[__METHOD__][$sig]))
+		if (isset(static::$loaded[__METHOD__][$selector]))
 		{
-			// Setup options object
-			$opt['active'] = (isset($params['active']) && ($params['active'])) ? (string) $params['active'] : '';
-
-			// Initialise with the Joomla specifics
-			$opt['isJoomla'] = true;
-
-			// Include the Bootstrap Tab Component
-			HTMLHelper::_('bootstrap.tab', '#' . preg_replace('/^[\.#]/', '', $selector), $opt);
-
-			// Set static array
-			static::$loaded[__METHOD__][$sig] = true;
-			static::$loaded[__METHOD__][$selector]['active'] = $opt['active'];
-
-			return LayoutHelper::render('libraries.html.bootstrap.tab.starttabset', ['selector' => $selector]);
+			return '';
 		}
+
+		// Setup options object
+		$opt['active'] = (isset($params['active']) && ($params['active'])) ? (string) $params['active'] : '';
+
+		// Initialise with the Joomla specifics
+		$opt['isJoomla'] = true;
+
+		// Include the Bootstrap Tab Component
+		HTMLHelper::_('bootstrap.tab', '#' . preg_replace('/^[\.#]/', '', $selector), $opt);
+
+		// Set static array
+		static::$loaded[__METHOD__][$selector]['active'] = $opt['active'];
+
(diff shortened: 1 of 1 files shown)


PhilETaylor:
Im not sure this is the right fix though... because this would create two elements with the same ID on the page. 

I think the right fix would be to throw an Exception because the Selector has already been used... hmmm.

PhilETaylor:
Im not convinced this is the full correct fix. Sure it fixes the PHP issue which was reported, but the developer (DX) experience is bad. 

The root cause is a developer trying to use the same `$selector` for two different Tab Sets. Which is not allowed. 

Paging @dgrammatiko who might have more ideas.

The problem is, is if a developer reuses the same selector more than once, we dont really want this PHP error, but equally we dont want an awkward missing tabs... 

For example 

```php

echo \Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet('asd', []);
echo \Joomla\CMS\HTML\Helpers\Bootstrap::addTab('asd', 'myid','mytitle');
echo 'aa aa aa aa';
echo \Joomla\CMS\HTML\Helpers\Bootstrap::endTab();
echo \Joomla\CMS\HTML\Helpers\Bootstrap::addTab('asd', 'myid2','mytitle2');
echo 'bb bb bb bb';
echo \Joomla\CMS\HTML\Helpers\Bootstrap::endTab();
echo \Joomla\CMS\HTML\Helpers\Bootstrap::endTabSet();


echo \Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet('asd', []);
echo \Joomla\CMS\HTML\Helpers\Bootstrap::addTab('asd', 'myid11','mytitle');
echo 'aaa aaa aaa aaa';
echo \Joomla\CMS\HTML\Helpers\Bootstrap::endTab();
echo \Joomla\CMS\HTML\Helpers\Bootstrap::addTab('asd', 'myid22','mytitle2');
echo 'bbb bbb bbb bbb';
echo \Joomla\CMS\HTML\Helpers\Bootstrap::endTab();
echo \Joomla\CMS\HTML\Helpers\Bootstrap::endTabSet();
```

would give this with this PR:


https://user-images.githubusercontent.com/400092/116470902-eca2ee80-a86b-11eb-9427-5806768de260.mp4

dgrammatiko:
@PhilETaylor there's some background here: https://github.com/joomla/joomla-cms/pull/32146 and https://github.com/joomla/joomla-cms/pull/32357

In short, although we tried to not break the B/C for tabs and modals factories we assumed that the old code was sufficient enough and there weren't any further improvements. That said I think devs should make sure that their element ids are totally unique, eg use a pattern like `company-product-specifier` which should guarantee uniqueness. I don't think that the core should go into this rabbit hole of ensuring the uniqueness of ids, it's not the core's responsibility it's down to the dev understanding what they're instructing the framework to do...

PhilETaylor:
> . I don't think that the core should go into this rabbit hole of ensuring the uniqueness of ids, it's not the core's responsibility 

But ... Im not happy leaving a PHP error, we would not do that anywhere else in Joomla

Can we throw a nice exception instead if someone uses the same id twice? or just ignore it and carry on silently outputting crap HTML that then its up to the developer to debug and fix?

dgrammatiko:
> Can we throw a nice exception instead if someone uses the same id twice?

That would be ok but manipulating the id to "fix" it magically is a route that the core should not follow.

PhilETaylor:
Awesome thanks. I'll add this back to my todo pile :)

PhilETaylor:
Replaced with  #33409

PR #33409, opened 2021-04-29
[4] Silently return early if this method has already been called for the given selector

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/33396 replacement for https://github.com/joomla/joomla-cms/pull/33397

@dgrammatiko @brainforgeuk

### Summary of Changes

Reemploying the same tactic that the `startAccordion` [method uses to return early](https://github.com/joomla/joomla-cms/blob/7d02ff86b956064e1d12bc386f61ec28069c7633/libraries/src/HTML/Helpers/Bootstrap.php#L665) and ignore the fact that the method has been called incorrectly, with the same selector, twice by a developer. 

Prevents outputting a PHP error message. 

Please read the background here https://github.com/joomla/joomla-cms/pull/33397#issuecomment-829137212

For element ID's a safe strategy is to use a pattern like `company-product-specifier` which should guarantee uniqueness

### Testing Instructions

add the code below somewhere on a page you are looking at, refresh and see the error.

```php
\Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet('asd', []);
\Joomla\CMS\HTML\Helpers\Bootstrap::startTabSet('asd', []);
```

Screenshot 2021-04-29 at 17 57 14

The page loads without a PHP error, albeit the developer still needs to fix their code as the second call to `startTabSet` doesn't actually do anything, but silently returns. 

This is by design, because of the history explained in the background here https://github.com/joomla/joomla-cms/pull/33397#issuecomment-829137212

None.

### DIFF
--- a/libraries/src/HTML/Helpers/Bootstrap.php
+++ b/libraries/src/HTML/Helpers/Bootstrap.php
@@ -811,23 +811,26 @@ HTMLSTR;
 	{
 		$sig = md5(serialize([$selector, $params]));
 
-		if (!isset(static::$loaded[__METHOD__][$sig]))
+		if (isset(static::$loaded[__METHOD__][$sig]))
 		{
-			// Setup options object
-			$opt['active'] = (isset($params['active']) && ($params['active'])) ? (string) $params['active'] : '';
-
-			// Initialise with the Joomla specifics
-			$opt['isJoomla'] = true;
-
-			// Include the Bootstrap Tab Component
-			HTMLHelper::_('bootstrap.tab', '#' . preg_replace('/^[\.#]/', '', $selector), $opt);
-
-			// Set static array
-			static::$loaded[__METHOD__][$sig] = true;
-			static::$loaded[__METHOD__][$selector]['active'] = $opt['active'];
-
-			return LayoutHelper::render('libraries.html.bootstrap.tab.starttabset', ['selector' => $selector]);
+			// Silently return early if this method has already been called for the given selector
+			return '';
 		}
+
+		// Setup options object
+		$opt['active'] = (isset($params['active']) && ($params['active'])) ? (string) $params['active'] : '';
+
+		// Initialise with the Joomla specifics
+		$opt['isJoomla'] = true;
+
+		// Include the Bootstrap Tab Component
+		HTMLHelper::_('bootstrap.tab', '#' . preg_replace('/^[\.#]/', '', $selector), $opt);
+
+		// Set static array
+		static::$loaded[__METHOD__][$sig] = true;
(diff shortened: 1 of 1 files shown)


libraries/src/HTML/Helpers/Bootstrap.php
```diff
@@ -811,23 +811,26 @@ public static function startTabSet($selector = 'myTab', $params = []) :string
 	{
 		$sig = md5(serialize([$selector, $params]));
 
-		if (!isset(static::$loaded[__METHOD__][$sig]))
+		if (isset(static::$loaded[__METHOD__][$sig]))
 		{
-			// Setup options object
-			$opt['active'] = (isset($params['active']) && ($params['active'])) ? (string) $params['active'] : '';
+			// Silently return early if this method has already been called for the given selector
+			return '';
```
dgrammatiko, line 817: If you return an empty string here you will end up with broken HTML (a surplus of `</div>`. You might want to either return directly
```html
<ul class="joomla-tabs nav nav-tabs" role="tablist"></ul> <!-- this is probably useless --> 
<div class="tab-content">
```

or adjust the layout `libraries.html.bootstrap.tab.starttabset` to echo the opening div.

PhilETaylor: or like you say, leave it to the developer to stop doing crap they should not be doing :-) 

This PR just implements the same as `startAccordian` does. So if there is a bigger fix you would like to implement in exchange for this PR then go for it :)

dgrammatiko: > So if there is a bigger fix you would like to implement in exchange for this PR then go for it :)

Nah, this is fine 🤣

PhilETaylor: It was fine for `startAccordian` ;-)

dgrammatiko: Yes this should alert the dev that probably something is wrong in their output. If you don't mind can you add in the description my comment from the other PR: 
```
For element ID's a safe strategy is to use a pattern like `company-product-specifier` which should guarantee uniqueness
```

Quy:
I have tested this item :white_check_mark: successfully on c7d4c1ff68c1fe60aebb7bd13f31f58124c4ced6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33409">issues.joomla.org/tracker/joomla-cms/33409</a>.</sub>

PhilETaylor:
Not tested by original reporting @BrainforgeUK and no other interest so closing this now.

Issue #33418, opened 2021-04-29
[4.0] Behaviour plugins


brianteeman:
### Steps to reproduce the issue
J4 has two new plugins - behaviour tags and behaviour versions that have a description of
Allows ~content~ items to be tagged.
Allows ~content~ items to be versioned.

### Expected result
~1. From the description I expect that disabling the plugins will remove the ability to version or tag a content item but that it still is possible to version or tag another item such as a contact.~
2. If I can't version an item then I would expect not to see a versions button in the toolbar or a version note field
3. If I did add a version note and then save I would expect to get some sort of message instead of the note being silently discarded
4. If I can't tag an item then I would expect not to see the tag field where I can select an existing tag
5. If I can tag an item and then save  I would expect to get some sort of message instead of the tag being silently discarded

### Actual result
~1. The plugins apply everywhere not just content~
2. version button displayed and active 
3. version note field displayed but cannot be updated
4. tags field is displayed where it appears that I can select a tag or even create a new one

### Why?
Why do we even have these plugins
If we do have these plugins I would expect them to be applied per component and configured in the integration tab of the component
~At an absolute minimum the description of the plugin should be updated to show that they apply to all items not just content items~

_Updated after #33420_

PhilETaylor:
> Why do we even have these plugins

The plugins only provide themselves the the onTable* hooks for Tables that implement the `TaggableTableInterface` or `VersionableTableInterface`

In order for these features to work 1) The plugins must be published 2) The PHP Classes need to implement the interfaces.

The only tables that implement these interfaces, and therefore have these features are: 

Component | VersionableTableInterface | TaggableTableInterface
----------- | ----------- | ----------- 
Content | Yes | Yes
Banners | Yes | No
Banner Clients | Yes | No
Contacts | Yes | Yes
Newsfeeds | Yes | Yes
Tags | Yes | No
User Notes | Yes | No
Categories | Yes | Yes

As you can see this is not the full list of Joomla components, I guess its a half implemented set of features.

brianteeman:
> As you can see this is not the full list of Joomla components, I guess its a half implemented set of features.

Well it wouldnt make sense to tag a tag ;)

None of this explains why its a plugin or explains the broken behaviour described

> The use of the word "content" in the phrase "Allows content items to be tagged" is applied to mean "The royal content" and not as in "Content = Articles"

Easy enough for us who speak english. I wonder how many translated sites will maintain that distinction

[...]

PR #33422, opened 2021-04-30
[4] Hide versions/tags features in Articles if plugins disabled

PhilETaylor:
Pull Request for part of Issue https://github.com/joomla/joomla-cms/issues/33418

### Summary of Changes

 if you disable the Taggable/Versionable plugins the features should "disappear" from the editing of an items that support those interfaces (see table below)

Versionable = Remove Versions Toolbar button, and version note form field
Taggable = Removes the Tags selector in the right hand site
If disabled, On save nothing is done with the version or tags. 

This is not my prettiest work, but is a stepping stone and a jump better than it was integrated.

Component | VersionableTableInterface | TaggableTableInterface | Frontend Also
----------- | ----------- | -----------  | ----------- 
Content | Yes | Yes | Yes
Banners | Yes | No | No
Banner Clients | Yes | No |No
Contacts | Yes | Yes | Yes
Newsfeeds | Yes | Yes | No
Tags | Yes | No | No
User Notes | Yes | No | No
Categories | Yes | Yes | No
All other extensions | No | No | No


brianteeman:
Almost.

I forgot to document that there is an enable/disable versions option in the component options as well. 
image

PhilETaylor:
I believe that option is also taken into account by my code as that is this part:

```php
$this->state->params->get('save_history', 0)
```

I have updated this PR to also remove these features on the frontend edit form

brianteeman:
the field is still displayed in the options when the plugin is disabled

PhilETaylor:
Ah now I understand. You want that option to be removed if the conditions are not met for it.

brianteeman:
Well it would confuse people if they set it to enabled there and it doesnt work ;)

PhilETaylor:
Im struggling to find an existing field that we currently do this for, I dont really want to have to invent or reinvent if you know of another field that magically disappears?

brianteeman:
from memory multilingual associations might

brianteeman:
That last change is not going to work as it is making the version history configurable only globally and not per component

PhilETaylor:
No what it achieved was putting it in EVERY extensions config - which was unintended,

ok Im going to give up on the Configuration screens. 

Looking under the hood this is all so badly implemented that its a complete mess.

There is no way to accurately determine which components are compatible with versioning. Including reusing the feature on custom extensions. 

The ArticleTable doesn't even implement VersionableTableInterface the way the other extensions do! 

Some extensions have the "Enable versions" on "Editing layout" tab of their config, others have it else where. If I move it to the Integrations tab that works great - unless there are already fieldsets there and then Joomla refuses to add more fieldsets.

I can revert the last change, which will get us something that is better than what we started with - but overriding the com_config forms  - in a reusable generic way - looks like its going to take all weekend of refactoring.

PhilETaylor:
I was so close...

PhilETaylor:
Versionable/Taggable are per TABLE, not per component, but yet the configuration is per component /facepalm

PhilETaylor:
ok Brian I made the impossible possible. (For Articles at the moment, and part of banners as a test) 

I have removed The two fields (Enable Versions, Max Versions) from the configuration of com_content in Joomla Global Config -> Articles -> Edit Layout.

I have removed The two fields (Enable Versions, Max Versions) from the configuration of com_banners in Joomla Global Config -> Banners -> History

Now, the configuration fields are part of the Versionable Plugin. Once the Versionable plugin is enabled, the fields will then show in the Joomla Global Config page for the component either 1) On the Integration Tab if it exists (Like in Articles) or 2) on a new tab (like in banners) 

This now means this is reusable too. 

Eventually I can do this for all the extensions that support Versionable in core (not many of them!) for for today this 2 min job has taken hours and I need a break. 

Screenshot 2021-04-30 at 14 29 11
Screenshot 2021-04-30 at 14 29 04

PhilETaylor:
Ok I think this is near complete now. Disabling the Behaviour plugins disables the features, and hides the features. The Tags and Versions config in the options screens also moved to the Integration tab

Quy:
Please change the legend to be `Versions`.

PhilETaylor:
There was a reason for this but I cannot remember what it was now. I will take a look again in the morning. I think it was something to do with the banner options page that doesn't have an integration tab But I will work it out again tomorrow￼￼

PhilETaylor:
The reason was, if there is no Integration Tab in Options (Like in Banner Options) then it uses that for the new Tab name. 

I guess this is ok to show Versions, as it will be the only option on that new tab it automatically creates. 

Screenshot 2021-05-02 at 16 01 09

Whoever thought forms, as XML, was a good idea should be shot :)

PhilETaylor:
I've been struggling to present this PR, its a million times better than the behaviour reported by Brian, but its still no where near a perfect abstraction of behaviours. 

What would be needed, would be to literally remove versions and tags, and rewrite it from scratch in a fully abstracted way, to the point where the plugins could be even deleted without effecting the site. 

Although this PR achieves its aim to remove the fields from components if the plugins are disabled, I still feel that the behaviour are far too coupled to the components, Ive done the best job with what was already there, but with a handful of new events that could be triggered (for example, for toolbars, passing them through triggered events) a better, cleaner abstraction of the behaviour could be done. 

The other issue is that the Tags/Version Note fields are not always in the same place in the edit forms, and so there is a little bit of hard coding already going on to place them in the place the designed wanted them (Prior to this PR)

PhilETaylor:
It also seems my stalker is back... after not learning his lesson @rdeutz 

Screenshot 2021-05-02 at 16 10 40

SharkyKZ:
This is not the correct implementation. Components should not be checking if the plugin is enabled.

PhilETaylor:
> This is not the correct implementation. Components should not be checking if the plugin is enabled.

Like I said.

Issue #33445, opened 2021-04-30
[4] Filter Tags can kill categories overview 


chmst:
### Steps to reproduce the issue
You should have one or more categories and tags (install blog sample data).
Open categories. Open Filter and filter for a tag. If nothing is found, the "New-Page" appears. 
No chance to get back to categories overview. 

### Expected result
If nothing is found, I expect a message as usual 

### Actual result
Result: The "New-Page" appears. 
No chance to come back to categories overview.

rjharishabh:
@chmst  Am I doing something wrong to reproduce the issue

https://user-images.githubusercontent.com/61203226/116749957-59121f00-aa1f-11eb-87b7-2878bf39603a.mp4

brianteeman:
Its a bug but hardly a release blocker

brianteeman:
@PhilETaylor @bakual This is a bug in the emptystate code for categories

Bakual:
I think this has to be fixed here:
https://github.com/joomla/joomla-cms/blob/6f98135cb5a64f354c97edd567bf3abdac83661d/administrator/components/com_categories/src/Model/CategoriesModel.php#L503-L510

Either clear also the "join", or maybe better just start with a new query instead of manipulating the existing one.

Bakual:
Btw same issue exists for other views where you can filter by tags (or probably any joined table property).

PhilETaylor:
ok ok ok give me a moment.

PhilETaylor:
Please test [4] remove join from isEmptyState calculation sql #33453

PhilETaylor:
This can be closed, and certainly is not a release blocker.

Quy:
Thank you!

PR #33453, opened 2021-04-30
[4] remove join from isEmptyState calculation sql

PhilETaylor:
Fixes https://github.com/joomla/joomla-cms/issues/33445

### DIFF
--- a/administrator/components/com_categories/src/Model/CategoriesModel.php
+++ b/administrator/components/com_categories/src/Model/CategoriesModel.php
@@ -17,6 +17,7 @@ use Joomla\CMS\Factory;
 use Joomla\CMS\Language\Associations;
 use Joomla\CMS\MVC\Factory\MVCFactoryInterface;
 use Joomla\CMS\MVC\Model\ListModel;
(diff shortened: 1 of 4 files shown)


joomdonation:
@PhilETaylor @Bakual I think we should re-think about the implementation of method `getIsEmptyState` in our model classes. As of right now, we assume that the method is always called after **getItems** method (mean we have $query property already initialized somewhere)

However, since it is a **Public** method, we could not assure that. For a public method like that, developer could call it anytime they want and not depend on the calling of other methods. For example, in any view class, if we call this command at the top:

```php
$this->isEmptyState = $this->get('IsEmptyState');
```
We will get error below:

`Call to a member function clear() on array`

To implement this method properly, we can borrow `getTotal` method from `ListModel` class.

richard67:
See also #33466 ... seems me and others missed that when testing/reviewing.

PhilETaylor:
@joomdonation please move discussion of removing `getIsEmptyState` content to #33466

This PR stands on its own to fix a specific issue, and works well.

Issue #33452, opened 2021-04-30
[4.0] [Media] Invalid error when renaming a file to russian letters


Kostelano:
As in https://github.com/joomla/joomla-cms/issues/33449, the error is related to Russian letters, since there is no problem with renaming the file to Latin letters. 

### Steps to reproduce the issue
Go to the Media manager and try to rename the file using Russian letters, for example, `ыыыыыыыы`. We already get errors here, but one of them looks strange.

> Move file is not possible as destination file already exists.

![Screenshot_4](https://user-images.githubusercontent.com/8440661/116754904-33cae400-aa12-11eb-9b3b-475666d42330.png)

I found out that this error is due to the fact that Joomla "does not understand" Russian letters and tries to rename the file to `png` without a name.

Then I just copy the `test__ыыы.png` file to the images folder locally and try to rename it. Joomla gives several error messages, nevertheless renames it to `png` (no name), thereby deleting MY file.

![Screenshot_1](https://user-images.githubusercontent.com/8440661/116755950-e9e2fd80-aa13-11eb-85ec-900985317f95.png)

Yes, I remember that it seems in Joomla 3.9.25 there was a change regarding the Russian-language file / directory headers (I could be wrong), but in any case, the current behavior is incorrect and needs to be improved.

PhilETaylor:
> I found out that this error is due to the fact that Joomla "does not understand" Russian letters and tries to rename the file to png without a name.

The "bug" is that you dont have the correct PHP Extensions installed. 

If you are going to use non Latin ASCII chars then you need ICONV and INTL PHP Extensions installed. 

Without them installed `ыыыыыыыы` will become `""` (Empty string)

With them installed  `ыыыыыыыы` will become `yyyyyyyyyy`

Kostelano:
How should the most common user understand this if Joomla runs its own processes without reporting anything?

I am not saying that this is a direct bug bug bug.

PhilETaylor:
I guess by better feedback like https://github.com/joomla/joomla-cms/pull/33451 however your example of "ыыыыыыыыaaaaaa" would be made safe as "aaaaaa" so the file should rename to "aaaaaa" unless that file already exists.

Kostelano:
I also think that a similar message will be enough. In any case, the most ordinary user will understand what Joomla wants from him :)

PhilETaylor:
[...]

Issue #33450, opened 2021-04-30
[4.0] [Mass Mail Users] Infinite repeating error when clicking the Submit button


Kostelano:
### Steps to reproduce the issue
Go to Users - Mailing list, try to send the mail, but do not enter the subject / text of the message. Each time you click the SUBMIT button, a new error message will be displayed.

![Screenshot_3](https://user-images.githubusercontent.com/8440661/116753128-47288000-aa0f-11eb-986d-c148a4401eed.png)

I can say for sure that in some other component in a similar situation (today I don’t remember exactly where), the error message is not duplicated, but one "hangs" until the required condition is met.

rjharishabh:
Similar issue #33306

Kostelano:
Also, you need to change the message type to Warning as a minimum. Now - just information for the user.

brianteeman:
@Kostelano you just beat me to adding that bug to this issue

brianteeman:
If anyone wants to fix this then its the javascript in `build\media_source\com_users\js\admin-users-mail.es6.js` that needs to be ammended

ghost:
~@Kostelano your report reminds on an issue reporte by @PhilETaylor, can you have a look if its the same? issue is #33306~

Edit because of comment below, thanks @brianteeman.

brianteeman:
@sandramay0905 similar result different cause

PhilETaylor:
Why is this form not validating on blur like other forms? why is it waiting to validate after the form is submitted? Is that just because no one has applied the Javascript based validation on this page and it was overlooked?

richard67:
Closing as having a pull request. Please test #33485 . Thanks in advance.

Kostelano:
Can we open it up again?

PR https://github.com/joomla/joomla-cms/pull/33485 was closed unfortunately.

rjharishabh:
Please open this issue

Hackwar:
I created a PR with #39374. Closing this issue.

PR #33454, opened 2021-04-30
[4][Media] Renaming files when containing 100% non-latin ascii

PhilETaylor:
Fixes https://github.com/joomla/joomla-cms/issues/33452

### Summary of Changes

If all that is left after making the file name safe is the extension, then error and say why. 

Also fixes a double slash issue `//` when a filename has been made save. 

### Testing Instructions

WIHTOUT iconv and intl PHP extensions, attempt to rename a FILE to `МИ.png`

Screenshot 2021-04-30 at 23 13 11

and the moved file is created at `/images/png`

Screenshot 2021-04-30 at 23 12 35


Kostelano:
I have tested this item :white_check_mark: successfully on 6dd7e1cddb1183542aeb5b651c7e19c212ac69a9<br /><br />Thank you!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33454">issues.joomla.org/tracker/joomla-cms/33454</a>.</sub>

Screenshot_1

Fedik:
ideally it should be transliterated, but an exception also enough for quick fix :)

PhilETaylor:
It can't be if you are missing the PHP extensions to actually do that!

Fedik:
> It can't be if you are missing the PHP extensions to actually do that!

It is still possible, we actually doing it for an alias field, 
However it does not cover everything, mostly only for installed languages that provide transliteration in their localise.php

https://github.com/joomla/joomla-cms/blob/e44a1d22202030f8e69ba578e972a152d52f0d62/libraries/src/Filter/OutputFilter.php#L85-L89

and 

https://github.com/joomla/joomla-cms/blob/3a5984d03366f707ef842ee48051af194f68848f/libraries/src/Language/Language.php#L396-L401

upd, example for uk-UA localise.php:
https://github.com/Joomla-Ukraine/uk-UA/blob/88ca9b96a3872de4e4971315d24465ee9b78b5e0/uk-UA_3.x/language/uk-UA/uk-UA.localise.php#L100-L109

plugins/filesystem/local/src/Adapter/LocalAdapter.php
```diff
@@ -546,10 +546,16 @@ public function move(string $sourcePath, string $destinationPath, bool $force =
 		$name     = $this->getFileName($destinationPath);
 		$safeName = $this->getSafeName($name);
 
+		// If transliterating could not happen, and 100% of the chars are filtered out, then throw an error. 3 = file extension
```
richard67:
```suggestion
		// If transliterating could not happen, and all characters except of the file extension are filtered out, then throw an error.
```

infograf768:
for j4, the priority for the alias as well as for upload image, is:

1 intl extension
2 language pack custom transliteration.
3 for alias only, if no ascii, the date of saving

PhilETaylor:
Well if that case the CMS should be cleaning up the file name before asking the `File` library to write the file. 

The CMS File library is not the place to be injecting lots of language specific code. 

`File::makeSafe` is not the place for language pack transliteration. 

Replacing a failed `makeSafe` with todays date would be worse in my opinion - someone attempting to rename a file, only for that file to not be renamed to what they asked, but to a date... is likely to confuse.

Fedik:
> File::makeSafe is not the place for language pack transliteration.

Correct.

Here we have `LocalAdapter::getSafeName()`.
https://github.com/joomla/joomla-cms/blob/ef05728f055127f77c9a2a271b68b32f6f1ce95e/plugins/filesystem/local/src/Adapter/LocalAdapter.php#L777-L779

Additionally, the same issue may affect all methods in `LocalAdapter`, not only `move()`:
```
createFolder()
createFile()
copy()
```

Maybe you better move an exception in to `LocalAdapter::getSafeName()` method, 
Ideally add transliteration there also :)

PhilETaylor:
` $lang->transliterate('МИМИМИМИ')` also returns a blank string...

PhilETaylor:
I have added the exceptions in the LocalAdapter makeSafe, and added language transliterate with another check as both `File::makeSafe` and $lang->transliterate('МИМИМИМИ') can return a blank string.

plugins/filesystem/local/src/Adapter/LocalAdapter.php
```diff
@@ -772,7 +778,15 @@ private function getEncodedPath(string $path): string
 	private function getSafeName(string $name): string
 	{
 		// Make the filename safe
-		$name = File::makeSafe($name);
+		if (!$name = File::makeSafe($name))
+		{
+			throw new \Exception(Text::_('COM_MEDIA_ERROR_MAKESAFE'));
+		}
+
+		if (!$name = Factory::getLanguage()->transliterate($name))
```
Fedik: After some rest, and look at it again,
I think it will not work as I expected, because `File::makeSafe` already return "empty" name, and so the transliterate will transliterate emptinesssss.

Sorry for confusing.

The first exception after `File::makeSafe` is good here.

PhilETaylor: removed

richard67:
I've updated the branch to latest 4.0-dev to get rid of the unrelated javascript-cs error in drone.

@Kostelano Could you redo your test? The PR has meanwhile received some change beside that branch update. Thanks in advance.

Kostelano:
So, it works a little differently than I expect.

I just take a file with a Latin name and try to ADD Russian letters to its name. I receive an error NOT declared in the PR, but already another one.

Screenshot_1

If I try to rename the file to a Russian name WITHOUT a combination with Latin names, I will already get the declared error.

Screenshot_2

It seems to me that both cases should somehow correspond to the same error.

Also, note that errors are duplicated on the same page, similar to the error reported at https://github.com/joomla/joomla-cms/issues/33450.

PhilETaylor:
If a file exists with just the latin chars, then that is correct, as makeSafe will remove the russian and leave the other chars (if your server doesn't have iconv/intl installed)

Fedik:
I have tested this item :white_check_mark: successfully on cee796056c5cd3034832fa0a030e1936e7ed8375<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33454">issues.joomla.org/tracker/joomla-cms/33454</a>.</sub>

Kostelano:
I have tested this item :white_check_mark: successfully on cee796056c5cd3034832fa0a030e1936e7ed8375<br /><br />Ok, the main issue is solved by this PR. Thank you!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33454">issues.joomla.org/tracker/joomla-cms/33454</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33454">issues.joomla.org/tracker/joomla-cms/33454</a>.</sub>

wilsonge:
Thanks!

PR #33451, opened 2021-04-30
[4] Folder creation - Feedback INTL and ICONV PHP Extensions not installed

PhilETaylor:
Pull Request for Issue #33449 .

### Summary of Changes

If you are on a server where ICONV AND INTL extensions are NOT installed and you try to create a folder in com_media with name `ыыыыыыыы` then it will fail, but tell you that it created the folder. 

This PR handles that, and tells you the real reason. 

### Testing Instructions

On a server where ICONV AND INTL extensions are NOT installed and you try to create a folder in com_media with name `ыыыыыыыы` then it will fail, but tell you that it created the folder. 

Apply PR

On a server where ICONV AND INTL extensions are NOT installed and you try to create a folder in com_media with name `ыыыыыыыы` then it will fail, but tell you that it failed to create the folder. 

### Actual result BEFORE applying this Pull Request

Screenshot 2021-04-30 at 22 22 46

### Expected result AFTER applying this Pull Request

Screenshot 2021-04-30 at 22 22 26

### Documentation Changes Required

none. 

Isolated code test proving this issue available here https://3v4l.org/BRcNa

PR #33457, opened 2021-04-30
[4] Add intl extension check to system info, and while there clean up deprecations/types/[]

PhilETaylor:
### Summary of Changes

As a result of https://github.com/joomla/joomla-cms/pull/33451 and https://github.com/joomla/joomla-cms/pull/33454 I noticed ICONV was checked but not INTL - as both are needed for `File::makeSafe` it makes sense to have both listed in the System Information for ease of reference. 

While in this code, took the opportunity to remove deprecated `getDBO` method calls and add param types and return types. 

Screenshot 2021-05-01 at 00 18 43

### Testing Instructions

Load  System Information in Joomla Admin - if it loads, great. Check PHP tab and see your status of INTL (checks for presence of the function `transliterator_transliterate` which is the one Joomla uses in `File::makeSafe`

Try JSON and TXT download.

Works - but intl extension not in the list

Still works - but calls one less deprecated function and code is slightly more modern and less warnings in phpStorm. 

None


administrator/language/en-GB/com_admin.ini
```diff
@@ -119,6 +119,7 @@ COM_ADMIN_HELP_USERS_USER_MANAGER_EDIT="Users: New/Edit"
 COM_ADMIN_HELP_USERS_USER_NOTES="Users: User Notes"
 COM_ADMIN_HELP_USERS_USER_NOTES_EDIT="Users: User Notes - New/Edit"
 COM_ADMIN_ICONV_AVAILABLE="Iconv Available"
+COM_ADMIN_INTL_AVAILABLE="Intl Available"
```
brianteeman: it would be so much better if "intl" was hardcoded. It shouldn't be translated

PhilETaylor: I just did the same as Iconv :)

brianteeman: Doesn't make it right

brianteeman:
I have tested this item :white_check_mark: successfully on 68d52d8174900012aa70528e65d5ae9460fba604<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33457">issues.joomla.org/tracker/joomla-cms/33457</a>.</sub>

administrator/components/com_admin/src/Model/SysinfoModel.php
```diff
@@ -32,23 +32,23 @@ class SysinfoModel extends BaseDatabaseModel
...
+	protected $config = [];
 
 	/**
 	 * Some system values
 	 *
 	 * @var    array
 	 * @since  1.6
 	 */
-	protected $info = array();
+	protected $info = [];
```
sandewt, line 51: See also line 68

`protected $phpInfoArray;`

should be

`protected $phpInfoArray = [];`

sandewt: Line 150

`protected $safeData;`

should be

`protected $safeData = [];`

richard67:
~~If this PR will be merged, the same check should also be added to the Joomla Update Component's Pre-Update Checks. This will have to be done in the 3.10-dev branch so it can be checked before updating from 3.10 to 4. I could do that if necessary.~~

~~@PhilETaylor Do I understand right that the INTL extension is only needed in J4, not in J3? Or is that wrong and it should also be checked in J3? If my understanding is right, then the future PR for the pre-update check should show the check for INTL only if the version to be updated to is J4, like we have it already with the database compatibility check for MS SQL Server.~~

Update: I just see the INTL is only needed for transliteration, so it makes sense to add it to the system info with this PR here, but it is _not a required setting_ to be checked by the Pre-Update Check. What could make sense is to add it to the _recommended_ settings in the Pre-Update Checker.

administrator/components/com_admin/src/Model/SysinfoModel.php
richard67: @sandewt I don't agree. This PR here changes `array()` to short array syntax `[]`. The changes which you suggest mean to change the initial value from unset/null to an empty array. If that is right or not has to be checked elsewhere but seems to me to be out of scope of this PR and has to be carefully checked.

richard67:
I have tested this item :white_check_mark: successfully on 68d52d8174900012aa70528e65d5ae9460fba604<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33457">issues.joomla.org/tracker/joomla-cms/33457</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33457">issues.joomla.org/tracker/joomla-cms/33457</a>.</sub>

richard67:
Drone failure is **_not_** related to this PR just by the way.

administrator/components/com_admin/src/Model/SysinfoModel.php
sandewt: @richard, indeed you're right.

richard67:
Restarted drone. All fine now.

Quy:
Thank you!

Issue #33466, opened 2021-05-01
[4.0] Fix the new emptyState feature


bembelimen:
The emptyState feature is really nice, but the database query has two issues:

#### First issue:

Go to a model and add the following code:
```
	public function getTest()
	{
		echo $this->_getListQuery()->dump();

		$this->getIsEmptyState();

		echo $this->_getListQuery()->dump();
	}
```

Now call in your view:
```$this->get('Test');```

#### 2nd issue:
have a query with 2 from or joins, grouping, having, etc.

#### First issue
twice the same query

#### Second issue
The method is generic enough to work all the time (Imho clearing the query is per se wrong, it should build up the query...but that's my opinion)

#### First issue
different queries

#### Second issue
It fails on so many instances (Joins, Grouping, several froms, ...)

### Additional comments
The emptyState check changes the original query instead of making a clone (which is still broken) or do a normal new query  with the correct syntax. => also check ```_getListCount(...)```

richard67:
We could fix issue #33386 with use of the empty state feature, too, if this issue here is solved.

PhilETaylor:
Not really a release blocker - as you can simply add "return false" at the top of the isEmptyState function and be done with it and release Joomla 4 without this feature. The feature is only a smoke screen, all previous behaviour is still there.

Everyone seems to have their opinion on what should be in that function, but I have not seen any real concrete examples from people yet. 

`getIsEmptyState()` method is written to check the empty state of the last SQL run. (Ie removing any limits from the SQL and counting the number of rows). It is not written to check if a model is an empty state, because Joomla abuses models with huge queries with multiple joins. As written, the query in `getIsEmptyState()` method does exactly what it was designed to do, that is, to take the last query run, clean it up, and to run it again to return a count of rows. 

Used like that, the method is working well. 

Used any other way, with false assumptions, like the one you are reporting in this issue, will result in different behaviour. 

Issue 1 - if you are going to misuse functions then its up to you to understand the knock on effect. `_getListQuery` is a `protected` method for a reason and `getIsEmptyState()` method was written to directly manipulate the LAST query before it was run.

Issue 2 - It does not "fail on so many instances". All reported "instances" have already been fixed within hours of being reported. 

`_getListCount` literally does something similar to the code in isEmprtyState.

richard67:
See also https://github.com/joomla/joomla-cms/pull/33453#issuecomment-830542323 .

PhilETaylor:
[...]

PR #33468, opened 2021-05-01
[4] EmptyState replace custom sql with reusing existing methods

PhilETaylor:
@richard67 Please test this PR on all empty states except Categories and Tags (which have custom `getIsEmptyState` methods Im working on. 

If this works, then great an https://github.com/joomla/joomla-cms/issues/33466 is resolved.


PhilETaylor:
In fact this doesn't work - because it reintroduced the but that when you filter by a tag, you get 0 rows and a empty state...

Issue #33266, opened 2021-04-23
[4] Empty State's Todo


PhilETaylor:
Menu items in need of [new blank state design](https://github.com/joomla/joomla-cms/pull/33264)

- [x] [4.0] Moving reusable parts of blankstate layout to a JLayout #33288 @Bakual Thanks!
- [x] [4] redo the logic behind Empty State Queries. #33471
- [X] [4] Remove disabled and useless buttons on empty state leaving just New #33380

Module Items 

 - [x] [4] Introducing empty state to modules #33472

Menu Items: 

- [x] Banners -> Banners #33269
- [x] Banners -> Clients https://github.com/joomla/joomla-cms/pull/33381 
- [x] Banners -> Tracks https://github.com/joomla/joomla-cms/pull/33381
- [x] Contacts -> Contacts  #33265
- [x] Newsfeeds -> Feeds #33268
- [x] Smart Search -> Index #33489
- [x] Smart Search -> Maps #33491
- [x] Smart Search -> Search Terms https://github.com/joomla/joomla-cms/pull/33495/files
- [x] Tags https://github.com/joomla/joomla-cms/pull/33286
- [x] Users -> User Notes https://github.com/joomla/joomla-cms/pull/33273
- [x] Privacy -> Requests https://github.com/joomla/joomla-cms/pull/33272
- [x] Privacy -> Consents https://github.com/joomla/joomla-cms/pull/33272
- [x] Messaging -> Private Messages  #33274
- [X] Site Modules (If mad enough to delete the 3 default) https://github.com/joomla/joomla-cms/pull/33398
- [X] System -> Clear Cache https://github.com/joomla/joomla-cms/pull/33325 @PhilETaylor 
- [X]  System -> Global Checkin https://github.com/joomla/joomla-cms/pull/33383
- [X] System -> Discover [4] Empty State for System-> Discover #33400
- [X] System -> Warnings #33411
- [X] System -> Content Security Policy  #33416
- [x] System -> Redirects https://github.com/joomla/joomla-cms/pull/33384
- [x] Restyle the Post Installation No Messages page (Blank State Style) #33287
- [x] [4] Empty State for Checking for updates #33474
- [x] Categories #33286
- [x] Banners -> Categories #33286
- [x] Contacts -> Categories #33286
- [x] Users -> User Note Categories #33286
- [x] Newsfeeds -> Categories #33286

# Fields, Field groups.

[...]

PR #33471, opened 2021-05-01
[4] redo the logic behind Empty State Queries.

PhilETaylor:
#33466 @bembelimen  @richard67 @joomdonation 

Fulfils these new requirements from others
- Works with Categories and Tags which need to extend the query
- Allow calling the method in any order
- Allow calling the method as many times as you like
- Don't change $this->query directly, clone it.
- Abit easier to extend/change the query to meet the need of child model class.
~~- Resets the user state, so clicking on the Call To Action (Add/New) button doesn't start a new item in the Trash~~


bembelimen:
Thank you for the improvements.

PhilETaylor:
But do they work :)

administrator/components/com_categories/src/Model/CategoriesModel.php
```diff
@@ -490,30 +491,28 @@ public function countItems(&$items, $extension)
...
-			->clear('order')
-			->clear('where')
-			->select('count(*)');
+		// Remove select because it contains fields from joined tables, that have been removed from join.
+		$query->clear('select');
 
-		$sql->where($this->_db->quoteName('extension') . ' = :extension')
-			->bind(':extension', $extension);
+		// Add back a simple count select.
+		$query->select('count(*)');
```
bembelimen: I think this happens in the ```_getListCount``` call anyways?

PhilETaylor: well without those lines in the Model override, it doesn't work :)

richard67:
Adding release blocker label as inherited from the issue.

brianteeman:
This isnt a release blocker. If this matches the criteria to block a release then the tag should be applied to 90% of pr

richard67:
Well I don't really care. It will be merged in time anyway if it's good. Will see if I can test tomorrow.

administrator/components/com_categories/src/Model/CategoriesModel.php
joomdonation: @PhilETaylor @bembelimen You should not have to do that. Guess you have to do that manually because you added `$query->clear('select');` in the above line.

libraries/src/MVC/Model/ListModel.php
```diff
@@ -149,27 +150,56 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
...
 	}
 
+	/**
+	 * Provide a query to be used to evaluate if this is an Empty State, can be overridden in the model to provide granular control.
+	 *
+	 * @return DatabaseQuery
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	protected function getEmptyStateQuery(): DatabaseQuery
```
joomdonation: I unsure if we could force return type to `DatabaseQuery` (that's the reason we have `instanceof ` check in the code). So better, don't force return type for now.

libraries/src/MVC/Model/ListModel.php
```diff
@@ -149,27 +150,56 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
...
-			->clear('order')
-			->clear('where')
-			->select('count(*)');
-
-		$this->_db->setQuery($sql);
-
-		return !($this->_db->loadResult() > 0);
+		static $state;
+
+		if (empty($state))
```
joomdonation: This is not right. In case $state = false, condition return true, too. I think we need to change it to `if ($state === null)`

joomdonation:
Could you please move reset states logic to a separate PR? My feeling is that `getIsEmptyState` should not do the reset states itself. It is up to caller to do that if it wants. Further more, I'm unsure if the implementation of resetState is right but that is a different discussion

For this one, could we please focus on fixing the empty state logic? One problem at a time.

administrator/components/com_categories/src/Model/CategoriesModel.php
PhilETaylor: There is a reason for every single line I have written. Why don't you actually try doing it yourself and you will see.￼

PhilETaylor:
So once again one person wants it and another person doesn't want it. I just can't win. It works, And has zero knock on effects.￼￼

libraries/src/MVC/Model/ListModel.php
PhilETaylor: I don't think it should be cached at all. Joomla is multi request, another admin could be making changes at the same time.￼ I only attempted to add caching because it was asked for￼

joomdonation: To be clear, this is cache for a single request only. For this I also don't think that it needs to be cached. However, if you decide to cache it, it should be implemented properly. Otherwise, just remove cache.

joomdonation:
> So once again one person wants it and another person doesn't want it. I just can't win. It works, And has zero knock on effects.￼￼

I just express my concern and want to move it to separate PR so that maintainers and review and make decision easier. At the end, you don't need to make everyone happy. If you think it is OK, just leave it in this PR and maintainers will decide.

administrator/components/com_categories/src/Model/CategoriesModel.php
joomdonation: I just take a quick look at it only. I will try it myself later today.

PhilETaylor: Instead of trying to do all this magic by manipulating the list query, I can see us moving to handcrafting a empty state sql for each and every model, so that it is correct for that model to satisfy everybody. 

All these clears are just a fudge￼￼￼￼, some magic, in an attempt to reuse an existing query - an empty state is NOT always the same as having no matching results from a list query, There is a lot more to take into account!￼￼￼ and when you start to clear things like joins and wheres to clean up the list query to convert it into an empty state query, you sometimes then have to add back the specific filter states you need in order to craft the correct sql query that would return the value of an empty state.￼￼￼￼￼￼

I have literally spent the last four weekends on these, less than 20, lines of code. If you feel that you have a better solution and I will not be offended to see it. But as it is 4:51 am, I am going to bed.￼

joomdonation: I wish we have a way to get the main table for the model and we can do this easily. But we will have to use what we are having now.

> But as it is 4:51 am, I am going to bed

Why stay up too late like this? Get some sleeps and we will continue later.

joomdonation: > There is a reason for every single line I have written. Why don't you actually try doing it yourself and you will see
I tried that and it worked.

joomdonation:
So I still believe (myself) that `resetState` should be implemented like that or is called within `getIsEmptyState` (maybe it is just me, don't worry). As I said, you don't have to make everyone happy. If you want to keep it, leave it as how it is. Other maintainers will review and make decision.

In case you agree with me, I made a PR to your branch https://github.com/PhilETaylor/joomla-cms/pull/7 which I think it will make this PR safe to merge. Please take a look at it (especially the change in Categories model which we discussed earlier in this PR) and if it works and you are happy, merge it.

richard67:
For me it's ok to remove the resetState here and do it with another PR, to come forward here. I don't even insist on my issue to be fixed, I just thought it could be a good idea.

libraries/src/MVC/Model/ListModel.php
bembelimen: It has to be cached, because the correct way would be not to do it in the getEmptyState method at all but use the getEmptyState method to do it somewhere else, so you have a second call...

I agree ([as mentioned here](https://github.com/joomla/joomla-cms/issues/33466#issuecomment-830655857)) it shouldn't be in this PR.

PhilETaylor:
I have just landed and will work on this today. Again. But the pr proposed by @joomdonation i think has a major flaw - but I will be verbose in explaining that when I'm next at a desk and not sat at the controls of an airplane flying

PhilETaylor:
I take it back. The concern I had was unfounded as using `_getListCount` takes my concern (about joined tables like tags still being left in the select, when they were removed as a join). `_getListCount` does this clearing of `select` and readds `select(*)` for us, 

I have merged your PR to my branch.

PhilETaylor:
I have removed limit and offset from being cleared in `getEmptyStateQuery` - these are removed by _getListCount anyway

libraries/src/MVC/Model/ListModel.php
```diff
@@ -149,27 +150,42 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
...
+	 *
+	 * @return DatabaseQuery
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	protected function getEmptyStateQuery(): DatabaseQuery
+	{
+		$query = clone $this->_getListQuery();
+
+		if ($query instanceof DatabaseQuery)
```
joomdonation, line 164: As we force return type to DatabaseQuery, I think we can remove this $query instanceof DatabaseQuery check now. After that, I think we are fine.

PhilETaylor: Done.

administrator/components/com_categories/src/Model/CategoriesModel.php
```diff
@@ -490,30 +491,22 @@ public function countItems(&$items, $extension)
...
-	 * Is this an empty state, I.e: no items of this type regardless of the searched for states.
+	 * Manipulate the query to be used to evaluate if this is an Empty State to provide specific conditions for this extension.
 	 *
-	 * @return boolean
+	 * @return DatabaseQuery
 	 *
 	 * @since __DEPLOY_VERSION__
 	 */
-	public function getisEmptyState()
+	protected function getEmptyStateQuery(): DatabaseQuery
```
joomdonation: Should we remove return type here as well to have it consistent with `ListModel` class? Same for TagsModel.

PhilETaylor:
Once this is merged I'll do a quick recap of all the issues and outstanding states in https://github.com/joomla/joomla-cms/issues/33266 and then I'll be stepping back and taking a break. Thanks for your support on these, I think Joomla is better for this new feature and only sad the rest of my work has not made it as far.

joomdonation:
I have tested this item :white_check_mark: successfully on 66b2c1cef42092b98d1ac7860c75232b3642e92b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33471">issues.joomla.org/tracker/joomla-cms/33471</a>.</sub>

joomdonation:
@PhilETaylor Thanks for working so hard for Joomla! these days.

richard67:
I've updated the branch to latest 4.0-dev to get rid of the unrelated javascript-cs error in drone and restored the previous test result in the issue tracker.

drmenzelit:
I have tested this item :white_check_mark: successfully on af0ddad90f87f707f1b9d72f46dec1e3ae37f098<br /><br />I tested all views with empty states and checked using search and filters. I couldn't find problems. And the problem with "disappearing" categories when using filter tag is gone.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33471">issues.joomla.org/tracker/joomla-cms/33471</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33471">issues.joomla.org/tracker/joomla-cms/33471</a>.</sub>

bembelimen:
Thank you!

PR #31048, opened 2020-10-11
[4.0] Add WebP support to Image library

SharkyKZ:
### Summary of Changes

Adds WebP support to image library.

### Testing Instructions

Upload a `webp` image to template directory.
Go to `com_templates`.
In configuration add `webp` to `Valid Image Formats` field.
Go to template files and try to edit the image.

### Actual result BEFORE applying this Pull Request

Error:

>Attempting to load an image of unsupported type image/webp 

### Expected result AFTER applying this Pull Request

Image can be opened and edited (but is saved in wrong format).

### Documentation Changes Required

IDK.

PR #33481, opened 2021-05-02
[4.0] fix webp support 

brianteeman:
_redo of #33090_

#31048 added support for webp images however it was not complete. MediaHelper::isImage was not updated to support webp. As a result while webp could be added to media manager as an allowed image format everything worked except for the display of the thumbnail.

As this PR fixes that we might as well add webp to the default list of supported image formats.

### To test

1. either do a fresh install with the prebuilt package OR apply this pr, run npm ci and go to the media manager options. delete all the values in the Legal Extensions (File Types), Legal Image Extensions (File Types) and Legal MIME Types and then save the options. This will reset the value to the new defaults.
2. Try to upload any webp image
3. Try to crop and resize the image
4. Try to rename the image
5. Get a shareable link and make sure it works

### Before
image

### After
image


richard67:
@brianteeman Could you update the branch of this PR to latest 4.0-dev of upstream so we get rid of the unrelated javascript-cs errors in drone? Alternatively let me know if I shall do that for you. Thanks in advance.

brianteeman:
done

build/media_source/com_joomlaupdate/js/default.es6.js
```diff
@@ -456,8 +456,8 @@ Joomla = window.Joomla || {};
     if (extensionData.compatibilityData.resultGroup === 3) {
       PreUpdateChecker.nonCoreCriticalPlugins.foreach((cpi) => {
         // TODO: Make this typesafe
-        if (PreUpdateChecker.nonCoreCriticalPlugins[cpi].package_id == extensionId
-            || PreUpdateChecker.nonCoreCriticalPlugins[cpi].extension_id == extensionId) {
+        if (PreUpdateChecker.nonCoreCriticalPlugins[cpi].package_id === extensionId
```
Quy: Please remove unrelated changes.

richard67: That seems to come from an attempt to update the branch to latest 4.0-dev for fixing appveyor failing. I had asked Brian to do that. But it seems it was done wrong, with a squash commit instead of just merging in the branch..

brianteeman: feel free to fix it

richard67: Done. It was just merge of the 4.0-dev branch from upstream into the branch of this PR, without squash, and then push.

richard67: @Quy: Fixed.

brianteeman: Thanks. I am going to have to disable phpcs-fixer tools as its causing all sorts of problems

Quy:
I get the following error with crop/resize:

brianteeman:
@quy I can not replicate that
before

Quy:
Open browser's console.
Resize/crop.
Click Save button.
See console.

brianteeman:
I understand what you said but I cannot replicate it

image

image

Quy:
It must be my local setup then. I will try again on a hosting site. Thanks.

Quy:
Same issue on cloudaccess.host. Hopefully, someone can test and confirm.

brianteeman:
can anyone else replicate the issue @quy reported

jwaisner:
@brianteeman , I was able to replicate what @Quy  experiences

image

This issue occurs though regardless of the patch. Seems like something else is going on.

image

Quy:
Please fix conflicts so I can retest to see I can reproduce the issue. Thanks.

installation/sql/mysql/base.sql
```diff
@@ -157,7 +157,7 @@ INSERT INTO `#__extensions` (`package_id`, `name`, `type`, `element`, `folder`,
...
+(0, 'com_media', 'component', 'com_media', '', 1, 1, 0, 1, 1, '', '{"upload_extensions":"bmp,csv,doc,gif,ico,jpg,jpeg,odg,odp,ods,odt,pdf,png,ppt,txt,webp,xcf,xls,BMP,CSV,DOC,GIF,ICO,JPG,JPEG,ODG,ODP,ODS,ODT,PDF,PNG,PPT,TXT,WEBP,XCF,XLS","upload_maxsize":"10","file_path":"images","image_path":"images","restrict_uploads":"1","allowed_media_usergroup":"3","check_mime":"1","image_extensions":"bmp,gif,jpg,jpeg,png","ignore_extensions":"","upload_mime":"image\\/jpeg,image\\/gif,image\\/png,image\\/bmp,application\\/msword,application\\/excel,application\\/pdf,application\\/powerpoint,text\\/plain,application\\/x-zip","upload_mime_illegal":"text\\/html"}', ''),
```
richard67: With the merge conflict resolving, the "webp" got lost in the parameters "image_extensions" and "upload_mime".

See here the change from the merge from 4.0-dev: https://github.com/joomla/joomla-cms/pull/33481/commits/7c56a8142ee1d55e7780a79b34e0cefd2c2c7873

installation/sql/postgresql/base.sql
```diff
@@ -163,7 +163,7 @@ INSERT INTO "#__extensions" ("package_id", "name", "type", "element", "folder",
...
+(0, 'com_media', 'component', 'com_media', '', 1, 1, 0, 1, 1, '', '{"upload_extensions":"bmp,csv,doc,gif,ico,jpg,jpeg,odg,odp,ods,odt,pdf,png,ppt,txt,webp,xcf,xls,BMP,CSV,DOC,GIF,ICO,JPG,JPEG,ODG,ODP,ODS,ODT,PDF,PNG,PPT,TXT,WEBP,XCF,XLS","upload_maxsize":"10","file_path":"images","image_path":"images","restrict_uploads":"1","allowed_media_usergroup":"3","check_mime":"1","image_extensions":"bmp,gif,jpg,jpeg,png","ignore_extensions":"","upload_mime":"image\\/jpeg,image\\/gif,image\\/png,image\\/bmp,application\\/msword,application\\/excel,application\\/pdf,application\\/powerpoint,text\\/plain,application\\/x-zip","upload_mime_illegal":"text\\/html"}', '', 0, 0),
```
richard67: See my previous review comment for MySQL. The same applies here.

installation/sql/mysql/base.sql
richard67: The reason is that it was not one of these merge conflicts where you can accept the change from the one or the other side. Both made changes in the same line, and so it really needs to put both changes together. This can't be done with a quick click in a GUI, it needs to edit the file and solve the conflict manually by merging both changes together into a single line.

Quy:
Here is the json response after saving the resized/cropped webp image. I don't know media manager to see how it handles the saved file.

> {"success":false,"message":"Unable to upload file.","messages":{"error":["Illegal mime type detected: application\/octet-stream"]},"data":null}

brianteeman:
can you please confirm the settings in com_media options

richard67:
@brianteeman The problem seems to be the MIME type. @Quy 's JSON response shows `Illegal mime type detected: application/octet-stream`. Your PR adds "image/webp" to the default value for the field "Allowed MIME types" (or so), but it doesn't add "application/octet-stream". I'm not sure if that MIME type should be added because I have no idea what else uses that and if it can be dangerous.

@Quy Can you check if it works after you've added "application/octet-stream" to the allowed MIME types in media manager config?

brianteeman:
it shouldnt add application/octet-stream.

brianteeman:
I think I know the cause - just checking

richard67:
Possibly we need to update some library used for MIME type detection because it doesn't know webp yet?

brianteeman:
no - thats not it - please wait

brianteeman:
So the problem sounds the same as the issue with pdf files back with a change in 3.7 
Can you confirm if fileinfo is enabled in php
Can you change this setting and see if it works
image

Quy:
> Can you confirm if fileinfo is enabled in php

Enabled.

> Can you change this setting and see if it works

`PHP Notice:  exif_imagetype(): Read error! in \libraries\src\Helper\MediaHelper.php on line 74`

The saved file no longer has the `image/webp` MIME type.
```
Folder
    local-images:/
Type
    File
Date Created
    2021-05-25 15:32
Date Modified
    2021-05-25 15:32
Dimensions
    - 
Size
    287.95 KB 
MIME Type
    application/octet-stream
Extension
    webp
```

Quy:
Works in Chrome v91.0.4472.7 but not in Firefox v9.0 on Windows 10 Pro.

brianteeman:
@quy can you please confirm the version of Firefox you are using?

Quy:
Firefox 89.0.1

brianteeman:
Still cannot replicate your problem

image

Quy:
So you can crop/resize/rotate and save & close/open the image again? I will test again shortly and let you know.

brianteeman:
ah I see - its failing at the save part

brianteeman:
Wondering if the save in the crop/resize plugin is actually saving it as a jpg but keeping the extension as webp which chrome has no problem with but firefox does??

Quy:
See https://github.com/joomla/joomla-cms/pull/33481#issuecomment-847979184

brianteeman:
Its an upstream problem with the cropper.js library

https://github.com/fengyuanchen/cropperjs/issues/800#issuecomment-864427190

brianteeman:
Could you please retest. I found and fixed an error in the js which I have now fixed and it would appear to resolve the issue. 

Note that if it doesnt then this js correction is still needed

Quy:
It is still an issue with crop.

Edit an image.
Crop the image.
Click Save & Close button.
See browser's console.

brianteeman:
are rotate and resize fixed?

Quy:
Same error for all 3 actions.

brianteeman:
Ok so the conclusion is that it is **not** a bug in the js here or the cropper library. Firefox does not identify webp images when used with the HTMLCanvasElement.toDataURL() method and it just falls back to png

Reference [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL)
Tests [Tests](https://kangax.github.io/jstests/toDataUrl_mime_type_test/)

image


image

richard67:
Reference [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL) says:

> If the requested type is not image/png, but the returned value starts with data:image/png, then the requested type is not supported.

I have no words. That seems to me like `if my function returns a "Successfully completed the xyz task." message for the "abc" task, the task has failed.`

brianteeman:
Closed as image editing for webp doesnt work with cropperjs and firefox windows

PR #33484, opened 2021-05-02
[4.0] JS codestyle

Hackwar:
This should fix the codestyle error Drone is complaining about.


joomdonation:
The current javascript code in com_joomlaupdate is completely broken. So we can merge this PR now to solve cs issue and try to fix the real js error later.

richard67:
My attempt to fix it was #33487 . I can re-open it if preferred by the release lead.

richard67:
Another attempt see #33488 .

HLeithner:
Merging this and hope that #33488 is ready soon.

Thanks

PR #33488, opened 2021-05-02
[4.0] Fix JS errors for com_joomlaupdate

joomdonation:
Pull Request for Issue # .

### Summary of Changes
There are currently many js errors in com_joomlaupdate. This PR attempts to fix this.

### Testing Instructions
I'm unsure as I don't understand fully how pre-update checks in com_joomlaupdate works yet. Maybe for now, code review would be enough because the current code is already broken.

### DIFF
--- a/administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
+++ b/administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
@@ -17,8 +17,8 @@ use Joomla\Component\Joomlaupdate\Administrator\View\Joomlaupdate\HtmlView;
 /** @var HtmlView $this */
 
 // JText::script doesn't have a sprintf equivalent so work around this
-Factory::getDocument()->addScriptDeclaration("var COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION = '" . Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-right large-icon" style="font-size:0.85rem"></span>', true) . "';");
-Factory::getDocument()->addScriptDeclaration("var COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION = '" . Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-up large-icon" style="font-size:0.85rem"></span>', true) . "';");
+Factory::getDocument()->addScriptOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-right large-icon" style="font-size:0.85rem"></span>', true));
+Factory::getDocument()->addScriptOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-up large-icon" style="font-size:0.85rem"></span>', true));
 Factory::getDocument()->addScriptOptions('nonCoreCriticalPlugins', $this->nonCoreCriticalPlugins);
 
 $compatibilityTypes = array(
--- a/administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
+++ b/administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
@@ -17,9 +17,9 @@ use Joomla\Component\Joomlaupdate\Administrator\View\Joomlaupdate\HtmlView;
 /** @var HtmlView $this */
 
 // JText::script doesn't have a sprintf equivalent so work around this
-Factory::getDocument()->addScriptOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-right large-icon" style="font-size:0.85rem"></span>', true));
-Factory::getDocument()->addScriptOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-up large-icon" style="font-size:0.85rem"></span>', true));
-Factory::getDocument()->addScriptOptions('nonCoreCriticalPlugins', $this->nonCoreCriticalPlugins);
+Factory::getDocument()->addScriptOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-right large-icon" style="font-size:0.85rem"></span>', true))
+	->addScriptOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-up large-icon" style="font-size:0.85rem"></span>', true))
+	->addScriptOptions('nonCoreCriticalPlugins', $this->nonCoreCriticalPlugins);
 
 $compatibilityTypes = array(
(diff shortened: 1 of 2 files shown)


build/media_source/com_joomlaupdate/js/default.es6.js
```diff
@@ -495,10 +495,12 @@ Joomla = window.Joomla || {};
               >
               <span class="icon-help"></span>
               ${Joomla.JText._('COM_JOOMLAUPDATE_VIEW_DEFAULT_HELP')}
-              </span>`);
-          const popoverElement = pluginTitleTableCell.find('.hasPopover');
-          popoverElement.style.cursor = 'pointer';
-          popoverElement.popover({ placement: 'top', trigger: 'focus click' });
+              </span>`;
+          const popoverElement = pluginTitleTableCell.querySelector('.hasPopover');
+          if (popoverElement) {
```
joomdonation, line 495: @dgrammatiko Do you know how to convert this block of code? The original code from 3.10.0 can be found here https://github.com/joomla/joomla-cms/blob/3.10-dev/media/com_joomlaupdate/js/default.js#L331-L333 . Not sure how to do this conversion on Joomla 4?

richard67:
@joomdonation It needs to fix javascript-cs, see https://ci.joomla.org/joomla/joomla-cms/42929/1/21 .

build/media_source/com_joomlaupdate/js/default.es6.js
richard67: Maybe use `let` instead of `const` so you get:

```
let popoverElement = pluginTitleTableCell.querySelector('.hasPopover');
if (popoverElement) {
  popoverElement.style.cursor = 'pointer';
  popoverElement.popover({placement: 'top', trigger: 'focus click'});
}
```

joomdonation: I think it is not related. I don't know how to add popover to the element here, so @dgrammatiko  or @Fedik or maybe @wilsonge could help us.

dgrammatiko: To initialize the popover use

```js
new bootstrap.Popover(popoverElement, {placement: 'top', trigger: 'focus click'});
```

dgrammatiko: Btw you need to include the popover in the layout HTMLHelper::_('bootstrap.popover')

build/media_source/com_joomlaupdate/js/default.es6.js
```diff
@@ -454,11 +454,10 @@ Joomla = window.Joomla || {};
 
     // Process the nonCoreCriticalPlugin list
     if (extensionData.compatibilityData.resultGroup === 3) {
-      PreUpdateChecker.nonCoreCriticalPlugins.foreach((cpi) => {
-        // TODO: Make this typesafe
-        if (PreUpdateChecker.nonCoreCriticalPlugins[cpi].package_id === extensionId
-            || PreUpdateChecker.nonCoreCriticalPlugins[cpi].extension_id === extensionId) {
-          document.getElementById(`#plg_${PreUpdateChecker.nonCoreCriticalPlugins[cpi].extension_id}`).remove();
+      PreUpdateChecker.nonCoreCriticalPlugins.forEach((plugin, cpi) => {
+        if (plugin.package_id.toString() === extensionId
```
wilsonge, line 454: Does toString work if it's acting on a string already (i'm sure there was a database type where integers' come out as a string)

build/media_source/com_joomlaupdate/js/default.es6.js
wilsonge: > Btw you need to include the popover in the layout HTMLHelper::_('bootstrap.popover')

I called WA addScript directly in the layout - because I don't think we need the HtmlHelper initialisation of the element?

build/media_source/com_joomlaupdate/js/default.es6.js
Fedik: yes it will work, `"blabla".toString()`

build/media_source/com_joomlaupdate/js/default.es6.js
joomdonation: @dgrammatiko Thanks, implemented and confirm that it works.

administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
```diff
@@ -19,7 +19,7 @@
 // JText::script doesn't have a sprintf equivalent so work around this
 Factory::getDocument()->addScriptDeclaration("var COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION = '" . Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_MORE_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-right large-icon" style="font-size:0.85rem"></span>', true) . "';");
```
dgrammatiko: Can we use also `$doc->addScriptOptions()` for these 2 strings here?

I kinda hate the var and the inline js just to pass data from PHP to the JS world, so 2005...

administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
```diff
@@ -19,7 +19,7 @@
...
 Factory::getDocument()->addScriptDeclaration("var COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION = '" . Text::sprintf('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION', '<span class="icon-chevron-up large-icon" style="font-size:0.85rem"></span>', true) . "';");
```
dgrammatiko: Same as above...

administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
joomdonation: Original code :).

dgrammatiko: I know and probably should have made that comment the other day but would be great if we could "fix" it in this PR. If not no worries

joomdonation:
@wilsonge As of right now, all the errors are fixed and I would say that it works in the same way with what we are having on 3.10.0. You can review the code and get it merged (I made basic test by installing older version of Akeeba Backup and JCE editor for testing)

However, there are still some problems:

- The layout of the page does not look good. We should have someone with frentend/design skill to take a look at it and sovle it
- During the code review, I can see a potential issue (comes from the original code, before we do this conversion to convert to vanilla js). See https://github.com/joomla/joomla-cms/blob/3.10-dev/media/com_joomlaupdate/js/default.js#L292

Basically, from what I understand, the loop over all the none core plugins and if that plugin belong to an extension, remove it from the array. But that could cause all plugins load being checked (causes by the removal mentioned). Maybe you can ask the developer who worked on the feature to take a look at it.

dgrammatiko:
> However, there are still some problems:

George had asked me to fix this after the BS js PR but I never made a PR because my solution (SPA basically)was touching the restore.php and that wasn't a discussion I wanted myself involved. This is (I think) the repo: https://github.com/joomla/joomla-cms/compare/4.0-dev...dgrammatiko:4.0-dev_joomlaUpdate?expand=1

At the very least check the changes in the controller and the way I pass data from PHP to JS. The current state is god awful...

joomdonation:
@dgrammatiko I haven't spent time to read the whole code to understand the process, so I could not comment on your solution right now. However, if you make a PR, I would be happy to review and test it. The current code is not looking good, I agree.

administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_preupdatecheck.php
joomdonation: @dgrammatiko Fixed that as suggested :)

build/media_source/com_joomlaupdate/js/default.es6.js
```diff
@@ -474,31 +473,33 @@ Joomla = window.Joomla || {};
...
               ${Joomla.JText._('COM_JOOMLAUPDATE_VIEW_DEFAULT_HELP')}
-              </span>`);
-          const popoverElement = pluginTitleTableCell.find('.hasPopover');
-          popoverElement.style.cursor = 'pointer';
-          popoverElement.popover({ placement: 'top', trigger: 'focus click' });
+              </span>`;
+          const popoverElement = pluginTitleTableCell.querySelector('.hasPopover');
+          if (popoverElement) {
+            popoverElement.style.cursor = 'pointer';
+            const popover = new bootstrap.Popover(popoverElement, { placement: 'top', html: true, trigger: 'focus click' });
```
joomdonation: @dgrammatiko Could you help with the CS issue related to this line of code ? Leave it like that, the CS error is https://ci.joomla.org/joomla/joomla-cms/43002/1/21 . If I change it to the code below, another CS error appear
```javascript
new bootstrap.Popover(popoverElement, { placement: 'top', html: true, trigger: 'focus click' });
```
Don't know how to make it happy :(

dgrammatiko: What's the error when you're using this:

```js
new bootstrap.Popover(popoverElement, { placement: 'top', html: true, trigger: 'focus click' });
```

joomdonation: I don't remember exactly but it does throws a message like you cannot create a new object without storing it in a variable :(.

dgrammatiko: Just mute it using: `// eslint-disable-next-line no-new` 

Unless this `bootstrap.Popover(popoverElement, { placement: 'top', html: true, trigger: 'focus click' });` works (I doubt it will)

joomdonation: Thanks @dgrammatiko . Had to mute it using `// eslint-disable-next-line no-new` as instructed.

build/media_source/com_joomlaupdate/js/default.es6.js
```diff
@@ -140,14 +140,14 @@ Joomla = window.Joomla || {};
         if (toggle.target.dataset.state === 'closed') {
           toggle.target.dataset.state = 'open';
           // eslint-disable-next-line max-len,no-undef
```
wilsonge: You can remove the `no-undef` ignore here (possible max-len too now this is a proper function - but less sure - same in the other places too where you're using the getOptions)

wilsonge:
Starting to look better. Thanks for fixing my mess trying to port this!

build/media_source/com_joomlaupdate/js/default.es6.js
joomdonation: @wilsonge Done.

wilsonge:
This is clearly an improvement merging this now

PR #33176, opened 2021-04-17
[4.0] Fixes radio fields layout

YatharthVyas:
Pull Request for Issue #33144

### Summary of Changes
This PR changes the CSS class and styling applied to the input type=radio's label.  Refer to https://github.com/joomla/joomla-cms/blob/2c302641aaad823ac91421a0b11f14e3f8e2fd27/layouts/joomla/form/field/radio/buttons.php#L49
Bootstrap suggests that you organize the radio options as:
```
<div class="form-check">
  <input class="form-check-input" type="radio" name="exampleRadios" id="exampleRadios1" value="option1" checked>
  <label class="form-check-label" for="exampleRadios1">
    Default radio
  </label>
</div>
<div class="form-check">
  <input class="form-check-input" type="radio" name="exampleRadios" id="exampleRadios2" value="option2">
  <label class="form-check-label" for="exampleRadios2">
    Second default radio
  </label>
</div>
```
This PR changes the `form-check` class applied to each option's `<label>`  to `form-check-label` and encloses each option's input-label pair in a `<div class="form-check">`

### Testing Instructions
Change the XML form for any plugin's admin console page to include a radio field like: (similar to the [example in Joomla 4 docs](https://docs.joomla.org/J4.x:Creating_a_Plugin_for_Joomla))
```
<field
  name="search_archived"
  type="radio"
  class="switcher"
  label="LABEL_NAME"
  default="0"
>
  <option value="1">JYES</option>
  <option value="0">JNO</option>
</field>
```
Note that there is **NO** `layout="joomla.form.field.radio.switcher"` part.
Please also test for `class="btn-group"` and `class="btn-group-yesno"`

For example, in the screenshots below I've removed line 37 from plugins/content/vote/vote.xml to achieve the desired testing condition
https://github.com/joomla/joomla-cms/blob/2c302641aaad823ac91421a0b11f14e3f8e2fd27/plugins/content/vote/vote.xml#L37

### Actual result BEFORE applying this Pull Request
image

### `no class`
image

### `btn-group`
image

### `btn-group-yesno`
image

### Documentation Changes Required
 `layout="joomla.form.field.radio.switcher"` should be mentioned in the [Joomla 4 Plugin Tutorial Page's XML Section](https://docs.joomla.org/J4.x:Creating_a_Plugin_for_Joomla) for making radio fields with 2 options.

PR #33503, opened 2021-05-03
[4.0] Fix Custom Classes for Radio Fields

YatharthVyas:
Pull Request for https://github.com/joomla/joomla-cms/pull/33176#issuecomment-831180478
Thanks for bringing it into my attention @regularlabs 

### Summary of Changes
Reverted `$divClass` back to `$class` for styling the outer div

### Testing Instructions
Make a new plugin XML form with a `radio` field or use an existing one for example:  `/plugins/content/vote/vote.xml`
https://github.com/joomla/joomla-cms/blob/8a1d803f59a669dfe4574867bf5e5b7d751f6d91/plugins/content/vote/vote.xml#L33

Replace the Radio field with the following code:
```
<field
	name="show_total_votes"
	type="radio"
	label="PLG_VOTE_TOTAL_VOTES_LABEL"
	class=""
	default="0"
	filter="integer"
>
	<option value="0">JHIDE</option>
	<option value="1">JSHOW</option>
</field>
```
Try out the following combinations for the `class=""` attribute: (Feel free to replace text-muted with any other custom class)

1. "btn-group"
2. "btn-group text-muted"
3. "btn-group btn-group-yesno"
4. "btn-group btn-group-yesno text-muted"
5. "text-muted"
6. And lastly, remove the class attribute and check

The Custom Class (text-muted) will not affect the output and it solely depends on `btn-group` and `btn-group-yesno`

The custom class will also be taken into consideration 

https://user-images.githubusercontent.com/53610833/116872439-81726700-ac33-11eb-8b58-78272893bf38.mp4

1. In [Joomla 4 Docs Plugin Tutorial Page](https://docs.joomla.org/J4.x:Creating_a_Plugin_for_Joomla), the XML form has mentioned `class="switcher"` for the radio field. This class is redundant and doesn't affect anything. I believe the author intended to write `layout="joomla.form.field.radio.switcher"`

2. While studying the code, I found out that classes, (Edit: also onchange and onclick) can also be applied to the `<option>` as `<option class="text-muted">` under the radio field. For example:
```
<field
	name="show_total_votes"
	type="radio"
	label="PLG_VOTE_TOTAL_VOTES_LABEL"
	default="0"
	filter="integer"
>
	<option value="0" class="text-muted">JHIDE</option>
	<option value="1">JSHOW</option>
</field>
```
will dislpay the output as:

image

The documentations doesn't mention this: (only attribute given here is value) so a new attribute for `class` (Edit: also `onchange` and `onclick`) must be mentioned here

image

Link: https://docs.joomla.org/Radio_form_field_type

The code that is responsible for checking the option tag's classes is: 
https://github.com/joomla/joomla-cms/blob/8a1d803f59a669dfe4574867bf5e5b7d751f6d91/layouts/joomla/form/field/radio/buttons.php#L113


YatharthVyas:
appveyor build is failing because the link https://getcomposer.org/composer-1.phar seems to be broken at the moment. I don't think any of the changes in my PR are the culprit for this one 🤔

regularlabs:
Fix looks good 👍

However, I do feel there is way too much magic going on in this file.
Personally, I think the HTML output should not rely on the class names.
A class name should just be that. Let CSS handle it from there.

Wouldn't it make more sense to just make an extra layout file for the 'btn-group' version, like there is one for the switcher?

YatharthVyas:
Making a new layout seems to be a tempting fix however, that would imply the usage of something on the lines of `layout="joomla.form.field.radio.btngroup"` attribute for this and it might cause issues in terms of backwards compatibility for all the fields out there that are working on `class="btn-group"`

regularlabs:
Then maybe add something like this to the top of the default layout:
```
if(strpos($class, 'btn-group')) {
   ... code to load the 'group' layout ...
   return;
}
```

regularlabs:
And maybe even have a separate yesno layout too:
```
if(strpos($class, 'btn-group-yesno')) {
   ... code to load the 'yesno' layout ...
   return;
}
if(strpos($class, 'btn-group')) {
   ... code to load the 'group' layout ...
   return;
}
```

YatharthVyas:
If I were to separate these out then the code would, of course, be more readable. However, because the underlying HTML markup and attributes except for the class are identical for all the cases, a lot of the code would be repeated in the same file, which in my opinion, isn't a good practice. This tantamounts to being a comparison between better code readability v/s less redundancy. Both have their pros but I'm biased to lean towards less redundancy here as it implies fewer changes in the code and is backed by the assurance that things are working as intended in the current state.

regularlabs:
Ok.

ceford:
I have tested this item :white_check_mark: successfully on ac524e0011e1f55bda3c57694ccf46c0a53cc607<br /><br />Styling works as described. I fixed the **Creating a Plugin for Joomla** tutorial to remove class switcher and use layout instead. I left the **Radio form field type** tutorial alone because ...<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33503">issues.joomla.org/tracker/joomla-cms/33503</a>.</sub>

Quy:
I have tested this item :white_check_mark: successfully on ac524e0011e1f55bda3c57694ccf46c0a53cc607<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33503">issues.joomla.org/tracker/joomla-cms/33503</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33503">issues.joomla.org/tracker/joomla-cms/33503</a>.</sub>

richard67:
Thanks!

YatharthVyas:
Thanks everyone!
Especially @ceford for the documentation fix

machadoug:
Thank you @YatharthVyas for this PR. 
I just noticed the **`active`** class doesn't change when I click on another option. 

I've added your code to /plugins/content/vote/vote.xml for testing purposes and it didn't work. See:

[![active class not changing]](https://user-images.githubusercontent.com/1613200/121271109-59f96480-c899-11eb-9db5-61fdcc78ca95.mp4)

Code added:
```
<field
	name="show_total_votes"
	type="radio"
	label="PLG_VOTE_TOTAL_VOTES_LABEL"
	default="0"
	filter="integer"
	class="btn-group btn-group-yesno"
>
	<option value="0" >JHIDE</option>
	<option value="1">JSHOW</option>
</field>
```

I'm not sure where/how to fix it and whether I should create a new issue or not.

YatharthVyas:
@machadoug Thank you for informing me about this. I'm afraid I might be doing something wrong because I was unable to replicate this in my local environment:

custom-class

### Note:
The code provided by you seems perfect and you've done everything correctly in the demo so there is no mistake made by you. However, I observed that the UI in the video you shared seems to belong to an older version of Joomla (before the Admin UI was repainted) so I think the issue might be occurring here because it is running an older version of J4.

Could you please try to update your build to the latest commit/nightly? I have a feeling it might resolve the issue. 

Please notify me in case the issue persists after updating.

### System Information:

image

My J4 Version is updated to the latest commit:  317f826319a11dd5ce3e6a4df1e599bfa3cc2d08

machadoug:
Thank you @YatharthVyas. 
The problem was fixed using the latest nightly build.

PR #33518, opened 2021-05-03
[4.0] update cacert.pem

brianteeman:
replaces a 4 year old cert with the latest one available from http://curl.haxx.se/ca/cacert.pem


wilsonge:
Thanks!

wilsonge:
If you want a small task. We're shipping with the composer ca bundle module (https://github.com/composer/ca-bundle ) anyhow which ships the mozilla CA. We can add an explicit dependency on it and just use the `CaBundle::getBundledCaBundlePath()` to avoid us having to update this (given we're not very good at updating it) and reduce duplicated files in the CMS

brianteeman:
Thanks for the merge. I will look at the task later

regularlabs:
Joomla 4 uses the ca-bundle cacert.pem (/libraries/vendor/composer/ca-bundle/res/cacert.pem)
This now causes issues with servers using updated Let's Encrypt certificates.

See: https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/
and https://medium.com/geekculture/will-you-be-impacted-by-letsencrypt-dst-root-ca-x3-expiration-d54a018df257

This also concerns Joomla 3 setups!

brianteeman:
see #34703

regularlabs:
Thanks

wojsmol:
@brianteeman @wilsonge  Certificate merged with this PR contains `DST Root CA X3` expired Let's Encrypt Root certificate - this will cause issues on openssl version 1.0.2 - see [Old Let’s Encrypt Root Certificate Expiration and OpenSSL 1.0.2](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/)

regularlabs:
Also something half-related... if I update my Joomla 4 setup to the nightly build, I still don't get a `libraries/src/Http/Transport/cacert.pem` file. 
All that Transport folder contains is:
image

So I guess there is something wrong in the Joomla 4 update process that doesn't add this file when it isn't there...
🤷🏻‍♂️

regularlabs:
And just to make clear what the effects of this issue are:
Currently all Joomla websites trying to 'Find updates' to extensions hosted on servers with updated Let's Encrypt certificates (like mine) will get error messages, stating they can't connect to the update site.
image

This is of course a serious issue. So I assume that Joomla will release new versions asap that fix this.

brianteeman:
probably best to create a new issue - not many people like me are subscribed to all messages

wojsmol:
We have a 2 issues here. I can create one related to expired Let's Encrypt root certificate.

wojsmol:
@regularlabs PR for 3.10 https://github.com/joomla/joomla-cms/pull/35781 - partial as in is 3.10 we have separate copy in composer/ca-bundle witch is a dependency of joomla/http. 
For 4.0-dev and up we use composer/ca-bundle directly -see https://github.com/joomla/joomla-cms/pull/34697.

regularlabs:
Awesome. Thanks. Hopefully, Joomla will release new versions asap to deal with this global issue that now affects millions.

wojsmol:
AFAIK we will be faster then CMS with name starting with W.

regularlabs:
Wix?

wojsmol:
@regularlabs We both know a connect name :)

Issue #33525, opened 2021-05-03
[4] Admin Module - Login Form 


PhilETaylor:
### Steps to reproduce the issue

You can add the Admin Module - Login Form  to your admin dashboards - this makes literally no sense! 

Also it states you cannot unpublish it - this is a lie, you can, you can click the cog and click unpublish ;) 

<img width="308" alt="Screenshot 2021-05-04 at 00 02 35" src="https://user-images.githubusercontent.com/400092/116943782-2c9f1280-ac6c-11eb-99eb-9cd69b94672f.png">

<img width="1597" alt="Screenshot 2021-05-04 at 00 03 42" src="https://user-images.githubusercontent.com/400092/116943797-3294f380-ac6c-11eb-8895-33d363fc73bf.png">

rjharishabh:
PR #33534

brianteeman:
Not every module is designed to work everywhere. Just as it would be crazy if you put the popular articles module in the toolbar position. The only thing that does need to be changed is the language string. Changing the text from "it can't be" to "it should not be"

PhilETaylor:
https://github.com/joomla/joomla-cms/pull/33534

PR #33534, opened 2021-05-04
[4.0] Admin module - Login Form

rjharishabh:
Pull Request for Issue #33525.

### Summary of Changes
Update Required Language files

### Testing Instructions
You can add the Admin Module - Login Form to your admin dashboards - this makes literally no sense!

Also it states you cannot unpublish it - this is a lie, you can, you can click the cog and click unpublish ;)

Please enable two factor Authentication

mod_login_before

mod_login__before

mod_login_after

mod_login__after

None


brianteeman:
Not every module is designed to work everywhere. Just as it would be crazy if you put the popular articles module in the toolbar position. The only thing that does need to be changed is the language string. Changing the text from "it can't be" to "it should not be"

rjharishabh:
It can be unpublished from Admin dashboard

brianteeman:
it can be but you should not

brianteeman (changes requested):
please revert all the tfa changes. they are not needed
image

rjharishabh:
No need of this PR

Issue #33528, opened 2021-05-03
[4] Database Password help text should be under the field not on top


PhilETaylor:
### Steps to reproduce the issue

Global Config -> system

### Expected result

Database Password help text should be under the field not on top to be consistent with everywhere else

### Actual result
Database Password help on top

<img width="880" alt="Screenshot 2021-05-04 at 00 12 53" src="https://user-images.githubusercontent.com/400092/116944531-95d35580-ac6d-11eb-955e-1f74707a489a.png">

YatharthVyas:
Edit: This seems to be an issue with all type="password" fields that have a description without rules=""
In such cases, an extra description is displayed on top of the field. I'm working on a fix 

<hr/>
Original message: 

It appears as if the person who wrote this code has intentionally taken extra efforts to hide the description from coming below it so they could place it above. Are you sure this was unintentional and needs to be fixed? (Maybe they kept it above the field because the message is an important warning and needs extra emphasis?)

This can be fixed by removing the `hiddenDescription` attribute and the markup above. I'd be happy to submit a PR for this if this issue can be confirmed.

https://github.com/joomla/joomla-cms/blob/3cebcc73cd03f015807444ce69e52da0f58c8932/administrator/components/com_config/forms/application.xml#L206

```
<field
	name="password"
	type="password"
	label="COM_CONFIG_FIELD_DATABASE_PASSWORD_LABEL"
	description="COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC"
	filter="raw"
	autocomplete="off"
	size="30"
	hiddenDescription="true"
	lock="true"
/>
```

Quy:
Please test PR #33536

PR #33536, opened 2021-05-04
[4.0] Fix Description for Password Fields that do not have rules

YatharthVyas:
Pull Request for Issue #33528
This also fixes an issue where Passwords fields without `rules="true"` had to include a misleading attribute: `hiddenDescription="true"` so that they could prevent the description being shown twice

**Edit**: Removed the dependency of rules on description as discussed in the conversation below. I've modified some of the testing conditions accordingly
Now, the description can be independent of the rules text and both can be used together with different text in a single field

### Summary of Changes
- Password fields use the rules attribute to check for the rules specified in the Description (These include max length, allowed chars, etc)
- If the rules attribute is set to true, then the description and rules are displayed above the password field like:

image

- Along with this, the description is displayed again below the field because of  the echo in renderfield.php layout which is common to all the types of fields https://github.com/joomla/joomla-cms/blob/3cebcc73cd03f015807444ce69e52da0f58c8932/layouts/joomla/form/renderfield.php#L52
- So unless the field has an attribute of hiddenDescription="true", the output turns out to be something like: 
 
image

- The bug occurs when you add a description without a rule. In this case, the description is repeated twice unless you add the hiddenDescription="true" attribute. But the catch here is that when you apply hiddenDescription, then the description that is present below the field is hidden while the description above the password field stays visible (resulting in issue #33528)
- Until now, fields with description and without rules had to use hiddenDescription="true" as a make-do way to show description only once. After this PR is merged, developers will no longer have to worry about why their description is being shown twice and they won't have to add a misleading hiddenDescription attribute to tix this.

#### Test for the fixed Issue:
Admin Panel -> System -> Global Config -> Server-> Check the Password Field under Database Fieldset

#### Test for checking this works globally
<ol>
<li> Make a new form / Edit an Existing form </li>
<li> Add two fields of type password such that:

| Field 1 contains: | Field 2 contains: |
| --- | --- |
| <ul type="none"><li> ✔️ description</li><li>❌ **no** rules attribute</li><li>❌ **no** hidden description attribute</li></ul>  |<ul type="none"><li> ✔️ description</li><li> ✔️ rules="true"</li><li> ❌ **no** hidden description attribute</li></ul>|
</li>
</ol>
<br/>
For Example:

```
<field
	name="password1"
	type="password"
	label="COM_CONFIG_FIELD_DATABASE_PASSWORD_LABEL"
	description="COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC"
	filter="raw"
	autocomplete="off"
	size="30"
	lock="true"
/>

<field
	name="password"
	type="password"
	label="JGLOBAL_PASSWORD"
	rules="true"
	description="COM_CONFIG_FIELD_DATABASE_PASSWORD_DESC"
	autocomplete="new-password"
	class="validate-password-strength"
	filter="raw"
	validate="password"
	strengthmeter="true"
	force="on"
	size="30"
/>
```

| Before | After |
| --- | --- |
| image | image |
| Field 1 shows duplicate descriptions and Field 2 shows description instead of rules because it uses the description text to display rules and the language constants for it is missing a %s | Field 1 is fixed and Field 2 now shows unique text for rules message and description message |

You can also test out different combinations of the following attributes:
- description (valid or empty)
- rules (true or false)
- hiddenDescription (true or false)

### Actual result BEFORE applying this Pull Request (For  #33528)
image

### Expected result AFTER applying this Pull Request (For  #33528)
image

None

### DIFF
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -92,7 +92,7 @@ $attributes = array(
 	strlen($hint) ? 'placeholder="' . htmlspecialchars($hint, ENT_COMPAT, 'UTF-8') . '"' : '',
 	!empty($autocomplete) ? 'autocomplete="' . $autocomplete . '"' : '',
 	!empty($class) ? 'class="form-control ' . $class . '"' : 'class="form-control"',
-	!empty($ariaDescribedBy) ? 'aria-describedby="' . $ariaDescribedBy . '"' : '',
+	!empty($ariaDescribedBy) ? 'aria-describedby="' . trim($ariaDescribedBy) . '"' : '',
 	$readonly ? 'readonly' : '',
 	$disabled ? 'disabled' : '',
 	!empty($size) ? 'size="' . $size . '"' : '',
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -135,7 +135,7 @@ if ($rules && !empty($description))
 	}
 }
 ?>
-<?php if ($rules) : ?>
+<?php if (!empty($description) && $rules) : ?>
 	<div id="<?php echo $name . '-desc'; ?>" class="small text-muted">
 		<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
 	</div>
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -135,13 +135,9 @@ if ($rules && !empty($description))
 	}
 }
 ?>
-<?php if (!empty($description)) : ?>
+<?php if ($rules) : ?>
 	<div id="<?php echo $name . '-desc'; ?>" class="small text-muted">
-		<?php if ($rules) : ?>
-			<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
-		<?php else : ?>
-			<?php echo Text::_($description); ?>
-		<?php endif; ?>
+		<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
 	</div>
 <?php endif; ?>
 
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -136,7 +136,7 @@ if ($rules)
 }
 ?>
 <?php if ($rules) : ?>
-	<div id="<?php echo $name . '-desc'; ?>" class="small text-muted">
+	<div id="<?php echo $name . '-rules'; ?>" class="small text-muted">
 		<?php echo Text::sprintf('JFIELD_PASSWORD_RULES_MINIMUM_REQUIREMENTS', implode(', ', $requirements)); ?>
 	</div>
 <?php endif; ?>
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -85,11 +85,14 @@ if ($lock)
 	$value = '';
 }
 
+$ariaDescribedBy = $rules ? $name . '-rules ' : '';
+$ariaDescribedBy .= !empty($description) ? $name . '-desc' : '';
+
 $attributes = array(
 	strlen($hint) ? 'placeholder="' . htmlspecialchars($hint, ENT_COMPAT, 'UTF-8') . '"' : '',
 	!empty($autocomplete) ? 'autocomplete="' . $autocomplete . '"' : '',
 	!empty($class) ? 'class="form-control ' . $class . '"' : 'class="form-control"',
-	!empty($description) ? 'aria-describedby="' . $name . '-desc"' : '',
+	!empty($ariaDescribedBy) ? 'aria-describedby="' . $ariaDescribedBy . '"' : '',
 	$readonly ? 'readonly' : '',
 	$disabled ? 'disabled' : '',
 	!empty($size) ? 'size="' . $size . '"' : '',
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -105,7 +105,7 @@ $attributes = array(
 	$dataAttribute,
 );
 
-if ($rules && !empty($description))
+if ($rules)
 {
 	$requirements = [];
 
@@ -135,9 +135,9 @@ if ($rules && !empty($description))
 	}
 }
 ?>
-<?php if (!empty($description) && $rules) : ?>
+<?php if ($rules) : ?>
 	<div id="<?php echo $name . '-desc'; ?>" class="small text-muted">
-		<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
+		<?php echo Text::sprintf('JFIELD_PASSWORD_RULES_MINIMUM_REQUIREMENTS', implode(', ', $requirements)); ?>
 	</div>
 <?php endif; ?>
 
--- a/administrator/components/com_config/forms/application.xml
+++ b/administrator/components/com_config/forms/application.xml
@@ -203,7 +203,6 @@
 			filter="raw"
 			autocomplete="off"
 			size="30"
-			hiddenDescription="true"
 			lock="true"
 		/>
 
--- a/administrator/components/com_users/forms/user.xml
+++ b/administrator/components/com_users/forms/user.xml
@@ -21,8 +21,6 @@
 			name="password"
 			type="password"
 			label="JGLOBAL_PASSWORD"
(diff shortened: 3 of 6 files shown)


YatharthVyas:
My latest commit also fixes one more and the only other occurrence of this issue:

Admin Panel -> Plugins -> Authentication - LDAP -> Check the password field

Before:

image


After:

image

layouts/joomla/form/field/password.php
```diff
@@ -135,13 +135,9 @@
...
 ?>
-<?php if (!empty($description)) : ?>
+<?php if ($rules) : ?>
 	<div id="<?php echo $name . '-desc'; ?>" class="small text-muted">
-		<?php if ($rules) : ?>
-			<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
-		<?php else : ?>
-			<?php echo Text::_($description); ?>
-		<?php endif; ?>
+		<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
```
Quy: Remove `description="JFIELD_PASSWORD_RULES_MINIMUM_REQUIREMENTS"` to get the following:

> PHP Notice:  Undefined variable: requirements in \layouts\joomla\form\field\password.php on line 140

YatharthVyas: Good Catch! The reason behind this is that the rules are displayed with the description and I forgot to include a check for `!empty($description)` in `if ($rules)`. I'll commit the fix. Thanks

richard67:
The PR might do what's desired as a visual result, but I'm not sure if it's good design to make the handling of the description dependent on the presence or absence of the rules. Isn't there another way, cleaner way to achieve the same visual result?

It's just a question. I am not a password fields expert or so, so I might be wrong with my concerns.

YatharthVyas:
@richard67 I hope I understood your point correctly, this PR doesn't change how description is normally shown as that is done via renderField.php which is a totally separate file: https://github.com/joomla/joomla-cms/blob/3cebcc73cd03f015807444ce69e52da0f58c8932/layouts/joomla/form/renderfield.php#L52

What this achieves is that it fixes a bug where the description was displayed on the top of the field a second time because of the else block that came in this piece of code:

```php
<?php if (!empty($description)) : ?>
	<div id="<?php echo $name . '-desc'; ?>" class="small text-muted">
		<?php if ($rules) : ?>
			<?php echo Text::sprintf($description, implode(', ', $requirements)); ?>
		<?php else : ?>
			<?php echo Text::_($description); ?>
		<?php endif; ?>
	</div>
<?php endif; ?>
```
This code is taken from password.php and it is present right above the markup for rendering the input tag. 

Before the PR, a second description was rendered if the condition `$rules` was "false" and `$description` was not empty resulting in something like: (refer to the first field), All the existing password fields used a hack where they added the attribute: `hiddenDescription="true"` to hide the bottom description and find the easy way.

image

richard67:
@YatharthVyas No need to explain the issue again. I did fully understand it. I just was asking if there is another way to fix it.

Quy:
Let's not make rules attribute be dependent on description attribute. 

Do this instead and remove the check for `$description`.

`<?php echo Text::sprintf('JFIELD_PASSWORD_RULES_MINIMUM_REQUIREMENTS', implode(', ', $requirements)); ?>`

YatharthVyas:
@richard67 Oh, I apologize. In the time I spent on this, I found that the only possible solution to fix this was by modifying the code that was involved in rendering the rules and omitting the seemingly unnecessary part from there. In terms of a better way to fix this, I am unable to come up with anything. I'll still try to brainstorm a little more around this but I doubt I'll be having any better luck 😞

richard67:
@YatharthVyas No need to apologize. If there is no better way then it's ok. But maybe @Quy 's [suggestion above](https://github.com/joomla/joomla-cms/pull/33536#issuecomment-832091471) could make us find a way?

YatharthVyas:
Thanks, @Quy and @richard67 
Got to learn something new working on this PR! 😃
I was hesitant about mapping to this language string of 'Minimum Requirements:' because I was concerned about Rules involving Maximum Character limit or something similar but turns out that the requirements array that holds all the rules is only based on:
- Min Length
- Min Symbols
- Min Integers
- Min Lowercase
- Min Uppercase

So, 'Minimum' stays common among all

Quy:
Perfect! Please remove where applicable.

```
description="JFIELD_PASSWORD_RULES_MINIMUM_REQUIREMENTS"
hiddenDescription="true"
```

Quy:
There is an issue to address when rules/description are specified where `id` is duplicate.

```
<div id="jform[password]-desc" class="small text-muted">
		<strong>Minimum Requirements</strong> — Characters: 12	</div>
```

```
<div id="jform[password]-desc">
				<small class="form-text">
					<strong>Minimum Requirements</strong> — %s				</small>
			</div>
```

Quy:
Almost there. If there is a rules attribute but no description, then `aria-describedby` should be the id to rules. But I don't how to handle it when both rules/description are specified.

YatharthVyas:
https://stackoverflow.com/questions/53398842/can-i-reference-multiple-elements-using-aria-describedby#:~:text=Yes.,into%20a%20single%20description%20string.

If both rules and description are valid, I could try writing both of them and separate them by a space. As an extra measure, I'll also ensure that a punctuation mark by adding one after %s in the language constant used for rules

Quy:
That is cool! It may not be necessary to add the period for rules so you can decide which is best.

richard67:
Nice team work guys just by the way. That's why I love Joomla, because it's community.

YatharthVyas:
I tested it for all cases and it seems to work fine.

Here's how it behaves when I've entered both rules as well as description:

image

richard67:
@YatharthVyas How does the generated HTML source look like?

I ask because the `$ariaDescribedBy .= !empty($description) ? $name . '-desc' : '';` concatenates without a space in-between, so I would expect it to result in `jform[password]-rulesjform[password]-descr`. Is that the case, and the clever browser fixed that? If so, code should be `$ariaDescribedBy .= !empty($description) ? ' ' . $name . '-desc' : '';`

Update: My suggested change would cause it end in a space if both $rules and $description are empty, so that would not be the perfect fix. It should be refined further.

But maybe I'm wrong and am just missing something, that always can be the case.

YatharthVyas:
@richard67 
There is a space after `-rules` in Line 88 of password.php (files changed) `$ariaDescribedBy = $rules ? $name . '-rules ' : '';`

Here's the markup generated in case both are present: 
```html
<input 
  type="password" 
  name="jform[password]" 
  id="jform_password" 
  value="" autocomplete="new-password"
  class="form-control js-password-strength validate-password-strength"
  aria-describedby="jform[password]-rules jform[password]-desc" 
  size="30" 
  maxlength="99" 
  data-min-length="12"
>
```

However, I'll add a trim() function to ensure that the extra trailing space isn't there incase of a tag with only rules without description

richard67:
> @richard67
> There is a space after `-rules` in Line 68 of password.php (files changed) `$ariaDescribedBy = $rules ? $name . '-rules ' : '';`

Seems I need glasses ;-) Haven't seen that space.

Quy:
I have tested this item :white_check_mark: successfully on 07694c52619619d9f52abafbed04d885b9ab7045<br /><br />Awesome work! Thank you!!!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33536">issues.joomla.org/tracker/joomla-cms/33536</a>.</sub>

richard67:
Maybe it needs to adjust the testing instructions to the changes for other tester(s).

YatharthVyas:
Testing Instructions have been updated.

saumyasarkar11:
I have tested this item :white_check_mark: successfully on 07694c52619619d9f52abafbed04d885b9ab7045<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33536">issues.joomla.org/tracker/joomla-cms/33536</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33536">issues.joomla.org/tracker/joomla-cms/33536</a>.</sub>

chmst:
Thank you all, great team work

PR #33545, opened 2021-05-04
[4.0] a11y - prevent double announcement of readmore

chmst:
Pull Request for Issue # .

### Summary of Changes
Readmore has an aria-label. Additionally, the text on the button is announced by the screenreader. This results in a double announcement. This PR hides the button-text for screenreaders.
Credits: Accessibilty Team

### Testing Instructions
Make Articles with Readmore. Use different variants
- Readmore with title
- Readmore without title
- Readmore with request for login
Check the sourcecode or use a screenreader or another a11y testing tool.  

The text on the Readmore button remains as before. 
But the screenreader announces always the Readmore-text plus titile.

### DIFF
--- a/layouts/joomla/content/readmore.php
+++ b/layouts/joomla/content/readmore.php
@@ -21,26 +21,34 @@ $direction = Factory::getLanguage()->isRtl() ? 'left' : 'right';
 <p class="readmore">
 	<?php if (!$params->get('access-view')) : ?>
 		<a class="btn btn-secondary" href="<?php echo $displayData['link']; ?>" aria-label="<?php echo Text::_('JGLOBAL_REGISTER_TO_READ_MORE') . ' ' . $this->escape($item->title); ?>">
-			<?php echo '<span class="icon-chevron-' . $direction . '" aria-hidden="true"></span>'; ?>
-			<?php echo Text::_('JGLOBAL_REGISTER_TO_READ_MORE'); ?>
+			<span class="aria-hidden">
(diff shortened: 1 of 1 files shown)


brianteeman:
Sorry this is not correct. The reported issue does not occur so this PR is not needed

The presence of the aria-label already ensures that it is the aria label that is announced and not the content of the link _(that is the actual intended purpose of the aria-label)_

Tested with Windows Narrator and NVDA which both only announce the aria-label

Tested with chrome browser which clearly shows that the only thing that is in the accessibility tree is the aria-label

image

(oops wrong image originally)

brianteeman:
It would be really interesting to know which testing tool and screen reader that have been used

brianteeman:
and to just so you dont have to take my word for it. the following link has been tested with most known browsers, os and screen reader combinations.

## Test link
     `   <a href='4567.html' aria-label='this is an aria label'>Click Here</a>`

## Test results with audio
https://www.powermapper.com/tests/screen-readers/labelling/a-aria-label/

chmst:
Thank you for testing! I don't hear the screenreader :)

brianteeman:
@chmst if you use nvda then you can have a speech viewer window. Everything spoken is printed in that window. It's what I use.

brianteeman:
This doc explains it https://www.accessibility-developer-guide.com/setup/screen-readers/nvda/#speech-viewer

PS I really want to know how the JAT thought this was a problem and what software they were using that indicated it was a problem so that we can prevent this happening again and to save you time writing code

PR #33578, opened 2021-05-06
[4.0] Atum preview images

brianteeman:
Updates the images used for the thumbnail and preview of the atum template after the repaint


wilsonge:
Thanks!

brianteeman:
thank you

PR #33558, opened 2021-05-05
[3.10] Correctly handle objects when storing data in tables with _jsonEncode being set

richard67:
Pull Request for Issue #33553 .

### Summary of Changes

Back-integrate from 4.0-dev to here (3.10-dev) the following changes from PR #25761 for fixing bind() with objects in the JTable class:
- Change order of processing in the bind() routine and add a `get_object_vars` if the `$src` parameter is an object, so that it is:
1. Check if the source value is an array or object and throw an exception if not.
2. If the ignore value is a string, explode it over spaces.
3. If the source value is an object, get its accessible properties with `get_object_vars`.
4. JSON encode any fields if required.
5. Bind the source value, excluding the ignored fields.

- Add a unit test for bind() with an object.

I'm wondering why the 4th step in the above list currently comes before the 1st without this PR.

### Testing Instructions

Step 1: On a current 3.10-dev branch or latest 3.10 nightly build, switch on ""Debug System" and set "Error Reporting" to "Maximum" or "Development".

Step 2: Add the following code to some administrator view:
```
// Test PR-33558 begin
JTable::addIncludePath(JPATH_ROOT . '/administrator/components/com_contact/tables');

$row1 = JTable::getInstance('Contact', 'ContactTable');

$contact1 = [
	'name'      => 'Bat Man',
	'alias'     => 'bat-man',
	'catid'     => 4,
	'params'    => ['show_contact_list' => 0, 'show_tags' => ''],
	'language'  => '*',
	'published' => 1,
	'access'    => 1,
];

$row1->bind($contact1);
$row1->store();

echo '<div>Contact 1 saved.<div>';

$row2 = JTable::getInstance('Contact', 'ContactTable');

$contact2            = new stdClass;
$contact2->name      = 'John Doe';
$contact2->alias     = 'john-doe';
$contact2->catid     = 4;
$contact2->params    = ['show_contact_list' => 0, 'show_tags' => ''];
$contact2->language  = '*';
$contact2->published = 1;
$contact2->access    = 1;

$row2->bind($contact2);
$row2->store();

echo '<div>Contact 2 saved.<div>';
// Test PR-33558 end
```
E.g. you can add it just before the following line in the users view's default template:
https://github.com/joomla/joomla-cms/blob/3.10-dev/administrator/components/com_users/views/users/tmpl/default.php#L20

The code creates or updates 2 new contact records in database, using bind() and store(), one contact given as array and the other one as object (stdClass).

Step 3: Navigate to the view modified in the previous step.
[...]

Issue #33553, opened 2021-05-05
[3.9] _jsonEncode in table not working with object


Flowman:
### Steps to reproduce the issue

When storing a object to the table and _jsonEncode is set on the table, it will fail with 'Cannot use object of type stdClass as array'.

It is because only supports arrays but the rest of the bind function supports both arrays and objects. https://github.com/joomla/joomla-cms/blob/2b7a904b9f981a4526e7ffa05c4aa83bbfc39617/libraries/src/Table/Table.php#L602

This has been fixed in 4.0

### Expected result

Storing the data should be successful. 

### Actual result

Error 'Cannot use object of type stdClass as array'

### System information (as much as possible)

Joomla 3.9.26

richard67:
In J4 it was fixed with #25761 . Unfortunately that PR doesn't tell anything about it in its title or description, it just mentions tests being added :-(
We should back-integrate the fix for handling objects from that PR into 3.10-dev.
I'll prepare a PR soon.

richard67:
The pull request with a fix for this issue for 3.10 is almost ready. It only needs testing instructions, but that would need to write some little PHP class and test program. @Flowman  Do you have some extension which we could use for testing?

richard67:
PR is #33558 .

Flowman:
@richard67 not really. I got this quiz or lms component I am building but its to large. 

But I know the changes you are doing is working, as I did the same hack to the core on my test site.

richard67:
Closing as having a pull request. Please test #33558 . Thanks in advance.

richard67:
New pull request is #33633 for 3.9. Please test. Thanks in advance.

PR #25761, opened 2019-08-01
[4.0] Adding further tests for Joomla\CMS\Table\Table::bind()

Hackwar:
This adds further tests for Table::bind(). These tests uncovered 3 issues in that method:

1. If neither an array nor an object is given to bind, the ```onTableBeforeBind``` event throws an exception, but not the ```InvalidArgumentException``` that ```bind()``` is supposed to throw. So the check for the "validity" of the input data is pulled to the front of the method.
2. Likewise the $ignore parameter. This can be both an array or a space separated string of field names. If you use that string variant, the events throw an error of invalid input.
3. If a field is marked as JSON encoded, but the input is an object, this code will fail because it treats the object as an array.

I still have to find a way to test the events.

PR #33633, opened 2021-05-07
[3.9] Correctly handle objects when storing data in tables with _jsonEncode being set

richard67:
Pull Request for Issue #33553 .

### Summary of Changes

Back-integrate from 4.0-dev to here (staging) the following changes from PR #25761 for fixing bind() with objects in the JTable class:
- Change order of processing in the bind() routine and add a `get_object_vars` if the `$src` parameter is an object, so that it is:
1. Check if the source value is an array or object and throw an exception if not.
2. If the ignore value is a string, explode it over spaces.
3. If the source value is an object, get its accessible properties with `get_object_vars`.
4. JSON encode any fields if required.
5. Bind the source value, excluding the ignored fields.

- Add a unit test for bind() with an object.

I'm wondering why the 4th step in the above list currently comes before the 1st without this PR.

### Testing Instructions

Step 1: On a current staging branch or latest 3.9.x nightly build, switch on ""Debug System" and set "Error Reporting" to "Maximum" or "Development".

Step 2: Add the following code to some administrator view:
```
// Test PR-33558 begin
JTable::addIncludePath(JPATH_ROOT . '/administrator/components/com_contact/tables');

$row1 = JTable::getInstance('Contact', 'ContactTable');

$contact1 = [
	'name'      => 'Bat Man',
	'alias'     => 'bat-man',
	'catid'     => 4,
	'params'    => ['show_contact_list' => 0, 'show_tags' => ''],
	'language'  => '*',
	'published' => 1,
	'access'    => 1,
];

$row1->bind($contact1);
$row1->store();

echo '<div>Contact 1 saved.<div>';

$row2 = JTable::getInstance('Contact', 'ContactTable');

$contact2            = new stdClass;
$contact2->name      = 'John Doe';
$contact2->alias     = 'john-doe';
$contact2->catid     = 4;
$contact2->params    = ['show_contact_list' => 0, 'show_tags' => ''];
$contact2->language  = '*';
$contact2->published = 1;
$contact2->access    = 1;

$row2->bind($contact2);
$row2->store();

echo '<div>Contact 2 saved.<div>';
// Test PR-33558 end
```
E.g. you can add it just before the following line in the users view's default template:
https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_users/views/users/tmpl/default.php#L20

The code creates or updates 2 new contact records in database, using bind() and store(), one contact given as array and the other one as object (stdClass).

Step 3: Navigate to the view modified in the previous step.
In my example: "Users -> Manage" (`administrator/index.php?option=com_users&view=users`).

Result: Error 'Cannot use object of type stdClass as array', see issue #33553 and section "Actual result BEFORE applying this Pull Request" below.

Step 4: Check in database if the two contacts have been created.

Result: Only the first contact "Bat Man" has been created.

Step 5: Delete that contact.

Step 6: Apply the patch of this PR.

Step 7: Navigate again to the view modified in the previous step, or reload the page if still there.

Result: No error, the debug output is shown, see section "Expected result AFTER applying this Pull Request" below.

Step 8: Check in database if the two contacts have been created.

Result: Both contacts "Bat Man" and "John Doe" have been created.

Storing data fails with error 'Cannot use object of type stdClass as array'.

The line number may be different to the one shown in the screenshot because that screenshot was made when testing PR #33558 which I had made for 3.10).

Storing data succeeds.

The debug output is shown.

None.

### Additional information

This change doesn't need to be merged up into 4.0-dev since it has been applied there already, see PR #25761 . The unit tests are implemented differently there.

I.e. when merging this up and getting conflicts, just chose to use the existing 4.0-dev code for these when resolving.

### DIFF
--- a/libraries/src/Table/Table.php
+++ b/libraries/src/Table/Table.php
@@ -594,18 +594,6 @@ abstract class Table extends \JObject implements \JObservableInterface, \JTableI
 	 */
 	public function bind($src, $ignore = array())
 	{
-		// JSON encode any fields required
-		if (!empty($this->_jsonEncode))
-		{
-			foreach ($this->_jsonEncode as $field)
-			{
-				if (isset($src[$field]) && is_array($src[$field]))
-				{
-					$src[$field] = json_encode($src[$field]);
-				}
-			}
-		}
-
 		// Check if the source value is an array or object
 		if (!is_object($src) && !is_array($src))
 		{
@@ -618,16 +606,28 @@ abstract class Table extends \JObject implements \JObservableInterface, \JTableI
 			);
 		}
 
+		// If the ignore value is a string, explode it over spaces.
+		if (!is_array($ignore))
+		{
+			$ignore = explode(' ', $ignore);
+		}
+
 		// If the source value is an object, get its accessible properties.
 		if (is_object($src))
 		{
 			$src = get_object_vars($src);
 		}
 
-		// If the ignore value is a string, explode it over spaces.
-		if (!is_array($ignore))
+		// JSON encode any fields required
+		if (!empty($this->_jsonEncode))
 		{
-			$ignore = explode(' ', $ignore);
+			foreach ($this->_jsonEncode as $field)
+			{
+				if (isset($src[$field]) && is_array($src[$field]))
+				{
+					$src[$field] = json_encode($src[$field]);
+				}
+			}
 		}
 
 		// Bind the source value, excluding the ignored fields.
--- a/tests/unit/suites/libraries/joomla/table/JTableTest.php
+++ b/tests/unit/suites/libraries/joomla/table/JTableTest.php
@@ -356,6 +356,48 @@ class JTableTest extends TestCaseDatabase
 	}
 
 	/**
+	 * Test for bind method with object.
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function testBindWithObject()
+	{
+		TestReflection::setValue($this->object, '_jsonEncode', array('params'));
+
+		$data = new stdClass;
+		$data->id1    = 25;
+		$data->id2    = 50;
+		$data->title  = 'My Title';
+		$data->params = array('param1' => 'value1', 'param2' => 25);
+
+		$this->object->bind($data);
+
+		$this->assertEquals(
+			25,
+			$this->object->id1
+		);
+
+		$this->assertEquals(
+			50,
+			$this->object->id2
+		);
+
+		$this->assertEquals(
+			'My Title',
+			$this->object->title
+		);
+
+		// Check the object is json encoded properly
+		$this->assertEquals(
+			'{"param1":"value1","param2":25}',
+			$this->object->params,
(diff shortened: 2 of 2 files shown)


joomdonation:
I have tested this item :white_check_mark: successfully on 7424cc40392d652b0825d1bf406ac0045474f61a<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33633">issues.joomla.org/tracker/joomla-cms/33633</a>.</sub>

libraries/src/Table/Table.php
```diff
@@ -618,16 +606,28 @@ public function bind($src, $ignore = array())
...
 		// If the source value is an object, get its accessible properties.
 		if (is_object($src))
 		{
 			$src = get_object_vars($src);
 		}
 
-		// If the ignore value is a string, explode it over spaces.
-		if (!is_array($ignore))
+		// JSON encode any fields required
+		if (!empty($this->_jsonEncode))
```
joomdonation, line 622: Maybe we should add is_array check here, so the could would be:

```php
if (is_array($this->_jsonEncode) && !empty($this->_jsonEncode))
```

richard67: why? an exception is raised it it's not an array, and then the values are extracted before. at this point it can not be anything else than an array.

joomdonation: No, there is not exception. If for some stupid reason, someone set $this->_jsonEncode for a table class to a string like 'params', there would be a warning message like this displayed:

`Warning: Invalid argument supplied for foreach() in D:\www\joomla\libraries\src\Table\Table.php on line 624`

richard67: And what is this then? https://github.com/joomla/joomla-cms/pull/33633/files#diff-052e470aee4351c5ff7b15c0ac2a9ac5e71c211729023bd607d1688e7e2d3adcR598

richard67: Ah wait you are talking about the params only.

richard67: I think that can be fixed by some other PR.

joomdonation: That's not important. You can change it if you want, otherwise, just leave it as how it is. I won't complain :).

alikon:
I have tested this item :white_check_mark: successfully on 7424cc40392d652b0825d1bf406ac0045474f61a<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33633">issues.joomla.org/tracker/joomla-cms/33633</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33633">issues.joomla.org/tracker/joomla-cms/33633</a>.</sub>

HLeithner:
thanks

PR #33685, opened 2021-05-09
[4.0] Remove BannersHelper class

joomdonation:
Pull Request for Issue # .

### Summary of Changes
The class `BannersHelper` is just an alias for `Joomla\Component\Banners\Administrator\Helper\BannersHelper` class and from I can see, it is not used anywhere, so I propose to remove it to to make the component looks a bit cleaner (it is strange to see a blank class extends the actual namespace class like that)

### Testing Instructions
Testing is not needed. Leave it to @wilsonge to decide if he wants to accept this removal PR.


joomdonation:
Ah, just saw the deprecated message, we could not remove it until 5.0

PR #33720, opened 2021-05-09
[3] Improve indexing performance of Smart Search

Hackwar:
### Summary of Changes
This PR improves the indexing performance of Smart Search in Joomla 3 by inserting smaller chunks into the database at once and generally using a slightly different receipe to handle the insertion of terms into the database. In my small tests here, it reduced the time to index the content of the testing data from around 16s to 13s.

### Testing Instructions
Use a website with content and smart search setup. The more content, the better. In the backend, go to the Smart Search component and clear the index. Go into the command line and call `php cli/finder_indexer.php` to run the indexing process. Notice the time it takes. Run these steps several times to get a reliable time for the indexing process.
Now apply the patch and again do these steps several times.

Content is indexed, but it takes long.

Content is still indexed, but it is about 20% quicker.

### DIFF
--- a/administrator/components/com_finder/helpers/indexer/indexer.php
+++ b/administrator/components/com_finder/helpers/indexer/indexer.php
@@ -518,32 +518,34 @@ abstract class FinderIndexer
 
 		if (($tokens instanceof FinderIndexerToken) === false)
 		{
-			// Break into chunks of no more than 1000 items
-			$chunks = count($tokens) > 1000
-				? array_chunk($tokens, 1000)
-				: array($tokens);
+			// Cloning a new query template is twice as fast as calling the clear function
+			$query = clone $this->addTokensToDbQueryTemplate;
 
-			foreach ($chunks as $chunkTokens)
+			// Iterate through the tokens to create SQL value sets.
+			foreach ($tokens as $token)
 			{
-				// Cloning a new query template is twice as fast as calling the clear function
-				$query = clone $this->addTokensToDbQueryTemplate;
+				$query->values(
+					$db->quote($token->term) . ', '
+					. $db->quote($token->stem) . ', '
+					. (int) $token->common . ', '
+					. (int) $token->phrase . ', '
+					. $db->escape((float) $token->weight) . ', '
+					. (int) $context . ', '
+					. $db->quote($token->language)
+				);
+				$values++;
 
-				// Iterate through the tokens to create SQL value sets.
-				foreach ($chunkTokens as $token)
+				if ($values > 0 && ($values % 128) == 0)
 				{
-					$query->values(
-						$db->quote($token->term) . ', '
-						. $db->quote($token->stem) . ', '
-						. (int) $token->common . ', '
-						. (int) $token->phrase . ', '
-						. $db->escape((float) $token->weight) . ', '
-						. (int) $context . ', '
-						. $db->quote($token->language)
-					);
-					++$values;
-				}
+					$db->setQuery($query)->execute();
+					$query->clear('values');
 
-				$db->setQuery($query)->execute();
+					// Check if we're approaching the memory limit of the token table.
+					if ($values > static::$state->options->get('memory_table_limit', 10000))
+					{
+						$this->toggleTables(false);
+					}
+				}
 			}
 		}
 		else


richard67:
@rigin Could you test this PR here? Thanks in advance.

rigin:
I will deploy another local copy of the site and experiment on it for some time.
After that, I will publish the report here.

richard67:
> I will deploy another local copy of the site and experiment on it for some time.
> After that, I will publish the report here.

@rigin Thanks in advance. I don't expect all to be solved what you have mentioned. But there should be a performance gain like described above in the description.

jsubri:
I have tested this item :white_check_mark: successfully on fa47381311f921378ad84338e76802e4c6e54cd0<br /><br />Before
Total Processing Time: 65.279 seconds
Peak memory usage: 33,554,432 bytes

After:
Total Processing Time: 47.213 seconds
Peak memory usage: 33,554,432 bytes

Batch 1 & 2 are sightly slower, then batches 3 - 8 are up to 40% faster
Thank you<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33720">issues.joomla.org/tracker/joomla-cms/33720</a>.</sub>

jsubri:
after a second run Batch 1 & 2 are not slower, all good :-)

rigin:
I am currently making a series of 3 measurements.
Tomorrow or the day after tomorrow it will be ready.

alikon:
I have tested this item :white_check_mark: successfully on fa47381311f921378ad84338e76802e4c6e54cd0<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33720">issues.joomla.org/tracker/joomla-cms/33720</a>.</sub>

alikon:
i'll wait the test result from @rigin  for setting this rtc

rigin:
Good afternoon.

Measurement results.

The original version.

*****************************
Total Processing Time: 48880.113 seconds.
Peak memory usage: 25,165,824 bytes

Total Processing Time: 48767.392 seconds.
Peak memory usage: 25,165,824 bytes

Total Processing Time: 45462.329 seconds.
Peak memory usage: 23,068,672 bytes

Total Processing Time: 44495.304 seconds.
Peak memory usage: 23,068,672 bytes
***************************
After the changes.

*******************
Total Processing Time: 38031.2 seconds.
Peak memory usage: 23,068,672 bytes

Total Processing Time: 37066.64 seconds.
Peak memory usage: 23,068,672 bytes

Total Processing Time: 37427.362 seconds.
Peak memory usage: 23,068,672 bytes
****************

Sorry for the delay. During the measurement process, my MariaDB settings were reset to the default state. (just a coincidence) I had to repeat the tests. But, the results have not changed much.


Also, I tested my database using the mysqltuner utility.

It issues a recommendation there.

join_buffer_size (> 100.0M, or always use indexes with joins)

Hackwar:
Thank you for your tests. I'd say that supports this PR. 😉 @alikon can you set this to RTC then?

rigin:
I have a question.

Is it possible to modify the cli script so that it performs indexing when the content plugin is disabled?

So that the user can save materials without indexing, and simultaneously conduct the indexing process using the cli script.

And also, whether it is possible to change the algorithm for selecting content for indexing, so that it discards materials that have a change time older than the time of index creation. So that the cli script indexes only the changed materials.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33720">issues.joomla.org/tracker/joomla-cms/33720</a>.</sub>

Hackwar:
I've not tested this in Joomla 3, but in Joomla 4 you don't need to have the content plugin enabled for the CLI script to work.

The indexer will not reindex content items which have not changed since last indexing. This is done through calculating the content to index and then comparing an MD5 hash of the current object and the one previously indexed. Processing the items to the point where the indexer can index them will take some time, but it should be at least 10 times faster than actually indexing it completely. Unfortunately the last-changed timestamp is not reliable and is not present in every component. You can do optimisations like this in custom finder plugins for your specific site if you want.

rigin:
I tested, the script just scrolls the content without indexing.

rigin:
#You can do optimisations like this in custom finder plugins for your specific site if you want. 

OK, I'll try it as soon as I have some free time.

Fedik:
> I tested, the script just scrolls the content without indexing.

you need to disable "Content - Smart Search" plugin, it responsible for indexing "on save".
but the plugin "Smart Search - Content" must be enabled, it responsible for Content indexing in general.

rigin:
Yes, you're right, I was wrong. This works for the cli.

rigin:
I figured out the reasons for the brakes.
I had the "Table Memory limit" parameter set too low in the smart search settings.
And accordingly, the #__finder_tokens and #__finder_tokens_aggregate tables were switched to MyISAM mode.

Here I described the situation and ways to solve it (in Russian)

https://rigin.net/joomla/obshchie-voprosy/oshibka-indeksirovaniya-the-table-finder-tokens-is-full.html

HLeithner:
thanks

rigin:
When indexing for smart search, sometimes there is an overflow error in the temporary tables #__finder_tokens and #__finder_tokens_aggregate .

These tables are designed for caching temporary data and have the MEMORY type, and it is banal that this memory itself is not enough.

MYSQL just ran out of memory for placing data in the heap. This can be easily changed in the MYSQL settings (in Linux, the default is /etc/my. cnf). The max_heap_table_size parameter is responsible for this-the maximum allowed size of the table stored in memory (of the MEMORY type). Its default value is 16 MB. Change it to max_heap_table_size=256M (for example)

However, there is an additional one. complexity, purely administrative-this parameter is usually the responsibility of the hosting administrator )))

But if you write the right letter, it will be changed there.


**Using the "Table Memory Limit" parameter.**

In the Smart search settings, on the Indexing tab, there is a parameter called "Table memory limit". Its essence is to prevent this error, although at the expense of a sharp decrease in the speed of the algorithm.

The indexing algorithm takes into account the number of records in this temporary table when scanning each material, and if this number exceeds this parameter, the table is converted to the MyISAM type, which prevents this error, but dramatically reduces the speed of the algorithm.

According to my estimates, each record in these tables requires approximately 1.2-1.5 KB per record, so this parameter can be set based on an approximate estimate - the value of max_heap_table_size divided by 1.5. If indexing works without errors, then the parameter can be increased, if errors occur, then reduced.

Shorter-it should be the maximum at which indexing errors do not occur.

Issue #33629, opened 2021-05-07
[4.0] [a11y] atum - links in lists have no underline 


chmst:
### Steps to reproduce the issue
Have a look on the system view or on other dashboards where lists and tables are mixed.

### Expected result
All links are underlined.

### Actual result
![grafik](https://user-images.githubusercontent.com/1035262/117438227-900d9680-af31-11eb-9823-9c01a9160985.png)

ciar4n:
Thank you @chmst. PR created.. #33808

PR #33752, opened 2021-05-10
[4.0] [a11y] atum Underlined links

himanshu007-creator:
Pull Request for Issue #33629 .

### Summary of Changes
The links in list earlier were not underlined.

### Testing Instructions
Apply this patch , compile css and have a look on links in dashboard 

Links were not underlined

Links are underlined

p2

### DIFF
--- a/administrator/templates/atum/scss/pages/_com_cpanel.scss
+++ b/administrator/templates/atum/scss/pages/_com_cpanel.scss
@@ -174,6 +174,10 @@
     &:hover {
       background: var(--toolbar-bg);
     }
+    
+   a[class^="flex-grow-1"] {
+     text-decoration: underline;
+    }
 
     .list-group-item a > span {
       &[class^="#{$jicon-css-prefix}-"],
--- a/administrator/templates/atum/scss/pages/_com_cpanel.scss
+++ b/administrator/templates/atum/scss/pages/_com_cpanel.scss
@@ -175,8 +175,8 @@
       background: var(--toolbar-bg);
     }
     
-   a[class^="flex-grow-1"] {
-     text-decoration: underline;
+    a[class^="flex-grow-1"] {
+      text-decoration: underline;
     }
 
     .list-group-item a > span {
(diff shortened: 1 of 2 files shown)


brianteeman:
Sorry but this is completely wrong. You should be using css here.

richard67:
@himanshu007-creator Sorry, but this PR is wrong. We don't use anywhere the `<u>` element to underline links, we use CSS for that.

Furthermore, as you can read here, it shall not be used for underlining links: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u

> Be careful to avoid using the `<u>` element with its default styling (of underlined text) in such a way as to be confused with a hyperlink, which is also underlined by default.

richard67:
@brianteeman Seems you were a few seconds faster than me ;-)

himanshu007-creator:
@richard67  @brianteeman  thanks for speedy feedback, really appreciate that!. I get your point. I will try to make changes using CSS only. I will surely work on it tomorrow :)

himanshu007-creator:
I have used BS5 <code>text-decoration-underline</code> for this. <a href="https://getbootstrap.com/docs/5.0/utilities/text/#text-decoration"><i>reference</i></a> 
is it good to go, or do i have to make changes in CSS itself?

brianteeman:
@himanshu007-creator this is how I would approach this problem.

1. Look at the other underlined links and see how they have been styled
2. Compare the two and then work to apply the **same** styling

richard67:
My gut feeling is that the styling is already somewhere, and it just needs to modify the CSS selector for it.

himanshu007-creator:
@richard67  i am thinking of this 
<code>   a[class^="flex-grow-1"] {
      text-decoration: underline;
    }
</code>

in file administrator/templates/atum/scss/pages/_com_cpanel.scss

administrator/templates/atum/scss/pages/_com_cpanel.scss
```diff
@@ -174,6 +174,10 @@
     &:hover {
       background: var(--toolbar-bg);
     }
+    
+   a[class^="flex-grow-1"] {
+     text-decoration: underline;
```
richard67:
```suggestion
    a[class^="flex-grow-1"] {
      text-decoration: underline;
```
Please indent in the same way as you can see in previous code parts. Code style is important for quick readability and so maintainability of source code.

Quy:
Closing in favor of PR #33808. Thanks.

Issue #33692, opened 2021-05-09
[4.0] Backend: Admin modules use language filter of site modules


ghost:
### Steps to reproduce the issue

1. Site Modules: Select Language English
2. Switch to Admin Module: Only Modules for English are shown.

For language filtering of admin modules the default-option is set to `No`.

The screencapture show:
1. Site Module without filter
2. Admin Module without filter
3. Site Module with language filter english
4. Admin Module 
5. Site Module with language filter persian
6. Admin Module

![1](https://user-images.githubusercontent.com/78906980/117571403-ad2b9c00-b0ce-11eb-83b5-1a8a95fdeb1a.gif)

### Expected result
Admin Module for all languages are shown.

### Actual result
Only admin modules are shown for language selected for site modules.

### System information (as much as possible)
> ![image](https://user-images.githubusercontent.com/78906980/117570867-6e94e200-b0cc-11eb-826d-9d9fb2202cd0.png)

### Additional comments

In j3 it works as expected.

Maybe its expected but its not possible to set different language filter for site and admin. I guess it's not needed.

chmst:
The filter is broken in multilingual sites. If the filter is extended in admin modus, the language filter is missing, the language is still in the state, but reset of the filter is not possible

joomdonation:
Please test PR https://github.com/joomla/joomla-cms/pull/33763

ghost:
Reopen please as pull request is closed.

alikon:
pr reopened no need to reopen this issue

PR #33763, opened 2021-05-11
[4.0] Fix Modules Filter

joomdonation:
Pull Request for Issue #33692.

### Summary of Changes
Currently, we use same session key for Admin and Site modules filter options value, thus causes unexpected behavior. One of the issue is https://github.com/joomla/joomla-cms/issues/33692 

There are other issues, too. For example, if you are on Site Modules and select a module position filter, then switch to Administrator, you won't see any modules... (because the system remember that site position, but of course, backend modules have different positions list. Same for module filter and menu item filter.

### Testing Instructions
1. Look at the issue https://github.com/joomla/joomla-cms/issues/33692 , confirm that the issue happens (you would need to have a multilingual website to confirm the issue)
2. Apply patch, confirm that the issue sorted.
3. Try to filter modules by position, by module types and make sure it is still working as expected, nothing should be broken
4. From Modules Management screen, choose a module position. Then press New button in toolbar, confirm that that module position is pre-selected on New module screen.

### DIFF
--- a/administrator/components/com_modules/src/Model/ModuleModel.php
+++ b/administrator/components/com_modules/src/Model/ModuleModel.php
@@ -639,7 +639,8 @@ class ModuleModel extends AdminModel
 			// Pre-select some filters (Status, Module Position, Language, Access Level) in edit form if those have been selected in Module Manager
 			if (!$data->id)
 			{
-				$filters = (array) $app->getUserState('com_modules.modules.filter');
+				$clientId = $app->input->getInt('client_id', 0);
+				$filters  = (array) $app->getUserState('com_modules.modules.' . $clientId . '.filter');
 				$data->set('published', $app->input->getInt('published', ((isset($filters['state']) && $filters['state'] !== '') ? $filters['state'] : null)));
 				$data->set('position', $app->input->getInt('position', (!empty($filters['position']) ? $filters['position'] : null)));
 				$data->set('language', $app->input->getString('language', (!empty($filters['language']) ? $filters['language'] : null)));
--- a/administrator/components/com_modules/src/Model/ModulesModel.php
+++ b/administrator/components/com_modules/src/Model/ModulesModel.php
@@ -87,7 +87,7 @@ class ModulesModel extends ListModel
 		{
 			$this->context .= '.' . $layout;
 		}
-		
+
 		// Make context client aware
 		$this->context .= '.' . $app->input->get->getInt('client_id', 0);
 
@@ -120,7 +120,7 @@ class ModulesModel extends ListModel
 		else
 		{
 			$clientId = (int) $this->getUserStateFromRequest($this->context . '.client_id', 'client_id', 0, 'int');
-			$clientId = (!in_array($clientId, array (0, 1))) ? 0 : $clientId;
+			$clientId = (!in_array($clientId, array(0, 1))) ? 0 : $clientId;
 			$this->setState('client_id', $clientId);
 		}
 
--- a/administrator/components/com_modules/src/Model/ModulesModel.php
+++ b/administrator/components/com_modules/src/Model/ModulesModel.php
@@ -87,6 +87,29 @@ class ModulesModel extends ListModel
 		{
 			$this->context .= '.' . $layout;
 		}
+		
+		// Make context client aware
+		$this->context .= '.' . $app->input->get->getInt('client_id', 0);
+
+		// Load the filter state.
+		$this->setState('filter.search', $this->getUserStateFromRequest($this->context . '.filter.search', 'filter_search', '', 'string'));
+		$this->setState('filter.position', $this->getUserStateFromRequest($this->context . '.filter.position', 'filter_position', '', 'string'));
+		$this->setState('filter.module', $this->getUserStateFromRequest($this->context . '.filter.module', 'filter_module', '', 'string'));
+		$this->setState('filter.menuitem', $this->getUserStateFromRequest($this->context . '.filter.menuitem', 'filter_menuitem', '', 'cmd'));
+		$this->setState('filter.access', $this->getUserStateFromRequest($this->context . '.filter.access', 'filter_access', '', 'cmd'));
+
+		// If in modal layout on the frontend, state and language are always forced.
+		if ($app->isClient('site') && $layout === 'modal')
+		{
+			$this->setState('filter.language', 'current');
+			$this->setState('filter.state', 1);
+		}
+		// If in backend (modal or not) we get the same fields from the user request.
+		else
+		{
+			$this->setState('filter.language', $this->getUserStateFromRequest($this->context . '.filter.language', 'filter_language', '', 'string'));
+			$this->setState('filter.state', $this->getUserStateFromRequest($this->context . '.filter.state', 'filter_state', '', 'string'));
+		}
 
 		// Special case for the client id.
 		if ($app->isClient('site') || $layout === 'modal')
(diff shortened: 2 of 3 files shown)


ghost:
I have tested this item :red_circle: unsuccessfully on 16f40f9343cd875934cab7c300daf72bd12a14d6<br /><br />Test using patchtester got same behaviour with and without pr. Test filter on language and position.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33763">issues.joomla.org/tracker/joomla-cms/33763</a>.</sub>

joomdonation:
Ah, no. The fix is not this simple. I will see if I can find a solution

joomdonation:
Close this for now. When I come up with a solution, I will re-open.

joomdonation:
@sandramay0905 Could you please give it another try?

ghost:
I have tested this item :white_check_mark: successfully on f449246adbe5181786568d7754e26d560c486e81<br /><br />Test Filters Site: Language, Position and combination > Admin display of modules is not effected.

Enabled on Admin language filtering: Filter on Persian and in site on english > works.

Thanks @joomdonation <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33763">issues.joomla.org/tracker/joomla-cms/33763</a>.</sub>

astridx:
Unfortunately, the context is hard coded in some places. These would have to be adapted with this solution.

**One** is here, for example:

https://github.com/joomla/joomla-cms/blob/92aa07576e8c99ad1ddd7abc316e01a0c6413ba8/administrator/components/com_modules/src/Service/HTML/Modules.php#L190

If this is not changed, the following can happen:

1. Make a new installaiton and filter modules



2. Move a module that is the only one on its position to another position.


3. After moving you see no module in the list.

joomdonation:
@astridx Do you know if there are other places affected by this change? I ask because you tried to fix this issue before, so you have more experience and could help.

joomdonation:
@astridx Unless I'm mistaken, I think the issue you are talking about happens before this PR:

- The modules list empty because there are no modules left for that position (after the last module from that position moved to new position)
- What's wrong here is the system still remember the original position (because it is stored in the session), but the positions filter does not contain that position anymore (the positions filter from dropdown only contains positions which has at least one module assigned to it). Maybe we can modify code of this method https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_modules/src/Field/ModulesPositionField.php#L120 , add the missing position (which is stored in the session to the array, same with the code at https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_modules/src/Service/HTML/Modules.php#L188-L197

Something like (quick draft code, clean up needed)
```php

$position = $app->getUserState('com_modules.modules.' . $clientId . '.filter.position');

if ($position)
{
	$found = false;

	// Add active position here
	foreach ($options as $option)
	{
		if ($option->value ===$position)
		{
			$found = true;
			break;
		}
	}

	if (!$found)
	{
		$options[] = \JHtml::_('select.option', $position, $position);
	}
}
```
How do you think about it?

ceford:
I have tested this item :white_check_mark: successfully on ca23363e1d745e08141e045614bcf32c890f0f5e<br /><br />I confirm that switching back and forth between Site and Administrator gives independent module lists.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33763">issues.joomla.org/tracker/joomla-cms/33763</a>.</sub>

astridx:
> @astridx Unless I'm mistaken, I think the issue you are talking about happens before this PR:

Now I'm not sure I understand you correctly. Without this PR, the error described in https://github.com/joomla/joomla-cms/pull/33763#issuecomment-838333863 does not occur, right?




> @astridx Do you know if there are other places affected by this change?

I don't know Joomla well. I searched for "com_modules.modules" and  found this place: https://github.com/joomla/joomla-cms/blob/9653158d1eb14190f66f68233fc44763dd67d884/administrator/components/com_modules/src/Model/ModuleModel.php#L642

But I don't know how important this is and I don't know if there are other places where the context might be put together with variables?

joomdonation:
> Now I'm not sure I understand you correctly. Without this PR, the error described in #33763 (comment) does not occur, right?

Actually, that error happens because this PR. I explained the issue and proposed the solution in my earlier comment https://github.com/joomla/joomla-cms/pull/33763#issuecomment-838805885. But I don't want to include that in this PR. I can make a new PR for that if needed.

> I don't know Joomla well. I searched for "com_modules.modules" and found this place:

=> That's not affected by the PR. It uses to auto-populate data when you create new module base on data from filter. For example, if you choose a Position in Modules Management page, that will be default position when you press New button to create a new module. That still works after this PR.

joomdonation:
@astridx Thinking more about the issue you described, I think after moving the modules to new position, we should set the position filter to that new position. That would be best solution, I think

joomdonation:
>But I don't know how important this is and I don't know if there are other places that might be put together with variables?

Actually, that is affected by this PR and fixed by this comment https://github.com/joomla/joomla-cms/pull/33763/commits/d47154e68a524d85f1d59ae9fbf622fdb98ce5dc

joomdonation:
@sandramay0905 @ceford Could you please help testing it again? There is a new commit which auto-populate data for New Module base on data from filter (step #4 in testing instructions). Compare to your previous test, that new step is needed.

The issue which @astridx mentioned is independent with this PR and I will make a new PR to fix it after this one merged (because the new code will depend on this PR)

Thanks.

ceford:
I have tested this item :white_check_mark: successfully on d47154e68a524d85f1d59ae9fbf622fdb98ce5dc<br /><br />Confirmed that it works and step 4 works.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33763">issues.joomla.org/tracker/joomla-cms/33763</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on d47154e68a524d85f1d59ae9fbf622fdb98ce5dc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33763">issues.joomla.org/tracker/joomla-cms/33763</a>.</sub>

joomdonation:
Thanks @ceford and @sandramay0905 for testing.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33763">issues.joomla.org/tracker/joomla-cms/33763</a>.</sub>

sonnemondundsterne:
Thanks astrid for finding possible issues.

Stuartemk:
It is assumed that (array) should no longer be used, instead it should be [], there are many (array) in the whole Joomla 4 cms and they are still adding more, perhaps there is something that I do not know and justify the use of (array). Please tell me if I am in error and if not, change all the (array) to [] or at least not continue with that obsolete practice.
They are doing a great job !

joomdonation:
@Stuartemk I don't understand your comment. Is this something related to this PR ?

richard67:
@Stuartemk Maybe you mix up type cast `(array) $somevariable` and array creation `array(1,2,3)`? Only the latter is replaced by short array syntax `[1,2,3]` in J4, but here in the diff of this PR I only see a type cast to array, and that's something completely different.

Stuartemk:
Maybe I'm confused, but I think this should be changed to be 100% compatible with php8 I think so, but I can be wrong

$filters  = (array) $app->getUserState('com_modules.modules.' . $clientId . '.filter');

 for this

$filters  = [ ] $app->getUserState('com_modules.modules.' . $clientId . '.filter');

Regards

joomdonation:
@Stuartemk No, it's not valid syntax. For type casting, we will still have to use the original code:

```php
$filters = (array) $app->getUserState('com_modules.modules.' . $clientId . '.filter');
```

richard67:
Nothing else than what I have said.

Quy:
Thank you!

PR #33769, opened 2021-05-11
Makes check for field type="Hidden" case-insensitive

regularlabs:
Fixes issue with field not being seen as hidden when using `type="Hidden"` instead of `type="hidden"`

You can test this by for instance - changing:
```
		<field
			name="update_site_id"
			type="hidden"
		/>
```
to:

```
		<field
			name="update_site_id"
			type="Hidden"
		/>
```
in administrator/components/com_installer/forms/updatesite.xml

Then go to administrator/index.php?option=com_installer&view=updatesites and click on one of the update sites.
Result is now:
image

After fix, that line is not visible (which is good).

### DIFF
--- a/libraries/src/Form/FormField.php
+++ b/libraries/src/Form/FormField.php
@@ -686,7 +686,7 @@ abstract class FormField
 		$this->repeat = ($repeat === 'true' || $repeat === 'multiple' || (!empty($this->form->repeat) && $this->form->repeat == 1));
 
 		// Set the visibility.
-		$this->hidden = ($this->hidden || (string) $element['type'] === 'hidden');
+		$this->hidden = ($this->hidden || strtolower((string) $this->element['type']) === 'hidden');
 
 		$this->layout = !empty($this->element['layout']) ? (string) $this->element['layout'] : $this->layout;
 


brianteeman:
Surely if you are making this case-insensitive then you need to make them all case insensitive?

regularlabs:
Well, all field types are already case-insensitive... kind of.
If the Field classname is `FoobarField`, then the casing doesn't matter.
So you can use `type="list"`, or `type="List"` or even `type="LiSt"`... but not `type="Hidden"`

Depending on your server (whether it can deal with case-insensitive files/folders) you **DO** need to use the exact casing when using other uppercase characters in the classname, like `FooBarField`. 
Then you must use `field="FooBar"` or `field="fooBar"`, but `field="foobar"` will fail.

Joomla uses a mix of the lowercase and title case class name (yes, bit of a mess):
image
image

Anyway, this PR is just to fix the issue with the `type="Hidden"`

regularlabs:
PS: Here you can already see that there are other internal checks on the type that are case-insensitive.
https://github.com/joomla/joomla-cms/blob/ad9eba8fc281b875a2b0d501be8a9213262516a9/libraries/src/Form/FormField.php#L908

brianteeman:
I have tested this item :white_check_mark: successfully on ad9eba8fc281b875a2b0d501be8a9213262516a9<br /><br />Marking as successful that it does what it says. reserving judgement due to lack of knowledge if this is the correct approach<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33769">issues.joomla.org/tracker/joomla-cms/33769</a>.</sub>

ceford:
I have tested this item :white_check_mark: successfully on ad9eba8fc281b875a2b0d501be8a9213262516a9<br /><br />Works as described.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33769">issues.joomla.org/tracker/joomla-cms/33769</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33769">issues.joomla.org/tracker/joomla-cms/33769</a>.</sub>

richard67:
Thanks!

regularlabs:
Thanks!

Issue #33490, opened 2021-05-02
[4] Deleting content maps only deletes a few (a tree branch) at a time when all is selected


PhilETaylor:
### Steps to reproduce the issue

Go to Content Maps -> Select All -> Click Delete

### Expected result

All are deleted

### Actual result

Some are deleted (the first top level parent and its children) only. 

Watch:

https://user-images.githubusercontent.com/400092/116826140-70215000-ab8a-11eb-85db-46aade6d8e19.mp4

ghost:
Confirm the issue.

PhilETaylor:
#33771

PR #33771, opened 2021-05-11
[4.0] Fix Delete All for Content Map

YatharthVyas:
Pull Request for Issue #33490

### Summary of Changes

<hr> 

#### Reason behind this issue

I debugged the array of id(s) which were queried for deletion and their query result:

```
Array of ids to be deleted: [5,4,10,14,13,12,11,15,7,6,9,8,3,2]
Status:
5 - done
4 - done
10 - failed
```

| id | Name | |
| --- | --- | --- |
| 5 | author/yatharthvyas | ✔️ Successfully deleted |
| 4 | author | ✔️ Successfully deleted but since it was a root level map, it also deleted all it's children along with it, so now id=10 was no longer present in the table |
| 10 | author/joomla | ❌ Failed because it was already deleted in the query of it's parent (id: 4).<br>This conditional block for key not in table returned false and ended the execution here skipping all remaining queries in the array. |

<hr> 

#### Changes Made:

The condition that checks if the key is in the table is now shifted outside the loop so when an element is deleted, that doesn't generate any error and return prematurely. Incase any key is not present, then it would be checked before deletion and the execution would stop then and there

<hr> 

#### Before testing, Please take a backup of `#_finder_taxonomy` table to avoid the loss of any data
1. Visit Backend
2. Components -> Smart Search -> Content Map
3. Ensure that you have sufficient data in this table with multiple parent-level and child-level maps (Installting Blog Sample Data will also work)
4. Click on the select all checkbox and proceed to delete them.

Before the PR, only the first parent and its children would get deleted and then the function would return false skipping the remaining elements.
After the PR, you should see the Empty State, ie. All items deleted successfully.

<hr>

Please refer to the video in #33490

<hr>

content_map_fixed

<hr>

None

### DIFF
--- a/administrator/components/com_finder/src/Model/MapsModel.php
+++ b/administrator/components/com_finder/src/Model/MapsModel.php
@@ -97,6 +97,18 @@ class MapsModel extends ListModel
 		// Include the content plugins for the on delete events.
 		PluginHelper::importPlugin('content');
 
+		// Iterate the items to check if all of them exists.
+		foreach ($pks as $i => $pk)
+		{
+			if (!$table->load($pk))
+			{
+				// Item is not in the table.
+				$this->setError($table->getError());
+
+				return false;
+			}
+		}
+
 		// Iterate the items to delete each one.
 		foreach ($pks as $i => $pk)
 		{
@@ -141,12 +153,6 @@ class MapsModel extends ListModel
 						$this->setError(Text::_('JLIB_APPLICATION_ERROR_DELETE_NOT_PERMITTED'));
 					}
 				}
-			}
-			else
-			{
-				$this->setError($table->getError());
-
-				return false;
 			}
 		}
 
--- a/administrator/components/com_finder/src/Model/MapsModel.php
+++ b/administrator/components/com_finder/src/Model/MapsModel.php
@@ -97,7 +97,7 @@ class MapsModel extends ListModel
 		// Include the content plugins for the on delete events.
 		PluginHelper::importPlugin('content');
 
-		// Iterate the items to check if all of them exists.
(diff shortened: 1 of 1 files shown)


administrator/components/com_finder/src/Model/MapsModel.php
```diff
@@ -97,6 +97,18 @@ public function delete(&$pks)
 		// Include the content plugins for the on delete events.
 		PluginHelper::importPlugin('content');
 
+		// Iterate the items to check if all of them exists.
```
brianteeman:
```suggestion
		// Iterate the items to check if all of them exist.
```

brianteeman:
I have tested this item :white_check_mark: successfully on 9b14e159c4bfe2a1f07010eac97e001be56b4b4f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33771">issues.joomla.org/tracker/joomla-cms/33771</a>.</sub>

administrator/components/com_finder/src/Model/MapsModel.php
YatharthVyas: Thanks

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33771">issues.joomla.org/tracker/joomla-cms/33771</a>.</sub>

Quy:
Thank you!

YatharthVyas:
Thanks everyone

PR #33773, opened 2021-05-11
[4.0] Repackage the js of fancy select field

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Remove the choices.js from the vendor folder
- Use `import` in the fancy select field

### Why?
Well, custom elements (and Web Components) are better if they don't depend on other scripts, eg append the HTML markup and the script element should be the only step for using them.

Also, personally, I'm very worried that the library is already abandoned and shipping it in the vendor folder the project creates the same situation that existed with chosen. Let's not do the same mistake. By using import to utilize choices we are encapsulating the whole thing ONLY in the fancy select script source and thus someone could swap the library or completely rewrite the custom element without any dependency. In short, it's more future proof this way...

### Testing Instructions
Apply that patch (needs `npm ci`
Check as many fancy select (the one with the search) in as many edit forms as possible
Screenshot 2021-05-11 at 13 54 15

Works

Works, 1 HTTP request less (but this PR is not about reducing the HTTP requests, HTTP3 is already here and the number of requests is completely irrelevant for performance)

No

@Fedik are you ok with this one?


Fedik:
Yes and no.
Even if you embed the choicesjs in to the custom element, the choicesjs should be available as standalone asset.
Developers may want to use somewhere in their extension.

dgrammatiko:
> Developers may want to use somewhere in their extension.

That's what I want to prevent. The script is abandoned so if Joomla ships with it will end up with the same mess as chosen:

Screenshot 2021-05-11 at 15 30 07

Fedik:
We already ships it, no way back :smile:

dgrammatiko:
Although I'm totally against this as it's very bad to ship dead code, I've reverted that part, so the script will be available in the vendor folder.

ceford:
I have tested this item :white_check_mark: successfully on 417e784fde0b03aedff523a9eafed41626e70149<br /><br />The fancy select fields I tried still work after applying the patch. I was rather expecting the vendor choices to disappear from the source but it is still there
> 	<script src="/j4test/media/vendor/choicesjs/js/choices.js?9.0.1" defer data-asset-name="choicesjs"></script>
I checked that it is not being used for this test (it is not) and guess some other component must be loading it - ->usePreset('chocesjs') is used in a couple of places.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33773">issues.joomla.org/tracker/joomla-cms/33773</a>.</sub>

Fedik:
> must be loading it - ->usePreset('chocesjs') is used in a couple of places.

yea can be,
@dgrammatiko please check where fancy select is included, 
and change `->usePreset('chocesjs')` to `->useStyle('chocesjs')`

ceford:
The file layouts/joomla/form/field/tag.php also seems to need ->useStyle('choicesjs') on line 129. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33773">issues.joomla.org/tracker/joomla-cms/33773</a>.</sub>

ceford:
I think that ->useStyle('choicesjs') needs to be removed everywhere (in atum). The choices.css is in atum/css and the choices.js is now in joomla-field-fancy-select.js - so calling choicesjs just loads the js file from vendor/choicesjs but it is not used. I am a bit out of my depth here!
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33773">issues.joomla.org/tracker/joomla-cms/33773</a>.</sub>

dgrammatiko:
Well, that won't happen for J4 then

PR #33786, opened 2021-05-11
Fixed minor problem in com_cpanel's layout

eopws:
### Summary of Changes

Fixed minor problem in com_cpanel by adding needed css

### Testing Instructions

Go to administrator > com_cpanel and hover on the "Add module to the dashboard" button

The box with cross inside has no changes in color when mouse is on the outer box
before

The box with cross inside has changed its color to white when mouse got on the outer box
after

### DIFF
--- a/administrator/templates/atum/scss/pages/_com_cpanel.scss
+++ b/administrator/templates/atum/scss/pages/_com_cpanel.scss
@@ -158,6 +158,10 @@
       color: hsl(var(--hue),30%,40%);
       margin-inline-end: .5rem;
     }
+
+    &:hover .cpanel-add-module-icon {
+      color: inherit;
+    }
   }
 }
 


brianteeman:
Why do you think the current behaviour is a problem. It is exactly the same as all the dashboard icons

eopws:
> Why do you think the current behaviour is a problem. It is exactly the same as all the dashboard icons

On my opinion when the box is white you can see it better

brianteeman:
So why are you not proposing to change all the icons?

eopws:
> So why are you not proposing to change all the icons?

I didn't spot that all the dashboard icons have the same problem.

eopws:
> So why are you not proposing to change all the icons?

I will open new PR to change all the icons. Thank you for remark.

brianteeman:
Have you not considered that this was a deliberate design decision?

eopws:
> Have you not considered that this was a deliberate design decision?

Maybe it is, but don't you think that it looks a bit weird?

brianteeman:
I am not a designer ;)

Quy:
The focus should be on the text and not the icon.

eopws:
> The focus should be on the text and not the icon.

So we shouldn't change icons, right?

Quy:
I am fine as it, but that is just my preference.

eopws:
I think I touched too subjective topic. Okay, let's wait for more people to tell their opinion.

ceford:
IMHO, with the background turning dark blue on hover, the icons should turn white.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33786">issues.joomla.org/tracker/joomla-cms/33786</a>.</sub>

ghost:
I'm not a designer and let decide @ciar4n (thanks for your work), who repaint the backend.

eopws:
> Have you not considered that this was a deliberate design decision?

It's the way the icons done in joomla 4.1

img

But joomla 4.1 has a lot of changes in design, so we still need to wait for a designer's decision

brianteeman:
> But joomla 4.1 has a lot of changes in design, so we still need to wait for a designer's decision

It shouldnt have any

Issue #33740, opened 2021-05-10
[4][Release Blocker] Unable to cleanly logout of Joomla when using pgsql


PhilETaylor:
### Steps to reproduce the issue

Install Joomla 4 with a Postgres database server
Login to Joomla admin
Attempt to logout.

### Expected result

No errors. Session destroyed correctly. Now back at the login screen.

### Actual result

<img width="1600" alt="Screenshot 2021-05-10 at 14 32 14" src="https://user-images.githubusercontent.com/400092/117667325-b2571c80-b19c-11eb-8f75-26edd0fcf334.png">

### System information (as much as possible)

Postgres

### Additional comments

I believe this is a release blocker, because if this was a mysql issue, and no one could cleanly logout from Joomla admin, then that would for sure be a release blocker and as Joomla advertises itself as being compatible with pgsql then it should be no difference and this should be a major issue to resolve quickly.

Unfortunately Ive not worked on pgsql since my days working at Skype.com and even then I did not really understand it :-( so need someone that does to correctly convert the session id to a string 

Discovered by @alikon white testing a redis session issue here https://github.com/joomla-framework/session/pull/51#issuecomment-836618757 but the redis part is totally unconnected to this issue.

peteruoi:
Maybe totally unrelated maybe not.
06 May 2021
PHP 7.4.19 Released!
The PHP development team announces the immediate availability of PHP 7.4.19. This release reverts a bug related to PDO_pgsql that was introduced in PHP 7.4.18.

The PHP development team announces the immediate availability of PHP 8.0.6. This release reverts a bug related to PDO_pgsql that was introduced in PHP 8.0.5.

PHP 8.0 users that use PDO_pgsql are encouraged to upgrade to this version.

richard67:
Maybe something wrong with template style parameter json in the installation SQL script for PostrgeSQL. Will check later today or tomorrow morning if I can reproduce and if so, find the reason and fix it (if it’s not the PDO issue mentioned above).

PhilETaylor:
@richard67 I think you are confused. This issue is about sessions and the fact pgsql doesn't store the session_id as a string, and `loadColumn` returns resources and not the value of the session_id. 

The "look and feel" of the admin template is forked to a different issue #33741 :)

richard67:
I see. Well I just had in mind that the last SQL stuff we did was for the repaint, so that might have lead me the wrong way.

PhilETaylor:
[...]

PR #33787, opened 2021-05-11
[4] Cast session_id to string before destroying it

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/33740

### Summary of Changes

Cast session_id to a string before we use it 

### Testing Instructions

 https://github.com/joomla/joomla-cms/issues/33740

Cant logout under some circumstances with pgsql

Can logout with no errors under some circumstances with pgsql

### DIFF
--- a/libraries/src/Session/SessionManager.php
+++ b/libraries/src/Session/SessionManager.php
@@ -66,7 +66,7 @@ final class SessionManager
 
 		foreach ($sessionIds as $sessionId)
 		{
-			if (!$this->destroySession($sessionId))
+			if (!$this->destroySession((string) $sessionId))
 			{
 				$result = false;
 			}


PhilETaylor:
Drone issue unrelated.

richard67:
@PhilETaylor Sure that is the solution? When I add a `error_log((string) $sessionId);` just before the changed line and have session handler set to filesystem so I can reproduce the problem, I get in the error log: `Resource id #386`.

With the code change I had proposed in the issue, I did get the correct session ID in the log.

PhilETaylor:
works-on-my-machine

No idea :)

richard67:
Well it doesn't cause that type error. But does it really destroy the session?

richard67:
With my code, the guest session was removed after login, with the code proposed here it isn't.

Update: I mean the session records in database when collection session metadata is switched on and the session handler is file system.

joomdonation:
I think this PR does not work. See https://www.php.net/manual/en/language.types.string.php#language.types.string.casting

> Resources are always converted to strings with the structure "Resource id #1", where 1 is the resource number assigned to the resource by PHP at runtime. While the exact structure of this string should not be relied on and is subject to change, it will always be unique for a given resource within the lifetime of a script being executed (ie a Web request or CLI process) and won't be reused

So from what I see, you don't see the PHP error because the code successfully convert the data to string, however, it return wrong Session ID as you can guess.

After some Google Search, I think we might have to accept the ugly code https://github.com/joomla/joomla-cms/issues/33740#issuecomment-838539919 and move on.

richard67:
@PhilETaylor We cast stuff (e.g. objects) to strings here and there, but on a quick search yesterday I haven't found any place where we do that for resources. If you find such a place, please report it with a new issue because that would be a bug, as far as I understand PHP docs.

richard67:
Alternative PR see #33819 .

brianteeman:
Can you update the title please. We should use tags to indicate release blocker

PhilETaylor:
Closing in favour of #33819

PhilETaylor:
> We should use tags to indicate release blocker

Not everyone has the permissions to add tags.

PhilETaylor:
Please remove the tag release blocker - this PR is now closed.

Issue #33801, opened 2021-05-11
[4.0] Fields title in user profile inside contact


drmenzelit:
### Steps to reproduce the issue

- Activate user profile
- Create a contact
- Set display user profile
- Create a custom field in users
- Let the field empty in the user's profile
- Create a contact menu item

### Expected result
Contact data are displayed, including user profile. No custom fields, because no value was entered in user profile

### Actual result
Contact data are displayed, including user profile. A `<h3>` title "Fields" is displayed in the user profile part of the contact although the field is empty

![grafik](https://user-images.githubusercontent.com/9153168/117873857-88871e00-b2a0-11eb-8a7e-104a604e1613.png)

The title is generated in com_contact\tmpl\contact\default_user_custom_fields.php line 37
It looks for existent `userFieldGroups `but not if the fields are filled.

### System information (as much as possible)
Joomla ‎4.0.0-beta8-dev

drmenzelit:
Closing as we have a PR #33818

PR #33818, opened 2021-05-12
[4.0] Display Custom Fields for Frontend Contact View Conditionally

YatharthVyas:
Pull Request for Issue #33801

### Summary of Changes
Refactored the foreach block. The loop iteration is executed if the field's value exists and its parent Field Group is to be displayed.

### Testing Instructions

1. Activate user profile
2. Create a contact
3. Set display user profile
4. Create a custom field in users (Multiple fields under multiple field groups are recommended for thorough testing)
5. Let the field empty in the user's profile
6. Create a contact menu item
7. Visit the contact created from Frontend

Before the PR, you will be able to see the field groups for all possible groups, including those left blank.
After the PR, Only those groups will be shown that have a field with a value.

<hr>

image

image

None


ghost:
I have tested this item :white_check_mark: successfully on 8da070d4f8bf455c08869a2aa051da3f91e1bcdb

#### Without pull request

> Screenshot_2021-05-12 Contact

#### With pull request

> Screenshot_2021-05-12 Contact(1)


#### With pull request, no empty field

> Screenshot_2021-05-12 Contact(2)



<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33818">issues.joomla.org/tracker/joomla-cms/33818</a>.</sub>

joomdonation:
I haven't tested but I think you are making more change than needed (and even the current code is having more lines of code than it should)

I think all you need to do is change the current block of code https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_contact/tmpl/contact/default_user_custom_fields.php#L29-L32 with something like:
```php
<?php if ($field->value) : ?>
		<?php $userFieldGroups[$field->group_title][] = $field; ?>
<?php endif; ?>
```
No need for defining an extra array, no need for initialize $userFieldGroups[$field->group_title] with an empty array like in the current code.

Or even better, you can move the $field->value check to the if check at the top https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_contact/tmpl/contact/default_user_custom_fields.php#L26, and then remove these lines of code https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_contact/tmpl/contact/default_user_custom_fields.php#L29-L31 and everything will just work well.

Hope I don't miss anything.

YatharthVyas:
Thanks, @joomdonation for the amazing solution. It worked out perfectly!

@sandramay0905 Thank you for testing. If you have time, could you please re-test.

Edit: Please wait before testing, I will be adding one more commit to remove some extra code

Edit 2: Thank you for your patience, you may test the code now

components/com_contact/tmpl/contact/default_user_custom_fields.php
```diff
@@ -23,7 +23,7 @@
 <?php endif; ?>
 
 <?php foreach ($this->contactUser->jcfields as $field) : ?>
-	<?php if (!in_array('-1', $displayGroups) && (!$field->group_id || !in_array($field->group_id, $displayGroups))) : ?>
+	<?php if (!in_array('-1', $displayGroups) && (!$field->group_id || !in_array($field->group_id, $displayGroups)) || !$field->value) : ?>
```
joomdonation: I think you have change the logic of this if block. The code should be something like below to keep same logic as before:

```php
<?php if (!$field->value || (!in_array('-1', $displayGroups) && (!$field->group_id || !in_array($field->group_id, $displayGroups)))) : ?>
```
But it is still a bit complicated. Maybe you can try to make it simpler if it is possible

joomdonation: You can try to test with:
```php
<?php if (!($field->value && (in_array('-1', $displayGroups) || in_array($field->group_id, $displayGroups)))) : ?>
```
It should work and easier to understand than the current code.

YatharthVyas: The current loop structure was something like:

```python
for loop:
    if (condition):
        continue;
    // block of code
```
so I decided of shifting it to:

```python
for loop:
    if (!condition):
        // block of code

```

Now, the challenge here was to find the negation of the condition and simplify it:

I derived it to be 
```php
<?php if ($field->value && (in_array('-1', $displayGroups) || ($field->group_id && in_array($field->group_id, $displayGroups)))) : ?>
```

And I also verified it using D`Morgan's Law:

```
~(~a or (~b and (~c or ~d)))
a and ~(~b and (~c or ~d))
a and (b or ~(~c or ~d))
a and (b or (c and d))

where

a = $field->value
b = in_array('-1', $displayGroups)
c = $field->group_id
d = in_array($field->group_id, $displayGroups)
```

YatharthVyas: So, the final statement reads as:

If the value of the field exists `AND` ( the default category is in $displayGroups `OR` ( the field group is not default `AND` the field group exists in $display groups)))

components/com_contact/tmpl/contact/default_user_custom_fields.php
```diff
@@ -23,13 +23,9 @@
 <?php endif; ?>
 
 <?php foreach ($this->contactUser->jcfields as $field) : ?>
-	<?php if (!in_array('-1', $displayGroups) && (!$field->group_id || !in_array($field->group_id, $displayGroups))) : ?>
-		<?php continue; ?>
+	<?php if ($field->value && (in_array('-1', $displayGroups) || ($field->group_id && in_array($field->group_id, $displayGroups)))) : ?>
```
joomdonation: I think it could be simplified a bit more:

```php
<?php if ($field->value && (in_array('-1', $displayGroups) || in_array($field->group_id, $displayGroups))) : ?>
```

joomdonation: So the field will be displayed if:

- It has value
- The menu item decide to display all field groups or the field belong to a group which is selected in menu item parameter

YatharthVyas: You are correct.
Tested the same and it works.
Thank you, once more!

joomdonation: Well done! Now we will just need user test. I'm sure @sandramay0905 will test it. If no one else tests, I will test on tomorrow.

joomdonation:
I have tested this item :white_check_mark: successfully on ec931310a7ce8f245c013a99cfb4c1048bbfe89c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33818">issues.joomla.org/tracker/joomla-cms/33818</a>.</sub>

ceford:
I created a Group, then went to create a Field and got a 500 error:

 >    500 Unknown column 'a.only_use_in_subform' in 'field list' 
 >JROOT/administrator/components/com_fields/src/View/Fields/HtmlView.php:90 

This before applying the patch. I just merged the latest origin into my local copy. I don't know what has wrong but I can't test this PR.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33818">issues.joomla.org/tracker/joomla-cms/33818</a>.</sub>

joomdonation:
@ceford Your Joomla is outdated. If it is possible, please download nightly build and re-install it. Or if you don't want to re-install, update to Joomla 4 nightly build, then run this SQL command to your database:

```sql
ALTER TABLE `#__fields` ADD COLUMN `only_use_in_subform` tinyint(1) NOT NULL DEFAULT 0;
```

Of course you need to replace #__fields with correct database table on your site (depends on the table prefix which you are using for your Joomla installation)

ghost:
I have tested this item :white_check_mark: successfully on ec931310a7ce8f245c013a99cfb4c1048bbfe89c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33818">issues.joomla.org/tracker/joomla-cms/33818</a>.</sub>

joomdonation:
Thanks !<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33818">issues.joomla.org/tracker/joomla-cms/33818</a>.</sub>

ceford:
It is not that my installation is out of date. I simply forgot that there is a Database Update option. Anyone, we have two tests so all good. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33818">issues.joomla.org/tracker/joomla-cms/33818</a>.</sub>

richard67:
Thanks!

YatharthVyas:
Thanks all

PR #33817, opened 2021-05-12
[4.0] sessionID & pgsql

alikon:
Pull Request for Issue #33740 . alternative to #33787

### Summary of Changes
Handle resources

### Testing Instructions
see #33740
use postgresql
change session to file
login/logout

### Actual result BEFORE applying this Pull Request
error

### Expected result AFTER applying this Pull Request
logout successfully

PR #33819, opened 2021-05-12
[4.0] Handle session ID being a stream to fix session manager problem on logout with PostgreSQL database

richard67:
Pull Request for Issue #33740 .

Alternative to #33787 and #33817 .

### Summary of Changes

When reading the session id from the session table on a PostgreSQL database using loadColumn(), the result is an array of resources and not of strings.

This PR here adds a check in the destroySessions routine of the session manager if the session ID is of type `resource` and the resource type is a stream, and if that is the case, reads the stream and use the result string for the destroySession call.

### Testing Instructions

See issue #33740 .

See issue #33740 .

No errors. Session destroyed correctly. Now back at the login screen.

None.

### DIFF
--- a/libraries/src/Session/SessionManager.php
+++ b/libraries/src/Session/SessionManager.php
@@ -66,6 +66,11 @@ final class SessionManager
 
 		foreach ($sessionIds as $sessionId)
 		{
+			if (is_resource($sessionId) && get_resource_type($sessionId) === 'stream')
+			{
+				$sessionId = stream_get_contents($sessionId);
+			}
+
 			if (!$this->destroySession($sessionId))
 			{
 				$result = false;


richard67:
Added release blocker label as inherited from the issue.

joomdonation:
I have tested this item :white_check_mark: successfully on fc07b6e547a4b780f91335858260c547a434c803<br /><br />Tested on both MySQLi and PostgreSQL.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33819">issues.joomla.org/tracker/joomla-cms/33819</a>.</sub>

PhilETaylor:
You gained a badge of honour...

Screenshot 2021-05-12 at 11 46 40

PhilETaylor:
You gained a badge of honour... (@rdeutz)

Screenshot 2021-05-12 at 11 46 40

richard67:
@SharkyKZ How should it be done in the right way?

richard67:
@PhilETaylor No need to ping Robert. Maybe we get some info why, or not. Let's wait and see what happens. I am really not the PHP guru, so I'm always open for suggestions/change requests/critics, from whomever it comes.

PhilETaylor:
> I'm always open for suggestions/change requests/critics, from whomever it comes.

But it never comes. Thats the problem. 

Its so disrespectful to simply thumb down PR's - constantly - with zero feedback and zero input. He has been banned and warned about his actions previously, but it seems he has not learned his lesson.

PhilETaylor:
The docblock states that the input to this method should be an array of strings, so I guess doing the conversion where you are goes against that, because they should be strings before this method is called - according to the documented param - but, as in all the docs in joomla methods, they could be wrong and can be changed. 

```
 * @param   string[]  $sessionIds  The session IDs to destroy.
 * ```

richard67:
If @alikon 's PR #33817 doesn't get a thumb down in a while, and Nicola adds the change at the other place which I have mentioned there with a comment. then I am happy to close this one in favour of Nicola's.

richard67:
Closing in favour or #33817 .

Issue #33823, opened 2021-05-12
4.0 Difference in height of visibility button and password input 


deveshprasad:
### Steps to reproduce the issue

Go to the main joomla website page and view on mobile view

### Expected result

![10 05 2021_20 55 28_REC](https://user-images.githubusercontent.com/63739986/117997238-6e147980-b360-11eb-8dee-4676cfd58fa4.png)

### Actual result

![11 05 2021_21 10 38_REC](https://user-images.githubusercontent.com/63739986/117997244-6f45a680-b360-11eb-93c8-7580bc930a6f.png)

### Additional comments

A very unnoticable issue but still an important one .

eopws:
1. Where is the place, you got the problem?
2. What is on screenshot that you provided in "System information" section?
3. If you provided screenshot where everything is fine, that probably means that you solved the issue, so why don't you open a pull request?

rjharishabh:
Please add 4.0 in the Title of the issue and change the System information

deveshprasad:
@eopws I have opened a pull request solving it just want to confirm on which branch should I work?
1)Login form on the joomla main webpage

richard67:
Closing as having a pull request. See #33824 .

PR #32697, opened 2021-03-15
Fixes button margin

srishty-07:
Pull Request for Issue # .

**Summary of Changes**
bootstrap margin class was added to the danger button

**Testing Instructions**
open the cassiopeia page[frontend] on (http://localhost/joomla-cms/index.php?option=com_config&view=modules&id=1&Itemid=101&return=aHR0cDovL2xvY2FsaG9zdC9qb29tbGEtY21zLw%3D%3D)
2)open edit module on the right hand side of the page.

set the toggle device toolbar to mobile view.
4)before applying the patch it would show improper margin between the buttons, but after applying the patch proper margin is there.

**Actual result BEFORE applying this Pull Request**

image

**Expected result AFTER applying this Pull Request**
image

PR #33824, opened 2021-05-12
Difference in height of visibility button and password input #33823

deveshprasad:
Pull Request for Issue #33823 

### Summary of Changes

Difference in the height of input and visibility button
Solved by adding a bootstrap bottom margin to 0.

### Testing Instructions

Go to the main joomla web page and view it on mobile view.

`<button type="button" class="btn btn-primary js-pstats-btn-allow-always m-1"><?php echo Text::_('PLG_SYSTEM_STATS_BTN_SEND_ALWAYS'); ?></button> <button type="button" class="btn btn-primary js-pstats-btn-allow-once m-1"><?php echo Text::_('PLG_SYSTEM_STATS_BTN_SEND_NOW'); ?></button> <button type="button" class="btn btn-primary js-pstats-btn-allow-never m-1" style="margin:0px"><?php echo Text::_('PLG_SYSTEM_STATS_BTN_NEVER_SEND'); ?></button>`

10 05 2021_20 55 28_REC

11 05 2021_21 10 38_REC

No changes in the documentation.

### DIFF
--- a/modules/mod_login/tmpl/default.php
+++ b/modules/mod_login/tmpl/default.php
@@ -52,7 +52,7 @@ Text::script('JHIDEPASSWORD');
 				<div class="input-group">
 					<input id="modlgn-passwd-<?php echo $module->id; ?>" type="password" name="password" autocomplete="current-password" class="form-control" placeholder="<?php echo Text::_('JGLOBAL_PASSWORD'); ?>">
 					<label for="modlgn-passwd-<?php echo $module->id; ?>" class="visually-hidden"><?php echo Text::_('JGLOBAL_PASSWORD'); ?></label>
-					<button type="button" class="btn btn-secondary input-password-toggle">
+					<button type="button" class="btn btn-secondary mb-0 input-password-toggle">
 						<span class="icon-eye icon-fw" aria-hidden="true"></span>
 						<span class="visually-hidden"><?php echo Text::_('JSHOWPASSWORD'); ?></span>
 					</button>


rjharishabh:
I have tested this item :red_circle: unsuccessfully on 48a28941bd8e2998b334a0291873f7a69da954cb<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33824">issues.joomla.org/tracker/joomla-cms/33824</a>.</sub>

rjharishabh:
It's the same as before 
erro

And please change your branch to something other, not 4.0-dev

rjharishabh (changes requested):
Please change bottom-margin to 0
See below comment

brianteeman:
The problem is caused by
```
@media (max-width: 767.98px)
.btn {
    margin-bottom: .25rem;
}
```

rjharishabh (approved):

rjharishabh:
Sorry, it's by mistake approved and I didn't revert it.
So, please don't consider these changes approved

deveshprasad:
I followed the GSOC guidelines to change the branch to 4.0-dev, so which branch should I choose?

richard67:
> I followed the GSOC guidelines to change the branch to 4.0-dev, so which branch should I choose?

@deveshprasad The branch is ok. It would have been better if you had created a new, extra branch for this PR based on the 4.0-dev branch, but as long as you don't make other PR's for other issues, it works like now, using your 4.0-dev branch.

Later when this PR is ok and tested and merged, the best is you delete your 4.0-dev branch and create it again based on the 4.0-dev branch of the CMS. Otherwise, if you don't do that and later create another PR based on your 4.0-dev branch which still contains your changes from this PR here, the other, future PR will be wrong.

But for now that should not be a problem.

rjharishabh:
Here you should change margin-bottom to 0

https://github.com/joomla/joomla-cms/blob/9653158d1eb14190f66f68233fc44763dd67d884/templates/cassiopeia/scss/vendor/bootstrap/_buttons.scss#L52

brianteeman:
> Here you should change margin-bottom to 0

1. If the intent is to remove the margin-bottom at that size then you should just remove the entire class not set it to 0
2. Whenever you remove something you should see why it was there in the first place. Everything has a purpose so you wouldnt want to break that with a change

The PR that introduced that code was #32697

brianteeman:
I have tested this item :red_circle: unsuccessfully on 48a28941bd8e2998b334a0291873f7a69da954cb<br /><br />Breaks all buttons see comments<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33824">issues.joomla.org/tracker/joomla-cms/33824</a>.</sub>

rjharishabh:
Then it can be corrected by add `mb-0` to it's markup. 
That's in password. php file.

brianteeman:
That would set a margin bottom at all screen sizes. Is that what you want to do?

deveshprasad:
So if the bootstrap margin is not the correct way and we should remove the entire class, and should directly set it to 0.
I tried applying bootstrap margin individually ( .btn) so that it won't affect the other buttons. But this is still not the correct way, 
so can we add a new class on this button and which has its margin in the CSS files equal to 0?

I have a similar issue which I solved with help of bootstrap, but if m-1 is not the correct way as it works on all responsive screens
then can I add a new class because removing the current one will remove all those styling related with that class.

Similar issue:

rjharishabh:
Same issue at the change password page in the frontend.
change-pass

If we add `mb-0` to this line then both problem will be solved, It's not affecting other screen sizes

https://github.com/joomla/joomla-cms/blob/85d42d25279aac2cfbe2d69a7568234d04c6a40b/layouts/joomla/form/field/password.php#L156

deveshprasad:
13 05 2021_15 45 42_REC
13 05 2021_20 50 17_REC


Similar issue  on mobile view .
If we want to target one particular screen without affecting other screen size, and by using bootstrap we can use 
mb-md-0 for the previous issue and mt-md-1 for this issue ?

PR #33887, opened 2021-05-14
Upgrade not downgrade appveyor

wilsonge:
test


wilsonge:
@HLeithner I guess we can do this

Issue #33717, opened 2021-05-09
Responsive Issues on different screen sizes


deveshprasad:
### Intro

My name is Devesh Prasad. I want to contribute to the GSOC 21 project of Joomla that is to add a frontend inline editing feature, after successful installation of code files in the localhost to get familiar with the codebase I have shortlisted some frontend responsive issue which can help me later on with the project of inline editing.
So, I would like to work on these issues.

### 1
![screen shot 2021-05-09 at 19 06 03](https://issues.joomla.org/uploads/1/17bc405527c5c2e23942b3befba68de2.png)

### 2
![screen shot 2021-05-09 at 19 06 03](https://issues.joomla.org/uploads/1/3826548fb831c33e8b6acc4db97b8d22.png)!

### 3
![screen shot 2021-05-09 at 19 06 03](https://issues.joomla.org/uploads/1/86927870f1ec43fa7ac34c3e9abe5cd4.png)!

### 4
![screen shot 2021-05-09 at 19 06 03](https://issues.joomla.org/uploads/1/781c99a3da4a11b76cdef80c2e62f18c.png)!

### 5
![screen shot 2021-05-09 at 19 06 03](https://issues.joomla.org/uploads/1/5a6fbb8c84dc45c5cfe4416f9840e9a4.png)

### Expected result

Responsive issues mentioned above to be solved with improved familiarity with the codebase.
By working on the margin and padding and other frontend issues as per the timetable I have mentioned in my GSOC proposal.

### Proposal Timeline Mentioned
Proposal link: https://docs.google.com/document/d/1a_Ix_XVIiXLxm1Teu-G5s_BU0DzWtgb5Q43F2VTkVA4/edit?usp=sharing

### Name: Devesh Prasad
Email: deveshprasad99911@gmail.com

Bond97:
I confirmed responsive issue 2, it happens when you use mobile view of Atum template and the resolution is less than the default mobile view in Firefox or Chromium-based browser. 

@ciar4n did a great job to repaint Admin UI, but he write, that media manager needs much more work to be fully responsive in mobile view.

deveshprasad:
### **To solve issue 1 above**
Which approach should I follow, 
1) By adding bootstrap margins to all the three buttons that is " m-1 " in all the three buttons as adding it on an individual will affect the alignment and gap ratio between all three buttons.
2) By using media queries in the SCSS  file so that when the screen size is a mobile view, the never button should have a top margin.

[...]

PR #33895, opened 2021-05-15
4.0-dev -> Responsive Issue 1 on different screen sizes #33717

deveshprasad:
Pull Request for Issue # .

### Summary of Changes
By adding bootstrap margins to all the three buttons that is " mb-1 " in all the three buttons as adding it on an individual will affect the alignment and gap ratio between all three buttons.

### Testing Instructions

Go to the administrator dashboard
View it on mobile view

13 05 2021_15 43 25_REC

No changes

### DIFF
--- a/administrator/components/com_postinstall/tmpl/messages/default.php
+++ b/administrator/components/com_postinstall/tmpl/messages/default.php
@@ -47,12 +47,12 @@ $adminFormClass = count($this->extension_options) > 1 ? 'form-inline mb-3' : 'vi
 			<div>
 				<?php echo Text::_($item->description_key); ?>
 				<?php if ($item->type !== 'message') : ?>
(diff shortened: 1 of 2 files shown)


richard67:
@deveshprasad In future, please use the "Pull Request for Issue # ." head line in the pull request to refer to the issue solved by that pull request so we know that and can close the issue and not have to spend time with searching.

In this case here it would be: Pull Request for Issue #33717 .

Leave a space after the number and GitHub will automatically make the right link of it.

Thanks in advance.

brianteeman:
Doing it this way means that you have to potentially do this to all buttons now and in the future. Also if I understand correctly it is only an issue at specific sizes whereas this way applies the change to all sizes.

Please see #33899 for my proposed alternative

PR #33901, opened 2021-05-15
[4.0] Bootstrap 5.0.1

dgrammatiko:
Pull Request for Issue # .


PhilETaylor:
```
Changelog
Fixed an issue where dropdowns wouldn’t close after clicking into an <input>
Validated inputs in .input-groups no longer render behind sibling elements
Prevent accent-bg from leaking in nested tables
Modal backdrops no longer throw Uncaught TypeError when initialized through JS
Refactored disposing properties into the base component
Extracted static DATA_KEY and EVENT_KEY to the base component
Merged transitionend listener callbacks into one method
Popovers and tooltips have a streamlined config property
Toasts no longer automatically hide on focus or hover
No longer redefining $list-group-color in the list group loop
Our docs and examples also received a few updates:

Fixed Sidebars example not rendering correctly in Chrome
Fixed RTLCSS stringMap configuration snippet
Updated offcanvas navbar example to prevent console error
Fixed miscellaneous typos, grammatical errors, and links in the Migration guide

```

HLeithner:
is it expected that npm deletes 9,926 rows from package-lock.json?

PhilETaylor:
> is it expected that npm deletes 9,926 rows from package-lock.json?

 All those that are removed are under the "packages" top level key. 

we still have 50 vulnerabilities (43 moderate, 7 high) mostly due to the reliance on insecure out of postcss

PhilETaylor:
I have tested this item :white_check_mark: successfully on 8d43c3c5265afd75eaf99027c1a8567793bc0b2d<br /><br />Checked change log, no showstoppers. Applied PR. Rebuilt all assets, played with Joomla admin for a while. All seems to function. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33901">issues.joomla.org/tracker/joomla-cms/33901</a>.</sub>

richard67:
> is it expected that npm deletes 9,926 rows from package-lock.json?

@HLeithner That comes from the lock file format changing back from version 2 to version 1 (`"lockfileVersion": 1`in line 4). Version 2 had lots of additional info like funding information. There still hasn't been made a decision if we change from 1 to 2. The reason why the 4.0-dev branch currently is 2 is because that might have happened with a past PR.

@wilsonge Correct me if that was all wrong what I wrote.

HLeithner:
hmm at least it should be consistent and not changed with every pull request

richard67:
Agree. But what can I do?

brianteeman:
If memory serves me right then this is something the production department were going to make a decision on. Either way it doesnt prevent this being merged

HLeithner:
@dgrammatiko can you please switch to type 2 format? so we don't jump between the formats?

dgrammatiko:
I have no idea how this can be done. Let me google it

brianteeman:
Its pointless

HLeithner:
> Its pointless

no idea why you think it's pointless but 🤷

brianteeman:
Because if you look at the history you will see it switching between lockfile 1 and lockfile 2 repeatedly with no errors or issues. Its fine to make a ruling on which version to require but until that date ...

HLeithner:
thanks @dgrammatiko 
right brian doesn't make sense when I'm the only maintainer that thinks about such things

dgrammatiko:
So it turns out that version 2 is not what it's supposed to be here (it implies Node >= 14 && npm >= 7 which are not the project's minis)

A fix is to require everybody to use node>= 14 and npm >= 7
But, a nicer solution is to setup a simple github action that will automatically update the -lock.json (basically only the action will ever update the locks)

Solutions like:
https://stackoverflow.com/questions/64813775/is-there-any-way-to-fix-package-lock-json-lockfileversion-so-npm-uses-a-specific

will make things a lot more confusing for devs not familiar with nvm etc

Issue #33861, opened 2021-05-14
[4.0] draggable rows - sorting order not correct


Ruud68:
### Steps to reproduce the issue
In Users > Fields create 3 custom fields: test 1, test 2, test 3 (e.g. type text)
when looking in the fields table you can see that all of them have set the ordering field to 0

now click on the sort column header to start sorting
drag test 1 to position 2

the displayed sort order is now:
test 2
test 1
test 3

now press f5 to refresh the screen
the sort order now is:
test 2
test 3
test 1

### Expected result
the values in the ordering field should be set to the correct sort order:
test 1 > 2
test 2 > 1
test 3 > 3

### Actual result
the values in the ordering field are not set to the correct sort order
test 1 > 3
test 2 > 1
test 3 > 2

### System information (as much as possible)
tested with todays (2021-05-14) nightly
php 8.0.5
mysql 8.0.25

brianteeman:
confirmed - something weird going on there

himanshu007-creator:
confirmed issue

joeforjoomla:
Confirmed, applies to core and also third-party extensions

joeforjoomla:
The problem is that the dragged element is always POSTed as the last one having the max order value
order[]: 16
cid[]: 30

joomdonation:
@Fedik or @dgrammatiko Could you please take a look? It seems the dragged row is also counted, so on this line https://github.com/joomla/joomla-cms/blob/4.0-dev/build/media_source/system/js/draggable.es6.js#L54 , `rows` contains more than one element than it should

Change this line https://github.com/joomla/joomla-cms/blob/4.0-dev/build/media_source/system/js/draggable.es6.js#L64 to 
```javascript
for (i = 0, l = rows.length - 1; l > i; i += 1) {
```
would fix the issue. But I guess it is not the right fix.

dgrammatiko:
@joomdonation looking good, probably I should go back to elementary school and relearn counting 😂

joomdonation:
@dgrammatiko No, your original code looks good. The issue we are having here is somehow, related to how `dragula` works. If we have table with 3 rows for example, that querySelectorAll methods 4 rows. I can only guess that dragula create one extra element which is used for drag and drop and that's the reason querySelectorAll return one extra row.

In short, you do not have to go back to  elementary school :D.

dgrammatiko:
> The issue we are having here is somehow, related to how dragula works. 

[...]

PR #33909, opened 2021-05-16
[4.0] Fix drag and drop re-ordering

joomdonation:
Pull Request for Issue #33861.

### Summary of Changes
This PR fixes the issue https://github.com/joomla/joomla-cms/issues/33861

### Testing Instructions

1. Confirm the issue described at https://github.com/joomla/joomla-cms/issues/33861
2. Download update package for this PR at https://ci.joomla.org/artifacts/joomla/joomla-cms/4.0-dev/33909/downloads/43761/Joomla_4.0.0-beta8-dev+pr.33909-Development-Update_Package.zip , go to System -> Update -> Joomla, upload this package ad run the update
3. Then check and confirm that the issue sorted.

### DIFF
--- a/build/media_source/system/js/draggable.es6.js
+++ b/build/media_source/system/js/draggable.es6.js
@@ -61,8 +61,8 @@ document.addEventListener('DOMContentLoaded', () => {
       }
 
       /**
-       * Get the order array. We have to exclude the extra DOM element that draggula is creating for the dragging effect,
-       * that's the reason l = rows.length - 1 instead of l = rows.length as usual
+       * Get the order array. We have to exclude the extra DOM element that draggula is creating
+       * for the dragging effect, that's the reason l = rows.length - 1 instead of l = rows.length as usual
        */
       for (i = 0, l = rows.length - 1; l > i; i += 1) {
         rows[i].value = i + 1;
--- a/build/media_source/system/js/draggable.es6.js
+++ b/build/media_source/system/js/draggable.es6.js
@@ -63,6 +63,7 @@ document.addEventListener('DOMContentLoaded', () => {
       // Get the order array
       for (i = 0, l = rows.length; l > i; i += 1) {
         // Skip a mirror element
+        // eslint-disable-next-line no-continue
         if (rows[i].closest('.gu-mirror')) continue;
 
         rows[i].value = i + 1;
--- a/build/media_source/system/js/draggable.es6.js
+++ b/build/media_source/system/js/draggable.es6.js
@@ -60,12 +60,11 @@ document.addEventListener('DOMContentLoaded', () => {
         inputRows.reverse();
       }
 
-      /**
-       * Get the order array. We have to exclude the extra DOM element that draggula is creating
-       * for the dragging effect, that's the reason l = rows.length - 1 instead of
-       * l = rows.length as usual
-       */
-      for (i = 0, l = rows.length - 1; l > i; i += 1) {
+      // Get the order array
+      for (i = 0, l = rows.length; l > i; i += 1) {
+        // Skip a mirror element
+        if (rows[i].closest('.gu-mirror')) continue;
+
         rows[i].value = i + 1;
         result.push(`order[]=${encodeURIComponent(rows[i].value)}`);
         result.push(`cid[]=${encodeURIComponent(inputRows[i].value)}`);
--- a/build/media_source/system/js/draggable.es6.js
+++ b/build/media_source/system/js/draggable.es6.js
@@ -62,7 +62,8 @@ document.addEventListener('DOMContentLoaded', () => {
 
       /**
        * Get the order array. We have to exclude the extra DOM element that draggula is creating
-       * for the dragging effect, that's the reason l = rows.length - 1 instead of l = rows.length as usual
+       * for the dragging effect, that's the reason l = rows.length - 1 instead of
+       * l = rows.length as usual
        */
       for (i = 0, l = rows.length - 1; l > i; i += 1) {
         rows[i].value = i + 1;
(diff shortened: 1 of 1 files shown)


ghost:
I have tested this item :white_check_mark: successfully on 34b45a035389b37dab4ba36b2f781f0b5117a790<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33909">issues.joomla.org/tracker/joomla-cms/33909</a>.</sub>

Fedik:
you have found bug correctly, but the fix not really ;)

You have to skip "mirror" row, not last row.
The mirrored row is removed in "dragend" event, it would be best to call the request here, however this event also triggered when user cancel sorting, so it a bit tricky.

Just make the loop like this:
```javascript
for (i = 0, l = rows.length; l > i; i += 1) {
  // Skip a mirror element
  if (rows[i].closest('.gu-mirror')) continue;

  rows[i].value = i + 1;
  result.push(`order[]=${encodeURIComponent(rows[i].value)}`);
  result.push(`cid[]=${encodeURIComponent(inputRows[i].value)}`);
}
```

joomdonation:
@Fedik No surprise with my beginner JS level :D. I asked @dgrammatiko before and he confirmed the fix looks OK. I tested and it seems it works, too. Maybe you can help with a proper fix? I will be happy to close this PR

Fedik:
just copy paste the code from my last comment ;)

joomdonation:
Thanks @Fedik. Tested and it worked, so I committed the change.

Ruud68:
Hi Tuan, Although this fixes the issue initially reported here: #33861 testing this a bit further exposes another bug in the logic:
create 6 custom fields, set the view list length to 5 (so only the first 5 are shown).
now when sorting on the first page, the sorting looks okay until you F5 then field number 6 turns out to be not the last but the first...
Should I create a new issue for this?

Fedik:
to fix eslint, add
```javascript
// Skip a mirror element
// eslint-disable-next-line no-continue
if (rows[i].closest('.gu-mirror')) continue;
```

_this `no-continue` rule is pointless, I would disable it globally_

joomdonation:
@Ruud68 I'm unsure If my js skill is enough to address that. So just leave this PR here for a while, If I cannot get it sorted, I will close and ask @Fedik or @dgrammatiko for help

One thing I don't understand is why ordering of fields are set to 0 when it is created? If we create 3 fields, I would expect ordering of the fields set to 1, 2, 3..... and so on. For case, I think the last field has ordering set to 0, so after the sorting, it still has ordering 0, while other fields on the first page has ordering set to right value and it causes the error

So maybe something related to field ordering when it is first saved?

Fedik:
> So maybe something related to field ordering when it is first saved?

This something in PHP side on save, I suspect.

joomdonation:
> This something in PHP side on save, I suspect.

I think so. But we might also have problem with client side, too. If from second page, we start ordering again with value 0 (or 1) for first row, we would get the issue (need to look at more carefully, just a quick analyst)

Ruud68:
makes me wonder what happens when you have a list of e.g. 100K items you sort, would it rewrite all the records ordering fields?
Just a thought as I have a customer who has that amount of sortable items...

joomdonation:
As I guessed, on new page (second, third page....), we re-start with ordering of items again from 1, so we have a big bug here :( .

Ruud68:
And to further thicken the plot: in J3 ordering is using negative numbers, so e.g. when you set field one below field 2 then field one gets ordering of 1 and field 2 gets ordering of -1.
It should also be tested how ordering of fields will handle these when upgrading to 4.0 and then reordering n 4. Will it correctly handle the negative numbers etc. as the logic of ordering has apparently changed.

joomdonation:
OK. I think this PR solves a valid issue. For re-ordering items logic, I open a separate issue https://github.com/joomla/joomla-cms/issues/33917 . @Ruud68 If you have any other concerns, please add it to the issue so that we can look at it later.

Fedik:
> we re-start with ordering of items again from 1

hmhm the script just send the final positions to the backend, and backend should recalculate the final order for given ID, depend from element position in request.
In theory.
Not sure how it work now, in backend.

joomdonation:
> Not sure how it work now, in backend

From my quick look, the posted ordering will be saved for each record (if the value for that change), then the ordering if related items will be re-calculated (I haven't looked at that re-calculate logic yet)

So from page 2 and page 3 for example, the item will have ordering again, set to 1, 2, 3......, so it is completely wrong. We need to look at it and correct it in a separate PR

Fedik:
I have tested this item :white_check_mark: successfully on a34e44b2f6dd5e0cbd2df3dbbb03ad9cfcd9d331<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33909">issues.joomla.org/tracker/joomla-cms/33909</a>.</sub>

joomdonation:
@sandramay0905 Could you please re-test this PR ? Just makes sure it works like in your previous test. There is another bug related to re-ordering items but we will address it in a different PR

ghost:
I have tested this item :white_check_mark: successfully on a34e44b2f6dd5e0cbd2df3dbbb03ad9cfcd9d331<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33909">issues.joomla.org/tracker/joomla-cms/33909</a>.</sub>

joomdonation:
Thanks for testing !<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33909">issues.joomla.org/tracker/joomla-cms/33909</a>.</sub>

Quy:
Thank you!

PR #33910, opened 2021-05-16
[4.0] Simplify populateState method for ActionLogs model

joomdonation:
Pull Request for Issue # .

### Summary of Changes
Our ListModel is smart enough to populate model states automatically, so we do not have to write code to populate value for individual states anymore. From what I see, manual populating model states like this is only needed for Joomla 3 because we have to support Hathor, a template which does not use search tools for filtering data.

This PR is made to see if it could be accepted. If it is accepted, similar clean up code be made for some other model classes in our codebase

### Testing Instructions
1. Apply patch
2. Access to Users -> User Actions Log, try to filter action logs search by searching, filter by component, date range... and make sure it is still working as expected.

### DIFF
--- a/administrator/components/com_actionlogs/src/Model/ActionlogsModel.php
+++ b/administrator/components/com_actionlogs/src/Model/ActionlogsModel.php
@@ -73,23 +73,6 @@ class ActionlogsModel extends ListModel
 	 */
 	protected function populateState($ordering = 'a.id', $direction = 'desc')
 	{
-		$app = Factory::getApplication();
-
-		$search = $app->getUserStateFromRequest($this->context . 'filter.search', 'filter_search', '', 'string');
-		$this->setState('filter.search', $search);
-
-		$user = $app->getUserStateFromRequest($this->context . 'filter.user', 'filter_user', '', 'string');
-		$this->setState('filter.user', $user);
-
-		$extension = $app->getUserStateFromRequest($this->context . 'filter.extension', 'filter_extension', '', 'string');
-		$this->setState('filter.extension', $extension);
-
-		$ip_address = $app->getUserStateFromRequest($this->context . 'filter.ip_address', 'filter_ip_address', '', 'string');
-		$this->setState('filter.ip_address', $ip_address);
-
-		$dateRange = $app->getUserStateFromRequest($this->context . 'filter.dateRange', 'filter_dateRange', '', 'string');
-		$this->setState('filter.dateRange', $dateRange);
-
 		parent::populateState($ordering, $direction);
 	}
 


PhilETaylor:
the only thing that is left is a call to the parent class - does that mean you can actually remove this whole method and allow inheritance to happen, meaning the parent method will get called instead auto-magically?

joomdonation:
@PhilETaylor No, we cannot remove it. We need to leave it here so that default ordering is `a.id desc` (see default values for the parameters of the method)

PhilETaylor:
Are those peoples different in the parent method then? Sorry I cannot check quite now as I am mobile, and driving

joomdonation:
Yes, Phil. For parent method, the default value for these parameters are null

`protected function populateState($ordering = null, $direction = null)`

PhilETaylor:
Awesome - thanks for explaining - I'll mark a test when I'm home later if still needed

joomdonation:
Thanks Phil. Maybe we should ask maintainers to see if they want this clean up first. @wilsonge @HLeithner Do you think we should go with this code clean up?

wilsonge:
Yeah if this works we should definitely clean it up this way. Just check what state the API filters on when your rolling this out (we can move it to the new search tools style state names - just need to make sure it's not breaking API filters)

joomdonation:
@wilsonge  The filters form within Users Action Logs still work in the same way as before. The only thing was changed is that if we remove these code, someone could not initialize value for a model state using this code (could be used before redirecting to com_actionlogs component)

```php
Factory::getApplication()->setUserState('com_actionlogs.actionlogsfilter.search', 'initialize search value');
```

They will have to use new code such as
```php
Factory::getApplication()->setUserState('com_actionlogs.actionlogs.filter', ['search' => 'initialize search value']);
```

We do not use any of these code in our code base. So will this change be OK?

joomdonation:
I don't understand your idea/request below, could you please explain more what I need to check to make sure it won't break our API?

>Just check what state the API filters on when your rolling this out (we can move it to the new search tools style state names - just need to make sure it's not breaking API filters)

wilsonge:
I think this specific PR is fine. But some of the API sets these older style filters (e.g. https://github.com/joomla/joomla-cms/blob/4.0-dev/api/components/com_content/src/Controller/ArticlesController.php#L65). We'll need to clean them up - I guess thinking about it we should be able to automatically allow using these filters from the XML files rather than doing it manually like we are there...

joomdonation:
@wilsonge I don't understand how our API works yet, so I could not comment on the linked code.

Back to this PR, all it does is remove the un-necessary code. There is no reason to keep the code which is not needed and we do not use it (we do not need these lines of code from the time search tools was added to Joomla, values for model states are now passed in **filter array**, not as individual scalar value like `$_POST['filter_search']` anymore). 

For more information, we do not keep these strange code anymore in our other components like **com_banners** , see https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_banners/src/Model/BannersModel.php#L292 

Would be great if we could clean up these code before RC release.

PR #33751, opened 2021-05-10
[4] New moduleposition below-top, option to switch off logo

chmst:
### Summary of Changes
Not all users want it have a logo in the header of their site. Cassipoeia has no option to switch off the display of a logo or description
This PR adds a param to the template where the user can switch off the logo.
It adds a new position "below-top" where users can add an own module. 
Using both - position "below-top" and logo is up to the user. 

The error page can be adapted if this is accepted.

### Testing Instructions
Apply the patch and play with the template style cassiopeia. Use the module position brand 

### Actual result BEFORE applying this Pull Request
There is always an area for the logo above the navigation. 

### Expected result AFTER applying this Pull Request

The user can swith off the logo
grafik

### Documentation Changes Required
yes

PR #31520, opened 2020-11-28
[4.0] Cassopeia template and blog sample data fixes and improvements

richard67:
Pull Request for Issues #30812 , #30680 , #30515 , #30442 , #30426 , #29504 , #29311 , #29138 , #25239 , #24129 and possibly more which haven't been identified yet.

#### Introduction

This pull request (PR) is the result of the work of the [Frontend Template Working Group (JEDT - FT)](https://volunteers.joomla.org/teams/frontend-template-working-group) in the recent few months for improving Cassiopeia, the default site template of Joomla 4.0.

Beside fixing template-specific issues and bugs, our task was to make Cassiopeia be a good showcase for new features in Joomla 4.0, together with improved sample data and without introducing any additional third-party dependencies.

Development took place in the [Cassiopeia repository on GitHub](https://github.com/joomla/cassiopeia), so there you can find why things have been done like you can see them here.

The implementation was mainly done by **_following contributors_** (in alphabetical order):
@chmst , @drmenzelit , @hans2103 and @richard67

But of course we had some help, therefore our **_special thanks_** goes to (in alphabetical order)
@brianteeman , @ciar4n , @helvecio , @infograf768 and @N6REJ
for advice, support, testing and contribution.

#### Improved Template Structure - HTML and SCSS

The structure of the template has got countless small changes and a new, clear structure for SCSS files. The header has got a new structure. Module positions are the same as before.

#### New Error and Offline Site

The error and offline pages have been rewritten so they match the structure of the template's index.php and use the new features like colour themes or fonts schemes.

#### Accessibility Improvements

Not only for the template itself (colours, font sizes), but also in layouts of the Joomla core many improvements for accessibility have been made, especially in the search module and search result page. 

- Colour contrast ratio is AA over the whole template.
- Font Size and Line Height are user friendly.
- For icons and images, aria labels and alt-texts are carefully used .
- All elements are included in landmarks .
- The elements and the dropdown menu (metismenu) are accessible also for keyboard users.
- Font size and family for frontend editing have been improved.

Many a11y improvements which we have found during this project were made in layouts of the Joomla core with other pull requests, but there are still some issues left in the Joomla frontend (e.g. structure of headings) where further improvement is needed.

#### Accessible Dropdown Menu

Currently Joomla 4 and the Cassiopeia template don't provide an accessible dropdown menu which can be used in horizontal or vertical orientation. Bootstrap classes like e.g `nav-pills` don't work like they did in Joomla 3.

[...]

PR #33930, opened 2021-05-17
[4.0] Fix initial value of the Cassiopeia template style parameter "brand" for new installation and update and use it in error.php, too

richard67:
Pull Request for https://github.com/joomla/joomla-cms/pull/33751#issuecomment-841432541 .

### Summary of Changes

1. Fix the outdated `params` values for the Cassiopeia template style in the base.sql files for new installation so they match to the default values in the XML file, including the changes from #33751 , #31520 and possibly others.

2. Use default value 1 in call to `$this->params->get` for this parameter so it defaults to "Yes" if the values is not set in database e.g. after update from 4.0 Beta 7 or 3.10, which is the same default value as defined in the XML for that parameter.

3. Use the parameter also for the error page as mentioned in the description of PR #33751 :

> The error page can be adapted if this is accepted.

4. Remove the `height: 100vh;` in the `.error_site` block of the `templates/cassiopeia/scss/blocks/_modifiers.scss` to get rid of the otherwise shown blue bar at the top or the error page when no header should be shown, see comment https://github.com/joomla/joomla-cms/pull/33930#issuecomment-842582987 below.

### Testing Instructions

Step 1: Use current 4.0-dev or the nightly build from May 18 or later to make a new installation.

Step 2: After the installation has finished, do not do anything else in backend.

Step 3: In frontend, check if the blue header section is shown.

Result: No blue header section.

Step 4: Apply the changes from this PR.

Step 5: Check again if the blue header section is shown in frontend.

Result: Blue header section with logo "Cassiopeia" and no site description is shown, like it was before PR #33751 was merged, because the default value in the `$this->params->get` call.

Step 6: Remove configuration.php, delete all tables in database and make again a new installation, still with the patch of this PR applied.

Step 7: After the installation has finished, do not do anything else in backend.

Step 8: In frontend, check if the blue header section is shown.

Result: Blue header section with logo "Cassiopeia" and no site description is shown, like it was before PR #33751 was merged, because the parameter has the right value in database.

Step 9: In backend, edit the Cassiopeia template style settings and check if the "Brand" parameter in advanced settings is switched on.

Result: The parameter is switched on.

Step 10: Check the error page by entering an unkown URL in frontend.

Result: Blue header section with logo "Cassiopeia" and no site description is shown.

Step 11: In backend, switch off the "Brand" parameter and save the changes.

Step 12: Check the front page and the error page on frontend if the blue header section is shown.

Result: The blue header is not shown anymore in frontend on normal and error page.

No blue header section with logo in frontend after a new installation or update from 3.10 or 4.0 Beta 7 and earlier without having saved template style settings.

See https://github.com/joomla/joomla-cms/pull/33751#issuecomment-841432541 .

The error page doesn't respect the "Brand" parameter.

Site shows the header with logo "Cassiopeia" on frontend after a new installation or update without having saved template style settings.

The error page respects the "Brand" parameter.

None.


richard67:
Drone error is not related to this PR.

hans2103:
> Result: The blue header is not shown anymore in frontend on normal and error page.

this is not correct. Due to templates/cassiopeia/scss/blocks/_modifiers.scss line 6 .error_site { height: 100vh; } there is a blue line, depending on the height of the viewport
https://github.com/joomla/joomla-cms/blob/5371b921a6486a23fc956ed3e362e6818df27bea/templates/cassiopeia/scss/blocks/_modifiers.scss#L6

by removal of the height, the blue bar is gone on the error page too. 
Is it within the scope of this PR? Don't know.. but is easy fix

hans2103 (approved):
see my comment... with or without the height I approve this PR. 
https://github.com/joomla/joomla-cms/pull/33930#issuecomment-842582987

richard67:
> 
> 
> > Result: The blue header is not shown anymore in frontend on normal and error page.
> 
> this is not correct. Due to templates/cassiopeia/scss/blocks/_modifiers.scss line 6 .error_site { height: 100vh; } there is a blue line, depending on the height of the viewport
> 
> https://github.com/joomla/joomla-cms/blob/5371b921a6486a23fc956ed3e362e6818df27bea/templates/cassiopeia/scss/blocks/_modifiers.scss#L6
> 
> by removal of the height, the blue bar is gone on the error page too.
> Is it within the scope of this PR? Don't know.. but is easy fix

Fixed, and PR description adapted.

hans2103:
I have tested this item :white_check_mark: successfully on 3a494e2fccad7729c0e129911e351314d5cc6eeb<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33930">issues.joomla.org/tracker/joomla-cms/33930</a>.</sub>

templates/cassiopeia/error.php
```diff
@@ -103,16 +103,18 @@
...
 	<header class="header container-header full-width">
-		<div class="grid-child">
-			<div class="navbar-brand">
-				<a class="brand-logo" href="<?php echo $this->baseurl; ?>/">
-					<?php echo $logo; ?>
-				</a>
-				<?php if ($this->params->get('siteDescription')) : ?>
-					<div class="site-description"><?php echo htmlspecialchars($this->params->get('siteDescription')); ?></div>
-				<?php endif; ?>
+		<?php if ($this->params->get('brand', 1)) : ?>
```
Fedik, line 106: I always wonder why do we have to set any initial "params" data in to database, 
as we can just have `$params->get('foobar', 'default')` , and leave a database empty, that can save much time. 

Just thoughts ;)

richard67: Well, I could have change the wrong data to an empty json instead of correct data, that's right. But I decided to stay consistent with what we have for the other template style.

wilsonge:
Thanks!

wilsonge:
Merging this with only one good test because I want to have the sql updates in place for an RC in the next week or so

richard67:
Thanks!

Issue #33917, opened 2021-05-16
[4.0] Review items re-ordering logic


joomdonation:
### Steps to reproduce the issue
While reviewing PR https://github.com/joomla/joomla-cms/pull/33909, we found potential issue with items re-ordering using drag and drop in Joomla 4. Main issue are:

- On page 2, page 3...., ordering of items are re-started from 1, so it causes wrong ordering saved if we change ordering of items from page 2, page 3.....
- Also, from @Ruud68 
> And to further thicken the plot: in J3 ordering is using negative numbers, so e.g. when you set field one below field 2 then field one gets ordering of 1 and field 2 gets ordering of -1.
It should also be tested how ordering of fields will handle these when upgrading to 4.0 and then reordering n 4. Will it correctly handle the negative numbers etc. as the logic of ordering has apparently changed.

Fedik:
hm, it seems PHP part for saveOrder is not changed,
on quick look it the same in J3 and J4 for `AdminController::saveOrderAjax()`, `AdminController::saveorder()`, `AdminModel::saveorder()`

joomdonation:
Yes, I think only client part was changed, so should not be hard to fix :). As mentioned, page 2, page 3 restart item ordering from 1 and it causes the issue. I guess instead of re-start with 1, it should be started with min ordering of the items on that page and it should be fine.

Ruud68:
What makes it difficult is that the first sort is different from the second sort. This because the sort order is set to 0 when an item is created.  This could be avoided if the initial sort order is set to the primary key on item creation.

richard67:
Set release blocker label as discussed with other maintainers.

Fedik:
okay, so it is JS, in J3 the order is recalculated before submit in
https://github.com/joomla/joomla-cms/blob/4172b68dac980a85c6070511622322006e98fb5b/media/jui/js/sortablelist.js#L215-L255

in J4 it always start fro 1

joomdonation:
That's right. And keep in might that in J3, we do not change ordering of all items like we are doing in J4. Change ordering of all items could cause performance issue during saving process (note that we also have row in same group, and ordering value of row from different groups should be be changed)

Fedik:
It changes the order for the active page, if I right understood. 
But need to check how it doing the calculation, it not that easy to get what there is written :smile:

joomdonation:
[...]

Issue #33936, opened 2021-05-18
[4.0] Possible Bug with ordering value in ArticlesModel query


joomdonation:
### Steps to reproduce the issue
I haven't had time to look at it yet but in ArticlesModel, we have two ordering fields in the query without using Alias, and it could cause wrong ordering value returned, thus causes error with Drag and Drop ordering in Articles management

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_content/src/Model/ArticlesModel.php#L255
https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_content/src/Model/ArticlesModel.php#L276

If someone could figure out why we need that ordering data (especially from #__content_frontpage table), that would be great.

dgrammatiko:
> If someone could figure out why we need that ordering data (especially from #__content_frontpage table)

Isn't this the messy `featured items`?

joomdonation:
> Isn't this the messy featured items?

I haven't had a chance to read the code myself carefully yet. However, because of this bug, the wrong ordering value for each article will be used on articles view, mean re-ordering articles in article management won't work properly. (I found this issue why trying to test the PR https://github.com/joomla/joomla-cms/pull/33937 )

richard67:
> > If someone could figure out why we need that ordering data (especially from #__content_frontpage table)
> 
> Isn't this the messy `featured items`?

@dgrammatiko Yes, `#__content_frontpage` is the table for that messy thing .. the table of which I wish it never had been created. The columns could have been added to the content table. This would maybe waste a few kilobytes space on a big table where the columns would be used only for a few % of all data, but it would have saved us from so much trouble. Well, all I can say is .. it wasn't me 😄

joomdonation:
I don't understand why we need an extra table `#__content_frontpage` like this. But I guess it is too late for us to do the removal, so I made PR https://github.com/joomla/joomla-cms/pull/34025 to fix this issue for now.

PR #33937, opened 2021-05-18
[4.0] Fix drag and drop ordering logic

joomdonation:
Pull Request for Issue #33917

### Summary of Changes
My attempt to fix drag and drop re-ordering issue https://github.com/joomla/joomla-cms/issues/33917 . Ping everyone to review code, especial @dgrammatiko @Fedik @Ruud68 who understand the issue and discussed about it in https://github.com/joomla/joomla-cms/issues/33917

### Testing Instructions
1. Download update package for this PR at https://ci.joomla.org/artifacts/joomla/joomla-cms/4.0-dev/33937/downloads/43843/Joomla_4.0.0-beta8-dev+pr.33937-Development-Update_Package.zip , go to System ->Update -> Joomla, upload and install that updated package

2. Go to Users -> Fields, create 7 or 8 user custom fields . (You just need to create 1 field, then use Save As Copy to create other fields to save time)

3. Click on Sort icon to sort fields by ordering, try to drag and drop to change of the field. Then re-fresh the page and make sure the ordering is sorted.
4. Choose 5 from Number records dropdown (so that the system will only display 5 records per page). Navigate to page 2 (that's the reason I asked you to create 7 or 8 fields). Then try to change ordering of items on that second page, refresh the page and make sure the ordering are kept (items ordering are saved properly)

- ordering data of all items changed for each re-order action. Could cause performance issue
- ordering data of items always re-started from 1, which cause wrong ordering value, especially when we re-order items from page 2, page 3....

- only ordering data of the necessary items are changed. 
- ordering data of items in next page (page 2, page 3...) are calculated base on it's existing ordering data and the item which is re-ordered in the action

### DIFF
--- a/build/media_source/system/js/draggable.es6.js
+++ b/build/media_source/system/js/draggable.es6.js
@@ -2,186 +2,182 @@
  * @copyright  (C) 2019 Open Source Matters, Inc. <https://www.joomla.org>
  * @license    GNU General Public License version 2 or later; see LICENSE.txt
  */
+// The container where the draggable will be enabled
+let url;
+let direction;
+let isNested;
+let dragElementIndex;
+let dropElementIndex;
+let container = document.querySelector('.js-draggable');
+const orderRows = container.querySelectorAll('[name="order[]"]');
 
-document.addEventListener('DOMContentLoaded', () => {
-  'use strict';
+if (container) {
+  /** The script expects a form with a class js-form
+   *  A table with the tbody with a class js-draggable
+   *                         with a data-url with the ajax request end point and
+   *                         with a data-direction for asc/desc
+   */
+  url = container.dataset.url;
+  direction = container.dataset.direction;
+  isNested = container.dataset.nested;
+} else if (Joomla.getOptions('draggable-list')) {
+  const options = Joomla.getOptions('draggable-list');
 
-  // The container where the draggable will be enabled
-  let url;
-  let direction;
-  let isNested;
-  let dragElementIndex;
-  let dropElementIndex;
-  let container = document.querySelector('.js-draggable');
-  const orderRows = container.querySelectorAll('[name="order[]"]');
-
-  if (container) {
-    /** The script expects a form with a class js-form
-     *  A table with the tbody with a class js-draggable
-     *                         with a data-url with the ajax request end point and
-     *                         with a data-direction for asc/desc
-     */
-    url = container.getAttribute('data-url');
-    direction = container.getAttribute('data-direction');
-    isNested = container.getAttribute('data-nested');
-  } else if (Joomla.getOptions('draggable-list')) {
-    const options = Joomla.getOptions('draggable-list');
-
-    container = document.querySelector(options.id);
-    /**
-     * This is here to make the transition to new forms easier.
-     */
-    if (!container.classList.contains('js-draggable')) {
-      container.classList.add('js-draggable');
-    }
-
-    ({ url } = options);
-    ({ direction } = options);
-    isNested = options.nested;
+  container = document.querySelector(options.id);
+  /**
+   * This is here to make the transition to new forms easier.
+   */
+  if (!container.classList.contains('js-draggable')) {
+    container.classList.add('js-draggable');
   }
 
-  if (container) {
-    // Add data order attribute for initial ordering
-    for (let i = 0, l = orderRows.length; l > i; i += 1) {
-      orderRows[i].setAttribute('data-order', i + 1);
-    }
+  ({ url } = options);
+  ({ direction } = options);
+  isNested = options.nested;
+}
 
-    // IOS 10 BUG
-    document.addEventListener('touchstart', () => {}, false);
+if (container) {
+  // Add data order attribute for initial ordering
+  for (let i = 0, l = orderRows.length; l > i; i += 1) {
+    orderRows[i].dataset.order = i + 1;
+  }
 
-    const getOrderData = (rows, inputRows, dragIndex, dropIndex) => {
-      let i;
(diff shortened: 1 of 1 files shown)


Fedik:
Looks good on quick look. Redo of j3 version?

Noticed you do not skip "mirror" element of dragula.js. You think it will not cause the same issue as was fixed before? in #33909
Btw, if do final calculation at "dragend" event then the "mirror" element will be already removed.

joomdonation:
> Looks good on quick look. Redo of j3 version?

Should work in the same way with J3 except that in J3, we submitted data of all records. This version, a bit improvement because we only submit data of the record needs to be updated only

> Noticed you do not skip "mirror" element of dragula.js. You think it will not cause the same issue as was fixed before

Could not say that I understand how dragula works. But base on my debug, that mirror element is always the latest row when we do `querySelectorAll`, so it is OK me (tried some console.log to check the submit data, tested and it works well in my test)

dgrammatiko:
> Noticed you do not skip "mirror" element of dragula.js. You think it will not cause the same issue as was fixed before?

Probably not as the rows are container children and the mirror element is before the body end...

`const rows = [].slice.call(container.querySelectorAll(orderSelector));`

dgrammatiko:
I have not tested this item.<br /><br />Categories are fine, articles not but there's an issue for this view: https://github.com/joomla/joomla-cms/issues/33936<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33937">issues.joomla.org/tracker/joomla-cms/33937</a>.</sub>

dgrammatiko:
I have tested this item :white_check_mark: successfully on 6b4d90efc002d82525da5584d9a75bdbd739ad6e<br /><br />Categories are fine, articles not but there's an issue for this view: https://github.com/joomla/joomla-cms/issues/33936<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33937">issues.joomla.org/tracker/joomla-cms/33937</a>.</sub>

dgrammatiko:
I have tested this item :white_check_mark: successfully on e61426b53c9290b343bb1203bbc9c91fadab2e6c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33937">issues.joomla.org/tracker/joomla-cms/33937</a>.</sub>

PhilETaylor:
Maybe one of these days the JSST @joomla/security team might actually acknowledge and reply to my email about a security issue related to this... 12 days and not even an acknowledgement or initial response.

dgrammatiko:
> but "generally related to reordering", sorry.

Is it J4 specific? (or in plain English did I messed up big time?)

PhilETaylor:
nothing to do with your work :) its in PHP :) - I've pinged them again today and @richard67 directly to see if I can get responses... its like a ghost town trying to contact the JSST

dgrammatiko:
>  its in PHP :)

I think I spotted it, it's a 1 line fix if I got it right?

PhilETaylor:
yes probably - but probably effects Joomla 3 also

dgrammatiko:
> yes probably - but probably effects Joomla 3 also

Confirmed

joomdonation:
@PhilETaylor @dgrammatiko If I understand that issue right, maybe it is not really a security because we at least check edit state permission during saving ordering process already

And maybe we should do the fix for J4 only. If we do that fix in J3, it might make third party extensions broken for that task (because there could be extensions do not pass that data in save ajax ordering request in URL)

Just blind guess :D.

PhilETaylor:
This is why we have a whole team dedicated, that should be addressing this themselves. Instead all of my emails have been ignored. 

The current state of that team is undeterminable.￼￼

If my emails about this minor issue are ignored, what is to say that major issues reported by others are also being ignored by the projects JSST team￼

The ACL check, is no substitution for correct request ￼Security.￼

PhilETaylor:
And yes this should be fixed in 3, as well! Not fixing security issues because they might affect 3PD is simply unacceptable￼￼, And this project has always made the right decision previously to implement request Security correctly￼

PhilETaylor:
And because of the nature of the security fix required, it has to be handled by the JSST team, who can then merge it by release lead￼ just before releasing the next version of Joomla! 3

PhilETaylor:
Finally got a reply from the team, they have confirmed the issue, and they are working on a fix for joomla 3.￼

Sorry to hijack this PR comment thread￼

joomdonation:
Good to know that. And I was also wrong when assumed that the fix is backward incompatible. Actually, J3 is working in different way with J4 (the whole form data is serialized and submitted, so the fix is safe for J3, too)

TLWebdesign:
I have tested this item :white_check_mark: successfully on e61426b53c9290b343bb1203bbc9c91fadab2e6c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33937">issues.joomla.org/tracker/joomla-cms/33937</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33937">issues.joomla.org/tracker/joomla-cms/33937</a>.</sub>

wilsonge:
Thanks! Good work!

PR #33946, opened 2021-05-18
[4][Security] Do not allow command line to be called from the web

PhilETaylor:
### Summary of Changes

StackOverflow emailed me this https://joomla.stackexchange.com/questions/29123/cli-directory-is-publicly-accessible-is-it-supposed-to-be-is-that-safe/29181

They were correct, that Joomla 4 did not have any test for CLI use when the scripts were called from the web.  

The code proposed is back ported from the [Joomla-framework `Joomla\Console\Application` constructor](https://github.com/joomla-framework/console/blob/537038ea7eb67a807c1b502801441206a16c56f7/src/Application.php#L190
) and so should be "good enough"

### Testing Instructions

Go to http://example.com/cli/joomla.php in a web browser. 

With debug on/error reporting = maximum 
Screenshot 2021-05-18 at 18 17 17

With debug off/error_reporting none

Screenshot 2021-05-18 at 18 18 00

White screen of death - which is a good thing.

// @joomla/security @wilsonge

### DIFF
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -107,7 +107,7 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 		?OutputInterface $output = null
 	)
 	{
-		// Close the application if we are not executed from the command line.
+		// Close the application if it is not executed from the command line.
 		if (!\defined('STDOUT') || !\defined('STDIN') || !isset($_SERVER['argv']))
 		{
 			$this->close();
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -107,6 +107,12 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 		?OutputInterface $output = null
 	)
 	{
+		// Close the application if we are not executed from the command line.
+		if (!\defined('STDOUT') || !\defined('STDIN') || !isset($_SERVER['argv']))
+		{
+			$this->close();
+		}
+
 		// Set up a Input object for Controllers etc to use
 		$this->input    = new \Joomla\CMS\Input\Cli;
 		$this->language = $language;


HLeithner:
is there a reason why you not simple do `if (\php_sapi_name() === 'cli') {` ?

PhilETaylor:
Just like I stated:

> The code proposed is back ported from the Joomla-framework Joomla\Console\Application constructor and so should be "good enough"

PhilETaylor:
It should also be noted that its the same code that is currently used in Joomla 3

https://github.com/joomla/joomla-cms/blob/241925d58aca0fe2b8d0d29f792ac7146e056b59/libraries/src/Application/CliApplication.php#L57

richard67:
I have tested this item :white_check_mark: successfully on 9926dbc78120f8f34df278f2cea5727ee1be0777<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33946">issues.joomla.org/tracker/joomla-cms/33946</a>.</sub>

HLeithner:
> Just like I stated:
> 
> > The code proposed is back ported from the [Joomla-framework Joomla\Console\Application constructor](https://github.com/joomla-framework/console/blob/537038ea7eb67a807c1b502801441206a16c56f7/src/Application.php#L190) and so should be "good enough"

^^ ok sorry missed this part, looks strange anyway but ok

PhilETaylor:
Im not saying its perfect, but at least its consistent :) 

There are a few uses of `php_sapi_name` in Joomla 3 but the only use of `php_sapi_name` in Joomla 4 appears to be in vendor libs.

alikon:
I have tested this item :white_check_mark: successfully on 9926dbc78120f8f34df278f2cea5727ee1be0777<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33946">issues.joomla.org/tracker/joomla-cms/33946</a>.</sub>

alikon:
perfectible  but better than before

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33946">issues.joomla.org/tracker/joomla-cms/33946</a>.</sub>

libraries/src/Application/ConsoleApplication.php
```diff
@@ -107,6 +107,12 @@ public function __construct(
 		?OutputInterface $output = null
 	)
 	{
+		// Close the application if we are not executed from the command line.
```
brianteeman:
```suggestion
		// Close the application if it is not executed from the command line.
```

PhilETaylor: same comment is used elsewhere, I just copy and pasted it :)

richard67:
Previous tests and RTC are still valid since last commit was just a change in a code comment.

wilsonge:
Somewhat unconvinced this belongs here vs in the entry file itself for the cli. But better than nothing for sure. easy to move around in the future. Thanks!

PhilETaylor:
its a direct copy of the same code in a constructor in the CliApplication.php 

https://github.com/joomla/joomla-cms/blob/ad06aa48f235dd09a8ca7181cf48f51ad0f6a291/libraries/src/Application/CliApplication.php#L111

wilsonge:
Sure but in that place every CLI app had a different entry point. In J4 the one true way (xD) is that we'll have a single entry point like site/admin is - so should be a bit simpler to manage and less confusing for 3rd party devs who don't need to look at the file. But anyhow it's fine for now

PR #33954, opened 2021-05-18
[4] Method call is provided 5 parameters, but the method signature uses 4 parameters

PhilETaylor:
needs careful code review

### DIFF
--- a/administrator/components/com_users/src/Model/UserModel.php
+++ b/administrator/components/com_users/src/Model/UserModel.php
@@ -1034,7 +1034,7 @@ class UserModel extends AdminModel
 		if (strpos($config, '{') === false)
 		{
 			$openssl         = new Aes($key, 256);
-			$mcrypt          = new Aes($key, 256, 'cbc', null, 'mcrypt');
+			$mcrypt          = new Aes($key, 256, 'cbc', 'mcrypt');
 
 			$decryptedConfig = $mcrypt->decryptString($config);
 


joomdonation:
I have tested this item :white_check_mark: successfully on 4020292bbc6c72ca1cdb8e9792a0dc72c9bff2cb<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33954">issues.joomla.org/tracker/joomla-cms/33954</a>.</sub>

joomdonation:
It is OK. But I have a question which I don't know the answer yet. Isn't Mcrypt not available in from PHP 7.2+ anymore? Maybe we will need to review this case more carefully.

PhilETaylor:
mcrypt can still be used after 7.2 - but you just have to add it manually to PHP if you want it.

This PR is not about mcrypt, its about instantiating the class with the correct number of params.

I believe its a typo Im removing.

joomdonation:
> This PR is not about mcrypt, its about instantiating the class with the correct number of params. I believe its a typo Im removing.

That's right. But the fact is that we still use Mcrypt here, so there could be fatal error in some case. We should review it and solve it (but of course in a different PR)

PhilETaylor:
replaced this PR with https://github.com/joomla/joomla-cms/pull/34147

PR #33960, opened 2021-05-18
[4] remove fake type hints

PhilETaylor:
you can't just append a type and hope Joomla converts to that type haha - fake security, the method call doesn't have this additional param 

code review

### DIFF
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -63,8 +63,8 @@ class ManageController extends ApiController
 			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0');
 		}
 
-		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')));
-		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')));
+		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status', null, 'int')));
+		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type', null, 'string')));
 
 		return parent::displayList();
 	}
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -48,7 +48,7 @@ class ManageController extends ApiController
 	 */
 	public function displayList()
 	{
-		$requestBool = $this->input->get('core', $this->input->get->get('core', null, 'string'));
+		$requestBool = $this->input->get('core', $this->input->get->get('core', null));
 
 		if (!is_null($requestBool) && $requestBool !== 'true' && $requestBool !== 'false')
 		{
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -48,7 +48,7 @@ class ManageController extends ApiController
 	 */
 	public function displayList()
 	{
-		$requestBool = $this->input->get('core', $this->input->get->get('core', null));
+		$requestBool = $this->input->get('core', $this->input->get->get('core', null, 'boolean'));
 
 		if (!is_null($requestBool) && $requestBool !== 'true' && $requestBool !== 'false')
 		{
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -48,7 +48,7 @@ class ManageController extends ApiController
 	 */
 	public function displayList()
 	{
-		$requestBool = $this->input->get('core', $this->input->get->get('core'));
+		$requestBool = $this->input->get('core', $this->input->get->get('core', null, 'string'));
 
 		if (!is_null($requestBool) && $requestBool !== 'true' && $requestBool !== 'false')
 		{
--- a/api/components/com_installer/src/Controller/ManageController.php
+++ b/api/components/com_installer/src/Controller/ManageController.php
@@ -60,11 +60,11 @@ class ManageController extends ApiController
 
 		if (!is_null($requestBool))
 		{
-			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
+			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0');
 		}
 
-		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
-		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'STRING');
+		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')));
+		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')));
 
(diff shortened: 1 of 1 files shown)


richard67:
@PhilETaylor Not sure if you are right here. See https://github.com/joomla-framework/input/blob/2.0-dev/src/Input.php#L166 .

PhilETaylor:
This is set, you linked to get

richard67:
Oh ... I need sleep. For set you are right.

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -60,11 +60,11 @@ public function displayList()
 
 		if (!is_null($requestBool))
 		{
-			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
+			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0');
 		}
 
-		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
-		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'STRING');
+		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')));
```
joomdonation: Instead of remove this, shouldn't we use correct filter for $this->input-get ?

From quick look, it seems filter.status and filter.type expect to receive string, see https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_installer/src/Model/ManageModel.php#L83-L84

PhilETaylor: This PR, like the others, six only to remove that which is blatantly wrong and should not be there and is currently not used. To that end this PR is quickly testable And mergeable￼￼

joomdonation: Do it in the right way, please. If you do not do it right now, no one will see this issue and correct it in the future.

PhilETaylor: > Do it in the right way, please

This was only merged a few days ago, it would have been better for it to have been done right then :)

joomdonation: @PhilETaylor Busy now, so could not review your last changes yet. Will do that later today. And I guess issue like this happens because other developers do not have good IDE like yours :)

PhilETaylor: I don't understand how or why anyone would ever try to write code without a decent code editor. Would a surgeon use a spade to perform heart surgery?

joomdonation: @PhilETaylor Not all our contributors have good IDE like PhpStorm, Zend Studio..., I can only assume that. They can write code using Visual Studio Code, Vim, Notepad++...., so issue like this happens and it is understandable.

PhilETaylor: This is why we have a PR review process isn't it ;-) At least 2 tests and a maintainer checking code before its committed... ... never mind.

Its been proved that the project doesnt care about code quality... Im only one voice screaming in the desert... all this should be fully automated.

joomdonation: We care :). That's the reason I did a real review to your PRs, not just give it success test :)

wilsonge: This ones on me for poor code review. Sorry. It’s the main thing I do before merging.

wilsonge (approved):

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -48,7 +48,7 @@ class ManageController extends ApiController
 	 */
 	public function displayList()
 	{
-		$requestBool = $this->input->get('core', $this->input->get->get('core'));
+		$requestBool = $this->input->get('core', $this->input->get->get('core', null, 'boolean'));
```
joomdonation, line 51: I would say that we are still having problem with the current code. The filer, if applied, should be applied for both `$this->input->get` and `$this->input->get->get` . I mean we should use the same filter. Since you compared with string in later code `$requestBool === 'true'`, I think you should use string filter here.

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -60,11 +60,11 @@ public function displayList()
 
 		if (!is_null($requestBool))
 		{
-			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
+			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0');
 		}
 
-		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
-		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'STRING');
+		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status', null, 'int')));
```
joomdonation, line 66: From what I see from ManageModel, `filter.status` should be `string`, not `int`. I think we can change this line to:

```php
$this->modelState->set('filter.status', $this->input->getString('status', $this->input->get->getString('status', '')));
```

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -60,11 +60,11 @@ public function displayList()
...
+		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type', null, 'string')));
```
joomdonation, line 67: I guess we will can change to:

```php
$this->modelState->set('filter.type', $this->input->getString('type', $this->input->get->getString('type', '')));
```

joomdonation (commented):
@wilsonge I'm not familiar with API request, so please make a review for this PR. Something is still not right here.

PhilETaylor:
No idea - but it is clear what was there before was clearly wrong and needs addressing. If this PR is incorrect, it doesnt change the fact that the current code is wrong :)

joomdonation:
Tried to look at it again but unsure what to do :(. I don't know why we need to attempt to get data from `$this->input` and if it does not exist, try to get it from `$this->input->get` ?

Isn't $this->input contains data from both POST and GET already? Or there is something special here :( ? In ArticlesController for example, we do not need to use `$this->input->get` at all. See https://github.com/joomla/joomla-cms/blob/4.0-dev/api/components/com_content/src/Controller/ArticlesController.php#L48-L79 .

wilsonge:
Honestly right now I can’t remember why this would be either. Pretty sure we just want to be checking GET. Will look again tomorrow with a fresh pair of eyes

joomdonation:
@wilsonge Now I understand more about webservice, I think just getting data from GET would be right. However, for shorter code, using just `$this->input->get` instead of `$this->input->get->get` could work, too. The code here could be simplified.

wilsonge:
> However, for shorter code, using just $this->input->get instead of $this->input->get->get could work, too

It can work but would allow query overriding which would be wrong. For the API we generally should be fairly strict about the data sources.

joomdonation:
This PR should be updated to use same method with https://github.com/joomla/joomla-cms/pull/34288 . We do not want to individual model states like this anymore.

PhilETaylor:
Closing as no possibility of being merged.

PR #33981, opened 2021-05-19
[4] s/setMvcFactory/setMVCFactory

PhilETaylor:
Change case of method. Doesn't change the way its called, but is "more correct' that the 33+ instances that call this method, call it with the correct case

Note that `getMVCFactory` has MVC in all caps, and so rather than change33+ cases to be `setMvcFactory` it seems the right code change is to change as this PR is.


wilsonge:
Thanks!

wilsonge:
Plus I agree with the fix - also consistent with the namespace declaration

PR #33988, opened 2021-05-19
[4] Method call is provided 1 parameters, but the method signature uses 0 parameters

PhilETaylor:
code review

### DIFF
--- a/plugins/system/highlight/highlight.php
+++ b/plugins/system/highlight/highlight.php
@@ -102,7 +102,7 @@ class PlgSystemHighlight extends CMSPlugin
 
 		// Adjust the component buffer.
 		$doc = $this->app->getDocument();
-		$buf = $doc->getBuffer('component');
+		$buf = $doc->getBuffer();
 		$buf = '<div class="js-highlight">' . $buf . '</div>';
 		$doc->setBuffer($buf, 'component');
 	}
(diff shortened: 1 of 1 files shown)


plugins/system/highlight/highlight.php
```diff
@@ -102,7 +102,7 @@ public function onAfterDispatch()
 
 		// Adjust the component buffer.
 		$doc = $this->app->getDocument();
-		$buf = $doc->getBuffer('component');
+		$buf = $doc->getBuffer();
```
wilsonge: OK So I *might* be wrong. But I think because this only runs on HTML pages (see line 63) this actually does have an affect - see https://github.com/joomla/joomla-cms/blob/staging/libraries/src/Document/HtmlDocument.php#L455

PhilETaylor: then either app or getDocument is returning the wrong type hint and phpStorm is complaining, because clicking getBuffer takes me to a method with no params

wilsonge: I think it takes you to \Joomla\CMS\Document\Document which is the parent class that getDocument will return. But HtmlDocument doesn't have a consistent method signature with it. You can just add a typehint for HtmlDocument *after* we've done the check on 63 we have a non-html document

PhilETaylor: type hint added. and you were right

wilsonge:
Thanks!

PR #33990, opened 2021-05-19
[4] correct property type

PhilETaylor:
code review

### DIFF
--- a/plugins/sampledata/blog/blog.php
+++ b/plugins/sampledata/blog/blog.php
@@ -58,7 +58,7 @@ class PlgSampledataBlog extends CMSPlugin
 	/**
 	 * Holds the menuitem model
 	 *
-	 * @var    MenusModelItem
+	 * @var    \Joomla\Component\Menus\Administrator\Model\ItemModel
 	 *
 	 * @since  3.8.0
 	 */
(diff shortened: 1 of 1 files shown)


plugins/sampledata/blog/blog.php
```diff
@@ -58,7 +58,7 @@ class PlgSampledataBlog extends CMSPlugin
 	/**
 	 * Holds the menuitem model
 	 *
-	 * @var    MenusModelItem
+	 * @var    \Joomla\Component\Menus\Administrator\Model\ItemModel
```
wilsonge: Can we import this please as it's also used to initialise the class below in the actual code

PhilETaylor: lol you told me off for importing previously and wanted them all long hand :)

wilsonge: If it's just a code comment with no concrete uses in the file I prefer inline. If it's used as a concrete class in the code I prefer an import.

Is that clear or just confusing more? (Also FYI this is just my personal preference and isn't an official position taken by Joomla etc etc)

PhilETaylor: imported

wilsonge:
Thankyou and sorry for being a pain :)

PR #33992, opened 2021-05-19
[4] add missing class import for method param type

PhilETaylor:
Code review


wilsonge:
I still hate whoever in PHP land decided we should import global things like Exceptions but 🤷‍♂️ thankyou!

PhilETaylor:
well it was other import, or add a `\` before Exception param type, and that would involve scrolling down so I took the easy option :)

PR #33994, opened 2021-05-19
[4] add missing class import for method param type and more cleanup

PhilETaylor:
code review

### DIFF
--- a/plugins/sampledata/blog/blog.php
+++ b/plugins/sampledata/blog/blog.php
@@ -19,7 +19,6 @@ use Joomla\CMS\Language\Text;
 use Joomla\CMS\Plugin\CMSPlugin;
 use Joomla\CMS\Plugin\PluginHelper;
 use Joomla\CMS\Session\Session;
-use Joomla\Component\Menus\Administrator\Model\ItemModel;
 use Joomla\Database\ParameterType;
 
 /**
@@ -59,7 +58,7 @@ class PlgSampledataBlog extends CMSPlugin
 	/**
 	 * Holds the menuitem model
 	 *
-	 * @var    ItemModel
+	 * @var    \Joomla\Component\Menus\Administrator\Model\ItemModel
 	 *
 	 * @since  3.8.0
 	 */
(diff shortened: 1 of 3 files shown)


plugins/sampledata/blog/blog.php
```diff
@@ -883,7 +884,7 @@ public function onAjaxSampledataApplyStep2()
 		$articleIds = $this->app->getUserState('sampledata.blog.articles');
 
 		// Get MenuItemModel.
-		$this->menuItemModel = new \Joomla\Component\Menus\Administrator\Model\ItemModel;
+		$this->menuItemModel = new ItemModel;
```
wilsonge: Totally upto you if you wanna call it out of scope. But we should boot this the same way we're doing in the other two plugins through bootComponent rather than directly

PhilETaylor: add an issue, im actually in bed with a laptop propped up on my pillow and falling asleep :)

plugins/sampledata/multilang/multilang.php
```diff
@@ -61,7 +61,7 @@ class PlgSampledataMultilang extends CMSPlugin
 	/**
 	 * Holds the menuitem model
 	 *
-	 * @var     MenusModelItem
+	 * @var     \Joomla\Component\Menus\Administrator\Model\ItemModel
```
wilsonge: This property is actually unused in this file. Can be deleted

joomdonation:
@PhilETaylor Could you please merge this with latest 4.0-dev so that we can review and get this merged? The change on the file plugins/sampledata/blog/blog.php  was implement on a different PR and not needed anymore.

PhilETaylor:
I have rebased this on current 4.0-dev

PhilETaylor:
looks like I screwed it up

PhilETaylor:
ok maybe now?

joomdonation:
> ok maybe now?

Maybe not. I'm unsure if the changes on plugins/sampledata/blog/blog.php are actual the changes you want (it was not there on your original PR and the change in line 305 I know it's wrong, the other two I'm unsure), so maybe you would want to revert it.

As I can see, there are more problems on plugins/sampledata/multilang/multilang.php. Do you want to look at it and and get it sorted in this PR? (wrong return type declaration, $lang = Factory::getLanguage()->getTag(); , then call $lang->load could cause fatal error....). If not, I will make a PR for that later.

PhilETaylor:
ok I'll close this and start from scratch later today

PR #33996, opened 2021-05-19
s/$stageTable/$workflowTable

PhilETaylor:
stageTable is not yet defined at this line, is the workflowTable that might have errored


wilsonge:
Thanks!

PR #33997, opened 2021-05-19
[4] update url to the current one

PhilETaylor:
code review


wilsonge:
Not sure it's the book store link :) But probably better than an invalid one

PR #33998, opened 2021-05-19
[4] s/http/https

PhilETaylor:
Inline with all the other links to the GPL html

`<license>https://www.gnu.org/licenses/gpl-2.0.html GNU/GPL</license>`


wilsonge:
Thanks!

PR #30707, opened 2020-09-21
[4.0] rename Joomla core classNames fas fa- back to icon- (and still use Font Awesome)

hans2103:
Pull Request for Issue # .

### Summary of Changes

All looked well when @N6REJ implemented Font Awesome in Joomla Administrator. Trouble began when the magic words "backward compatibility" where mentioned. During Joomla 4 both Font Awesome and icons from 3th party extensions with the Joomla 3 prefix "icon-" should be loaded. 

As Font Awesome documentation states in the first line of _[Basic Use](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use)_: 
> You can place Font Awesome icons just about anywhere using a style prefix and the icon’s name. 

A prefix and an icon name. 

By default this prefix is set to `fa`.
There is a variable in the Joomla Administrator template Atum where you can set the prefix. 
https://github.com/joomla/joomla-cms/blob/eb9f540453f76898cf5b0e86f7408179e6550885/administrator/templates/atum/scss/_variables.scss#L128
This prefix is set to `fa` and can be changed to whatever. 

To fix all 3th party backwards compatibility shizzle this PR replaces all instances of `fas fa-` by `icon-` and still uses FontAwesome. 

### Testing Instructions

- Browse through Joomla Administrator and notice the presence of icons. 
- Install Akeeba and / or JCE and see their icons too. 
- Apply this PR
- Run `npm run build:css`
- Repeat the first step and browse through Joomla Administrator and notice the presence of icons.  
  - even the icons for Akeeba and / or JCE are present. 

### Actual result BEFORE applying this Pull Request

Schermafbeelding 2020-09-21 om 08 38 45

### Expected result AFTER applying this Pull Request

Schermafbeelding 2020-09-21 om 08 25 17

### Documentation Changes Required

I do think so... :-) But I don't know where. Not familiar with the Joomla Font Awesome documentation. Sorry

PR #34046, opened 2021-05-20
[4.0] Change fa-8x to icon-8x

Quy:
### Summary of Changes
Change the last instances of `fa-`.

Change `fa-8x` to `icon-8x`.
Apply coding style.

### Testing Instructions
Install the prebuilt package.

or

Click `Post Installation Messages` icon.
Click `Hide all messages` button.
Apply PR.
Run `npm run build:css`

Bell icon should be the same size before/after PR.

### DIFF
--- a/administrator/components/com_postinstall/tmpl/messages/default.php
+++ b/administrator/components/com_postinstall/tmpl/messages/default.php
@@ -27,7 +27,7 @@ $adminFormClass = count($this->extension_options) > 1 ? 'form-inline mb-3' : 'vi
 
 <?php if (empty($this->items)) : ?>
 	<div class="py-5 text-center">
-		<span class="fa-8x icon-bell mb-4" aria-hidden="true"></span>
+		<span class="icon-8x icon-bell mb-4" aria-hidden="true"></span>
 		<h1 class="display-5 fw-bold"><?php echo Text::_('COM_POSTINSTALL_LBL_NOMESSAGES_TITLE'); ?></h1>
 		<div>
 			<p class="lead mb-4">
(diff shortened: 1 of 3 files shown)


brianteeman:
1. repeating my objection to this pointless change
2. its only a partial job - what if in 2 days someone submits a pr with fa-7x or any other valid and working fa class that has not been duplicated as per this pr
3. https://github.com/joomla/joomla-cms/blob/a07bc7796fc460881ebdffed5bf85d1cf3f0d218/layouts/joomla/content/emptystate.php#L39

Quy:
Then they can use `fa-7x`, but at least in the core, we are being consistent.

C-Lodder:
@Quy You want to start moving away from `icon-xx`. It was kept purely for 3rd party compatability reasons. There is absolutely no reason for core to use `icon-` over `fas-`

richard67:
@Quy SCSS - CS: https://ci.joomla.org/joomla/joomla-cms/44115/1/20

Quy:
PR #30707 changed `fas-` to `icon-`. Is it too late to revert before RC?

brianteeman:
That other pr was reverting poorly implemented code despite multiple people pointing that out in advance ;(

icon- was created to be a bc compatibility layer with iconmoon. it was never intended to be a conversion between fontawesome syntax and iconmoon syntax which is what this pr (and others) are doing

Quy:
Let's do it right out of the gate. 

@wilsonge Please confirm to revert #30707.

brianteeman:
reverting that PR will not achieve that

Quy:
Then I am lost what the solution should be.

brianteeman:
The solution is not to do anything

Quy:
So it is ok to have mixed `icon-` and `fa-` in the backend?

brianteeman:
yes - I see no reason why not. Especially for the utility classes

Quy:
Closing then. Thanks!

PR #34060, opened 2021-05-20
[4] Delete whole method from class as it contains calls to methods that dont exist.

PhilETaylor:
Code review. 

Not a typo. 

This method cannot be used - as `_buildQuery` method doesnt even exist. 

I do not believe this method is ever called, and after considerable testing I cannot locate any code or gui interface that would call it.

Happy to be proved wrong, but even if something did call it, they would get a PHP error because `_buildQuery` method doesnt exist as a string in the whole of a Joomla 4 site.

### DIFF
--- a/components/com_content/src/Model/ArchiveModel.php
+++ b/components/com_content/src/Model/ArchiveModel.php
@@ -130,34 +130,6 @@ class ArchiveModel extends ArticlesModel
 	}
 
 	/**
-	 * Method to get the archived article list
-	 *
-	 * @access public
-	 * @return array
-	 */
-	public function getData()
-	{
-		$app = Factory::getApplication();
-
-		// Lets load the content if it doesn't already exist
-		if (empty($this->_data))
-		{
-			// Get the page/component configuration
-			$params = $app->getParams();
-
-			// Get the pagination request variables
-			$limit      = $app->input->get('limit', $params->get('display_num', 20), 'uint');
-			$limitstart = $app->input->get('limitstart', 0, 'uint');
-
-			$query = $this->_buildQuery();
-
-			$this->_data = $this->_getList($query, $limitstart, $limit);
-		}
-
-		return $this->_data;
-	}
-
-	/**
 	 * Model override to add alternating value for $odd
 	 *
 	 * @param   string   $query       The query.


PhilETaylor:
AppVeyour failure unrelated to PR content

```
git clone -q https://github.com/joomla/joomla-cms.git C:\projects\joomla-cms
7fatal: unable to access 'https://github.com/joomla/joomla-cms.git/': Could not resolve host: github.com
8Command exited with code 128
```

chmst:
I have tested this item :white_check_mark: successfully on 127e4a7c558287d1f3e43aa63f43db15cb8e6210<br /><br />Code review<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34060">issues.joomla.org/tracker/joomla-cms/34060</a>.</sub>

joomdonation:
How do you think about remove `_getList` method from same class, too ? There is no reason we have to override that method like we are doing at the moment. I would vote to take this change to remove the method before 4.0 release.

PhilETaylor:
If they are generically all the same and there is a parent method that does the same thing then they should be removed.

I would say though that that should be a different PR. This PR is focused on a method that clearly cannot ever be used as the `_buildQuery` method doesnt exist anywhere in code, so would cause a PHP Fatal error if this code was ever used and so its very obvious that this method has not been called in its current state for many years and can be removed

joomdonation:
I have tested this item :white_check_mark: successfully on 127e4a7c558287d1f3e43aa63f43db15cb8e6210<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34060">issues.joomla.org/tracker/joomla-cms/34060</a>.</sub>

joomdonation:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34060">issues.joomla.org/tracker/joomla-cms/34060</a>.</sub>

PhilETaylor:
>  from same class, too 

Ah I missed that you mean "this class", the one I changed. Yes it looks odd having a odd setting, that can be done in the view if it was actually needed.

joomdonation:
@PhilETaylor Maybe a different PR. Sometime, we have strange code in our codebase. For now, this PR is RTC.

PhilETaylor:
> @PhilETaylor Maybe a different PR. Sometime, we have strange code in our codebase. For now, this PR is RTC.

This is https://github.com/joomla/joomla-cms/pull/34118

richard67:
Thanks!

PR #34062, opened 2021-05-20
[4] Correctly return from display method, correct the doc for that too - manually.

PhilETaylor:
### Summary of Changes

Not a typo. 70 changed files with 128 additions and 124 deletions. 

Not a simple search and replace of everything. 

`parent::display()` in most extensions HtmlView.php files calls `\Joomla\CMS\MVC\View::display()` as `parent::display()`  which finally echo's a $result. 

An echo is not a return, so the second `parent::display()` returns `void` to the first, which then doesnt return anything (correctly returns void already) back to the HTMLView and so its plainly wrong to again then return the non-existing void back up the call stack. 

Therefore `return parent::display($tpl)` is simply wrong, proved by removing the return in this PR. 

This PR then updates the docblocks correctly. 

THERE ARE A COUPLE OF GOTCHAS, this is not a simple search and replace. 

### Testing Instructions

Code review

Apply PR - Joomla still works. 

### note...

This is different to the use of `return parent::display();` in Controllers which is correct, as that is returning an instance of `BaseController` to provide a fluid interface for chaining methods. Whereas this is not the case in the views.

### DIFF
--- a/administrator/components/com_categories/src/View/Categories/HtmlView.php
+++ b/administrator/components/com_categories/src/View/Categories/HtmlView.php
@@ -84,7 +84,7 @@ class HtmlView extends BaseHtmlView
 	 *
 	 * @param   string|null  $tpl  The name of the template file to parse; automatically searches through the template paths.
 	 *
-	 * @return  mixed  A string if successful, otherwise an Error object.
+	 * @return  void
 	 */
 	public function display($tpl = null)
 	{
@@ -139,7 +139,7 @@ class HtmlView extends BaseHtmlView
 			}
 		}
 
-		return parent::display($tpl);
+		parent::display($tpl);
 	}
 
 	/**
--- a/administrator/components/com_categories/src/View/Categories/HtmlView.php
+++ b/administrator/components/com_categories/src/View/Categories/HtmlView.php
@@ -84,6 +84,8 @@ class HtmlView extends BaseHtmlView
 	 *
 	 * @param   string|null  $tpl  The name of the template file to parse; automatically searches through the template paths.
 	 *
+	 * @throws  GenericDataException
+	 *
 	 * @return  void
 	 */
 	public function display($tpl = null)
--- a/administrator/components/com_cache/src/View/Cache/HtmlView.php
+++ b/administrator/components/com_cache/src/View/Cache/HtmlView.php
@@ -83,11 +83,11 @@ class HtmlView extends BaseHtmlView
 	 *
 	 * @param   string  $tpl  The name of the template file to parse; automatically searches through the template paths.
 	 *
-	 * @return  mixed  A string if successful, otherwise an Error object.
+	 * @return  void
 	 *
 	 * @since   1.6
 	 *
-	 * @throws  Exception
+	 * @throws  GenericDataException
 	 */
 	public function display($tpl = null): void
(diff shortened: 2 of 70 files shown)


PhilETaylor:
rebased after @brianteeman preparedocument PR to fix my conflicts.

administrator/components/com_categories/src/View/Categories/HtmlView.php
```diff
@@ -84,7 +84,7 @@ class HtmlView extends BaseHtmlView
 	 *
 	 * @param   string|null  $tpl  The name of the template file to parse; automatically searches through the template paths.
 	 *
-	 * @return  mixed  A string if successful, otherwise an Error object.
+	 * @return  void
```
joomdonation, line 89: Could you please add @throws  GenericDataException here (same as you did for administrator/components/com_cache/src/View/Cache/HtmlView.php) ? Hope you can do the same for other classes you work on in this PR.

PhilETaylor: phpStorm wants 

```
@throws  GenericDataException|Exception
or 
@throws  Exception
```

This is because accessing the db has an uncaught Exception that can happen. 

GenericDataException is an instance of RuntimeException which is an instance of Exception - this is why phpStorm doesnt moan if you only have Exception in the throws list. 

Then change I made to throws, not being one of the options above, is already complaining. So basically we need to choose one of the above two, and stick to it. One means I revert 2 files, and one means 70 file changes to be made :)

joomdonation: Ah, right. I don't know which one is good, too. Guess it should be `GenericDataException|Exception` as you pointed out, but I don't have a strong idea about it. @wilsonge Decision, please?

PhilETaylor: I think `GenericDataException|Exception` is "more correct" and should be used, just means 70 more file changes on a sunday :)

joomdonation: I agree. But please wait for decision from @wilsonge. You would kill me if you changed these 70 lines, and then he asks you to revert :D.

joomdonation: @wilsonge @HLeithner Could we have guidance about what to do with throws tag here? Would love to get this PR completes so that IDE stops complaining everyday :D.

HLeithner: The correct way would be to list all Exceptions that could occur but I think that's a bigger task in the cms then change 70 files? I mean not many functions have a throws attribute with proper exceptions that could be thrown right?

PhilETaylor: The docblocks are a complete mess throughout the project. The sheer number of mistakes is staggering.

PhilETaylor: This PR was not just about the documentation, it was to remove the fake returning from the parent

wilsonge:
This will do as a starting point. The exception handling is much more tricky and a bigger piece of work - think this is good enough standalone.

Issue #33179, opened 2021-04-18
[4.0] Files uploaded to templates with upper case extensions don't appear in folder tree.


particthistle:
### Steps to reproduce the issue
- Prepare a file and have an uppercase file extension eg sample.PNG
- System > Site Templates (or Administrator Templates)
- Click on template to modify template files
- Click New File to upload the new file to the template

### Expected result
Uploaded file appears in the template folder tree so you can further perform actions on the file.

### Actual result
File is uploaded, and success message is displayed, but the file does not appear in the folder tree.
![image](https://user-images.githubusercontent.com/5515866/115143699-36315380-a08c-11eb-99bd-059df666ac96.png)

### Additional comments
Occurs with any file type where the extension is uppercase.

Though you can add the upper case extensions to your list of allowed extensions to fix it, this is not necessarily the best solution, and instead perhaps the extension needs to be converted to lower case on upload?

Discovered while testing https://github.com/joomla/joomla-cms/pull/33087

ghost:
Upload 1.PNG show now:

> ![image](https://user-images.githubusercontent.com/78906980/116781353-91971480-aa82-11eb-8606-93de7b36bdba.png)

As user its hard to know that the file-extension has to be lowercase.

brianteeman:
please test #34065

richard67:
Closing as having a pull request. Please test #34065 . Thanks in advance.

brianteeman:
Please re-open

richard67:
Re-opening as the pull request has been closed.

joomdonation:
PR https://github.com/joomla/joomla-cms/pull/34095 should address this issue. Please test, thanks !

PR #34065, opened 2021-05-20
[4.0] template manager uppercase filenames

brianteeman:
All uploaded files are saved as lowercase.

Try to upload a file with an uppercase extension eg test.JPG

It fails

Apply PR

Try to upload a file with an uppercase extension eg test.JPG

Success - saved as test.jpg

Pull Request for Issue #33179

### DIFF
--- a/administrator/components/com_templates/src/Helper/TemplateHelper.php
+++ b/administrator/components/com_templates/src/Helper/TemplateHelper.php
@@ -83,7 +83,7 @@ abstract class TemplateHelper
 			}
 		}
 
-		if ($file['name'] !== File::makeSafe($file['name']) || preg_match('/\s/', File::makeSafe($file['name'])))
+		if (strtolower($file['name']) !== File::makeSafe($file['name']) || preg_match('/\s/', File::makeSafe($file['name'])))
 		{
 			$app = Factory::getApplication();
 			$app->enqueueMessage(Text::_('COM_TEMPLATES_ERROR_WARNFILENAME'), 'error');


ghost:
I have tested this item :white_check_mark: successfully on 18d57f52572c8bb47d52a6d2c8355703a071e5d1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34065">issues.joomla.org/tracker/joomla-cms/34065</a>.</sub>

joomdonation:
From a quick code review, this PR looks wrong to me. It will prevent file such as **Test.jpg** from being uploaded and it could not be right ?

If I understand the problem right, what is happening right now is the file **test.JPG** still being uploaded, but it is invisible right now and the right fix would be make it visible ? If so, the right fix would be modify `checkFormat` method in **TemplateModel** https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_templates/src/Model/TemplateModel.php#L1972 so that file extension is case-insensitive (mean both Test.jpg and Test.JPG will be displayed in template manager).

Modify that method to the following code should work:

```php
protected function checkFormat($ext)
{
		if (!isset($this->allowedFormats))
		{
			$params       = ComponentHelper::getParams('com_templates');
			$imageTypes   = explode(',', $params->get('image_formats'));
			$sourceTypes  = explode(',', $params->get('source_formats'));
			$fontTypes    = explode(',', $params->get('font_formats'));
			$archiveTypes = explode(',', $params->get('compressed_formats'));

			$this->allowedFormats = array_merge($imageTypes, $sourceTypes, $fontTypes, $archiveTypes);
			$this->allowedFormats = array_map('strtolower', $this->allowedFormats);

		}

		return in_array(strtolower($ext), $this->allowedFormats);
}
```

brianteeman:
> If I understand the problem right, what is happening right now is the file test.JPG still being uploaded, but it is invisible right now and the right fix would be make it visible ?

No. In my tests using testone.JPG and Testtwo.JPG both gave the following error message and were not uploaded

>Invalid file name. Please correct the name of the file and upload again.
Failed to upload file.


Happy to admit my code might not be correct but after testing your suggestion the error remained.

With this PR both were uploaded successfully as testone.jpg and testtwo.jpg

However I now see that a file uploaded to the server directly that is mixed case is not displayed in the folder tree - with or without this pr or your suggestion

joomdonation:
Hmm. I just tried to use today nightly build and file is uploaded 
success_uploaded

It is just not being displayed (due to the file extension case) and with the suggested code, it is being displayed. Maybe I mis-understand the issue somehow.

brianteeman:
I rechecked and still cant upload

The problem with seeing the file in the tree turns out to be super super easy
image

I am going to close this as its obviously not the right way.

joomdonation:
@brianteeman I think I know the reason of the error. You cannot upload file with upercase character because this block of code https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Filesystem/File.php#L86-L90

Not sure why we decide to make the filename become lowercase in that method like that. Seems a bug to me.

PR #34069, opened 2021-05-21
[4.0] refactor the media adapters

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/26750#issuecomment-739248621.

### Summary of Changes

This is a proposal to change the adapters from ie: `local-0` to ie: `local-images` as sated https://github.com/joomla/joomla-cms/issues/26750#issuecomment-739248621.
Probably if this is ok (I think the second part would be better if it's `md5($name)`) would be merged to #33724 for easier testing?

@joomdonation @Fedik @bembelimen your thoughts here?

### Testing Instructions
Goto: `administrator/index.php?option=com_plugins&view=plugins&filters=[type=filesystem]`
Edit `FileSystem - Local` plugin by adding a new path, eg `installation`
Check the media manager
Re-edit the previous plugin by changing the order of the folders (first installation, second images)
Check again the media manager still operates fine

Also, edit any article (or create a new one):
- with tinMCE try to upload an image with drag and drop
- try to select an image for intro or full position

There shouldn't be any difference for end-users. This PR just ensures that the adapters are not index related.


richard67:
@dgrammatiko The linter is still a bit unhappy: https://ci.joomla.org/joomla/joomla-cms/44184/1/21 :
```
/********/src/administrator/components/com_media/resources/scripts/store/state.es6.js
  13:15  error  'key' is assigned a value but never used  no-unused-vars
```

dgrammatiko:
> @dgrammatiko The linter is still a bit unhappy

That was an easy fix but this PR is breaking B/C. (Anyways I was asked if I could help to transition from the index-based naming to this new name-based, so I guess this is already justified)

administrator/components/com_media/src/Controller/ApiController.php
```diff
@@ -374,7 +374,7 @@ private function checkContent()
 	 */
 	private function getAdapter()
 	{
-		$parts = explode(':', $this->input->getString('path', ''), 2);
+		$parts = explode(':', $this->input->getString('path', ''));
```
Fedik: better keep it with limit 2,
other places also

Fedik:
looks good, if it works ;) 

I not remember where exactly, in past I seen places where second part of the adapter is cast to integer,
I mean stuff like:
```
$blabla = (int) $part[1];
```

administrator/components/com_media/src/Provider/ProviderManager.php
```diff
@@ -89,7 +89,7 @@ public function getProvider($id)
 	 */
 	public function getAdapter($name)
 	{
-		list($provider, $account) = array_pad(explode('-', $name, 2), 2, null);
```
Fedik: this was made to be sure the array always have 2 items,
Because you will get PHP notice `Undefined $account` when someone will submit invalid string, eg `blabla:/`

dgrammatiko:
> I not remember where exactly, in past I seen places where second part of the adapter is cast to integer,

A quick search returned no casting `(int)` results in com_media or the plugin

administrator/components/com_media/src/Provider/ProviderManager.php
dgrammatiko: Do you want me to restore it?

Fedik:
> A quick search returned no casting

Maybe I mixed something, thanks for checking

administrator/components/com_media/src/Provider/ProviderManager.php
dgrammatiko: Restored

plugins/filesystem/local/local.php
```diff
@@ -94,7 +94,9 @@ public function getAdapters()
 			{
 				$directoryPath = JPATH_ROOT . '/' . $directoryEntity->directory;
 				$directoryPath = rtrim($directoryPath) . '/';
-				$adapters[]    = new \Joomla\Plugin\Filesystem\Local\Adapter\LocalAdapter($directoryPath, $directoryEntity->directory);
+				$adapters[$directoryEntity->directory] = new \Joomla\Plugin\Filesystem\Local\Adapter\LocalAdapter(
```
joomdonation: Almost there. To make it works well (hard to explain), we will have to change this block of code to:

```php
foreach ($directories as $directoryEntity)
{
	if ($directoryEntity->directory)
	{
		$directoryPath = JPATH_ROOT . '/' . $directoryEntity->directory;
		$directoryPath = rtrim($directoryPath) . '/';

		$adapter = new \Joomla\Plugin\Filesystem\Local\Adapter\LocalAdapter(
			$directoryPath, $directoryEntity->directory
		);

		$adapters[$adapter->getAdapterName()] = $adapter;
	}
}
```
Now, the name of the adapters have this format: **local-images**, **local-sampledata**
images, sampledata is returned by $adapter->getAdapterName() method, so later, to allow the system to find the right adapter from path in the request, we will have to use `$adapter->getAdapterName()` as key name of the item.

joomdonation: I hope this is understandable. Hard to explain but it is still not clear, I will try to find a better way to explain. I only see this problem while checking to see why dropbox adapter does not work after this change.

administrator/components/com_media/resources/scripts/store/state.es6.js
```diff
@@ -6,12 +6,25 @@ if (options.providers === undefined || options.providers.length === 0) {
...
 
+/**
+ * Get the drives
+ *
+ * @param  {Array}  adapterNames
+ * @param  {String} adapterName
+ *
+ * @return {Array}
+ */
+const getDrives = (adapterNames, adapterName) => {
```
joomdonation: The parameters name of function here is a bit confusing. How about changing it to const getDrives = (adapterNames, provider). In the server side, we call `local`, `dropbox`  as providers, so I think we should make the same change here to make it consistent.

plugins/filesystem/local/local.php
dgrammatiko: Done

administrator/components/com_media/resources/scripts/store/state.es6.js
dgrammatiko: I'm not sure if we need to change anything here as the object (client-side) remains the same:
Screenshot 2021-05-21 at 17 37 55

dgrammatiko: Ok, done, arguments renamed

joomdonation:
Thanks @dgrammatiko. All good. I will now run a final test.

joomdonation:
Sorry. Could you please do a final change https://github.com/joomla/joomla-cms/blob/4.0-dev/layouts/joomla/form/field/media.php#L107 ? Change `local-0` to `local-images`

Later it will be changed anyway. But this is to avoid when your PR merged, people use nightly build won't get error while trying to select an image for article.

joomdonation:
I have tested this item :white_check_mark: successfully on f194566f20b195a85ec1ebe587e2e0d3ac7bfeae<br /><br />Thanks, all good now. Great work !<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34069">issues.joomla.org/tracker/joomla-cms/34069</a>.</sub>

dgrammatiko:
@richard67 I guess this one should have a RB tag as it partially solves one of the issues of https://github.com/joomla/joomla-cms/issues/26750 The complete solution will be a combo of this and https://github.com/joomla/joomla-cms/pull/33724 from @joomdonation. (I guess you can also skip as most probably people are already aware of the state and why this PR created in the first place 🤷‍♂️)

richard67:
I have tested this item :white_check_mark: successfully on 04fa696d0228795689f1b6bec9916707143b899f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34069">issues.joomla.org/tracker/joomla-cms/34069</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34069">issues.joomla.org/tracker/joomla-cms/34069</a>.</sub>

joomdonation:
@rdeutz @wilsonge Could we have this PR merged ? I have an open RB PR which depends on this one and want to have it finished ASAP.

Fedik:
Thanks!

PhilETaylor:
hold the front page...

brianteeman:
I got this error a few times but its gone now

image

PhilETaylor:
Thats what I got for ages... - its cache though... its still calling urls with 

`administrator/index.php?option=com_media&format=json&task=api.files&path=local-0:/`

seems Joomla 4 cache busing is not that good at the moment.

dgrammatiko:
There's a sessionStorage entry, so unless you do some update without killing the session then this shouldn't be a problem

PhilETaylor:
actually Im not convinced it is cache... go to Home Dashboard, and click the media icon in the dashboard

Links to http://127.0.0.1:4444/administrator/index.php?option=com_media

PhilETaylor:
ah - let me try killing that

PhilETaylor:
yup that was it - clear as day it had local-0 in the sessionStorage entry,

PhilETaylor:
So if someone had a Joomla 4 beta, and had been to media in this session, and upgrades to Joomla 4 RC1 they COULD have this problem right?

dgrammatiko:
Yeah, but I think it's highly unlikely to hit all these requirements at once...

PhilETaylor:
> Yeah, but I think it's highly unlikely to hit all these requirements at once...

/bookmarked.... 

Screenshot 2021-05-22 at 19 31 19

brianteeman:
Highly unlikely as in two people already did

richard67:
Is there anything we can do against it?

dgrammatiko:
> Is there anything we can do against it?

We could empty the sessionStorage key before the update, but it won't cover all the updade paths

dgrammatiko:
But we can ask people to log out after the update and re-login (in case they used com_media **right** before the update). No code needed

PR #34074, opened 2021-05-21
[4] print_popup function missing, provide exception

PhilETaylor:
code review.

This is not used by Joomla at all, it might be used by 3PD **BUT** it would never work and would generate a PHP Fatal error as the method doesnt exist in the class. 

Somewhere previously someone decided to deprecate the `print_popup` method for removal in Joomla 5. 

This method calls `\Joomla\Component\Content\Administrator\Service\HTML\Icon::print_popup` which doesnt exist. 

Rather than write that method from scratch, for an obviously deprecated future, I have taken the decision to throw an Exception, instead of a PHP Fatal error which seems to make more sense. 

Without this change, if a 3PD were to implement `JHtmlIcon::print_popup` static method, they would get a PHP Fatal error but now they will get accurate feedback

### DIFF
--- a/components/com_content/helpers/icon.php
+++ b/components/com_content/helpers/icon.php
@@ -72,7 +72,7 @@ abstract class JHtmlIcon
 	 */
 	public static function print_popup($article, $params, $attribs = array(), $legacy = false)
 	{
-		return self::getIcon()->print_popup($article, $params, $attribs, $legacy);
+		throw new \Exception('Sorry, Joomla 4 no longer supports print_popup');
 	}
 
 	/**


richard67:
@PhilETaylor You wrote in the description:

> Without this change, if a 3PD were to implement `JHtmlIcon::print_popup` static method, they will get accurate feedback and not a PHP Fatal error.

Is it a typo and you meant "With this change, ..."?

PhilETaylor:
I obviously got distracted half way through the sentence. thanks.

richard67:
I have tested this item :white_check_mark: successfully on 6365e9f896b466c2e40885c7b539422e173fff77<br /><br />Code review:
- The `self::getIcon()` call returns a new object of class `\Joomla\Component\Content\Administrator\Service\HTML\Icon` here https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_content/helpers/icon.php#L102 .
- The class is defined here https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_content/src/Service/HTML/Icon.php . It doesn't have a `print_popup` method and has no parent class which you have one.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34074">issues.joomla.org/tracker/joomla-cms/34074</a>.</sub>

wilsonge:
Thanks!

PR #34075, opened 2021-05-21
[4] Push deprecation of CMSObject to 5.0.0 instead of 4.0.0

PhilETaylor:
Code review

It seems previously someone was optimistic that they could remove `CMSObject` in Joomla 4.0.0 

As this clearly has not happened, and Joomla 4.0.0 very much still depends on `CMSObject` I have updated the `@deprecated` tag to the next major series, which will be the next opportunity to remove this class.

### DIFF
--- a/libraries/src/Object/CMSObject.php
+++ b/libraries/src/Object/CMSObject.php
@@ -17,7 +17,7 @@ namespace Joomla\CMS\Object;
  * and an internal error handler.
  *
  * @since       1.7.0
- * @deprecated  4.0.0
+ * @deprecated  5.0.0
  */
 class CMSObject
 {


brianteeman:
deprecated in 4 removed in 5

PhilETaylor:
open to interpretation unfortunately

There are MANY interpretations in Joomla of the use of the `@deprecated` tag

Eg:

```
* @deprecated  5.0 Without replacement

* @deprecated 4.0.0

```

PhilETaylor:
https://pear.php.net/package/PhpDocumentor/docs/latest/phpDocumentor/tutorial_tags.deprecated.pkg.html

```
Document elements that have been deprecated and should not be used as they may be removed at any time from a future version
```

So the "future version" given is the version that the code can be removed from. Is my interpretation.

brianteeman:
There was a production department decision on this many many years ago. I think by @chrisdavenport

PhilETaylor:
Search for `@deprecated` ... soooo painful to see. 

Screenshot 2021-05-21 at 13 08 40

brianteeman:
Found it

6.2 https://developer.joomla.org/development-strategy.html

PhilETaylor:
Shall I point out all the ways that section is wrong haha... 

> An element can be marked as being deprecated in a new minor release or the first release of a new major version. Advice on how to amend code that is currently using an element that is marked as deprecated MUST be added to the documentation.

And yet there are 1000s without this documentation after the tag

> Previously deprecated elements that are no longer used in the product itself MUST be removed in the first version of a new major series but MUST NOT be removed in a minor or patch release.

Like 4.0.0...... yet there are loads that have not been and we are about to go to RC.

> Code that is deprecated will have a @deprecated tag added to the appropriate docblock and it will be noted in the release notes.

Never happens.

PR #34078, opened 2021-05-21
[4] Try to load a language string from the installation language var will be removed in 4.0.

PhilETaylor:
Code review.

Code says it will be removed in 4.0.

This PR removes it.

### DIFF
--- a/libraries/src/Installer/Adapter/LanguageAdapter.php
+++ b/libraries/src/Installer/Adapter/LanguageAdapter.php
@@ -906,38 +906,6 @@ class LanguageAdapter extends InstallerAdapter
 			$contentLanguageNativeTitle = $siteLanguageManifest['nativeName'];
 		}
 
-		// Try to load a language string from the installation language var. Will be removed in 4.0.
-		if ($contentLanguageNativeTitle === $contentLanguageTitle)
-		{
-			$manifestfile = JPATH_INSTALLATION . '/language/' . $tag . '/langmetadata.xml';
-
-			if (!is_file($manifestfile))
-			{
-				$manifestfile = JPATH_INSTALLATION . '/language/' . $tag . '/' . $tag . '.xml';
-			}
-
-			if (file_exists($manifestfile))
-			{
-				$installationLanguage = new Language($tag);
-				$installationLanguage->load('', JPATH_INSTALLATION);
-
-				if ($installationLanguage->hasKey('INSTL_DEFAULTLANGUAGE_NATIVE_LANGUAGE_NAME'))
-				{
-					// Make sure it will not use the en-GB fallback.
-					$defaultLanguage = new Language('en-GB');
-					$defaultLanguage->load('', JPATH_INSTALLATION);
-
-					$defaultLanguageNativeTitle      = $defaultLanguage->_('INSTL_DEFAULTLANGUAGE_NATIVE_LANGUAGE_NAME');
-					$installationLanguageNativeTitle = $installationLanguage->_('INSTL_DEFAULTLANGUAGE_NATIVE_LANGUAGE_NAME');
-
-					if ($defaultLanguageNativeTitle !== $installationLanguageNativeTitle)
-					{
-						$contentLanguageNativeTitle = $installationLanguage->_('INSTL_DEFAULTLANGUAGE_NATIVE_LANGUAGE_NAME');
-					}
-				}
-			}
-		}
-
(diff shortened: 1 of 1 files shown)


richard67:
Comments can be wrong. Wouldn't it be better to do a real test to make sure it's really to be removed?

brianteeman:
> Comments can be wrong. Wouldn't it be better to do a real test to make sure it's really to be removed

and then update the comment if appropriate

PhilETaylor:
Did you test before commenting? Do you think I submit PRs without testing?

richard67:
> Did you test before commenting? Do you think I submit PRs without testing?

@PhilETaylor No, haven't tested yet. I just asked to be sure you've tested because testing instructions are only about code review.

PhilETaylor:
The clue is `JPATH_INSTALLATION` which means that this block of code can only ever be sensibly run when installing Joomla.

I was able to install Joomla 4 with French, and Persion languages with no errors during the installer process. 

Screenshot 2021-05-21 at 13 27 29

And I was able to complete the install and install the multi lingual test data with no issues after deleting the installation folder.

Screenshot 2021-05-21 at 13 27 52

richard67:
> I was able to install Joomla 4 with French, and Persion languages with no errors during the installer process.

@PhilETaylor Then I hope it would not be the German language which would break it 😜 (joking).

Thanks for reporting back.

infograf768:
@PhilETaylor 
(from ipad) Have you checked Content Languages Native Names?

PhilETaylor:
And how would a person do that? No idea.

infograf768:
@PhilETaylor 
Made some research on this and found out the reason for that code.
We `could` indeed take it off from 4.0 as I expect that totally new language packs for that version will contain the `nativeName` metadata in its **site** `langmetadata.xml`.

The reason we have the code in 3.x is because some languages were not updated and do not contain that metadata.

Example for Galician site `gl-ES.xml` ( http://joomlacode.org/gf/download/frsrelease/19631/159913/gl-ES_joomla_lang_full_3.3.1v2.zip ): 
```
<metadata>
		<name>Galician (Galego)</name>
		<tag>gl-ES</tag>
		<rtl>0</rtl>
		<locale>gl_ES.utf8, gl_ES.UTF-8, gl_ES.UTF-8@euro, gl_ES, gal_ES, gl, galician</locale>
		<firstDay>1</firstDay>
		<weekEnd>0,6</weekEnd>
	</metadata>
```

But it contains the installation key which is used as fallback to create the `Title Native` for the Content Language grace to the code.

```
; IMPORTANT NOTE FOR TRANSLATORS: Do not literally translate this line, instead add the localised name of the language. For example Spanish will be Español
INSTL_DEFAULTLANGUAGE_NATIVE_LANGUAGE_NAME="Galego"
```

Which gives in 3.x

Screen Shot 2021-05-22 at 09 31 42

So for 3.x, we have 2 fallbacks:

Use the metadata [name] if no Site metadata{nativeName] and if no specific `INSTL_DEFAULTLANGUAGE_NATIVE_LANGUAGE_NAME` key.

If the code is deleted and you install in J4 the 3.x Galician pack, the Title Native will be `Galician (Galego)` which obviously is not native at all...

**If whoever now in charge** makes sure that the `metadata[nativeName]` is present for ALL new J4 lang packs, we can delete that code.

Hope it helps.

richard67:
@infograf768 Thanks for your research and detailed information. So the question is if we can be sure about the metadata[nativeName] being present for ALL new J4 lang packs, and this is related to #21806 being solved, i.e. the new download infrastructure being ready and true J4 language packs being provided there.

@wilsonge Any news on that?

infograf768:
IMHO, it does not so much depends on the availability of the new download infrastructure (evidently as totally necessary), but of a control of the new packs.

wilsonge:
> If whoever now in charge makes sure that the metadata[nativeName] is present for ALL new J4 lang packs, we can delete that code.

I have no clue who's in charge now. @marcodings can you provide insight please

PhilETaylor:
Paging @marcodings again... else lets just merge this or close it. No point dragging it out for another decade.

infograf768:
The language packs or installation ini & xml files are not controlled anymore by Production. I would not expect @marcodings to reply to you as he is the one who managed for Programs to be responsible for Core Translations.

PhilETaylor:
So that begs the question: Who is actually responsible now? Who should this issues be addressed to?

PR #34092, opened 2021-05-22
[4.0] Use MVCFactory to create model

joomdonation:
Pull Request for Issue # .

### Summary of Changes
In Joomla 4, the right way to create a model object is use `MVCFactory` after `bootComponent`. For example https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_categories/src/Helper/CategoriesHelper.php#L99

However, there are still some places in our code use old method (create model class directly using `new ModelClassName` syntax). This PR improves code to make it consistent in our whole codebase.

### Testing Instructions
The modified code is random in our CMS, so I think the best way is someone with code review skill to review the code and make sure it is all correct. Or maybe @wilsonge will need to review and merge base on code review.

### DIFF
--- a/administrator/components/com_actionlogs/src/Plugin/ActionLogPlugin.php
+++ b/administrator/components/com_actionlogs/src/Plugin/ActionLogPlugin.php
@@ -13,7 +13,6 @@ namespace Joomla\Component\Actionlogs\Administrator\Plugin;
 
 use Joomla\CMS\Factory;
 use Joomla\CMS\Plugin\CMSPlugin;
-use Joomla\Component\Actionlogs\Administrator\Model\ActionlogModel;
 
 /**
  * Abstract Action Log Plugin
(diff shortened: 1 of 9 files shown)


richard67:
@joomdonation There are PHP code style errors: https://ci.joomla.org/joomla/joomla-cms/44216/1/6

wilsonge:
I’m happy with most places on review.

Only thing I wouldn’t mind tested is the script.php part doing a J3 to j4 upgrade - just because I don’t know exactly where that runs in the upgrade cycle and whether we’ll be exclusively using the j4 files everywhere (I think it will be fine - but better safe than sorry)

joomdonation:
@wilsonge I installed Joomla 3.10.0 nightly, tested the update process and it is working well, so I think it is fine.

For technical information, the code which is changed here will be executed when `finaliseUpgrade` method of `UpdateModel` is called. It is called when:

- The system calls `finalise` method of `UpdateController`. This is run in Joomla context, so it is fine (I tested in above test) (somewhere during update process, there is a redirection to index.php?option=com_joomlaupdate&task=update.finalise)
- And in UpdateCoreCommand https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Console/UpdateCoreCommand.php#L220 . For console commands, I don't understand the workflow of the process, so I don't not know if it works or not.

If it is needed, I will remove the code which I changed in com_admin/script.php to be safe. Let me know if you want me to do that.

wilsonge:
I’m happy with that explanation. Thanks!

PR #34147, opened 2021-05-23
[4.0.0] Deprecate mcrypt use in decrypting settings for 2fa

PhilETaylor:
### Summary of Changes

The mcrypt extension has been abandonware for over a decade now, and was also fairly complex to use. It has therefore been deprecated in favour of OpenSSL, where it will be removed from the core and into PECL in PHP 7.2. It was deprecated in PHP 7.1

There is one use in Joomla 4 that can still be triggered, and that is to decrypt the configuration and OTEP values from the database, if previously encrypted with mcrypt. 

If this case is triggered then Joomla decrypts using mcrypt and re-encrypts using openssl using the AES Class. 

**THIS CODE HAS NEVER WORKED ANYWAY SINCE BEING ADDED** due to a typo in the constructor arguments  when setting up the `$mcrypt` instance. The number of params provided meant that the `mcrypt' value last in the params list was never used, meaning that `$mcrypt` was always an instance using the openssl adapter. 

As Joomla 4.0.0 is about to be released (cough, maybe some time, but the RC will be) we need to slip these deprecated tags in else it will be another decade before we can remove this code. 

@wilsonge urgent review please :) :) :)

### Testing Instructions

Code review. 

// cc @nikosdion @joomdonation 
// replacement for https://github.com/joomla/joomla-cms/pull/33954

### DIFF
--- a/administrator/components/com_users/src/Model/UserModel.php
+++ b/administrator/components/com_users/src/Model/UserModel.php
@@ -1043,7 +1043,7 @@ class UserModel extends AdminModel
 			// Attempt to decrypt using the mcrypt adapter, under normal circumstances this should fail (We no longer use mcrypt adapter to encrypt).
 			$decryptedConfig = $mcrypt->decryptString($config);
 
-			// If we were able to decrypt correctly, { will be in the config (JSON String), so lets update to openssl adapter use.
+			// If we were able to decrypt using the mcrypt adapter, { will be in the config (JSON String), so lets update to openssl adapter use.
 			if (strpos($decryptedConfig, '{') !== false)
 			{
 				// Data encrypted with mcrypt, decrypt it, and then convert to openssl.
--- a/administrator/components/com_users/src/Model/UserModel.php
+++ b/administrator/components/com_users/src/Model/UserModel.php
@@ -1040,7 +1040,7 @@ class UserModel extends AdminModel
 
 			$decryptedConfig = $mcrypt->decryptString($config);
 
-			// If we were able to decrypt correctly, { will be in the config, so lets get rid of mcrypt and update to openssl encryption.
+			// If we were able to decrypt correctly, { will be in the config, so lets get rid of mcrypt and update to openssl adapter use.
 			if (strpos($decryptedConfig, '{') !== false)
 			{
 				// Data encrypted with mcrypt, decrypt it, and then convert to openssl.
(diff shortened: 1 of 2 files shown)


administrator/components/com_users/src/Model/UserModel.php
```diff
@@ -1034,19 +1034,22 @@ public function getOtpConfig($userId = null)
 		if (strpos($config, '{') === false)
 		{
 			$openssl         = new Aes($key, 256);
-			$mcrypt          = new Aes($key, 256, 'cbc', null, 'mcrypt');
+
+			// Deal with legacy mcrypt encrypted data, and convert it to openssl encrypted data.
+			$mcrypt          = new Aes($key, 256, 'cbc', 'mcrypt');
```
richard67:
```suggestion
			$mcrypt = new Aes($key, 256, 'cbc', 'mcrypt');
```
If not directly below another assignment anymore, it doesn't need to align the "=".

PhilETaylor: yup I was just adding more comments - sorry for the delay

richard67:
I have tested this item :white_check_mark: successfully on e9ab4ee902e93d9c8bf582bbbeb4e3d03036bd55<br /><br />Code review.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34147">issues.joomla.org/tracker/joomla-cms/34147</a>.</sub>

PhilETaylor:
drone failure unrelated to PR content :)

wilsonge:
More than happy to merge the deprecation stuff. Honestly the constructor thing is one of those things where - do we break things more by fixing it than it was before xD #scaredMaintainer :)

PhilETaylor:
well its clearly never worked ever... so maybe I just document that it has never ever worked and put verbose notes to say its never ever worked and a note to say to remove it in 5.0.0 rather than fix it. 

What it means is that any OTP config, currently encrypted with mcrypt, will not be able to be decrypted and migrated to openssl. Its a very very fringe case anyway. 

Obviously the correct course of action is to fix it properly. :-)

PhilETaylor:
I have moved the deprecation to here: https://github.com/joomla/joomla-cms/pull/34159 for you to merge on sight.

PhilETaylor:
and the reset of the commenting is here: https://github.com/joomla/joomla-cms/pull/34160

So we dont actually fix the problem anymore, just document it, ready for deletion in 5.0.0 and thus I'll close this PR and we can leave this bug alone.

PR #34152, opened 2021-05-24
[3] Fix Delete Batch for Content Maps

YatharthVyas:
### Summary of Changes
Redo of PR #33771 for J3

#### Before testing, Please take a backup of `#_finder_taxonomy` table to avoid the loss of any data

1. Visit Control Panel (Backend)
2. Top Navbar -> Extensions -> Plugins -> Search 'Content - Smart Search' -> Enable the plugin
3. Components -> Smart Search -> Content Map
4. Ensure that you have sufficient data in this table with multiple parent-level and child-level maps (Installting Blog Sample Data will also work)
5. Click on the select all checkbox and proceed to delete them.

Before the PR, only the first parent and its children would get deleted and then the function would return false skipping the remaining elements.
After the PR, you should see the Empty State, ie. All items deleted successfully.

maps_after

None

### DIFF
--- a/administrator/components/com_finder/models/maps.php
+++ b/administrator/components/com_finder/models/maps.php
@@ -86,6 +86,17 @@ class FinderModelMaps extends JModelList
 		// Include the content plugins for the on delete events.
 		JPluginHelper::importPlugin('content');
 
+		// Iterate the items to check if all of them exist.
+		foreach ($pks as $i => $pk)
+		{
+			if (!$table->load($pk))
+			{
+				$this->setError($table->getError());
+
+				return false;
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


richard67:
@YatharthVyas Content maps are created by using the Smart Search component. If you go to that component in backend, there might be an alert telling that some plugin is not enabled. Enable that. Then you can go to the index page of smart search and start indexing, if it hasn't done before. After that there should be content maps.

YatharthVyas:
Thanks @richard67 
I enabled the Smart Search Component and the content maps were generated. I couldn't replicate the issue reported for J4 in J3 so there is no need for this PR anymore. Everything is working as it should.

PR #34157, opened 2021-05-24
[4.0] Sidebar badge contrast [a11y]

brianteeman:
On a multilingual site there is a badge for the default language menu item. This is failing the contrast checks.

this is an scss change so dont forget to rebuild the css for testing

image

### DIFF
--- a/administrator/templates/atum/scss/blocks/_sidebar.scss
+++ b/administrator/templates/atum/scss/blocks/_sidebar.scss
@@ -185,7 +185,7 @@
   .badge {
     align-self: center;
     margin: 0 .3rem .25rem;
-    background-color: var(--atum-bg-dark-50);
+    background-color: var(--atum-bg-dark-60);
   }
 }
 


infograf768:
Suggest also adding a white border as the background would not stand much vs the sidebar background color.

```
background-color: var(--atum-bg-dark-60);
border: 1px solid white;
```
which would give

Screen Shot 2021-05-24 at 10 29 02

instead of this with this pr

Screen Shot 2021-05-24 at 10 32 16

brianteeman:
Not required for accessibility. That's a styling change which is beyond the scope of this pr and should be considered as a separate issue as it would impact the consistency of badge styling elsewhere

infograf768:
> should be considered as a separate issue as it would impact the consistency of badge styling elsewhere

I  doubt about that. `main-nav` is only used by the admin mod_menu.

css is in _sidebar.scss
```
// Sidebar navigation
.main-nav {

[...]

.badge {
    align-self: center;
    margin: 0 .3rem .25rem;
    background-color: var(--atum-bg-dark-50);
  }
```
Just say you don't want to do the change in a a11y PR. That's enough.

brianteeman:
I meant consistency with badges everywhere not just sidebar.

eopws:
Maybe out of topic, but why duplicate language tag?

infograf768:
> Maybe out of topic, but why duplicate language tag?

It is not a duplicate. The tag added to the menu title comes from the multilingual sample data and it is there on purpose.
On a real site it is not necessary.

wilsonge:
Thanks!

brianteeman:
thank you

Issue #34117, opened 2021-05-23
[4.0] Global Configuration - Sidebar collapse toggle incorrectly initialized for smaller screens


YatharthVyas:
### Steps to reproduce the issue

1. Visit the Global Configuration page in the Backend.
2. Switch to the mobile view using Devtools.
3. **Refresh the page while being in the mobile view** or navigate outside and back to this page to imitate how a mobile user would enter this page.
4. Notice that the sidebar dropdown is initially expanded (Contrary to `aria-expanded="false"`) and when you press the toggle button, it goes through an animation to stay in the expanded state (class changes from `sidebar-nav` to `sidebar-nav collapsing` and then to `sidebar-nav collapse show`)

### Expected result
When you resize the screen after it has loaded, then the sidebar is initially hidden and the toggle button behaves as intended.

![config_menu2](https://user-images.githubusercontent.com/53610833/119253328-5e9df780-bbce-11eb-86f5-37ca45e25023.gif)

### Actual result
On refreshing the page in mobile view (without resizing), the menu is expanded by default and the first toggle button press doesn't collapse it.

![config_menugif](https://user-images.githubusercontent.com/53610833/119252855-12ea4e80-bbcc-11eb-9026-5ec85741ac55.gif)

### System information (as much as possible)
Google Chrome on Windows 10

PhilETaylor:
Confirmed on real mobile iPhone 12

https://user-images.githubusercontent.com/400092/119260036-e41ea900-bbc8-11eb-9a56-b44b16ccec90.mov

eopws:
> 4\. class changes from `sidebar-nav` to `sidebar-nav collapsing` and then to `sidebar-nav collapse show`

I have tried to add `sidebar-nav collapse show` to the list initially and it works and doesn't ruin anything, but I don't know whether it is a good solution or not?

eopws:
IMHO, the "toggle menu" button has to have margin-left.

infograf768:
> IMHO, the "toggle menu" button has to have margin-left.

If it does in LTR, then it should have a margin-right in RTL

YatharthVyas:
> > 4. class changes from `sidebar-nav` to `sidebar-nav collapsing` and then to `sidebar-nav collapse show`
> 
> I have tried to add `sidebar-nav collapse show` to the list initially and it works and doesn't ruin anything, but I don't know whether it is a good solution or not?

[...]

PR #33877, opened 2021-05-14
Fix the mobile menu wrongly expanding the logo

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/pull/33754#issuecomment-841373252 and #34105

### Summary of Changes

> Joomla logo is expanded pushing the cog icon out of view and toolbar options are displayed by default.

### Testing Instructions

Shrink the width of your browser to a mobile width and open the menu/toolbar. The logo should not expand

PR #34158, opened 2021-05-24
[4.0] Sidebar collapse toggle incorrectly initialized for smaller screens

eopws:
Pull Request for Issue #34117.

### Summary of Changes

Page resize logic updated. Now the resize function called when page loaded in order to solve whether the menu should be collapsed or not (according to `aria-expanded="false"` on mobile the menu should be collapsed initially).

### Testing Instructions

1. Visit the Global Configuration page in the Backend.
2. Switch to the mobile view using Devtools.
3. Refresh the page while being in the mobile view or navigate outside and back to this page to imitate how a mobile user would enter this page.
4. Notice that the sidebar dropdown is initially collapsed (Contrary to the previous behaviour) and when you press the toggle button animation is ok contrary to the previous behaviour.

test

after

Added the `resizeAction` function and a comment to it. Please comment if documentation is incorrect (e.g. english grammar)

### DIFF
--- a/build/media_source/templates/administrator/atum/js/template.es6.js
+++ b/build/media_source/templates/administrator/atum/js/template.es6.js
@@ -149,14 +149,19 @@ function setMobile() {
   }
 
   if (small.matches) {
-    if (sidebarNav) sidebarNav.classList.add('collapse');
     if (subhead) subhead.classList.add('collapse');
     if (sidebarWrapper) sidebarWrapper.classList.add('collapse');
   } else {
-    if (sidebarNav) sidebarNav.classList.remove('collapse');
     if (subhead) subhead.classList.remove('collapse');
     if (sidebarWrapper) sidebarWrapper.classList.remove('collapse');
   }
+
+  if (mobile.matches) {
+    if (sidebarNav) sidebarNav.classList.add('collapse');
+  } else {
+    if (sidebarNav) sidebarNav.classList.remove('collapse');
+  }
+
   changeLogo('closed');
 }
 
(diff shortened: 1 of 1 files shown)


dgrammatiko:
There's a;ready a PR for this: https://github.com/joomla/joomla-cms/pull/33877

dgrammatiko:
Please undo the changes here there's a PR already: https://github.com/joomla/joomla-cms/pull/33877

eopws:
> Please undo the changes here there's a PR already: #33877

Ok.

eopws:
@dgrammatiko
Maybe I am wrong, but it seems that the #33877 PR doesn't solve the #34117 issue problem ・_・

dgrammatiko:
> doesn't solve the #34117 issue problem

Of course, it doesn't because the template CANNOT and SHOULD NOT interfere with specific page/component behaviours. This PR is totally wrong, you're deferring a responsibility of a specific page to the template, that's not how things should work...

eopws:
> This PR is totally wrong, you're deferring a responsibility of a specific page to the template, that's not how things should work...

Not at all, the logic of this PR is to call 'resize' function when a page loads.

dgrammatiko:
Please close it, it's wrong...

dgrammatiko:
So, now #33877 should fix also #34117, please test

eopws:
> So, now #33877 should fix also #34117, please test

Yes, it works, but for really small screens and still have the issue for screens beetwen 575.98px and 992px (sizes from code).

test

eopws:
> So, now #33877 should fix also #34117, please test

Also it works only on resize. Please try to reload page while using dev-tools with mobile preview turned on and you are still gonna have the issue.

dgrammatiko:
> using dev-tools with mobile preview turned on

Try to use real mobile size, not the emulation...

eopws:
> Try to use real mobile size, not the emulation...

I'm currently not able to. Hope it works.
Probably it is a dev-tools's bad =)

PR #34159, opened 2021-05-24
[4] Deprecate MCrypt Class for removal in 5.0.0

PhilETaylor:
A fork of https://github.com/joomla/joomla-cms/pull/34147 for @wilsonge to merge without fear of breaking anything else. 

Code review

### DIFF
--- a/libraries/src/Encrypt/AES/Mcrypt.php
+++ b/libraries/src/Encrypt/AES/Mcrypt.php
@@ -16,6 +16,8 @@ use Joomla\CMS\Encrypt\Randval;
  * Mcrypt implementation
  *
  * @since    4.0.0
+ *
+ * @deprecated 4.0.0 will be removed in 5.0.0
  */
 class Mcrypt extends AbstractAES implements AesInterface
 {


libraries/src/Encrypt/AES/Mcrypt.php
```diff
@@ -16,6 +16,8 @@
  * Mcrypt implementation
  *
  * @since    4.0.0
+ *
+ * @deprecated 4.0.0 will be removed in 5.0.0
```
richard67, line 20: @wilsonge Shouldn't that be `* @deprecated  5.0 Will be removed without replacement` like we have it at other places?

PhilETaylor: An argument already had with @brianteeman who pointed to the crappy code standards in Joomla that state that what you say is wrong :-(

PhilETaylor: Section 6.2 of https://developer.joomla.org/development-strategy.html

richard67: I know that, but it doesn't clearly say how the deprecated tag has to look like and which version it shall contain, the one when it's deprecated or the one for which it is deprecated so it will be replaced or removed, or I am too silly to understand it.

PhilETaylor: No you are not silly, its just a very very badly written document, just like most of the development strategy and the coding standards... thats why I spell out exactly when the code will be removed.

PhilETaylor: see comments on https://github.com/joomla/joomla-cms/pull/34075 which is an example where CMSObject says `@deprecated 4.0.0` but "apparently" that means I cannot remove it in 4.0.0 I have to wait until 5.0.0

wilsonge:
Thanks!

PR #34160, opened 2021-05-24
[4] Document broken code and schedule it for removal in 5.0.0

PhilETaylor:
Forked from the comments on #34147 for @wilsonge to merge without fear of breaking anything else.

Code review.


wilsonge:
Thanks. I mean your fix is almost certainly right - just not something I want to merge without decently extensive testing so close to RC

Issue #34162, opened 2021-05-24
[4] Use Joomla\Input\Cli instead, despite that class not existing


PhilETaylor:
On Code review the class `\Joomla\CMS\Filter\InputFilter\Cli` has been deprecated, and every property and method has been marked as 

`* @deprecated  5.0  Use Joomla\Input\Cli instead`

The problem is that there is no such class as `Joomla\Input\Cli` which is very confusing for developers... 

https://github.com/joomla/joomla-cms/blob/f64e0186e41df3d68fa6b14d185cb4e60593cab9/libraries/src/Input/Cli.php#L19

brianteeman:
This has been resolved with #34306 and can be closed

PR #34163, opened 2021-05-24
[4] Remove non existing class import

PhilETaylor:
Code review

This class doesnt exist, and is not used in the file anyway as the only one that is, is `\Joomla\CMS\Input\Cli` which itself is deprecated with a note to use this non existing class `\Joomla\Input\Cli` /facepalm

Related to https://github.com/joomla/joomla-cms/issues/34162

### DIFF
--- a/libraries/src/Application/CliApplication.php
+++ b/libraries/src/Application/CliApplication.php
@@ -22,7 +22,6 @@ use Joomla\DI\ContainerAwareTrait;
 use Joomla\Event\DispatcherAwareInterface;
 use Joomla\Event\DispatcherAwareTrait;
 use Joomla\Event\DispatcherInterface;
-use Joomla\Input\Cli;
 use Joomla\Input\Input;
 use Joomla\Registry\Registry;
 use Joomla\Session\SessionInterface;


joomdonation:
I have tested this item :white_check_mark: successfully on 1abc6272f3501d13c34dc8c2fa83bdf1fa1de4e1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34163">issues.joomla.org/tracker/joomla-cms/34163</a>.</sub>

Quy:
I have tested this item :white_check_mark: successfully on 1abc6272f3501d13c34dc8c2fa83bdf1fa1de4e1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34163">issues.joomla.org/tracker/joomla-cms/34163</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34163">issues.joomla.org/tracker/joomla-cms/34163</a>.</sub>

rdeutz:
Maybe better to remove the full class path here https://github.com/joomla/joomla-cms/blob/1abc6272f3501d13c34dc8c2fa83bdf1fa1de4e1/libraries/src/Application/CliApplication.php#L129 instead of removeing the class import?

PhilETaylor:
I think you need more coffee. 

`use Joomla\Input\Cli;` is being removed by they PR because that class doesn't exist at all. 

`\Joomla\CMS\Input\Cli` is being used on line 129 COULD be imported, but then so could all the Session classes above it and I get told off by "maintainers" for importing classes because "that causes too much work with merging up" - and that would be out of scope of this already RTC PR. 

Removing a reference to a class that totally doesnt exist, is more important I would say then simplifying fully qualified class names. 

However you are the maintainer, and you have "Allow edits by maintainers" rights.. .so do as you wish to this already RTC PR.

rdeutz:
Good suggestion with the coffee

PR #34164, opened 2021-05-24
[4] Another set of fixes for Undefined Classes in docblocks

PhilETaylor:
Code review

All these fix Undefined Class issues in docblocks

Lumped them altogether to avoid being told off for too many PRs ...


wilsonge:
Thanks!

PR #34165, opened 2021-05-24
[4] Fix fatal error due to Undefined Class not imported 

PhilETaylor:
Code review

`use Joomla\CMS\Language\Text;` is missing which is used on line 61 when setting the error text message.


Quy:
Thanks!

PR #34167, opened 2021-05-24
[4] remove undefined call to parent::loadConfiguration that would cause fatal error

PhilETaylor:
Code review.

There is no such method called `loadConfiguration` in the parent class.

In fact there is no such method called `loadConfiguration` in the whole of Joomla including its installed vendors, apart from this method that being called. 

If this method were to be called ever, it would result in a fatal error


wilsonge:
This is just oversight from the cleanup from it originally extending the platform application daemonset i suspect.

wilsonge:
Thanks!

PR #34171, opened 2021-05-24
[4] Remove unused file

PhilETaylor:
### Summary of Changes

Deleted the file as per the instructions contained in the file:

```
// @TODO Remove when webcomponents get merged into the core
```

### Testing Instructions

Im guessing.. paging @dgrammatiko our resident Webcomponents expert... 

Install Joomla. Create some validation errors on installing Joomla on purpose.


brianteeman:
it is referencing a layout that doesnt exist 
`
		return LayoutHelper::render('installation.template.html.system.message', $displayData);`

PhilETaylor:
@brianteeman Further reason to delete this file then right?

dgrammatiko:
Probably the todo is wrong but this PR is looking fine

richard67:
It would have to be added to the list of files to be deleted in script.php .. but maybe it's better if we do that later with the tool which is in development before @wilsonge builds the RC.

PhilETaylor:
Did you miss the fact this is in the installation folder? Does that make a difference?

wilsonge:
Merging on review. Indeed as installation nothing will go into script.php :)

richard67:
Ahh good .. installation :-) sometimes we are lucky

PR #34172, opened 2021-05-24
[4] Condition is always 'true' because '!$menuElement' is already 'false' at this point

PhilETaylor:
Code review.

Diff makes this change look worse than it really is haha

### DIFF
--- a/libraries/src/Installer/Adapter/ComponentAdapter.php
+++ b/libraries/src/Installer/Adapter/ComponentAdapter.php
@@ -1097,63 +1097,42 @@ class ComponentAdapter extends InstallerAdapter
 		// Let's figure out what the menu item data should look like
 		$data = array();
 
-		if ($menuElement)
+		// I have a menu element, use this information
+		$data['menutype']     = 'main';
+		$data['client_id']    = 1;
+		$data['title']        = (string) trim($menuElement);
+		$data['alias']        = (string) $menuElement;
+		$data['type']         = 'component';
+		$data['published']    = 1;
+		$data['parent_id']    = 1;
+		$data['component_id'] = $componentId;
+		$data['img']          = ((string) $menuElement->attributes()->img) ?: 'class:component';
+		$data['home']         = 0;
+		$data['path']         = '';
+		$data['params']       = '';
+
+		if ($params = $menuElement->params)
 		{
-			// I have a menu element, use this information
-			$data['menutype']     = 'main';
-			$data['client_id']    = 1;
-			$data['title']        = (string) trim($menuElement);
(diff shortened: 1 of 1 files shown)


richard67:
I have tested this item :white_check_mark: successfully on 068d79d2ecc9b54d33110397c277665eeda549d0<br /><br />With a good merge tool like e.g. Beyond Compare the diff looks much better than on GitHub.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34172">issues.joomla.org/tracker/joomla-cms/34172</a>.</sub>

wilsonge:
you can just add ?w=1 on the end of the URL to ignore whitespace ;) 

So the only thing that kinda concerns me is when Nic rewrote this code (https://github.com/joomla/joomla-cms/pull/5499) he clearly seemed to think there was some specific use case here because he commented it all up. I mean it might well just be an oversight but is it possible to have an empty menu element tag that will be an empty XMLElement or something stupid and non-strict type checks will treat them differently...

PhilETaylor:
Before my modified code is 

```php
// Just do not create the menu if $menuElement not exist		
if (!$menuElement)		
{			
    return true;		
}
```

Therefore there is no way for the modified removed code to run... even phpStorm says so :)

richard67:
Right. If there were typesafe comparisons used, then a null could make a difference, but as it is now, I think the PR here is right.

wilsonge:
> Before my modified code is

I understand - but that line was added in that linked PR. That's the only reason I'm hesitating a bit. You're probably right. Just trying to figure out any weird edge cases in my head

joomdonation:
I have tested this item :white_check_mark: successfully on 068d79d2ecc9b54d33110397c277665eeda549d0<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34172">issues.joomla.org/tracker/joomla-cms/34172</a>.</sub>

joomdonation:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34172">issues.joomla.org/tracker/joomla-cms/34172</a>.</sub>

joomdonation:
@wilsonge Reviewed the code carefully. This PR is right and safe to merge. You can trust three of us :D.

wilsonge:
fiiine. i conceed :) thanks guys!

PR #34173, opened 2021-05-24
[4] Cleanup response from deleting /installation folder

PhilETaylor:
### Summary of Changes

Clean up a TODO left in the installation process on the task of deleting the /installation folder. 

Note that there is no point enqueuing a message, because when you destroy a session, you destroy that messageQueue ! 

### Testing Instructions

Install Joomla 4 taking notice of the ajax response when deleting the installation folder on the last step

A full HTML page is returned from the delete installation folder ajax

A simple JSON page is returned from the delete installation folder ajax

none

### DIFF
--- a/installation/src/Controller/InstallationController.php
+++ b/installation/src/Controller/InstallationController.php
@@ -256,24 +256,10 @@ class InstallationController extends JSONController
 
 		/** @var \Joomla\CMS\Installation\Model\CleanupModel $model */
 		$model = $this->getModel('Cleanup');
-		$success = $model->deleteInstallationFolder();
-
-		// If an error was encountered return an error.
-		if (!$success)
-		{
-			$this->app->enqueueMessage(Text::sprintf('INSTL_COMPLETE_ERROR_FOLDER_DELETE', 'installation'), 'warning');
-		}
+		$model->deleteInstallationFolder();
 
 		$this->app->getSession()->destroy();
 
-		$r = new \stdClass;
-		$r->view = 'remove';
-
-		/**
-		 * TODO: We can't send a response this way because our installation classes no longer
-		 *       exist. We probably need to hardcode a json response here
-		 *
-		 * $this->sendJsonResponse($r);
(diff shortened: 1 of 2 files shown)


installation/src/Controller/InstallationController.php
```diff
@@ -256,24 +256,10 @@ public function delete()
...
 		/** @var \Joomla\CMS\Installation\Model\CleanupModel $model */
 		$model = $this->getModel('Cleanup');
-		$success = $model->deleteInstallationFolder();
-
-		// If an error was encountered return an error.
-		if (!$success)
-		{
-			$this->app->enqueueMessage(Text::sprintf('INSTL_COMPLETE_ERROR_FOLDER_DELETE', 'installation'), 'warning');
-		}
+		$model->deleteInstallationFolder();
```
wilsonge: Can we return a hardcoded json response if this fails like we do below?

PhilETaylor: will need to see if the JS actually looks at the AJAX response enough to display the error... else it would be pointless. I doubt it does given that before this PR the ajax returned a whole HTML page response.

PhilETaylor: Done.

PhilETaylor:
Now if the folder fails to delete you get a nice JS message, and you can retry as many times as you like by pressing the red button. 

Screenshot 2021-05-25 at 13 58 34


And if it succeeds then you get the normal hiding of the red button and dev mode message

Screenshot 2021-05-25 at 14 01 47

wilsonge:
Looks great - trying to figure out if we should be checking the `error` property or if this is one of these things that's good enough for now given the rarity that this will fail in the first place.

PhilETaylor:
> Looks great - trying to figure out if we should be checking the error property or if this is one of these things that's good enough for now given the rarity that this will fail in the first place.

Added this now.

wilsonge:
Thanks!

PR #34216, opened 2021-05-25
[4.0] tinyMCE 5.8.1

brianteeman:
Version 5.8.1 - May 20, 2021
Fixed
An unexpected exception was thrown when switching to readonly mode and adjusting the editor width.
Content could be lost when the pagebreak_split_block setting was enabled.
The list-style-type: none; style on nested list items was incorrectly removed when clearing formatting.
URLs were not always detected when pasting over a selection. Patch contributed by jwcooper.
Properties on the OpenNotification event were incorrectly namespaced.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9168,9 +9168,9 @@
       "dev": true
     },
     "node_modules/tinymce": {
-      "version": "5.8.0",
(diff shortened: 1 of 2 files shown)


wilsonge:
Thanks!

brianteeman:
thanks

PR #17677, opened 2017-08-22
[4.0] Change the lookup order for the favicon.ico

Bakual:
Pull Request for Issue #17629 .

### Summary of Changes
This PR changes the lookup order for the favicon so it first looks in the root folder and afterwards in the template one.
Currently, the lookup works the other way (first template, then root).

I think the current way has historic reasons because (I think) we used to ship a favicon.ico in root in earlier versions. The favicon in the template folder then allowed to override that favicon.

Today we don't ship a favicon in root, but we ship one in the template folders, thus it imho makes more sense to reverse the override place.

### Testing Instructions
* Put different favicon.ico files into the root folder and the active template folder.
* Check which favicon is used

### Expected result
The one from root is used, overriding the one from the template

### Actual result
The one from the template is used, overriding the one from root.

### Documentation Changes Required
Don't know if that hehavior is documented somewhere. If yes, it should be adjusted.
Also it should be documented as a changed behaviour between 3.x and 4.0.

Issue #17629, opened 2017-08-20
user.ico so favicon.ico gets not overwritten


jjsjjs:
### Steps to reproduce the issue
Use the standard templates, update Joomla

### Expected result
your favicon is gone

### Actual result
The Joomla favicon is displayed instead of your own

### System information (as much as possible)
Every version of Joomla

### Additional comments
Is it possible to create a user.ico so the favicon is not getting overwritten every time there is an update?
This is happening when you make use of protostar or beez template. It's most annoying and easily forgotten to copy again and again and again. Thanks!

brianteeman:
No its not possible as favicon is an internet standard

The only option is to use a copy of the default template

jjsjjs:
You could leave it out of every update, and only put it in when a big update (read new Joomla version) is there. Updating the favicon every time is not bringing anything, but hassle for every Joomla user who uses the standard (and mostly adjusted) templates.

brianteeman:
Re-opening

That might be possible.

jjsjjs:
Would be great, thank you.

C-Lodder:
Then the same would go for all the LESS files too. If you want to customize the default Protostar template, you should copy it. Same applies if you want to change the favicon

jjsjjs:
Maybe that's true. I don't know if the LESS files are also altered as much as the template.css, which has support for user.css

mbabker:
The right way to do things if you are going to use the core templates is to duplicate them.  We can not (and should not) pick and choose which files are "important" for an update and which files are "important" to not override because a user might have made some changes.  So it is an all or nothing agreement.

brianteeman:
Closing

franz-wohlkoenig:
closed as stated above.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/17629">issues.joomla.org/tracker/joomla-cms/17629</a>.</sub>

jjsjjs:
ok fair enough. Thanks.

Bakual:
From memory, if you put a favicon into the root directory of your Joomla, the one from the template folder gets "overriden" and the one in root will be used.

mbabker:
Well, close.  If one doesn't exist in the template and one exists at the root, that will be used.

[...]

PR #34221, opened 2021-05-25
Switch favicon lookup to prefer the icon in Joomla root

mustaqahmed:
Change favicon lookup order to check Joomla root folder before the template folder.

### Summary of Changes
Currently by default the favicon is chosen from the default template folder.  If a site replaces the icon in the same folder, the new icon is gone after the template is updated.  On the other hand, if the site puts the custom icon in the Joomla root, the current lookup order doesn't pick the icon because the template folder is considered first.

This PR switches the favicon lookup order to pick the icon in the Joomla root first if one exists.  This addresses the Joomla forum comment [here](https://forum.joomla.org/viewtopic.php?t=937844#p3551480).

### Testing Instructions
If we have two different favicons in Joomla root folder and in the template folder, this change will pick the one in Joomla root.

The favicon in the template folder is shown on the published site.

The favicon in the Joomla root folder is shown on the published site.

Not sure if favicon lookup order is documented somewhere or not.

### DIFF
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -670,24 +670,17 @@ class HtmlDocument extends Document
 			ob_end_clean();
 		}
 
-		// Try to find a favicon by checking the root then the template folder
-		$icon = 'favicon.ico';
+		// Try to find a favicon by checking the root and then the template folder
+		$icon = '/favicon.ico';
 
-		if (file_exists($icon))
+		foreach (array(JPATH_BASE, $directory) as $dir)
 		{
-			$this->addFavicon($icon);
-		}
-		else
-		{
-			foreach (array($directory, JPATH_BASE) as $dir)
+			if (file_exists($dir . $icon))
 			{
-    			if (file_exists($dir . '/' . $icon))
-				{
-					$path = str_replace(JPATH_BASE, '', $dir);
-					$path = str_replace('\\', '/', $path);
-					$this->addFavicon(Uri::base(true) . $path . '/' . $icon);
-						break;
-				}
+				$path = str_replace(JPATH_BASE, '', $dir);
+				$path = str_replace('\\', '/', $path);
+				$this->addFavicon(Uri::base(true) . $path . $icon);
+				break;
 			}
 		}
 
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -670,17 +670,24 @@ class HtmlDocument extends Document
 			ob_end_clean();
 		}
 
-		// Try to find a favicon by checking the template and root folder
-		$icon = '/favicon.ico';
+		// Try to find a favicon by checking the root then the template folder
+		$icon = 'favicon.ico';
 
-		foreach (array($directory, JPATH_BASE) as $dir)
+		if (file_exists($icon))
 		{
-			if (file_exists($dir . $icon))
+			$this->addFavicon($icon);
+		}
+		else
+		{
+			foreach (array($directory, JPATH_BASE) as $dir)
 			{
-				$path = str_replace(JPATH_BASE, '', $dir);
-				$path = str_replace('\\', '/', $path);
-				$this->addFavicon(Uri::base(true) . $path . $icon);
-				break;
+    			if (file_exists($dir . '/' . $icon))
+				{
+					$path = str_replace(JPATH_BASE, '', $dir);
+					$path = str_replace('\\', '/', $path);
+					$this->addFavicon(Uri::base(true) . $path . '/' . $icon);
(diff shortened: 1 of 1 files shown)


brianteeman:
Doesnt it already do that? In my test it does

richard67:
@mustaqahmed Your pull request (PR) has several mistakes.

1. When you create a PR, the description already contains headings like "Summary of Changes", "Testing Instructions". They are not there for nothing. They are meant to be completed by the corresponding text.

You might find bad examples of PR's from other people where testing instructions are not provided. In some case this might be ok because it's simple changes only, but you should use them as example for your PR's. You should always provide a description of changes and testing instructions.

2. Your PR has code style errors. You can see that by going down to the bottom of the PR at GitHub and check the section "Some checks were not successful". Here now it shows that a step called "continuous-integration/drone/pr" has failed.

If you follow the "Details" link right beside that failed step, you will come to a page with a list of checks on the left hand side, failed ones in red and passed ones in green colour.

Here for this PR the "phpcs" step fails, which means "PHP code style". If you select that step in the list on the left hand side, the right part of the windows shows the log file with the details. You might have to scroll up a bit to see that. Here the link to this log: https://ci.joomla.org/joomla/joomla-cms/44426/1/3 .

It shows that you have used spaces instead ot tabulators to indent code.

You can also see that by the wrong indentation shown when you check the changes on GitHub: https://github.com/joomla/joomla-cms/pull/34221/files#diff-1f3d4b0ff411d275d94c1538ef487ef75d4450d772ad2cf712b5796467917f6dR684 .

These code style errors need to be fixed if this PR shall be accepted.

3. As @brianteeman already pointed out, there is no problem with the favicon in the root folder being not found, so please check if your PR really fixes any issue.

We require that a pull request has been tested by the author before it is submitted.

This brings us back to the 1st point, providing testing instructions.

When testing yourself it helps you to wrote the instructions.

mustaqahmed:
@brianteeman @richard67, please take a look again.  My first commit was done in a hurry only to understand why the simple fix was missed.  I have now updated the PR to properly fix the lookup order and also fill the description fields.

sakiss:
IMHO the ability of a template to override a globally set favicon is welcome and does not need to change. Usually different templates provide new styling that also affects the site's branding.
Any anomaly caused by that (e.g. a template is shipped with a favicon), is caused by the implementation of the template developers and not by the feature itself.

brianteeman:
@sakiss you miss the point. Joomla does not provide a favicon in the root so there is no globally set favicon. A site owner should be able to set a favicon in the root to safely override any favicon in the template directory without it being replaced on a template update. The only question here is does that happen right now and if not how do we make it happen

sakiss:
@brianteeman Thanks for the info. I am aware of that. My post was about keeping the template's ability to override a globally set favicon (i.e. the opposite of the goal of that PR). If we want to be consistent with that approach, we may have to remove the favicon file from the templates and provide a global one.

> The only question here is does that happen right now and if not how do we make it happen

In my tests it works that way. The template's one overrides the global.

brianteeman:
what globally set favicon?? I dont see one

mustaqahmed:
I think "globally set" here means site owner creating the file `root/favicon.ico`.  IMHO this file should override the template's icon because removing the icon in the template folder _does not_ work today across updates.

Oh, BTW, I just spotted [this landed PR](https://github.com/joomla/joomla-cms/pull/17677) doing the exact same thing but to 4.0!  Not sure what it means for my PR.

richard67:
> Oh, BTW, I just spotted [this landed PR](https://github.com/joomla/joomla-cms/pull/17677) doing the exact same thing but to 4.0! Not sure what it means for my PR.

@mustaqahmed It might mean that is was decided that the issue has to be fixed, but that the fix is a backwards compatibility (b/c) break and so had to go into version 4.

The discussion in issue #17629 linked by that PR looks as if this was the case.

richard67:
@Bakual Do you remember why you had to rebase your PR #17677 to 4.0-.dev? Was it a b/c issue? In this case this PR here should be closed. Or was it not, and you think it could still be fixed in 3.x?

Bakual:
I did it from the beginning for 4.0 because it is a minor B/C issue. We should not change this in 3.x at this point.

richard67:
@Bakual Thanks for feedback and confirmation.

@mustaqahmed As you can see in the comment above, we can't change that in Joomla 3 due to B/C (backwards compatibility) reasons. So I think your PR should be closed. That can happen sometimes, also to experienced contributors. I hope you are not too disappointed. I think this PR was at least a good training for making other PR's, and I'm looking forward to more contributions from you, and thank for your this one.

mustaqahmed:
That's all fair.  Thanks for the engaging discussion on my trial PR.

Issue #34228, opened 2021-05-26
[4.0] HTMLHelper::_('script' ...) B/C issue?


Ruud68:
### Steps to reproduce the issue
There are two JS files in the extension:
- (full) abc/123/file1.js
- (minified) abc/123/file1.min.js

load the (full) js file via HTMLHelper (note: relative = false because the file is NOT located in the media folder)

```
HTMLHelper::_('script', 'abc/123/file1.js', ['relative' => false]);
```

### Expected result

The file abc/123/file1.js should be loaded in the head of the page

### Actual result

Joomla 3: correct! abc/123/file1.js is loaded
Joomla 4: INCORRECT > the minified file is loaded abc/123/file1.min.js

### System information (as much as possible)
J3.9.27 and J4.0.0-beta8-dev

### Additional comments
J4 forces the loading of minified files when they are found based on the full filename. This behavior can only be toggled via the global JDEBUG setting, this way it is for extension developers NOT possible to load the full version based on an extension setting.

Loading the file via J4 webassetmanager has the same behavior as the uses HTMLHelper::_('script', ...) to to get the file to load.

richard67:
My personal opinion is that this issue should be closed as expected behaviour.

@dgrammatiko What do you think?

dgrammatiko:
It is expected behaviour. If a minified file exists (and not in debug mode) that one will be served.

Also please DON'T serve static assets outside of the media folder!!!

richard67:
Closing as expected behaviour.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34228">issues.joomla.org/tracker/joomla-cms/34228</a>.</sub>

Ruud68:
> It is expected behaviour. If a minified file exists (and not in debug mode) that one will be served.
> 
> Also please DON'T serve static assets outside of the media folder!!!

file and locations where only there for showing the issue :0 not a path I would use myself :)

> Closing as expected behaviour.

although 'expected', still different as to what the same function does in J3.
In my experience there is a difference in expected behavior and changed behavior :)
Not saying that the issue should be resolved, but at least this should be described as changed behavior when going to J4.

If it wasn't an issue I would have not ran into it when migrating my extensions, not being able to pick which file (full / minified) other then with the all or nothing approach (JDEBUG) is breaking functionality for me :S

so please reopen :)

[...]

PR #34232, opened 2021-05-26
[4.0] Respect detectDebug

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/34228 .

### Summary of Changes
Fixes the ordering 

### Testing Instructions
Check that `HTMLHelper::_('script', 'media/system/js/showon.js', ['detectDebug' => false]);` respects the rule and the non minified is served

Also check that `HTMLHelper::_('script', 'media/system/js/showon.js');` loads the minified 
Enable the DEBUG and check that the non minified is served

### DIFF
--- a/libraries/src/HTML/HTMLHelper.php
+++ b/libraries/src/HTML/HTMLHelper.php
@@ -1280,20 +1280,20 @@ abstract class HTMLHelper
 
 			if ($debugMode)
 			{
-				return self::checkFileOrder($minifiedPath, $nonMinifiedPath);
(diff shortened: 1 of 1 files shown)


Ruud68:
This doesn't work:

```
HTMLHelper::_('script', 'media/system/js/showon.js');
```
loads showon.js when JDEBUG is false and showon.min.js when JDEBUG is true.

The way it was was correct, this patch 'inverts' the file selection.

dgrammatiko:
Closing as it's not really working (?)

PR #34130, opened 2021-05-23
[4] Remove method from exception messages

PhilETaylor:
### Summary of Changes

Continuing the slow progress of removing PHP Internal Class and Method names from user visible error messages. 

Including class/method names and namespaces is "probably" ok for framework packages aimed at developers - but class/method names should never be shown to end users in error messages. (E.g When was the last time an Amazon shopping error page told you which class and method the error was in? it never does!) 

### Testing Instructions

mainly code review. 

https://example.com/administrator/index.php?option=com_media&format=json&task=api.files&path=local-images:/sampledata/../banner_cradle.jpg&url=asddfsdfg

### Actual result BEFORE applying this Pull Request

Screenshot 2021-05-23 at 14 43 31

### Expected result AFTER applying this Pull Request

Screenshot 2021-05-23 at 14 48 02

PR #34233, opened 2021-05-26
[4.0] Make sending attachments with the Mail Templates component more secure

richard67:
Pull Request for Issue #34185 .

### Summary of Changes

This pull request (PR) adds validation to the "Attachments Folder" option of the Mail Templates (com_mails) component so it falls under the same restrictions like the "Path to Images Folder" option of the Media Manager, i.e.
- The attachments folder has to be below the Joomla root folder.
- Certain system folders like "administrator", "api", "cache", and so on are not allowed.

In addition, it adds checks for the parameter from database in case if a path to the "Attachments Folder" has been saved in database which is not valid, which can happen before the patch of this PR is applied, or after that if someone hacks your database. These checks are added at two places:
- When loading the form to edit mail templates, the fields for adding attachments are not included in the form if the "Attachments Folder" is not valid or equal to the Joomla root.
- Sending email will be rejected with an exception if the "Attachments Folder" is not valid or equal to the Joomla root or if the full path to the attachment file is not valid or the file doesn't exist.

Finally, it makes sure that the file name used for the email attachment (Field "File Name" of an attachment in the mail template form), which can be different to the real name of the file on the server, doesn't use a different file name extension than the real file, by stripping off any extension from the attachment name and appending the extension of the real file.

What this PR does not fix:
- You can save the options form with "Attachments Folder" which doesn't exist.
- You can enter a "File Name" for an attachment which is not a valid file name on any client OS and so might make problems when the email receiver wants to save the attachment with that suggested name.

### Testing Instructions

The test starts on an installation of current 4.0-dev branch or latest 4.0 nightly build _without_ the patch of this PR applied which has mailing configured so that it works.

All actions are performed in backend or your local email client.

1. Go to "System -> Templates -> Mail Templates" and edit the options.

2. In field Attachments Folder, enter a valid folder below the Joomla root, e.g. "images", and save and close.

3. Make sure that the previously entered folder contains some files which can be attached to email, e.g some images.

4. Edit one of the mail templates, e.g. "Global Configuration: Test Mail".

5. At the bottom in section "Attachments", use the "File" field to select a file which has a file name extension, e.g. ".jpg" for an image.

6. In field "Filename" enter a file name using a different extension. E.g. if the attachment has extension ".jpg", use ".txt".

7. Save the changes.

8. Cause sending an email to an address of which you can receive email and read it in an email client.

9. When having received the email, open it in your email client and check the name of the attachment.
Result: The attachment has the name and extension as entered in step 3 in field "Filename".
This can be used to send files like "virus.exe" as an attachment "nice-image.jpg".
It can also cause problems when saving the file locally and then trying to open it on an OS (Windows) which uses the extension to determine the file type and to decide what to do with it, i.e. which program to use to open that file.

10. Go back to "System -> Templates -> Mail Templates" and edit again the component's options.

11. In field "Attachments Folder, enter a relative path to a folder which exists and is accessible for the webserver and PHP but outside the Joomla root, using "../" to break out the Joomla root, like it is described in issue #34185.

12. Save the changes.
Result: You can save the options.

13. Edit a mail template, e.g. "Global Configuration: Test Mail", and check if you can chose attachments from the folder specified in step 11.
Result: You can chose attachments from that folder.

14. Select an attachment and save the changes.

15. Cause a mail to be sent with that mail template.
Result: The mail is sent with the attachment specified in step 14.

16. Apply the patch of this PR.

17. Cause again a mail to be sent with that mail template.
Result: The mail is not sent. You see an error alert.

18. Edit again the mail template chosen in step 13 and check if you can chose attachments.
Result: The fields for choosing attachments are not shown.

19. Go back to "System -> Templates -> Mail Templates" , open again the component's options and try to save.
Result: You can't save the options due to invalid "Attachments Folder".

20. Try to use one of the system folders which are in the exclude list here: https://github.com/joomla/joomla-cms/pull/34233/files#diff-2792d8a5dc948a15bef321b2d150882221ceeb12488953ed502422a601235bd6R48
Result: You can't save the options due to invalid "Attachments Folder".

21. Change the "Path to Images Folder" to a valid path inside the Joomla root which contains files, and save the options.
Result: You can save the options.

22. Edit one of the mail templates, e.g. "Global Configuration: Test Mail".

23. At the bottom in section "Attachments", use the "File" field to select a file which has a file name extension, e.g. ".jpg" for an image.

24. In field "Filename" enter a file name using a different extension. E.g. if the attachment has extension ".jpg", use ".txt".

25. Save the changes.

26. Cause sending an email to an address of which you can receive email and read it in an email client.

27. When having received the email, open it in your email client and check the name of the attachment.
Result: The attachment has the name as entered in step 24 in field "Filename" but without the extension plus the extension of the real file appended. E.g. if the real file on the server was "joomla_black.png" and you have entered "blabla.txt" in field "Filename", the attachment will be named "blabla.png".

28. Edit again one of the mail templates, e.g. "Global Configuration: Test Mail".

29. At the bottom in section "Attachments", use the "File" field to select a file which has a file name extension, e.g. ".jpg" for an image.

30. In field "Filename" enter a file name using no extension.

31. Save the changes.

32. Cause sending an email to an address of which you can receive email and read it in an email client.

33. When having received the email, open it in your email client and check the name of the attachment.
Result: The attachment has the name as entered in step 30 in field "Filename" plus the extension of the real file appended. E.g. if the real file on the server was "joomla_black.png" and you have entered "blabla" in field "Filename", the attachment will be named "blabla.png".

Any folder can be specified as attachment folder in mail template options. This includes the possibility to break outside the Joomla root.

Files can be attached to email with a different file name extension than the attached file has on the server, e.g. you can attach a file "virus.exe" with name "funny-image.jpg".

The attachment folder parameter for mail templates is validated in the same way as it is for the "images" folder in the media manager options, i.e. it has to be in the Joomla root or below, and it excludes system folders like "administrator", "api", "cache", and so on.

If for some reason an invalid attachment folder parameter has been saved in database, e.g. on a previous 4.0 Beta version or a current 4.0-dev or nightly build which doesn't include the patch from this PR someone has used such a bad path, or someone hacked it in your database, then it is checked in the mail template's templates model and cleared if invalid.

Finally the path is checked when attaching files to mails on sending. If there for some reason the path is not valid, an exception is raised, so sending the mail is aborted.

When files are attached to mail, the file name extension of the real file on server side is appended to the (possibly different) file name of the attachment, if necessary, so it's not possible anymore that the attachment has a different extension than the file being attached.

Possibly there is already documentation for the mail templates, and it tells that you can use any folder, also outside the Joomla root. If this is the case: It needs to be updated.

### DIFF
--- a/administrator/components/com_mails/config.xml
+++ b/administrator/components/com_mails/config.xml
@@ -45,7 +45,7 @@
 			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
 			size="40"
 			validate="filePath"
-			exclude="administrator|api|bin|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
+			exclude="administrator|api|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
 		/>
 
 	</fieldset>
--- a/administrator/components/com_mails/config.xml
+++ b/administrator/components/com_mails/config.xml
@@ -44,6 +44,8 @@
 			label="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_LABEL"
 			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
 			size="40"
+			validate="filePath"
+			exclude="administrator|api|bin|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
 		/>
 
 	</fieldset>
--- a/administrator/components/com_mails/config.xml
+++ b/administrator/components/com_mails/config.xml
@@ -43,6 +43,7 @@
 			type="text"
 			label="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_LABEL"
 			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
+			filter="path"
 			size="40"
 			validate="filePath"
 			exclude="administrator|api|bin|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
--- a/administrator/components/com_mails/config.xml
+++ b/administrator/components/com_mails/config.xml
@@ -43,7 +43,6 @@
 			type="text"
 			label="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_LABEL"
 			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
-			filter="path"
 			size="40"
 			validate="filePath"
 			exclude="administrator|api|bin|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
--- a/administrator/components/com_mails/src/Model/TemplateModel.php
+++ b/administrator/components/com_mails/src/Model/TemplateModel.php
@@ -13,6 +13,7 @@ namespace Joomla\Component\Mails\Administrator\Model;
 
 use Joomla\CMS\Component\ComponentHelper;
 use Joomla\CMS\Factory;
+use Joomla\CMS\Filesystem\Path;
 use Joomla\CMS\Form\Form;
 use Joomla\CMS\Language\LanguageHelper;
 use Joomla\CMS\Language\Text;
@@ -113,7 +114,16 @@ class TemplateModel extends AdminModel
 			$form->removeField('copyto', 'params');
 		}
 
-		if (!$params->get('attachment_folder') || !is_dir(JPATH_ROOT . '/' . $params->get('attachment_folder')))
+		try
+		{
+			$attachmentPath = Path::check(JPATH_ROOT . '/' . $params->get('attachment_folder'));
+		}
+		catch (\Exception $e)
+		{
+			$attachmentPath = '';
+		}
+
+		if (!$attachmentPath || !is_dir($attachmentPath))
 		{
 			$form->removeField('attachments');
 		}
@@ -122,7 +132,7 @@ class TemplateModel extends AdminModel
 			$field = $form->getField('attachments');
 			$subform = new \SimpleXMLElement($field->formsource);
 			$files = $subform->xpath('field[@name="file"]');
-			$files[0]->addAttribute('directory', JPATH_ROOT . '/' . $params->get('attachment_folder'));
+			$files[0]->addAttribute('directory', $attachmentPath);
 			$form->load('<form><field name="attachments" type="subform" '
 				. 'label="COM_MAILS_FIELD_ATTACHMENTS_LABEL" multiple="true" '
 				. 'layout="joomla.form.field.subform.repeatable-table">'
--- a/administrator/components/com_mails/src/Model/TemplateModel.php
+++ b/administrator/components/com_mails/src/Model/TemplateModel.php
@@ -130,7 +130,7 @@ class TemplateModel extends AdminModel
 			$attachmentPath = '';
 		}
 
-		if (!$attachmentPath || $attachmentPath === JPATH_ROOT || !is_dir($attachmentPath))
+		if (!$attachmentPath || $attachmentPath === Path::clean(JPATH_ROOT) || !is_dir($attachmentPath))
 		{
 			$form->removeField('attachments');
 
--- a/administrator/components/com_mails/src/Model/TemplateModel.php
+++ b/administrator/components/com_mails/src/Model/TemplateModel.php
@@ -114,7 +114,7 @@ class TemplateModel extends AdminModel
 			$form->removeField('copyto', 'params');
 		}
 
-		if (!$params->get('attachment_folder'))
+		if (!trim($params->get('attachment_folder')))
 		{
 			$form->removeField('attachments');
 
--- a/administrator/components/com_mails/src/Model/TemplateModel.php
+++ b/administrator/components/com_mails/src/Model/TemplateModel.php
@@ -123,7 +123,7 @@ class TemplateModel extends AdminModel
 
 		try
 		{
-			$attachmentPath = rtrim(Path::check(JPATH_ROOT . '/' . $params->get('attachment_folder')), '/');
+			$attachmentPath = rtrim(Path::check(JPATH_ROOT . '/' . $params->get('attachment_folder')), \DIRECTORY_SEPARATOR);
 		}
 		catch (\Exception $e)
 		{
--- a/administrator/components/com_mails/src/Model/TemplateModel.php
+++ b/administrator/components/com_mails/src/Model/TemplateModel.php
@@ -114,32 +114,39 @@ class TemplateModel extends AdminModel
 			$form->removeField('copyto', 'params');
 		}
 
+		if (!$params->get('attachment_folder'))
+		{
+			$form->removeField('attachments');
+
+			return $form;
+		}
+
 		try
 		{
-			$attachmentPath = Path::check(JPATH_ROOT . '/' . $params->get('attachment_folder'));
+			$attachmentPath = rtrim(Path::check(JPATH_ROOT . '/' . $params->get('attachment_folder')), '/');
 		}
 		catch (\Exception $e)
 		{
 			$attachmentPath = '';
 		}
 
-		if (!$attachmentPath || !is_dir($attachmentPath))
+		if (!$attachmentPath || $attachmentPath === JPATH_ROOT || !is_dir($attachmentPath))
 		{
 			$form->removeField('attachments');
+
+			return $form;
 		}
-		else
-		{
-			$field = $form->getField('attachments');
-			$subform = new \SimpleXMLElement($field->formsource);
-			$files = $subform->xpath('field[@name="file"]');
-			$files[0]->addAttribute('directory', $attachmentPath);
-			$form->load('<form><field name="attachments" type="subform" '
-				. 'label="COM_MAILS_FIELD_ATTACHMENTS_LABEL" multiple="true" '
-				. 'layout="joomla.form.field.subform.repeatable-table">'
-				. str_replace('<?xml version="1.0"?>', '', $subform->asXML())
-				. '</field></form>'
-			);
-		}
+
+		$field = $form->getField('attachments');
+		$subform = new \SimpleXMLElement($field->formsource);
+		$files = $subform->xpath('field[@name="file"]');
+		$files[0]->addAttribute('directory', $attachmentPath);
+		$form->load('<form><field name="attachments" type="subform" '
+			. 'label="COM_MAILS_FIELD_ATTACHMENTS_LABEL" multiple="true" '
+			. 'layout="joomla.form.field.subform.repeatable-table">'
+			. str_replace('<?xml version="1.0"?>', '', $subform->asXML())
+			. '</field></form>'
+		);
 
 		return $form;
 	}
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -477,7 +477,8 @@ class MailTemplate
 	}
 
 	/**
-	 * Check and if necessary fix the file name of an attachment so that the attached file has the same extension as the source file, and not a different file extension
+	 * Check and if necessary fix the file name of an attachment so that the attached file
+	 * has the same extension as the source file, and not a different file extension
 	 *
 	 * @param   string  $file  Path to the file to be attached
 	 * @param   string  $name  The file name to be used for the attachment
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -287,7 +287,7 @@ class MailTemplate
 
 		$folderPath = rtrim(Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder')), '/');
 
-		if ($folderPath && $folderPath !== JPATH_ROOT && is_dir($folderPath))
+		if ($folderPath && $folderPath !== Path::clean(JPATH_ROOT) && is_dir($folderPath))
 		{
 			$attachments = (array) json_decode($mail->attachments);
 		}
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -285,7 +285,7 @@ class MailTemplate
 			$this->mailer->addReplyTo($this->replyto->mail, $this->replyto->name);
 		}
 
-		$folderPath = rtrim(Path::check($config->get('attachment_folder')), '/');
+		$folderPath = rtrim(Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder')), '/');
 
 		if ($folderPath && $folderPath !== JPATH_ROOT && is_dir($folderPath))
 		{
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -12,6 +12,7 @@ namespace Joomla\CMS\Mail;
 
 use Joomla\CMS\Component\ComponentHelper;
 use Joomla\CMS\Factory;
+use Joomla\CMS\Filesystem\File;
 use Joomla\CMS\Filesystem\Path;
 use Joomla\CMS\Language\Text;
 use Joomla\CMS\Mail\Exception\MailDisabledException;
@@ -297,7 +298,7 @@ class MailTemplate
 
 					if (is_file($filePath))
 					{
-						$this->mailer->addAttachment($filePath, $this->setAttachmentName($filePath, $attachment->name ?? $attachment->file));
+						$this->mailer->addAttachment($filePath, $this->getAttachmentName($filePath, $attachment->name));
 					}
 				}
 			}
@@ -307,7 +308,7 @@ class MailTemplate
 		{
 			if (is_file($attachment->file))
 			{
-				$this->mailer->addAttachment($attachment->file, $this->setAttachmentName($attachment->file, $attachment->name));
+				$this->mailer->addAttachment($attachment->file, $this->getAttachmentName($attachment->file, $attachment->name));
 			}
 			else
 			{
@@ -485,29 +486,17 @@ class MailTemplate
 	 *
 	 * @since   4.0.0
 	 */
-	protected function setAttachmentName($file, $name)
+	protected function getAttachmentName($file, $name)
 	{
-		// If no name is given use the basename of the file to be attached
+		// If no name is given, do not process it further
 		if (!trim($name))
 		{
-			return basename($file);
+			return '';
 		}
 
-		// Get file extension of the file to be attached
-		$extensionFile = pathinfo($file, PATHINFO_EXTENSION);
+		$ext = File::getExt($file);
 
-		// If the file has no extension there is nothing to do with the name
-		if (!$extensionFile)
-		{
-			return $name;
-		}
-
-		// Make sure to attach the file with the same extension as the attached file has (case-insensitive)
-		if (strtolower(pathinfo($name, PATHINFO_EXTENSION)) !== strtolower($extensionFile))
-		{
-			return $name . '.' . $extensionFile;
-		}
-
-		return $name;
+		// Strip off extension from $name and append extension of $file, if any
+		return File::stripExt($name) . ($ext ? '.' . $ext : '');;
 	}
 }
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -285,25 +285,21 @@ class MailTemplate
 			$this->mailer->addReplyTo($this->replyto->mail, $this->replyto->name);
 		}
 
-		$attachments = array();
-
-		if ($config->get('attachment_folder'))
+		if (trim($config->get('attachment_folder')))
 		{
 			$folderPath = rtrim(Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder')), \DIRECTORY_SEPARATOR);
 
 			if ($folderPath && $folderPath !== Path::clean(JPATH_ROOT) && is_dir($folderPath))
 			{
-				$attachments = (array) json_decode($mail->attachments);
-			}
-		}
+				foreach ((array) json_decode($mail->attachments) as $attachment)
+				{
+					$filePath = Path::check($folderPath . '/' . $attachment->file);
 
-		foreach ($attachments as $attachment)
-		{
-			$filePath = Path::check($folderPath . '/' . $attachment->file);
-
-			if (is_file($filePath))
-			{
-				$this->mailer->addAttachment($filePath, $this->setAttachmentName($filePath, $attachment->name ?? $attachment->file));
+					if (is_file($filePath))
+					{
+						$this->mailer->addAttachment($filePath, $this->setAttachmentName($filePath, $attachment->name ?? $attachment->file));
+					}
+				}
 			}
 		}
 
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -285,7 +285,7 @@ class MailTemplate
 			$this->mailer->addReplyTo($this->replyto->mail, $this->replyto->name);
 		}
 
-		$attachments = array()
+		$attachments = array();
 
 		if ($config->get('attachment_folder'))
 		{
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -285,13 +285,24 @@ class MailTemplate
 			$this->mailer->addReplyTo($this->replyto->mail, $this->replyto->name);
 		}
 
-		$path = Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder') . '/');
+		$folderPath = rtrim(Path::check($config->get('attachment_folder')), '/');
 
-		foreach ((array) json_decode($mail->attachments)  as $attachment)
+		if ($folderPath && $folderPath !== JPATH_ROOT && is_dir($folderPath))
 		{
-			if (is_file($path . $attachment->file))
+			$attachments = (array) json_decode($mail->attachments);
+		}
+		else
+		{
+			$attachments = array();
+		}
+
+		foreach ($attachments as $attachment)
+		{
+			$filePath = Path::check($folderPath . '/' . $attachment->file);
+
+			if (is_file($filePath))
 			{
-				$this->mailer->addAttachment($path . $attachment->file, $attachment->name ?? $attachment->file);
+				$this->mailer->addAttachment($filePath, $attachment->name ?? $attachment->file);
 			}
 		}
 
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -507,7 +507,7 @@ class MailTemplate
 		}
 
 		// Make sure to attach the file with the same extension as the attached file has (case-insensitive)
-		if (strtolower(pathinfo($name, PATHINFO_EXTENSION)) !== $extensionFile)
+		if (strtolower(pathinfo($name, PATHINFO_EXTENSION)) !== strtolower($extensionFile))
 		{
 			return $name . '.' . $extensionFile;
 		}
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -303,7 +303,7 @@ class MailTemplate
 
 			if (is_file($filePath))
 			{
-				$this->mailer->addAttachment($filePath, $attachment->name ?? $attachment->file);
+				$this->mailer->addAttachment($filePath, $this->setAttachmentName($filePath, $attachment->name ?? $attachment->file));
 			}
 		}
 
@@ -311,7 +311,7 @@ class MailTemplate
 		{
 			if (is_file($attachment->file))
 			{
-				$this->mailer->addAttachment($attachment->file, $attachment->name);
+				$this->mailer->addAttachment($attachment->file, $this->setAttachmentName($attachment->file, $attachment->name));
 			}
 			else
 			{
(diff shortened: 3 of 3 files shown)


brianteeman:
I have tested this item :white_check_mark: successfully on ef76d64f637c3c08f4a40ee0dba5d94fd4dfbb1d<br /><br />I am marking this as successful test as it does what it says but its not a very friendly experience as there is no clue what folders are allowed and which are not.

Please consider changing the field to type folderlist<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

richard67:
@brianteeman I agree a folder list would be more comfortable. Right now it is the same here as it is in the media manager options, so it's consistent with that. I think if we change it, we should change it also there.

brianteeman:
agree - thats why I marked it a successful test and that can be changed in both places in another pr

dgrammatiko:
I have tested this item :white_check_mark: successfully on ef76d64f637c3c08f4a40ee0dba5d94fd4dfbb1d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

PhilETaylor:
> chmst removed PR-4.0-dev  Release Blocker  labels 6 minutes ago

This is still a release blocker until tested and merged

richard67:
@PhilETaylor The issue tracker bot has removed it when setting RTC. I‘ve added it back.

richard67:
@PhilETaylor What's not working?

richard67:
@PhilETaylor If you think your issue #34236 justifies to give this PR here a negative test without any further explanation then I think you are wrong because this PR still fixes the issue as described. If it doesn’t, then please report back your observations.

PhilETaylor:
patience... Im typing as fast as I can!

PhilETaylor:
(sorry, slow to type up the results below, I'm not well, but wanted to mark as unsuccessful test quickly to prevent merge).

Unfortunately this fix is not enough. 

1) Side issue of no consequence is here https://github.com/joomla/joomla-cms/issues/34236

2) The reason this PR is not good enough is that the validation when using the value from the database and sending the the actual email means that I can still send /etc/passwd file to myself.

This means if a hacker can SQL inject, he can attach any files files to any email. 

Using your PR code, I was able to still attach /etc/passwd to my emails.

My exact replication is this.

 - My JPATH_BASE is /application
 - Set the Attachments folder to be media (which any sane Joomla admin could do)
 - Edit the test email, drop down the files and you will see the index.html (from the media folder) - right click and inspect that field, set the value to be relative to your media folder (so in my case ../../etc/passwd) - this replicates a SQL injection (but also shows that you can do this by editing and manipulating the field in Joomla admin, which again you should not be able to do!) 
 - Go to global config and click test mail - look at the email.

Screenshot 2021-05-26 at 20 51 52

1) Need to validate the file names when saving the Mail Template, the value selected in the dropdown should be a valid file on the hard disk in the attachment folder (in my case media) - not just checking if it is a `is_file($path.$file)` (not sure if ANY CHECKS at all are actually done on that input when saving?

2) Need to REVALIDATE the filename after pulling it back from the database and when using it in MailTemplate.php just before attaching it to the email.  

At the moment you have this code, my notes on this code are comments below. This code is from /libraries/src/Mail/MailTemplate.php

```php

// In my case $path is now JPATH_BASE/media/
		$path = Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder') . '/');

		foreach ((array) json_decode($mail->attachments)  as $attachment)
		{

// need a check here to test if $attachment->file is a file IN THE FOLDER `attachment_folder`. It should be a single 
// filename with no path at all. `filename.extension` only, and should exist in the array of 
//    `scandir(JPATH_ROOT . '/' . $config->get('attachment_folder'))` - rather than using `is_file` to check if the file exists. 

// At the moment, here $path . $attachment->file = JPATH_BASE/media/../../etc/passwd
 // So this is TRUE - thats bad! 
			if (is_file($path . $attachment->file))
			{
// and then the file is attached, even though its not in the attachment_folder. $attachment->file = JPATH_BASE/media/../../etc/passwd
				$this->mailer->addAttachment($path . $attachment->file, $attachment->name ?? $attachment->file);
			}
		}
``

File: libraries/src/Mail/MailTemplate.php [hunk 1]
```diff
@@ -283,7 +285,7 @@ public function send()
 			$this->mailer->addReplyTo($this->replyto->mail, $this->replyto->name);
 		}
 
-		$path = JPATH_ROOT . '/' . $config->get('attachment_folder') . '/';
+		$path = Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder') . '/');
```
dgrammatiko:
```suggestion
		$path = Path::check(JPATH_ROOT . '/' . str_replace('../', '', $config->get('attachment_folder') . '/'));
```

@PhilETaylor is this ok?

PhilETaylor: In my exact case - probably - but remember Im not a hacker, Im sure a hacker has better skills than I do and can use more than `../` - it also a tacky way of doing it. The value should be validated into and validated out of the db before even being used in a file path.

PhilETaylor:
Also, if I dont provide an attachments_folder in options - the root JPATH_BASE folder is used... so... I ... can email my configuration.php :) :) 

Screenshot 2021-05-26 at 21 11 07

richard67:
@PhilETaylor As as I see you describe a new exploit now. The one you‘ve described in issue #34185 is solved here.

richard67:
If you specify an empty value you should not see any fields for attaching files in the Mail template form. Please test this and not just make assumptions.

PhilETaylor:
> If you specify an empty value you should not see any fields for attaching files in the Mail template form. Please test this and not just make assumptions.

Is my screenshot not enough proof for you that I tested it??

Would a video prove it to you?


https://user-images.githubusercontent.com/400092/119725150-822f9f00-be67-11eb-8a16-3547687f02cf.mp4

richard67:
> If you specify an empty value you should not see any fields for attaching files in the Mail template form. Please test this and not just make assumptions.

Hmm, no, that might be wrong. But I‘m too tired now.

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

PhilETaylor:
> @PhilETaylor As as I see you describe a new exploit now. The one you‘ve described in issue #34185 is solved here.

No - the exact same bug - just expounding on it.

The only thing that is fixed, is now you cannot see the files from the /etc/ folder in a select box. You can still manually type a value in and use it, but you cannot see the file names. 

Screenshot 2021-05-26 at 21 17 04

administrator/components/com_mails/src/Model/TemplateModel.php
```diff
@@ -113,7 +114,16 @@ public function getForm($data = array(), $loadData = true)
 			$form->removeField('copyto', 'params');
 		}
 
-		if (!$params->get('attachment_folder') || !is_dir(JPATH_ROOT . '/' . $params->get('attachment_folder')))
+		try
+		{
+			$attachmentPath = Path::check(JPATH_ROOT . '/' . $params->get('attachment_folder'));
```
PhilETaylor: This will not throw an exception if `attachment_folder` is blank - it will return JPATH_BASE, and this is why configuration.php and the root files can be seen in the dropdown.

administrator/components/com_mails/src/Model/TemplateModel.php
```diff
@@ -113,7 +114,16 @@ public function getForm($data = array(), $loadData = true)
...
+		}
+		catch (\Exception $e)
+		{
+			$attachmentPath = '';
+		}
+
+		if (!$attachmentPath || !is_dir($attachmentPath))
```
PhilETaylor: here you could add a check to see if `$attachmentPath === JPATH_BASE` and if thats true then remove the field.

libraries/src/Mail/MailTemplate.php
PhilETaylor: No this is not good enough, as its not the `attachment_folder` Im hacking anymore, Im hacking the filename in the dropdown and making it relative :) so no.

administrator/components/com_mails/src/Model/TemplateModel.php
richard67: JPATH_BASE? Not JPATH_ROOT?

brianteeman:
None of this is preventing me from entering a file as the attachment for a specific email that is in a forbidden location

administrator/components/com_mails/src/Model/TemplateModel.php
PhilETaylor: sorry I use them interchangeably when I should not. I mean the path to the site = JPATH_ROOT

libraries/src/Mail/MailTemplate.php
richard67: Yes and this is what I mean, it is a different exploit than the one I am fixing here because here I am fixing the attachment_folder.

PhilETaylor: Do i get 2 brownie points for this then then...If only Joomla had some kind of dedicated security team....

administrator/components/com_mails/src/Model/TemplateModel.php
richard67: Possibly it would have to be `$attachmentPath === JPATH_ROOT . '/' here` .. so maybe it would be easier just to check for `!$params->get('attachment_folder')`?

PhilETaylor: what if attachment_folder was two spaces? that would pass still right? and that would be bad.

PhilETaylor: Oh dear - even after a Path::check - if the attachment_folder is two spaces - Path Check validates it as ok and returns `/application/  /` :-( another bug to raise...

brianteeman:
As I repeatedly have pointed out the validation of data is missing in many places and quite often (as seen here) relies 100% on someone not manipulaating the form before submitting it. Instead of validating the actual data ssubmitted it assumes that if you submitted it then its ok.

richard67:
> Please consider changing the field to type folderlist

@brianteeman I've just checked that and it seems a folderlist is just a list of folders in a parent folder without the possibility to go deeper. Maybe you meant a media field, where we can navigate through a folder structure? If so, then it might be that this is limited to a base folder defined in media manager options. But with this I'm not sure.

brianteeman:
inspect the form and change the value of one of the possible files to a file and path that you should not have access to.
Save and then check the stored value for the attachment in the database

richard67:
> inspect the form and change the value of one of the possible files to a file and path that you should not have access to.
> Save and then check the stored value for the attachment in the database

@brianteeman I guess that doesn't refer to my comment before yours where I asked about the folderlist field and media field.

PhilETaylor:
> folderlist field and media field.

Are the submitted values of those fields actually validated on save? Else if they can be manipulated in the dom we are back to square one

richard67:
> > folderlist field and media field.
> 
> Are the submitted values of those fields actually validated on save? Else if they can be manipulated in the dom we are back to square one

@PhilETaylor I think with the media field I was wrong because it's used to select files and not folders, and the folderlist is not appropriate for the reason I stated above, so we forget about these now.

I will see if I can add validation for the other issues here but it's late now and I'm tired and gotta work tomorrow so it may take time. If someone is faster: Go for it.

PhilETaylor:
There is no rush. I'm in bed too and have a flight plan filed for 10am so need to sleep. Thanks for working on this, I'm hoping to have the weekend off this week haha

brianteeman:
> > folderlist field and media field.
> 
> Are the submitted values of those fields actually validated on save? Else if they can be manipulated in the dom we are back to square one

Not that I could see

richard67:
I just see folderlist can be recursive, so it would really be more comfortable here where we want to be limited to the Joomla root or below. Of course it still would need to do some validation in the model before saving and more. Will work on that tomorrow or later.

brianteeman:
May I make a suggestion. 
The ability to send an attachment is a new feature. Instead of taking up a lot of time in the last few days before RC trying to secure it just pull it out of 4.0.
Then it can be worked on at leisure for 4.1
No one will miss a feature that they never had and that has questionable benefits. Have you ever wanted to add an attachment to one of Joomla's automated emails? I haven't.

richard67:
I think I can fix it, but if I don‘t get it fixed until weekend we can wipe it out.

joomdonation:
@richard67 I made a PR to your branch https://github.com/richard67/joomla-cms/pull/17 for some more validations. Could you please review?

richard67:
@joomdonation Thanks for the PR, but I was already working on a slightly different way to solve it. It does the same but with a bit different code.

brianteeman:
Please let us know when it is ready for retesting

richard67:
> Please let us know when it is ready for retesting

Will do so.

Following has been done with the last changes:
- Input filtering has been added to the attachments folder field of the component's configuration form so that invalid characters or spaces only result in the form being saved with empty value for that field, like it is with other input filters, and this will then result in email attachments are disabled.
- The attachments folder check on loading the mail template form has been improved so that an empty or otherwise bad value is correctly detected and the fields for attachments are not added to the form. This also ensures that it's not possible to browse the Joomla root folder for adding attachments.
- The attachments folder check on sending email has been improved so that it's not possible to attach files from the Joomla root folder.
- For each attachment, the complete path is checked so that it's not possible anymore to send email with not existing attachments.

Following remains to be done:
- I think about adding a custom exception so we see a useful exception message when sending email is aborted due to invalid folder or file for attachments (which at this point could only come from a bad parameter in database from past or a hack).
- Testing instructions need to be adapted to changes.

richard67:
I have to test it on Windows, too, not only on Linux. Stay tuned.

richard67:
Ok, I'm ready with the folder and the file name on server side.

But as @brianteeman correctly observed, you can enter in the mail template form for an attachment anything you want in the text field labelled with "File Name". And this field is then used as the file name for the attachment in the email, so when you try to save that file after having received the email, you get proposed the weird and possibly (depending on the OS where the email client is used) invalid name.

So it needs some validation and maybe also input filtering for that field, too.

richard67:
... And that field should not allow to change the extension. Not sure if I can fix that here or if it is better to make a separate PR for that.

brianteeman:
@richard67 that wasn't quite what I meant but its a valid concern

I have checked what I can with this and it appears to be ok now but I am in no way an expert at this type of thing.

(still not convinced on the value of this feature)

richard67:
@brianteeman Currently you can with my PR now safely chose a file for attachment and then enter a completely weird file name with which it will be attached in the email. That means we can attach a file "virus.exe" from the server to the email with a name "naked-girl.jpg" and when the received wants to check the attachment it is "naked-girl.jpg" (or on Windows when hiding extensions of known file types just "naked-girl") 🤣

richard67:
@brianteeman P.S. What I describe in my previous comment is a very valuable feature for certain people 😈  I just don't think they should use Joomla for their business.

brianteeman:
in the context of the mail templates that we have I just dont see the use case

libraries/src/Mail/MailTemplate.php
```diff
@@ -283,21 +285,33 @@ public function send()
...
-			if (is_file($path . $attachment->file))
+			$folderPath = rtrim(Path::check(JPATH_ROOT . '/' . $config->get('attachment_folder')), \DIRECTORY_SEPARATOR);
+
+			if ($folderPath && $folderPath !== Path::clean(JPATH_ROOT) && is_dir($folderPath))
+			{
+				$attachments = (array) json_decode($mail->attachments);
+			}
+		}
+
+		foreach ($attachments as $attachment)
```
Quy: Move inside `if` statement?

richard67: No, not necessary because the attachments array will be empty.

richard67:
I think I have to revert the input filter because it disallows folder names with non ASCII characters, e.g. Greece or Russian letters.

libraries/src/Mail/MailTemplate.php
Quy: I was referring to moving it inside the above if statement.

richard67: yes and that's not right because the array will be empty anyway if the parameter was not set or not right

richard67:
@Quy I've refactored the code and followed your suggestion to move the loop inside the if even if I'm not a fan of too deeply nested if's, but it's a matter of taste.

@joomdonation You should be happy too because I've decided to use trim as you had suggested after I had to revert the input filtering.

Will update testing instructions a bit later and report back when it's ready for testing.

richard67:
@brianteeman @joomdonation @PhilETaylor @Quy (in alphabetical order): Ready for testing. Please test. Many test steps in the instructions, but less time consuming than it looks like.

brianteeman:
Is it the intended behaviour that if you select a file from the list called fred.jpg and enter a filename of mary.png the attachment is named as mary.png.jpg ?

brianteeman:
As stated before the user experience of setting the folder in the options is very frustrating as you dont know why the folder input was rejected

brianteeman:
otherwise I have checked what I can with this and it appears to be ok now but I am in no way an expert at this type of thing.

richard67:
> Is it the intended behaviour that if you select a file from the list called fred.jpg and enter a filename of mary.png the attachment is named as mary.png.jpg ?

@brianteeman Yes. As described in the description, this PR makes sure that you can't send the attachment with a different file name extension than the real file has (if it has one).

> As stated before the user experience of setting the folder in the options is very frustrating as you dont know why the folder input was rejected

I fully agree, but I can't fix this with this PR here.

richard67:
@brianteeman So was your test a successful test or not? Am asking because I see no result marked in the tracker.

brianteeman:
It is neither. I didn't find anything wrong but I don't feel qualified enough to say this is secure

richard67:
Another option than fixing this attachment folder option for mail templates would be to make it use the media field to select attachments and remove that attachments folder option from the mail component's options. But that would be a bigger change and need more time, so I just wanted to quickly fix the security hole.

PhilETaylor:
Sorry for my absence/delay. Some thoughts on the current state after this PR would be merged...



> Sending email will be rejected with an exception if the "Attachments Folder" is not valid or equal to the Joomla root or if the full path to the attachment file is not valid or the file doesn't exist.

> Finally the path is checked when attaching files to mails on sending. If there for some reason the path is not valid, an exception is raised, so sending the mail is aborted.


1) This exception is unhandled and comes from `Path::check` -  That might be ok... , but in Global Configuration it displays as a HTTP Error and not an exception message when sending the test mail. 

Screenshot 2021-05-28 at 17 11 46

It would be interesting to see other places in Joomla handling the failure to send a Mail Template because of this exception being thrown and not handled... Its not like Joomla to throw an unhandled exception. 

2) If your provided custom file name ends in a period, then you get customfilename..jpg as an extension. 

3) Out of scope of this PR: `validate="filePath"` doesnt validate correctly, but for example using ftp://example.com is a valid attachment_folder at the moment - added more examples to  #34236

4) I can still right click the dropdown list of files, and change the value of the dropdown, and click save, and its inserted into the db, despite my input being outside of the Joomla site. E.g attachments_folder = images, and I edit the files list and change joomla_black.png to ../../joomla_black.png and can save the Mail Template with no errors. This should not be allowed. 

5) After doing 4 above, the image is obviously bad as ../../joomla_black.png would take you out of the joomla file system, this causes Path::check to throw an exception "Joomla\CMS\Filesystem\Path::check() - Use of relative paths not permitted", which is not caught and handled (see 1), leading to a PHP fatal error (or on my set up, when sending the test email from Global Config, I get a "An error has occurred while fetching the JSON data: HTTP 502 status code. Bad Gateway" error message shown on the page) 

6) If, after resolving the folder, and resolving the filename, and putting them together and not finding that file on the hard disk - no feedback is provided to say the attachments were not valid, and not attached, and the email is sent without attachments, silently. Im not sure this is what "A Joomla Admin" would expect. If they have a mail that is configured to send an attachment, and it doesnt (because Admin B deleted the file!) then the mail sending should fail right? Sending the mail with no attachment, when one is confiugred, just seems wrong. Happy to be overruled on this. 

7. Are you sure the "media field to select attachments." would solve the issue,

brianteeman:
re 4
I had the same as you in that I could put the invalid data into the database BUT when the email was sent it was ignored and I didnt get an error message

PhilETaylor:
Summary: This PR fixes the security issue. 

However further refinement is needed generally, and Joomla should be doing these kinds of code reviews before features are merged to the core, and not 4 weeks after a RC is due. The JSST should be taking an active part in the development of the product and doing proactive code reviews (as it is in their role to do so) and not ignoring Joomla 4.

PhilETaylor:
I have tested this item :white_check_mark: successfully on 818407d0066f91f3abb46da9c45a5ff95e67db60<br /><br />Great work. Sometimes when given a pig, bacon sandwiches is all you can smell. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

richard67:
@wilsonge I'd like to ask you to check this PR and decide what to do with it, merge and further improve, or better change this email attachment feature of the mail component to not use an own attachment folder parameter and own file name field but the media field, or even remove this attachment feature.

Personally if we had enough time I'd tend to the latter, i.e. change or remove the feature,

richard67:
@brianteeman @PhilETaylor Just to be sure you get me right: I agree with all you find which is not good with this function, I just don't have the time and maybe also not the knowledge to fix them all. And yes, we should have reviewed and fixed it long time ago.

PhilETaylor:
This could be merged as it is. That would fix the release blocking security issue.

joomdonation:
I could not do a full code review right now. But I hope @wilsonge will do a code review before merging. From my quick look, I have few small complains:

- Why do you use `setAttachmentName` as method name? Maybe `getAttachmentName` would be better. set indicates that we set something, but the actual logic here is that we calculate the attachment name base on the real attachment file and the alternative name
- Why do you use `$attachment->name ?? $attachment->file` in the setAttachmentName call? I think we just need to pass `$attachment->name` only and in case it empty, just return '' and the the mailer would use the name from `$attachment->file`. That will save us from some extra unnecessary code in case $attachment->name empty (which I expect in most case)
- We have a library method File::getExt to get file extension. I'm unsure which one is better (compare with pathinfo) but normally, I would use the build in library method if available.

Overall, this PR is good and solve the security issue. I'm unsure if my mentioned issues above are right, but @wilsonge should review and make decision.

PhilETaylor:
> Why do you use setAttachmentName 

Quick easy change @richard67 

> Why do you use $attachment->name ?? $attachment->file

These are being passed to `setAttachmentName` you mentioned above, and not as you think to the `addAttachment` method call. I think you missed that.

richard67:
@brianteeman @joomdonation @PhilETaylor @Quy (in alphabetical order): Ready for testing again.

I've implemented @joomdonation 's suggestion to rename `setAttachmentName` to `getAttachmentName` and have simplified the check for the attachment name (field "Filename") and the file name extension handling. Tesing instructions have been adapted and simplified a bit.

Meanwhile I am working on another PR to implement validation for file and folder paths including the check if the file or folder exists so it would not be possible anymore to enter a not existing folder name for attachments.

brianteeman:
> Meanwhile I am working on another PR to implement validation for file and folder paths including the check if the file or folder exists so it would not be possible anymore to enter a not existing folder name for attachments

And what if the file or folder is removed later?

richard67:
> And what if the file or folder is removed later?

@brianteeman That's a different thing which requires validation when the folder is used, what this PR here does for the mail attachments. But my new PR will solve @PhilETaylor 's issue that you can enter not existing folders and save the form and get the message that the form is saved and all looks fine.

joomdonation:
I have tested this item :white_check_mark: successfully on cd0c6d8c9889e6e2906c1ab10085dda65d284911<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

brianteeman:
The more I look at this the more I ask myself what is the point of the filename field. What purpose does it serve? What is the usecase that it satisfies

richard67:
@brianteeman You can send an attachment with language-specific file name.

brianteeman:
ok - an edge case that I didnt consider but a valid one - thanks

joomdonation:
@PhilETaylor Could you please test this PR?

PhilETaylor:
Ive just landed a short while ago and still at the airport. Once Im home and showered its in my list.

191455391_10159213209351870_902310662463094669_n

richard67:
@PhilETaylor Could you test this PR again? Thanks in advance.

administrator/components/com_mails/config.xml
```diff
@@ -44,6 +44,8 @@
 			label="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_LABEL"
 			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
 			size="40"
+			validate="filePath"
+			exclude="administrator|api|bin|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
```
PhilETaylor:
```suggestion
			exclude="administrator|api|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
```

There is no `bin` folder in Joomla 4 and the admin script deletes all joomla 3 files from that folder, Im guessing we are keeping this in here for those that have additional scripts in the `bin` during 3->4 migration ? if you are happy its here then Im happy, but if its just a copy and paste issue then remove it.

libraries/src/Mail/MailTemplate.php
```diff
@@ -464,4 +475,28 @@ public static function deleteTemplate($key)
 
 		return $db->execute();
 	}
+
+	/**
+	 * Check and if necessary fix the file name of an attachment
```
PhilETaylor:
```suggestion
	 * Check and if necessary fix the file name of an attachment so that the attached file has the same extension as the source file, and not a different file extension
```

libraries/src/Mail/MailTemplate.php
```diff
@@ -464,4 +475,28 @@ public static function deleteTemplate($key)
...
+	 *
+	 * @param   string  $file  Path to the file to be attached
+	 * @param   string  $name  The file name to be used for the attachment
+	 *
+	 * @return  string  The corrected file name for the attachment
+	 *
+	 * @since   4.0.0
+	 */
+	protected function getAttachmentName($file, $name)
```
PhilETaylor:
```suggestion
	protected function getAttachmentName(string $file, string $name): string
```

As this is a new method, lets type hint it all the way :)

PhilETaylor:
I have tested this item :white_check_mark: successfully on cd0c6d8c9889e6e2906c1ab10085dda65d284911<br /><br />This PR is great enough to be merged, a lot of hard work by Richard and the initial security issue is well catered for and fixed now.. Just some notes:

1) #34263 still occurs when trying to send test email from admin when db contains manipulated ../../etc/passwd file name - however, on frontend when trying to attach ../../etc/passwd to an email (tested with password reset) you correctly get "Joomla\CMS\Filesystem\Path::check() - Use of relative paths not permitted" (See #34130) and "Failed sending email."

2) There is a logic issue where, if you set a correct folder, and set a correct attachment file to a mail, and then go an rename the folder on the hard disk - so it no longer exists as configured as the attachment folder, then the mails no longer show the configured files as attachments (which are still stored in the db). Thats not the fault of this PR

3) Further to 2 above, If there is a set a correct folder, and set a correct attachment file to a mail, and then the folder is inaccessable for some reason - the email is still sent, just with no attachment, and does not throw an exception that it could not find the file to attach. The "Summary of Changes" state:

> Sending email will be rejected with an exception [...] if the full path to the attachment file is not valid or the file doesn't exist.

That doesnt happen. (This was point 6 in the comment https://github.com/joomla/joomla-cms/pull/34233#issuecomment-850529623)

4) If my filename is "FileWithNoExtensionIsValidTest" and my name is "mine.txt" the file is attached with name "mine". That was not expected, as the original filename had no extension. But Im ok with it being this way. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

administrator/components/com_mails/config.xml
richard67: The bin folder is there because this exclude was copied from a place where it existed before, also in J3. Will remove it here.

richard67:
@PhilETaylor I've applied your suggestions. Could you and @joomdonation briefly test again? Thanks in advance.

PhilETaylor:
I have tested this item :white_check_mark: successfully on e4b46ea9a8fb835127d98920cf2d16a9bddc45ab<br /><br />Im still happy after changed made in e4b46ea9a8fb835127d98920cf2d16a9bddc45ab<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

richard67:
@PhilETaylor For point 6 of your comment about check if the folder exists at the end see my new PR #34285 . For the issue #34262 with filelist field, and that applies also to folderlist, too, see my new PR #34284 .

PhilETaylor:
I only mentioned point 6 because it was specifically called out as a change in this PR and therefore expected that this PR would fix it. I was aware that you had other PRs open - thanks for those. Some hard work here while Ive been out having fun... 

I'll take these other PR's for a spin shortly. Not sure when @wilsonge is tagging the RC, but this PR absolutely needs to be in the RC, the others not so much so.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34233">issues.joomla.org/tracker/joomla-cms/34233</a>.</sub>

richard67:
Previous test result are still valid because the changes after that were just changes in code comments, type hint for the new method and code style.

PhilETaylor:
> The more I look at this the more I ask myself what is the point of the filename field. What purpose does it serve? What is the usecase that it satisfies

The additional oversight is that placeholders cannot be used in the name field, like they can in the mail body/subject. 

Thinking how this feature might be used, the 3pd might want to add dynamic detail to the file name. E.g Joe Blogs buys an ebook, and so the filename becomes `Order123_Bloggs_book.pdf` dynamically using the placeholders. 

Forked to https://github.com/joomla/joomla-cms/issues/34287

PhilETaylor:
Screenshot 2021-05-30 at 17 05 00
🎉🎉

wilsonge:
Thanks!

richard67:
Thanks all!

PR #34257, opened 2021-05-28
[4.0] mod_popular with disabled hits

brianteeman:
When hits are disabled for articles (in the article options) it makes no sense to display the Most Read Articles module in the site or admin as the contents will never be updated.

This PR changes the output of the module so that instead of a list of non-updating articles a message is displayed.

### Site Module
image

### Admin Module
image


chmst:
The module seems to be very old and there are only few use cases when it makes sense, with or without hit counting. I suggest removing it from the dashboard for new installations. @richard67 what do you think?

richard67:
> The module seems to be very old and there are only few use cases when it makes sense, with or without hit counting. I suggest removing it from the dashboard for new installations. @richard67 what do you think?

@chmst I think that with hit count it sill makes sense and I would like to have that module on my dashboard because I have it for my homepage, too. But that's just my personal opinion, I have nothing to decide and I don't want to have anything to decide.

brianteeman:
I actually use it a lot - so maybe others so as well - dont really care if its default or not. Either way its beyond the scope of this PR to remove it

chmst:
It is not in scope of this PR, just a matter of taste.

brianteeman:
we really have to get out of the habit of out of scope comments. Far too often they completely detract from the actual PR. my 2c

Abernyte-Git:
I have tested this item :white_check_mark: successfully on 1f63709d9ccef7e55cccab10e4b566225ee869b2<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34257">issues.joomla.org/tracker/joomla-cms/34257</a>.</sub>

ghost:
I have tested this item :white_check_mark: successfully on 1f63709d9ccef7e55cccab10e4b566225ee869b2<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34257">issues.joomla.org/tracker/joomla-cms/34257</a>.</sub>

Quy:
Should the module be displayed on the front end in this case?

brianteeman:
Absolutely yes. Think like a user.

I turn off hits in January. In December I decide to add the module but it doesn't appear anywhere - joomla is broken.
This way it appears and they are informed that they need to re-enable hits

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34257">issues.joomla.org/tracker/joomla-cms/34257</a>.</sub>

chmst:
Thanks!

brianteeman:
Thanks

brianteeman:
@SharkyKZ is it your mission in life to just sit there and put the thumbs down. This is an open repo - if you have a comment to make then please do so.

SharkyKZ:
With maintainers like these it's pointless.

richard67:
Thanks for the flowers.

brianteeman:
I dont believe anyone is a mind reader. If there is something wrong then say it

PhilETaylor:
For once - I agree with Sharky. This broke things. PR incoming :-(

brianteeman:
thats why you have to speak and not thumbs

PhilETaylor:
Detail: https://github.com/joomla/joomla-cms/pull/34318

PhilETaylor:
#34320 also moves the output rendering out of the Model helper where it is "more correct" to be outputting output, rather than in the M of the MVC

PR #34260, opened 2021-05-28
Added more context to Filesystem::delete exception

echterhoff:
I would prefer to know where a problematic file is located during a failed update.

Pull Request for Issue # .

### Summary of Changes

Added a check for JDEBUG and present a full path of the undeletable file to the user in case the system runs in debug mode.

### Testing Instructions

Make a core file inaccessible to the web server and run a core refresh or update. In case it can't delete a file, it will raise an error with just the file name or in debug mode with the full path name.

Just a filename without any information where to find that file.

A full path to the file...

### DIFF
--- a/libraries/vendor/joomla/filesystem/src/File.php
+++ b/libraries/vendor/joomla/filesystem/src/File.php
@@ -123,7 +123,7 @@ class File
 
 			if (!Path::canChmod($file))
 			{
-				throw new FilesystemException(__METHOD__ . ': Failed deleting inaccessible file ' . $filename);
+				throw new FilesystemException(__METHOD__ . ': Failed deleting inaccessible file ' . ( defined('JDEBUG') && JDEBUG ? $file : $filename) );
 			}
 
 			// Try making the file writable first. If it's read-only, it can't be deleted
@@ -134,7 +134,7 @@ class File
 			// as long as the owner is either the webserver or the ftp
 			if (!@ unlink($file))
 			{
-				throw new FilesystemException(__METHOD__ . ': Failed deleting ' . $filename);
+				throw new FilesystemException(__METHOD__ . ': Failed deleting ' . ( defined('JDEBUG') && JDEBUG ? $file : $filename) );
 			}
 		}
 


richard67:
@HLeithner @nibra Shouldn't this change be made upstream here https://github.com/joomla-framework/filesystem and then a new release of the upstream package should be fetched into the CMS? Or shall the change be made at both places? Or is it ok to make it only here like this PR does?

Quy:
@richard67 This should be done upstream and not here.

echterhoff:
Should I request the pull on the upstream package? If so, its no problem.

I was not sure wheter the use of JDEBUG constant actually belongs to the filesystem sources or not...

echterhoff:
@PhilETaylor not sure if I missed something but I changed the message only for the debug mode. Throwing errors at developers that are meaningless is quite pointless. The exception stack in debug mode tells me enought to know which method throwed what from where but in this case actually nothing about the file that is the reason for the exception.

nibra:
The Framework is completely unaware of JDEBUG.

echterhoff:
2021-05-28 20_26_01-Error_ 0 Joomla_Filesystem_File__delete_ Failed deleting inaccessible file webli

PhilETaylor:
@echterhoff Sorry if I confused you, I was making a general point about the whole throwing of exceptions and not specifically about your changes. 

The point I was making is that this:

```php
throw new FilesystemException(__METHOD__ . ': Failed deleting inaccessible file ' . $filename);
```

should be this:

```php
throw new FilesystemException('Failed deleting inaccessible file ' . $filename);
```

and that the `__METHOD__ ` should never be part of a string message from an exception (and even more importantly, should never then be rendered to the user to tell the user the exception - The internals of code should never be shown to users - when was the last time you got an error page on Amazon.com and they told you that the error was in a method named `chargeTheUser20PercentMore()` .. you would never see PayPal.com show you their method names or class names... but yet Joomla and Joomla Framework classes do this a lot...

HLeithner:
> and that the `__METHOD__ ` should never be part of a string message from an exception (and even more importantly, should never then be rendered to the user to tell the user the exception - The internals of code should never be shown to users - when was the last time you got an error page on Amazon.com and they told you that the error was in a method named `chargeTheUser20PercentMore()` .. you would never see PayPal.com show you their method names or class names... but yet Joomla and Joomla Framework classes do this a lot...

ever used a java website ;-)

@echterhoff I think we can change $filename to $file in the framework but this would and in an information disclosure. 

What we can do is to extend the `FilesystemException` adding the relevant information to it (like the full path) and then expose this information in the cms if needed.

I'm closing this pr because it can't be merged but discussion should continue.

Issue #34236, opened 2021-05-26
[4] filepath form field validation rule doesnt reject invalid folders.  


PhilETaylor:
### Steps to reproduce the issue

Ensure you dont have a folder called `ABCDEF` in the root of your site. 

add `	validate="filePath"` to an xml form field like below

```xml
<field
			name="attachment_folder"
			type="text"
			label="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_LABEL"
			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
			size="40"
			validate="filePath"
			exclude="administrator|api|bin|cache|cli|components|includes|language|layouts|libraries|modules|plugins|templates|tmp"
		/>
```

Use that form and enter `ABCDEF` as the field value.

### Expected result

`ABCDEF` is rejected as it doesnt exist, therefore its an invalid location 

### Actual result

`ABCDEF` passes validation. 

`https://raw.githubusercontent.com/4Hackerz/C99-Shell/master/` passes validation

`ftp://example.com` passes validation

`,http://example.com` passes validation

### Bonus

Backticks are also accepted in the file path and the following is a valid folder name EDIT (hmm hard to add backticks in GitHub issues, here is an image) 

<img width="379" alt="Screenshot 2021-05-26 at 20 39 07" src="https://user-images.githubusercontent.com/400092/119720882-6ece0500-be62-11eb-99eb-fa5f7624da47.png">

joomdonation:
PR is https://github.com/joomla/joomla-cms/pull/34271 . Please test.

Fedik:
the path rule do not supposed to check if path exists, but to check if there stuff like `blabla/path`, that all.

> Backticks are also accepted 

They are valid in folder name.

Do not see issue here.

PhilETaylor:
Then you probably have not lived long enough and do not understand the point I was making.

In this context it's "probably" ok, however it would be highly frowned upon to use backticks in file names and folder names and anything else that is then evaluated - for example an filtered, and Unescaped, passing to passthur or exec or system as part of a command eg to move a file with mv 

And yes, I have seen vulnerable code from 3pd that does exactly this!

Fedik:
yea, but well, you want to much from "filePath" rule  :smile: 
It just to checking if the given string looks like a path, no more no less.

For what you asked it better to create more obvious rule like "pathExist" or something.

PhilETaylor:
[...]

PR #34270, opened 2021-05-29
[4.0] Improve file path rule

joomdonation:
Pull Request for Issue #34236.

### Summary of Changes
This PR makes small improvement to `FilePathRule` so that the entered path not only must be an internal path but also needs to be a valid/existing folder.

### Testing Instructions
1. Access to Content -> Media, click on Options button in the toolbar
2. Change **Path to Files Folder** from images to a random string like aaa (a none existing folder)
3. Before patch: The change is saved
4. After patch: The change is not saved, there is an error message displayed **Invalid field: Path to Files Folder**


joomdonation:
Sorry, wrong branch.

Issue #34269, opened 2021-05-29
[4.0] Web services don't always comply with JSON:API


pjdevries:
### Steps to reproduce the issue

Access a `PATCH` or `DELETE` endpoint. 

### Expected result

A valid [JSON:API](https://jsonapi.org/) document response.

### Actual result
`PATCH`: `[]`
`DELETE`: `1`

wilsonge:
Agree patch needs fixing.

To me delete should be returning the empty response as https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/MVC/Controller/ApiController.php#L322 doesn’t output anything. Should be an easy fix somewhere

wilsonge:
OK just tested this - for me I'm not getting any response back from delete (as expected).

Obviously as you say patch is incorrect

alikon:
so for the PATCH
this will be enough  in your opinion ? @wilsonge @pjdevries 
after this line 
https://github.com/joomla/joomla-cms/blob/884d197bc5bd9face0130508577ce622b9bc626d/libraries/src/MVC/Controller/ApiController.php#L396

adding  `$this->displayItem($recordId);   `

pjdevries:
@alikon Makes sense, since `POST` is handled in a similar fashion. But I'm not the expert, so I leave it up to @wilsonge.

alikon:
better discuss with code :smile: 
see #34313

pjdevries:
> OK just tested this - for me I'm not getting any response back from delete (as expected).

Don't want to nit pick, bu no response is not entirely as expected. That is to say, if we want to comply with the JSON:API completely. Take a look at [JSON:API](https://jsonapi.org/format/#crud-deleting-responses) says about `DELETE` responses. A 200 response should return 'only top-level meta data'.

wilsonge:
> Don't want to nit pick, bu no response is not entirely as expected. That is to say, if we want to comply with the JSON:API completely. Take a look at JSON:API says about DELETE responses. A 200 response should return 'only top-level meta data'.

I agree but I'm getting back a 204 not a 200 (which is expected as we're setting that here https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/MVC/Controller/ApiController.php#L322) and that's totally allowed by JSON API in that section.

Would be useful to see what request you're getting a 200 from on a delete. Because I can't seem to reproduce that.

pjdevries:
[...]

Issue #34261, opened 2021-05-28
[4.0] - Web service PATCH requests not validatated properly


pjdevries:
### Steps to reproduce the issue

Create an article or obtain the `id` of an existing article.
Obtain an active Joomla Api Token.
Run the following cURL command after replacing **[your J4 website]** with the name of your website, **[your article id]** with the above mentioned article `id` and **[your user token]** with the above mentioned Joomla Api Token.
```
curl --location --request PATCH 'https://[your J4 website]/api/index.php/v1/content/articles/[your article id]' \
--header 'Content-Type: application/json' \
--header 'X-Joomla-Token: [your user token]' \
--data-raw '{
    "state": 1
}'
```

### Expected result
'[]'

### Actual result
```
{
    "errors": [
        {
            "title": "Field required: Title\nField required: Category"
        }
    ]
}
```

### Additional comments

For a large part, web services use existing, non web service specific logic. In this particular case, the `Joomla\CMS\MVC\Controller\ApiController::save()` method obtains a `Joomla\CMS\Form\Form` object from the administrator `ArticleModel` and calls that model's `validate` method. Unfortunately that method seems to expect a valid article and not a just one or more fields to update (PATCH) an existing article with.

alikon:
https://github.com/joomla/joomla-cms/blob/f61b9ae293e289a4a9177e6cdea12d8576f46419/libraries/src/MVC/Controller/ApiController.php#L436-L437

a quick & dirty hack could be to have a specific form for the api 
`Form::addFormPath(JPATH_COMPONENT_ADMINISTRATOR . '/forms/api');`

pjdevries:
@alikon I'm not sure I grasp your intention. Do you propose to add specific forms with the sole purpose of validating web service request data? To be honest, I am not a huge fan of using form definitions in the context of web services. @joomdonation suggested something similar in #34101. But maybe you two are right and I am totally wrong.

Anyhow, even if we use forms for this purpose, how do you distinguish different requests (`POST`, `PATCH`) in order to do the appropriate validations?

joomdonation:
Adding a new form does not help, I think. I think an acceptable solution would be pre-load $data array with the data which we are having in the database for the record at the moment. So adding these lines of code below after this line https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/MVC/Controller/ApiController.php#L432 might help

```php
if ($recordKey && $table->load($recordKey))
{
	$fields = $table->getFields();

[...]

PR #34274, opened 2021-05-29
[4][webservices] PATCH verb

alikon:
Pull Request for Issue #34261 +  #34269 

### Summary of Changes
fix PATCH
return the record in case of success
return a 404 in case of not found

### Testing Instructions
see #34261

PATCH an existing article
PATCH an unexisting article

you must add title and category_id to change only 1 field

you can PATCH the fields in the body request

### DIFF
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -372,7 +372,7 @@ class ApiController extends BaseController
 
 		$recordId = $this->input->getInt('id');
 
-		if (!$recordId)
+		if ($recordId && !$table->load($recordId))
 		{
 			throw new Exception\ResourceNotFound(Text::_('JLIB_APPLICATION_ERROR_RECORD'), 404);
 		}
@@ -394,6 +394,7 @@ class ApiController extends BaseController
 		}
 
 		$this->save($recordId);
+		$this->displayItem($recordId);
 
 		return $this;
 	}
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -431,20 +431,19 @@ class ApiController extends BaseController
 		$checkin    = property_exists($table, $table->getColumnAlias('checked_out'));
 		$data[$key] = $recordKey;
 
-		
 		if ($this->input->getMethod() === 'PATCH')
 		{
 			if ($recordKey && $table->load($recordKey))
 			{
 				$fields = $table->getFields();
-			
+
 				foreach ($fields as $field)
 				{
 					if (array_key_exists($field->Field, $data))
 					{
 						continue;
 					}
-				
+
 					$data[$field->Field] = $table->{$field->Field};
 				}
 			}
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -431,25 +431,22 @@ class ApiController extends BaseController
 		$checkin    = property_exists($table, $table->getColumnAlias('checked_out'));
 		$data[$key] = $recordKey;
 
-		if ($recordKey && $table->load($recordKey))
+		
+		if ($this->input->getMethod() === 'PATCH')
 		{
-			$fields = $table->getFields();
-
-			foreach ($fields as $field)
+			if ($recordKey && $table->load($recordKey))
 			{
-				if (array_key_exists($field->Field, $data))
+				$fields = $table->getFields();
+			
+				foreach ($fields as $field)
 				{
-					continue;
+					if (array_key_exists($field->Field, $data))
+					{
+						continue;
+					}
+				
+					$data[$field->Field] = $table->{$field->Field};
 				}
-
-				$fieldValue = $table->{$field->Field};
-
-				if (is_string($fieldValue) && is_array(json_decode($fieldValue)))
-				{
-					$fieldValue = json_decode($fieldValue);
-				}
-
-				$data[$field->Field] = $fieldValue;
 			}
 		}
 
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -372,7 +372,7 @@ class ApiController extends BaseController
 
 		$recordId = $this->input->getInt('id');
 
-		if ($recordId && !$table->load($recordId))
+		if (!$recordId)
 		{
 			throw new Exception\ResourceNotFound(Text::_('JLIB_APPLICATION_ERROR_RECORD'), 404);
 		}
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -431,6 +431,29 @@ class ApiController extends BaseController
 		$checkin    = property_exists($table, $table->getColumnAlias('checked_out'));
 		$data[$key] = $recordKey;
 
+		if ($recordKey && $table->load($recordKey))
+		{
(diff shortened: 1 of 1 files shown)


pjdevries:
Like I proposed in issue #34261, I would prefer to incorporate this logic in the `edit()` as opposed to the `save()` method. After all, we only need it for `PATCH` requests. I would add the logic in a separate method, because I like modularity and keep the original method as clean as possible, and call it from the `edit()` method just before line https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/MVC/Controller/ApiController.php#L396.

Not critical but personal preferences :)

pjdevries:
@alikon Not sure how you did it, but a `PATCH` request with patch-126 gives me a 404.

pjdevries:
> @alikon Not sure how you did it, but a `PATCH` request with patch-126 gives me a 404.

Silly me this time. patch-126 does not have the pluralised endpoints :)

pjdevries:
> Like I proposed in issue #34261, I would prefer to incorporate this logic in the `edit()` as opposed to the `save()` method. 

This proves a bit more complicated than I imagined. Several api controllers override the `save()` method, setting the `data` input attribute. It is possible to supplement and overwrite the original raw json input data, but that feels like a bad idea.

So let's stick to the current implementation, as suggested by @joomdonation. I tested this PR and it does solve the `PATCH` validation issue.

libraries/src/MVC/Controller/ApiController.php
```diff
@@ -431,6 +431,28 @@ protected function save($recordKey = null)
...
+			foreach ($fields as $field)
+			{
+				if (array_key_exists($field->Field, $data))
+				{
+					continue;
+				}
+
+				$fieldValue = $table->{$field->Field};
+
+				if (is_string($fieldValue) && is_array(json_decode($fieldValue)))
```
wilsonge: What's this working around? Seems like a weird piece of code :)

wilsonge:
> This proves a bit more complicated than I imagined. Several api controllers override the save() method, setting the data input attribute. It is possible to supplement and overwrite the original raw json input data, but that feels like a bad idea.

I've fixed most of these in https://github.com/joomla/joomla-cms/pull/34280 - just leaves the OverridesController which fully overrides this method.

> I would prefer to incorporate this logic in the edit() as opposed to the save() method. After all, we only need it for PATCH requests

So I agree with this but like mentioned it's pretty hard because we don't actually grab any data in the edit method and like you say passing it around is tricky. However in the save method what we can and should do is check the request type (`if $this->getInput()->getMethod() === 'PATCH')` before we try and run any of this in save save function.

libraries/src/MVC/Controller/ApiController.php
alikon: yes indeed

alikon:
request type check added

libraries/src/MVC/Controller/ApiController.php
```diff
@@ -431,6 +431,24 @@ protected function save($recordKey = null)
 		$checkin    = property_exists($table, $table->getColumnAlias('checked_out'));
 		$data[$key] = $recordKey;
 
+		if ($this->input->getMethod() === 'PATCH')
+		{
+			if ($recordKey && $table->load($recordKey))
+			{
+				$fields = $table->getFields();
```
wilsonge, line 439: Is this any different to $table->getProperties() ?

pjdevries: > Is this any different to $table->getProperties() ?

They definitely do two different things, but I guess the end result is the same most of the time. But since `Joomla\CMS\Object\CMSObject` is deprecated, I think it's the right choice to favor `Table::getFields()` over `CMSObject::getProperties()`.

pjdevries:
I tested the previously failing `PATCH` request and it now succeeds.

joomdonation:
@pjdevries Could you please help testing to see with the PATCH request, tags data for article is kept or lost? I am asking for this because the tags data is not populated in $data array for our current code, so I'm worry that it might be lost (sorry I could not check it myself, too busy today)

libraries/src/MVC/Controller/ApiController.php
joomdonation: @wilsonge Just for your information, I was suggesting this code (without testing) because I thought that in the table, there could be some JSON encoded data for some fields, mean the original data for these fields before saving is array, so it is decoded and restore as array to have it saves properly

I don't know if that is needed or not. Guess we need to test and make sure the data for json encoded fields in the table still being saved properly after PATCH request.

pjdevries:
> 
> 
> @pjdevries Could you please help testing to see with the PATCH request, tags data for article is kept or lost? I am asking for this because the tags data is not populated in $data array for our current code, so I'm worry that it might be lost (sorry I could not check it myself, too busy today)

I'm not sure that's relevant for this issue. The problem here is unnecessary validation of otherwise required attributes. For `articles` those would be `title` and `catid` for instance. Validation of those is required for new articles, but not when using `PATCH` to change one or more not required attributes. So I think supplementing the request data just like we do know, solvers this problem. Since tags are not required attributes, I think there is no problem there.

libraries/src/MVC/Controller/ApiController.php
```diff
@@ -372,7 +372,7 @@ public function edit()
 
 		$recordId = $this->input->getInt('id');
 
-		if (!$recordId)
+		if ($recordId && !$table->load($recordId))
```
wilsonge: So if we remove this - what actually fails?

Also this will now 'succeed' if the `recordId` is null

alikon: it gives 500 when you PATCH a non existent article

and 

but $table->load(null) return false  so you got 400

wilsonge: https://github.com/joomla/joomla-cms/commit/d87d441f0b032bcd9ab71ef3dea46376de77612a 500 fixed :)

alikon: :+1: 
removed

wilsonge:
Thanks!

marconte21870:
The same error:"{"errors":[{"title":"Field required: Title\nField required: Category"}]}"
when i send curl for new article:
curl -X POST [myhttspsite]/joomla/api/index.php/v1/content/articles
-H "Content-Type: application/json"
-H "X-Joomla-Token: mytoken"
-d "{'alias': 'my-article','articletext': 'My text','catid':'2,'language': '*','metadesc': '','metakey': '','title': 'Here's an article'}"

alikon:
double check the `'` vs `"`
this work fine 

`'{"alias": "my-article","articletext": "My text","catid": 2,"language": "*","metadesc": "","metakey": "","title": "Here is an article"}'`

plus this pr was for **PATCH** not for  **POST**

marconte21870:
Thanks for the reply, I didn't find anything on POST, as it seemed like a similar error I thought you could help me; I changed 'VS "but nothing are two idays that I try by changing all the url parameters, I try with postman with php file but nothing

alikon:
```
`curl --location --request POST 'http://[yoursite]/api/index.php/v1/content/articles' \
--header 'Content-Type: application/json' \
--header 'X-Joomla-Token:  [yourtoken]' \
--data-raw '{"alias": "my-article-test","articletext": "My test text","catid": 2,"language": "*","metadesc": "","metakey": "","title": "Here is my test article"}'`
```

just exported from postam

marconte21870:
I solved with your advice, it works but on postman it does not work, it is not important postman but it is important that it works on php and my n8n platform,
a sincere thank you

Issue #34262, opened 2021-05-28
[4][Security] filelist field type doesnt validate the input before saving


PhilETaylor:
### Steps to reproduce the issue

A good example of this is in https://github.com/joomla/joomla-cms/pull/34233, specifically this comment https://github.com/joomla/joomla-cms/pull/34233#issuecomment-850529623 where you can right click and modify the value of the selected item, and click save, and it is saved in the database. 

you do that like this video, and then click save, and then inspect your db and you will see the faked value added to the db

https://user-images.githubusercontent.com/400092/120015480-474e7800-bfdb-11eb-96a6-28ae194d008b.mp4

### Expected result

That any POSTed value is revalidated against the available options presented when the form was rendered before saving to the db.

### Actual result

The unfiltered value is saved to the db

brianteeman:
I gave up trying to get a response to this. I think I have 3 open and 2 closed issues =- the last of which was closed "now that j4 was as broken as j3 "

Fedik:
The field does not validate anything if it not configured to.

> where you can right click and modify the value of the selected item, and click save, and it is saved in the database.

For this we have a rule `Joomla\CMS\Form\Rule\OptionsRule`, to validate whether the value from the list.
That should be used when such validation needed.

https://github.com/joomla/joomla-cms/blob/3a5984d03366f707ef842ee48051af194f68848f/libraries/src/Form/Rule/OptionsRule.php

PhilETaylor:
> For this we have a rule Joomla\CMS\Form\Rule\OptionsRule

Oh sexy... pity its not actually used though :-)

richard67:
The OptionsRule only works for list fields but not for filelist fields as far as I remember (but I might be wrong).

But I'm currently working on new validation rules.

Fedik:
OptionsRule should work for all fields that provide `$field->options` property. (At least how it looks in the code).

Filelist field extend the list field, and so also should work.

richard67:
For the example mentioned here, we could also use the new "FilePathExistsRule" rule (i.e. `validate="filePathExists"`) which will be added by my PR #34277 .

@PhilETaylor Would my PR #34277 close this issue here, or do you want to wait with that until it has been merged and the new rule is used at all places where it's necessary?

richard67:
Hmm, for filelist and folderlist fields, my new validation rules aren't really necessary because the files or folders have to exist for being added to the options.

But the OptionsRule would work.

So there are 2 ways to go:
[...]

PR #29721, opened 2020-06-20
[4.0] Filesystem plugin: limit folders choice to images and any custom folder

infograf768:
### Summary of Changes
Makes no sense to let choose core folders (except `images`) to contain media files.

### Testing Instructions
Display the filesystem local plugin after creating a custom root folder

### Before patch
Screen Shot 2020-06-20 at 11 09 46

### After patch

Screen Shot 2020-06-20 at 11 11 38

Therefore if we do this
Screen Shot 2020-06-20 at 11 14 26

we will get this
Screen Shot 2020-06-20 at 11 13 58

### NOTE:
We could also add recursing as proposed and closed here: https://github.com/joomla/joomla-cms/pull/29719 as this limits drastically the number of sub folders to fetch, preventing overload.

If we do, we would get:

Screen Shot 2020-06-20 at 08 06 48

PR #34277, opened 2021-05-29
[4.0] New form field validation rules for paths to existing folders or files

richard67:
Pull Request for Issue #34262 and possibly others.

### Summary of Changes

Currently, when you enter a valid path to a not existing file or folder, e.g. the path to files and images in the com_media options or the attachment selection in com_mails, you can save the form without any notice that the file or folder doesn't exist.

See issue #34262 for the filelist field or the discussion in PR #34233 for the attachment folder field in com_mails options, the discussions or issues related to com_media folders I have to search for, but I'm sure there are some.

This pull request (PR) fixes that by adding two new validation rules, which can be used with text fields or filelist fields or folderlist fields.
- "FilePathExistsRule" for paths to files
- "FolderPathExistsRule" for paths to folders

Both rules extend from rule "FilePathRule" and first call the check method of this parent.

After that, the file or folder is checked for existence if the field is required.

This PR adds use of these rules at following places, other places can be done in later PR's:

- "Path to Files Folder" and "Path to Images Folder" text fields in Media Manager (com_media) options
- "Attachments Folder" text field in Mail Templates (com_mails) options
- "File" filelist field for an attachment in a Mail Template

### Testing Instructions

The test starts on an installation of current 4.0-dev branch or latest 4.0 nightly build without the patch of this PR applied.

All actions are performed in backend.

1. Go to media manager options.

2. In one or both of the fields "Path to Files Folder" and "Path to Images Folder", enter a path which is valid e.g. regarding forbidden characters but which doesn't exist below the Joomla root, e.g. "gaga" or "images/gaga".

3. Save and close.
Result: Changes are saved, green success message shown.

4. Go to "System -> Templates -> Mail Templates" and edit the options.

5. In field "Attachments Folder", enter a path which is valid e.g. regarding forbidden characters but which doesn't exist below the Joomla root, e.g. "gaga" or "images/gaga".

6. Save the change.
Result: Changes are saved, green success message shown.

7. Change field "Attachments Folder" back to a path which is valid and exists and which contains files which can be attached to email, e.g. "images", and save and close.

8. Go to "System -> Templates -> Mail Templates" and edit one of the mail templates, e.g. "Global Configuration: Test Mail".

9. At the bottom in section "Attachments", add an attachment but don't select a file, or select a file and modify the value to an invalid one in the DOM using your browser's developer tools like it is described in issue #34262 .

10. Save the change.
Result: Changes are saved, green success message shown.

11. Apply the patch of this PR.

12. Go to media manager options which have still the path to not existing folder(s) from step 2.

13. Without making any changes, use the "Save" button.
Result: The form is not saved, you get an alert about invalid field.

14. Change the path(s) so they are valid and point to existing folder(s), and save and close.

15. Edit again the mail template you have edited in step 8 but don't make changes.

16. Use the "Save" button.
Result: The form is not saved, you get an alert about invalid field.

17. Select a file for the attachment, and use again the "Save" button.
Result: The form is saved.

- "File" filelist field for an attachment in a Mail Template: See issue #34262 .
- "Path to Files Folder" and "Path to Images Folder" text fields in Media Manager (com_media) options and "Attachments Folder" text field in Mail Templates (com_mails) options: See section "What this PR does not fix" in PR #34233 .

Will be added soon.

### Other information

The "FilePathExistsRule" doesn't work with fields of type "filelist" if the "stripext" element of that field is set to "true".

Documentation of form field validation rules in J4 needs to be extended by the two new rules, including what's mentioned in the previous section "Other information".

### DIFF
--- a/libraries/src/Form/Rule/FilePathExistsRule.php
+++ b/libraries/src/Form/Rule/FilePathExistsRule.php
@@ -10,6 +10,8 @@ namespace Joomla\CMS\Form\Rule;
 
 defined('JPATH_PLATFORM') or die;
 
+use Joomla\CMS\Filesystem\File;
+use Joomla\CMS\Filesystem\Path;
 use Joomla\CMS\Form\Form;
 use Joomla\CMS\Form\Rule\FilePathRule;
 use Joomla\Registry\Registry;
@@ -52,8 +54,20 @@ class FilePathExistsRule extends FilePathRule
 		}
 
 		// In case of a file list field we might have a directory, otherwise it's Joomla root
-		$parentPath = $element['directory'] ?? JPATH_ROOT;
+		if (isset($element['directory']) && trim((string) $element['directory']))
+		{
+			$parentPath = Path::clean((string) $element['directory']);
 
-		return is_file($parentPath . '/' . $value);
+			if (strpos($parentPath, Path::clean(JPATH_ROOT)) !== 0)
+			{
+				$parentPath = JPATH_ROOT . '/' . $parentPath;
+			}
+		}
+		else
+		{
+			$parentPath = JPATH_ROOT;
+		}
+
+		return File::exists($parentPath . '/' . $value);
 	}
 }
--- a/libraries/src/Form/Rule/FilePathExistsRule.php
+++ b/libraries/src/Form/Rule/FilePathExistsRule.php
@@ -43,6 +43,14 @@ class FilePathExistsRule extends FilePathRule
 			return false;
 		}
 
+		// If the field is empty and not required, the field is valid.
+		$required = ((string) $element['required'] == 'true' || (string) $element['required'] == 'required');
+
+		if (!$required && empty($value))
+		{
+			return true;
+		}
+
 		// In case of a file list field we might have a directory, otherwise it's Joomla root
 		$parentPath = $element['directory'] ?? JPATH_ROOT;
 
--- a/libraries/src/Form/Rule/FilePathExistsRule.php
+++ b/libraries/src/Form/Rule/FilePathExistsRule.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  Copyright (C) 2005 - 2020 Open Source Matters, Inc. All rights reserved.
+ * @license    GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\CMS\Form\Rule;
+
+defined('JPATH_PLATFORM') or die;
+
+use Joomla\CMS\Form\Form;
+use Joomla\CMS\Form\Rule\FilePathRule;
+use Joomla\Registry\Registry;
+
+/**
+ * Form Rule class for the Joomla Platform.
+ *
+ * @since  4.0.0
+ */
+class FilePathExistsRule extends FilePathRule
+{
+	/**
+	 * Method to test if the file path is valid and points to an existing file in or below the Joomla root or a directory
+	 *
+	 * @param   \SimpleXMLElement  $element  The SimpleXMLElement object representing the `<field>` tag for the form field object.
+	 * @param   mixed              $value    The form field value to validate.
+	 * @param   string             $group    The field name group control value. This acts as an array container for the field.
+	 *                                       For example if the field has name="foo" and the group value is set to "bar" then the
+	 *                                       full field name would end up being "bar[foo]".
+	 * @param   Registry           $input    An optional Registry object with the entire data set to validate against the entire form.
+	 * @param   Form               $form     The form object for which the field is being tested.
+	 *
+	 * @return  boolean  True if the value is valid and points to an existing file in or below the Joomla root, false otherwise.
+	 *
+	 * @since   4.0.0
+	 */
+	public function test(\SimpleXMLElement $element, $value, $group = null, Registry $input = null, Form $form = null)
+	{
+		if (!parent::test($element, $value, $group, $input, $form))
+		{
+			return false;
+		}
+
+		// In case of a file list field we might have a directory, otherwise it's Joomla root
+		$parentPath = $element['directory'] ?? JPATH_ROOT;
+
+		return is_file($parentPath . '/' . $value);
+	}
+}
--- a/libraries/src/Form/Rule/FilePathExistsRule.php
+++ b/libraries/src/Form/Rule/FilePathExistsRule.php
@@ -17,7 +17,7 @@ use Joomla\Registry\Registry;
 /**
  * Form Rule class for the Joomla Platform.
  *
- * @since  4.0.0
+ * @since  __DEPLOY_VERSION__
  */
 class FilePathExistsRule extends FilePathRule
 {
@@ -34,7 +34,7 @@ class FilePathExistsRule extends FilePathRule
 	 *
 	 * @return  boolean  True if the value is valid and points to an existing file in or below the Joomla root, false otherwise.
 	 *
-	 * @since   4.0.0
+	 * @since   __DEPLOY_VERSION__
 	 */
 	public function test(\SimpleXMLElement $element, $value, $group = null, Registry $input = null, Form $form = null)
 	{
--- a/libraries/src/Form/Rule/FilePathExistsRule.php
+++ b/libraries/src/Form/Rule/FilePathExistsRule.php
@@ -2,7 +2,7 @@
 /**
  * Joomla! Content Management System
  *
- * @copyright  Copyright (C) 2005 - 2020 Open Source Matters, Inc. All rights reserved.
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
  * @license    GNU General Public License version 2 or later; see LICENSE.txt
  */
 
--- a/libraries/src/Form/Rule/FolderPathExistsRule.php
+++ b/libraries/src/Form/Rule/FolderPathExistsRule.php
@@ -10,6 +10,8 @@ namespace Joomla\CMS\Form\Rule;
 
 defined('JPATH_PLATFORM') or die;
 
+use Joomla\CMS\Filesystem\Folder;
+use Joomla\CMS\Filesystem\Path;
 use Joomla\CMS\Form\Form;
 use Joomla\CMS\Form\Rule\FilePathRule;
 use Joomla\Registry\Registry;
@@ -52,8 +54,20 @@ class FolderPathExistsRule extends FilePathRule
 		}
 
 		// In case of a folder list field we might have a directory, otherwise it's Joomla root
-		$parentPath = $element['directory'] ?? JPATH_ROOT;
+		if (isset($element['directory']) && trim((string) $element['directory']))
+		{
+			$parentPath = Path::clean((string) $element['directory']);
 
-		return is_dir($parentPath . '/' . $value);
+			if (strpos($parentPath, Path::clean(JPATH_ROOT)) !== 0)
+			{
+				$parentPath = JPATH_ROOT . '/' . $parentPath;
+			}
+		}
+		else
+		{
+			$parentPath = JPATH_ROOT;
+		}
+
+		return Folder::exists($parentPath . '/' . $value);
 	}
 }
--- a/libraries/src/Form/Rule/FolderPathExistsRule.php
+++ b/libraries/src/Form/Rule/FolderPathExistsRule.php
@@ -24,7 +24,7 @@ use Joomla\Registry\Registry;
 class FolderPathExistsRule extends FilePathRule
 {
 	/**
-	 * Method to test if the folder path is valid and points to an existing folder (directory) in or below the Joomla root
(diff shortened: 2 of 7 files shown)


libraries/src/Form/Rule/FilePathExistsRule.php
```diff
@@ -0,0 +1,59 @@
...
+
+		if (!$required && empty($value))
+		{
+			return true;
+		}
+
+		// In case of a file list field we might have a directory, otherwise it's Joomla root
+		$parentPath = $element['directory'] ?? JPATH_ROOT;
+
+		return is_file($parentPath . '/' . $value);
```
Fedik: probably more safe will be use File::exists(), 
that should cleaning the case like `JPATH_ROOT . '../../../blabla/path'`

richard67: That check for cleaning is already done in the `parent::test` call.

Fedik: hm, or maybe I wrong,

richard67: That's why you will see we use `is_file` and `is_dir` after having used `path::check`, what the parent's test routine does.

richard67: If you are convinced, please mark the discussion as resolved.

Fedik: I have missed parent test sorry.

richard67:
I will add some new unit tests soon.

brianteeman:
Why do you have to add "validate="folderPathExists" to the xml ? What is the usecase where you would ever not want to validate it

richard67:
> Why do you have to add "validate="folderPathExists" to the xml ? What is the usecase where you would ever not want to validate it

@brianteeman We already do it that way for the images and files paths of the media manager with the rule `FilePathRule`, and I think in this context this question has already been discussed.

brianteeman:
Just because it was done before doesnt make it right

richard67:
@brianteeman https://github.com/joomla/joomla-cms/pull/34257#issuecomment-850353673 .

brianteeman:
it absolutely is in the scope of this pr. You are adding a validation rule - great. But why is it optional? Thats what this pr is doing and its making it off by default

richard67:
@brianteeman The scope of this PR is to improve existing mechanisms, not to establish new ones. A form field used for folders cannot know if we want to enter the path to an existing folder or to a folder which shall be created. Being able to chose different validation rules makes it possible to use the same field for both purposes. Making it possible to define that in the XML enables 3rd party extension developers to use own rules or no rules.

brianteeman:
🤦

richard67:
Unit tests have been added. Ready for testing.

Fedik:
> Why do you have to add "validate="folderPathExists" to the xml ? What is the usecase where you would ever not want to validate it

How else should it be?

It is how validation works in Joomla in general. The PR is good for what it doing.

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,59 @@
...
+		// If the field is empty and not required, the field is valid.
+		$required = ((string) $element['required'] == 'true' || (string) $element['required'] == 'required');
+
+		if (!$required && empty($value))
+		{
+			return true;
+		}
+
+		// In case of a folder list field we might have a directory, otherwise it's Joomla root
+		$parentPath = $element['directory'] ?? JPATH_ROOT;
```
joomdonation: Do we have a valid use case for directory attribute like this? I'm unsure if we ever have a case which directory set to an absolute folder like it is programmed/checked here.

tests/Unit/Libraries/Cms/Form/Rule/FolderPathExistsRuleTest.php
```diff
@@ -0,0 +1,91 @@
...
+	public function dataTest(): array
+	{
+		$xml = new \SimpleXMLElement('<field
+			name="file_path"
+			type="text"
+			label="COM_MEDIA_FIELD_PATH_FILE_FOLDER_LABEL"
+			description="COM_MEDIA_FIELD_PATH_FILE_FOLDER_DESC"
+			size="50"
+			default="images"
+			validate="filePathExists"
```
PhilETaylor:
```suggestion
			validate="folderPathExists"
```

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: @joomdonation Test it with a field of type "folderlist" which has this attribute set.

libraries/src/Form/Rule/FilePathExistsRule.php
```diff
@@ -0,0 +1,59 @@
...
+defined('JPATH_PLATFORM') or die;
+
+use Joomla\CMS\Form\Form;
+use Joomla\CMS\Form\Rule\FilePathRule;
+use Joomla\Registry\Registry;
+
+/**
+ * Form Rule class for the Joomla Platform.
+ *
+ * @since  4.0.0
```
PhilETaylor: Im assuming this is ok as it probably will be 4.0.0 and no need for `__DEPLOY_VERSION__`? (repeated throughout the new files)

richard67: For 4.0.0 it doesn't make a big difference.

PhilETaylor: No worries.

libraries/src/Form/Rule/FolderPathExistsRule.php
joomdonation: @richard67 For `directory` attribute like that, please note that it could be relative folder to root, not always absolute folder (I could say that relative to JPATH_ROOT is real use case). See https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/Field/FolderlistField.php#L190-L195

richard67: I see. Will check and provide a fix.

libraries/src/Form/Rule/FilePathExistsRule.php
joomdonation: The suggestion from @Fedik is still valid, I think. We should use `File::exists()` instead of `is_file` here because from what I can see `$parentPath . '/' . $value` has not had Path::clean yet (so it could have problem with \ character on Linux).

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,59 @@
...
+
+		return is_dir($parentPath . '/' . $value);
```
joomdonation: Better use `Folder::exists` instead of `is_dir` here. The reason is same with one of my comment above.

richard67: then the same should apply to the is_file in the other one ... but the check in the parent class called before all that did already a path::clean on the $value part.

richard67: Done.

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: Done.

richard67:
Ready for testing. I think.

richard67:
Just thinking that this might be a bit over engineered with the check for the directory element, because that’s only needed for filelist and folderlist fields where the files or folders have to exist for getting into the list field, so a validation with the options rule should be enough for these 2 kinds of fields, and the new validation rules added by this PR here would be needed only for text fields.  It i have to check that first to be sure.

@wilsonge What’s your opinion on this?

richard67:
On the other side, a filelist or folderlist field can contain additional values, like "none selected", and the new validation rules would make sure that something was selected, while the options rule wouldn't.

libraries/src/Form/Rule/FilePathExistsRule.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  Copyright (C) 2005 - 2020 Open Source Matters, Inc. All rights reserved.
```
brianteeman:
```suggestion
 * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
```

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,73 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  Copyright (C) 2005 - 2020 Open Source Matters, Inc. All rights reserved.
```
brianteeman:
```suggestion
 * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
```

tests/Unit/Libraries/Cms/Form/Rule/FilePathExistsRuleTest.php
```diff
@@ -0,0 +1,93 @@
+<?php
+/**
+ * @package     Joomla.UnitTest
+ * @subpackage  HTML
```
brianteeman:
```suggestion
 * @subpackage  Rule
```

tests/Unit/Libraries/Cms/Form/Rule/FolderPathExistsRuleTest.php
```diff
@@ -0,0 +1,91 @@
+<?php
+/**
+ * @package     Joomla.UnitTest
+ * @subpackage  HTML
```
brianteeman:
```suggestion
 * @subpackage  Rule
```

richard67:
Thinking over if more, we should use text fields for files or folders only where we want to create new files or folders but not for selecting files or folders, and for creating the existing FilePathRule is enough.

Selecting files or folders should be done either with filelist and folderlist fields or in case of files with a media field, and in case of folders it would be nice to have a kind of media folder field which works like the media field, you can browse and select a folder.

There is PR #29721 which does that for the folders of the file system plugin, but it has an issue with the "exclude" element being checked as regular expression and so not working as expected.

So I am thinking about closing this PR and make another one for using `validation="options"` for filelist and folderlist fields where it is missing, and maybe help with fixing PR #29721 or another way not to use text fields for choosing existing files or folders.

brianteeman:
I have tested this item :red_circle: unsuccessfully on f4442288dfac92cf992fd4269dfbdf894abcec0f<br /><br />Additional tests

1. Remove current path and save - saved with default value - good
2. Remove current path and replace with two spaces - saved - bad<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34277">issues.joomla.org/tracker/joomla-cms/34277</a>.</sub>

richard67:
@brianteeman Could you explain your additional test more detailed? I don't understand what you mean with "Remove current path". Which field was it, and did you manipulate the value in DOM, or the "directory" element of the filelist field for the attachment?

brianteeman:
before

richard67:
@brianteeman I can confirm your finding, but I also can confirm it on a 4.0-dev branch without this PR being applied so the field still uses the FilePathRule. So it's not caused by or specific to the new rules added by this PR.

brianteeman:
The new rule is supposed to ensure that only a valid path is saved. But as shown you can save invalid paths

richard67:
@brianteeman The same applies to the old FilePathRule, and this is where the issue is located.

richard67:
@brianteeman But thanks for your review suggestions and the finding. I hope I did not waste too much of your time with this PR in case if I decide to close it due to my thoughts in my previous comments.

richard67:
@brianteeman The problem with and without this PR is that `Path::check` or `Path::clean` result in JPATH_ROOT being used if the value is empty or consists of spaces only, and that's valid, and so validation passes. That's nothing I can change with this PR except if I change the new rules to allow only files in folder below the Joomla root but not in the Joomla root, and not allow empty value or spaces which result in the Joomla root being used, in opposite to the existing FilePathRules which allows that.

richard67:
@brianteeman Final one: For files we are ok because they can't be empty, and for folder we should not allow the Joomla root. The old FilePathRule was not aware of if we want to check a file or a folder, and so you are right: It has to be fixed here in the PR for the FolderPathExists" rule.

richard67:
Hmm, it would need to set `required="true"` for the field to not allow empty field or spaces. But that would mean the form doesn't save and reset to the default when using an empty field.

Why an empty field is set to the default value and a field with spaces only is not when saving, that's another issue which can't really be solved with this PR here.

richard67:
@brianteeman Ok, I've fixed it finally.

richard67:
Closing in favour of PR's #34285 and #34284 . Please test. Thanks in advance, and thanks for all reviews and hints so far.

PR #34285, opened 2021-05-30
[4.0] New form field validation rule "FolderPathExistsRule"

richard67:
Pull Request for https://github.com/joomla/joomla-cms/pull/34233#issuecomment-850529623 point 6.

### Summary of Changes

Currently we have three places where we use text fields to specify the path to a folder which has to exist and be below the Joomla root:
- Field "Attachments Folder" in the Mail Templates component (com_mails) options.
- Fields "Path to Files Folder" and "Path to Images Folder" in the Media Manager (com_media) options.

This is bad because text fields should only be used to define names for new files or folders to be created, and because there is no validation done if the folder really exists if we use text fields to define paths to existing folders.

But currently we have no comfortable way to recursively, i.e. including all sub folders, navigate through all folders in the Joomla root.

For the mean time, as long as we don't have a better solution, this PR here adds a new validation rule "FolderPathExistsRule" to be used in such cases and which checks if the folder exists below the Joomla root, in addition to the validation which is done by the "FilePathRule".

The FilePathRule validation should not be extended to check if a file or folder exists because it shall still be possible to use it for validating paths to not existing files or folders.

The new rule added by this PR should not be used with folderlist fields because 1. it doesn't make sense for them, the "OptionsRule" should be used to validate these fields, and 2. it would not work for these fields because the "directory" element of such fields is not handled by the new rule.

In addition to these changes, this PR removes the "bin" folder from the existing "exclude" elements of the 2 text fields "Path to Files Folder" and "Path to Images Folder" in the Media Manager (com_media) options, because in J4 there is no "bin" folder anymore, not even after updates from 3.10.

### Testing Instructions

The test starts on an installation of current 4.0-dev branch or latest 4.0 nightly build without the patch of this PR applied.

All actions are performed in backend.

1. Go to media manager options.

2. In one or both of the fields "Path to Files Folder" and "Path to Images Folder", enter a path which is valid e.g. regarding forbidden characters but which doesn't exist below the Joomla root, e.g. "gaga" or "images/gaga".

3. Save and close.
Result: Changes are saved, green success message shown.

4. Go to "System -> Templates -> Mail Templates" and edit the options.

5. In field "Attachments Folder", enter a path which is valid e.g. regarding forbidden characters but which doesn't exist below the Joomla root, e.g. "gaga" or "images/gaga".

6. Save the change.
Result: Changes are saved, green success message shown.

7. Apply the patch of this PR.

8. Go to media manager options which have still the path to not existing folder(s) from step 2.

9. Without making any changes, use the "Save" button.
Result: The form is not saved, you get an alert about invalid field.

10. Change the path(s) so they are valid and point to existing folder(s), and save and close.
Result: Changes are saved, green success message shown.

11. Go to "System -> Templates -> Mail Templates" and edit the options which have still the path to not existing folder(s) from step 5.

12. Without making any changes, use the "Save" button.
Result: The form is not saved, you get an alert about invalid field.

13. Change the path so it is valid and points to an existing folder which is in the exclude list, e.g. "administrator", and use the "Save" button.
Result: The form is not saved, you get an alert about invalid field.

14. Change the path so it is valid and points to an existing folder, and save and close.
Result: Changes are saved, green success message shown.

See section "What this PR does not fix" in PR #34233 : You can specify paths to not existing folder for the three fields.

You can't specify paths to not existing folder for the three fields. The form can't be saved, and an alert is shown, telling which field is invalid.

When specifying paths to existing folders, the fields can be saved.

Documentation of form field validation rules in J4 needs to be extended by the new rule, including to mention that it only makes sense for text fields and will not work and not make sense with filelist or filderlist fields.


libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,67 @@
...
+	{
+		if (!parent::test($element, $value, $group, $input, $form))
+		{
+			return false;
+		}
+
+		// If the field is empty and not required, the field is valid.
+		$required = ((string) $element['required'] == 'true' || (string) $element['required'] == 'required');
+
+		if (!$required && empty($value))
```
Fedik: required already tested before executing the validator rule:
https://github.com/joomla/joomla-cms/blob/bbd7b55c174b4011b8bac1e5d24deae5ef671a76/libraries/src/Form/FormField.php#L1219-L1225

if it required and empty then it will not reach `FolderPathExistsRule::test()`

so it enough just:
```
if ($value === '' || $value === null)
{
  return true;
}
```

richard67: Right. Done.

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		// If the field is empty and not required so the previous test hasn't failed, the field is valid.
+		if ($value === '' || $value === null)
+		{
+			return true;
+		}
+
+		$pathCleaned = rtrim($value, \DIRECTORY_SEPARATOR);
+		$rootCleaned = rtrim(Path::clean(JPATH_ROOT), \DIRECTORY_SEPARATOR);
```
PhilETaylor, line 55: "JPATH_ROOT" is not allowed might be a better way to explain that

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,65 @@
...
+
+		// Joomla root is not allowed
```
PhilETaylor:
```suggestion
		// JPATH_ROOT is not allowed
```

tests/Unit/Libraries/Cms/Form/Rule/FolderPathExistsRuleTest.php
```diff
@@ -0,0 +1,91 @@
...
+			[false, $xml, 'Εικόνες'],
+			[false, $xml, 'Изображений'],
+		];
+	}
+
+	/**
+	 * Tests the FolderPathExistsRule::test method.
+	 *
+	 * @param   string             $expected  The expected test result
+	 * @param   \SimpleXMLElement  $element  The SimpleXMLElement object representing the `<field>` tag for the form field object.
```
PhilETaylor:
```suggestion
	 * @param   \SimpleXMLElement  $element   The SimpleXMLElement object representing the `<field>` tag for the form field object.
```

tests/Unit/Libraries/Cms/Form/Rule/FolderPathExistsRuleTest.php
```diff
@@ -0,0 +1,91 @@
...
+	 * @param   mixed              $value    The form field value to validate.
```
PhilETaylor:
```suggestion
	 * @param   mixed              $value     The form field value to validate.
```

richard67:
Wait guys I have to fix another mistake first.

tests/Unit/Libraries/Cms/Form/Rule/FolderPathExistsRuleTest.php
```diff
@@ -0,0 +1,91 @@
...
+			description="COM_MEDIA_FIELD_PATH_FILE_FOLDER_DESC"
+			size="50"
+			default="images"
+			validate="folderPathExists"
+			exclude="administrator|media"
+		/>'
+		);
+
+		return [
+			[true, $xml, ''],
```
PhilETaylor, line 46:
```suggestion
			[true, $xml, ''],
			[false, $xml, JPATH_ROOT],			
```

Add additional test for "Joomla root is not allowed"

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,65 @@
...
+defined('JPATH_PLATFORM') or die;
+
+use Joomla\CMS\Filesystem\Folder;
+use Joomla\CMS\Filesystem\Path;
+use Joomla\CMS\Form\Form;
+use Joomla\CMS\Form\Rule\FilePathRule;
+use Joomla\Registry\Registry;
+
+/**
+ * Form Rule class for the Joomla Platform.
```
PhilETaylor: This is a forked lib for the CMS now, and not for the Platform.

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license    GNU General Public License version 2 or later; see LICENSE.txt
```
PhilETaylor:
```suggestion
 * @license     GNU General Public License version 2 or later; see LICENSE.txt
```

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: The comment was copied from the FilePathRule 😞

PhilETaylor:
Something not right - I cannot save `images` or `media` :-( 

Screenshot 2021-05-30 at 16 23 54

PhilETaylor:
This is because `$pathCleaned` = `media` and so doesnt exist as its not an absolute path.

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,65 @@
...
+			return false;
+		}
+
+		// If the field is empty and not required so the previous test hasn't failed, the field is valid.
+		if ($value === '' || $value === null)
+		{
+			return true;
+		}
+
+		$pathCleaned = rtrim($value, \DIRECTORY_SEPARATOR);
```
PhilETaylor:
```suggestion
		$pathCleaned = rtrim(JPATH_ROOT . '/' . Path::clean($value), \DIRECTORY_SEPARATOR);
```
Something like this...

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,65 @@
...
+		$rootCleaned = rtrim(Path::clean(JPATH_ROOT), \DIRECTORY_SEPARATOR);
```
PhilETaylor, line 61: Do we need to reclean `JPATH_ROOT`? Surely its been cleaned before setting as a constant?

libraries/src/Form/Rule/FolderPathExistsRule.php
PhilETaylor: yeah looks like all the rules need updating :)

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: We do that at other places and I think it's needed because it's not cleaned when setting it.

libraries/src/Form/Rule/FolderPathExistsRule.php
```diff
@@ -0,0 +1,71 @@
+<?php
+/**
+ * Joomla! Content Management System
```
PhilETaylor, line 3: Also think this copyright header (and all the form rule headers) need to be standard Joomla ones?

libraries/src/Form/Rule/FolderPathExistsRule.php
PhilETaylor: 🤦‍♂️ haha

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: This is the standard for J4, as far as I know. It has been fixed by Brian in the previous PR #34277 .

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: I've done it here, feel free to make a PR for the others.

richard67:
> This is because `$pathCleaned` = `media` and so doesnt exist as its not an absolute path.

Has been fixed with my last commits. Please check again.

libraries/src/Form/Rule/FolderPathExistsRule.php
richard67: I have fixed that already after I had typed a comment for you that I need to fix a mistake and you shall wait.

richard67:
> Something not right - I cannot save `images` or `media` :-(

@PhilETaylor Has been fixed meanwhile, please test again.

administrator/components/com_mails/config.xml
```diff
@@ -44,6 +44,8 @@
 			label="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_LABEL"
 			description="COM_MAILS_CONFIG_FIELD_ATTACHMENT_FOLDER_DESC"
 			size="40"
+			validate="folderPathExists"
```
PhilETaylor, line 47:
```suggestion
			validate="folderPathExists"
			default="media"
```

Just a thought?

PhilETaylor: actually scrap that.

PhilETaylor:
I have tested this item :white_check_mark: successfully on ee0f510d093092bd0c448be260e7d103b804fd8f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34285">issues.joomla.org/tracker/joomla-cms/34285</a>.</sub>

libraries/src/Form/Rule/FolderPathExistsRule.php
brianteeman: My comment in the other PR was for the copyright line not this one.

PhilETaylor: Im happy to leave this out of this PR - as its a bigger issue - fork it to a new issue

richard67: Thanks.

joomdonation:
I have tested this item :white_check_mark: successfully on ee0f510d093092bd0c448be260e7d103b804fd8f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34285">issues.joomla.org/tracker/joomla-cms/34285</a>.</sub>

richard67:
Previous test results are still valid because the commit after that was just a clean branch update for solving a conflict caused by my other, merged PR.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34285">issues.joomla.org/tracker/joomla-cms/34285</a>.</sub>

richard67:
Thanks all!

PR #34288, opened 2021-05-30
[4.0] Centralized, maintainable handling of web service filtering

pjdevries:
Pull Request for Issue #34100

### Summary of Changes

Centralize filtering of data sets for relevant web service components. In this PR those are:
- com_banners
- com_categories
- com_contacts
- com_content
- com_fields
- com_tags
- com_users

Filters must be passed in accordance with the [JSON:API]()https://jsonapi.org/format/#fetching-filtering) specifications. Mapping of filter request parameters is done by means of an array, providing mappings between query parameter names and model state names. The mapping also contains information about the data type for proper input filtering. A map element looks like this:
```
'query param name' => [
    'name' => 'filter.model_supported_filter_attribute_name',
    'type' => 'input_filer_type'
]
```

The actual mapping is done in a utility method of `Joomla\CMS\MVC\Controller\ApiController` which is called from its `displayList()` method.

### Testing Instructions
Call any known list endpoint with one or more applicable filter parameters. For example:

{{base_path}}/api/index.php/v1/content/article?filter[category_id]=10
{{base_path}}/api/index.php/v1/content/categories?filter[search]=Uncategorised
etc.

As far as I know, no documentation for filtering exists yet.


libraries/src/MVC/Controller/ApiController.php
```diff
@@ -554,4 +565,28 @@ protected function preprocessSaveData(array $data): array
...
+
+		foreach ($this->modelFilterStateMap as $apiFilterName => $modelState)
+		{
+			if (!array_key_exists($apiFilterName, $apiFilterInfo) || trim($apiFilterInfo[$apiFilterName]) !== '')
+			{
+				continue;
+			}
+
+			$this->modelState->set($modelState['name'],
+				$filter->clean($apiFilterInfo[$apiFilterName], $modelState['type']));
```
richard67:
```suggestion
			$this->modelState->set(
				$modelState['name'],
				$filter->clean($apiFilterInfo[$apiFilterName], $modelState['type'])
			);
```
Fix to make PHP code style check pass in drone.

wilsonge:
This looks pretty sensible to me and I like it :)

Two things:

1. I think there should be a setting that allows you to override the state name - so you can have a filter that corresponds to a different state name if desired.
2. I know I said I wasn't wedded to the e.g. category_id thing - but i think it should be consistent. If we are updating the filter we should update the view to match it.

pjdevries:
> This looks pretty sensible to me and I like it :)

Thanx for the positive feedback. Honorable mention goes to @joomdonation for suggesting the concept to start with.

>     1. I think there should be a setting that allows you to override the state name - so you can have a filter that corresponds to a different state name if desired.

Not sure I understand. State names are hard coded in the models. Why would anyone want to map a filter to a different state name than what the filter is intended for? Isn't it up to the server to define the api and for the client to simply consume it the way it is defined? Or am I misunderstanding what you mean?

>     2. I know I said I wasn't wedded to the e.g. category_id thing - but i think it should be consistent. If we are updating the filter we should update the view to match it.

I definitely agree the api should be consistent. The views completely escaped my attention. 

There are two parts to this consistency: 
- Consistent attribute names in requests and responses per component.
- Consistent attribute names across all components.

What do you propose?

libraries/src/MVC/Controller/ApiController.php
pjdevries: @richard67 How do I know when or what to distribute over multiple lines? The [Joomla! Coding Standards](https://developer.joomla.org/coding-standards/php-code.html) does not mention anything about proper formatting function calls. What makes it worse, is that the [Basic Guidelines](https://developer.joomla.org/coding-standards/basic-guidelines.html) state the following:

> There is no maximum limit for line lengths in files, however, a notional value of about 150 characters is recommended to achieve a good level of readability without horizontal scrolling. Longer lines are permitted if the nature of the code for individual lines requires it and line breaks would have an adverse affect on the final output (such as for mixed PHP/HTML layout files).

Are there guidelines on how to please the drone? :D

richard67: > Are there guidelines on how to please the drone? :D

@pjdevries Yes: The the log file from drone. It can be found when following the "Details" right beside the drone step at the bottom of the PR (might need to use "show all checks" first to expand them) and then in the list on the left hand side select the failed check, in this case it was PHPCS to see the log file at the right side.

pjdevries: @richard67 Thanx for the feed back. Not the answer I was hoping for though :) I would prefer to provide drone acceptable code right away. Is there no documentation for the code style the drone expects?

alikon:
1 small step at a time

alikon (approved):

wilsonge:
> Not sure I understand. State names are hard coded in the models. Why would anyone want to map a filter to a different state name than what the filter is intended for? Isn't it up to the server to define the api and for the client to simply consume it the way it is defined? Or am I misunderstanding what you mean?

No I totally misread the code. I didn't realise the key in the array was the key in the api request 🤦  The code is totally fine as you've done it.

> Consistent attribute names in requests and responses per component.

For now just in the components where there is already filtering change the models (and possibly their relations code) to point to your new expected property names.

> Consistent attribute names across all components.

Aside from conventions it's pretty hard to normalise things across components. I mean I guess you can build some of it out of traits but I think it's probably going to cause more trouble than just having clear conventions.

pjdevries:
@wilsonge So you want to see the same attribute names in `jsonapiView` lists `$fieldsToRenderItem`,  `$fieldsToRenderList and `$relationship` as in the controller's `$queryFilterModelStateMap` list. Is that correct?

PR #34191, opened 2021-05-24
[4] cleanup more of libraries/vendor before release

PhilETaylor:
Code review.

Although this doesnt drastically reduce the size of the final zip files, it does significantly cleanup the libraries/vendor folders removing all kinds of build files that just are not needed on a live Joomla 4.x site. 

The complete list of files this deletes are:

```
libraries/vendor/algo26-matthias/idna-convert/CODE_OF_CONDUCT.md
libraries/vendor/dealerdirect/phpcodesniffer-composer-installer/CODE_OF_CONDUCT.md
libraries/vendor/softcreatr/jsonpath/CODE_OF_CONDUCT.md
libraries/vendor/spomky-labs/cbor-php/CODE_OF_CONDUCT.md
libraries/vendor/algo26-matthias/idna-convert/docker-compose.yml
libraries/vendor/algo26-matthias/idna-convert/phpunit.xml
libraries/vendor/phpunit/php-timer/phpunit.xml
libraries/vendor/phpunit/php-token-stream/phpunit.xml
libraries/vendor/phpunit/php-code-coverage/phpunit.xml
libraries/vendor/phpunit/php-file-iterator/phpunit.xml
libraries/vendor/sebastian/comparator/phpunit.xml
libraries/vendor/sebastian/global-state/phpunit.xml
libraries/vendor/sebastian/object-enumerator/phpunit.xml
libraries/vendor/sebastian/object-reflector/phpunit.xml
libraries/vendor/sebastian/code-unit-reverse-lookup/phpunit.xml
libraries/vendor/sebastian/diff/phpunit.xml
libraries/vendor/sebastian/recursion-context/phpunit.xml
libraries/vendor/sebastian/type/phpunit.xml
libraries/vendor/sebastian/environment/phpunit.xml
libraries/vendor/sebastian/exporter/phpunit.xml
libraries/vendor/wamania/php-stemmer/README.md
libraries/vendor/myclabs/deep-copy/README.md
libraries/vendor/php-cs-fixer/diff/README.md
libraries/vendor/ralouphie/getallheaders/README.md
libraries/vendor/joomla-projects/joomla-browser/README.md
libraries/vendor/codeception/codeception/ext/README.md
libraries/vendor/codeception/codeception/src/Codeception/Util/README.md
libraries/vendor/codeception/codeception/src/Codeception/Step/README.md
libraries/vendor/codeception/codeception/src/Codeception/README.md
libraries/vendor/codeception/codeception/src/Codeception/Lib/README.md
[...]

PR #34289, opened 2021-05-30
[4.0] Update files and folder deletion in script.php to recent changes

richard67:
Pull Request for Issue # .

### Summary of Changes

This pull request (PR) adapts the lists of files and folders to be deleted on update in `script.php` to recent changes:
- Add files being removed from builds which results from PR #34191 - the `/libraries/vendor/bin` folder will not be deleted because it might contain other files added by the site admin.
- Add the 3.9.x update SQL scripts which come from PR #33212 .
- Remove deletion of the `/bin` folder for the same reasons as it is with the `/libraries/vendor/bin` folder.

### Testing Instructions

Code review.

Files which have been removed with PR #34191 in the 4.0-dev branch and the update SQL scripts "3.9.27-2021-04-20.sql" which have been added to staging by PR #33212 are not removed on update if they are present.

Files which have been removed with PR #34191 in the 4.0-dev branch and the update SQL scripts "3.9.27-2021-04-20.sql" which have been added to staging by PR #33212 are removed on update if they are present.

None.


wilsonge:
> The tool from PR #25559 used to create the lists adds folder /libraries/vendor/bin because that folder is removed from the builds since PR #34191 has been merged.

bin folder indeed should be removed - we don't really want to be dealing with the symlinks on different installs. Plus tbh just outright not sure for security we want to come close to exposing these binaries anyhow.

> That was right for those which have been released before, but the ones which were added after the last Beta 7 don't need to be in the list because we don't support updates between or from nightly builds.

This is correct. Don't bother adding them. In the grand scheme of things it really won't matter.

richard67:
~~Not sure if we can delete the "bin" folder if it contains links which are not removed before by the script because they are not part of any package. I have to test that. See also issue #34298 .~~

Update: Sorry, was mixing up "bin" with "libraries/vendor/bin". All ok.

brianteeman:
the question on #34290 is still valid even if you were thinking of the wrong folder

richard67:
> the question on #34290 is still valid even if you were thinking of the wrong folder

@brianteeman That's right. It's just nothing I should change with this PR here.

wilsonge:
Thanks!

richard67:
Thanks too, especially to the testers 😆

PR #34295, opened 2021-05-30
[4.0] Pattern attribute is missing from the number from field

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Adds the missing attribute in the rendered HTML

### Testing Instructions

Review, it's a no brainer


Quy:
From developer.mozilla.org

> Do not support use of the pattern attribute for making entered values conform to a specific regex pattern.

> The rationale for this is that number inputs won't be valid if they contain anything except numbers, and you can constrain the minimum and maximum number of valid digits using the min and max attributes (as explained above).

dgrammatiko:
> The rationale for this is that number inputs won't be valid if they contain anything except numbers, and you can constrain the minimum and maximum number of valid digits using the min and max attributes (as explained above).

I can type `bad` in Safari and the form will submit fine but the value is not numeric. Specs are not always followed by browsers

PR #34302, opened 2021-05-30
[4.0] Point to J4 translations list

wilsonge:
Pull Request for Issue #21806.

### Summary of Changes
This sets up the language files in J4 to point to a J4 language server.

NOTE: This file doesn't exist yet - it will in the next 24 hours. this should NOT be merged until that file exists

### Testing Instructions
Ensure Joomla is updated to point to the new update list.

None


administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-05-30.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE `#__update_sites`
+   SET `location` = 'https://update.joomla.org/language/translationlist_4.xml'
+ WHERE `update_site_id` = 2;
```
richard67: You could have added that to the existing 4.0.0-2021-05-21.sql script as we don't support updating from nightlies.

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-05-30.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE "#__update_sites"
+   SET "location" = 'https://update.joomla.org/language/translationlist_4.xml'
+ WHERE "update_site_id" = 2;
```
richard67: You could have added that to the existing 4.0.0-2021-05-21.sql script as we don't support updating from nightlies.

administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-05-30.sql
richard67: And maybe renamed it and add a comment to it with the new script name like we have the other comments after we combined the SQL.

But that's just a nice to have, the PR here is good as it is.

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-05-30.sql
richard67: See also additional comment in my other review comment above.

wilsonge:
@richard67 any chance you can PR to my branch. I've completely destroyed my PHP setup last night trying to update some homebrew stuff :(

richard67:
For the update SQL, can we be sure that the ID of the update site which we have to change is always 2, also on installation with a longer update history?

richard67:
On my home page the ID is 3.

alikon:
use the `location` in the where condition instead of `update_site_id`

richard67:
@alikon You are right. I've just fixed that.

Issue #34248, opened 2021-05-27
[4.0] global menu options


brianteeman:
### Step 1
Go to the global options for a menu item and set a value for all 4 options
![chrome_2021-05-27_18-32-30](https://user-images.githubusercontent.com/1296369/119875499-b914bc00-bf1e-11eb-9df5-48b3e524c3bc.png)

### Step 2
Create a new menu item and check the settings in the page display tab - expected behaviour would be to see the values from step 1

![chrome_2021-05-27_18-34-03](https://user-images.githubusercontent.com/1296369/119875638-e5303d00-bf1e-11eb-9130-6f4fc3d99e73.png)

###  Step 3
Now we know that the page heading and page class global settings are ignored - what about the browser title that is at least displayed in the menu item options. Nope thats ignored

![chrome_2021-05-27_18-38-32](https://user-images.githubusercontent.com/1296369/119875731-0002b180-bf1f-11eb-9d26-666e748c5782.png)

### Summary
The only one of the four options that actually does anything is the third one "Show Page Heading" but even that doesnt really doing anything as it is meaningless without a value for  "Page Heading"

### Conclusion
There is zero point in having any of these global options for a menu item as they dont work as their is no code to use them and they can be removed - _**unless I have missed something**_

tonypartridge:
They used to work in J3, so I’d expect them to work in J4... we have actually be using them for meta management and page H1’s. 

And page class to define custom classes within the theme... 

Wouldn’t it make more sense to add the support rather than remove it?

brianteeman:
are you sure they work in j3? The broken behaviour seems identical to me.

chmst:
They don't work in J3 either, at least in content. @tonypartridge magine that all sites have the same title? a11y and seo ....

chmst:
Finally found what happens and where. 

Before a PR I would like to get your opinion if we can remove the site title form gobal params. Equal title for all pages is an accessibility issue if it works. @brianteeman 
 <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34248">issues.joomla.org/tracker/joomla-cms/34248</a>.</sub>

brianteeman:
I agree equal titles is wrong. Hard to confirm that this param creates that though as I dont see where it can be used

chmst:
[...]

PR #34311, opened 2021-05-31
Global menu options

chmst:
Pull Request for Issue #34248 .

### Summary of Changes
Fix missing useglobal attributes to site display params.
Remove site_title form global params

Fix missing param pageclass_sfx.

### Testing Instructions
Testing is playing with global setting of menu options and setting of params in menuItems. 

grafik

- The site title (in meta data of the head)
- The site title on the page if it is not hidden
- The pageclass in the source code
- Note: not all views use the pageclass_sfx. This is not in scope of this PR.

Settings of global params is ignored. see #34248 .

If global Params are set for Menu and menu Item has no own param for this
then the global param is used

if a menu Item has set own Param,
then own Params are used


brianteeman:
@chmst I understand that this code will make the options work. It confirms what I reported that they dont currently work as there is no code to use them and the same is true in J3. So as no one reported an issue lets remove them instead of fixing them as they cant have been used. Am I missing something?

ghost:
@chmst no test cause its a draft, right?

drmenzelit:
> @chmst no test cause its a draft, right?

right

chmst:
Sorry I forgot to activate this PR. @sandramay0905 , @brianteeman a test would be highly appreciated, I think it is now as it should be.

ghost:
> The site title (in meta data of the head)

You mean setting in `Global Configuration` > `Site` > `SEO` > `Site Name in Page Titles` to `After` or `Before`?

> The site title on the page if it is not hidden

You mean setting in `Global Configuration` > `Site` > `SEO` > `Site Name in Page Titles` to `No`?

> Note: not all views use the pageclass_sfx. This is not in scope of this PR.

I test on a category-blog-view and can't find `myGlobalPageClass` which i set in menu-options so i guess this is one of the views they don't use pageclass_sfx. Can you please say which view use this sfx?

brianteeman:
page class - have not been able to get it to be used anywhere

page title - it works but I dont see the use case for this - ever

show page title - works as before

As stated in the original report #3428 the two non working options have not worked for years (and no one commented). So surely the fix is to remove them instead of wasting time trying to fix.

brianteeman:
I have tested this item :red_circle: unsuccessfully on e203d389d910e847206ac92837042eb2adc97780<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34311">issues.joomla.org/tracker/joomla-cms/34311</a>.</sub>

chmst:
Pageclass_sfx is used on single item pages only.

PR #34313, opened 2021-05-31
[4][webservices]PATCH fixes

alikon:
Pull Request for Issue #34269 .

### Summary of Changes
return the record in case of success
return a 404 in case of not found

### Testing Instructions

- PATCH an existing article
- PATCH an unexisting article

### DIFF
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -372,7 +372,7 @@ class ApiController extends BaseController
 
 		$recordId = $this->input->getInt('id');
 
-		if (!$recordId)
+		if ($recordId && !$table->load($recordId))
 		{
 			throw new Exception\ResourceNotFound(Text::_('JLIB_APPLICATION_ERROR_RECORD'), 404);
 		}
@@ -394,6 +394,7 @@ class ApiController extends BaseController
 		}
 
 		$this->save($recordId);
+		$this->displayItem($recordId);
 
 		return $this;
 	}


wilsonge:
In terms of what it does JSON API says the following: 
> 200 OK
> If a server accepts an update but also changes the resource(s) in ways other than those specified by the request (for example, updating the updated-at attribute or a computed sha), it MUST return a 200 OK response. The response document MUST include a representation of the updated resource(s) as if a GET request was made to the request URL.
>
> A server MUST return a 200 OK status code if an update is successful, the client’s current fields remain up to date, and the server responds only with top-level meta data. In this case the server MUST NOT include a representation of the updated resource(s).
> 
> 204 No Content
> If an update is successful and the server doesn’t update any fields besides those provided, the server MUST return either a 200 OK status code and response document (as described above) or a 204 No Content status code with no response document.

So I think if we decide to take the convention that if the document doesn't change we do return the 200 with a body rather than try and do the 204 this should be fine and ready for testing.

libraries/src/MVC/Controller/ApiController.php
```diff
@@ -372,7 +372,7 @@ public function edit()
 
 		$recordId = $this->input->getInt('id');
 
-		if (!$recordId)
+		if ($recordId && !$table->load($recordId))
```
wilsonge, line 375: I think we can omit this change. This will be part of the other PATCH PR (where we load the item to merge the data before saving).

alikon: no it's needed i think cause if you PATCH  a non existent you got a 500 "Internal server error" instead of 404 Resource not found
 anyway i'll close this one and i'll merge in #34274

alikon:
closing this  1 better to have 1 pr only for the same file

PR #34337, opened 2021-06-01
[4] Fix fatal error by calling correct class name

PhilETaylor:
Code review. 

`Route::getNewsfeedRoute` is invalid 

The method `getNewsfeedRoute` is in the `RouteHelper.php` file in the same namespace


wilsonge:
Thanks!

PR #34338, opened 2021-06-01
[4] Prevent fatal error by importing \Exception

PhilETaylor:
Code review

`Exception` is used 4 times in this file, but its undefined. 

three in docblocks, no worries, but undefined as not in the same namespace. 

One of those uses is in a try/catch block on line 102, therefore to prevent that not working we need the class imported.


wilsonge:
Thanks!

PR #34339, opened 2021-06-01
[4] Prevent fatal error by importing use CategoryFactoryInterface

PhilETaylor:
Code review

`CategoryFactoryInterface` is used in the constructor however that class is not defined as this file is in the namespace `Joomla\Component\Tags\Site\Service` and `CategoryFactoryInterface` is in the namespace `Joomla\CMS\Categories`

So we need to import the class to prevent fatal errors if someone ever used the constructor and provided a valid value for `$categoryFactory`


wilsonge:
Thanks!

Issue #34263, opened 2021-05-28
[4] Unhandled exceptions on sending mails + Mail Test AJAX error.


PhilETaylor:
### Steps to reproduce the issue

Issue Forked from https://github.com/joomla/joomla-cms/pull/34233#issuecomment-850529623

It is possible to generate an uncaught exception in Mail Templates when `path::check` detects snooping 

This issue is two parted

1) The Joomla global configuration test mail button - when pressed, an exception occurs that is unhandled gives the wrong error, when it should tell you the exception message

<img width="785" alt="Screenshot 2021-05-28 at 17 11 46" src="https://user-images.githubusercontent.com/400092/120016231-3ce0ae00-bfdc-11eb-9fa7-de55276e0f57.png">

2) More generically - any Mail sending with Mail Templates could throw an exception, and that is unhandled within the mail sending, so the code could fatal error on any mail being sent when an exception is thrown (mainly from `Path::check`- I dont have steps to replicate this but I think my logic is right. 

### Expected result

Exception is handled and returned 

### Actual result

1) An error has occurred while fetching the JSON data: HTTP 502 status code. Bad Gateway" error message shown on the page

2) PHP Fatal error.

Fedik:
As far as I remember such JSON error happen for broken response, in your case `502 Bad gateway`, server just crashes.
It not Exception.

Fedik:
like this:
![image](https://user-images.githubusercontent.com/1568198/120031099-fe111f00-c000-11eb-867e-802131982e23.png)

PhilETaylor:
yup - well thats bad too haha... The exception is being thrown as I debugged that far, but you are right I did not debug all the way to the response. 

This should be fixed though, its a valid bug

Fedik:
Need a way to replicate it ;) 

We have global exception handler, that in theory should catch everything. Maybe server just decided to die :smile:

PhilETaylor:
Ive not had much time this week sorry, i should have more time over the weekend and next week to catch up with this.

Fedik:
Take your time, nothing really critical here ;)

dgrammatiko:
What's the endpoint for the Ajax request?

Obviously there should be a try/catch when trying to parse a JSON response and pretty sure that there are many instances that missing it

Fedik:
> Obviously there should be a try/catch

There is try/catch, and then "catch" shows error which @PhilETaylor posted.
It not JS, it server side error.

[...]

PR #34340, opened 2021-06-01
[4] When using path::clean, a security method, correctly throw a useable exception. 

PhilETaylor:
Pull Request for Issue #34263

### Summary of Changes

When using `Path::clean()` method, with an invalid path (Like `../../../../../../`) it is designed to throw an exception. 

The problem is that someone used code = 20 in the exception.

If your code calls the code below, and doesnt catch the exception then your website will not show an error message, it will show a 502 Bad Gateway (on nginx)

```php
Path::clean('../../../../../../');
```

When you have a php-fpm set up, where the web server passes the request "upstream" to the "gateway" (i.e php-fpm running somewhere, or somehow else) then PHP-FPM will return error code `20` when this exception is thrown causing an error like:

```
webserver_1  | 2021/06/01 15:10:43 [error] 31#31: *109 upstream sent invalid status "20" while reading response header from upstream, client: 172.19.0.1, server: , request: "GET /administrator/index.php HTTP/1.1", upstream: "fastcgi://172.19.0.6:9000", host: "127.0.0.1:4444"
```

### Testing Instructions

If your code calls the code below, and doesnt catch the exception then your website will not show an error message, it will show a 502 Bad Gateway (on nginx)

```php
Path::clean('../../../../../../');
```

 502 Bad Gateway 

 Correctly displayed Exception that can be handled 

// @Fedik @dgrammatiko

### DIFF
--- a/libraries/src/Filesystem/Path.php
+++ b/libraries/src/Filesystem/Path.php
@@ -212,8 +212,7 @@ class Path
 				sprintf(
 					'%s() - $path is not a string',
 					__METHOD__
-				),
-				20
+				)
 			);
 		}
 
--- a/libraries/src/Filesystem/Path.php
+++ b/libraries/src/Filesystem/Path.php
@@ -173,8 +173,7 @@ class Path
 				sprintf(
 					'%s() - Use of relative paths not permitted',
 					__METHOD__
-				),
-				20
+				)
 			);
 		}
 
@@ -187,8 +186,7 @@ class Path
 					'%1$s() - Snooping out of bounds @ %2$s',
 					__METHOD__,
 					$path
-				),
-				20
+				)
 			);
 		}
 


Fedik:
I would change to 418, but removing is okay

PhilETaylor:
`418 I'm a teapot` is a HTTP Error code, not a PHP Exception code ;-)

Fedik:
yes, that in result will be HTTP response status code ;)

PhilETaylor:
On reflection of your joke with HTTP Codes, Im not sure this PR is the right approach... Im thinking ow maybe something in the Symfony error handler is converting a User-Defined Exception code to a HTTP Response code... 

Let me dig into Symfony Error Handler more...

Fedik:
I have tested this item :white_check_mark: successfully on cf1b76446a2fa7975c91d2d9c442d36e39313a69<br /><br />on review<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34340">issues.joomla.org/tracker/joomla-cms/34340</a>.</sub>

Fedik:
> maybe something in the Symfony error handler is converting a User-Defined PHP Exception code to a HTTP Response code.

Anything can be, 
but I never seen such, most of scripts just use an exception code directly for HTTP response

PhilETaylor:
Confirmed... in the same nginx/php-fpm set up running, using code:

```
<?php
error_reporting(E_ALL);
ini_set('display_errors',1);

throw new Exception('asd',20);
```

I correctly get

```
Fatal error: Uncaught Exception: asd in /application/ex.php:5 Stack trace: #0 {main} thrown in /application/ex.php on line 5
```

So something in Joomla deep in the unhandled error handler is causing a HTTP Response of code 20... 

This needs researching and fixing.... (Cause we cannot trust 3PD to catch Exceptions because even Joomla doesnt ;-()

PhilETaylor:
I must have had my coffee.... 

I was right

> On reflection of your joke with HTTP Codes, Im not sure this PR is the right approach... Im thinking now maybe something in the Symfony error handler is converting a User-Defined PHP Exception code to a HTTP Response code...



https://github.com/symfony/symfony/blob/0d3befa984bab8a1ef9304dfced1d9d9b8fc8d2e/src/Symfony/Component/ErrorHandler/Exception/FlattenException.php#L78

Fedik:
if you curious, here is where header sets, for an exception
https://github.com/joomla/joomla-cms/blob/3a5984d03366f707ef842ee48051af194f68848f/libraries/src/Document/ErrorDocument.php#L117-L132

it should be 500 if exception code is less than 400, that strange

PhilETaylor:
I dont think it goes anywhere near that, especially as the status of 20 is less than 400 and so that code would reset it to 500... Its the Symfony error handler (The red screen of death)... I need more coffee before going that low into the code again,[ last time I barely made it out alive](https://issues.joomla.org/tracker/joomla-cms/30056)...

dgrammatiko:
One more: https://github.com/joomla/joomla-cms/blob/cf1b76446a2fa7975c91d2d9c442d36e39313a69/libraries/src/Filesystem/Path.php#L216 ??

PhilETaylor:
yeah that one probably needs to go too, but let me look under the red screen of death and see if we (Joomla) can handle it correctly first.

richard67:
Does this PR solve issue #34263 so it can be closed?

PhilETaylor:
Please leave open for now, Im using it as a reminder to check that all Mail Sending in Joomla is catching newly added Exceptions that can be raised in MailTemplates.php and below :(

wilsonge:
I suspect 20 comes from the old JError days fwiw and the status codes never got changed.

PhilETaylor:
This can be merged now...

wilsonge:
Thanks!

PR #34343, opened 2021-06-01
[4] fix undefined namespace autocompletion in IDEs

PhilETaylor:
Code review

A bunch of classes in comments that are not correctly namespace and so generate warnings/errors in IDE's and other tools about the namespaces and classes not existing.


wilsonge:
Thanks!

Issue #34351, opened 2021-06-02
Joomla Update pre-update checks for third party extensions are wrong


nikosdion:
### Steps to reproduce the issue

* Install any extension compatible with Joomla 4, such as the latest versions of JCE.
* Edit Version.php to fake the current Joomla version as 4.0.0-beta8 (which will allow Joomla Update to find an update to the CMS version 4.0.0-rc1, the current version at the time of this writing).
* Go to System, Update, Extensions and click on Check for Updates to verify Joomla has the latest extension update information.
* Go to System, Update, Joomla.
* Look at the Pre-update Check results.

### Expected result

The third party extension (JCE) which is [definitely compatible with Joomla 4.0, per its XML update source file,](https://cdn.joomlacontenteditor.net/updates/xml/editor/pkg_jce_pro.xml) appears as such.

### Actual result

You get a misleading and scary message titled “JCE Pro Extension Package  Potentially Serious Problem.  More Information.” which says that you need to update the extension and if you don't the update will fail.

Moreover, when you click on More Detail you get even more confusing and wrong information. You are told that the currently installed version is the latest version installed and it has "No compatibility information" with 4.0.0-rc1 even though the XML update source clearly states that 4.0 is an accepted target.

### System information (as much as possible)

Completely irrelevant. This is a JavaScript issue affecting all environments.

### Additional comments

I had reported this issue back in the Alpha 10 or Alpha 12 days (it was a long time ago...). I was told that my update source was wrong — even though it was valid, marked the extensions as compatible with Joomla 4 and I could prove it by the fact that my extension updates worked in Joomla 4. This time I am filing the same bug report using a non-Akeeba extension, proving the point that the bug does exist, has been there for at least two years and needs to be fixed. Sorry it took so long, no developer but yours truly would start supporting Joomla 4 before it even hit beta.

Here's how I debugged this issue and how to fix it.

Opening your browser's dev tools and going to the Network tab you will see that the AJAX URL is glaringly wrong: `http://localhost/joomla4/administrator/index.php?option=com_joomlaupdate&task=update.fetchextensioncompatibility&joomla-target-version=4.0.0-rc1joomla-current-version=4.0.0-rc1extension-version=2.9.7&extension-id=317`

Have you spotted the missing ampersands yet? So, let's see where these come from.

[...]

PR #34354, opened 2021-06-02
Joomla Update pre-update checks for third party extensions are wrong

nikosdion:
Pull Request for Issue #34351 .

### Summary of Changes

Added the missing ampersands in the Joomla Update Javascript, fixing the Pre-update checks for third party extensions.

### Testing Instructions

* Install any extension compatible with Joomla 4, such as the latest versions of JCE or Akeeba Backup.
* Edit libraries/src/Version.php to fake the current Joomla version as 4.0.0-beta8 (which will allow Joomla Update to find an update to the CMS version 4.0.0-rc1, the current version at the time of this writing).
* Go to System, Update, Extensions and click on Check for Updates to verify Joomla has the latest extension update information.
* Go to System, Update, Joomla.
* Look at the Pre-update Check results.

You are told that the third party extension you installed needs to be updated because it's not compatible with Joomla 4. When you click More Detail you see that the latest version and the currently installed one are the same but there is "no compatibility information" which is obviously wrong.

Make sure to Follow the **last three steps** of the instructions.

Now you see that the third party extension is listed as No Update Required. Woo-hoo!

None required.

### DIFF
--- a/build/media_source/com_joomlaupdate/js/default.es6.js
+++ b/build/media_source/com_joomlaupdate/js/default.es6.js
@@ -302,8 +302,8 @@ Joomla = window.Joomla || {};
     Joomla.request({
       url: `${PreUpdateChecker.config.serverUrl
       }&joomla-target-version=${encodeURIComponent(PreUpdateChecker.joomlaTargetVersion)
-      }joomla-current-version=${PreUpdateChecker.joomlaCurrentVersion
-      }extension-version=${node.getAttribute('data-extension-current-version')
+      }&joomla-current-version=${PreUpdateChecker.joomlaCurrentVersion
+      }&extension-version=${node.getAttribute('data-extension-current-version')
       }&extension-id=${encodeURIComponent(node.getAttribute('data-extension-id'))}`,
       onSuccess(data) {
         const response = JSON.parse(data);


joeforjoomla:
This works and fix the bug with ampersands... but what about the scary message “Extension XXX Potentially Serious Problem. More Information.”? Seriously we want that everytime that there is a Joomla 4 update and a system plugin has no update informations, the user will be warned about potentially breaking the Joomla site or update? 
Keep in mind that there are many extensions developers that provide plugins bundled as part of a main component which is the one subject of the updater.

zero-24:
Can we please branch this question out to its own issue? This PR here fixes an valid issue to be merged after tests. The question mentiond by you is also the case for 3.10 so i would propose this to be moved to its own issue. I'm happy to explain the reasons and discuss alternative solutions when we have better proposals. Thanks @joeforjoomla

nikosdion:
@joeforjoomla I can address your issue to the extent it touches the scope of my PR. When you install a `package` type extensions all of its extensions (e.g. component, plugins, modules, ...) are "assigned" to the package ID in the `#__extensions` table, see the `package_id` column. According to my tests with JCE and my own software these sub-extensions do **not** appear in the pre-update check. So, by fixing the problem this PR addresses any canonical 3PD Joomla extension installation should appear correctly in the pre-update checks.

wilsonge:
Thankyou @nikosdion - sorry must have mucked this one up in the conflicts from 3.10

nikosdion:
@wilsonge It's OK. My original report was two pronged, one question being about sub-extensions (I had the same concern as @joeforjoomla). The other one was about the actual checks which originally were brushed off as a problem with the extension XML update file, then there was an oblique reference to a possible issue with the URL. It ended up not being fixed, I guess? At least now I could test with JCE, not just my stuff, to make sure that the problem is definitely not between this chair and keyboard ;)

I will probably come back with more small bugs. I am in the process of converting Admin Tools to core MVC and figuring out the transition from FOF-based to core MVC-based extensions. I end up hitting all the weird things in extensions installation and updates nobody could have reasonably tested during the alpha and beta.

I currently have an issue with an extension version X listing update site A and extension version Y listing site B resulting in both sites being kept. I will be diving into that in a week or so.

joeforjoomla:
ewdw

joeforjoomla:
@nikosdion there is still a javascript error after this fix (see image)

It seems that there are no elements having id 'plg_xxx' so the remove function fails, i guess that line should be removed.

image

Issue #34358, opened 2021-06-02
[4.0] Non default path's causes problems using the media manager


pjdevries:
### Steps to reproduce the issue

Change both path's of the Media Manager configration, i.e. 'Path to Files Folder' and  'Path to Images Folder'.

Open Media Manager.

### Expected result

A list of files in folder 'Path to Files Folder', if there are any, or an empty folder otherwise.

### Actual result

The result depends on proper configuration of the `FileSystem - Local` plugin. If no directory is chosen, which is the default, the plugin returns `[{"directory": "images"}]` and the Media Manager shows nothing.

### System information (as much as possible)

A fresh Joomla! 4.0 installation, or at least with unaltered  `FileSystem - Local` plugin configuration.

### Additional comments

My suggestion is to return Media Manager's 'Path to Files Folder' setting from `FileSystem - Local` as a default and _not_ `[{"directory": "images"}]`.

richard67:
@pjdevries Will that be solved, too, with your PR #34314 ? If so, this issue can be closed.

pjdevries:
@richard67 Unfortunately this will not be solved with PR  #34314. It's a minor fix though, which I don't mind doing if that's o.k.

richard67:
> ...  which I don't mind doing if that's o.k.

@pjdevries You mean you wanna fix it with another PR? That would be very o.k.

pjdevries:
@richard67 Yeah, that's what I meant. I'll do it right away.

alikon:
please test #34364

PR #34364, opened 2021-06-02
[4.0] Return sensible default from 'FileSystem - Local' plugin.

pjdevries:
Pull Request for Issue #34358 .

### Summary of Changes

Changed the way directory settings are obtained in the `FileSystem - Local` plugin. If mo directory has been explicitly chosen yet, return Media Manager's 'Path to 'Path to Files Folder' setting.

### Testing Instructions

Make sure you have a fresh Joomla! 4.0 installation, or at least one with an unaltered `FileSystem - Local` plugin configuration.

Change both path's of the Media Manager configuration, i.e. 'Path to Files Folder' and 'Path to Images Folder'.

Open Media Manager.

**Beware of caching when fiddling with the Media Manager and its path settings!** (see also #34239)

The result depends on proper configuration of the FileSystem - Local plugin. If no directory is chosen, which is the default, the plugin returns [{"directory": "images"}] and the Media Manager shows nothing.

A list of files in folder 'Path to Files Folder', if there are any, or an empty folder otherwise.

It's probably a good thing to document this somewhere, preventing (new) users to get confused.


plugins/filesystem/local/local.php
```diff
@@ -79,30 +83,56 @@ public function getDisplayName()
...
 
 		return $adapters;
 	}
+
+	/**
+	 * Return plugin directory paramater settings or sensible default
+	 *
+	 * @return array
+	 *
+	 * @since 4.0
```
Quy:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

pjdevries: Thanx for reviewing. I didn't encounter that `__DEPLOY_VERSION__` anywhere in the code yet. What's wrong with `4.0` if that is the version it was introduced in first?

brianteeman: It is shorthand that we encourage people to use when submitting a pull request. When it is merged and a release is made it is automatically replaced with the exact release number. This is just to prevent you having to change the number if it doesn't make the intended release. Sometimes things take longer than expected or an unexpected release is made.

pjdevries: @brianteeman Makes sense. From now on, I'll use `__DEPLOY_VERSION__` instead of a fixed version number for new items with an `@since` doc element. And I'll change it in this PR :)

plugins/filesystem/local/local.php
```diff
@@ -79,30 +83,56 @@ public function getDisplayName()
 	public function getAdapters()
 	{
 		$adapters = [];
-		$directories = $this->params->get('directories', '[{"directory": "images"}]');
```
Fedik, line 82: `'[{"directory": "images"}]'` is not a random thing, but default field value of the "directories" field 
https://github.com/joomla/joomla-cms/blob/df39c83d4923a3c06a889cfc32b5461af663f16b/plugins/filesystem/local/local.xml#L31-L34

Fedik: well, I have to read first :smile:

Fedik:
> If mo directory has been explicitly chosen yet, return Media Manager's 'Path to 'Path to Files Folder' setting.

I am doubt here, the plugin should use own options. There may be another filesystem plugins.

> If no directory is chosen, which is the default, the plugin returns [{"directory": "images"}] and the Media Manager shows nothing.

I not checked how it connected, but probably the folder should come from filesystem plugins only.

The path option in the new Media Manager configuration seems does not make sense anymore.

pjdevries:
> > If mo directory has been explicitly chosen yet, return Media Manager's 'Path to 'Path to Files Folder' setting.
> 
> I am doubt here, the plugin should use own options. There may be another filesystem plugins.

If I understand it correctly:
- The `Local` plugin supports local files.
- The Media Manager also supports local files.

So I think it makes sense to use Media Manager's path setting as a default for the `Local` plugin, 

> > If no directory is chosen, which is the default, the plugin returns [{"directory": "images"}] and the Media Manager shows nothing.
> 
> I not checked how it connected, but probably the folder should come from filesystem plugins only.

I _did_ check. After changing paths in the Media Manager, it did not work anymore until I selected the proper directory in the `Local` plugin.

> The path option in the new Media Manager configuration seems does not make sense anymore.

That may well be, but it is there and can be used. As long as that's the case, I think it must behave in a predictable manner.

Fedik:
> If I understand it correctly:
> The Local plugin supports local files.
> The Media Manager also supports local files.

The thing that  Media Manager supports depend from enabled FileSystem plugins.
On its own it does not support anything.

richard67:
I agree with @Fedik . You can install a filesystem plugin e.g. for using Dropbox, and then you can set paths to files and images folders in the media manager options. With the change which is suggested here, the local filesystem plug-in would use the path to a folder on Dropbox as directory, and that would be wrong.

pjdevries:
@Fedik @richard67 I'm not sure I understand. Is it possible to have multiple active filesystem plugins? If so, how does the Media Manager distinguish between them, i.e. know which one to use?

richard67:
@pjdevries Yes, the media manager can use multiple different filesystems. Each one is a base folder in the media manager. See the testing instructions in this PR: #33724 . In media manager you see 2 parent folders where you can navigate to children, the local "Images" and the "Dropbox" folder when doing as described in that PR.

richard67:
P.S.: Same when e.g. inserting an image into an article: You can use images from the local filesystem as well as from Dropbox when using that dropbox plugin (which is not ready yet and was developed just for testing, but I'm sure sooner or later you will find more such plugins in the extensions directoy).

plugins/filesystem/local/local.php
```diff
@@ -1,18 +1,21 @@
 <?php
 /**
- * @package     Joomla.Plugin
- * @subpackage  FileSystem.Local
+ * @package         Joomla.Plugin
+ * @subpackage      FileSystem.Local
  *
  * @copyright   (C) 2017 Open Source Matters, Inc. <https://www.joomla.org>
- * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ * @license         GNU General Public License version 2 or later; see
```
brianteeman, line 7: please revert this change

pjdevries: > please revert this change

Not so hasty. The issue still exists. If this isn't the right solution, let's think of a better one.

But first some thoughts.

Thanks to @richard67 explanation, I now have a better understanding. I also better understand whu @Fedik says the path option in the new Media Manager configuration is obsolete. I not only tend to agree, but it makes things confusing. Maybe remove them, in which case the original 'images' as the default return value from the `Local` filesystem plugin makes much more sense.

On the other hand, as long as those paths in the Media Manager configration are there, why not use them the way I proposed. Is it that much different than the original code, returning 'images' as a default? After all, it's just a default. On a fresh Joomla! 4 installation, only the `Local` filesystem plugin is enabled by default but not yet explicitly configured. Only under those circumstances will the default be returned and returning the Media Manager path seems just as good (or bad) as 'images' :)

But I'm happy to change the default to something more sensible. Any suggestions?

Fedik:
I think we need to consider removing `file_path` and `image_path` from Media config.
I do not see a reason why we need them, but I may be wrong.

pjdevries:
> I think we need to consider removing `file_path` and `image_path` from Media config.
> I do not see a reason why we need them, but I may be wrong.

From what I understand from @richard67's explanation, I think you are right. If Media Manager no longer uses those paths but only the adapter settings, the issue that resulted in this PR vanishes automatically. A default return value of 'images' from the `Local` filesystem plugin, also makes sense then, making this PR unnecessary and @brianteeman a happy dude :)

pjdevries:
@richard67 
> See the testing instructions in this PR: #33724

Seemed like a good idea to play a bit with multiple adapters, but I get this known error: 'Call to undefined method GuzzleHttp\Utils::chooseHandler()'.

richard67:
> Seemed like a good idea to play a bit with multiple adapters, but I get this known error: 'Call to undefined method GuzzleHttp\Utils::chooseHandler()'.

@pjdevries That's an issue with that plugin which was a quick made thing just for testing. You can find the solution here: https://github.com/joomla/joomla-cms/pull/33724#issuecomment-847081056

pjdevries:
> That's an issue with that plugin which was a quick made thing just for testing. You can find the solution here: [#33724 (comment)](https://github.com/joomla/joomla-cms/pull/33724#issuecomment-847081056)

Yes, works! Cool!

plugins/filesystem/local/local.php
pjdevries: > Not so hasty. The issue still exists. If this isn't the right solution, let's think of a better one.

Apologies to @brianteeman for not looking more closely what he was referring to.

Issue #32820, opened 2021-03-23
Users page is too wide


mariantanase:
### Steps to reproduce the issue
1. Fresh Joomla 4.0.0 beta 8 dev
2. In admin backend, go to "Users". The details page user list is too wide and need to make horizontal scroll.

### Actual result
![screen shot 2021-03-23 at 09 29 39](https://issues.joomla.org/uploads/1/58584d9253ac28a44a31b4dddd191214.png)

### System information (as much as possible)
PHP 7.4.13
Screen resolution: 1280 x 1024

PhilETaylor:
Gets even worse if you enable the 2FA features as another column is added. 

Discussions:
https://github.com/joomla/joomla-cms/discussions/31915
https://github.com/joomla/joomla-cms/discussions/32549

sksuryan:
Interesting, because I worked on a similar issue with the mail template, but I'm not able to reproduce this one.

Yash-g17:
@mariantanase  I'd love to solve this  can you please assign this to me ?

PhilETaylor:
The Joomla Project doesn't assign issues to contributors in that way. Just work on it and submit a PR :-) 

https://github.com/joomla/joomla-cms/blob/staging/.github/CONTRIBUTING.md

Yash-g17:
@PhilETaylor How do I figure out which file this issue is pointing towards ?

PhilETaylor:
There appears to be a certain breakpoint where the horizontal scroll takes place. 

https://user-images.githubusercontent.com/400092/112767389-68f8a680-900e-11eb-9907-9c0b0d10d5a1.mp4

meinhoonharsh:
> @PhilETaylor How do I figure out which file this issue is pointing towards ?

@PhilETaylor I am also wants to know, How can we find  which we have to edit to make changes in something we want, I got css, scss and js files through the get parameters in the URL, but unable to find the HTML for anything

PhilETaylor:
Sorry, Its not for me to teach you the Joomla architecture or how to use the search :) 

Start your quest in the 4.0-dev branch here: https://github.com/joomla/joomla-cms/tree/4.0-dev/administrator/components/com_users/tmpl/users

meinhoonharsh:
@PhilETaylor Thank you, It helped alot 😁

ghost:
@Yash-g17  
> How do I figure out which file this issue is pointing towards ?

[...]

PR #34387, opened 2021-06-03
[4.0] Views UI overflow fixed

ThisIsFaar:
Pull Request for Issue #32820 .

### Summary of Changes

manage users section is fixed, now any number of columns can be added without worsing UI.

### Before Fix

Before

### After Fix

After

### DIFF
--- a/administrator/components/com_privacy/src/View/Request/HtmlView.php
+++ b/administrator/components/com_privacy/src/View/Request/HtmlView.php
@@ -120,7 +120,7 @@ class HtmlView extends BaseHtmlView
 	 */
 	protected function addToolbar()
 	{
-		Factory::getApplication()->set('hidemainmenu', true);
+		Factory::getApplication()->input->set('hidemainmenu', true);
 
 		// Set the title and toolbar based on the layout
 		if ($this->getLayout() === 'edit')


ghost:
I have tested this item :white_check_mark: successfully on be23d924c8765e1bc6cc54a6f96796c8dca0ce9e<br /><br />Works also on list-views for articles, categories.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34387">issues.joomla.org/tracker/joomla-cms/34387</a>.</sub>

brianteeman:
This applies to all views not just users - is that intentional ?

Please try to do this without using inline styles

Abernyte-Git:
I can still get one break point where the horizontal scroll bar appears.  Is this expected behaviour? If yes I can agree a successful test.
screen shot 2021-06-04 at 07 25 14<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34387">issues.joomla.org/tracker/joomla-cms/34387</a>.</sub>

ThisIsFaar:
> This applies to all views, not just users - is that intentional ?
> 
> Please try to do this without using inline styles

**I also thought the same but then I applied to all the views because,**

**The Problem:** currently manage users fields have lots of columns which was the fault of overflow as all the columns will not fit under the parent div.

So Why I don't apply this fix only for the manage users part?
**here's the reason:**
1. maybe in future updates if any other views also got more columns then this issue will come again in the story.
2. not only in future updates, let's just take an example If any user using some plugin that will add/update columns for the functionality/required purpose to any of the views, then this issue will also currently faced by those users.

and about Inline CSS, I think that for just a single property Inline is the best, or else I have to declare and apply with class or id

ThisIsFaar:
> I can still get one break point where the horizontal scroll bar appears. Is this expected behaviour? If yes I can agree a successful test.
> screen shot 2021-06-04 at 07 25 14
> 
> This comment was created with the [J!Tracker Application](https://github.com/joomla/jissues) at [issues.joomla.org/tracker/joomla-cms/34387](https://issues.joomla.org/tracker/joomla-cms/34387).

can you please describe a bit more, means draw some arrows for your issue on the ss, because I don't get your point.

dgrammatiko:
Just add the `overflow: scroll;` here: https://github.com/joomla/joomla-cms/blob/1d0510539f501e79b3836e1a548430e615c7a53a/administrator/templates/atum/scss/blocks/_global.scss#L118-L122

brianteeman:
inline styles are wrong for many reasons. not least of which is that they are very difficult to override

ThisIsFaar:
> inline styles are wrong for many reasons. not least of which is that they are very difficult to override

Yes, I totally agree. thanks for reminding me.

ThisIsFaar:
> Just add the `overflow: scroll;` here:
> 
> https://github.com/joomla/joomla-cms/blob/1d0510539f501e79b3836e1a548430e615c7a53a/administrator/templates/atum/scss/blocks/_global.scss#L118-L122

hello @dgrammatiko, @brianteeman  thank you for the help, I have done the changes, now I have to make another PR, or I can edit this PR. sry, it sounds silly but I am currently in the learning stage of the PullRequest Scenario.

drmenzelit:
You can make new commits to this PR directly. Maybe you should change the title of the PR if you are changing all views and not only users.

ThisIsFaar:
@drmenzelit, @brianteeman @dgrammatiko  I have done changes, but continuous-integration/drone/pr shows `build is failing` what mistake I am doing?

dgrammatiko:
Check the logs: https://ci.joomla.org/joomla/joomla-cms/44832/1/20

Also please read https://github.com/joomla/joomla-cms/blob/4.0-dev/build/README.md

The order of the declarations is important, lint the file `npm run lint:css -- --fix`

ThisIsFaar:
@drmenzelit, @brianteeman @dgrammatiko I have done the changes, please review!

Quy:
There are horizontal/vertical scrollbars that don't do anything.

34387-vertical

34387-horizontal

ThisIsFaar:
> There are horizontal/vertical scrollbars that don't do anything.
> 
> 34387-vertical
> 
> 34387-horizontal

Those horizontal/vertical scrollbars are responsible for the Inner/Main Container to be scrollable flow.

scrollbars reason

brianteeman:
If there is now overflow then they should not be displayed

ThisIsFaar:
@dgrammatiko @brianteeman @drmenzelit can anyone explain the steps/stages in "pull request to be merged".

ThisIsFaar:
> If there is now overflow then they should not be displayed

Okay, I will try to fix this!

noxidsoft:
I like the scroll-bar solution, why would we hide columns, how do we know they exist if they are hidden on smaller screens?

PR #34406, opened 2021-06-04
[4.0] remove com_admin scss

brianteeman:
When this css was added the table headers of the template were not bold. Since they are now we don't need this extra css file.

Note that th in the header is still font-weight: 700 but the th in the body will be 500 - this is consistent with all other tables now.

### DIFF
--- a/administrator/templates/atum/scss/pages/_com_admin.scss
+++ b/administrator/templates/atum/scss/pages/_com_admin.scss
@@ -1,9 +1,0 @@
-// com_admin
-
-.sysinfo {
-  .table {
-    th {
-      font-weight: $font-weight-bold;
-    }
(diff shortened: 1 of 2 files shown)


Quy:
I have tested this item :white_check_mark: successfully on 765c3bfc977459aabe7b773b426a53a1f7257cf8<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34406">issues.joomla.org/tracker/joomla-cms/34406</a>.</sub>

wilsonge:
Thanks!

brianteeman:
every little bit helps

Issue #31435, opened 2020-11-20
[3.9] com_contact icon settings are not available in the administrator contact form even thought they are defined


shoulders:
### Steps to reproduce the issue
create a new contact or edit a contact and you will find there are no options to define the different types of icons as you can in the component global options
![image](https://user-images.githubusercontent.com/319997/99802004-d52e3f80-2b2e-11eb-833e-a48345de04b7.png)

These icons parameters are defined in `administrator/models/forms/contact.xml` but are not rendered in the form to set icons on a per contact basis.

https://github.com/joomla/joomla-cms/blob/a5b37ed5b199b7489e9f91b35a47209947d41c02/administrator/components/com_contact/models/forms/contact.xml#L238-L296

### Expected result
If the options are defined in the **contact.xml** they should be rendered or the referenced should be removed.

### Actual result
These fields are not rendered but are referenced in the template.

### System information (as much as possible)

Setting | Value
-- | --
 
PHP Built On | Windows NT E6540 10.0 build 19041 (Windows 10) AMD64
Database Type | mysql
Database Version | 5.5.5-10.4.10-MariaDB
Database Collation | utf8mb4_unicode_ci
Database Connection Collation | utf8mb4_general_ci
PHP Version | 7.3.12
Web Server | Apache/2.4.41 (Win64) OpenSSL/1.1.1c PHP/7.3.12
WebServer to PHP Interface | apache2handler
Joomla! Version | Joomla! 3.9.22 Stable [ Amani ] 6-October-2020 15:00 GMT
Joomla! Platform Version | Joomla Platform 13.1.0 Stable [ Curiosity ] 24-Apr-2013 00:00 GMT
User Agent | Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:82.0) Gecko/20100101 Firefox/82.0

### Additional comments
I had a quick look and it appears this issue is present in J4.x

drmenzelit:
What do you mean? Do you want to have different icons in each individual contact?

shoulders:
Do you want to have different icons in each individual contact? = this is what I am discussing.  😄 

I am not for or against this feature, it is just these input definitions are present and are not used.

so one of the following is required:

1. remove the unused definitions in `contact.xml` and just have the global options defined in the component.
2. correct the edit template to use these input fields

It is probably easier to remove the values in the `contact.xml` as I am not sure anyone has ever been able to use them on a per contact basis. However I would like to point you to the install.sql file which shows these options:

[...]

PR #34418, opened 2021-06-05
[4.0] com contact unused fields

brianteeman:
The form for an individual contact defines fields for all the icon images so that you can override them at the contact level as well as the component level.

However the fields are never displayed so they are never used. I have gone back over 8 years in the history and I cant see them being used at any point in that time period.

This pr removes the fields from the xml. No point in maintaining code that is never used


chmst:
The icons are used in single contacts. Here for example I have added an image as icon-address

grafik

brianteeman:
how? By using the form in the component options? That hasnt changed.

chmst:
Clobal confiuration - options. sry .. will be back later

brianteeman:
Thought so. I spent a stupid amount of time checking this

chmst:
Don't blame me ;) I spent an insane amout of time to find this: it is the marker-class
https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_contact/tmpl/contact/default_address.php#L27

brianteeman:
They're still global params aren't they?. Unless you have also found out where you can set the params for an individual contact

chmst:
There is no chance for individual contacts. I think that this was long time before my "active time". 
The default icons were somehow knuffy - 1990 style. And web site admins should have own icons for all their contacts (for all contacts).

brianteeman:
> There is no chance for individual contacts. I think that this was long time before my "active time".

As I said I went back 8 years without finding it

> The default icons were somehow knuffy - 1990 style. And web site admins should have own icons for all their contacts (for all contacts).

Agreed and they still can. This PR doesnt change that. The only thing this pr does is to remove a bunch of fields that are defined in xml and nowhere else

chmst:
Removing these fields would be a B/C break. If websites have adde here own icons. If they are removed, the lyouts can be made much easier.

brianteeman:
> Removing these fields would be a B/C break. 

How? they are NOT used. The ones that are used are in the global component options

brianteeman:
Or am I missing something

chmst:
Params are used here: 
https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_contact/src/View/Contact/HtmlView.php#L266 

Maybe I am missing something? I never use contact without own override, so am not expert here.

brianteeman:
Those icons are defined in the global options for the contact component aren't they?

Thats not the fields this PR is removing.

Please open joomla in your browser. You can set custom icons in the component options. There is nowhere you can set custom icons for an individual contact.

Now apply this PR. Nothing has changed. Therefore the fields in this xml are not used and can be deleted.

Quy:
See if #31435 confirms this.

brianteeman:
bingo

chmst:
My bad - I mixed up the xml files. You are right.

Fedik:
~the language strings also need to be removed or?~

ah I see, it also in component config, still in use

brianteeman:
Now we know the fields that are being removed can we get some tests please

Quy:
Should the info be updated/removed as this is the only place that these settings can be configured?

Should this be fixed in J3 too?

chmst:
@Quy yes, the message is wrong.

brianteeman:
And it was always bad english. ;)

I could remove it but then there is a consistency issue with the other tabs. How about simply

These settings apply to a single contact

chmst:
I don't use this feature, so cannot say where is the advantage if every contact has own icons. You could change the message and say "these settings are for all contacts and cannot be set individually" (in correct english)?

administrator/components/com_contact/config.xml
```diff
@@ -396,7 +396,7 @@
 	<fieldset
 		name="Icons"
 		label="COM_CONTACT_ICONS_SETTINGS"
-		description="COM_CONTACT_FIELD_CONFIG_INDIVIDUAL_CONTACT_DESC"
+		description="COM_CONTACT_ICONS_SETTINGS_DESC"
```
Quy, line 399: Why not keep the old language key and update its description?

brianteeman: because it is used in other places

chmst:
I have tested this item :white_check_mark: successfully on 59413069748e294c806de5fa32c849591d56d4db<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34418">issues.joomla.org/tracker/joomla-cms/34418</a>.</sub>

ChristineWk:
I have tested this item :white_check_mark: successfully on 59413069748e294c806de5fa32c849591d56d4db<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34418">issues.joomla.org/tracker/joomla-cms/34418</a>.</sub>

ChristineWk:
with Patch:

screen shot 2021-06-07 at 11 56 34<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34418">issues.joomla.org/tracker/joomla-cms/34418</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34418">issues.joomla.org/tracker/joomla-cms/34418</a>.</sub>

brianteeman:
thank you

PR #34419, opened 2021-06-05
[4.0] download key tip

brianteeman:
A tooltip is displayed for a missing download key that is not correct.

### Testing Instructions
Install this fake [ "paid download" ](http://updates.myoldsite.com/file_null-1.0.zip)extension Null file. Thanks @nikosdion for the file.

Go to System > Update > Extensions

Observe the alert and the tooltip

image

The tooltip should not be displayed here as its the tooltip for the update sites view and the instructions will not work on this page.

~This PR removes the tooltip from this page. I chose not to add a new tooltip with the correct information as it would just be a duplicate of the alert.~

Update
replaced the tooltip with a direct link to the extension update site page

Update 2
On the page Extensions: Update Sites the tooltip has been converted from a bootstrap tooltip
image

### DIFF
--- a/administrator/components/com_installer/tmpl/update/default.php
+++ b/administrator/components/com_installer/tmpl/update/default.php
@@ -101,12 +101,7 @@ $listDirn  = $this->escape($this->state->get('list.direction'));
 										<?php if($item->isMissingDownloadKey): ?>
 										<?php HTMLHelper::_('bootstrap.popover', 'span.hasPopover', ['trigger' => 'hover focus']); ?>
 										<span class="badge bg-danger">
-											<span class="hasPopover"
-												  title="<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL') ?>"
-												  data-bs-content="<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_TIP') ?>"
-											>
-												<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?>
-												</span>
+											<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?>
 										</span>
 										<?php endif; ?>
 									</th>
--- a/administrator/components/com_installer/tmpl/update/default.php
+++ b/administrator/components/com_installer/tmpl/update/default.php
@@ -103,7 +103,7 @@ $listDirn  = $this->escape($this->state->get('list.direction'));
 												<?php $url = 'index.php?option=com_installer&view=updatesite&layout=edit&update_site_id=' . $item->update_site_id; ?>
 											<?php } ?>
 											<span class="badge bg-danger">
-													<a href="<?php echo $url; ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
+													<a class="text-decoration-none text-white" href="<?php echo $url; ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
 											</span>
 										<?php endif; ?>
 									</th>
--- a/administrator/components/com_installer/tmpl/update/default.php
+++ b/administrator/components/com_installer/tmpl/update/default.php
@@ -99,12 +99,8 @@ $listDirn  = $this->escape($this->state->get('list.direction'));
 											<?php endif; ?>
 										</div>
 										<?php if($item->isMissingDownloadKey): ?>
-											<?php { ?>
-												<?php $url = 'index.php?option=com_installer&view=updatesite&layout=edit&update_site_id=' . $item->update_site_id; ?>
-											<?php } ?>
-											<span class="badge bg-danger">
-													<a class="text-decoration-none text-white" href="<?php echo $url; ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
-											</span>
+											<?php $url = 'index.php?option=com_installer&task=updatesite.edit&update_site_id=' . (int) $item->update_site_id; ?>
+											<a class="btn btn-danger btn-sm text-decoration-none" href="<?php echo $url; ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
 										<?php endif; ?>
 									</th>
 									<td class="d-none d-md-table-cell">
--- a/administrator/components/com_installer/tmpl/update/default.php
+++ b/administrator/components/com_installer/tmpl/update/default.php
@@ -100,8 +100,8 @@ $listDirn  = $this->escape($this->state->get('list.direction'));
 										</div>
 										<?php if($item->isMissingDownloadKey): ?>
 											<?php $url = 'index.php?option=com_installer&task=updatesite.edit&update_site_id=' . (int) $item->update_site_id; ?>
-											<a class="btn btn-danger btn-sm text-decoration-none" href="<?php echo $url; ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
-										<?php endif; ?>
+											<a class="btn btn-danger btn-sm text-decoration-none" href="<?php echo Route::_($url); ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
+											<?php endif; ?>
 									</th>
 									<td class="d-none d-md-table-cell">
 										<?php echo $item->client_translated; ?>
--- a/administrator/components/com_installer/tmpl/update/default.php
+++ b/administrator/components/com_installer/tmpl/update/default.php
@@ -99,7 +99,6 @@ $listDirn  = $this->escape($this->state->get('list.direction'));
 											<?php endif; ?>
 										</div>
 										<?php if($item->isMissingDownloadKey): ?>
-										<?php print_r($item); ?>
 											<?php { ?>
 												<?php $url = 'index.php?option=com_installer&view=updatesite&layout=edit&update_site_id=' . $item->update_site_id; ?>
 											<?php } ?>
--- a/administrator/components/com_installer/tmpl/update/default.php
+++ b/administrator/components/com_installer/tmpl/update/default.php
@@ -99,10 +99,13 @@ $listDirn  = $this->escape($this->state->get('list.direction'));
 											<?php endif; ?>
 										</div>
 										<?php if($item->isMissingDownloadKey): ?>
-										<?php HTMLHelper::_('bootstrap.popover', 'span.hasPopover', ['trigger' => 'hover focus']); ?>
-										<span class="badge bg-danger">
-											<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?>
-										</span>
+										<?php print_r($item); ?>
+											<?php { ?>
+												<?php $url = 'index.php?option=com_installer&view=updatesite&layout=edit&update_site_id=' . $item->update_site_id; ?>
+											<?php } ?>
+											<span class="badge bg-danger">
+													<a href="<?php echo $url; ?>"><?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?></a>
+											</span>
 										<?php endif; ?>
 									</th>
 									<td class="d-none d-md-table-cell">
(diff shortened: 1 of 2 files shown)


nikosdion:
Or we could simply add `data-bs-placement="bottom"` to have the popover appear below the Badge.

The point of the popover is that newcomers to Joomla do not understand that the Download Key is required to install the update. Yeah, sure there's the warning at the top but experience says they don't ever read it. The point of the popover is that they are far more likely to move their pointer over the red text than look somewhere else on the page to understand what the heck is that message.

I also stand by my statement nearly 2 years prior that the Update Site edit page should allow developers to display a custom message (via a translation screen) to direct the user as to where their Download Key can be found. The Joomla interface, as it is implemented right now, expect the user to _know_ things. My experience doing support with paying customers is that they do not understand the relation between purchasing something and having to not only install it but also type something somewhere for updates to work. Even more so when they've hired a site integrator to build their site and 6 to 12 months later they try to take over maintenance of the site and have no idea what is going on. Giving folks some hints about what they need to do helps A LOT. I observed an ~80% reduction in support tickets regarding download keys and failed updates when I started nagging people to enter their Download IDs and told them where to find them on my site.

brianteeman:
Nik

read the tooltip. on this **update extensions** page the text is not correct. it is correct on the **update sites** page where it is also displayed.

It says click on the update sites title. This view does not have an update site title or anything that is clickable
image

The alert that is displayed on this page contains a link that takes you to the update sites page where the update site title exists and is clickable.

Happy to have a tooltip on this page but the text I am deleting actually creates issues instead of your intention of solving them

nikosdion:
You make a very fair point that the tooltip content is misleading and its placement is awkward. I agree. You can have a different tooltip content there, e.g. “Updates for this extension will not work until you enter its Download Key in the Update Sites page. Please check the message above the list of updates for instructions.”

However, the tooltip and its content are implementation details which are up for discussion. Let's first focus on why a tooltip is there in the first place.

Joomla's extensions update process is idiosyncratic. It diverges from both update models users are familiar with: update in the software (everything is managed in the software being updated) and updates as part of a centralised software management facility / App Store (where the user needs to do nothing and updates appear “automatically”). It is very reasonable that the average user who's not overly familiar with Joomla's update model can't make the connection between the badge, the top of the page, the Update Sites page, the Download Key and something on my or any other 3PD's site. This makes updates look scary, opaque and burdensome. We need to hold the user's hand and walk them through the process.

Here's how I envisioned it about two years ago:

* The badge is strategically placed in proximity to the checkbox they need to click to update the software. It also stands out.
* They will very likely hover over the badge because it's a Red Angry Thing. The tooltip will tell them to look at the top of the page.
* The top of the page tells them to enter the Download Key and has a link. Hopefully, they click on it.
* They see the list of update sites which don't have a Download Key. If they're lost, they will hover over the Red Angry Thing there.
* That tooltip tells them to click on the Update Site title and enter the Download Key.
* (missing step) The Update Site edit page tells them where to get the Download Key from.
* They paste the Download Key and click on Save & Close.
* Now they can start over the update process.

This process is multistep but the steps lead from one to the next. This makes people think it's easy since, as you know, clarity is what people prefer over brevity.

I am not particularly attached to having a tooltip. If there's another way, e.g. making the Red Angry Thing a link, I actually prefer it; tooltips require interaction which poses a discoverability issue. They are also not very accessible which is another problem. The point is, how can we hold the user by the hand and help them navigate a scary and unfamiliar process? I am open to suggestions as long as it doesn't require the user to magically make a connection between two unrelated parts of the page because I know they won't and that will be a problem I will have to provide support for 😉

brianteeman:
the angry red text can easily be a link if thats what you want. I just wanted to remove the incorrect instructions in the tooltip

nikosdion:
I'd prefer it a link, if it shows up legible. I believe the link should have the `text-decoration-none` class as well to prevent the entire badge from becoming underlined and hard to read. I can't remember if two years ago I didn't know about this class or if it didn't exist in BS4.

brianteeman:
easy enough - almost done

brianteeman:
~Changed to a link **but** need a little assistance making the link work and not give a direct access is not allowed~

brianteeman:
All changes made as requested - please test

Abernyte-Git:
I have tested this item :white_check_mark: successfully on 8f1ecdd371baa7d5f2c48c4f3d49149796ec1b2e<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

administrator/components/com_installer/tmpl/update/default.php
```diff
@@ -99,15 +99,8 @@
...
-										<?php HTMLHelper::_('bootstrap.popover', 'span.hasPopover', ['trigger' => 'hover focus']); ?>
-										<span class="badge bg-danger">
-											<span class="hasPopover"
-												  title="<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL') ?>"
-												  data-bs-content="<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_TIP') ?>"
-											>
-												<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?>
-												</span>
-										</span>
+											<?php $url = 'index.php?option=com_installer&task=updatesite.edit&update_site_id=' . (int) $item->update_site_id; ?>
```
nikosdion, line 102: Please change that to `$url = \Joomla\CMS\Router\Route::_('index.php?option=com_installer&task=updatesite.edit&update_site_id=' . (int) $item->update_site_id);`. All administrator URLs in Joomla 4 are going through the URL router which addresses some issues with less common server setups and/or accidental redirections to the site's front-end. In the future we might even see SEF URLs in the backend.

nikosdion (changes requested):
One minor change and we're good to go!

administrator/components/com_installer/tmpl/update/default.php
joomdonation: Actually, it could be shorter because we have `use Joomla\CMS\Router\Route;` at the top already

```php
$url = Route::_('index.php?option=com_installer&task=updatesite.edit&update_site_id=' . (int) $item->update_site_id);
```

However, we should only display the link it the current user has edit permission instead of always displaying it like currently. I think you can copy this block of code (contains permission check) https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_installer/tmpl/updatesites/default.php#L94-L104 and adjust it to have it works with your PR.

nikosdion: It's not necessary. Only Super Users can install extensions — including updates — which means that they already have all permissions. 

We had changed that sometime around 2.5 if I recall correctly because installing extensions is a privileged operation with security repercussions. Back in the Joomla 1.5 days even Administrators could install extensions. I had demonstrated that I could create a "fake" component package which would create a Super User account under my control and then uninstall itself, leaving no trace (except the Super User). Using the same method I could also change an existing Super User and block all other Super Users, essentially performing a hostile takeover. This led to extensions being installable only by Super Users, the most trusted tier with absolute site access.

joomdonation: @nikosdion As of right now, I can see that permission is configurable, so we can still configure permission to allow administrator to do anything.
permission

But yes, the default permission when we install Joomla is that only Super Users can do these tasks.

nikosdion: I can confirm that the hardcoded ACL checks have been removed and we now do regular ACL checks. I'll report this to the JSST. It's a major security issue.

brianteeman: I will leave that to them and declare it out of scope for this PR

brianteeman:
This is ready for testing. The security comment is out of scope for this pr

administrator/components/com_installer/tmpl/update/default.php
joomdonation: Could you please at least move the url inside the `Route::_`  call like this https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_installer/tmpl/updatesites/default.php#L96 ? We do not want to define an one time use variable (the $url variable in this case).

brianteeman: @joomdonation this is no different to lots of other code and it is much more readable

joomdonation: I don't know if we code guideline but I'm sure we advice to not define onetime use variable on other PRs before. For readable, the code like this https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_installer/tmpl/updatesites/default.php#L95-L101 looks good to me. But, that's personal preference.

ChristineWk:
Have the following current state (see screenshots).
Patch is enabled - Joomla 4.0.0-rc2-dev from today:

screen shot 2021-06-06 at 13 05 51

screen shot 2021-06-06 at 13 06 06

screen shot 2021-06-06 at 13 06 24


 <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

brianteeman:
something not right there as you are displaying a tooltip that has been removed

Quy:
Go to Update Sites to see tooltip.

Abernyte-Git:
I get the tooltip no longer showing as per expected at Update-Extension: Update but it does show at Extensions: Update Sites.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on 46ae40adede3046ca1c6b741bafd6e886d8ccddf<br /><br />Works like described! <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

richard67:
I have tested this item :white_check_mark: successfully on 46ae40adede3046ca1c6b741bafd6e886d8ccddf<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

richard67:
@joomdonation Setting RTC here because it does what it says. If you want, make a PR later with your suggested change after that one got merged.

brianteeman:
Updated to convert the tooltip on the update sites page

administrator/components/com_installer/tmpl/updatesites/default.php
```diff
@@ -114,14 +114,12 @@
...
-										<span class="badge bg-danger">
-											<span class="hasPopover"
-													title="<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL') ?>"
-													data-bs-content="<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_TIP') ?>"
-											>
+										<span class="badge bg-danger" tabindex="0">
 											<?php echo Text::_('COM_INSTALLER_DOWNLOADKEY_MISSING_LABEL'); ?>
-											</span>
 										</span>
+										<div role="tooltip" id="tip<?php echo $i; ?>">
```
Quy: See line 130. Duplicate `id`.

RickR2H:
I have tested this item :white_check_mark: successfully on c7d6deca113e025c2aa240db1224c8cb4602d112<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

brianteeman:
would be great if this could get merged today before the language freeze

richard67:
I have tested this item :white_check_mark: successfully on c7d6deca113e025c2aa240db1224c8cb4602d112<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34419">issues.joomla.org/tracker/joomla-cms/34419</a>.</sub>

richard67:
> would be great if this could get merged today before the language freeze

The PR doesn't have a language change anymore after it had been changed.

Quy:
Thanks!

brianteeman:
finally

Issue #34362, opened 2021-06-02
[4.0] After update from 3.10, Global Checkin shows all items of the "#__extensions" table as checked out, but they aren't


richard67:
### Steps to reproduce the issue

Update a clean 3.10 Alpha 6 or latest 3.10- nightly or a current 3.10-dev branch to 4.0 RC 1 or latest 4.0 nightly.

For updating to 4.0 RC 1 you can use custom URL https://update.joomla.org/core/test/310to4_list.xml .

After the update has finished, go to "System -> Maintenance -> Global Check-in".

### Expected result

Nothing to be checked in which was not to be checked in before the update.

### Actual result

A lot of items (215 on a clean and empty installation) are shown to be checked it for the the `#___extensions` table.

In fact these are all records of that table.

If you go to "System -> Manage -> Extensions", everything looks ok, nothing is checked out.

If you check in database you will see that the `checked_out` column is zero for all extensions, but the `checked_out_time` column doesn't have value NULL but has the old style (peudo) null dates ('0000-00-00 00:00:00' on MySQL or MariaDB and '1970-01-01 00:00:00' on PostgreSQL databases).

Checking in the items in Global Check-In works, i.e. after that they are not shown as checked out anymore.

### System information (as much as possible)

Any system after update from current 3.10-dev to current 4.0-dev or corresponding latest releases or nightlies.

### Additional comments

The  the update SQL script for changing to NULL values works well, but the `checked_out_time` column of the `#__extensions` table seems to be updated AFTER the update SQL scripts have run.

You can verify this by adding a `die` call just after this line, i.e. just after the update SQL scrips have been processed:

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php#L708

Then pack an update packages with that modified file and use that package for updating a 3.10 using Upload & Update.

Then check in database and you will see all records in the `#__extensions` table have a `checked_out_time` of NULL, like it should be in J4.

But I wasn't able (yet?) to find out where it happens.

**_Update 2021-06-04:_** If I add a die statement later below the following line, then the values have the old '0000-00-00 ...' in database:

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php#L740

So the error happens somewhere inside here: https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_admin/script.php#L78

[...]

PR #34433, opened 2021-06-06
[4.0] Reinitialise the installer's cached extensions table's properties after a core update.

richard67:
Pull Request for Issue #34362 .

### Summary of Changes

At the end of a core update, especially to a new major version, it may happen that there are structural changes on columns of the extensions table applied by the update SQL scripts, e.g. like real NULL values and no default value for the `checked_out_time` column.

But the installer object still uses old, cached information about these columns even if you create a new installer object, and so uses the old default values for such columns when loading the extensions table in its constructor.

When after running the update SQL scripts the manifest cache of all core extensions is updated, this results in the old default values being written for columns like e.g. the `checked_out_time` column, where the old zero date '0000-00-00 00:00:00' on MySQL and MariaDB or '1970-01-01 00:00:00' on PostgreSQL are used when storing the table.

This result in issue #34362 .

This pull request (PR) fixes that by calling the `getFields` method of the installer's extensions table object with the `$reload` parameter set to `true` just after the schema updates have run, so the table columns properties are read again from database with the new table structure.

That `getFields` is the one of the parent class here: https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Table/Table.php#L242

### Testing Instructions

1. Update a clean 3.10 Alpha 6 or latest 3.10- nightly or a current 3.10-dev branch to 4.0 RC 1 or latest 4.0 nightly.

2. After the update has finished, go to "System -> Maintenance -> Global Check-in".
Result: See section "Actual result BEFORE applying this Pull Request" below.

3. Update a clean 3.10 Alpha 6 or latest 3.10- nightly or a current 3.10-dev branch to the patched package built buy drone for this PR using either the custom update URL or the zip package.

4. After the update has finished, go to "System -> Maintenance -> Global Check-in".
Result: See section "Expected result AFTER applying this Pull Request" below.

A lot of items (215 on a clean and empty installation) are shown to be checked it for the the `#___extensions` table.

In fact these are all records of that table.

If you go to "System -> Manage -> Extensions", everything looks ok, nothing is checked out.

If you check in database you will see that the `checked_out` column is zero for all extensions, but the `checked_out_time` column doesn't have value NULL but has the old style (peudo) null dates ('0000-00-00 00:00:00' on MySQL or MariaDB and '1970-01-01 00:00:00' on PostgreSQL databases).

Checking in the items in Global Check-In works, i.e. after that they are not shown as checked out anymore.

Nothing to be checked in which was not to be checked in before the update (which should be nothing when updating from a clean 3.10).

None.

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -91,6 +91,9 @@ class JoomlaInstallerScript
 			// Informational log only
 		}
 
+		// Reinitialise the installer's extensions table's properties.
+		$installer->extension->getFields(true);
+
 		// Ensure we delete the repeatable fields plugin before we remove its files
 		$this->uninstallRepeatableFieldsPlugin();
 
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -91,9 +91,6 @@ class JoomlaInstallerScript
 			// Informational log only
 		}
 
-		// Reinitialise the installer's extensions table's properties.
-		$installer->extension->getFields(true);
-
 		// Ensure we delete the repeatable fields plugin before we remove its files
 		$this->uninstallRepeatableFieldsPlugin();
 
--- a/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php
+++ b/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php
@@ -715,6 +715,9 @@ ENDDATA;
 			return false;
 		}
 
+		// Reinitialise the installer's extensions table's properties.
+		$installer->extension->getFields(true);
+
 		// Start Joomla! 1.6.
 		ob_start();
 		ob_implicit_flush(false);


joomdonation:
I have tested this item :white_check_mark: successfully on c42b507bf339c89690dd19499c4c22fc7a2383ac<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34433">issues.joomla.org/tracker/joomla-cms/34433</a>.</sub>

PhilETaylor:
I have tested this item :white_check_mark: successfully on c42b507bf339c89690dd19499c4c22fc7a2383ac<br /><br />Code review only<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34433">issues.joomla.org/tracker/joomla-cms/34433</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on c42b507bf339c89690dd19499c4c22fc7a2383ac<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34433">issues.joomla.org/tracker/joomla-cms/34433</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34433">issues.joomla.org/tracker/joomla-cms/34433</a>.</sub>

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34433">issues.joomla.org/tracker/joomla-cms/34433</a>.</sub>

richard67:
Sorry all but after talk with @wilsonge I had to make a change. I've moved the fix from `script.php` to the joomlaupdate mode.

The fix is still the same and happens at the same point of the update, just the code has been moved.

Please test again. Thanks in advance.

wilsonge:
I'm happy with just the code moving to merge after the previous tests. Thanks all!

richard67:
Thanks all!

PR #34482, opened 2021-06-10
[4.0] Fix searchtool JS error

joomdonation:
Pull Request for Issue # .

### Summary of Changes
Currently, if we use searchtools script but there are no filter fields for some reasons (it could happen with third party extensions like mine), there could be javascript errors in browser console like this:

```javascript
Uncaught TypeError: Cannot read property 'forEach' of undefined
    at Searchtools.checkActiveStatus (searchtools.js?2de576a4eb7e64e845e8e1a5fafd9f831d1aec5c:264)
    at new Searchtools (searchtools.js?2de576a4eb7e64e845e8e1a5fafd9f831d1aec5c:194)
    at HTMLDocument.onBoot (searchtools.js?2de576a4eb7e64e845e8e1a5fafd9f831d1aec5c:504)
```

We have the code to check and make sure filter fields exist on other places like:

https://github.com/joomla/joomla-cms/blob/4.0-dev/build/media_source/system/js/searchtools.es6.js#L177
https://github.com/joomla/joomla-cms/blob/4.0-dev/build/media_source/system/js/searchtools.es6.js#L248
....

So I think it makes sense to have the same check in `checkActiveStatus` method, too.

### Testing Instructions
I think this issue only happens for third party extensions only, so there is nothing to check from core. So we will have to rely on code review here.

@dgrammatiko Could you please take a look at this to see if the change is OK ?

### DIFF
--- a/build/media_source/system/js/searchtools.es6.js
+++ b/build/media_source/system/js/searchtools.es6.js
@@ -286,18 +286,16 @@ Joomla = window.Joomla || {};
 
     // eslint-disable-next-line class-methods-use-this
     checkActiveStatus(cont) {
-      const els = this.getFilterFields();
+      const els = [].slice.call(this.getFilterFields());
       let activeFilterCount = 0;
 
-      if (els) {
-        els.forEach((item) => {
-          if (item.classList.contains('active')) {
-            activeFilterCount += 1;
-            cont.filterButton.classList.remove('btn-secondary');
-            cont.filterButton.classList.add('btn-primary');
-          }
-        });
-      }
+      els.forEach((item) => {
+        if (item.classList.contains('active')) {
+          activeFilterCount += 1;
+          cont.filterButton.classList.remove('btn-secondary');
+          cont.filterButton.classList.add('btn-primary');
+        }
+      });
 
       // If there are no active filters - remove the filtered caption area from the table
       if (activeFilterCount === 0) {
--- a/build/media_source/system/js/searchtools.es6.js
+++ b/build/media_source/system/js/searchtools.es6.js
@@ -289,13 +289,15 @@ Joomla = window.Joomla || {};
       const els = this.getFilterFields();
       let activeFilterCount = 0;
 
-      els.forEach((item) => {
-        if (item.classList.contains('active')) {
-          activeFilterCount += 1;
-          cont.filterButton.classList.remove('btn-secondary');
-          cont.filterButton.classList.add('btn-primary');
-        }
-      });
+      if (els) {
+        els.forEach((item) => {
+          if (item.classList.contains('active')) {
+            activeFilterCount += 1;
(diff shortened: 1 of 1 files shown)


build/media_source/system/js/searchtools.es6.js
```diff
@@ -289,13 +289,15 @@ Joomla = window.Joomla || {};
       const els = this.getFilterFields();
```
dgrammatiko, line 289: Using `const els = [].slice.call(this.getFilterFields());` you can skip the if part

joomdonation: Thanks @dgrammatiko for better code, tested and confirmed it works.

richard67:
I have tested this item :white_check_mark: successfully on 7714b66d964405636409a82b0035317359d2f72c<br /><br />Code review and research what `[].slice.call(...)` does, and noticed by review of diverse other PR's that we use it already elsewhere with success.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34482">issues.joomla.org/tracker/joomla-cms/34482</a>.</sub>

richard67:
@dgrammatiko Could you test this PR by review? Thanks in advance.

dgrammatiko:
I have tested this item :white_check_mark: successfully on 7714b66d964405636409a82b0035317359d2f72c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34482">issues.joomla.org/tracker/joomla-cms/34482</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34482">issues.joomla.org/tracker/joomla-cms/34482</a>.</sub>

richard67:
Thanks @dgrammatiko .

dgrammatiko:
@richard67 a more general note here, the `[].slice.call(...)` forces the `HTMLElement.querySelectorAll()` to return an `Array`. The difference is that `HTMLElement.querySelectorAll()` will return either:
- `undefined` is no element exists
- a node list (which could be treated as an array)

So in the case that `HTMLElement.querySelectorAll()` returns `undefined` the code needs a conditional but if we have converted it to an array (in this case will be an empty array) the code won't break...

richard67:
> So in the case that `HTMLElement.querySelectorAll()` returns `undefined` the code needs a conditional but if we have converted it to an array (in this case will be an empty array) the code won't break...

@dgrammatiko Yes that's how I understand it.

Quy:
Thanks!

PR #34499, opened 2021-06-13
[4.0] Unused strings in joomla.ini (admin / api / site)

Kostelano:
### Summary of Changes
I am engaged in translation. Found a lot of unused constants in joomla.ini.

I delete them in the PR, then copy / paste to the API directory.

And added a constant for the AUTHOR meta tag, as using a generic constant is not a good idea. For example, in Russian, we translate all meta-tags "Метатег Robots", "Метатег Keywords" ... And suddenly, among the meta-data, there is just the line AUTHOR.


richard67:
Are you sure that these strings are all not used? In some (rare) cases language constants are combined in code from parts, so you won’t find the complete language constant when searching for it.

brianteeman:
also please note I have personally never removed strings from joomla.ini as thats a generic language file that is used by extensions

Kostelano:
I know that part of a string is sometimes used, so I tested strings in a similar way. Of course, I would be glad if someone re-checks, but I think that everything is in order here.

As for other extensions, it is certainly impossible to guess here, but I do not think that many extensions use the lines of the main file. Besides, it is not a good idea to "reserve" a lot of unused lines of Joomla itself for some third-party extension, since the number of such lines will grow forever.

Here in the PR there are lines that were probably used back in Joomla 1.0 :).

brianteeman:
> Here in the PR there are lines that were probably used back in Joomla 1.0 :).

Probably - even more reason that they are used elsewhere.

> As for other extensions, it is certainly impossible to guess here, but I do not think that many extensions use the lines of the main file.

And that's a guess ;)

It took me 15 years to delete a string from  specific language files.

Issue #34510, opened 2021-06-14
[4.0] Pre-update check "More Detail" and "Less Detail" toggles of the PHP checks are broken.


richard67:
### Steps to reproduce the issue

On a current 4.0-dev branch which includes the latest commits from stuff having been merged up from 3.10-dev today, make a new installation if you haven't done one yet.

Then go to the Joomla Update Component's options and set the update channel to the custom URL of the nightlies so that the pre-update checker is shown after saving.

Now toggle a few times one of the "More Detail" and "Less Detail" links in the legends of the 2 PHP checks "Required PHP & Database Settings" and "Recommended PHP Settings" at the top.

_Important:_ Click on the arrow symbol for toggling, not on the "More Detail" or "Less Detail" texts.

### Expected result

The toggle just changes the text and the arrow icon.

### Actual result

The link text is added again and again with each toggle:

![2021-06-14_pre-update-check-js-problem](https://user-images.githubusercontent.com/7413183/121872272-5cc6e000-cd05-11eb-99d3-d11e2489abda.png)

### System information (as much as possible)

Current 4.0-dev branch from today, June 14 2021.

### Additional comments

Might be something to be fixed in the JS of the Joomla Update Component.

brianteeman:
it is a change in the markup that is impacting on the js. probably after the recent merge from @wilsonge

richard67:
> probably after the recent merge from @wilsonge

@brianteeman Remove the "probably" ;-)

richard67:
Strange: Now after having been logged out for a while, it works, and I can't reproduce this issue here.

richard67:
I've just tried again with a fresh and clean 4.0-dev and still can't reproduce it, so I guess it was some kind of caching issue.

joomdonation:
I use nightly build and don't see the issue, too. So we are fine.

Quy:
To reproduce it, click on the arrow and not the text.

richard67:
>To reproduce it, click on the arrow and not the text.

@Quy Confirmed.

joomdonation:
PR https://github.com/joomla/joomla-cms/pull/34516 should fix this issue. Please test.

PR #34516, opened 2021-06-14
[4.0] Fix com_joomlaupdate js error

joomdonation:
Pull Request for Issue #34510.

### Summary of Changes
This PR fixes few js error for com_joomlaupdate. See https://github.com/joomla/joomla-cms/issues/34510 for more details. It also fixes an error in the console when the page is loaded (You need to install a third party extension like JCE to see this error - or code review)

### Testing Instructions
1. See https://github.com/joomla/joomla-cms/issues/34510 to understand the issue
2. Download update package for this PR, go to System -> Update -> Joomla to install the update package for this PR (or apply patch, and run npm command, sorry I don't remember the command to run)
3. Check and confirm the issue is gone.

### DIFF
--- a/build/media_source/com_joomlaupdate/js/default.es6.js
+++ b/build/media_source/com_joomlaupdate/js/default.es6.js
@@ -99,17 +99,17 @@ Joomla = window.Joomla || {};
     [].slice.call(document.querySelectorAll('.settingstoggle')).forEach((el) => {
       el.style.float = 'right';
       el.style.cursor = 'pointer';
-      el.addEventListener('click', (toggle) => {
+      el.addEventListener('click', () => {
         const settingsfieldset = el.closest('fieldset');
-        if (toggle.target.dataset.state === 'closed') {
-          toggle.target.dataset.state = 'open';
-          toggle.target.innerHTML = Joomla.sanitizeHtml(Joomla.getOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION'));
+        if (el.dataset.state === 'closed') {
+          el.dataset.state = 'open';
+          el.innerHTML = Joomla.sanitizeHtml(Joomla.getOptions('COM_JOOMLAUPDATE_VIEW_DEFAULT_EXTENSIONS_SHOW_LESS_COMPATIBILITY_INFORMATION'));
           settingsfieldset.querySelectorAll('.settingsInfo').forEach((fieldset) => {
             fieldset.classList.remove('hidden');
           });
         } else {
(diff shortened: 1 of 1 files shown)


richard67:
@joomdonation It fixes the error in my issue, but regarding the console error there is still `Uncaught TypeError: document.getElementById(...) is null` here: https://github.com/joomla/joomla-cms/pull/34516/files#diff-550b2e9757e1fef80e3714191e4f70f3dd9bd7441ff3e7679698de031551e346R420

joomdonation:
@richard67 I don't see the issue myself for some reasons. Maybe it's because I have some third party extension installed (JCE, and my own extensions). But I saw that error before and the modification in that line of code fixes that. The # character is needed for old code (jQuery) but not needed on document.getElementById

richard67:
> @richard67 I don't see the issue myself for some reasons. Maybe it's because I have some third party extension installed (JCE, and my own extensions). But I saw that error before and the modification in that line of code fixes that. The # character is needed for old code (jQuery) but not needed on document.getElementById

@joomdonation No need to explain me the "#" change, I understand that already.

joomdonation:
> No need to explain me the "#" change, I understand that already.

Sorry, I just tried to explain the reason for the change on that line of code.

richard67:
@joomdonation Yes but unfortunately it still doesn't work. Seems there is something more wrong with that element id.

richard67:
I can't see any html element having an id "plg_226", which is the one for which the document.getElementById call fails. 226 is the id of the JCE extension in my installation.

joomdonation:
Somehow, it relates to logic of com_joomlaupdate which I don't understand yet. The element is only available if the layout administrator/components/com_joomlaupdate/tmpl/joomlaupdate/default_update.php is loaded , but in this case, it is not loaded for some reasons (maybe it is because no update found for that extension, will need to look more at com_joomlaupdate to understand)

richard67:
@joomdonation One solution could be that you revert that one change which is not related to the issue and which does not work for me, so the other change can be tested before RC 2 is made tonight, and for the remaining js error make a new issue.

joomdonation:
@richard67 OK. We will handle that change on a separate PR.

richard67:
I have tested this item :white_check_mark: successfully on 8f8f971dc6941d772d56d5c57c32ec6f819cc738<br /><br />Fixes #34510 .<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34516">issues.joomla.org/tracker/joomla-cms/34516</a>.</sub>

wilsonge:
Thanks!

PR #34538, opened 2021-06-17
[4.0] multilingual urls

### LINKED ISSUE #34523: [4.0] When Language Filter with sef_suffix are active base URL's are like /en.html
Make a multilingual website:
- Install new language
- Create a content language
- Enable the Language Filter plugin (associations = yes, makes it easier to see the result)
- Activate _sef_ + _sef_suffix_ (+ _sef_rewrite_) in the global configuration

### Expected result
Associations and links in mod_languages to the default (home) pages of different languages should be:
- https://www.example.com/en/
- https://www.example.com/es/
- https://www.example.com/nl/

The .html suffix only with deep links to non default pages:
- https://www.example.com/en/my-article.html

### Actual result
Associations and links in mod_languages to the default pages are like:
- https://www.example.com/en.html
- https://www.example.com/es.html
- https://www.example.com/nl.html

- Joomla! 4.0.0-rc1 Release Candidate [ Mañana ] 1-June-2021 20:36 GMT
- PHP 7.4.10
- nginx/1.19.2
- fpm-fcgi

### Additional comment
I was wondering if the problem lies in the Language Filter plugin or in the _sef_suffix_ logic.

brianteeman:
Hopefully this fixes #34523

Hopefully there are unit tests in existence which will verify that

For human testing
Create a multlingual site and check that the urls for the home pages are correct
Test with both suffix enabled and disabled

Before this PR the url would be example.com/fr.html
After this PR the url would be example.com/fr/

### DIFF
--- a/libraries/src/Router/SiteRouter.php
+++ b/libraries/src/Router/SiteRouter.php
@@ -500,7 +500,7 @@ class SiteRouter extends Router
 
 		// Set query again in the URI
 		$uri->setQuery($query);
-		$uri->setPath(trim($route, '/'));
+		$uri->setPath($route);
 	}
 
 	/**


ghost:
I have tested this item :white_check_mark: successfully on d1cbbed232cc3477b2de2521a327339e69ccade9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34538">issues.joomla.org/tracker/joomla-cms/34538</a>.</sub>

libraries/src/Router/SiteRouter.php
```diff
@@ -500,7 +500,7 @@ public function buildSefRoute(&$router, &$uri)
 
 		// Set query again in the URI
 		$uri->setQuery($query);
-		$uri->setPath(trim($route, '/'));
+		$uri->setPath($route);
```
richard67, line 503: @brianteeman Maybe better use `$uri->setPath(ltrim($route, '/'));` so still a leading slash would be trimmed, only trailing slash not?

brianteeman: I dont know

richard67: Me neither ... was just an idea to keep impact little compared to existing code. I'd really like @Hackwar to have a look on it.

richard67:
Unfortunately this change also affects other pages than the home page.

On a copy of my homepage updated to 4.0 RC 2, I get `https://test7.richard-fath.de/de/music/` for a categroy list with this PR applied, and not `https://test7.richard-fath.de/de/music.html` like it is on my J3 homepage or on the J4 copy without this PR.

brianteeman:
That's why we should not be relying solely on human  tests and should have automated testing

richard67:
I have tested this item :red_circle: unsuccessfully on d1cbbed232cc3477b2de2521a327339e69ccade9<br /><br />See my previous comment.

For other testers: It's not enough just to test the home pages which have the issue. If such changes are done with the router, it needs to test other pages, too.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34538">issues.joomla.org/tracker/joomla-cms/34538</a>.</sub>

ghost:
@richard67 

> For other testers: It's not enough just to test the home pages which have the issue. If such changes are done with the router, it needs to test other pages, too.

Test instructions by @brianteeman:

> For human testing
Create a multlingual site and check that the urls for the home pages are correct
Test with both suffix enabled and disabled.

richard67:
@sandramay0905 Sorry, did not want to blame you or anyone. Only wanted to say sometimes it needs to test a bit more than instructions say.

libraries/src/Router/SiteRouter.php
elpescador-nl: This change makes the trailing slash reappear in the home/default page like https://example.com/en/ 👍
...but also in https://example.com/en/deep-link/ whereas the deep link **should not have** this trailing slash: https://example.com/en/deep-link and with _sef_suffix_ active it should be https://example.com/en/deep-link.html

richard67:
@sandramay0905 P.S. ... and thanks a lot for your testing activities in general. It really helps us a lot and is much appreciated.

brianteeman:
without automated tests and documentation of how the urls are supposed to work its virtually impossible

richard67:
Alternative PR is #34558 . It will be ready for test soon.

~~Update: No, doesn't work (yet).~~

Update 2: Now it works. I only don't know if my PR is the right way to fix the issue.

PR #34571, opened 2021-06-20
[4.0] Plural for password rules

Kostelano:
### Summary of Changes
Make it possible to change the string of password parameters for each language separately.

For example, in Russian it will look much better if you change the wording to "N characters ...". Changes for a bunch of lines:

```
JFIELD_PASSWORD_RULES_CHARACTERS="Characters: %d"
JFIELD_PASSWORD_RULES_DIGITS="Numbers: %d"
JFIELD_PASSWORD_RULES_LOWERCASE="Lower Case: %d"
...
JFIELD_PASSWORD_RULES_SYMBOLS="Symbols: %d"
JFIELD_PASSWORD_RULES_UPPERCASE="Upper Case: %d"
```

How to check? Change the password requirements in the user settings - add required characters, numbers, upper / lower case, etc.

Break, for example, the line below (each in its own language) and apply the correct wording: 

`JFIELD_PASSWORD_RULES_CHARACTERS="Characters: %d"`

=>

```
JFIELD_PASSWORD_RULES_CHARACTERS_1="%d character"
JFIELD_PASSWORD_RULES_CHARACTERS_2="%d characters"
JFIELD_PASSWORD_RULES_CHARACTERS="%d characters"
```
Turn on registration on the site and try to register a new user. Pay attention to the password hints.

### DIFF
--- a/layouts/joomla/form/field/password.php
+++ b/layouts/joomla/form/field/password.php
@@ -114,27 +114,27 @@ if ($rules)
 
 	if ($minLength)
 	{
-		$requirements[] = Text::sprintf('JFIELD_PASSWORD_RULES_CHARACTERS', $minLength);
+		$requirements[] = Text::plural('JFIELD_PASSWORD_RULES_CHARACTERS', $minLength);
 	}
 
 	if ($minIntegers)
 	{
-		$requirements[] = Text::sprintf('JFIELD_PASSWORD_RULES_DIGITS', $minIntegers);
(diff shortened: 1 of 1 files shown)


brianteeman:
I believe the reason why it was not created as a plural was that unlike a message this is more like the output of a field.

Characters: 1
Symbols: 2
Lowercase: 3
Uppercase: 2
Numbers: 1

and not
1 character
2 symbols
3 lowercase
2 uppercase
1 number

_(but I don't have strong feelings either way)_

PR #34577, opened 2021-06-20
[4.0] Remove com_messages.ini in frontend

Kostelano:
### Summary of Changes
In the front, the old format of variables remains in the file + 1 line is not used in Joomla 4.

Someone can tell WHERE we use this file? I'm not sure that we need it, since the localization of private messages is carried out by a file from the admin panel (yes, in fact, private messages are only there).


brianteeman:
I'm not sure if that file is ever used

Kostelano:
I tested again and couldn't find where the file is being used. I suggest deleting it. Updated PR.

brianteeman:
any file that is deleted needs to be added to the list in com_admin/script.php

Kostelano:
@brianteeman Thanks, updated.

brianteeman:
I have tested this item :white_check_mark: successfully on c341ecee70738d8ac2d51690d8d2b334fa874da3<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34577">issues.joomla.org/tracker/joomla-cms/34577</a>.</sub>

wilsonge:
Thanks!

PR #34580, opened 2021-06-21
[4.0] Typo: descending, not decending

Bakual:
Typo: descending, not decending

Can be merged by review.

Thanks to @horus68 for spotting it while translating.


bembelimen:
Thanks

PR #34585, opened 2021-06-21
Ignore null values in subform rule

laoneo:
### Summary of Changes
The subform rule can get **null** as value to validate. This throws the following warning:
_Warning: foreach() argument must be of type array|object, null given in /var/www/html/j3/libraries/src/Form/Rule/SubFormRule.php on line 52_

~@Fedik can you help here with a test case in core?~

### Testing Instructions
- Add the following code to the file modules/mod_custom/mod_custom.xml after line 27:  
```
<field type="subform" name="subform" multiple="true" label="subform" 
  validate="subform" filter="unset">
  <form>
    <field type="text" name="text" label="text" />
  </form>
</field>
```
- Create a new custom module and save it

The warning mentioned above is added to the logs of the server.

No Warning.

### DIFF
--- a/libraries/src/Form/Rule/SubFormRule.php
+++ b/libraries/src/Form/Rule/SubFormRule.php
@@ -45,7 +45,12 @@ class SubformRule extends FormRule
 		{
 			throw new \UnexpectedValueException(sprintf('%s is no subform field.', $element['name']));
 		}
-
+		
+		if ($value === null)
+		{
+			return true;
+		}
+		
 		$subForm = $field->loadSubForm();
 
 		// Multiple values: Validate every row.


Fedik:
@laoneo do you have a default value for subform field?

Fedik:
the error probably because of wrong JSON  in default value : https://github.com/joomla/joomla-cms/blob/9bd9e5223520f71f1ab6b235e932dbbafd21ff7a/libraries/src/Form/Field/SubformField.php#L445-L456

you can get this error with this field:
```xml
<field type="subform" name="subform" multiple="true" label="subform" 
  validate="subform" default="{'blabla', 'bad json'}">
  <form>
    <field type="text" name="text" label="text" />
  </form>
</field>
```

laoneo:
No default value. The problem is that I do an filter="unset" before. So it is really **NULL** as value.

Fedik:
hmhm, maybe a good idea also modify https://github.com/joomla/joomla-cms/blob/9bd9e5223520f71f1ab6b235e932dbbafd21ff7a/libraries/src/Form/Field/SubformField.php#L451-L455

to:
```php
// The string is invalid json
if (!$value)
{
  return $this->multiple ? [] : null;
}
```

laoneo:
The test function needs to return true or false.

Fedik:
> The problem is that I do an filter="unset" before. So it is really NULL as value.

okay, I not sure we have it in core, but such may be used in other extensions also
then maybe this:

Place it in mod_custom:

```xml
<field type="subform" name="subform" multiple="true" label="subform" 
  validate="subform" filter="unset">
  <form>
    <field type="text" name="text" label="text" />
  </form>
</field>
```

Add `dump($value);exit;` before return here:
https://github.com/joomla/joomla-cms/blob/368a1f39a04d4d6adfabb8f2327988e1bd83f3ab/libraries/src/Form/Rule/SubformRule.php#L86-L88

And try save the module,
**Before** : you will get warning, then dump result;
**After**: only dump result

laoneo:
Ok thanks, updated the testing instructions.

Fedik:
I have tested this item :white_check_mark: successfully on 6a9ebdc09ac8de7ea2b9bbc99c1214c15c5c4561<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34585">issues.joomla.org/tracker/joomla-cms/34585</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on 6a9ebdc09ac8de7ea2b9bbc99c1214c15c5c4561<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34585">issues.joomla.org/tracker/joomla-cms/34585</a>.</sub>

libraries/src/Form/Rule/SubFormRule.php
```diff
@@ -45,7 +45,12 @@ public function test(\SimpleXMLElement $element, $value, $group = null, Registry
 		{
 			throw new \UnexpectedValueException(sprintf('%s is no subform field.', $element['name']));
 		}
-
+		
+		if ($value === null)
```
wilsonge, line 49: We should check if it is null and the field isn’t required I guess (like in other fields)

wilsonge (approved):

libraries/src/Form/Rule/SubFormRule.php
HLeithner: I thought the same, was if any of the fields, but required on subform is already broken, you can't save the form if you set the required attribute.

Fedik: this is done in the upper class,
if the field is required it will not reach the rule

https://github.com/joomla/joomla-cms/blob/db04e2f6aa525dbf6e30b262265fbf8a3d01d843/libraries/src/Form/Form.php#L2134-L2143

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34585">issues.joomla.org/tracker/joomla-cms/34585</a>.</sub>

libraries/src/Form/Rule/SubFormRule.php
laoneo: Exactly @Fedik. That's why I didn't add it here.

HLeithner:
Thanks!

PR #34588, opened 2021-06-21
[4.0] Add support full typizations for base class Table.

korenevskiy:

.
.
.
.


richard67:
@korenevskiy You have made your PR against the staging branch so it shows more than 500 changed files and lots of conflicts. Please rebase to 4.0-dev or make a new PR.

korenevskiy:
> @korenevskiy You have made your PR against the staging branch so it shows more than 500 changed files and lots of conflicts. Please rebase to 4.0-dev or make a new PR.

Thanks.

PR #34605, opened 2021-06-23
[4.0] Fix condition for readmore in newsflash

chmst:
Pull Request for Issue # .

### Summary of Changes
Newsflash module always shows the readmore button, also if it is switched off

### Testing Instructions
Use blog sample data. Swich off the readmore-button in options of mod_articles_news.

Readmore button in the newsflash module is always displayed.

Readmore is displayed only if the param is set


chmst:
@drmenzelit - your find :) 

There is another issue which needs more investigation (b/c). The second parameter which is used in com_content "show the title" is missing in the module. .

brianteeman:
Came back to report that I couldnt see the problem and  I see you have closed it

Issue #34496, opened 2021-06-12
[4.0] Impossible to differentiate between a selectable option and a disabled option in a select html tag


obuisard:
### Steps to reproduce the issue

Make sure the new workflow feature is not enabled on the site.
Go to Content -> Categories.
Edit a category. Go into the workflow tab.
Open the select box 'Workflows'.

### Expected result

You can differentiate between enabled and disabled options.

### Actual result

You cannot differentiate between enabled and disabled options. In this example (sorry, I could not find any better one), '-Workflows-' is disabled in the list.

![screen shot 2021-06-12 at 19 30 23](https://issues.joomla.org/uploads/1/bf22190ac77aaf6cebcdbe55a28c3d5f.png)

### System information (as much as possible)

Any Joomla 4 site.

### Additional comments

The core should take into account this kind of situation (even if rare), especially because third-party extensions may later use 'disabled' options in their fields.

chmst:
In this case, the inherited basic workflow is selectable. This is the Joomla default (publish/ unpublish / trash / archive) and cannot be disabled. 
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34496">issues.joomla.org/tracker/joomla-cms/34496</a>.</sub>

obuisard:
Thank you Christiane for your comment, I appreciate it.

This is actually _not_ my point here: I am talking about the fact that there is no CSS property in the admin template that allows users to differentiate between an option that is enabled and an option that is disabled. In the old school way, we would 'gray out' the disabled field, but this is not enough to make it accessible (not enough contrast).<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34496">issues.joomla.org/tracker/joomla-cms/34496</a>.</sub>

brianteeman:
Please test #34616

obuisard:
Closing this PR in favor of #38817

PR #34616, opened 2021-06-24
[4.0] Disabled options

brianteeman:
Adds a background to disabled items in an option list

PR for #34496

_Not convinced of the need for this myself someone else can decide_

### Before
image

### After
image

### DIFF
--- a/administrator/templates/atum/scss/vendor/bootstrap/_custom-forms.scss
+++ b/administrator/templates/atum/scss/vendor/bootstrap/_custom-forms.scss
@@ -72,5 +72,9 @@
   option {
     color: var(--atum-text-dark);
     background-color: $white;
+
+    &:disabled {
+      background-color: var(--atum-bg-dark-10);
+    }
   }
 }


ChristineWk:
I have tested this item :white_check_mark: successfully on 693a1cc91a73f096a0df0b055d255a352bfd2245<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34616">issues.joomla.org/tracker/joomla-cms/34616</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on 693a1cc91a73f096a0df0b055d255a352bfd2245<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34616">issues.joomla.org/tracker/joomla-cms/34616</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34616">issues.joomla.org/tracker/joomla-cms/34616</a>.</sub>

angieradtke:
> Not convinced of the need for this myself someone else can decide
@brianteeman 

A bit difficult. With the background, the disabled element looks emphasized even more.  I think the background should not be changed. The text color could be lighter, so that sees that thist item  is not selectable.
 But for people with visual impairments this could be difficult.  As always there is no perfect solution

Quy:
Maybe remove the `- Workflows -` option which is redundant since the label is `Worklows`.

chmst:
grafik

Similar fields have headlines and does not need extra background.

drmenzelit:
I'm confused... is `- Workflow -` an option that can be enabled? Or it is a label for the options below?

Quy:
It is hardcoded. I assume it is acting as an `<optgroup>`.

https://github.com/brianteeman/joomla-cms/blob/a5e3a3d1a1ad915b97cd7e65989e197a6756e6ac/administrator/components/com_content/src/Helper/ContentHelper.php#L179

wilsonge:
I have to say I don't think this change looks good - to me it draws attention to the disabled option and borderline makes it look selected.

Is our problem on the original issue just on Safari that the default text colour looks too similar to the default disabled text color? Because if so maybe we just need to tweak the text colour rather than add a background

obuisard:
> Maybe remove the `- Workflows -` option which is redundant since the label is `Worklows`.

That's not what this PR is about. The template needs to allow disabled options to be easily distinguished from options that are selectable. The workflow here is just an example of such a case where an option is disabled and yet you can't tell until you want to click on it.

drmenzelit:
@obuisard as @Quy and @chmst stated above, there is not an option and can't be enabled, the code is wrong `- Workflows -` acts like a title of the group below, like `-- From Component --` in the screenshot from @chmst

obuisard:
> @obuisard as @Quy and @chmst stated above, there is not an option and can't be enabled, the code is wrong `- Workflows -` acts like a title of the group below, like `-- From Component --` in the screenshot from @chmst 

I understand. The point here is not to fix the workflow selection but the template in general, which lacks styling for the disabled state.

PR #34314, opened 2021-05-31
[4.1] Media web service implementation

pjdevries:
[Joomla-Mono-Vertical-logo-light-background-en.png.base64.txt](https://github.com/joomla/joomla-cms/files/6602319/Joomla-Mono-Vertical-logo-light-background-en.png.base64.txt)
[Vertical-logo-light-background-en.png.base64.txt](https://github.com/joomla/joomla-cms/files/6582729/Vertical-logo-light-background-en.png.base64.txt)

Pull Request for Issue #34224 .

### Summary of Changes

My take on a media web service implementation. It relies on com_media's `ApiModel` to provide the logic. Two service models, `MediumModel` and `MediaModel`, provide wrappers around `ApiModel` to provide an interface as expected by the standard web service base classes.

The service recognizes the following endpoints:
```
GET {{base_path}}/api/index.php/v1/media
GET {{base_path}}/api/index.php/v1/media/{path}
POST {{base_path}}/api/index.php/v1/media
PATCH {{base_path}}/api/index.php/v1/media/{path}
DELETE {{base_path}}/api/index.php/v1/media/{path}
```

`{path}` parameters in the above endpoints represent paths of existing files or folders, relative to the media base folder (images by default). They are used in single item requests instead of the usual numeric item `id`, since media items don't have id's.

The web service mostly recognizes the same json request parameters as the com_media `ApiModel`. These are:

`path`: A file or folder path wherever appropriate. For instance, in case of the `list GET` request it points to the base folder, from which to retrieve the list of files and/or folders. Or the destination of a file or folder for a `POST` or `PATCH` request.

`url`: Instructs com_media to include a separate `url` attribute in the resulting file objects. Has no effect on folder objects.

`temp`: Instructs com_media to include a separate `tempUrl` attribute in the resulting file objects. I have no idea what the added value for this parameter is. Has no effect on folder objects.

`content`: Media content in base64 format for `POST` or `PATCH` requests.

In places where `path` is used, it can be prefixed with `adapter:` to access files managed using that specific adapter. For eaxample: `local-images:/banners/osmbanner1.png`, `dpdropbox-Dropbox:/fo/bar/baz.jpg`, etc. If no adapter is specified, `local-images` is assumed.

These are issues I struggled with during the implementation:
- The web service API heavily relies on the notion that all data comes from a database.
- Media files have no database entries and thus no record id.
- `LocalAdapter::createFolder()` ignores the result of `File::create()`, so the sweb service has no clue if creating a new file or folder succeeded.
[...]

PR #34634, opened 2021-06-26
[4.0] Select image/audio/video/document file from Media Manager

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/discussions/32642 and https://www.facebook.com/groups/joomlanospam/posts/10157777796480997/?__cft__[0]=AZXC8G44TJLPO0QeEbytZwSVaZX2YpNFHjOUcx1xUqyJjBUZGqgh763LWT-Lx2xKyvJfIm-LQO6wUquCrqneOLqSAlx3AOpRDAxnMrliJHTTK1AIkT4Mh-BVfJCJjMVFEyDYelefuWE8UMA3X6_-XUFRasmSPT8VNhUVlOln0Ujj4LCrwnBYH_QLfColkn2QEkk&__tn__=%2CO%2CP-R

Apparently, the Joomla community finds it totally fine to use software that is **based on a javascript that's 2 major versions behind**. It's not fine for me thus the PR...

### Summary of Changes
- Reworked the media select Javascript
- Changed the one field for all the supported extensions to multiple fileds in Media Manager (this needs to be communicated as it's a B/C break, and also it needs updates to the sql files)
- Reworked/added file types in the vue App (the files need to enable the preview)

Please extract this 
[Archive.zip](https://github.com/joomla/joomla-cms/files/6720757/Archive.zip)
 to the images directory, or use your own mp3/mp4/pdf files

### Testing Instructions
- Apply the PR or downlad the installable from github

- If you're applying the PR then go to the Media manager options and 
- Make sure the legal mime types are: `image/jpeg,image/gif,image/png,image/bmp,audio/mp3,video/mp4,application/msword,application/excel,application/pdf,application/powerpoint,text/plain,application/x-zip`
- the video extensions are `mp4`
- the audio extensions are `mp3`
- the document extensions are `odg,odp,ods,odt,pdf,png,ppt,txt,xcf,xls,csv`
- and finally the images extensions are: `bmp,gif,jpg,jpeg,png,ico` **This is IMPORTANT**
- 
- Go to tinyMCE and:

- Select the image from the CMS Content dropdown
- Then select the video file, click on the controls and embed checkboxes, change the width and height and then click select image at the bottom of the modal
- Press couple returns type `Video Link` and select the text you just wrote
- Then select the video file, without selecting the controls and embed checkboxes (width and height is irrelevant here) and then click select image at the bottom of the modal
- Check the produced HTML by clicking the toggle button under the editor

- Select the image from the CMS Content dropdown
- Then select the audio file, click on the controls and embed checkboxes and then click select image at the bottom of the modal
- Press couple returns type `Audio Link` and select the text you just wrote
- Then select the audio file, without selecting the controls and embed checkboxes and then click select image at the bottom of the modal
- Check the produced HTML by clicking the toggle button under the editor

- Select the image from the CMS Content dropdown
- Then select the PDF file, click on the embed checkboxe  change the width and height and then click select image at the bottom of the modal
- Press couple returns type `PDF Link` and select the text you just wrote
- Then select the PDF file, without selecting the embed checkboxe and then click select image at the bottom of the modal
- Check the produced HTML by clicking the toggle button under the editor

- your final html should look like:
```html
<p><video controls="controls" width="800" height="600">
<source src="/images/flower.mp4" type="video/mp4" /></video></p>
<p> </p>
<p><a href="images/flower.mp4" download="">Video Link</a></p>
<p> </p>
<p><audio src="images/t-rex-roar.mp3" controls="controls"></audio></p>
<p> </p>
<p><a href="images/t-rex-roar.mp3" download="">Audio Link</a></p>
<p> </p>
<p><object data="images/core_xy_exploded_infographic1.pdf" type="application/pdf" width="800" height="600">
You don't have a pdf plugin, but you can <a download="" href="/images/core_xy_exploded_infographic1.pdf"> download the pdf file.</a></object></p>
<p> </p>
<p><a href="images/core_xy_exploded_infographic1.pdf" download="">PDF Link</a></p>
```

- Check that the insertion of an image is still ok
- Check that the media FIELDS (intro/fulltext images) are still functional
- Finally save the article and visit it in the front end. ~That's it.~ 

- Edit the definition of the intro image https://github.com/joomla/joomla-cms/blob/7f5dd8ee5f6a690a88f05926bef6cb8456055bb8/administrator/components/com_content/forms/article.xml#L735-L739 by adding `types="images,audios"`
- Test the field, you should be able now to select images and audio files, also try to upload an unsupported file (eg video and pdf)
- Repeat with different combinations `types="images,audios,videos,documents"`

Impossible to select/embed anything other than an image file

It's ok to select/embed audio, video, document files
Screenshot 2021-06-26 at 23 04 08
Screenshot 2021-06-26 at 23 04 38
Screenshot 2021-06-26 at 23 04 28
Screenshot 2021-06-26 at 23 04 18

Front end:
j4 local_index php_teststststtsts

There are few todos that I skipped here:
- sql files
- remove renamed files
- rename the xtd button image to media as the image is confusing now

@wilsonge @laoneo any feedback?


brianteeman:
looking forward to testing this tomorrow

administrator/components/com_media/src/Model/ApiModel.php
```diff
@@ -533,17 +533,52 @@ private function isMediaFile($path)
...
-				'upload_extensions',
-				'bmp,csv,doc,gif,ico,jpg,jpeg,odg,odp,ods,odt,pdf,png,ppt,txt,xcf,xls,BMP,CSV,DOC,GIF,ICO,JPG,JPEG,ODG,ODP,ODS,ODT,PDF,PNG,PPT,TXT,XCF,XLS'
+			$this->allowedExtensions = array_merge(
+				array_map(
+					'trim',
+					explode(
+						',',
+						ComponentHelper::getParams('com_media')->get(
+						'image_extensions',
+						'bmp,csv,doc,gif,ico,jpg,jpeg'
```
brianteeman: doc isnt an image ;)

brianteeman:
Tested using your supplied files. Only the pdf uploads. The mp3 and mp4 give a generic cant be uploaded message.
Tested a larger 5mb mp4 and it doesnt appear to do anything. No upload, no error, nothing
Sometimes after one of the failures the session is killed and you have to log in again

brianteeman:
The problem was in your test instructions. Better to clear the existing values in the options and then save in order to get the defaults. If it had been customised at all it would fail

brianteeman:
Click on a pdf file and then click on the three dots
Check the console
```
media-manager.min.js?6a2a37958931d4cd9934c0d049a970a8:1 Uncaught (in promise) TypeError: Cannot read property 'focus' of undefined
    at Proxy.<anonymous> (media-manager.min.js?6a2a37958931d4cd9934c0d049a970a8:1)
```

brianteeman:
mp4 files have previews but there is no preview icon (you need to double click on the file to see the preview)

pdf files do not have previews but you can still double click on the file and the previous item is displayed

administrator/components/com_media/src/Model/ApiModel.php
denysdesign: Somebody drawing or embedding image in doc 😏😂

brianteeman:
If you dont check the box "Use native elements" before inserting then nothing added to the article

richard67:
For files and folder deletion on update with script.php you can leave it to @wilsonge and me. One of us will run the tool to generate the lists of files and folders when this PR has been merged and before the next release after that. So it does not need to be done in this PR. But if you want to do it here because you wanna make it right, I can help later.

For SQL scripts we have to decide how much we want to touch the media manager options on update. Since it will be at least risky and hacky trying to add single values to the json, we might have to completely replace the params so people would lose their settings on update if they have changed some. I can help with making the update SQL but not with 
making the decision if we want that or not.

brianteeman:
@richard67 whichever is decided on the sql there should be a postinstallation message. either saying the settings have been reset to the defaults or saying that you can update the settings to add support for pdf, etc

richard67:
> @richard67 whichever is decided on the sql there should be a postinstallation message. either saying the settings have been reset to the defaults or saying that you can update the settings to add support for pdf, etc

Yeah, that makes sense. Personally I'd prefer the "We don't touch the settings and let the postinstall message tell people that they can update their settings to add support for PDF and so on" approach.

It can be that we have already an update SQL touching these parameters when updating from 3.10. If this is the case, we could modify that one to set them right. I have to check if there is one.

brianteeman:
IIRC we discussed this when I added webp support (or tried to)

richard67:
I've just checked: It seems there is no update SQL script where we already touch the params of the com_media extension. So there is no need to modify such a script.

dgrammatiko:
> If you dont check the box "Use native elements" before inserting then nothing added to the article

Fixed, there are 3 modes: embed, insert asa link with predefined text, insert as a link with selected text from the editor (the 2nd was missing)

> mp4 files have previews but there is no preview icon (you need to double click on the file to see the preview)

~The preview needs refactoring so it can handle all the file types, also I thought that removing the magnifying glass I removed the preview but I totally forgot the double click~ FIXED


> Click on a pdf file and then click on the three dots
> Check the console

Some side effect from copy/pasting, ~investigating~ FIXED

> The problem was in your test instructions. Better to clear the existing values in the options and then save in order to get the defaults. If it had been customised at all it would fail

Will update the instructions for existing installations

A comment about SQL files and B/C:
- Since J4 supports multiple local adapters the `images` adapter should be sniffed on `update` time and the default directory for media files should be `media-uploads` or something similar. Having audio, video or other files under a directory `images` is weird but maybe I'm heretic here. My point is that this can be done in a non B/C breaking way with a post install script for updates. Also doing it this way means that we get rid of runtime code that we have right now to do the adjustment, sounds more solid solution to me anyways.
Anyways I'm ok whatever the decision is here, post install message or not or whatever...

administrator/components/com_media/src/Model/ApiModel.php
dgrammatiko: It's called text art 😂. Obviously copy/paste mistake

richard67:
> A comment about SQL files and B/C:
> 
>     * Since J4 supports multiple local adapters the `images` adapter should be sniffed on `update` time and the default directory for media files should be `media-uploads` or something similar. Having audio, video or other files under a directory `images` is weird but maybe I'm heretic here. My point is that this can be done in a non B/C breaking way with a post install script for updates. Also doing it this way means that we get rid of runtime code that we have right now to do the adjustment, sounds more solid solution to me anyways.
>       Anyways I'm ok whatever the decision is here, post install message or not or whatever...

@dgrammatiko It would be good to have that post install script or procedure. But that would maybe be better a new routine in the script.php than an update SQL script. If necessary @joomdonation could have a look on it.

brianteeman:
Preview of video now working correctly

Preview of pdf - almost - the preview opens but its black. Clicking on rotate makes it appear (maybe that will give a clue)

Audio file will not upload

dgrammatiko:
> Audio file will not upload

Make sure the legal mime types are: `image/jpeg,image/gif,image/png,image/bmp,audio/mp3,video/mp4,application/msword,application/excel,application/pdf,application/powerpoint,text/plain,application/x-zip`

brianteeman:
Checked that twice before posting

`image/jpeg,image/gif,image/png,image/bmp,audio/mp3,video/mp4,application/msword,application/excel,application/pdf,application/powerpoint,text/plain,application/x-zip`

brianteeman:
Article intro and full image
This works when selecting an image (as before)
However you can select a non image file - instead of some type of error message saying "this isnt an image etc" it just silently returns

dgrammatiko:
>  it just silently returns

Yes, I couldn't decide what would be the best scenario for this case so I silently ignored anything non-image. This might need an alert or something similar.

> Checked that twice before posting

There's something fishy as the tests are also failing for no obvious reason

richard67:
@dgrammatiko Here the screenshot from drone from system tests: https://ci.joomla.org/artifacts/joomla/joomla-cms/4.0-dev/34634/system-tests/45401/MediaListCest.showsDefaultFilesAndFolders.mysql.fail.png

It says "syntax error, unexpected ')'".

Here the part of the log with the test where it fails:

```
MediaListCest: Test that it shows the joomla default media files and folders.
89s
325Signature: MediaListCest:showsDefaultFilesAndFolders
89s
326Test: tests/Codeception/acceptance/administrator/components/com_media/MediaListCest.php:showsDefaultFilesAndFolders
89s
327Scenario --
89s
328 I do administrator login 
89s
329  [Cookies] [{"name":"0cd014c79b960a69d3b33ac5245b0833","value":"2c944nm5pml9uatbrlgsjngk43","path":"/","domain":"localhost","expiry":1624893164,"secure":false,"httpOnly":false}]
89s
330  [Snapshot] Restored "ci-admin" session snapshot
89s
331 I create directory "images/test-dir"
89s
332   I get suite configuration 
89s
333   I get suite configuration 
89s
334 Created /tests/www/test-install/images/test-dir
89s
335 I am on page "administrator/index.php?option=com_media&..."
89s
336  [GET] http://localhost/test-install/administrator/index.php?option=com_media&path=local-images:/
89s
337 I wait for media loaded 
90s
338   I wait for element {"class":"media-loader"},3
90s
339   I wait for element not visible {"class":"media-loader"}
90s
340   I wait 1
90s
341 I see contents ["banners","headers","sampledata","joomla_black.png","powered_by.png"]
91s
342   I see element {"class":"media-browser-item"}
91s
343 I delete directory "images/test-dir"
91s
344   I get suite configuration 
91s
345 Deleted /tests/www/test-install/images/test-dir
91s
346 I execute js "window.sessionStorage.removeItem("joomla...."
91s
347  Screenshot and page source were saved into '/********/src/tests/Codeception/_output/' dir
91s
348 FAIL 
```

Does that help?

dgrammatiko:
> Does that help?

Yup, turns out PHP is far away from becoming a functional programming language...

richard67:
@dgrammatiko Now only analysis4x is failing in drone, which is some security stuff using RIPS and is not related to your PR.

There was a general problem with that, but that should be solved now. So if you wanna get rid of the red test and see all green, just update your branch to latest 4.0-dev. But it's not absolutely necessary because we know the drone failure here is that thing.

dgrammatiko:
> Article intro and full image
> This works when selecting an image (as before)
> However you can select a non image file - instead of some type of error message saying "this isnt an image etc" it just silently returns

Now the media field will display **only** images (this is configurable in the code and can be adjusted to fit any combo)

> Audio file will not upload

For this, I need some help, maybe @laoneo ?

Fedik:
Looks good to me (except broken upload for audio/video).

I would suggest to add `webm, mov` for video, and `ogg` for audio, from beginning. Or later we will get a complains about it.

dgrammatiko:
> Looks good to me (except broken upload for audio/video).

Technically this bug is unrelated to this PR, it just happens to become obvious from the changes here 

> I would suggest to add webm, mov for video, and ogg for audio, from beginning. Or later we will get a complains about it.

Yeah, also `webp` images and maybe `avif` (avif is already supported on GD and Imagick but right now you have to compile from source to enable support, but will be available in some next minor release)

brianteeman:
There is already an open PR for webp. Its just stuck in an open state because cropper.js doesnt work in firefox for windows.

dgrammatiko:
@brianteeman can you do a simple check: in your apache configuration there should be a file `mime.types`, could you compare it with this one 
[mime.types.txt](https://github.com/joomla/joomla-cms/files/6725305/mime.types.txt) (backup your existing, remove the `.txt` and place it in the same folder).
I have no clue if php is sniffing the mime types from apache but if it does then there isn't a problem but rather a misconfiguration (I can upload mp3/mp4 without any problems here)

brianteeman:
Rebuilding my site at the moment but I think I may have found the problem

mime.types says 
`audio/mpeg					mpga mp2 mp2a mp3 m2a m3a`

This PR says
`audio\mp3`

Checking the [spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) `audio\mpeg`

So the problem does appear to be related to the pr and not the server

dgrammatiko:
Good catch!

brianteeman:
I can now add mp3 

What is the intended behaviour with the mp3 preview. I just get a filename
image

brianteeman:
image

I find this confusing. I expected the description to be an explanation of the option. The last sentence really describes what to do if you dont want a d/l link.

I would probably have two options
**Embed**
_Use native elements audio, video or object_

**Download**
_Use a download link_

dgrammatiko:
> What is the intended behaviour with the mp3 preview. I just get a filename

Weird, you should get: 
Screenshot 2021-06-28 at 12 41 14

and a markup: 

Screenshot 2021-06-28 at 12 41 34

Does your apache `mime.types` has uncommented the mp3 part? (I think you can drop a mimes.type in the root of the joomla folder, eg grab the one from a couple of comments above)

dgrammatiko:
> I would probably have two options

You mean something like a switcher or use one of the 2 proposals?

brianteeman:
Weird - the preview works now

> You mean something like a switcher or use one of the 2 proposals?

switcher/checkbox/radio - dont care but something that means the user is actively selecting embed or download

N6REJ:
Nice to finally see this.

dgrammatiko:
> switcher/checkbox/radio - dont care but something that means the user is actively selecting embed or download

I went for a radio and it looks like:
Screenshot 2021-06-28 at 14 35 17
Screenshot 2021-06-28 at 14 35 26

brianteeman:
Is there a reason you went for a fixed size embed instead of a responsive embed? (https://embedresponsively.com/)

dgrammatiko:
> Is there a reason you went for a fixed size embed instead of a responsive embed?

The wide/height are not for sizing but rather (as with images) to inform the browser what's the aspect ratio (plain English to allow the browser to correctly calculate the empty space till the video/object is loaded)

For example the code that linked does this (creates an empty 16/9 parent div and then inserts the iframe into it. This is a hack, we don't need to do this anymore, browsers respect and retrieve the correct aspect ratio from the width/height attributes. 

A good resource for width/height from Jen Simmons: https://speakerdeck.com/jensimmons/using-the-html-width-and-height-attributes-to-improve-web-page-loading

brianteeman:
> Is there a reason you went for a fixed size embed instead of a responsive embed?

and thats the good reason - thanks

brianteeman:
I have tested this item :white_check_mark: successfully on 08ce8176e10e9f52f49b97d9491cd36f78677277<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34634">issues.joomla.org/tracker/joomla-cms/34634</a>.</sub>

administrator/components/com_media/config.xml
```diff
@@ -56,6 +48,16 @@
...
 
+		<field
+			name="restrict_uploads_extensions"
+			type="text"
+			label="COM_MEDIA_FIELD_RESTRICT_UPLOADS_EXTENSIONS_LABEL"
+			description="COM_MEDIA_FIELD_RESTRICT_UPLOADS_EXTENSIONS_DESC"
+			size="50"
+			default="bmp,gif,jpg,jpeg,png,ico,mp3,mp4,odg,odp,ods,odt,pdf,png,ppt,txt,xcf,xls,csv"
+			showon="restrict_uploads:1"
+			/>
```
Quy, line 59:
```suggestion
		/>
```

Quy:
```suggestion
		/>
```

administrator/components/com_media/tmpl/file/default.php
```diff
@@ -25,41 +25,47 @@
...
-	'uploadPath'              => $this->file->path,
-	'editViewUrl'             => Uri::base() . 'index.php?option=com_media&view=file' . ($tmpl ? '&tmpl=' . $tmpl : ''),
-	'allowedUploadExtensions' => $params->get('upload_extensions', ''),
-	'maxUploadSizeMb'         => $params->get('upload_maxsize', 10),
-	'contents'                => $this->file->content,
+	'apiBaseUrl'         => Uri::base() . 'index.php?option=com_media&format=json' . $imgOnly,
+	'csrfToken'          => Session::getFormToken(),
+	'uploadPath'         => $this->file->path,
+	'editViewUrl'        => Uri::base() . 'index.php?option=com_media&view=file' . ($tmpl ? '&tmpl=' . $tmpl : '') . $imgOnly,
+	'imagesExtensions'   => explode(',', $params->get('image_extensions', 'bmp,gif,jpg,jpeg,png,ico')),
```
Quy, line 50:
```suggestion
	'imagesExtensions'   => explode(',', $params->get('image_extensions', 'bmp,gif,jpg,jpeg,png,webp')),
```

Quy:
Maybe swap `Embed` and  `Download` radio buttons since `Enable Controls` relates to `Embed`.

dgrammatiko:
> Maybe swap Embed and Download radio buttons since Enable Controls relates to Embed.

Done

administrator/components/com_media/tmpl/file/default.php
```diff
@@ -25,41 +25,47 @@
...
+	'audioExtensions'    => explode(',', $params->get('audio_extensions', 'mp3')),
+	'videoExtensions'    => explode(',', $params->get('video_extensions', 'mp4')),
+	'documentExtensions' => explode(',', $params->get('document_extensions', 'odg,odp,ods,odt,pdf,png,ppt,txt,xcf,xls,csv')),
```
Quy:
```suggestion
	'documentExtensions' => explode(',', $params->get('doc_extensions', 'odg,odp,ods,odt,pdf,png,ppt,txt,xcf,xls,csv')),
```

administrator/components/com_media/tmpl/file/default.php
```diff
@@ -25,41 +25,47 @@
...
-	'allowedUploadExtensions' => $params->get('upload_extensions', ''),
-	'maxUploadSizeMb'         => $params->get('upload_maxsize', 10),
-	'contents'                => $this->file->content,
+	'apiBaseUrl'         => Uri::base() . 'index.php?option=com_media&format=json' . $imgOnly,
+	'csrfToken'          => Session::getFormToken(),
+	'uploadPath'         => $this->file->path,
+	'editViewUrl'        => Uri::base() . 'index.php?option=com_media&view=file' . ($tmpl ? '&tmpl=' . $tmpl : '') . $imgOnly,
+	'imagesExtensions'   => explode(',', $params->get('image_extensions', 'bmp,gif,jpg,jpeg,png,ico')),
+	'audioExtensions'    => explode(',', $params->get('audio_extensions', 'mp3')),
+	'videoExtensions'    => explode(',', $params->get('video_extensions', 'mp4')),
```
Quy: Sync default values?

dgrammatiko: Done

Quy:
Edit an article.
Click CMS Content > Media
Click on an MP3 file.
Click `Embed` radio button.
Click `Enable Controls` checkbox.
The checkbox should not disappear.

dgrammatiko:
> The checkbox should not disappear.

Fixed, thanks

Quy:
Checked/unchecked `Enable Controls` produce the same markup.

`<p><audio src="images/t-rex-roar.mp3" controls="controls"></audio></p>`

dgrammatiko:
> Checked/unchecked Enable Controls produce the same markup.

Fixed, thanks

Quy:
Missing icon in list view.

34634-icon

alikon:
can you take a look at #34314 @dgrammatiko  please

dgrammatiko:
> Missing icon in list view.

Fixed

> can you take a look at #34314

I've seen it already but I don't think I'm the right guy for feedback there. My feeling is that we might want to control the contents of a folder that is returned (mind there's no permissions as in other components so things might have to be done different, eg utilising the same technic as: https://github.com/joomla/joomla-cms/pull/34634/files#diff-8404917e9476fe232ef9c49920ced95e9ce5825e5efd60f955ecde240a2a090aL533-R583 ) or wait till a db layer is added in com_media (or I'm totally wrong...).

Quy:
Select an MP3 file.
Click `Embed` radio button.
Don't click `Enable Controls`.

Here is the markup, but nothing is displayed.

`<audio src="images/t-rex-roar.mp3"></audio>`

dgrammatiko:
> Here is the markup, but nothing is displayed.
> `<audio src="images/t-rex-roar.mp3"></audio>`

Option to add or not controls removed. To my defend here, if you have a JS player you only need the tag with the source as the rest are often controlled by the JS player. Anyhow, if in the future people asking for this option the code is here...

administrator/language/en-GB/com_media.ini
```diff
@@ -48,6 +52,8 @@ COM_MEDIA_FIELD_PATH_FILE_FOLDER_LABEL="Path to Files Folder"
 COM_MEDIA_FIELD_PATH_IMAGE_FOLDER_DESC="Enter the path to the images folder relative to the root of your webspace. This path <strong>has to be the same as path to files (default) or to a subfolder of the path to file folder.</strong> Note: Do not start the path with a slash!"
 COM_MEDIA_FIELD_PATH_IMAGE_FOLDER_LABEL="Path to Images Folder"
 COM_MEDIA_FIELD_RESTRICT_UPLOADS_DESC="Restrict uploads for lower than manager users to images if Fileinfo or MIME Magic isn't installed."
+COM_MEDIA_FIELD_RESTRICT_UPLOADS_EXTENSIONS_DESC="Restrict uploads for lower than manager users to files in the list."
+COM_MEDIA_FIELD_RESTRICT_UPLOADS_EXTENSIONS_LABEL="Allowed extensions"
```
Quy:
```suggestion
COM_MEDIA_FIELD_RESTRICT_UPLOADS_EXTENSIONS_LABEL="Allowed Extensions"
```

brianteeman:
Embedding an object requires and alt  or equivalent see 
https://accessibilityinsights.io/info-examples/web/object-alt/

brianteeman:
The preview of a video is different to a preview of an image with regards to width and the position of the close button. I am assuming its some css

image

image

dgrammatiko:
> Embedding an object requires and alt or equivalent see

~I will implement this later on today~ Implemented

dgrammatiko:
> The preview of a video is different to a preview of an image with regards to width and the position of the close button. I am assuming its some css

The current implementation is using inline style background for the image, thus the difference

Quy:
Close icon overlaps filename.

34634-close

micker:
Hello i tested work very good for me, just 2 notices :
- the additionnal windows doesn't close when we close media
- it will be cool to open media popup when we do right click or double click on media easier to change option
thanks a million @dgrammatiko Now we have a real MEDIA manager !!!

dgrammatiko:
> the additionnal windows doesn't close when we close media

I don't understand what you mean, the media manager opens in a modal, not a new window...

> it will be cool to open media popup when we do right click or double click on media easier to change option

This is not possible for 2 main reasons:
- there is always an extra step after selecting a file (provide alt text or embed or link for download) so it will never be one-step process
- there are 2 client-side applications: the media manager and a tiny Custom Element that asks for the extra info. These 2 cannot be combined into one as the one runs in an iframe and the other one on the main document. To get everything in the same window requires a huge refactor of few parts (doable but not a couple of days work...)

micker:
Ok for point 2
For point sorry for my Bad explain i mean the small zone with option (embed or download) this zone doesnt close when WE close media

dgrammatiko:
>  i mean the small zone with option (embed or download) this zone doesnt close when WE close media

Can you record a video demonstrating what's not working? Seems fine here

micker:
i do small video, that the additionnal option zone
https://www.loom.com/share/b378c2a3c7304547ad577516f65c4742
thnaks

dgrammatiko:
> i do small video, that the additionnal option zone

Ok, I see what you mean and already fixed 3 cases that the aditional data should be dismissed but as long as some file is selected this will be visible. You can click on the title `Aditional Data` though, and that will be minimised. I can't do anything more here

micker:
Ok that already better !

dgrammatiko:
@micker can you please mark you test on the issues site?

brianteeman:
I have tested this item :white_check_mark: successfully on 42d28ef15eb33ab497af98ef671762a4039dcb56<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34634">issues.joomla.org/tracker/joomla-cms/34634</a>.</sub>

micker:
I have tested this item ✅ successfully on 42d28ef too

richard67:
@micker Could you please go to the issue tracker here and mark your test result in the right way? Just copying the message from another test is not sufficient. G to https://issues.joomla.org/tracker/joomla-cms/34634 , use the "Test this" button at the top left corner, select the appropriate test result and then submit. Thanks in advance.

micker:
I have tested this item :white_check_mark: successfully on 42d28ef15eb33ab497af98ef671762a4039dcb56<br /><br />Tested and perfect
Sorry Richard a mistake<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34634">issues.joomla.org/tracker/joomla-cms/34634</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34634">issues.joomla.org/tracker/joomla-cms/34634</a>.</sub>

wilsonge:
This seems to important to not merge right now given it has b/c implications. Although it's not great given our point of RC. Thankyou so much @dgrammatiko for your hard work on this!

brianteeman:
woohoo!!!

dgrammatiko:
Thanks

@wilsonge the B/C break is the change of some of the com_media params and to be honest here even what I did here is not perfect. An ideal solution is to have a data structure like
```json
{
  supported: [
    {
      extension: 'png',
      mine: 'image/png',
      preview: true,
      upload: true
    }
  ]
}
```
That will allow to remove the hardcoded preview extensions in the com_media app (vue) and also if done right (subforms?) it will be easier in the backend (eg have a pool that users enable/disable, instead of typing tings)

api/language/en-GB/joomla.ini
```diff
@@ -241,11 +241,22 @@ JFIELD_LOGOUT_REDIRECT_URL_LABEL="Logout Redirect"
...
 JFIELD_MEDIA_CLASS_LABEL="Image Class"
 JFIELD_MEDIA_FIGURE_CAPTION_LABEL="Figure Caption"
 JFIELD_MEDIA_FIGURE_CLASS_LABEL="Figure Class"
+JFIELD_MEDIA_HEIGHT_LABEL="Height"
 JFIELD_MEDIA_LAZY_LABEL="Image will be lazyloaded"
 JFIELD_MEDIA_SUMMARY_LABEL="Additional Data"
+JFIELD_MEDIA_WIDTH_LABEL="Width"
+JFIELD_MEDIA_TITLE_LABEL="Title"
+; Do not translate the text between the {}
+JFIELD_MEDIA_UNSUPPORTED="You don't have a {extension} plugin, but you can {tag} download the {extension} file.</a>"
```
Kostelano, line 259: I see that it is combined. But is everything true in this line? Confused by the end of the line.

@dgrammatiko

richard67: Hmm, yes ... where does that closing anchor tag belong to?

Kostelano: There is also the same ending in another line. Copy / paste

richard67: @Kostelano Maybe you can make an issue, or if you have more time make a PR to fix it? It would need to get that text shown and then check the original markup. Browser inspection tools might repair the invalid markup and so not show any mistake, so it could be tricky.

dgrammatiko: The first part is here https://github.com/joomla/joomla-cms/blob/83e3dd33a2853a72939fb4e3d682f90ee2ad414d/build/media_source/system/js/fields/joomla-media-select.w-c.es6.js#L235

richard67: I see ... not nice, but ok ... thanks for reporting back.

dgrammatiko: I just realized that the closing tag shouldn't be in the ini file. Well, my bad 😞

richard67: You can make a PR ... but it should be  soon because there might be language freeze for J4 in near future.

administrator/components/com_media/resources/scripts/components/browser/items/document.vue
```diff
@@ -0,0 +1,221 @@
+<template>
+  <div
+    class="media-browser-doc"
+    @dblclick="openPreview()"
+    @mouseleave="hideActions()"
+  >
+    <div class="media-browser-item-preview">
+      <div class="file-background">
+        <div class="file-icon">
+          <span class="fas fa-file-pdf" />
```
Quy, line 10: Can you please improve this to reflect their respective icons instead of a PDF icon for all document types? Thanks.

dgrammatiko: What would be the appropriate icon? Also you can PR the change yourself nothing too fancy here 😉

Quy: pdf icon for .pdf
word icon for .doc
excel icon for .xls
otherwise default to generic file icon for the other extensions

richard67: Don’t forget docx, docm, xlsx and xlsm

brianteeman: You can do this in pure css so that the markup doesnt have to contain everything. See this example https://gschoppe.com/uncategorized/add-auto-sensing-file-type-icons-to-lists-of-downloads-with-fontawesome-and-css/

dgrammatiko: So change line 10 to ` <span v-bind:class="getFileClass" />`
and add after line 44 (after the data(){} ):

```js
  computed: {
    /* Get the hashed URL */
    getFileClass: (item) => {
      switch (item.mime_type) {
        case 'application/pdf':
          return 'fas fa-file-pdf';
        case 'application/msword':
        case 'application/vnd.openxmlformats-officedocument.wordprocessingml.document':
          return 'fas fa-file-word';
        case 'application/vnd.ms-excel':
        case 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet':
          return 'fas fa-file-excel';
        case 'application/vnd.ms-powerpoint':
        case 'application/vnd.openxmlformats-officedocument.presentationml.presentation':
          return 'fas fa-file-powerpoint';
        default:
          return 'fas fa-file';
      }
    },
  },
```

Modify it to your taste

dgrammatiko: > You can do this in pure css so that the markup doesnt have to contain everything.

Actually, on the web, the extensions are totally FOOBAR. The only way to tell what a type of a file really is, is by checking the mime type. The mime type is not available to the HTML/CSS unless the application actually passes it either as a class or a data attribute. 

What I'm saying is that you can have a totally valid PDF (or whatever) without any extension. Actually, Joomla should never ever do anything based on the extensions as this is a huge SECURITY risk. 

/rant

Quy: > and add after line 44 (after the data(){} ):

Should it be after line 177 in the same file?

dgrammatiko: yes: https://github.com/joomla/joomla-cms/blob/b84fc74c882b0b86d220bf6c5b9999e9c1a17958/administrator/components/com_media/resources/scripts/components/browser/items/document.vue#L177

Issue #34446, opened 2021-06-07
[4.0] Frontend: Tag in Article miss top-shadow


ghost:
### What needs to be fixed
Use sample data and open for example `sample-layouts/category-list/welcome-to-your-blog.html`:

![image](https://user-images.githubusercontent.com/78906980/120991149-a18ccd00-c781-11eb-81dd-1ed496b39918.png)

### Why this should be fixed

Look better.

### How would you fix it
Don't know.

brianteeman:
The cause is this line of css but I have no idea what removing it will do elsewhere

```
.com-content-article ol,.com-content-article ul {
     overflow:hidden; 
}
```

drmenzelit:
This line of code was introduced with PR #33624

shahwezahmad:
i want to work on this issue assign me  this

shahwezahmad:
I am interested in this issue assign me t

ghost:
@shahwezahmad Thanks for your interest on this issue. You don't need to be assign to create a pull request.

shahwezahmad:
I am new in open source field  can you suggest me any project to contribute
or anything  which is helpful  to grow me   in this field

ghost:
Similar on "Read more"-Button but not sure if it's worth to open a new issue:

![image](https://user-images.githubusercontent.com/78906980/121773927-7dfdc400-cb7f-11eb-873c-34afcb67e9b6.png)

jatinS-dev:
hello, I want to work on this issue. Can anyone help me how to start?

brianteeman:
Please test #34636

richard67:
Closing as having a pull request. Please test #34636 . Thanks in advance.

PR #33560, opened 2021-05-05
[4.0] Cassiopeia - Add overflow hidden to ul and ol

YatharthVyas:
Pull Request for Issue #33501

### Summary of Changes
Added `overflow: hidden` to `<ul>` and `ol` globally.
Lists that are to the right of images having `float: left` retain their indentations for li, all other lists are not affected by this.

### Testing Instructions
Prerequisites: Install blog sample data
- Apply Patch
- `npm run build:css`
- Frontend: Visit the blog 'Your Modules' and check the list indentation
- Also, check any blog article that has a list without an image. In this case, everything should remain as is.

### Actual result BEFORE applying this Pull Request
image

### Expected result AFTER applying this Pull Request
image

### Documentation Changes Required
None

PR #34636, opened 2021-06-27
[4.0] specific css

brianteeman:
This PR corrects the change in #33560 which caused the issue reported in #34446

To test with the sample data.
Check that the readmore button on blog views has a complete border when it is active
Check that the tag buttons have a complete border when it is active

Dont forget to rebuild the scss

### DIFF
--- a/templates/cassiopeia/scss/blocks/_global.scss
+++ b/templates/cassiopeia/scss/blocks/_global.scss
@@ -119,7 +119,7 @@ a {
   }
 }
 
-.com-content-article {
+.com-content-article__body {
   ol,
   ul {
     overflow: hidden;


RickR2H:
The overflow of the tags is fixed, but not on the read more button.
The problem is the overflow hidden in _modifiers.scss on line 114
screen shot 2021-06-29 at 21 24 16<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34636">issues.joomla.org/tracker/joomla-cms/34636</a>.</sub>

PR #34646, opened 2021-06-28
[5.2] Fixed missing maxlength attribute in textarea.php

machadoug:
### Summary of Changes
The Max Length was printing just the NUMBER as the attribute instead of maxlength="NUMBER"

### Testing Instructions
See screenshot

### Actual result BEFORE and AFTER applying this Pull Request

https://user-images.githubusercontent.com/1613200/123715488-c9d28d80-d84e-11eb-831f-d359b3a1f4a5.mp4

None


layouts/joomla/form/field/textarea.php
```diff
@@ -78,7 +78,7 @@
 	!empty($autocomplete) ? 'autocomplete="' . $autocomplete . '"' : '',
 	$autofocus ? 'autofocus' : '',
 	$spellcheck ? '' : 'spellcheck="false"',
-	$maxlength ?: '',
+	$maxlength ? 'maxlength="'.$maxlength.'"' : '',
```
Quy:
```suggestion
	$maxlength ? 'maxlength="' . $maxlength . '"' : '',
```

Quy:
I am not able to reproduce this issue.

Created a custom text area field with maxlength value.
The markup has the maxlength attribute.

machadoug:
@Quy , according to the textarea field layout code this will only work if the ```$maxlength``` has the entire attribute instead of just an integer, for example, if the ```$maxlength = 'maxlength="1000"'``` 

I just noticed there's no ```@var``` documentation, but IMHO it should work just like the other variables, which contain the value only instead of the ```attribute_name="attribute_value"```

Quy:
Your PR will break the custom textarea field. See markup.

```
<textarea name="jform[com_fields][text-area]" id="jform_com_fields_text_area"   class="form-control"           maxlength=" maxlength="1000""   ></textarea>
```

brianteeman:
The character count requires two attributes in the xml
```
				maxlength="160"
				charcounter="true"
```

Quy:
@machadoug You have the same issue with `rows` (4) and `cols` (20) attributes displaying only the values. It is not happening in core so it must be on your end. Please provide steps/code to reproduce.

Fedik:
@machadoug I suspect you use the textarea layout with your own field,
because for core `textarea` field the `maxlength` is set in the class:

https://github.com/joomla/joomla-cms/blob/3544d8bef8f066a51463b0ee2fc079ddb494d93f/libraries/src/Form/Field/TextareaField.php#L180-L189

It looks strange to me, but it works like that since ~3.7 version

machadoug:
I was indeed using a custom Textarea field in my extension. I was under the impression that we should use only the attribute value, not ```attribute="value"```. Like the onchange, onclick and autocomplete attributes.

Maybe we should include something in the code documentation explaining it's a string and not an integer?
```php
 * @var   string   $rows         rows attribute for the field.
 * @var   string   $cols         cols attribute for the field.
 * @var   string   $maxlength    maxlength attribute for the field.
```

Fedik:
>  I was under the impression that we should use only the attribute value, not attribute="value". Like the onchange, onclick and autocomplete attributes.

That is correct expectation. 
But I cannot say why it like that for textarea.

Would be good to fix it. Only need to care about backward compatibility here.
The fix should move this 
https://github.com/joomla/joomla-cms/blob/3544d8bef8f066a51463b0ee2fc079ddb494d93f/libraries/src/Form/Field/TextareaField.php#L179-L181
to the layout.

And in layout doing something like:
```php
$maxlengthAttr = '';

if ($maxlengt)
{
  $maxlengthAttr = is_numeric($maxlength) ? ' maxlength="' . $maxlength . '"' : $maxlength;
}

```
The same for rows and cols attribute.

machadoug (commented):
Added attributes to doc block and fixed according to @Fedik suggestions.

I just didn't include the changes to joomla-cms/libraries/src/Form/Field/TextareaField.php because I don't know how to include it to this PR.

Fedik:
@machadoug I have added that changes to your PR

It ready for testing:
Make sure textarea field still working as before.

layouts/joomla/form/field/textarea.php
```diff
@@ -65,8 +68,8 @@
 }
 
 $attributes = array(
-	$columns ?: '',
-	$rows ?: '',
+	$columns ? (is_numeric($columns) ? ' cols="' . $columns . '"' : $columns) : '',
+	$rows	? (is_numeric($rows) ? ' rows="' . $rows . '"' : $rows) : '',
```
Quy:
```suggestion
	$columns ? (is_numeric($columns) ? 'cols="' . $columns . '"' : $columns) : '',
	$rows ? (is_numeric($rows) ? 'rows="' . $rows . '"' : $rows) : '',
```

layouts/joomla/form/field/textarea.php
```diff
@@ -78,7 +81,7 @@
 	!empty($autocomplete) ? 'autocomplete="' . $autocomplete . '"' : '',
 	$autofocus ? 'autofocus' : '',
 	$spellcheck ? '' : 'spellcheck="false"',
-	$maxlength ?: '',
+	$maxlength ? (is_numeric($maxlength) ? ' maxlength="' . $maxlength . '"' : $maxlength) : '',
```
Quy:
```suggestion
	$maxlength ? (is_numeric($maxlength) ? 'maxlength="' . $maxlength . '"' : $maxlength) : '',
```

layouts/joomla/form/field/textarea.php
```diff
@@ -65,8 +68,8 @@
 }
 
 $attributes = array(
-	$columns ?: '',
-	$rows ?: '',
+	$columns ? (is_numeric($columns) ? 'cols="' . $columns . '"' : $columns) : '',
```
Quy: Is it necessary to check if numeric?

Fedik: I think yes, because in past TextareaField set it as string `cols="12"` instead just number `12`,
just to be safe :)

machadoug: @Fedik That's correct.

HLeithner:

chmst:
I have tested this item :white_check_mark: successfully on 7b37d37f0eafea0cbaf1dfe13247e53004f8c162<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34646">issues.joomla.org/tracker/joomla-cms/34646</a>.</sub>

libraries/src/Form/Field/TextareaField.php
```diff
@@ -175,15 +175,10 @@ protected function getLayoutData()
     {
         $data = parent::getLayoutData();
 
-        // Initialize some field attributes.
-        $columns      = $this->columns ? ' cols="' . $this->columns . '"' : '';
-        $rows         = $this->rows ? ' rows="' . $this->rows . '"' : '';
-        $maxlength    = $this->maxlength ? ' maxlength="' . $this->maxlength . '"' : '';
-
         $extraData = array(
```
Quy:
```suggestion
        $extraData = [
```

libraries/src/Form/Field/TextareaField.php
```diff
@@ -175,15 +175,10 @@ protected function getLayoutData()
...
-            'maxlength'    => $maxlength,
-            'rows'         => $rows,
-            'columns'      => $columns,
+            'maxlength'    => $this->maxlength,
+            'rows'         => $this->rows,
+            'columns'      => $this->columns,
             'charcounter'  => $this->charcounter
         );
```
Quy:
```suggestion
        ];
```

layouts/joomla/form/field/textarea.php
```diff
@@ -65,8 +68,8 @@
 }
 
 $attributes = array(
```
Quy:
```suggestion
$attributes = [
```

layouts/joomla/form/field/textarea.php
```diff
@@ -78,7 +81,7 @@
     !empty($autocomplete) ? 'autocomplete="' . $autocomplete . '"' : '',
     $autofocus ? 'autofocus' : '',
     $spellcheck ? '' : 'spellcheck="false"',
-    $maxlength ?: '',
+    $maxlength ? (is_numeric($maxlength) ? 'maxlength="' . $maxlength . '"' : $maxlength) : '',
     !empty($counterlabel) ? $counterlabel : '',
     $dataAttribute,
 );
```
Quy:
```suggestion
];
```

libraries/src/Form/Field/TextareaField.php
```diff
@@ -175,16 +175,11 @@ protected function getLayoutData()
...
-
         $extraData = [
-            'maxlength'   => $maxlength,
-            'rows'        => $rows,
-            'columns'     => $columns,
-            'charcounter' => $this->charcounter,
+            'maxlength'    => $this->maxlength,
+            'rows'         => $this->rows,
+            'columns'      => $this->columns,
+            'charcounter'  => $this->charcounter,
```
Quy:
```suggestion
            'maxlength'   => $this->maxlength,
            'rows'        => $this->rows,
            'columns'     => $this->columns,
            'charcounter' => $this->charcounter,
```

HLeithner:

HLeithner:

Quy:
I have tested this item :white_check_mark: successfully on 82dc13fe6063701228c166d24b5c26826d503946<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34646">issues.joomla.org/tracker/joomla-cms/34646</a>.</sub>

pe7er:
Thank you for submitting this PR.

However, do we really need it? 
The file ``/libraries/src/Form/Field/TextareaField.php`` is already casting the parameters ``rows``, ``columns`` and ``maxlength`` to integers, isn't it?
```php
public function __set($name, $value)
    {
        switch ($name) {
            case 'rows':
            case 'columns':
            case 'maxlength':
                $this->$name = (int) $value;
                break;

            case 'charcounter':
                $this->charcounter = strtolower($value) === 'true';
                break;

            default:
                parent::__set($name, $value);
        }
    }
```

rdeutz:
Just tested it with a custom field, seems to me obsolte now. Closing it. Can be reopened if I missed something

Thanks.

joomdonation:
This is a backward incompatible change if someone override the textarea layout for some reasons.  The PR was from long time ago and he developer could easily modified his field to pass data in the required format like how we do in core. In addition, we can make a new PR to add the missing docblock for $rows, $columns and $maxlength variables and clearly state that it needs to be a string in the format attributename="value"

rdeutz:
I have looked into it again. It is consistend (kind of) for some attributes we are setting the "attribute=value" in the fields class and for some other attibutes we do it in the layout. E.g. cols, rows, maxlenght, inputmode are set in the class while onchange, onclick are set in the layout. It would be good, if we had it done always in the same way but changing this now will break a lot of sites. So I am closing this (again).

Issue #34527, opened 2021-06-16
SEF Not Working For Tags In Category Blog


David2376:
### Steps to reproduce the issue
Create a tag in Components -> Tags
Assign tag to an article
Create a menu item of category blog and set to display the tag created above
Navigate to the menu item

### Expected result
URL should be SEF

### Actual result
URL is not SEF.
A query is attached to the URL 
?filter_tag[0]=1

### System information (as much as possible)
Joomla 4.0.0-rc1
PHP 7.4.16

### Additional comments
I have found this issue dating back to early versions of Joomla. It appears that it was fixed but is broken again in all Joomla 4 releases.

ditsuke:
Hello, I've been attempting to fix this issue! 
I've gone through site router and then the routers for `com-content` and `com-tags` but haven't been able to find the SEF route build rules/code responsible for handling tags or filters yet. Can someone help me find the responsible code?

Quy:
Please test PR #34652. Thanks.

wilsonge:
Reopening this as the proposed patch isn't acceptable. But this bug still needs fixing

brianteeman:
Closed as resolved with #40225

PR #34652, opened 2021-06-29
[4.0] Clean filter_tag query from SEF URLs

ditsuke:
Pull Request for Issue #34527.

This should fix occurrence of unwanted queries in Menu links with SEF URLs enabled.

### Summary of Changes
Unset filter_tag query in SiteRouter's SEF route builder.

### Testing Instructions
Create menu item that filters articles by tag. If SEF is enabled, URLs currently retain an unwanted `filter_tag`
query.

Menu items that link to tag filtered category blogs are polluted with `filter_tag` query with SEF URLs enabled.

Menu items that link to tag filtered results shouldn't have redundant `filter_tag` query.

None


RickR2H:
I have tested this item :white_check_mark: successfully on 3ac41dedf391c137ec5a1cc88e86f918f0a869ad<br /><br />Patch removed the unwanted query.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34652">issues.joomla.org/tracker/joomla-cms/34652</a>.</sub>

Fedik:
You should not do it in SiteRouter, 
it need be done in component router level, somehow

richard67:
@Hackwar Could you have a look on this PR here and the linked issue #34527 ? Is it really an issue, and is this here the right way to fix it?

ditsuke:
> You should not do it in SiteRouter,
> it need be done in component router level, somehow

Would it be more appropriate if I implement it in `Router\Rules\NomenuRules`' build rule or a new `RulesInterface` implementation in `com_content` ?

Hackwar:
This is definitely not the right place to remove this. This would be something that would have to be implemented by the component router. This would have to probably be fixed in the StandardRules.php. But please don't simply remove the filter_tag, but cycle through the menu items query elements and remove them one by one.

libraries/src/Component/Router/Rules/StandardRules.php
```diff
@@ -211,16 +209,11 @@ public function build(&$query, &$segments)
...
-					unset($query['layout']);
-				}
+				$this->cleanupQuery($query, $item);
 
 				return;
 			}
-
-			if (isset($query[$view->key]) && $item->query[$view->key] == (int) $query[$view->key])
+			elseif (isset($query[$view->key]) && $item->query[$view->key] === (int) $query[$view->key])
 			{
```
richard67: The previous “if” ends with a “return” statement, so there is no need to change the “if” to an “elseif” here.

ditsuke: Yes I realized, but an `elseif` felt more readable within the context. Do you propose I should change it back?

richard67: Yes I propose to change back because there were deliberate changes in past for such cases. It is not really better readable for people who understand PHP and can see the return statement 3 lines or so above.

ditsuke: Reverted.

ditsuke:
@Hackwar I've implemented changes in the class you suggested. Can you have a look again?

libraries/src/Component/Router/Rules/StandardRules.php
```diff
@@ -308,11 +298,38 @@ public function build(&$query, &$segments)
...
-				unset($query['layout']);
+				continue;
+			}
+			elseif (isset($item->query[$key]) && $item->query[$key] === $value)
+			{
+				unset($query[$key]);
 			}
 		}
 	}
+
```
richard67: Remove this empty line.

libraries/src/Component/Router/Rules/StandardRules.php
```diff
@@ -308,11 +298,38 @@ public function build(&$query, &$segments)
...
+	 */
+	private function cleanupQuery(array &$query, MenuItem $item)
+	{
+		foreach ($query as $key => $value)
+		{
+			 /*
+			 ** These keys are either needed by other build rules or
+			 ** handled exceptionally
+			 */
+			// ? should the 'view' key be included?
```
richard67: If it is not clear to you whether the view shall be included or not, a comment buried in the source code is not the right way to ask.

Please remove that comment.

@Hackwar Can you advise regarding the view parameter? Thanks in advance.

libraries/src/Component/Router/Rules/StandardRules.php
```diff
@@ -308,11 +298,38 @@ public function build(&$query, &$segments)
...
-			if (isset($query['layout']) && $mLayout === $query['layout'])
+	/**
+	 * Strips $query of unneeded parameters
+	 *
+	 * @param   array    $query   The query to clean
+	 * @param   MenuItem $item    The menu item corresponding to the query
+	 *
+	 * @return void
+	 *
+	 * @since 4.0
```
Quy:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

wilsonge:
This just doesn't look right to me. Query parameters are used in various places even in core (e.g. the layout param, tpl param, I'd also guess (but not 100%) that the pagination on tables uses it too). And we can't predict what 3rd parties are going to do. What problem are you trying to solve here?

libraries/src/Component/Router/Rules/StandardRules.php
ditsuke: Thank you, I've made the change. :)

libraries/src/Component/Router/Rules/StandardRules.php
ditsuke: Apologies! That was not supposed to be in the PR, I've cleaned it up.

ditsuke:
> This just doesn't look right to me. Query parameters are used in various places even in core (e.g. the layout param, tpl param, I'd also guess (but not 100%) that the pagination on tables uses it too). And we can't predict what 3rd parties are going to do. What problem are you trying to solve here?

@wilsonge 
I think it's undesirable to have redundant queries in SEF URLs when the corresponding menu entry already has them. The linked issue had an instance of the `filter_tag` queries being retaining when the SEF route already accounted for it. 
My initial change only stripped the `filter_tag` queries (in the SiteRouter) but on suggestion of Hackwar I implemented stripping of all queries also found identical in the Menu entry in StandardRules. If this is breaking something, I suppose we need to place the cleanup at the step where other routines are already supposed to have dealt with the queries in their own ways and as this to ensure the final URL reflected in the front end is clean (it _will_ retain queries the menu entry doesn't have).

As for the layout param, if you look at `cleanupQuery()` the behavior is consistent with how it was being unset before.

joomleb:
Hi guys, 
tested with success. Please, Will this file be added to the 4.0 stable release ?

Hackwar:
This is not the right solution to the problem. Please don't add this. It breaks more than it fixes.

ditsuke:
> 
> 
> This is not the right solution to the problem. Please don't add this. It breaks more than it fixes.

@Hackwar I would really like to get this issue fixed. If this approach breaks things, can you please point me in the right direction to solve this?

joomleb:
> This is not the right solution to the problem. Please don't add this. It breaks more than it fixes.

Hi @Hackwar 
I suppose you are referring to:
> This is definitely not the right place to remove this. This would be something that would have to be implemented by the component router. This would have to probably be fixed in the StandardRules.php. But please don't simply remove the filter_tag, but cycle through the menu items query elements and remove them one by one.

Please, What breaks did you find?

libraries/src/Component/Router/Rules/StandardRules.php
```diff
@@ -308,10 +298,35 @@ public function build(&$query, &$segments)
...
+	 * @since __DEPLOY_VERSION__
+	 */
+	private function cleanupQuery(array &$query, MenuItem $item)
+	{
+		foreach ($query as $key => $value)
+		{
+			 /*
+			 ** These keys are either needed by other build rules or
+			 ** handled exceptionally
+			 */
```
PhilETaylor:
```suggestion
			 // These keys are either needed by other build rules or handled exceptionally
```

joomleb:
Hi guys, 
Is there any chance to see this fix added in 4.1 version ?

joomleb:
Hi guys,
sorry, but, What about this fix and 4.1 version release ? Will it be added ?
What is planned for this issue ?

joomleb:
Hi guys,
Please, Is there any merge plan here to the next upcoming ?

brianteeman:
Presumably not unless the comment from @Hackwar is resolved

joomleb:
Hi @brianteeman 
thanks for reply. Yes, it seems like @ditsuke is still waiting more details from @Hackwar...

HLeithner:

level420:
So @Hackwar did not answer since about a year now. I'm also waiting for this fix, but nothing happens. How to proceed with this issue/PR?

joomleb:
> > This is not the right solution to the problem. Please don't add this. It breaks more than it fixes.
> 
> @Hackwar I would really like to get this issue fixed. If this approach breaks things, can you please point me in the right direction to solve this?

We stop here, @ditsuke comment on 16 Aug 2021


> So @Hackwar did not answer since about a year now. I'm also waiting for this fix, but nothing happens. How to proceed with this issue/PR?

A bug discussed over and over by many people and literally "abandoned".
Since August 2005 it is the first time that I happen to see such a thing in Joomla...

joomleb:
@ditsuke  @chmst  @HLeithner  @rdeutz 
I suppose it is time to change [PR-4.2-dev](https://github.com/joomla/joomla-cms/labels/PR-4.2-dev) label to [PR-4.3-dev](https://github.com/joomla/joomla-cms/labels/PR-4.3-dev) here.

Please, Does anyone have any news on this ticket ?

HLeithner:
> I suppose it is time to change [PR-4.2-dev](https://github.com/joomla/joomla-cms/labels/PR-4.2-dev) label to [PR-4.3-dev](https://github.com/joomla/joomla-cms/labels/PR-4.3-dev) here.

Unlikely it goes to 4.3 nor 4.4, when rebased I would change it to 5.0, but that has to be clarified first

wilsonge:
I've reopened the original issue and am closing this PR. I'm sorry but forcibly removing all query params without a match in the router at the library level is a recipe to breaking things in 3rd party components - many of which use query params for things such as pagination.

HLeithner:
@wilsonge if I read the code correctly only query parameter which are set in the menu will be removed and only if the value is equal. At least what I have seen in the `cleanupQuery()` function

What did I missed?

wilsonge:
https://github.com/joomla/joomla-cms/pull/34652/files#diff-10c2ab3e362cc27a7f0e2b18ee0b59c909041a35db8c9c00abe617e8cd982061R279

Maybe I'm loosing it but that keeps the 3 mentioned ones and unsets everything else? Hence `filter_tag` isn't explicitly listed despite that's the problem it solves

HLeithner:
Nope, look at https://github.com/joomla/joomla-cms/pull/34652/files#diff-10c2ab3e362cc27a7f0e2b18ee0b59c909041a35db8c9c00abe617e8cd982061R281

it only unset get entries which are not in the menu item explicit set and has the same value as the get parameter provided to the router.

joomleb:
Hi guys,
this should be solved by the #40225 fixing, Right?

ditsuke:
> Hi guys,
> this should be solved by the #40225 fixing, Right?

Should be. It's essentially the same as this one except it doesn't generalise for parameters beyond filter_tag.

Issue #34639, opened 2021-06-28
[4.0] Not possible to use an icon as menu item


chmst:
### Steps to reproduce the issue
A user wants to build a menu with icons instead of texts. 

He enters "icon-eye" as link class (or another class see issue #30793)
and sets displlay menuTitle off

![grafik](https://user-images.githubusercontent.com/1035262/123604516-302dc080-d7fb-11eb-9de3-968e78db2116.png)

### Expected result
Same behaviour as with images:  
If display menuTitle is off, show only the icon with visually hidden Menu Title as Text

### Actual result
The icon is displayed (no aria-hidden = true), the menu Item title also is displayed (but with wrong font) also if "display menu title" is off

![grafik](https://user-images.githubusercontent.com/1035262/123610057-86513280-d800-11eb-8c40-b986e88ea887.png)

brianteeman:
Has this ever been possible? I know in Joomla 3 I used an image to achieve this

brianteeman:
The issue of the wrong font when the icon class is used has been raised before. I believe it is caused because fontawesome is classed as a serif font

chmst:
Same in J3 - probably never tried this before without override.

chmst:
If it were possible to use an icon in the same way as an image it can be made a11y and displayed independent of the menu item title.

brianteeman:
At this time that would need quite a bit of markup change in the menu module. Because of the way the icon changes the font it needs be a separate class and not a class on the link itself. Which also means it would have to be a separate field so that you could still add a link class

drmenzelit:
https://github.com/joomla/joomla-cms/issues/32471

brianteeman:
The _problem_ is that the link class adds a class to the entire link when all you want is to add a new span just for the icon. This could be achieved with an override. To change it for core it would require a lot more as you would need to keep the link class and then add another field for an icon class and its associated markup.

Another option without changing markup is you could change the css as below. Replacing Roboto with your own site font

```
.fa, .fas, [class*=" icon-"], [class^=icon-] {
    font-family: Font Awesome\ 5 Free, Roboto;
    font-weight: 900;
}
```

chmst:
Please test #34658

Issue #32471, opened 2021-02-20
[4.0b7] adding an icon changes the font


Magnytu2:
Hello following the addition of an icon on a menu link or a module, the font changes.

![Capture d’écran 2021-02-19 à 09 17 15](https://user-images.githubusercontent.com/26197929/108592983-3ef9f780-7371-11eb-8ea9-5a79c8a17412.png)

ReLater:
I presume that you used the field "Link Class" of the menu item where you entered a font class like "fa fa-home". Because it's the class of the anchor `<a>` I would say the behavior is correct somehow and I would use an override. 

Setting no image but fill the Image Class and use it in override as a SPAN class if no image provided.

Or maybe easier: CSS to force that only the `::before` part of the `<a>` uses `font-family: "Font Awesome 5 Free";`.

As long as there is no specific setting "Icon Class" or something similiar I think any other solution in core would be just strange when a user defines a "Link Class"

Magnytu2:
You are a thousand times right. You must actually use Font Awesome 5. fa fa-homme> fas fa-home
How to put a small space to the right of the icon? One space is not enough.

Magnytu2:
@ReLater After checking, it works fine on Safari, but not Firefox or Chrome.
![Capture d’écran 2021-02-21 à 16 45 39](https://user-images.githubusercontent.com/26197929/108630322-6b427080-7464-11eb-834e-2f39a5880ea7.png)

ReLater:
The CSS is clear here. 

The HTML

`<a href="/" class="fa fa-home" aria-current="page" tabindex="0">Home</a>`

The CSS for `<a>` AND its `::before` part
```
.fa, .fas, [class^="icon-"], [class*=" icon-"] {
    font-family: "Font Awesome 5 Free";
    font-weight: 900;
}
```
From my point of view your "Accueil" is not the same font-family like "Blog" (compare the lower `l`s.) Any browser falls back in another font-family in cases like this if it doesn't know what to do ;-)

drmenzelit:
@N6REJ as the icon expert can you maybe have a look here?

brianteeman:
iirc you need to use just "fa-home" as the class and then make sure that there is a class which will still target the icon something like `.mod-menu li [class^="fa-"]::before`

its a css issue not anything really specific to fonts

chmst:
Cosing as the pr  #34658 should solve this issue

PR #34658, opened 2021-06-30
[4.x] Menu Items with icon only

chmst:
Pull Request for Issue #34639 .

### Summary of Changes
Enable usage of icons only in frontend menus. The user can now define an icon (for example icon-home or fas fa-user ) and build a menu using only icons without text.

NOTE: This PR does NOT care for code which seems copied for many years from one menuType to the other and results in lots of  code. 

The PR
- Adds a new field to all  menu types: menu_icon_css
- Uses this in all menuTypes in the same way - this means that the menuItems which now are different in some way now have the same behaviour.

A menuItem can be just the title
or be an image with or without the title 
or be an icon with or without the title (this is new).

If a menu Item is only an icon, the menuitem title is added visually hidden (a11y)
If a menuItem is not hidden, the icon gets the aria-hidden="true"

### Testing Instructions
Build menus for example with blog sample data.
Change menu Items in main menu and blog menu (dropdown) and play with params: 

Before patch- see #34639 .

Example 

grafik

yes


dgrammatiko:
Imposing loading the font awesome icon in the front end due to some option on the back end is a **terrible idea**, (also this will never work on any template that doesn't load any icon fonts).
Either:
- use Jlayouts for the markup
- use svg icons, it's 2021 and we should stop using Icon fonts in the front end!

brianteeman:
if you want to use an svg you can already do this.
this pr does not impose font awesome at all. it will work with any icon font the template wants to use
obviously the template would have to support an icon font for it to work

brianteeman:
I have tested this and it does exactly what it says it will do. This is a good addition

chmst:
@dgrammatiko we don't add icon fonts. 

The issue was: users now add icons via the class for the link (see the issue #30793, #32471 and others ). 
This causes a11y error and makes it impossible to build a  menu with icons only. 
This PR only let the user separate the icon from menu title. 

I agree that this should be a Layout - at the moment the code is repeated in different places. But this is not in scope of this PR

brianteeman:
To avoid confusion with people trying to type a path to an icon etc why not label it icon class

image

chmst:
@brianteeman thank you, will do in the next step. There is another language string 
COM_MENUS_ITEM_FIELD_ICON_TITLE_LABEL="Link Icon Class"
Maybe we can use this one in all layouts.

ChristineWk:
> 
> 
> I have tested this and it does exactly what it says it will do. This is a good addition

confirm above.

> 
> 
> @dgrammatiko we don't add icon fonts.
> 
> The issue was: users now add icons via the class for the link (see the issue #30793, #32471 and others ).
> This causes a11y error and makes it impossible to build a menu with icons only.
> This PR only let the user separate the icon from menu title.
> 
> I agree that this should be a Layout - at the moment the code is repeated in different places. But this is not in scope of this PR

tested on 4.0.0-rc4-dev and confirm:
a) icon, e.g. fa fa-home & no title = OK.
b) icon, e.g. fa fa-search & no title = OK, but changes the font, as it is class for a link. (it's known, see above issues)

Above tests, shows only a summary.

chmst:
@ChristineWk thanks for testing. 
Your case b) If the user enters an icon class into the link - this is a problem (the same as in blog layout, if a user says num_columns = 3 and enters columns-2 into the class field.  No clue what to do then.

@brianteeman your opinion, please: We cannot avoid that a user adds an icon class AND an image. I had the image with higher proirity. 

These language keys exist already
COM_MENUS_ITEM_FIELD_ICON_TITLE_DESC="If an icon class is entered, it takes precedence over the link image."
COM_MENUS_ITEM_FIELD_ICON_TITLE_LABEL="Link Icon Class"

brianteeman:
> We cannot avoid that a user adds an icon class AND an image.

Why not? 
```
			<field
				name="menu_image"
				type="media"
				label="COM_MENUS_ITEM_FIELD_MENU_IMAGE_LABEL"
				showon="menu-anchor_icon:"
			/>
```

Also for consistency please rename the new fields menu_* instead of menu-*

ChristineWk:
With Patch got now (for menus) in frontend:

`Notice: Undefined property: Joomla\CMS\Menu\MenuItem::$menu_icon in /example.org/templates/cassiopeia/html/mod_menu/dropdown-metismenu_component.php on line 44 link of menu items`

will try to load a new Nightly and try to remove the overrides I did before

chmst:
@brianteeman the showon is possible of course, but li feel it more confusing than helpful, especially if a user changes his mind while he is building a menu. 

The more I dive into these scripts, the more I see how it is not consistent. For example - the menuItem title also is used as alt-text for the image. This is not really wrong, but also is not as clear as we have it now for other images. 

As @dgrammatiko says, these different menuIten type layouts should be improved, but not now in J4.0

modules/mod_menu/tmpl/default_component.php
```diff
@@ -41,21 +41,38 @@
...
+		// The link is an icon
+		if ($itemParams->get('menu_text', 1))
+		{
+			// If the link text is to be displayed, the icon is added with aria-hidden
+			$linktype = '<span class="p-2 ' . $item->menu_icon . '" aria-hidden="true"></span>' . $item->title;
+		}
+		else
+		{
+			// If the icon itself is the link, it needs a visually hidden text
+			$linktype = '<span class="p-2 ' . $item->menu_icon . '" ></span><span class="visually-hidden">' . $item->title . '</span>';
```
Quy:
```suggestion
			$linktype = '<span class="p-2 ' . $item->menu_icon . '"></span><span class="visually-hidden">' . $item->title . '</span>';
```

Quy: Please change the other instances too.

brianteeman:
I have tested this item :white_check_mark: successfully on 5a8291ca01f7969929b721cd9c822342666fe991<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34658">issues.joomla.org/tracker/joomla-cms/34658</a>.</sub>

ChristineWk:
Still getting:

`Notice: Undefined property: Joomla\CMS\Menu\MenuItem::$menu_icon in /example.org/templates/cassiopeia/html/mod_menu/dropdown-metismenu_component.php on line 44 link of menu items`

Tried also with Prebuilt Package and then installed Nightly from today.

chmst:
Thanks @ChristineWk! 
Don't know why this was not in the latest commit. It happened when there were old menuItems.

ChristineWk:
Thanks @chmst 
Notice has disappeared & the font in the Search menu is also back to normal. 
But you can still see the title. Will keep trying. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34658">issues.joomla.org/tracker/joomla-cms/34658</a>.</sub>

chmst:
I did not change the title. This is not in scope of this PR.

N6REJ:
glad to see this becoming part of core.  I used to have to do work arounds to make this happen.

Quy:
No translation for menu type URL: `COM_MENUS_ITEM_FIELD_ICON_LABEL`

modules/mod_menu/src/Helper/MenuHelper.php
```diff
@@ -170,11 +170,11 @@ public static function getList(&$params)
 					// We prevent the double encoding because for some reason the $item is shared for menu modules and we get double encoding
 					// when the cause of that is found the argument should be removed
 					$item->title          = htmlspecialchars($item->title, ENT_COMPAT, 'UTF-8', false);
+					$item->menu_icon      = htmlspecialchars($itemParams->get('menu-icon', ''), ENT_COMPAT, 'UTF-8', false);
```
Quy:
```suggestion
					$item->menu_icon      = htmlspecialchars($itemParams->get('menu_icon', ''), ENT_COMPAT, 'UTF-8', false);
```

modules/mod_menu/tmpl/default_component.php
```diff
@@ -41,21 +41,38 @@
...
+		// The link is an icon
+		if ($itemParams->get('menu_text', 1))
+		{
+			// If the link text is to be displayed, the icon is added with aria-hidden
+			$linktype = '<span class="p-2 ' . $item->menu_icon . '" aria-hidden="true"></span>' . $item->title;
+		}
+		else
+		{
+			// If the icon itself is the link, it needs a visually hidden text
+			$linktype = '<span class="p-2 ' . $item->menu_icon . '"></span><span class="visually-hidden">' . $item->title . '</span>';
```
Quy: Add `aria-hidden="true"`?

administrator/components/com_menus/forms/item_alias.xml
```diff
@@ -44,6 +44,14 @@
 				name="menu-anchor_css"
 				type="text"
 				label="COM_MENUS_ITEM_FIELD_ANCHOR_CSS_LABEL"
+				validate="CssIdentifier"
+			/>
+
+			<field
+				name="menu_icon"
+				type="text"
+				label="COM_MENUS_ITEM_FIELD_ICON_TITLE_LABEL"
```
Quy, line 53: Add `COM_MENUS_ITEM_FIELD_ICON_TITLE_DESC`?

modules/mod_menu/tmpl/default_url.php
```diff
@@ -31,21 +31,38 @@
 
 $linktype = $item->title;
 
-if ($item->menu_image)
+if ($item->menu_image || $item->menu_icon )
```
Quy: Change to this syntax?

```
if ($item->menu_icon)
{
	
}
elseif ($item->menu_image)
{
	
}
```

administrator/components/com_menus/forms/item_heading.xml
```diff
@@ -17,6 +17,13 @@
 				validate="CssIdentifier"
 			/>
 
+			<field
+				name="menu_icon"
```
Quy: Rename `menu_icon_css`?

administrator/components/com_menus/forms/item_alias.xml
chmst: This description is not correct for what I changed. Unfortunately the whole code for icons is not consistent, so this message is wrong.

mariantanase:
I have tested this item :white_check_mark: successfully on #34658

administrator/components/com_menus/forms/item_alias.xml
Quy: Why is it not correct? Reading the code, it appears to be doing exactly this.

> COM_MENUS_ITEM_FIELD_ICON_TITLE_DESC="If an icon class is entered, it takes precedence over the link image."

chmst: It is the other way around. If an image is entered, it takes priority. From my experience with sites, an image as mneu item is more common than only an icon. 
Of course we can change this, matching the older description. I have. 

It also would be possible to enable both - image and icon. But this needs a refactoring of the code.

Quy: Yes, you are right! I don't see a use case to enable both at the same time.

modules/mod_menu/tmpl/default_component.php
chmst: thx. done

administrator/components/com_menus/forms/item_alias.xml
chmst: added description and changed preference

administrator/components/com_menus/forms/item_heading.xml
chmst: done

chmst:
Changed as reqired. Thank you @Quy!

Thanks also @mariantanase for testing. I have improved the code with suggestions. It would be great if you could repeat the test (when @Quy has approved the changes)

administrator/components/com_menus/forms/itemadmin_alias.xml
```diff
@@ -27,6 +27,13 @@
 				label="COM_MENUS_ITEM_FIELD_ANCHOR_TITLE_LABEL"
 			/>
 
+
+			<field
```
Quy:
```suggestion
			/>

			<field
```

modules/mod_menu/tmpl/default_component.php
```diff
@@ -41,17 +41,31 @@
...
 	else
 	{
-		$linktype = HTMLHelper::_('image', $item->menu_image, $item->title);
+		// If the icon itself is the link, it needs a visually hidden text
+		$linktype = '<span class="p-2 ' . $item->menu_icon . '" aria-hidden="true"></span><span class="visually-hidden">' . $item->title . '</span>';
 	}
+}
+elseif ($item->menu_image)
+{
+	// The link is an image, maybe with an own class
```
Quy:
```suggestion
	// The link is an image, maybe with its own class
```

modules/mod_menu/tmpl/default_heading.php
```diff
@@ -15,17 +15,31 @@
...
 	else
 	{
-		$linktype = HTMLHelper::_('image', $item->menu_image, $item->title);
+		// If the icon itself is the link, it needs a visually hidden text
+		$linktype = '<span class="p-2 ' . $item->menu_icon . '" aria-hidden="true"></span><span class="visually-hidden">' . $item->title . '</span>';
 	}
+}
+elseif ($item->menu_image)
+{
+	// The link is an image, maybe with an own class
```
brianteeman:
```suggestion
	// The link is an image, maybe with its own class
```

modules/mod_menu/tmpl/default_separator.php
```diff
@@ -15,17 +15,31 @@
...
 	else
 	{
-		$linktype = HTMLHelper::_('image', $item->menu_image, $item->title);
+		// If the icon itself is the link, it needs a visually hidden text
+		$linktype = '<span class="p-2 ' . $item->menu_icon . '" aria-hidden="true"></span><span class="visually-hidden">' . $item->title . '</span>';
 	}
+}
+elseif ($item->menu_image)
+{
+	// The link is an image, maybe with an own class
```
brianteeman:
```suggestion
	// The link is an image, maybe with its own class
```

modules/mod_menu/tmpl/default_url.php
chmst: done, thx.

mariantanase:
I have re-tested this item ✅ successfully on #34658 

Anyway, I think is better to say:
"If a **Link Icon Class** is entered, it takes precedence over the **Link Image**."
instead of:
"If an icon class is entered, it takes precedence over the link image."

Or, better that that, consider to add that text as a tooltip for "Link Icon Class", to reduce the vertical space between two fields.

chmst:
Thanks for testing. At the moment no changes in language strings are possible https://github.com/joomla/joomla-cms/discussions/34685 . So you could open an issue for a later version.

brianteeman:
__Could someone remove the "updates requested" label please

hans2103 (approved):

hans2103:
I have tested this item :white_check_mark: successfully on 97fee991ed2e0240b2f7f1c2702c57d16c1a3b1f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34658">issues.joomla.org/tracker/joomla-cms/34658</a>.</sub>

richard67:
@mariantanase I have altered your test result in the issue tracker so it is properly counted. Next time please us the "Test this" button in the issue tracker to submitt your test result. Making a comment with just a green check mark isn't conuted right by our tools. Thanks in advance, and thanks for testing.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34658">issues.joomla.org/tracker/joomla-cms/34658</a>.</sub>

richard67:
@wilsonge @bembelimen Would you say this is a bug fix so it can go into 4.0.x, or is it a new feature so it has to go into 4.1? I've set the RLDQ label because I think it needs that decision.

bembelimen:
4.1 from my point of view.

chmst:
It is a new feature from my point of view

bembelimen:
Thx

PR #34691, opened 2021-07-02
Remove cacerts bundled in Joomla and replace with composer package

wilsonge:
### Summary of Changes
Replaces the outdated cacerts bundled with Joomla HTTP and FOF Download with the latest versions (from 2015 and 2017 respectively)

### Testing Instructions
Ensure that the HTTP Package and Downloads package continue to function and use the updated CA Trust

None


HLeithner:
I would like to move this to 3.9.29 (3.9.28 is ready for building final packages) or better 3.10 you remove the old files which could lead to a failure for 3rd party packages depending on the file location.

Maybe it would be better to maintain the 2 files with a composer post update script till EOL of 3.10.

Fedik:
I think @HLeithner is right,
Will be a bad idea to remove this files while 3.9.x.

But,  maybe, can add the bundle in parallel :)

HLeithner:
I merged #34693 so 3.9.28 has a up to date version.

wilsonge:
Closing as alternate PR was merged - if this is a b/c break then it's useless. We can't use composer hooks to update the files because we don't autoload our code with composer in staging.

PR #34709, opened 2021-07-04
make status order icon responsively correct in mobile view

pinak3748:
Pull Request for Issue # .

### Summary of Changes
On every page which has **status column** in which the status icon was not properly aligned in **mobile view**.
The order icon was going below the status.

### Testing Instructions
You have to have some entry on every page that has a status column.

before

after


brianteeman:
Please redo this PR without using inline styles

pinak3748:
ok

brianteeman:
Actually just tested it and this doesn't work as it assumes that the text is always 6 characters. Depending on the language it may not be.

So this is a valid issue but setting a width=100px is not the correct solution

brianteeman:
You can probably achieve what you want by removing the following

https://github.com/joomla/joomla-cms/blob/27851349c71580e4957efd0233a97ac45db22a2b/administrator/templates/atum/scss/vendor/bootstrap/_table.scss#L17-L20

PR #34715, opened 2021-07-05
[4.0] Finder.js to declare awesomplete dependency

sakiss:
Pull Request for Issue # .

### Summary of Changes
The finder.js depends on "awesomplete" but this dependency relies on the layouts to load it.

### Testing Instructions
Run `npm install` to re-create the media folder.
Test the suggestions in the smart search module and the search form in the results page.

The smart search suggestions to work.

The smart search suggestions to work.

No


wilsonge (approved):

Fedik:
> ... but this dependency relies on the layouts to load it.

This was made for purpose. When `show_autosuggest` is disabled then we do not need to load "Awesomplete".

This PR can be closed.

Fedik:
Or

You can make an additional asset (copy of existing `com_finder.finder`), that have Awesomplete dependency, `com_finder.finder-autocomplete`

Then in layout do switch:
```
if show_autosuggest
  use 'com_finder.finder-autocomplete'
else
  use 'com_finder.finder'
```

sakiss:
> This was made for purpose. When show_autosuggest is disabled then we do not need to load "Awesomplete".
Yes but it is error prone. 

Given that in the majority of the cases the suggestions will be enabled and the "Awesomplete" script is 7.5KB, i find the pros more than the cons in accepting that. Any feedback is welcome from other developers.


> You can make an additional asset (copy of existing com_finder.finder), that have Awesomplete dependency, com_finder.finder-autocomplete

Interesting approach. But pretty much the only thing that the finder.js does, is the auto-complete. I am not sure that has to be loaded at all if the auto-complete is off.

Fedik:
> But pretty much the only thing that the finder.js does, is the auto-complete. I am not sure that has to be loaded at all if the auto-complete is off.

It is doing something, check the script. It not only for Autocomplete. 
Otherwise it would be loaded only when `show_autosuggest` is "On"

PR #34743, opened 2021-07-09
[4.1] [GSoC 21] Core Enhancement 1 - Improve Menu-Module Assignment

YatharthVyas:
Project Repo PR: https://github.com/joomla-projects/gsoc21_core-enhancements/pull/1
Plan Document: https://docs.google.com/document/d/1Pl8JGa2hkYkmJzQOn9_mS8a4imDmqc2a/edit#heading=h.gewhdn9rqyeh

<hr>

### Frontend Flow:
This PR re-invents the unconventional method of placing modules from the Backend Dashboard to an easier to visualize series of steps where the users can visit a particular page in the Frontend (while logged in to an admin account) and from there, they can select the menu (automatically determined based on the page) and position of the Module by the means of buttons that reflect the output of where the module will be placed if the user proceeds to create it.

- **New Plugin**
    * Introduce a new system plugin `plg_system_addmodulebutton` for displaying the Add Module Button
    * This button is appended to the end of the `<main>` tag which is omnipresent on all pages. 
    * This button is only displayed when a user logs into the frontend using an account that has Module Add/Edit Permissions
    * Pressing this button adds `?pm=1` as a GET parameter in the URL. This is similar to the `?tp=1` parameter that is used for displaying [template positions](https://docs.joomla.org/Finding_module_positions_on_any_given_page)

addModbtn

- **Place Module** (`?pm=1`) **View** 
    * This view displays the current page will all possible template positions
    * Each position will have a button corresponding to it. This button can be used to select this position for placing the new module.
    * These buttons (to select position) redirect the user to a backend page having a URL like `administrator/index.php?option=com_modules&task=module.selectPosition&position=footer&menu=101` . This URL has 3 parts:
        1. `task=module.selectPosition` is used to call the selectPosition Controller in com_modules/module
        2. `&position=footer` is used to pass the position as a GET param
        3. `&menu=101` is used to pass the menu id of the current page as a GET param

image

- **New Controller: /com_modules/Module/selectPosition**
   * Saves the Position and Menu ID which is passed as a GET param in URL by the previous step in the User State,
   * Redirects to Module Select View where user can select the Module Type
   * After the Module Type has been selected, these state items (Position and Menu ID) is used to pre-fill the Position Name and Menu ID

Here is a video demonstration of the overall frontend flow :

https://user-images.githubusercontent.com/53610833/125059779-525ff200-e0c9-11eb-89c4-e60ccb5e19d4.mp4
 
<hr>

### Backend Flow:
Editing Modules in the Backend Admin Panel turned out to be a little tricky for the user if the template positions were not easy to visualize or if the User was unaware of the layout of the template positions. The old way for a user to get around this difficulty was to 
1. Visit the Global Configuration -> Templates -> Enable Template Positions
2. Visit Frontend
3. Append `?tp=1` as a param in the Page's URL (Might not be easy for users who do not have a technical background)
4. Remember the position name (Note: This view only shows active positions so they cannot even know where inactive positions are. Inactive means those positions that do not have a module to show in it)
5. Go back to the Admin Dashboard and select this position

To simplify all this into an easier and efficient method, a new modal has been added to select the positions via their preview. The video below demonstrates this new modal:

https://user-images.githubusercontent.com/53610833/125062858-783ac600-e0cc-11eb-9955-9fb4aa60c897.mp4

<hr>

### Testing Instructions
- Run `npm ci` because this PR involves re-building of assets
- Either run the SQL update script or download and install https://drive.google.com/file/d/1R4962C-MFFS236UW0cTBPn48DucG0-Qr/view?usp=sharing for the Add Module Button Plugin
- Test the frontend and backend placement flow summarized above. (The videos might be helpful)

<hr>

### Actual result AFTER applying this Pull Request
Refer to the two videos in the Summary of Changes section.

<hr>

I'm not sure but I think this should be included in the Joomla Docs as a User Guide for the new feature.

<hr>

### Mentors
@chmst @nibra @bembelimen
(and thanks @richard67)


richard67:
@YatharthVyas It needs to add the new plugin to the list of core extensions here: https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Extension/ExtensionHelper.php#L273 .

micker:
Whooo

brianteeman:
Some initial observations - maybe more for the mentors than you.

As I stated before I do not like that this is dependent on the template being used. Surely if we can have a preview of template positions and a module edit button without requiring a change to the template it can be done here as well.

Why do I have to go to the very bottom of the page to "turn this on" - imagine navigating there without using a mouse. My first impression was that this was the only area I could insert a module. Do we even need this step? We don't need an additional step for editing modules.

When I select a "place module here" the site changes to the admin (without any notice), I have to log in and there is no method to return to the site once the module has been inserted. Why not use the existing frontend module editing form and existing authorisation - you just need to add a "select module" to that.

Why is this controlled and managed indpendently to the edit module options?

Final observation is regarding the confusion for users when adding a module who dont understand the difference between a menu and a page. For example if you have a blog category and then select an article and you see the add module button they will think they are adding it to the page with the article and will be surprised to see it has been added to every article in the blog

image

YatharthVyas:
Thank you for this detailed and constructive feedback! I'm happy to learn where I can improve.
There are some points that I wish to address before my mentors comment on it

> As I stated before I do not like that this is dependent on the template being used. Surely if we can have a preview of template positions and a module edit button without requiring a change to the template it can be done here as well.

I was unable to figure out a workaround to this. I will need some guidance from my mentors here. Thanks!

> Why do I have to go to the very bottom of the page to "turn this on" - imagine navigating there without using a mouse. My first impression was that this was the only area I could insert a module. Do we even need this step? We don't need an additional step for editing modules.

Excellent Idea. I will try to re-position the button to be similar to the screenshot

> When I select a "place module here" the site changes to the admin (without any notice), I have to log in and there is no method to return to the site once the module has been inserted. Why not use the existing frontend module editing form and existing authorisation - you just need to add a "select module" to that.

I thought about this, but the frontend edit module view doesn't have all the fields. To be specific, the Menu Assignment Tab wasn't present in the frontend edit view. I thought that this has been deliberately not included and hence I decided to proceed with redirection to the admin panel. Transferring this to the frontend can be done but it will definitely require some extra efforts and careful inspection of what is being copy-pasted from the administrator files which might get a little complicated. I would want to hear my mentor's opinion on this too.

YatharthVyas:
image

Updated the Position and Style of the Add Module Button

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -78,10 +78,11 @@ public function onBeforeRender()
 		if ($showAddModuleBtn)
 		{
 			// Add Script Options to pass the Button label Language Constant
-			$this->app->getDocument()->addScriptOptions('js-addModuleBtn', array(
-				'btnLabel' => Text::_('PLG_SYSTEM_ADD_MODULE_BUTTON_LABEL'),
-				'btnDescription' => Text::_('PLG_SYSTEM_ADD_MODULE_BUTTON_DESCRIPTION')
-			));
+			$config = array(
+				'btnLabel' 			=> Text::_('PLG_SYSTEM_ADD_MODULE_BUTTON_LABEL'),
+				'btnDescription' 	=> Text::_('PLG_SYSTEM_ADD_MODULE_BUTTON_DESCRIPTION')
```
richard67:
```suggestion
				'btnLabel'       => Text::_('PLG_SYSTEM_ADD_MODULE_BUTTON_LABEL'),
				'btnDescription' => Text::_('PLG_SYSTEM_ADD_MODULE_BUTTON_DESCRIPTION')
```
It seems you still haven't adjusted your editor or IDE to show you where you have tabs and where spaces. Here you have a mix of both. Please don't use tabs for such alignments inside of the code, e.g. for aligning the "=" of several assignment statements in consecutive lines. For such purpose use spaces. Tabs are used to indent code line from left, not to adjust code in the middle of a line somewhere.

YatharthVyas: Oops I will configure my IDE as per your suggestion right away
Thanks

richard67:
> image
> 
> Updated the Position and Style of the Add Module Button

The "Add Module" label and "Add a new module here" description being shown below each other for that button now seem to be redundant to me.

Why not just use label "Add New Module Here" or "Create Module Here"?

@brianteeman What do you suggest as native speaker?

chmst:
I am not native speaker, but just wanted to say the same.  
"Create a module" is better.

brianteeman:
the l;anguage string is the least important thing to resolve on this pr and can be looked at later

chmst:
> I thought about this, but the frontend edit module view doesn't have all the fields. To be specific, the Menu Assignment Tab wasn't present in the frontend edit view. I thought that this has been deliberately not included and hence I decided to proceed with redirection to the admin panel. Transferring this to the frontend can be done but it will definitely require some extra efforts and careful inspection of what is being copy-pasted from the administrator files which might get a little complicated. I would want to hear my mentor's opinion on this too.

I think, we should keep this out from this PR

administrator/language/en-GB/plg_system_addmodulebutton.ini
```diff
@@ -0,0 +1,11 @@
+; Joomla! Project
+; (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_ADDMODULEBUTTON="System - Add Module Button"
+PLG_SYSTEM_ADD_MODULE_BUTTON_LABEL="Create a module"
```
brianteeman:
```suggestion
PLG_SYSTEM_ADD_MODULE_BUTTON_LABEL="Create a Module"
```

YatharthVyas: Thanks

YatharthVyas:
After discussing @brianteeman's suggestions (Thank you for the same) with my Mentors, here's where we stand:

## 1. Changes in Template File

We discussed that the changes made in Cassiopeia's template files are fine since the feature doesn't depend upon it, ie. The code will still work for external templates that do not use it, just that the select-able positions would be limited to active positions only. Moreover, the [Joomla Docs](https://docs.joomla.org/Finding_module_positions_on_any_given_page) mention that:
> Note that in some circumstances there may be module positions available that are not outlined in red. This can happen when a template defines those module positions as conditional on there being modules enabled in that position. If there are no modules enabled in that position, the template may adapt and the position will not be visible.

so each template can customize this and we're doing the same.
| Before | After |
| ------------- | ------------- |
| image | image |
| Shows Only Active Positions | Shows All Positions |

<br>

## 2. Redirection to the admin panel issue
Quoting what I wrote earlier:

> I thought about this, but the frontend edit module view doesn't have all the fields. To be specific, the Menu Assignment Tab wasn't present in the frontend edit view. I thought that this has been deliberately not included and hence I decided to proceed with redirection to the admin panel. Transferring this to the frontend can be done but it will definitely require some extra efforts and careful inspection of what is being copy-pasted from the administrator files which might get a little complicated. 

While the suggestion is very much apt and valid, Mentors agree that the missing extra tab is out of this PR's scope and it might be better to leave this as is in its current state.

<br>

## 3. Confusion regarding Menu and Page

This sure needs to be worked upon. I believe some rephrasing will be required in the button's tooltip description to clear this confusion. I believe that the feedback from the community would be the best source of inspiration for this.

<br>

### Fixed
- Position of the Module Button and it's label text was not very user friendly and that has been updated.


<hr>

I would like to discuss the future scope of this PR by requesting the maintainers to give their feedback on any other possible improvements that I might've missed out. Thank you

Dako390:
I have tested this item :white_check_mark: successfully on 0f735dbb8d54941022c4ceb2b0b2c8b6cfb382cd<br /><br />Works as intended.

One possible addition to it: I would love to be able to have the option "Place Module here" while adding a module to the dashboard.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34743">issues.joomla.org/tracker/joomla-cms/34743</a>.</sub>

YatharthVyas:
Thank you for testing!


> One possible addition to it: I would love to be able to have the option "Place Module here" while adding a module to the dashboard.

That sounds like a good idea. I will probably open up a new PR in the future to extend this feature to the Dashboard.

administrator/components/com_modules/tmpl/module/preview_positions.php
```diff
@@ -0,0 +1,42 @@
...
+$this->fieldsets = $this->form->getFieldsets();
+echo $this->form->renderField($templateField);
+?>
+<div class="jviewport-height90">
+	<iframe 
+		src=<?php echo $iframeBaseURL . '?pm=1&edit=1&templateStyle='; ?>
+		id="module-position-select" 
+		name="module-position-select" 
+		title="module-position-select"
+		height="100%"width="100%">
```
Quy:
```suggestion
		src="<?php echo $iframeBaseURL . '?pm=1&edit=1&templateStyle='; ?>
		id="module-position-select" 
		name="module-position-select" 
		title="module-position-select"
		height="100%"
		width="100%">
```

YatharthVyas: Thank you

scout507:
Everything seems to work as intended. It's a nice feature and easy to use.

Some suggestions:

- The 'Add Module' button would be a perfect fit for a floating button. Something like this accessability button:

Screenshot 2021-08-17 111903



- I haven't found a way to cancel the 'place module' mode. There should probably be a cancel button to improve usability.

- The 'place module here' button should probably be tied to the positions, rather than individual modules, since it doesn't seem to actually place it in the order you'd expect (above or below the selected module).

- Also, when you are not loged in as Admin in the Front-End but trying to use the 'Select From Template View' in the Back-End, you get an error stating that you need to be loged in first, so you have to login in the Template View in order to use this feature:


Screenshot 2021-08-17 101713


That beeing said, all these things are just suggestions to improve the useability of this feature, which is otherwise working just fine.

scout507:
I have not tested this item.<br /><br />Everything seems to work as intended. It's a nice feature and easy to use.

Some suggestions:

- The 'Add Module' button would be a perfect fit for a floating button. Something like this accessability button:


- I haven't found a way to cancel the 'place module' mode. There should probably be a cancel button to improve usability.

- The 'place module here' button should probably be tied to the positions, rather than individual modules, since it doesn't seem to actually place it in the order you'd expect (above or below the selected module).

- Also, when you are not loged in as Admin in the Front-End but trying to use the 'Select From Template View' in the Back-End, you get an error stating that you need to be loged in first, so you have to login in the Template View in order to use this feature:



That beeing said, all these things are just suggestions to improve the useability of this feature, which is otherwise working just fine.  


<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34743">issues.joomla.org/tracker/joomla-cms/34743</a>.</sub>

YatharthVyas:
Thank you soo much for such a comprehensive feedback! 😃 
<br>
 
> * The 'Add Module' button would be a perfect fit for a floating button. Something like this accessability button:

Nice Idea. I will discuss this with my mentors. I think we had decided to place the button in its current position because it resonated with the position fo the frontend editing button
<br>

> * I haven't found a way to cancel the 'place module' mode. There should probably be a cancel button to improve usability.

Currently, pressing the back button on your browser would do the job of cancelling it but your feedback has helped me understand that the back button way isn't the best and it needs some sort of a cancel button. I appreciate this idea and I will think about the best way to implement this!
<br>

> * The 'place module here' button should probably be tied to the positions, rather than individual modules, since it doesn't seem to actually place it in the order you'd expect (above or below the selected module).

I 100% agree with your point here, but there's just one tiny reason why I didn't decide to go forward with this. And that's because we would need to edit the template file in order to add this functionality and modifying the template file would be an issue for 3PD (third-party developers) as they would have to adapt to it accordingly. In the current iteration, everything would work without the need for 3PD having to modify their templates.
<br>

> * Also, when you are not loged in as Admin in the Front-End but trying to use the 'Select From Template View' in the Back-End, you get an error stating that you need to be loged in first, so you have to login in the Template View in order to use this feature:

I had deliberately kept this check as I intended it to be more like an informatory warning which would prevent any person without permissions from being able to modify module positions while guiding the users with the access in modifying it.

<br>

I want to thank you again for taking time to review this PR and provide such a helpful and detailed feedback. I'm really grateful!

build/media_source/layouts/js/chromes/outline.es6.js
```diff
@@ -0,0 +1,40 @@
+/**
+  * @copyright  (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+  * @license    GNU General Public License version 2 or later; see LICENSE.txt
+  */
+
+(() => {
+  'use strict';
+
+  /**
```
PhilETaylor, line 9:
```suggestion
   /**
```

build/media_source/com_modules/js/admin-modules-preview_positions.es6.js
```diff
@@ -0,0 +1,21 @@
...
+  */
+
+/**
+ * Add a keyboard event listener to the Select Template Style element.
+ *
+ * This script is meant to be loaded deferred. This means that it's non-blocking
+ * (the browser can load it whenever) and it doesn't need an on DOMContentLoaded event handler
+ * because the browser is guaranteed to execute it only after the DOM content has loaded, the
+ * whole point of it being deferred.
+ *
```
PhilETaylor:
```suggestion
```

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * Place Module Button
+ *
+ * @copyright (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license GNU General Public License version 2 or later; see LICENSE.txt
+ * @since  __DEPLOY_VERSION__
```
PhilETaylor, line 8:
```suggestion
 *
 * @since  __DEPLOY_VERSION__ 
```

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+defined('_JEXEC') or die;
+use Joomla\CMS\Helper\ContentHelper;
+use Joomla\CMS\Language\Text;
+use Joomla\CMS\Plugin\CMSPlugin;
+use Joomla\CMS\WebAsset\WebAssetManager;
+/**
```
PhilETaylor, line 17:
```suggestion

/**
```

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class PlgSystemAddModuleButton extends CMSPlugin
+{
+	/**
+	 * Load plugin language files automatically
+	 *
+	 * @var    boolean
+	 * @since  3.9.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -0,0 +1,92 @@
...
+	 */
+	protected $autoloadLanguage = true;
+
+	/**
+	 * Application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  3.2
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -0,0 +1,92 @@
...
+	 */
+	protected $app;
+
+	/**
+	 * Listener for the `onBeforeRender` event
+	 *
+	 * @return  void
+	 *
+	 * @since   1.0
```
PhilETaylor:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

plugins/system/addmodulebutton/addmodulebutton.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<extension type="plugin" group="system" method="upgrade">
+	<name>plg_system_addmodulebutton</name>
+	<author>Joomla! Project</author>
+	<creationDate>June 2021</creationDate>
```
PhilETaylor, line 5: Might need revisiting :)

YatharthVyas: Good spot!
I'll remember to revisit this when the PR gets 2 successful tests. 
Thanks

plugins/system/addmodulebutton/addmodulebutton.php
YatharthVyas: Thank you for all the suggestions!

brianteeman:
Trying to edit a module in the front endyou are presented with the module edit form. If you then click on the "select from template" button a modal is opened - it should NOT have the admin interface elements highlighted in the screenshot below

image

brianteeman:
_deleted this as its not related to this pr_

brianteeman:
Please change this button text to "Select Position From Template Preview" or even "Select Position From Template"

image

brianteeman:
Disable Frontend Editing in global configuration

image


Expected behaviour is for the add module functionality not to be active. It is :(

image

brianteeman:
I was really confused about this button. I thought when it said "add a new module here"  the _here_ meant something connected to the area on the screen that the button is located. I finally understand that there is just one button on the page.

> > This button is appended to the end of the <main> tag which is omnipresent on all pages.

1. The text should be "Add a new module" - the "here" is wrong and confusing
2. You cannot guarantee that **main** is present in every template. 
3. The button should be placed in a corner just as debug, accessibility buttons are. it refers to the entire page not the part of the page inside main. Look at the cassiopeia screenshot below of the sample data homepage. The button location makes no sense at all.
4. The button should be consistently placed on the screen no matter what page you are on - clearly it isnt

Inked2021-11-20_20-38-39

brianteeman:
Final comment on the frontend creation of modules.

Click on the button
-> select the position 
-> log in to the admin
-> select the module
-> complete the module details and save

You are now stuck in the admin interface. 
To return to the site you have to press back 4 times

Expected behaviour would be to return to the site and now you can see what you have done

YatharthVyas:
Thank you soo much @brianteeman for such a thorough feedback! I agree with all the points that you stated. I will start working on these improvements.

administrator/components/com_modules/layouts/joomla/form/field/modulespositionedit.php
```diff
@@ -76,13 +76,38 @@
...
+
+$modalUrl = 'index.php?option=com_modules&view=module&layout=preview_positions&id=1&client_id=' . $clientId;
+
+// &tmpl=component doesn't redirect if the user isn't logged into backend hence we are adding it conditionally
+$modalUrl = $app->isClient('site') ? 'administrator/' . $modalUrl : $modalUrl . '&tmpl=component';
 
+?>
+<joomla-field-fancy-select <?php echo implode(' ', $attributes); ?>>
+<?php
+echo HTMLHelper::_('select.groupedlist', $positions, $name, array(
```
PhilETaylor, line 90: convert to short array syntax for Joomla 4+

administrator/components/com_modules/layouts/joomla/form/field/modulespositionedit.php
```diff
@@ -76,13 +76,38 @@
...
+		'list.attr'   => implode(' ', $selectAttr),
+	)
+);
+?>
+</joomla-field-fancy-select>
+<?php
+echo HTMLHelper::_(
+	'bootstrap.renderModal',
+	'Modal_position',
+	array(
```
PhilETaylor, line 102: convert to short array syntax for Joomla 4+

administrator/components/com_modules/tmpl/module/preview_positions.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_modules
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
```
PhilETaylor, line 6: Probably going to be 2022 now haha

plugins/system/addmodulebutton/addmodulebutton.php
```diff
@@ -0,0 +1,94 @@
...
+
+				return;
+			}
+		}
+
+		// Display the Add Module Button
+		if ($showAddModuleBtn)
+		{
+			// Add Script Options to pass the Button label Language Constant
+			$config = array(
```
PhilETaylor, line 83: convert to short array syntax for Joomla 4+

brianteeman:
Could someone add the "conflicting files" label please

HLeithner:

chmst:
Unfortunately there are too many open questtions with 3rd party templates and template frameworks, no chance to make it compatible with all possible constellations.

PR #34608, opened 2021-06-23
Fix metismenu caret wrapping

RickR2H:
### Summary of Changes
When an menu item has a longer title then the other items in the Metismenu dropdown, the caret will wrap to the next line. This PR fixes this issue.

### Testing Instructions
install the Blog Sample Data.
Open the menu "Main Menu Blog".
Open the menu item "Articles" and rename it to "Articles Longer Title To Test".
The caret now wraps underneath the menu item title.
Please test all drop down possibilities in the menu. This is easily done by batch copying a menu item with multiple sub-items.

In the last commits I added fixes for long titles. So please also add some items with longer titles (6 or 7 words)

### Actual result BEFORE applying this Pull Request
before

### Expected result AFTER applying this Pull Request
after

### Documentation Changes Required
No

PR #34750, opened 2021-07-10
[4.0] Revert PR #34608 "Fix metismenu caret wrapping" because it has broken styling of the active menu item

richard67:
Pull Request for Issue # .

### Summary of Changes

This reverts PR #34608 "Fix metismenu caret wrapping" because it has broken styling of the active menu item.

The PR has added a span between the `li` element and the link but did not adjust existing SCSS for this, so the parent selectors `&.` don't match anymore for the menu item types modified by that PR. Instead of correcting SCSS that PR has added more SCSS, which leads to redundant hardly maintainable code, and the additional SCSS is not complete.

As a result of this, at least the styling of the active menu item has been broken by that PR.

It can be that the PR has broken more based on parent selectors which don't match anymore, but I haven't tested everything. RTL could be one thing, especially for menu items of level 3 or more.

Therefore it is the best to revert that PR for now.

### Testing Instructions

Make a new installation and then install Sample Data Blog.

After that, create a few copies of the Main Menu Blog and put them on different positions. Change the layout from "collapsible-dropdown" to "dropdown" for the copies, and for some of them use menu class "menu-horizontal".

Then check if a menu item with a link (i.e. type "URL" or "Component") is underlined when it is active or one of its submenu items is active, like is it done with the menu items without link (type "Heading" or "Separator").

While the active menu item and its parents are underlined underlined when the menu item is of a type without a link (type "Heading" or "Separator"), this doesn't happen when the menu item is of a type with a link (i.e. type "URL" or "Component") which have been modified by PR #34608 .

See the red marks in the following screenshot: 

As said, for menu items without a link which have not been modified by PR #34608 it works:

The above screenshots shows examples top menu items because that's the most visible thing, but of course parent menu items at lower levels should work the same.

All types of menu items and their parents are underlined (with a pseudo element in the header and with text decoration elsewhere) in the same way when they are the active menu item.

The following screenshot shows the example for a top menu item because that's the most visible thing, but of course parent menu items at lower levels should work the same.

None.

### DIFF
--- a/templates/cassiopeia/html/mod_menu/dropdown-metismenu_component.php
+++ b/templates/cassiopeia/html/mod_menu/dropdown-metismenu_component.php
@@ -68,11 +68,9 @@ elseif ($item->browserNav == 2)
 	$attributes['onclick'] = "window.open(this.href, 'targetWindow', '" . $options . "'); return false;";
 }
 
-echo '<span class="mm-toggler-component__wrapper">';
 echo HTMLHelper::link(OutputFilter::ampReplace(htmlspecialchars($item->flink, ENT_COMPAT, 'UTF-8', false)), $linktype, $attributes);
 
 if ($showAll && $item->deeper)
 {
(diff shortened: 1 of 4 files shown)


wilsonge:
Thanks!

Issue #22231, opened 2018-09-18
Invalid SQL created by removeOrphanNodes function


He-Man321:
### Steps to reproduce the issue
1) Log in to /Administrator as Admin
2) Go to the Articles and create a new one. Give it a title and Click Save & Close.
3) Open the same article again and click Save & Close again.

### Expected result
No error.

### Actual result
Error
Save failed with the following error: SQLSTATE[HY000]: General error: 1093 You can't specify target table 'abc_finder_taxonomy' for update in FROM clause

### System information (as much as possible)
Tested in joomla 3.8.8 and 3.8.12 running on Windows 2012R2 with MySQL 5.7.9 and PHP 7.2.2.

### Additional comments
This only happens when the Smart Search plug-in is enabled.

I have established that the cause of this is the removeOrphanNodes() function in administrator\components\com_finder\helpers\indexer\taxonomy.php.  Back in 3.4.5 this function deleted from the table without using a subquery but at some point after that (and before 3.7.5 I am told) it was changed to use the subquery.  However, https://dev.mysql.com/doc/refman/5.6/en/update.html confirms that in MySQL "You cannot update a table and select from the same table in a subquery".  This causes the query to fail on a MySQL system.
I have discussed this in more detail on the forum here: https://forum.joomla.org/viewtopic.php?t=965486

csthomas:
@alikon The problem start at Joomla 3.7.0 (https://github.com/joomla/joomla-cms/pull/12313)

alikon:
@csthomas maybe it depends from the mysql version on 5.7 works without issue

from the op text it's ambiguos the issue arise on 5.6 or in 5.7  more detail are needed....

csthomas:
I assume this is something like https://stackoverflow.com/questions/4429319/you-cant-specify-target-table-for-update-in-from-clause/14302701

He-Man321:
Yes, this isn't a MySQL version issue; there are Stack Overflow pages with people moaning about MySQL not allowing updates to tables in a where subquery going back to 2006 and beyond. It just isn't something MySQL supports, so I suspect this will happen with all versions.

alikon:
on Mysql 5.7.23 and on Maria DB 10.3.2 it works 
so it is a version issue
but this apart , this is a real issue  cause for https://docs.joomla.org/Technical_requirements
we still support 5.1. +  on 3.x :cry:

He-Man321:
[...]

PR #34763, opened 2021-07-12
[4.0] Fix SQL error 1093 "You can't specify target table ..." when updating from 3.10 with certain versions of MySQL 5.7

richard67:
Pull Request for Issue #32888 .

### Summary of Changes

This pull request (PR) adds proper names quoting and correct use of table aliases to the SQL statement for inserting the new template styles when updating from 3.10 to 4.

This fixes issue #32888 , which happens only on particular versions of MySQL 5.7 and results in the update from 3.10 to 4 being broken because the SQL updates fail with SQL error 1093 when executing the statement modified by this PR.

For sure MySQL version 5.7.14 which has been reported in issue #32888 has that problem, and very likely also version 5.7.9 which has been reported in issue #22231 , which has the same cause.

With MySQL 5.6 I could not reproduce the problem, and later versions of MySQL than the one mentioned above seem not to have the problem either.

The same changes as for MySQL are done for the PostgreSQL update script for consistency and for being safe from PostgreSQL becoming more strict in future regarding correct quoting of names which are equal to reserved words (e.g. "count").

For reproducing the issue it needs a MySQL database server version 5.7.14.

Because I haven't found any XAMPP version which includes this MySQL version, it needs to set up a MySQL server of that version.

You can find the downloads here: https://downloads.mysql.com/archives/installer/ .

Make sure that you have selected version 5.7.14 before downloading.

If that link doesn't work for you (not sure if it requires an Oracle Tech Network account), you can also download it from here: https://test5.richard-fath.de/mysql-installer-community-5.7.14.0.msi (MD5 checksum is b4df226b58390f8b3d02d264a785e715).

Make an installation of MySQL server including the client tools, if necessary on a separate (virtual) computer if you don't want to mess up your existing installation e.g. of XAMPP, and then use that database server from remote, i.e. not use "localhost" but the FQDN of that database server for your Joomla installation.

1. Make a new installation of current 3.10-dev or latest nightly or the last alpha version, using a PHP version which is supported by Joomla 4 and the database server prepared as described above.

2. Try to update that 3.10 installation to the latest 4.0 nightly build.

Result: See issue #32888. The update fails with SQL error "1093 You can't specify target table '#__template_styles' for update in FROM clause", and frontend and backend are broken.

When checking for the last SQL statement logged in the update log "administrator/logs/joomla_update.php", it is the statement coming just before the one modified by this PR.

#### Test 2: Verify that this PR fixes the issue

1. Repeat the previous test 1 starting at the same starting point and using again an empty database in the same MySQL 5.7.14 server, but this time use the update package or custom update URL created by Drone for this PR.

Result: The update succeeds. There is only the usual alert about the missing template found, which disappears after navigating around in backend.

2. Go to "System -> Maintenance -> Database" and check if there are database errors shown.

Result: There is one problem shown for the CMS.

3. Verify that it is only the problem with not matching CMS version which is normal when using development packages by hovering over the "One Problem" button.

Result: Only the usual version problem.

4. Verify if the new J4 templates are default templates.

Result: Atum is the default backend template, and Cassiopeia is the default frontend template.

#### Test 3: Verify that the SQL statement modified by this PR still does what it shall do

1. Start again with a 3.10 and an empty database and attempt to update to the patched J4 package for this PR , but before updating create a copy of the Protostar template and make this copy the default frontend template.

2. Verify if the new J4 templates are default templates.

Result: Atum is the default backend template, but Cassiopeia is not the default frontend template. This is still the copy of Protostar, and so frontend will be broken.

Alternatively you could install a 3rd party template on 3.10 before updating, which is compatible to 3.10 and 4, if you can find such a template, and make that the default.

Testing the same for the backend template will be harder. You need to find one which is compatible to 3.10 and 4 and make that the default backend template before updating.

The result in general should be:

If the Protostar or the beez3 template is the default frontend template on 3.10, Cassiopeia will be the default frontend template after the update, otherwise not, and if the Isis or the Hathor template is the default backend template on 3.10, Atum will be the default backend template after the update, otherwise not.

#### Test 4: Verify that nothing is broken for other MySQL versions or PostgreSQL

Execute test 3 on any other version of MySQL than those which have the issue, or on any PostgreSQL version.

The result should be the same.

When using certain versions of MySQL 5.7, e.g. for sure 5.7.14, as database server, an update from any 3.10 version to current J4 is broken because the SQL updates fail with SQL error 1093 when executing the statement modified by this PR at the very beginning of the database updates.

When using the button to go to the control panel, the backend is broken and looks like this:

Updates from 3.10 to 4 succeed with these MySQL versions, and for other versions of MySQL or for MariaDB or PostgreSQL nothing changes.

None.

### DIFF
--- a/administrator/components/com_admin/sql/updates/mysql/4.0.0-2018-03-05.sql
+++ b/administrator/components/com_admin/sql/updates/mysql/4.0.0-2018-03-05.sql
@@ -20,8 +20,8 @@ INSERT INTO `#__extensions` (`name`, `type`, `element`, `folder`, `client_id`, `
 ('cassiopeia', 'template', 'cassiopeia', '', 0, 1, 1, 0, '{}', '{}', '', '', 0, '0000-00-00 00:00:00', 0, 0);
 
 INSERT INTO `#__template_styles` (`template`, `client_id`, `home`, `title`, `params`) VALUES
-('atum', 1, (CASE WHEN (SELECT count FROM (SELECT count(`id`) AS count FROM `#__template_styles` WHERE home = '1' AND client_id = 1 AND `template` IN ('isis', 'hathor')) as c) = 0 THEN '0' ELSE '1' END), 'atum - Default', '{}'),
-('cassiopeia', 0, (CASE WHEN (SELECT count FROM (SELECT count(`id`) AS count FROM `#__template_styles` WHERE home = '1' AND client_id = 0 AND `template` IN ('protostar', 'beez3')) as c) = 0 THEN '0' ELSE '1' END), 'cassiopeia - Default', '{}');
+('atum', 1, (CASE WHEN (SELECT b.`count` FROM (SELECT count(a.`id`) AS `count` FROM `#__template_styles` a WHERE a.`home` = '1' AND a.`client_id` = 1 AND a.`template` IN ('isis', 'hathor')) AS b) = 0 THEN '0' ELSE '1' END), 'atum - Default', '{}'),
+('cassiopeia', 0, (CASE WHEN (SELECT d.`count` FROM (SELECT count(c.`id`) AS `count` FROM `#__template_styles` c WHERE c.`home` = '1' AND c.`client_id` = 0 AND c.`template` IN ('protostar', 'beez3')) AS d) = 0 THEN '0' ELSE '1' END), 'cassiopeia - Default', '{}');
 
 --
 -- Move mod_version to the right position for the atum template
--- a/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2018-03-05.sql
+++ b/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2018-03-05.sql
@@ -20,8 +20,8 @@ INSERT INTO "#__extensions" ("name", "type", "element", "folder", "client_id", "
 ('cassiopeia', 'template', 'cassiopeia', '', 0, 1, 1, 0, '{}', '{}', '', '', 0, '1970-01-01 00:00:00', 0, 0);
 
 INSERT INTO "#__template_styles" ("template", "client_id", "home", "title", "params") VALUES
-('atum', 1, (CASE WHEN (SELECT count FROM (SELECT count("id") AS count FROM "#__template_styles" WHERE home = '1' AND client_id = 1 AND "template" IN ('isis', 'hathor')) as c) = 0 THEN '0' ELSE '1' END), 'atum - Default', '{}'),
-('cassiopeia', 0, (CASE WHEN (SELECT count FROM (SELECT count("id") AS count FROM "#__template_styles" WHERE home = '1' AND client_id = 0 AND "template" IN ('protostar', 'beez3')) as c) = 0 THEN '0' ELSE '1' END), 'cassiopeia - Default', '{}');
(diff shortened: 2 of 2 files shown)


richard67:
I think it's a release blocker if updates break on particular MySQL versions which are supported by J4.

chmst:
Tested the SQL on a mariaDB in a J3 application and works fine. 

```
INSERT INTO `#__template_styles` (`template`, `client_id`, `home`, `title`, `params`) VALUES
('atum', 1,(
CASE
	WHEN (
		SELECT b.`count` 
			FROM (
				SELECT count(a.`id`) AS `count` 
				FROM `#__template_styles` a 
				WHERE a.`home` = '1' 
				AND a.`client_id` = 1 
				AND a.`template` IN ('isis', 'hathor')
			) 
		AS b) = 0 
	 THEN '0' 
	 ELSE '1' 
 END
 ), 
 'atum - Default', '{}');
('cassiopeia', 0, (
CASE 
	WHEN (
		SELECT d.`count` 
			FROM (
				SELECT count(c.`id`) AS `count` 
				FROM `#__template_styles` c 
				WHERE c.`home` = '1' 
				AND c.`client_id` = 0 
				AND c.`template` IN ('protostar', 'beez3')
			) 
		AS d) = 0 
	THEN '0' 
	ELSE '1' 
END
), 'cassiopeia - Default', '{}');




```

richard67:
> Tested the SQL on a mariaDB in a J3 application and works fine.

Thanks. Not a full test for the issue, but at least we know nothing is broken for MariaDB.

obuisard:
So far:
test 1 - same issue as before, confirms the problem
test 2 - the PR fixes the issue on MySQL 5.7.14. The update went as smoothly as can be.
Note: I do not any database error (I used the latest J!4 dev nightly and swapped the file with the fix for the MySQL update file).
test 3 - verification about the default template (frontend): success
I will report later on test 4.

richard67:
That sounds good so far.

richard67:
Tomorrow night RC 4 will be released, so there is not much time if we want to get this fix into RC 4.

richard67:
@obuisard Have you meanwhile found time to complete your test? If so, please mark your test result in the issue tracker here https://issues.joomla.org/tracker/joomla-cms/34763 by using the "Test this" button, selecting the appropriate test result and submitting. There would be a change to get it into RC 4 today but we need at least your test completed. Thanks in advance.

joomdonation:
I don't have MySQL 5.7.14 to test. However, I tested update with MySQL 5.7.29 and everything is working as expected.

obuisard:
I have tested this item :white_check_mark: successfully on 7b9fa98a19fd9d5a133736c9f2d2c79007dbc954<br /><br />I have tested with MySQL 5.7.14 ONLY. Results came as expected throughout Test 1 to 3.
Right now I do not have the ability to test other MySQL or PostgreSQL versions.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34763">issues.joomla.org/tracker/joomla-cms/34763</a>.</sub>

richard67:
I think as the others have tested the other databases, we can say that's one successful test for all in total.

richard67:
@obuisard Thanks for testing.

obuisard:
> @obuisard Thanks for testing.

You are welcome @richard67. This was a difficult one, thank you for looking into it!

wilsonge:
Thanks!

richard67:
Because there is another issue which results in the same SQL error 1093 but could only be reproduced with MySQL 5.7.9 and not with the 5.7.14 used here, I have checked if this fix here also works for MySQL 5.7.9.

Thanks god it does, at least with the database server being 5.7.9 (the webserver and so the db client were newer).

PR #33234, opened 2021-04-22
[4.0] Redo of the custom elements Alert and Tab

dgrammatiko:
Pull Request for Issue # .

# WIP don't test yet

# This is a PROXY only to get human tests. The real PRs are on the other repo.

### Summary of Changes
- A refactor of these two elements as the code is nowhere near production grade
- PR on the real repo: https://github.com/joomla-projects/custom-elements/pull/162

### Testing Instructions

Check that alerts working fine. Check that tabs work fine.

PR #33170, opened 2021-04-16
[4.0] use sass-embedded, reducing build time by 20-30%

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

- Switch to the `sass-embedded` package instead of the dart-sass (it's still dart sass under the hood, but that's the technical part)
- Info: https://github.com/sass/embedded-host-node
- info: https://twitter.com/SassCSS/status/1383151350851072003

### Testing Instructions

Run `npm ci` and compare the build times

### Documentation Changes Required

No, packages are supposed to be easily interchangeable

PR #34781, opened 2021-07-14
[4.0] Alerts custom element redo

dgrammatiko:
Redo of Pull Request #33234 .

## This one should be treated as Release Blocker 
(The alert code wasn't production-grade...)

### Summary of Changes
The Alert Custom Element had some known issues and this PR is pulling a version that fixes all of them https://github.com/joomla-projects/custom-elements/pull/179 

### Testing Instructions
Go to front end and backend and check that alerts are functioning as before (there are no styling changes, although the source is using CSS Vars for almost all parameters)

You can fire an alert using the browser's console with these snippets: 

```js

// Danger
Joomla.renderMessages({danger: ['Something happen']})
// Warning
Joomla.renderMessages({warning: ['Something happen']})
// Success
Joomla.renderMessages({success: ['Something happen']})
// Info
Joomla.renderMessages({info: ['Something happen']})

```

Screenshot 2021-07-14 at 12 38 57
Screenshot 2021-07-14 at 12 39 12

The way that messages are rendered is exactly the same (only one internal change, not exposed to the API), so there is not anything to document.

@wilsonge DO NOT MERGE THIS WITHOUT MERGING THE OTHER PR CREATE A RELEASE AND CHANGE THE PACKAGE.JSON TO USE A VERSION!!!!


richard67:
@dgrammatiko Npm still fails in drone.

Update: Would not be a problem as such because composer.lock will be updated anyway later when a new release has been made of custom_elements, but the failing npm causes other tests like the system tests not to run, so if it is possible to fix it, it would help with that.

dgrammatiko:
> Npm still fails in drone.

I saw that but I have no other tricks here (we need to install an npm package using git but for some reason, the drone script doesn't support that). If you know another way please edit the packge.json directly

PS it should be the same reason why the other PR for the SCSS compiler also failed (spawn... )

PS it turns out that drone is using the node alpine container which doesn't have git (https://github.com/nodejs/docker-node/blob/fd130acf063b312355a5d88d51716db3ff34ae49/14/alpine3.11/Dockerfile)... I can propose a docker container based on node 14 alpine with git, this will probably solve the problem here but also in https://github.com/joomla/joomla-cms/pull/33170

wilsonge:
Do we not need the `close-text` attribute on the hardcoded alert elements around core?

dgrammatiko:
Are there any hard coded custom elements? If so yes, it's for a11y

@wilsonge there is only one instance with `dismiss` attribute and now it's also patched

wilsonge:
I have tested this item :white_check_mark: successfully on 2c92281a9a00c15be7f928ebe25264286899e3d9<br /><br />Navigated around the logins, stats plugin alert and javascript generated alerts (e.g. when trying to save an article with no title). All seemed fine afterwards<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34781">issues.joomla.org/tracker/joomla-cms/34781</a>.</sub>

wilsonge:
Thanks!

dgrammatiko:
Thanks!

PR #34797, opened 2021-07-15
fixed getMetaData function undefined when user use Ajax call back function

bigbearman:
…ction

When the user uses the ajax call back function getList on this module. this will return 500 because $app->getDocument() return null so getMetaData function undefined.

Pull Request for Issue # .

### DIFF
--- a/modules/mod_banners/src/Helper/BannersHelper.php
+++ b/modules/mod_banners/src/Helper/BannersHelper.php
@@ -14,6 +14,7 @@ namespace Joomla\Module\Banners\Site\Helper;
 use Joomla\CMS\Application\CMSApplication;
 use Joomla\CMS\Component\ComponentHelper;
 use Joomla\CMS\Environment\Browser;
+use Joomla\CMS\Factory;
 use Joomla\Component\Banners\Site\Model\BannersModel;
 use Joomla\Registry\Registry;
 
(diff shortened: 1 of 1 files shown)


richard67:
@kienduongbk Please provide testing instructions. The head lines for that in the PR description are not there without a reason.

richard67:
@kienduongbk `$app->getDocument()` should work in J4, so please provide testing instructions or a detailed description on what you are doing to get that issue so we can see how it happens that`$app->getDocument()` return null in your scenario. Thanks in advance.

bigbearman:
@richard67 Thank you for answer my issue.

I want to get a buffer of module banners for preview on the backend. So I build a function to call back on my plugin to get that.
module: mod_banners
URL: domain.com/index.php?action=getmodule&moduleid=xxx
I used ModuleHelper::renderModule($module, array('style'=>$style)) for return content. 

https://prnt.sc/1czftq3

Fedik:
> So I build a function to call back on my plugin to get that.

Here is your issue, you hacked in normal runtime request and trying to call your code before the document is ready.

Please have a look how to use ajax with your plugin or module https://docs.joomla.org/Using_Joomla_Ajax_Interface

What you trying to fix here is not Joomla issue.

bigbearman:
@Fedik Thanks.
 I fixed it.

Issue #34810, opened 2021-07-17
[4.0] Image Preview For Media Field


joomdonation:
### Steps to reproduce the issue
1. Create a new banner
2. Select an image for it
3. Save it
4. Edit the banner again and look at the Image

### Expected result
Image previous is being displayed properly

### Actual result
At first, it is displaying the text **No Image Select** (as if the path to the image is wrong) and then about less than one second later, it is being displayed properly

### System information (as much as possible)
Joomla 4 nightly build

### Additional comments
Maybe when the page is loaded, the wrong preview image is used and the right path is updated later?

dgrammatiko:
> Image previous is being displayed properly

Loads fine here (tested with local adapter only)

brianteeman:
Not seen any problem here either

joomdonation:
![banner_image](https://user-images.githubusercontent.com/977664/126060215-b14d850a-10c8-4bd9-bf62-60727eb44087.gif)

Hope it is a problem on my computer only. Here is what happens when I edit a banner which has image selected.

chmst:
I see the "no image selectedt" message for a moment, then the banner. The image is ok. FF on win10.



![banner](https://user-images.githubusercontent.com/1035262/126061776-f83ed229-def4-4ebe-bdaf-d734dc499cff.gif)

dgrammatiko:
Ok, observing the page source and the DOM after the JS was executed I have
<img width="1201" alt="Screenshot 2021-07-18 at 12 21 32" src="https://user-images.githubusercontent.com/3889375/126063721-ea42cc18-15f5-4e35-9a08-bf566e6e1463.png">

<img width="1100" alt="Screenshot 2021-07-18 at 12 21 48" src="https://user-images.githubusercontent.com/3889375/126063726-c87e2368-e62f-45ce-a27d-cec92da4c094.png">

Basically it's down to these lines https://github.com/joomla/joomla-cms/blob/3071cd68a5d17e9616144e5ee003f143122c2aa6/layouts/joomla/form/field/media.php#L77-L81

We either have to pass the image URL from the HTML helper or remove the hash part manually before checking with the `file_exists()`

joomdonation:
I made PR https://github.com/joomla/joomla-cms/pull/34833 to fix the issue. Please help testing. Thanks !

PR #34833, opened 2021-07-19
[4.0] Fix image preview for media form field

joomdonation:
Pull Request for Issue #34810.

### Summary of Changes
The value stored in media form field (at least for image) contains adapter information (for example, images/headers/blue-flower.jpg#joomlaImage://local-images/headers/blue-flower.jpg?width=700&height=180), thus file_exist checks on this line https://github.com/joomla/joomla-cms/blob/4.0-dev/layouts/joomla/form/field/media.php#L77 return false and it results in preview image is not being displayed properly as described in the issue https://github.com/joomla/joomla-cms/issues/34810

This PR solves it by clean the value (removes adapter information) before file_exists check. A new method is added to MediaHelper class so that we can use on other places when it is needed (or even could be used by third party extensions developer - for example, we need to check and make sure the selected image is valid before resizing...)

### Testing Instructions
1. Look at https://github.com/joomla/joomla-cms/issues/34810, confirm the issue
2. Apply patch, confirm that the issue is sorted.

### DIFF
--- a/layouts/joomla/form/field/media.php
+++ b/layouts/joomla/form/field/media.php
@@ -11,6 +11,7 @@ defined('_JEXEC') or die;
 
 use Joomla\CMS\Component\ComponentHelper;
 use Joomla\CMS\Factory;
+use Joomla\CMS\Helper\MediaHelper;
 use Joomla\CMS\HTML\HTMLHelper;
 use Joomla\CMS\Language\Text;
 use Joomla\CMS\Router\Route;
@@ -73,12 +74,19 @@ switch ($preview) {
 }
 
 // Pre fill the contents of the popover
-if ($showPreview) {
-	if ($value && file_exists(JPATH_ROOT . '/' . $value)) {
-		$src = Uri::root() . $value;
-	} else {
+if ($showPreview)
+{
+	$cleanValue = MediaHelper::getCleanMediaFieldValue($value);
+
+	if ($cleanValue && file_exists(JPATH_ROOT . '/' . $cleanValue))
+	{
+		$src = Uri::root() . $cleanValue;
+	}
+	else
+	{
 		$src = '';
 	}
+
 	$width = $previewWidth;
 	$height = $previewHeight;
 	$style = '';
--- a/layouts/joomla/form/field/media.php
+++ b/layouts/joomla/form/field/media.php
@@ -80,7 +80,7 @@ if ($showPreview)
 
 	if ($cleanValue && file_exists(JPATH_ROOT . '/' . $cleanValue))
 	{
-		$src = Uri::root() . $cleanValue;
+		$src = Uri::root() . $value;
 	}
 	else
 	{
--- a/libraries/src/Helper/MediaHelper.php
+++ b/libraries/src/Helper/MediaHelper.php
@@ -464,4 +464,25 @@ class MediaHelper
 
 		return false;
 	}
+
+	/**
+	 * Helper method get clean data for value stores in a Media form field by removing adapter information
+	 * from the value if available (in this case, the value will have this format:
+	 * images/headers/blue-flower.jpg#joomlaImage://local-images/headers/blue-flower.jpg?width=700&height=180)
+	 *
+	 * @param   string  $value
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION
+	 */
+	public static function getCleanMediaFieldValue($value)
+	{
+		if ($pos = strpos($value, '#'))
+		{
+			return substr($value, 0, $pos);
+		}
+
+		return $value;
+	}
 }
--- a/libraries/src/Helper/MediaHelper.php
+++ b/libraries/src/Helper/MediaHelper.php
@@ -474,7 +474,7 @@ class MediaHelper
 	 *
 	 * @return  string
 	 *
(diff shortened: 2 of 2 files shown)


richard67:
@joomdonation I guess it was the auto correct by the spell checker which changed „preview“ to „previous“ if the title here.

joomdonation:
@richard67 Thanks. Maybe it was a typo.

layouts/joomla/form/field/media.php
```diff
@@ -73,12 +74,19 @@
...
-	if ($value && file_exists(JPATH_ROOT . '/' . $value)) {
-		$src = Uri::root() . $value;
-	} else {
+if ($showPreview)
+{
+	$cleanValue = MediaHelper::getCleanMediaFieldValue($value);
+
+	if ($cleanValue && file_exists(JPATH_ROOT . '/' . $cleanValue))
+	{
+		$src = Uri::root() . $cleanValue;
```
dgrammatiko:
```suggestion
		$src = Uri::root() . $value;
```

I would prefer if we keep the original value here, eg: `$src = Uri::root() . $value;` to. match the rendered preview source from JS land. Check my images in your issue

dgrammatiko (changes requested):

joomdonation:
@dgrammatiko Thanks. I didn't know that you need full value for preview

dgrammatiko:
>  I didn't know that you need full value for preview

Actually, we don't, but changing the $src in that point changes also the value in the input (and I guess we need the full value there). Basically edit a banner select a new image, save close, reopen and save (you lose the extra bits), this is what the proposed change prevents (I have no clue if it will be a problem, just did run the code in my head, so I might be wrong)

joomdonation:
>Actually, we don't, but changing the $src in that point changes also the value in the input 

Hmm. Change the $src variable won't change the value displaying in the input. The value in the input is still taken from $value variable, I think.

dgrammatiko:
> Hmm. Change the $src variable won't change the value displaying in the input. The value in the input is still taken from $value variable, I think.

You're right, I need more coffee, sorry

joomdonation:
Hmm. Somehow, the src of the preview image still use the unclean value. Maybe you set it automatically base on the value from input in the JS code?

dgrammatiko:
> Hmm. Somehow, the src of the preview image still use the unclean value. Maybe you set it automatically base on the value from input in the JS code?

Yes, the preview src is just the source + `Uri:root(true)`

https://github.com/joomla/joomla-cms/blob/703484715a3de6fff26a7566e1ff779b99fc4dab/build/media_source/system/js/fields/joomla-field-media.w-c.es6.js#L212

Now that I think about it it's good to have the same URL both from PHP and JS otherwise there will be 2 downloads for the same image

joomdonation:
Main question here is we have source from php, then why do we need to calculate src js again?

dgrammatiko:
> Main question here is we have source from php, then why do we need to calculate src js again?

The update of the preview happens as part of the field's custom element initialization (`connectedCallback()`) and I guess the idea was to resolve external adapters here (but I doubt that the code does that). Anyways the line for triggering the JS resetting the preview html is here: 
https://github.com/joomla/joomla-cms/blob/703484715a3de6fff26a7566e1ff779b99fc4dab/build/media_source/system/js/fields/joomla-field-media.w-c.es6.js#L126

dgrammatiko:
I have tested this item :white_check_mark: successfully on e904530b42eb40605f3acfe5f75578bf340db23d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34833">issues.joomla.org/tracker/joomla-cms/34833</a>.</sub>

libraries/src/Helper/MediaHelper.php
```diff
@@ -464,4 +464,25 @@ public static function isValidLocalDirectory($directory)
...
+	/**
+	 * Helper method get clean data for value stores in a Media form field by removing adapter information
+	 * from the value if available (in this case, the value will have this format:
+	 * images/headers/blue-flower.jpg#joomlaImage://local-images/headers/blue-flower.jpg?width=700&height=180)
+	 *
+	 * @param   string  $value
+	 *
+	 * @return  string
+	 *
+	 * @since   __DEPLOY_VERSION
```
Quy:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

richard67:
I've restored the previous test result since the change after that was only in a comment.

Quy:
I have tested this item :white_check_mark: successfully on 88d662abd52414e97fedc13c90463abea334f372<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34833">issues.joomla.org/tracker/joomla-cms/34833</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34833">issues.joomla.org/tracker/joomla-cms/34833</a>.</sub>

wilsonge:
Thanks!

PR #34853, opened 2021-07-20
[4.0] unused legacy rtl scss

brianteeman:
As far as I can tell none of the classes removed here are actually used.
code review

### DIFF
--- a/administrator/templates/atum/scss/template-rtl.scss
+++ b/administrator/templates/atum/scss/template-rtl.scss
@@ -21,59 +21,11 @@ ul {
   left: auto;
 }
 
-.switcher {
-  float: right;
-}
-
 .menu-collapse {
   right: 0;
   left: auto;
   margin-right: -7.5px;
   margin-left: 0;
-}
-
-// Status
-.status {
-  margin-right: 250px;
-  margin-left: 0;
-
-  .closed & {
-    margin-right: 50px;
-    margin-left: 0;
-  }
-}
-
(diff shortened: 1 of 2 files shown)


administrator/templates/atum/scss/template-rtl.scss
```diff
@@ -21,61 +21,13 @@ ul {
...
-.sidebar-wrapper[data-hidden="1"] ~ .status {
-  margin-right: 0;
-}
-
-// com_cpanel
-.com_cpanel {
-  .card {
-    .list-group {
-      > .published {
-        border-right: 5px solid var(--success);
```
wilsonge, line 55: Strictly on code review. This looks like the counter part to https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/templates/atum/scss/pages/_com_cpanel.scss#L86 so superficially it is being used (either that or both parts aren't being used :) )

wilsonge:
On code review alone I think the published and unpublished *might* be being used. Agreed that the rest is not

brianteeman:
Would be good if someone could find out where its being used as I couldnt find it.

wilsonge:
I have to say I don't see it either. But in this case we should remove the LTR SCSS too rather than it being some halfway house

brianteeman:
will update the pr

brianteeman:
Updated as requested

wilsonge:
Thanks!

brianteeman:
thanks

PR #34859, opened 2021-07-21
[4.0] tags images

brianteeman:
This PR fixes a bug in RTL as shown in the images by using logical css properties instead of adding additional rtl specific classes. There is no visible change in LTR

### Before
image

### After
image


wilsonge:
Thanks!

brianteeman:
merci

PR #34860, opened 2021-07-21
[4.0] users groups margin

brianteeman:
This PR fixes a bug in RTL as shown in the images by using logical css properties instead of adding additional rtl specific classes.

### Before/After LTR
image

### Before TRL
image

### After RTL
image


wilsonge:
Thanks!

brianteeman:
small steps

PR #34862, opened 2021-07-22
Fixed the aria-describedby not being set correctly

roland-d:
### Summary of Changes
In a repeatable subform we replace the `X` with the number of the row being created, however this is not done for the `aria-describedby`. This fixes that and also updates the `aria-describedby` tag.

### Testing Instructions
1. Login to the Joomla 4 backend
2. Go to System -> Global Configuration
3. Click on Users at the bottom of the list on the left
4. Click on the `Email Domain Options` tab
5. Click on the + to add a new row
6. Inspect the Domain name input field
7. Notice that it says `<input type="text" name="jform[domains][__field10][name]" id="jform_domains____field10__name" value="" class="form-control required" aria-describedby="jform[domains][__field1X][name]-desc" required="">`
8. Apply the PR
9. Reload the page
10. Click on the + to add a new row
11. Inspect the Domain name input field
12. Notice that it now says `<input type="text" name="jform[domains][__field10][name]" id="jform_domains____field10__name" value="" class="form-control required" aria-describedby="jform[domains][__field10][name]-desc" required="">`

The `aria-describedby` is not updated
`<input type="text" name="jform[domains][__field10][name]" id="jform_domains____field10__name" value="" class="form-control required" aria-describedby="jform[domains][__field1X][name]-desc" required="">`

The `aria-describedby` is updated
`<input type="text" name="jform[domains][__field10][name]" id="jform_domains____field10__name" value="" class="form-control required" aria-describedby="jform[domains][__field10][name]-desc" required="">`

None

### DIFF
--- a/build/media_source/system/js/fields/joomla-field-subform.w-c.es6.js
+++ b/build/media_source/system/js/fields/joomla-field-subform.w-c.es6.js
@@ -270,6 +270,7 @@
       haveName.forEach((elem) => {
         const $el = elem;
         const name = $el.getAttribute('name');
+        const aria = $el.getAttribute('aria-describedby');
         const id = name
           .replace(/(\[\]$)/g, '')
           .replace(/(\]\[)/g, '__')
@@ -333,6 +334,10 @@
         $el.name = nameNew;
         if ($el.id) {
           $el.id = idNew;
+        }
+
+        if (aria) {
+          $el.setAttribute('aria-describedby', nameNew);
         }
 
         // Check if there is a label for this input
--- a/build/media_source/system/js/fields/joomla-field-subform.w-c.es6.js
+++ b/build/media_source/system/js/fields/joomla-field-subform.w-c.es6.js
@@ -337,7 +337,7 @@
         }
 
         if (aria) {
-          $el.setAttribute('aria-describedby', nameNew);
+          $el.setAttribute('aria-describedby', `${nameNew}-desc`);
         }
 
         // Check if there is a label for this input


brianteeman:
/me confused you write `aria-controlledby` but the expected result shows `aria-describedby`

dgrammatiko:
@roland-d I guess there should be a helper function to deal with all the aria-* attributes:
- aria-labeledby
- aria-discribedby
- aria-controls
- ariacontrolledby
- ...

I mean this PR is fine but since you're doing this maybe patch all the cases in one unique way?


PS What I mean was something like:
```js
const updateAria = (sourceElement, targetElement) => {
  [].slice.call(sourceElement.attributes)
    .filter(attribute => attribute.startsWith('aria-'))
   .map(attr => targetElement(attr, sourceElement.getAttribute(attr);
}

// Call it like
updateAria(source, target);
```

roland-d:
@brianteeman I have no idea where that came from, but it has to be `aria-describedby`. I updated the text to reflect the correct name.

@dgrammatiko I don't mind doing all the cases if someone can tell me what they are and where to find them. I just found this as I am having troubles with subform and this stood out.

dgrammatiko:
MDN https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA is not very helpful as there's no list with all the attributes, or I couldn't find it. Also looking at my code the `.map(attr => targetElement(attr, sourceElement.getAttribute(attr);` should update the target element attribute to the new state, so it needs a bit more code there...

brianteeman:
please remember that the first rule of aria is not to use it (unless you have to)

dgrammatiko:
> please remember that the first rule of aria is not to use it (unless you have to)

@brianteeman you're right but we are **not** adding aria, we are copying/patching any existing attributes from the original form element to reflect the copied version.

brianteeman:
was just a reminder to those that know and possibly new information for those that dont

ceford:
I have tested this item :white_check_mark: successfully on cd1f121d57f78ed368cb8435e82fc771b22a8966<br /><br />Works as described, and the next field added is 11.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34862">issues.joomla.org/tracker/joomla-cms/34862</a>.</sub>

wilsonge:
Thanks!

Issue #34866, opened 2021-07-22
[4.0] Category sort settings in menu item get ignored 


HLeithner:
### Steps to reproduce the issue
Create a menuitem of type "List All Categories in an Article Category Tree"
Change "Blog Layout" -> "Category Order" to "Title Alphabetical"

### Expected result
Categories are sorted by title

### Actual result
Categories are still sorted by ordering (or at least I think it's ordering)

### System information (as much as possible)
4.0-rc4

joomdonation:
At first, I thought this would be easy to fix by providing an order by providing an order by option to Joomla\CMS\Categories\Categories class instead of hard code order by clause by `c.lft` (ordering) https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Categories/Categories.php#L297 . Sadly, that won't work. Look like categories needs to be ordered by ordering like that to have the categories tree built properly

So look like the only way to implement this ordering setting is using `usort` function to sort the data before it is displayed (doing that in the layout or inside the Categories model). Not sure if it is something which we should try to implement.

chmst:
@HLeithner @joomdonation Just tested in a J3.9 and it did not work there either. Or am I missing something?

joomdonation:
@chmst Looking at our current code, I would say that this feature never worked !

brianteeman:
The option you refer to is not for changing the order of the categories on the first page of this menu item,

> Change "Blog Layout" -> "Category Order" to "Title Alphabetical"

 These options are also used when you select one of the category links, on the first page and/or thereafter, unless they are changed for a specific menu item.

![image](https://user-images.githubusercontent.com/1296369/126859010-8fcf3503-7b29-47a3-9401-e7d60ef54a57.png)

joomdonation:
@brianteeman Shouldn't the order (if could be implemented) be applied for categories from all of these layouts? I mean it should be moved to Shared tab, same with in component Options

![component_options](https://user-images.githubusercontent.com/977664/126859383-01650565-e1e2-473b-9eb6-d4a2c97b3b25.png)


Anyway, from what I see from the code, we don't handle this parameter in our code.

brianteeman:
i am still trying to work out how you use the blog and list layouts

joomdonation:
Sorry, I don't understand your last comment.

chmst:
Yesterday I spent a huge amount of time with this issue.
What I think now: The categories are displayed as a tree. No ordering is here possible and never was possible.

[...]

PR #34883, opened 2021-07-23
[4.2] Fix code logic issues in Categories model

joomdonation:
Pull Request for Issue # .

### Summary of Changes
In frontend Categories model of our components (com_content in this PR), there are few logic issues:

- `$params` in `getItems` method use params from the active menu item instead of calculated params stored in model states. With code like that, it will have wrong value for parameters set to **Use Global** in the menu item parameters (as it Use Global, it should get value from component Options but that's not working here because we are getting params directly from active menu item at the moment).

-  The returned categories list does not depend on value of two model states `filter.access` and `filter.published`. So if someone uses this model class and want to get list of unpublished categories (by setting `filter.published` state to 0, it won't work)

This PR fixes the two above logic issues in the code. If it is accepted, the similar changes should be done for other categories.

### Testing Instructions
This is just code logic issues and does not have much effect how the system works, so there is not anything difference which testers can see. The only thing you can test is apply PR, create a menu item to link to **List All Categories in an Article Category Tree** and see that it is still working as expected.

I hope someone could review code, too.


richard67:
Are there any issues which are solved with this fix? If so, it would be good to find and close them.

joomdonation:
@richard67 I don't think so. It is just code logic I found when I tried to find a solution for the issue https://github.com/joomla/joomla-cms/issues/34866 .

components/com_content/src/Model/CategoriesModel.php
```diff
@@ -110,21 +109,13 @@ public function getItems($recursive = false)
 
 		if (!isset($this->cache[$store]))
 		{
-			$app = Factory::getApplication();
-			$menu = $app->getMenu();
-			$active = $menu->getActive();
+			$params = $this->getState('params');
```
laoneo: If you load the model with ignore_request true, this can be null.

components/com_content/src/Model/CategoriesModel.php
```diff
@@ -110,21 +109,13 @@ public function getItems($recursive = false)
...
-			{
-				$params = $active->getParams();
-			}
-			else
-			{
-				$params = new Registry;
-			}
-
-			$options = array();
+			$options               = array();
```
laoneo: Short array

components/com_content/src/Model/CategoriesModel.php
joomdonation: @laoneo Thanks. Was thinking about that, too. Now set default value for it.

components/com_content/src/Model/CategoriesModel.php
joomdonation: Changed. But I prefer we do it using a phpcs fixer tool at the time we are ready to make change in the whole project.

components/com_content/src/Model/CategoriesModel.php
laoneo: I would go with the active menu, to be on the safe side.

joomdonation: Actually, when you use the model from outside, the active menu item is not reliable (it could be from a page from different extension, for example, when model code is used inside a module)

So you should pass the model state with the right value you expect. Or if it is really needed, we should default to com_content parameters.

laoneo: I mean I agree that the model should not be aware of menu items, but you have then here a BC break. At least I would then get the com_content params if the params from the state is null.

joomdonation: I think we can default to com_content parameters. Or in case params state is not set, we can set `$options['countItems'] = 0;`, same as how we do in Category model here https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_content/src/Model/CategoryModel.php#L373-L383

We should have consistent code in all places instead of having different behaviors in different places at the moment :( . Could we have a decision here?

laoneo: What you need for a decision here? Either do it the same in all models to have it consistent and clean or fix this one here without a BC break. But you can't just fix one which breaks bc clearly.

joomdonation: I want to know if we can do the same like we do in Category model https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_content/src/Model/CategoryModel.php#L373-L383. So in case custom code does not set params state, we just set `$options['countItems'] = 0; `

socke300:
I have tested this item :white_check_mark: successfully on da420f11fdedaa738c8920cefd9d8d23fd18979a<br /><br />Hey, I have tested the changes and tried different settings. 
I didn't notice anything else, except for the problem with sorting. #34866
It is still working as expected.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34883">issues.joomla.org/tracker/joomla-cms/34883</a>.</sub>

richard67:
@BertaOctech I just notice you have marked a test result, but not with the "Test this" button. It seems you have used "Alter test". Does that mean you have tested this pull request with success?

BertaOctech:
I have tested this item :white_check_mark: successfully on da420f11fdedaa738c8920cefd9d8d23fd18979a<br /><br />@richard67 
Hi,

Yes, you are right.

I did test the patch and I could not find anything wrong with it.

Cheers,

Berta<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34883">issues.joomla.org/tracker/joomla-cms/34883</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34883">issues.joomla.org/tracker/joomla-cms/34883</a>.</sub>

richard67:
@wilsonge I leave the decision to you with respect of the above review discussion.

bembelimen:
For me it's an absolutely valid PR which makes it more correct. So definitely a "yes" to get it in core, but a "no" for 4.1.x. So move to 4.2 for now (and most likely it has to go to 5.0 as @laoneo mentioned.

laoneo:
Removed RTC as it is easy to do it in a BC way. When done we can add RTC again.

richard67:
> Removed RTC as it is easy to do it in a BC way. When done we can add RTC again.

@laoneo For removing RTC it needs to change status in the issue tracker, not just to remove the label.

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34883">issues.joomla.org/tracker/joomla-cms/34883</a>.</sub>

PR #34884, opened 2021-07-23
After Switching Admin Language, in the message displayed, show Native Title instead of Title [in English]

imanickam:
Pull Request for Issue # .

It is not an issue; but, an improvement.

### Summary of Changes

For example, if German is Switched as the back-end Administrator language, following message is displayed:
Die Administrationssprache wurde auf "German (Germany)" geändert. 

With the change proposed, the following message would be displayed:
Die Administrationssprache wurde auf "Deutsch (Deutschland)" geändert.

When the message is displayed in the native language (e.g., German), it would be nice/better to show the language's Native Name/Title instead of the Name/Title in English.

### Testing Instructions

1. Log into Joomla! 4's back-end (_site-url/administrator_) with an account that has Super User privileges
2. Install the German language by going to the screen "Extensions: Languages" (System => Install -> Languages)
3. Go to the screen "Languages: Installed (Site)" (System => Manage -> Languages)
4. Choose "Administrator" from the dropdown (that show the client) - The screen "Languages: Installed (Administrator)" would be shown and the list of languages installed that has Administrator pack would be shown
5. Choose the language German and click on the button "Switch Language"

Now the German language would be the back-end administrator language.  Also, a message confirming the language switch would be displayed as shown below in the German language.

`Die Administrationssprache wurde auf "German (Germany)" geändert. `

Notice that even though the message is in German, the language name shown is in English.

NOW, APPLY THE PATCH WITH THE CHANGE PROPOSED AND REPEAT THE TEST STEPS 1 TO 5 AS LISTED ABOVE.

After, German language is made as the back-end administrator language, following message would be displayed confirming the language switch.

`Die Administrationssprache wurde auf "Deutsch (Deutschland)" geändert.
`

**Notice the difference in the name of the language displayed between the messages before and after the change.**

Notice that the message is in the native language (e.g, German) whereas the language name shown is in English.

Original_Language_Title_Shown

Notice that the message is in the native language (e.g., German) and the language name shown is also in the native language (e.g., German).

Revised_Language_Native-Title_Shown

Probably Joomla! 4 documentation screenshot in case if the switching message is shown in the screenshot.

### DIFF
--- a/administrator/components/com_languages/src/Controller/InstalledController.php
+++ b/administrator/components/com_languages/src/Controller/InstalledController.php
@@ -12,7 +12,7 @@ namespace Joomla\Component\Languages\Administrator\Controller;
 \defined('_JEXEC') or die;
 
 use Joomla\CMS\Factory;
-use Joomla\CMS\Installer\Installer;
+use Joomla\CMS\Language\LanguageHelper;
 use Joomla\CMS\Language\Language;
 use Joomla\CMS\Language\Multilanguage;
 use Joomla\CMS\Language\Text;
@@ -92,8 +92,8 @@ class InstalledController extends BaseController
 			$file = JPATH_ADMINISTRATOR . '/language/' . $cid . '/' . $cid . '.xml';
 		}
 
-		$info         = Installer::parseXMLInstallFile($file);
-		$languageName = $info['name'];
+		$info = LanguageHelper::parseXMLLanguageFile($file);
+		$languageName = $info['nativeName'];
 
 		if ($model->switchAdminLanguage($cid))
 		{
--- a/administrator/components/com_languages/src/Controller/InstalledController.php
+++ b/administrator/components/com_languages/src/Controller/InstalledController.php
@@ -92,7 +92,7 @@ class InstalledController extends BaseController
 			$file = JPATH_ADMINISTRATOR . '/language/' . $cid . '/' . $cid . '.xml';
 		}
 
-		$info = LanguageHelper::parseXMLLanguageFile($file);
+		$info         = LanguageHelper::parseXMLLanguageFile($file);
 		$languageName = $info['nativeName'];
 
 		if ($model->switchAdminLanguage($cid))


richard67:
Can we be sure that all language packs provide the `nativeName` in their XML?

administrator/components/com_languages/src/Controller/InstalledController.php
```diff
@@ -92,8 +92,8 @@ public function switchAdminLanguage()
 			$file = JPATH_ADMINISTRATOR . '/language/' . $cid . '/' . $cid . '.xml';
 		}
 
-		$info         = Installer::parseXMLInstallFile($file);
-		$languageName = $info['name'];
+		$info = LanguageHelper::parseXMLLanguageFile($file);
+		$languageName = $info['nativeName'];
```
richard67:
```suggestion
		$info         = LanguageHelper::parseXMLLanguageFile($file);
		$languageName = $info['nativeName'] ?? $info['name'];
```
1. Restore the alignment of "=" like it was before.
2. To me it seems that we can't rely on the nativeName always being defined, see e.g. this code here https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Language/LanguageHelper.php#L52
or here https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_languages/src/Model/InstalledModel.php#L194

richard67 (dismissed):
See comment below my review suggestion.

Bakual:
In J4, you can assume `nativeName` to be always present. In J3 there can be old packs which don't have it (it was added in a J3 release) but it's a required field since quite some time.
So if it's missing, it's a bug in the language pack.

administrator/components/com_languages/src/Controller/InstalledController.php
richard67: Hmm I'm just told that the nativeName is a required field in XML of J4 languages, so this PR might be ok as it is (except maybe the small code style change).

administrator/components/com_languages/src/Controller/InstalledController.php
```diff
@@ -92,8 +92,8 @@ public function switchAdminLanguage()
 			$file = JPATH_ADMINISTRATOR . '/language/' . $cid . '/' . $cid . '.xml';
 		}
 
-		$info         = Installer::parseXMLInstallFile($file);
-		$languageName = $info['name'];
+		$info = LanguageHelper::parseXMLLanguageFile($file);
```
richard67:
```suggestion
		$info         = LanguageHelper::parseXMLLanguageFile($file);
```
Align "=" with the one in the row below.

imanickam (commented):
Aligned the = sign as suggested.

administrator/components/com_languages/src/Controller/InstalledController.php
imanickam: Aligned the "=" as suggested.

richard67:
I have tested this item :white_check_mark: successfully on 800730dca386a95d176a53121d8c701f46efef51<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34884">issues.joomla.org/tracker/joomla-cms/34884</a>.</sub>

richard67:
I've tested with all currently available J4 languages, which currently are:

2021-07-24_Language-switching_languages-available

It seems the Portuguese language pack uses wrong "_QQ_" in the language string for the message:

2021-07-24_Language-switching-errors_pt-PT

But it seems that's already fixed on Crowdin and so should be ok in future versions of that language pack.

ceford:
I have tested this item :white_check_mark: successfully on 800730dca386a95d176a53121d8c701f46efef51<br /><br />Works with German, French and Persion<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34884">issues.joomla.org/tracker/joomla-cms/34884</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34884">issues.joomla.org/tracker/joomla-cms/34884</a>.</sub>

brianteeman:
surely the native title for french is wrong and still in english

brianteeman:
and welsh

richard67:
Yes ... they have to fix that in their translation packages.

wilsonge:
Thanks!

PR #34888, opened 2021-07-24
[4.0] Codemirror update

brianteeman:
21-07-2021: Version 5.62.2:

- lint addon: Fix a regression that broke several addon options.

20-07-2021: Version 5.62.1:

- vim bindings: Make matching of upper-case characters more Unicode-aware.
- lint addon: Prevent options passed to the addon itself from being given to the linter.
- show-hint addon: Improve screen reader support.
- search addon: Avoid using innerHTML.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2755,9 +2755,9 @@
       }
     },
     "node_modules/codemirror": {
-      "version": "5.62.0",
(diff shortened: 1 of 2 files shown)


brianteeman:
Thanks

wilsonge:
Thanks!

PR #34891, opened 2021-07-25
[4.0] Improve $app->getParams logic

joomdonation:
Pull Request for Issue # .

### Summary of Changes
The `getParams` method in `SiteApplication ` class https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Application/SiteApplication.php#L284 return component parameters after merging it with active menu item parameters (the menu item parameters take higher priority). However, it does not check  and make sure the menu item is linked to current component before merging, and this PR just adds that missing check. 

### Testing Instructions
This need code review from developers understand how parameters is handle in Joomla!


winniecarole:
The improvement is okay but it doesn't decrease the code length.
image

PR #34892, opened 2021-07-25
Simplify com_content params model state

joomdonation:
Pull Request for Issue # .

### Summary of Changes
The $params variable returned from $app->getParams(); already contains component parameters merged with active menu item parameters, so no need to merge it with active menu item parameters again. 

For reference, we do it right in Categories model https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_content/src/Model/CategoriesModel.php#L69-L70

### Testing Instructions
Code review only.

PR #34894, opened 2021-07-25
Simplify params model state for com_contact, com_newsfeeds

joomdonation:
Pull Request for Issue # .

### Summary of Changes
Use $app->getParams() to get $params variable. Less code, consistent with other model classes.

### Testing Instructions
Code review should be enough.


components/com_contact/src/Model/CategoryModel.php
```diff
@@ -267,20 +267,9 @@ protected function getListQuery()
 	protected function populateState($ordering = null, $direction = null)
 	{
 		$app = Factory::getApplication();
-		$params = ComponentHelper::getParams('com_contact');
```
Quy: Please remove `use Joomla\CMS\Component\ComponentHelper;`.

joomdonation: Thanks @Quy.

wilsonge:
https://github.com/joomla/joomla-cms/blob/staging/components/com_content/models/category.php#L117-L127 We have the same code in com_content at the moment.

Newsfeeds doesn't even merge the menu specific items :) https://github.com/joomla/joomla-cms/blob/staging/components/com_newsfeeds/models/category.php#L197

Overall I think this is right - we just need to get everything consistent at once.

joomdonation:
@wilsonge We handle it different in com_content, thus I made a separate PR https://github.com/joomla/joomla-cms/pull/34892 for com_content.

For newsfeeds, yes, we have it in different way, too. So 3 components, should work in the same way but we handle it different in each component :D. I will add the same change to com_newsfeeds into this PR

joomdonation:
OK. I use similar code for com_newsfeeds. Assume it is accepted, we will have consistent code for the three components :).

HLeithner:

rdeutz (approved):

drmenzelit (approved):

PR #34901, opened 2021-07-25
[4.0] Pass $app to Module Helper

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR modify some of our frontend modules code to pass application option (which is available via $app variable) to module helper methods. This helps:

- Using injected $app object instead of having to call Factory::getApplication() to get application object (isn't it the purpose of our new module dispatcher architecture?) ?
- Allow replacing deprecated Factory::getUser() method with $app->getIdentity();

As these helper methods change in this PR is public method, I'm unsure if changing parameters here are consider backward incompatible changes. If it is, please let me know and I will close the PR. Just want to improve the code a bit.

### Testing Instructions
- Code review
- Or publish all the modules changed in this PR and make sure it is still working as expected.


richard67:
> As these helper methods change in this PR is public method, I'm unsure if changing parameters here are consider backward incompatible changes. If it is, please let me know and I will close the PR. Just want to improve the code a bit.

@wilsonge Your opinion?

modules/mod_articles_popular/mod_articles_popular.php
```diff
@@ -22,6 +22,6 @@
 	return;
 }
 
-$list = ArticlesPopularHelper::getList($params);
+$list = ArticlesPopularHelper::getList($params, $params);
```
dgrammatiko:
```suggestion
$list = ArticlesPopularHelper::getList($params, $app);
```

wilsonge:
> @wilsonge Your opinion?

This should be fine. Extensions officially anyhow don't fall under the b/c pledge and tbh no one should be using the modules helpers in their own extensions.

modules/mod_articles_news/src/Helper/ArticlesNewsHelper.php
```diff
@@ -30,16 +29,15 @@ abstract class ArticlesNewsHelper
 	/**
 	 * Get a list of the latest articles from the article model
 	 *
-	 * @param   \Joomla\Registry\Registry  &$params  object holding the models parameters
+	 * @param   \Joomla\Registry\Registry  &            $params  object holding the models parameters
```
Quy: Please fix `&`. Do the the same in the other files.

laoneo:
I like the change. What about making the app an optional parameter and when not set it is fetched from Factory? Pretty sure there are extensions which are using the helpers, especially the ones from content. We can add a warning when the app is not set that in J5 it will be a required parameter.

joomdonation:
> What about making the app an optional parameter and when not set it is fetched from Factory? 

I'm unsure about this one. The data returned from the method usually depend on parameters from the module itself, so I'm unsure if third party developers should use module helpers, they should use model instead. @wilsonge How do you think about  this ?

laoneo:
> so I'm unsure if third party developers should use module helpers

Even when they shouldn't use them, pretty sure they do. There is no reason to break here. We have been very carefully to make as less BC breaks as possible for J3 extensions to run on J4. Honestly I do not understand why now in RC phase we are leaving this path.

joomdonation:
@laoneo I'm just asking for decision from @wilsonge as he stated earlier

> This should be fine. Extensions officially anyhow don't fall under the b/c pledge and tbh no one should be using the modules helpers in their own extensions.

The reason is because if it is allowed, I would like to make our modules consistent. Right now, we code it in different way in different modules. I asked the question is yesterday https://github.com/joomla/joomla-cms/discussions/34907

If we decide to keep it BC, I will make the change, no problem. Just wait for final decision.

laoneo:
It would be nice to have them all working the same way but please no BC breaks even when it doesn't go under our BC promise. It should be possible to use them in J3 and J4 the same thats what I'm asking for.

HLeithner:
I would like to get rid of everything that's called static. No Idea if this is possible in joomla but the DI introduced is the first step. So this PR doesn't make sense at this time.

Maybe we can talk about this concept.

joomdonation:
@HLeithner Isn't this PR removes static calls `Factory::getApplication()` and `Factory::getUser()` ?

HLeithner:
@joomdonation true but you change a static class and we should get rid of static classes at least that's my opinion.

So just changing this function call is a bit unnecessary.  if we are able to remove all static classes and ask the booted component for the helper which returns the "helper" or anything other object that can do the same which already knows everything about the the current joomla instance.

modules/mod_articles_news/src/Helper/ArticlesNewsHelper.php
```diff
@@ -30,16 +29,15 @@ abstract class ArticlesNewsHelper
 	/**
 	 * Get a list of the latest articles from the article model
 	 *
-	 * @param   \Joomla\Registry\Registry  &$params  object holding the models parameters
+	 * @param   \Joomla\Registry\Registry               &$params  object holding the models parameters
+	 * @param   \Joomla\CMS\Application\CMSApplication  $app     The application
```
PhilETaylor, line 33:
```suggestion
	 * @param   \Joomla\CMS\Application\CMSApplication  $app      The application
```

modules/mod_articles_popular/src/Helper/ArticlesPopularHelper.php
```diff
@@ -28,14 +28,13 @@ abstract class ArticlesPopularHelper
 	/**
 	 * Get a list of popular articles from the articles model
 	 *
-	 * @param   \Joomla\Registry\Registry  &$params  object holding the models parameters
+	 * @param   \Joomla\Registry\Registry               &$params  object holding the models parameters
+	 * @param   \Joomla\CMS\Application\CMSApplication  $app     The application
```
PhilETaylor, line 32:
```suggestion
	 * @param   \Joomla\CMS\Application\CMSApplication  $app      The application
```

modules/mod_languages/src/Helper/LanguagesHelper.php
```diff
@@ -30,16 +30,16 @@ abstract class LanguagesHelper
 	/**
 	 * Gets a list of available languages
 	 *
-	 * @param   \Joomla\Registry\Registry  &$params  module params
+	 * @param   \Joomla\Registry\Registry               &$params  module params
+	 * @param   \Joomla\CMS\Application\CMSApplication  $app     The application
```
PhilETaylor, line 34:
```suggestion
	 * @param   \Joomla\CMS\Application\CMSApplication  $app      The application
```

modules/mod_stats/src/Helper/StatsHelper.php
```diff
@@ -27,13 +27,13 @@ class StatsHelper
 	/**
 	 * Get list of stats
 	 *
-	 * @param   \Joomla\Registry\Registry  &$params  module parameters
+	 * @param   \Joomla\Registry\Registry               &$params  module parameters
+	 * @param   \Joomla\CMS\Application\CMSApplication  $app     The application
```
PhilETaylor, line 31:
```suggestion
	 * @param   \Joomla\CMS\Application\CMSApplication  $app      The application
```

joomdonation:
We will convert our modules to new structure, so the change propose in this PR is not needed anymore. Closing.

PR #34905, opened 2021-07-25
[4.0]Translate menus

brianteeman:
On a site with multiple admin languages you would expect that the name of the menu could be translated.

## Test instructions
Create an admin menu and give it a title such as JADMINISTRATOR
Create a site menu and give it a title such as JSITE

## Expected behaviour
The names of both menus are translated

## Actual behaviour
Only the name of the admin menu is translated

## Solution
~I have no idea why but~ there is an extra space at the end of `$item->title` so the string being passed through JText is "JSITE " which obviously doesn't match. ~I have no idea why there is not an extra space for the admin menu~

image

This PR trims `$item->title` so that now the site menu name is also translated.

~No idea if this is a hack/bandage or the correct solution but its been bugging me for hours and I can't find the space~

### Before
image

image

### After
image

image

### DIFF
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -989,7 +989,7 @@ class MenusHelper extends ContentHelper
 		// Translate attributes for iterator values
 		foreach ($replace as $var => $val)
 		{
-			$item->title   = str_replace("{sql:$var}", $val, $item->title);
+			$item->title   = str_replace("{sql:$var}", $val, trim($item->title));
 			$item->element = str_replace("{sql:$var}", $val, $item->element);
 			$item->link    = str_replace("{sql:$var}", $val, $item->link);
 			$item->class   = str_replace("{sql:$var}", $val, $item->class);
@@ -998,7 +998,6 @@ class MenusHelper extends ContentHelper
 		}
 
 		$item->setParams($params);
-
 		return $item;
 	}
 }
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -989,7 +989,7 @@ class MenusHelper extends ContentHelper
 		// Translate attributes for iterator values
 		foreach ($replace as $var => $val)
 		{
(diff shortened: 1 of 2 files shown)


administrator/components/com_menus/src/Helper/MenusHelper.php
```diff
@@ -998,7 +998,6 @@ protected static function parseXmlNode($node, $replace = array())
 		}
 
 		$item->setParams($params);
-
```
richard67: Please revert removal of the empty line before the return because it is required for code style.

administrator/components/com_menus/src/Helper/MenusHelper.php
```diff
@@ -998,7 +998,7 @@ protected static function parseXmlNode($node, $replace = array())
...
 		return $item;
+
```
richard67: The empty line is needed before and not after the return;

administrator/components/com_menus/src/Helper/MenusHelper.php
```diff
@@ -989,7 +989,7 @@ protected static function parseXmlNode($node, $replace = array())
 		// Translate attributes for iterator values
 		foreach ($replace as $var => $val)
 		{
-			$item->title   = str_replace("{sql:$var}", $val, $item->title);
+			$item->title   = str_replace("{sql:$var}", $val, ($item->title));
 			$item->element = str_replace("{sql:$var}", $val, $item->element);
```
richard67: The last commit removed the trim but the brackets are still there. Am confused.

richard67:
@brianteeman Can it be that the space comes from here? https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_menus/presets/default.xml#L147

brianteeman:
Will check in the morning but looks like it could be

brianteeman:
Thanks @richard67 well spotted.

richard67:
@brianteeman I know it's not related to this PR. but as you are touching that file, maybe you could fix the typo in the comment here, too?

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_menus/presets/default.xml#L133

"The repeated elements are place" should be changed to "The repeated elements are placed".

wilsonge:
Thanks!

brianteeman:
thx

PR #34906, opened 2021-07-26
[4.0] Use getIdentity for plugins

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR does two things:
- Add $app property to plugins (when it is missing) and use that injected application object to replace Factory::getApplication()
- Use $this->app->getIdentity() to get current user object instead of the deprecated method Factory::getUser();

### Testing Instructions
Add it is small changes but target multiple plugins, I prefer to have careful code review here.


wilsonge:
> I excludes system plugins from these changes because I'm unsure if it is safe to use $this->app->getIdentity() method in system plugins yet.

You've included a system plugin :D I'm also unsure though. Action logs (and user plugins) will also be loaded on update - hence why they get the extra warnings in the pre-update checker - so this probably does need a test unless you remove all of those cases

plugins/editors/tinymce/tinymce.php
```diff
@@ -174,7 +174,7 @@ public function onDisplay(
 			return $editor;
 		}
 
-		$user     = Factory::getUser();
```
wilsonge, line 177: There's another instance in `getGlobalFilters`

plugins/system/privacyconsent/privacyconsent.php
```diff
@@ -284,7 +284,7 @@ public function onAfterRoute()
 			return;
 		}
 
-		$userId = Factory::getUser()->id;
+		$userId = $this->app->getIdentity()->id;
```
wilsonge, line 287: Another one on [Line 715](https://github.com/joomla/joomla-cms/blob/a909f52d1d8ba6e1b47ea0c38745c6c38d652767/plugins/system/privacyconsent/privacyconsent.php#L715). But assuming by your comment you didn't mean to commit this

plugins/user/token/token.php
```diff
@@ -244,7 +244,7 @@ public function onContentPrepareForm(Form $form, $data): bool
 
 		// No token: no reset
 		$userTokenSeed = $this->getTokenSeedForUser($userId);
-		$currentUser   = Factory::getUser();
+		$currentUser   = $this->app->getIdentity();
```
wilsonge, line 247: Also Line 551 https://github.com/joomla/joomla-cms/blob/a909f52d1d8ba6e1b47ea0c38745c6c38d652767/plugins/user/token/token.php#L551

wilsonge (commented):
overall seems ok. But need to test the 3.x update case and a couple of places missed.

plugins/editors/tinymce/tinymce.php
joomdonation: That's static method, so we could not use `$this->app->getIdentity()`. It is now replaced with `Factory::getApplication()->getIdentity()`.

plugins/system/privacyconsent/privacyconsent.php
joomdonation: That's different code (return User object of given User ID). That would have to be something different and should be in a separate PR. On this PR, I only replace the code to get current logged in user only.

plugins/user/token/token.php
joomdonation: Same with above. It should be address in a different PR. And just for more information, we do not have equivalent code to replace this deprecated code yet. That code would have to replace with something like:

```php
Factory::getContainer()->get(UserFactoryInterface::class)->loadUserById($userId);
```

Or 
```php
$this->app->getContainer()->get(UserFactoryInterface::class)->loadUserById($userId);
```

The difference is that the original cache the user object by ID, but the new code always create a new user, even same user id is passed for second time (a bit of un-optimized).

One more thing, although userFactory is passed into application, it is private property for some reasons, so to access to user factory, we still have to access it via container instead of injected property as adviced.

joomdonation:
> You've included a system plugin :D I'm also unsure though

After looking at our Application classes more carefully, I see that the `Identity` is available very early in application life cycle. It is available in initialise phase, right after **Factory::getUser()** line is called https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Application/SiteApplication.php#L621 , long before system plugins are imported and triggered, so it is safe to use **$this->app->getIdentity();** to get current logged in user.

With that said, I do the replacement in system plugins, too. Will test with update from Joomla 3.10 to 4.0 to make sure everything is still working OK.

joomdonation:
So I tested the update from 3.10.0-dev to the package on PR and it is working OK. Tested several features in the CMS like login, logout, action logs, editor-xtds... changes by this PR and it is working OK, too.

laoneo (approved):

plugins/editors-xtd/article/article.php
```diff
@@ -22,6 +21,14 @@
  */
 class PlgButtonArticle extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/content/pagenavigation/pagenavigation.php
```diff
@@ -24,6 +24,14 @@
  */
 class PlgContentPagenavigation extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/editors-xtd/contact/contact.php
```diff
@@ -22,6 +21,14 @@
  */
 class PlgButtonContact extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/editors-xtd/image/image.php
```diff
@@ -22,6 +21,14 @@
  */
 class PlgButtonImage extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/editors-xtd/menu/menu.php
```diff
@@ -22,6 +21,14 @@
  */
 class PlgButtonMenu extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/editors-xtd/module/module.php
```diff
@@ -22,6 +21,14 @@
  */
 class PlgButtonModule extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

plugins/editors-xtd/pagebreak/pagebreak.php
```diff
@@ -21,6 +20,14 @@
  */
 class PlgButtonPagebreak extends CMSPlugin
 {
+	/**
+	 * The application object
+	 *
+	 * @var    \Joomla\CMS\Application\CMSApplication
+	 * @since  4.0.0
```
PhilETaylor:
```suggestion
	 * @since  __DEPLOY_VERSION__
```

HLeithner:

laoneo:
Can you fix the conflicts here? Keep in mind that the service provider plugins have the app already injected. Would love to get this merged.

joomdonation:
There are so many conflicts now, so I'm closing this PR. Will review it at a later time base on latest code from plugins which we are having now.

Issue #34908, opened 2021-07-26
[4.0] Error "The template for this display is not available" after update from 3.10


Bond97:
Maybe that error it's ok, but I want to create issue. Or it's not ok?

### Steps to reproduce the issue
1. Install clean 3.9.28
2. Update to 3.10 alpha 9
3. Set update url https://update.joomla.org/core/test/310to4_list.xml
4. Update to Joomla 4 rc-4
5. See error at the screenshot on "Actual result"

### Expected result
Update process without any errors

### Actual result
![изображение](https://user-images.githubusercontent.com/78050514/126943186-53b967c0-ff4e-47dc-a462-c78214319717.png)

### System information (as much as possible)

PHP 8.0.8, Mysql 5.7.34

Kostelano:
I was getting the same error when I upgraded from 3.9 to 3.10 to 4.0.

richard67:
It is well known that this is shown one time after the update and then disappears and doesn't come again when you start to navigate around in backend.

Kostelano:
Everything is correct. However, I'm starting to curl my fingers how many more of these Issue will be created upon the release of the stable version 4.0.

richard67:
Yes, that would be a pain. I'm just clarifying if we have some docs about it or if we need some or need more.

Bond97:
@richard67 maybe it will be better to show full description about this? When Joomla 4 goes to stable version, users will be ask questions about it.

For example, "The template Isis is not supported with Joomla 4 and now, after upgrade from Joomla 3.10, it was deleted"

richard67:
> For example, "The template Isis is not supported with Joomla 4 and now, after upgrade from Joomla 3.10, it was deleted"

It's an error message shown in general when the template is not available. We don't have one for that special case of just having updated from 3.10 to 4.0. So if we just change the message text as suggested, the message will be wrong in other cases of missing templates.

brianteeman:
So maybe we should look at changing that so it is a specific error with a specific message OR work to prevent that message appearing in the first place.

brianteeman:
Looking at the code and considering what situation would trigger this message I would suggest not outputting a message at all by removing the following line or replacing it with a log message (ie not displayed on screen)

https://github.com/joomla/joomla-cms/blob/eeb71ee627f4328474edac300bce05bfb6d28623/libraries/src/Application/AdministratorApplication.php#L275

[...]

PR #34912, opened 2021-07-26
[4.0] Change message to log

brianteeman:
Pull Request for Issue #34908 .

When you update from 3.10 you immediately get an error message `The template for this display is not available.`

On an update this is a false positive.

This PR changes it from a visible message to a log

https://github.com/joomla/joomla-cms/issues/34908#issuecomment-886514944

testing is time consuming for such a minor change and can probably be accepted by code review

### DIFF
--- a/libraries/src/Application/AdministratorApplication.php
+++ b/libraries/src/Application/AdministratorApplication.php
@@ -273,7 +273,7 @@ class AdministratorApplication extends CMSApplication
 		if (!is_file(JPATH_THEMES . '/' . $template->template . '/index.php')
 			&& !is_file(JPATH_THEMES . '/' . $template->parent . '/index.php'))
 		{
-			Log::add(Text::_('JERROR_ALERTNOTEMPLATE'), Log::WARNING, 'jerror');
+			Log::add(Text::_('JERROR_ALERTNOTEMPLATE'), Log::WARNING, 'Update');
 
 			$template->params = new Registry;
 			$template->template = 'atum';
--- a/libraries/src/Application/AdministratorApplication.php
+++ b/libraries/src/Application/AdministratorApplication.php
@@ -17,6 +17,7 @@ use Joomla\CMS\Filter\InputFilter;
 use Joomla\CMS\Input\Input;
 use Joomla\CMS\Language\LanguageHelper;
 use Joomla\CMS\Language\Text;
+use Joomla\CMS\Log\Log;
 use Joomla\CMS\Plugin\PluginHelper;
 use Joomla\CMS\Router\Router;
 use Joomla\CMS\Session\Session;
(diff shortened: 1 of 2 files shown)


libraries/src/Application/AdministratorApplication.php
```diff
@@ -272,7 +273,8 @@ public function getTemplate($params = false)
 		if (!is_file(JPATH_THEMES . '/' . $template->template . '/index.php')
 			&& !is_file(JPATH_THEMES . '/' . $template->parent . '/index.php'))
 		{
-			$this->enqueueMessage(Text::_('JERROR_ALERTNOTEMPLATE'), 'error');
+			Log::add(Text::_('JERROR_ALERTNOTEMPLATE'),Log::WARNING,'jerror');
```
dgrammatiko:
```suggestion
			Log::add(Text::_('JERROR_ALERTNOTEMPLATE'), Log::WARNING, 'jerror');
```

brianteeman: I did wonder about that but as it was a direct copy paste I didnt.

richard67:
> testing is time consuming for such a minor change and can probably be accepted by code review

Can be easily tested by using the update package or custom update URL build by drone.

richard67:
@brianteeman I've just tested updating a clean, current 3.10-dev to the update package built by drone for this PR.

There is still an alert, only it has changed from red to yellow.

2021-07-26_test-pr-34912_1

No idea where that comes from. Either it is issued at another place again, or the log is somehow used for alerts, too.

richard67:
The 'jerror' last parameter makes it go into an alert. ~~Maybe use 'error', or if that works 'Update'? In the latter case it should go then into the Joomla Update log.~~

Fedik:
> This PR changes it from a visible message to a log

it does not change, all `jerror` logs goes to message
https://github.com/joomla/joomla-cms/blob/ed9d8e877152cb5404953ec696f8027896a3adf8/libraries/bootstrap.php#L77-L82

brianteeman:
Updated to generate a build so I can test "Update" which is how its used elsewhere

Fedik:
And changing to `Update` is not allowed ;)
Because `AdministratorApplication::getTemplate()` is general method, so error is joomla runtime error, not specificity "update" error

The same error we have when you trying to load a non existing user with `JUser::getInstance(<random id>)`

brianteeman:
ok I give up

Fedik:
> ok I give up

that  was to fast :)

I would remove this code
https://github.com/joomla/joomla-cms/blob/ed9d8e877152cb5404953ec696f8027896a3adf8/libraries/bootstrap.php#L77-L82
And then make the changes you made here.

But not sure if it allowed :)

PR #34914, opened 2021-07-26
[4.0] Remove jerror logging in to messagequeue logger

Fedik:
Pull Request for Issue #34908

### Summary of Changes

I have removed messagequeue logger from bootstrap, and added changes from #34912 by @brianteeman 

### Testing Instructions
Please follow #34908

Logging of `jerror` can be configured in global configuration:
Screenshot_2021-07-26_15-33-40

### DIFF
--- a/libraries/bootstrap.php
+++ b/libraries/bootstrap.php
@@ -74,12 +74,6 @@ if (in_array('phar', stream_get_wrappers()))
 // Define the Joomla version if not already defined.
 defined('JVERSION') or define('JVERSION', (new JVersion)->getShortVersion());
 
-// Set up the message queue logger for web requests
-if (array_key_exists('REQUEST_METHOD', $_SERVER))
-{
-	JLog::addLogger(['logger' => 'messagequeue'], JLog::ALL, ['jerror']);
-}
-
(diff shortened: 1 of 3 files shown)


wilsonge:
I think this change is too much. That's a too big a b/c break at this stage. I think @brianteeman 's PR is the right lines. We probably just need a general log file to be created. There are other system level logs we throw in various places for the category. Obviously I understand jerror is not good.

Fedik:
>  That's a too big a b/c break at this stage.

I think no one will cry for this kind of logging (with use of an alert), 
anyway it is useless for an average User :)

As you decide.

@brianteeman PR also okay, but then we need another category, that not a `jerror` and not `update`.
Or can just remove the log from that place, in general

wilsonge:
> I think no one will cry for this kind of logging (with use of an alert),
> anyway it is useless for an average User :)

I agree but I know several extension take advantage of it. It's just not worth the pain.

> @brianteeman PR also okay, but then we need another category, that not a jerror and not update.

agreed

> Or can just remove the log from that place, in general

Probably worth keeping in case of any very weird edge cases.

PR #34918, opened 2021-07-27
[4.0] Remove temp url code in adapter interface as it is not used

laoneo:
### Summary of Changes
Removes the temporary url code as it was used in the past in an unit test. This test got removed already.
The code has ben introduced in 77e9696a9e4c3d0b89065350d5f88df2e078b30e.

### Testing Instructions
Browse around in media manager.

Everything works.

Everything works.


wilsonge:
Thanks!

PR #34942, opened 2021-07-27
[4.0] Atum: Prevent warping custom logos in header

crystalenka:
Pull Request for Issue #34355 .

### Summary of Changes
By adding a few lines of CSS, we are able to ensure that custom logos in the admin template are not warped even if they don't fit the exact dimensions of the space available.

### Testing Instructions

1. Go to `System > Templates > Administrator Template Styles`.
2. Open a style of your choice and navigate to `Image Settings`.
3. Scroll down to "Brand Large". Upload an image with dimensions other than 30x150 pixels.
4. Scroll down to "Brand Small". Upload an image that is not square aspect ratio.
5. Save and close.

### Actual result BEFORE applying this Pull Request
"Brand Large" and "Brand Small" images are stretched and warped to fit specific pixel dimensions regardless of their original proportions or size.

### Expected result AFTER applying this Pull Request
"Brand Large" and "Brand Small" display with the correct aspect ratio, within the space available.

### Documentation Changes Required
None.

PR #34943, opened 2021-07-27
[4,0] #34942 fix

brianteeman:
Merged in haste.

This ensures it works in RTL

### DIFF
--- a/administrator/templates/atum/scss/blocks/_header.scss
+++ b/administrator/templates/atum/scss/blocks/_header.scss
@@ -54,7 +54,15 @@
       height: 30px;
       margin: 0 .35rem;
       object-fit: contain;
-      object-position: left center;
+
+      [dir=ltr] & {
+        object-position: left center;
+      }
+
+      [dir=rtl] & {
+        object-position: right center;
+      }
+
       &.logo-collapsed {
         display: none;
       }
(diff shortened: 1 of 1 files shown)


richard67:
I have tested this item :white_check_mark: successfully on 25d77a0f7319295021f4491b63e689a367a0fb52<br /><br />Code review only, but that should be sufficient here.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34943">issues.joomla.org/tracker/joomla-cms/34943</a>.</sub>

wilsonge:
Removed the dir=ltr because we don't have that in other places and technically if dir is absent it gives a fallback (not sure practically on our template if that can ever happen but better safe than sorry). It's probably technically the right thing - but if we do it here we should probably do it all over the shop rather than in this one place.

brianteeman:
We do it the correct way like this in lots of places. Not happy you changed the pr and merged

richard67:
I remember having seen both ways, so it seems we are not consistent with that.

brianteeman:
yes - everything created by one person was done the wrong way or just shoved in a separate rtl file. Which is why there were so many bugs to fix as you don't know there is a need to update in other places. bad practice leads to bad code leads to more work.

</rant>

wilsonge:
I just searched and we currently use dir=ltr in 3 places in all the SCSS twice in joomla-tab scss override and once in quickicons. I very specifically said that what you did is probably technically right - just it's not good to just add it to one more place - it should be done at once or not at all.

Issue #34936, opened 2021-07-27
[4.0] Underlined link to all notes


Kostelano:
### Steps to reproduce the issue
Users, create a new note. Go back to the users page, notice that there is a style inconsistency for links in the dropdown. 

![Screenshot_1](https://user-images.githubusercontent.com/8440661/127217578-8ec681e2-bc45-48b0-8aba-1cc0290ee254.png)

brianteeman:
Please test #36947

PR #34945, opened 2021-07-28
[4.0] Underline issue solved for users manage

rhythmjain32:
Pull Request for Issue #34936

### Summary of Changes
There was an underline on Add Note button on the users manage page.  

before

underlin-cms

### DIFF
--- a/administrator/templates/atum/scss/vendor/bootstrap/_table.scss
+++ b/administrator/templates/atum/scss/vendor/bootstrap/_table.scss
@@ -2,6 +2,7 @@
 
 .table {
   thead {
+
     th {
       font-size: $label-font-size;
     }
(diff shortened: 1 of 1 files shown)


brianteeman:
I have tested this item :red_circle: unsuccessfully on 7c11de96060342a13f74d77c0355a9b5eb85384f<br /><br />When changing a generic css class you should check and see where else it is used. As can be seen on your screenshot the change has been applied to the name as well which is not desired. Even more obvious that this change is not correct on all other list views<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34945">issues.joomla.org/tracker/joomla-cms/34945</a>.</sub>

PR #34946, opened 2021-07-28
[4.0] Do not crash when no adapters ar available

laoneo:
### Summary of Changes
When a filesystem plugin is loaded before the local one which has no disks, the media manager crashes.

@dgrammatiko any comment?


dgrammatiko:
Maybe render something in the browser instead of throwing an error? Ppl will hardly get what’s wrong if there’s no visual signal. Good catch by the way

richard67:
@laoneo The javascript linter complains: https://ci.joomla.org/joomla/joomla-cms/46228/1/21

wilsonge:
Thanks!

Issue #34953, opened 2021-07-28
J4RC5 Inconsistency in property name and label


BrainforgeUK:
### What needs to be fixed
&lt;field
	name="siteDescription"
	type="text"
	label="TPL_CASSIOPEIA_TAGLINE"
/&gt;

The field name and label are inconsistent.
Not minor difference, as elsewhere, but totally different.

### Why this should be fixed

Easier reuse in child templates.

### How would you fix it

Change the field name to **siteTagline**
Also change the related CSS class **site-description** to **site-tagline**

Some may argue for changing the label instead - not my first choice.
I prefer the word 'Tagline' as displayed on the style configuration page.
If site developer / admin want to change the text they will use language override anyway.

### Side Effects expected

Any existing child templates - but as this is at RC stage the author should be anticipating such changes.

brianteeman:
Do you mean for example that
```
			name="loggable_verbs"
			label="COM_ACTIONLOGS_LOGGABLE_VERBS"
```

ie label should be "COMPONENT_FIELDNAME"

If so then then there has never been a standard for this and you will find a large number of labels that do not follow this _rule_

OR do you mean something else??

BrainforgeUK:
A standard / recommendation for this as you suggest could be a good direction to go in.
Makes it easier looking for every use of LOGGABLE_VERBS (ignoring case) in an IDE.

In this particular case I was thinking of changing the field definition to:

&lt;field
name="siteTagline"
type="text"
label="TPL_CASSIOPEIA_TAGLINE"
/&gt;

In line with your comment this would be even better!
But then where do you stop changing labels everywhere ...
Maybe just make a change when it crosses one's path, like this one.

&lt;field
name="siteTagline"
type="text"
label="TPL_CASSIOPEIA_SITETAGLINE"
/&gt;

brianteeman:
> Maybe just make a change when it crosses one's path, like this one.

That just makes the situation worse.

Changing any of this at this time (rc5) is never going to happen. Realistically the next time this could be changed would be joomla 5 (imho)

richard67:
I see not really an issue in sense of a bug in it. Shall we move it do discussions?

BrainforgeUK:
Worth keeping in mind and include in discussions about future changes to standards / recommendations.

In meantime created pull request #34955 

I've not changed the CSS classes or _header.css
Be better if site-description changed to site-tagline for consistency.
Will see if I can add that to pull request.

richard67:
[...]

PR #34957, opened 2021-07-28
[4.0] Changed siteDescription field to siteTagline

BrainforgeUK:
Pull Request for Issue #34953  .

### Summary of Changes

Changed field name **siteDescription** to **siteTagline**
Changed CSS c;lass **site-description** to **site-tagline**

### Testing Instructions

Changing tagline of template style in back-end style editor will change the tagline in front-end.

The tagname changes on the front-end.

The tagname changes on the front-end, as before.
CSS class **site-tagline** can be used to change the appearance of the tagline


brianteeman:
I still do not understand the reason behind this change, its not for consistency so why? Please enlighten me. What am I missing.

brianteeman:
Note that this PR is incomplete as there are sql changes that need to be made as well both for new installs and upgrades.

BrainforgeUK:
Consistency - change applies only within the Cassiopeia template.

SQL changes
(1) Cannot see any need for such on new installs.
(2) For production site upgrades would need and to change the field name in the params column of the template styles table. - as J4 is still at RC stage ...

brianteeman:
Consistency with what?

as a clean install sets a param with the existing fieldname then of course it needs to be changed

BrainforgeUK:
Reopened.

chmst:
@BrainforgeUK could you please resolve the conflicts?

HLeithner:

drmenzelit:
Sorry that it took so long to make a decision, but we can't accept this PR, it implicates too many conflicts for existent websites using Cassiopeia and /or child templates.

PR #34974, opened 2021-07-29
[4.0] Menu preset import. Add separator counter for reliable alias creation.

### LINKED ISSUE #34964: [4] Administrator menu. Import preset. "The alias 1627514907-04 is already being used"
- The issue is for sure not reproducable for everybody. As far as I see it has to do with the performance of the server environment.
- Current nightly. Very little installed and very few entries in the database.
- Go to `Menus > Manage`.
- Set filter on `Administrator` to create a backend menu.
- Click button `New`.
- Enter `Title` and `Alias`.
- Field `Import a preset`: Select `Preset Alternative Main Menu`.
- `Save & Close` or just `Save`

### Expected result
- All menu items from preset saved.

### Actual result
- I get a message like
```
Menu was saved but failed to import the preset: The alias 1627514907-04 is already being used by - menu item in the asdfgh menu (remember it may be a trashed item).
```
(`asdfgh` is the alias/menutype I entered in this case).

The image shows several tests that resulted in different numbers of menu items but the last one (red arrow) where I have manipulated tha alias creation in `MenusHelper.php` for `separator` types before.

I located the error here: https://github.com/joomla/joomla-cms/blob/4.0.0-rc5/administrator/components/com_menus/src/Helper/MenusHelper.php#L509

Looks like microtime(true) is not reliable enough in faster environments to create unique aliases.

- 10.5.11-MariaDB-1:10.5.11+maria~focal-log
- PHP 8.0.6 (Linux dd51328 5.4.0-80-generic #   90-Ubuntu SMP Fri Jul 9 22:49:44 UTC 2021 x86_64 )

### Additional comment
-It is normal that in each additional menu with preset altenate.xml 1 more menu entry is added, but only 1.

ReLater:
Fixes https://github.com/joomla/joomla-cms/issues/34964

### Summary of Changes
- Add a variable that count up `separator` menu items to create a guaranteed unique alias for each of them.
- I haven't established the `$separatorCount` variable in method `installPresetItems()` because it looks like presets are imported stepwise (several calls of this method that reset the counter to 0). Maybe overcautious?

### Testing Instructions
- See instructions ín issue on https://github.com/joomla/joomla-cms/issues/34964 please.

- See image in issue on https://github.com/joomla/joomla-cms/issues/34964 please.

- See image below with 4 successful tests that failed before.
- At least 87 menu items imported with preset alternate.xml.
- it is expected that each test run will add one more menu item entry (cross-reference to already existing menus).

### DIFF
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -54,7 +54,7 @@ class MenusHelper extends ContentHelper
 	 *
 	 * @var  integer
 	 *
-	 * @since   4.0.0
+	 * @since   __DEPLOY_VERSION__
 	 */
 	protected static $separatorCount = 0;
 
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -50,6 +50,15 @@ class MenusHelper extends ContentHelper
 	protected static $presets = null;
 
 	/**
+	 * Counts separator menu items during preset imports for unique alias creation.
+	 *
+	 * @var  integer
+	 *
+	 * @since   4.0.0
+	 */
+	protected static $separatorCount = 0;
+
+	/**
 	 * Gets a standard form of a link for lookups.
 	 *
 	 * @param   mixed  $request  A link string or array of request variables.
@@ -468,9 +477,10 @@ class MenusHelper extends ContentHelper
 	 */
 	protected static function installPresetItems($node, $menutype)
 	{
-		$db    = Factory::getDbo();
-		$query = $db->getQuery(true);
-		$items = $node->getChildren();
+		$db             = Factory::getDbo();
+		$query          = $db->getQuery(true);
+		$items          = $node->getChildren();
+		$separatorCount = 0;
 
 		static $components = array();
 
@@ -506,7 +516,7 @@ class MenusHelper extends ContentHelper
 			{
 				// Do not reuse a separator
 				$item->title = $item->title ?: '-';
-				$item->alias = microtime(true);
+				$item->alias = microtime(true) . '-' . self::$separatorCount++;
 			}
 			elseif ($item->type == 'heading' || $item->type == 'container')
 			{
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -477,10 +477,9 @@ class MenusHelper extends ContentHelper
 	 */
 	protected static function installPresetItems($node, $menutype)
 	{
-		$db             = Factory::getDbo();
-		$query          = $db->getQuery(true);
-		$items          = $node->getChildren();
-		$separatorCount = 0;
+		$db    = Factory::getDbo();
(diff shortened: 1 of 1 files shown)


administrator/components/com_menus/src/Helper/MenusHelper.php
```diff
@@ -49,6 +49,15 @@ class MenusHelper extends ContentHelper
 	 */
 	protected static $presets = null;
 
+	/**
+	 * Counts separator menu items during preset imports for unique alias creation.
+	 *
+	 * @var  integer
+	 *
+	 * @since   4.0.0
```
Quy:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

administrator/components/com_menus/src/Helper/MenusHelper.php
```diff
@@ -49,6 +49,15 @@ class MenusHelper extends ContentHelper
...
+	 */
+	protected static $separatorCount = 0;
```
joomdonation, line 59: Could you please explain why we need to use this static property of the class instead of a normal variable inside the method?

ReLater: See description of pr

richard67: @joomdonation See the description: "I haven't established the `$separatorCount` variable in method `installPresetItems()` because it looks like presets are imported stepwise (several calls of this method that reset the counter to 0). Maybe overcautious?" Makes sense to me. If it's ok for you, too, please mark the conversation as revolved.

joomdonation: @richard67 I haven't had time to look at the whole process yet to confirm, but please make it as revolved. If I have time, I will review it later.

alikon:
i'm unable to replicate

richard67:
Well, the instructions in the issue say that it might be hard to replicate.

ReLater:
<strike>Quote from the PHP manual:
` microtime(bool $as_float = false): string|float`

> If as_float is set to true, then microtime() returns a float, which represents the current time in seconds since the Unix epoch accurate to the nearest microsecond.

Emphasis on "time in seconds"! The value is rounded to seconds. This means that it is not incomprehensible if double values come out when a loop is run through faster if we use `microtime(true)`</strike>

When I test with a loop like this:

```
$withTrue = [];
$woTrue = [];

for ($i = 0; $i < 100000; $i++)
{
	$withTrue[] = microtime(true);
	//$woTrue[] = microtime();
}

echo '$withTrue: <pre>' . print_r($withTrue, true) . '</pre>';
// echo '$woTrue: <pre>' . print_r($woTrue, true) . '</pre>';

exit;
```

OR use a counted output: which outputs just two values that are never equal in my test cases:

```
echo 'count($withTrue): <pre>' . print_r(count($withTrue), true) . '</pre>';

$filter = array_unique($withTrue);

echo 'count($filter): <pre>' . print_r(count($filter), true) . '</pre>';
exit;
```

- $withTrue case: I need several thousand loops before I get the first different time value. With the last test I needed ~16.000 loops.
- $woTrue case: Also not reliable. Several blocks with several double time values. With just 100 loops => 75 double time values.

PR #34885, opened 2021-07-23
[4.0] Redo the image editing api

dgrammatiko:
Pull Request is the second part for  the Tabs reworked code: https://github.com/joomla/joomla-cms/pull/34813

### Summary of Changes
Some changes to the API:
- Now it's async (used to be synchronous, but that won't work for many cases, current libraries, etc)
- Fixing all the problems with:
- - Empty fields on load
- - Save needed a click away to update the image
- - other inconsistencies

#### This PR is not testable with the downladable version

Apply this PR!!!
You need to run `npm install`

Try to edit some images and using all the existing options

### Actual result BEFORE applying this Pull Request

Many bugs

### Expected result AFTER applying this Pull Request

No (obvious) bugs

### Documentation Changes Required

The old registration was something like:
```js
  // Register the Events
  Joomla.MediaManager.Edit.crop = {
    Activate(mediaData) {
      // Initialize
      initCrop(mediaData);
    },
    Deactivate() {
      if (!Joomla.MediaManager.Edit.crop.cropper) {
        return;
      }
      // Destroy the instance
      Joomla.MediaManager.Edit.crop.cropper.destroy();
    },
  };
```

Which is changed to:

```js
// Register the Events
window.addEventListener('media-manager-edit-init', () => {
  Joomla.MediaManager.Edit.plugins.crop = {
    Activate(image) {
      return new Promise((resolve /* , reject */) => {
        init(image);
        Joomla.MediaManager.Edit.plugins.crop.cropper
          .setAspectRatio(formElements.cropAspectRatioOption.value);
        resolve();
      });
    },
    Deactivate(image) {
      return new Promise((resolve /* , reject */) => {
        if (image.cropper) {
          image.cropper.destroy();
          delete Joomla.MediaManager.Edit.plugins.crop.cropper;
        }
        resolve();
      });
    },
  };
}, { once: true });

```

So most of the changes are just changes in the initialization step:
- there's an event for the registration, on the window, named 'media-manager-edit-init'
- The plugin name is now one level deeper, from `Joomla.MediaManager.Edit.crop` to `Joomla.MediaManager.Edit.plugins.crop`
- Both the `Activate` and `Deactivate` functions should return a promise.

@wilsonge sorry...

PR #34994, opened 2021-07-30
[4.0] Fix the Reset button

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/pull/34885#issuecomment-889971654

### Summary of Changes

- Fixes the reset button

### Testing Instructions

Try edit some image and hit the reset button

No

@laoneo

### DIFF
--- a/build/media_source/com_media/js/edit-images.es6.js
+++ b/build/media_source/com_media/js/edit-images.es6.js
@@ -137,6 +137,8 @@ class Edit {
     this.current.contents = `data:image/${this.fileType};base64,${this.options.contents}`;
     // Reactivate the first plugin
     this.imagePreview.setAttribute('src', this.current.contents);
+    this.imagePreview.setAttribute('width', this.imagePreview.naturalWidth);
+    this.imagePreview.setAttribute('height', this.imagePreview.naturalHeight);
     document.getElementById('myTab').activateTab(0, false);
     this.history = {};
   }
--- a/build/media_source/com_media/js/edit-images.es6.js
+++ b/build/media_source/com_media/js/edit-images.es6.js
@@ -26,6 +26,7 @@ class Edit {
       extension: this.extension,
       contents: `data:image/${this.fileType};base64,${this.options.contents}`,
     };
+    this.previousPluginDeactivated = new Promise((resolve) => resolve);
     this.history = {};
     this.current = this.original;
     this.plugins = {};
@@ -42,7 +43,6 @@ class Edit {
 
     // Once the DOM is ready, initialize everything
     customElements.whenDefined('joomla-tab').then(async () => {
-      const pluginsForInit = [];
       const tabContainer = document.getElementById('myTab');
       const tabsUlElement = tabContainer.firstElementChild;
       const links = [].slice.call(tabsUlElement.querySelectorAll('button[aria-controls]'));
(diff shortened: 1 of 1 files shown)


laoneo:
Reset event is triggered, but when I resize first and then do a reset it jumps to the first tab but does not revert the resize.

laoneo:
Also when I put the crop plugin on the second position and resize first and open the edit page the canvas from the crop plugin is shown initially.

dgrammatiko:
> Also when I put the crop plugin on the second position and resize first and open the edit page the canvas from the crop plugin is shown initially.

Yeah, the whole thing is supposed to be asynchronous now but I guess I have to redo the activation/deactivation parts more efficiently. I Will update this PR later

dgrammatiko:
> Reset event is triggered, but when I resize first and then do a reset it jumps to the first tab but does not revert the resize.

Hmm, weird are the width/height values correct after reset (eg the initial ones)? If so it's just the preview image...

laoneo:
Perhaps it has to do with the async, let you do the fix there and then we see if it still is an issue.

dgrammatiko:
@laoneo can you check if https://github.com/joomla/joomla-cms/pull/34994/commits/28da0e463f2e40b3aff65e79a442eeaf0b3c4efe solves the issues? Thanks

RickR2H:
@dgrammatiko I've tested this PR and the fix works. I noticed that after pressing reset, I get redirected to the first tab. Is this intended behavior?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34994">issues.joomla.org/tracker/joomla-cms/34994</a>.</sub>

laoneo:
It works better and the events are fired. I have the same behavior as @RickR2H that it jumps to the first tab. The form controls are also not reset (not sure if this should be part of this pr). In the following image I rotated and then hit reset, the image is like the original, but the form control angle not:

image

The problem I have is that my plugins depend on the form controls, so reset has basically no effect as on activate I do read them again.

Nice solution would be when the core would reset the form controls to the initial value, not so nice would be when each plugin has to do that by its own.

dgrammatiko:
> I get redirected to the first tab. Is this intended behavior?
>  I have the same behavior as @RickR2H that it jumps to the first tab.

This is intended behavior. There is an open PR that adds undo and redo, which will let you stay in the same tab

> The form controls are also not reset (not sure if this should be part of this pr)
> Nice solution would be when the core would reset the form controls to the initial value

@wilsonge and I had a quick chat on this and we decided to skip the resetting of the forms (it wasn't implemented also before all these changes). 
Unfortunately, the Edit class is just orchestrating things so having it to hold the state of the form elements would be a big stretch and it probably fail miserably in many cases (Web Components with shadow DOM, fields that are just an entry point for Vue, React, Svelte, etc). I think the easier solution is to have each plugin register its own initial state and have one more function in the API (eg Reset) that when called it will restore all the initial values to the respected form elements. Automation comes with limitations...
Anyways if George is ok I can add these here or in another PR, it's not much code anyways

richard67:
@laoneo @RickR2H With respect to @dgrammatiko 's answer and if everything else was ok, would you say your test was a successful test? If so, please mark your test result in the issue tracker. Thanks in advance.

laoneo:
I'm fine with a Reset function in the API. But switching to the first tab should definitely be fixed as it is cleearly a disturbing behavior. @richard67 till then for me no successful test.

richard67:
@dgrammatiko The js linter complains about missing semicolon: https://ci.joomla.org/joomla/joomla-cms/46304/1/21

```
/********/src/build/media_source/com_media/js/edit-images.es6.js
  145:9  error  Missing semicolon  semi
```

richard67:
I have tested this item :white_check_mark: successfully on 68c6485a4e9e9f9aec98fa2bb25b8f032bb2217b<br /><br />"Reset" button works fine for "rotate" and "resize", and active tab stays the same when using it.

For "crop" the "reset" button does nothing, but I think that's by design because there is nothing really cropped as long as the change is not saved.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34994">issues.joomla.org/tracker/joomla-cms/34994</a>.</sub>

richard67:
@RickR2H Could you test again? Thanks in advance.

@laoneo If you are available, could you spend the time for a test, too, in case if Rick isn't available.

RickR2H:
I have tested this item :white_check_mark: successfully on 68c6485a4e9e9f9aec98fa2bb25b8f032bb2217b<br /><br />Pressing reset button will reset the original preview of the image for rotation and scaling. 
Values in the text boxes are not updated as stated  before.
Pressing reset will leave you in the current tab.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34994">issues.joomla.org/tracker/joomla-cms/34994</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34994">issues.joomla.org/tracker/joomla-cms/34994</a>.</sub>

laoneo:
When I activate my media edit plugins I get now the following error on page load:
image

Not sure if it is related...?

richard67:
@laoneo Is your branch up to date? Maybe it is missing this fix: #35058 . Or if you use the complete branch of this PR maybe that branch is not up to date and so is missing the fix?

laoneo:
Yes it is up to date...

richard67:
@laoneo Does that happen also without the PR?

laoneo:
Without the patch, the error is a bit different 
image

laoneo:
@dgrammatiko are these errors related to the recent changes or something old?

laoneo:
@wilsonge I would merge this one and then we can open another issue for the other error.

laoneo:
Here we go https://github.com/joomla/joomla-cms/issues/35075

wilsonge:
Thanks!

Issue #34992, opened 2021-07-30
Joomla 4: plugin uninstallation fails if the plugin files have been deleted from the server


nikosdion:
### Steps to reproduce the issue

Install a plugin. Any plugin.

Delete its folder. For example, if you have a system plugin named `foo` delete the folder `plugins/system/foo`.

Go to System, Manage, Extensions and try to uninstall the plugin.

### Expected result

The dead plugin record is removed from `#__extensions` and I possibly get a warning. Like in Joomla 3.

### Actual result

Call to a member function attributes() on null

Call stack

`#` | Function | Location 
-- | -- | --
1 | () | JROOT/libraries/src/Installer/Adapter/PluginAdapter.php:424 |
2 | Joomla\CMS\Installer\Adapter\PluginAdapter->setupUninstall() | JROOT/libraries/src/Installer/InstallerAdapter.php:1188
3 | Joomla\CMS\Installer\InstallerAdapter->uninstall() | JROOT/libraries/src/Installer/Installer.php:784
4 | Joomla\CMS\Installer\Installer->uninstall() | JROOT/administrator/components/com_installer/src/Model/ManageModel.php:268
5 | Joomla\Component\Installer\Administrator\Model\ManageModel->remove() | JROOT/administrator/components/com_installer/src/Controller/ManageController.php:119
6 | Joomla\Component\Installer\Administrator\Controller\ManageController->remove() | JROOT/libraries/src/MVC/Controller/BaseController.php:730
7 | Joomla\CMS\MVC\Controller\BaseController->execute() | JROOT/libraries/src/Dispatcher/ComponentDispatcher.php:146
8 | Joomla\CMS\Dispatcher\ComponentDispatcher->dispatch() | JROOT/libraries/src/Component/ComponentHelper.php:389
9 | Joomla\CMS\Component\ComponentHelper::renderComponent() | JROOT/libraries/src/Application/AdministratorApplication.php:143
10 | Joomla\CMS\Application\AdministratorApplication->dispatch() | JROOT/libraries/src/Application/AdministratorApplication.php:186
11 | Joomla\CMS\Application\AdministratorApplication->doExecute() | JROOT/libraries/src/Application/CMSApplication.php:278
12 | Joomla\CMS\Application\CMSApplication->execute() | JROOT/administrator/includes/app.php:63
13 | require_once() | JROOT/administrator/index.php:32

### System information (as much as possible)

Joomla 4.0-RC5
PHP 7.4.20

### Additional comments

[...]

PR #34996, opened 2021-07-31
[4.0] Fix fatal error while uninstall deleted plugin

joomdonation:
Pull Request for Issue #34992.

### Summary of Changes
Currently, if we uninstall a plugin which was deleted for some reasons (or at least the manifest file was deleted), we will get fatal error because in the PluginAdater, we have code which didn't handle the case manifest not found.

This PR just adds some check to prevent these fatal error and allows these plugins still be uninstalled properly (mean the folder being deleted, the record is deleted from extensions table). 

I think we can do it better. For example, in **getElement**, we can just return `$this->extension->folder` (it's plugin group) and in the code to detect plugin group, we can use `$this->extension->folder`, too, instead of getting these information from manifest object. But this is just handle some special case, so I think it is fine.

### Testing Instructions
1. Try to delete a plugin folder (or delete the xml file of that plugin)
2. Uninstall that plugin
3. Before patch: you will get fatal error
4. After patch, the plugin will be uninstalled. Joomla also shows a warning to tell users that manifest not found.

### DIFF
--- a/libraries/src/Installer/Adapter/PluginAdapter.php
+++ b/libraries/src/Installer/Adapter/PluginAdapter.php
@@ -232,7 +232,7 @@ class PluginAdapter extends InstallerAdapter
 	 */
 	public function getElement($element = null)
 	{
-		if (!$element)
+		if (!$element && $this->getManifest())
 		{
 			// Backward Compatibility
 			// @todo Deprecate in future version
@@ -421,7 +421,10 @@ class PluginAdapter extends InstallerAdapter
 		$this->parent->findManifest();
 		$this->setManifest($this->parent->getManifest());
 
-		$this->group = (string) $this->getManifest()->attributes()->group;
+		if ($this->getManifest())
+		{
+			$this->group = (string) $this->getManifest()->attributes()->group;
+		}
 
 		// Attempt to load the language file; might have uninstall strings
 		$this->loadLanguage($this->parent->getPath('source'));


richard67 (approved):

richard67:
Grr, I had forgotten to select test result.

richard67:
I have tested this item :white_check_mark: successfully on 2e02c39a1fffa086a61166580cf5b825c4270f3f<br /><br />If complete folder is missing, a warning alert "Joomla\CMS\Filesystem\Folder::delete: Path is not a folder." is shown.

If only the manifest XML is missing, a warning alert "JInstaller: :Install: Can't find XML setup file." is shown.

In both cases the plugin has been properly uninstalled.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34996">issues.joomla.org/tracker/joomla-cms/34996</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 2e02c39a1fffa086a61166580cf5b825c4270f3f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34996">issues.joomla.org/tracker/joomla-cms/34996</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34996">issues.joomla.org/tracker/joomla-cms/34996</a>.</sub>

richard67:
@wilsonge I'd like to have that in 4.0 so I've set the 4.0 milestone.

wilsonge:
Fair enough. I'm down with that. Thankyou!

awslabspl (commented):
COM returned ALLOK

PR #34999, opened 2021-07-31
[4.0] Check target file extension of com_media rename action

SniperSister:
### Summary of Changes
Added a check in the rename action of com_media to make sure that the new extension does not violate the list of allowed extensions.

### Testing Instructions
1. Open browser dev tools
2. Open the rename dialog in com_media and hit save without changing the name
3. Inspect that last request in the network tab of the dev tools, edit it and change the file extension of the new_path parameter in the request body to a disallowed value, i.e. php
4. Verify that the file was renamed
5. Apply patch
6. Redo 1-3
7. Verify that the file could not be renamed anymore

No check of extension

Check of extension


brianteeman:
Why is it not using the list of filters here https://github.com/joomla/joomla-cms/pull/35001

Why are the two lists different eg php5, php6 etc

Surely there should be just one list

SniperSister:
I have added the extensions for php3-8, good catch! @brianteeman @richard67

richard67:
I have tested this item :white_check_mark: successfully on 16cf3077ab445d0bd5e97dbca962e45664db844c<br /><br />Tested as described in testing instructions.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34999">issues.joomla.org/tracker/joomla-cms/34999</a>.</sub>

brianteeman:
not sure where your list comes from but I would add
.MSI – A Microsoft installer file. 
.JAR – .JAR files contain executable Java code. If you have the Java runtime installed, .JAR files will be run as programs.

but there are a lot more https://www.lifewire.com/list-of-executable-file-extensions-2626061

HLeithner:
> not sure where your list comes from but I would add
> .MSI – A Microsoft installer file.
> .JAR – .JAR files contain executable Java code. If you have the Java runtime installed, .JAR files will be run as programs.
> 
> but there are a lot more https://www.lifewire.com/list-of-executable-file-extensions-2626061

Don't know ether but msi (not sure about jar) will not be execute by a webserver.

brianteeman:
> Don't know ether but msi (not sure about jar) will not be execute by a webserver.

Neither would many of the others (unless it is a windows webserver)

SniperSister:
@brianteeman @HLeithner added MSI and JAR

brianteeman:
I still dont understand why this list is different to the list in the other PR or why this list is even used when the other list could be. 

One list, less mistakes.

wilsonge:
@SniperSister maybe we should create the static list of executables in the framework input filter package and then they can be maintained as a single list that way as @brianteeman says rather than having our duplicate here in the MediaHelper class?

brianteeman:
@wilsonge thats what I keep saying but get ignored https://github.com/joomla/joomla-cms/pull/35001 is a partially identical list (and should be identical)

SniperSister:
@wilsonge it's not a duplicate, it's only partially identical with the list in the InputFilter focussing on file types that might be executable directly within the web server, resulting in code executions. The list in the mediahelper also includes file types that are able to execute code not only in the webserver but also in a general OS scenario or even in the browser

brianteeman:
still better to have just one to avoid mistakes or just have the list in mediahelper extend the inputfilter


/me doesnt see/agree with your supposed differences - especially for windows servers

SniperSister:
> or just have the list in mediahelper extend the inputfilter

That's fine for me

@wilsonge your call. Let me know how you want it solved.

wilsonge:
> or just have the list in mediahelper extend the inputfilter

This makes sense to me too. Let's go with it!

wilsonge:
Whilst your at it can you just expand the comment on the class property to mention your explanation so it's super clear for people in the future. Thankyou :)

wilsonge:
@SniperSister please can you look at this. It needs to be ready by tuesday for the final RC or it's not going in :)

SniperSister:
@wilsonge done!

richard67:
I have tested this item :white_check_mark: successfully on 4b6b32e98bd4637838fcafd9d742db08030a4981<br /><br />Checked with values from both arrays (input filter and media helper) that the manipulated request results in a 500 response and the file is not renamed.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/34999">issues.joomla.org/tracker/joomla-cms/34999</a>.</sub>

PR #35027, opened 2021-08-02
The order of panel widgets in the system section

korenevskiy:
image

Colleagues, by what principle are the link panels in the system section sorted?
I would prefer the SETUP and MANAGE panels to be the first.
I feel tension every time I open a panel and try to find the right panel somewhere in the middle of other panels. And if the screen is small, then you will have to search for the most popular panel by rewinding the page.
.
Let's place the SETUP and MANAGE panels first.

I marked with numbers how I would like to see the order. But, I may be inaccurate. The main thing is that I would like to see the SETUP and MANAGE the in  first position.

### DIFF
--- a/administrator/components/com_menus/presets/system.xml
+++ b/administrator/components/com_menus/presets/system.xml
@@ -19,99 +19,6 @@
 	</menuitem>
 
 	<menuitem
-		title="MOD_MENU_MAINTAIN"
-		type="heading"
-		icon="tools"
-		>
-		<menuitem
-			title="MOD_MENU_CLEAR_CACHE"
-			type="component"
-			element="com_cache"
-			link="index.php?option=com_cache"
-			permission="core.manage;com_cache"
-		/>
-
-		<menuitem
-			title="MOD_MENU_SYSTEM_INFORMATION_DATABASE"
-			type="component"
-			element="com_installer"
-			link="index.php?option=com_installer&amp;view=database"
-			permission="core.manage;com_installer"
-			ajax-badge="index.php?option=com_installer&amp;task=database.getMenuBadgeData&amp;format=json"
-		/>
-
-		<menuitem
-			title="MOD_MENU_GLOBAL_CHECKIN"
-			type="component"
-			element="com_checkin"
-			link="index.php?option=com_checkin"
-			permission="core.manage;com_checkin"
-			ajax-badge="index.php?option=com_checkin&amp;task=getMenuBadgeData&amp;format=json"
-		/>
-	</menuitem>
-
-	<menuitem
-		title="MOD_MENU_INFORMATION"
-		type="heading"
-		icon="info-circle"
-		>
-		<menuitem
-			title="MOD_MENU_INFORMATION_WARNINGS"
-			type="component"
-			element="com_installer"
-			link="index.php?option=com_installer&amp;view=warnings"
-			permission="core.manage;com_installer"
-			ajax-badge="index.php?option=com_installer&amp;task=getMenuBadgeData&amp;format=json"
-		/>
-
-		<menuitem
-			title="MOD_MENU_INFORMATION_POST_INSTALL_MESSAGES"
-			type="component"
-			element="com_postinstall"
-			link="index.php?option=com_postinstall"
-			permission="core.manage;com_postinstall"
-			ajax-badge="index.php?option=com_postinstall&amp;task=getMenuBadgeData&amp;format=json"
-		/>
-
-		<menuitem
-			title="MOD_MENU_SYSTEM_INFORMATION_SYSINFO"
-			type="component"
-			element="com_admin"
-			link="index.php?option=com_admin&amp;view=sysinfo"
-			permission="core.admin"
-		/>
-	</menuitem>
-
-	<menuitem
-		title="MOD_MENU_INSTALL"
-		type="heading"
-		icon="upload"
-		permission="core.manage;com_installer"
-		>
-		<menuitem
-			title="MOD_MENU_INSTALL_EXTENSIONS"
-			type="component"
-			element="com_installer"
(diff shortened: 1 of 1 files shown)


RickR2H:
I have tested this item :white_check_mark: successfully on 47ec0e35b4cff050d3552b98c0c26b6ef39feb98<br /><br />Test is successful. The thing I don't know if it is a good idea to change the order due to personal preference. but that's not up to me to decide.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35027">issues.joomla.org/tracker/joomla-cms/35027</a>.</sub>

korenevskiy:
> I have tested this item ✅ successfully on [47ec0e3](https://github.com/joomla/joomla-cms/commit/47ec0e35b4cff050d3552b98c0c26b6ef39feb98)
> 
> Test is successful. The thing I don't know if it is a good idea to change the order due to personal preference. but that's not up to me to decide.
> 
> This comment was created with the [J!Tracker Application](https://github.com/joomla/jissues) at [issues.joomla.org/tracker/joomla-cms/35027](https://issues.joomla.org/tracker/joomla-cms/35027).

I started using J4 in combat mode. Creating a website for sales. For the necessary cases, I had to go very, very many times to the CMS system section. I will say that it is very inconvenient when the MANAGE block jumps and then is located somewhere in the middle of the screen.
Which order would you personally prefer?

RickR2H:
I agree that the block jumping is annoying. Clicked wrong a couple of times myself, so I get your point. I don't have a personal preference regarding the order, that's up to the maintainers to decide. Maybe it is a better option to create a solution which prevent the jumping in total.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35027">issues.joomla.org/tracker/joomla-cms/35027</a>.</sub>

chmst:
Every project has other requirements , especially during  development  The best solution for you, @korenevskiy woluld be an own backend menu so you can have your favourite menu items in the menu bar itself.

RickR2H:
@dgrammatiko The masonry grid is build with Javascript which is loaded after the page load. That's why the jumping occurs. Is there a way to prevent the jumping? In the Atum template I see some references to probably a prior/fallback solution for the layout using CSS column-count. Why use JS as we can implement a CSS option? Hope you can help...

screen shot 2021-08-03 at 10 21 04<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35027">issues.joomla.org/tracker/joomla-cms/35027</a>.</sub>

richard67:
> > I have tested this item ✅ successfully on [47ec0e3](https://github.com/joomla/joomla-cms/commit/47ec0e35b4cff050d3552b98c0c26b6ef39feb98)

We ~~assume~~ require the authors of pull requests to provide only code they have tested. But the authors shall not mark tests for their own pull requests in the issue tracker, because these tests will not count. A PR needs 2 human tests by other people than the author.

dgrammatiko:
> Why use JS as we can implement a CSS option?

Not my decision, if you go back in the PR implementing it you'll see my comments against it. Also, there's no way to fix the jumping unless you add display none till the JS evaluates (which is a terrible way to fix it and shouldn't be done)

RickR2H:
@dgrammatiko As long as we don'n have a clever solution, we are stuck with the jumping. Annoying, but not the biggest problem. @ricardo1709 please test so we have two human tests.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35027">issues.joomla.org/tracker/joomla-cms/35027</a>.</sub>

richard67:
> > > I have tested this item ✅ successfully on [47ec0e3](https://github.com/joomla/joomla-cms/commit/47ec0e35b4cff050d3552b98c0c26b6ef39feb98)
> 
> We ~assume~ require the authors of pull requests to provide only code they have tested. But the authors shall not mark tests for their own pull requests in the issue tracker, because these tests will not count. A PR needs 2 human tests by other people than the author.

@korenevskiy Sorry, my mistake. I haven't noticed it was a citation of @RickR2H 's test. To me it looked like it was your test at a first view.

rdeutz:
This is for sure a UX issue but it will not get solved by resorting. Someone else might want another order and at the end of the day is is a question of personal preference. We decided to let it as it is.

@korenevskiy maybe start a discussion to find a clever way that suits all needs.

richard67:
Discussion exists already here https://github.com/joomla/joomla-cms/discussions/35019 .

korenevskiy:
> This is for sure a UX issue but it will not get solved by resorting. Someone else might want another order and at the end of the day is is a question of personal preference. We decided to let it as it is.
> 
> @korenevskiy maybe start a discussion to find a clever way that suits all needs.

Of course, there will always be someone who wants a different order.
I propose to vote.

brianteeman:
No need to vote  thats not how we do things - things will be even slower to develop

korenevskiy:
> No need to vote thats not how we do things - things will be even slower to develop

The Masonry layout implies automatic placement depending on the height of the block. So that the lower border of the parent's space is aligned with the content.
Thus, the order of the blocks was imposed on us by the JS Masonry script. At your discretion. But in the cards there were blocks in which the number of elements is the same. Even they can't be swapped at our discretion. The JS script has its own logic.
I turned off Masonry. Sorted the blocks as the creator/maintainer of the sites.
As a result, it was possible to remove the block jumps when the page was loading.
Please test the PR

https://github.com/joomla/joomla-cms/pull/35040

Issue #35010, opened 2021-08-01
J4 RC5 a11y - increae color contrast for drop down


coolcat-creations:
### Steps to reproduce the issue
<img width="745" alt="grafik" src="https://user-images.githubusercontent.com/828371/127768018-cc30d2f8-4074-4f7a-be9a-20039e4e6e96.png">
<img width="285" alt="grafik" src="https://user-images.githubusercontent.com/828371/127768027-34f22511-9ffe-4fa1-aeb5-34df3adcd0d9.png">

The color contrast in the green drop downs is not enough

### Expected result
White and a darker green would work
<img width="756" alt="grafik" src="https://user-images.githubusercontent.com/828371/127768070-9a34fc5e-d673-40ad-9673-40e1a0aace2c.png">

### Actual result
Contrast Checker fails

### System information (as much as possible)
Joomla 4 RC5 and Firefox Developer Edition

brianteeman:
I'm a bit confused - which green are you referring to. If it is the green box with the black indicator then you cut the graphic before it shows the relevant part ;)

![image](https://user-images.githubusercontent.com/1296369/127768338-943aa25a-199b-49eb-b8ac-15bafcdfd78c.png)

If its the green text  on the white background then it is also fine

![image](https://user-images.githubusercontent.com/1296369/127768483-6af97475-1da0-4fda-8b55-62b5a219f1b6.png)

coolcat-creations:
I tested the black chevron on the green and It shows both fail in the color contrast checker?

chmst:
The black chevron on dark green indeed has bad contrast.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35010">issues.joomla.org/tracker/joomla-cms/35010</a>.</sub>

brianteeman:
It is not text. It's a user interface component which as shown in the screenshot passes because the contrast only needs to be 3:1

coolcat-creations:
Ok, got you - however the contrast is so weak that it confused even me and I am not color blind. this could be optimized or not?

brianteeman:
The problem is that because of the _strange_ way the arrow has been created that has to stay as black and changing the green is the only option but that has a knock on effect throughout the template.

For me the only real solution is to fix the way the arrow is created

coolcat-creations:
Seems we could change the css to 

[...]

PR #35052, opened 2021-08-05
White Color contrast for Dropdown

Krshivam25:
Pull Request for Issue #35010 

color2

color

No

### DIFF
--- a/administrator/components/com_banners/forms/banner.xml
+++ b/administrator/components/com_banners/forms/banner.xml
@@ -43,7 +43,6 @@
 			name="state"
 			type="list"
 			label="JSTATUS"
-			class="form-select-color-state"
 			size="1"
 			default="1"
 			validate="options"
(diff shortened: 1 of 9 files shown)


brianteeman:
There are 440 instances of form-select-color-state

Krshivam25:
> There are 440 instances of form-select-color-state

Yes, correct! I will check and fix this again.

brianteeman:
Hint this is not a suitable solution

dgrammatiko:
Use the system colours Canvas and/or CanvasText, should be the solution here. Info: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#system_colors

coolcat-creations:
Hi @Krshivam25 thank you - actually it would be good if the green remains but the arrow would be white.

brianteeman:
The challenge is that you cannot just change the arrow to white for the dropdown. The dropdown can have multiple states each with their own colour so if you change it to white it will be almost invisible in some states.

image
image
image

brianteeman:
this should be closed as it is completely the wrong approach

PR #34813, opened 2021-07-17
[4.0] Redo of custom elements tabs

dgrammatiko:
Pull Request for many Issues originating from the tabs script

fixes:
https://github.com/joomla/joomla-cms/issues/33229
https://github.com/joomla/joomla-cms/issues/34492
https://github.com/joomla/joomla-cms/issues/31583

### Summary of Changes
- The logic was rewritten from scratch
- Same funtionality
- Better code
- Support for dynamically added tabs
- A11y fixes (based on the WAI-ARIA recommendations)

Changes of the source js/css can be found: https://github.com/joomla-projects/custom-elements/pull/180

### Testing Instructions
Test each and every form in the backend that has tabs

### Expected result AFTER applying this Pull Request

Tabs View:
Screenshot 2021-07-17 at 14 06 02

Screenshot 2021-07-17 at 14 06 14

Shrink to accordion:
Screenshot 2021-07-17 at 14 07 03

Screenshot 2021-07-17 at 14 07 03

### Documentation Changes Required

- Added the recall and breakpoint attributes to the expected values for the creation of the tabs (UiTabs::)
- The `section` element (the tab content) is renamed to `joomla-tab-element`

PR #35058, opened 2021-08-06
[4.0] Fix missing npm update of custom elements

richard67:
Pull Request for Issue # .

### Summary of Changes

In package.json we currently have `"joomla-ui-custom-elements": "0.2.0",` but in package-lock.json we have `"joomla-ui-custom-elements": "0.1.0 ",`:

https://github.com/joomla/joomla-cms/blob/4.0-dev/package.json#L48

https://github.com/joomla/joomla-cms/blob/4.0-dev/package-lock.json#L27

So it seems that the change from PR #34813 later got lost, possibly when merging PR #34888 .

This means if you work on the current 4.0-dev branch you still have version 0.1.0 of custom elements after an `npm ci`, but after an `npm install` you would have version 0.2.0.

This pull request (PR) here fixes that.

One way how this could have happened is that the latter PR had a conflict for the package-lock.json when it was merged. Or something else happened later.

=> Ping @wilsonge .

### Testing Instructions

Code review.

In package-lock.json we have `"joomla-ui-custom-elements": "0.1.0 ",`, and so nightly build packages contain that version of Custom Elements, and the next RC will contain it, too, if this PR here will not be merged before, and so the changes from PR #34813 might not work.

Custom Elements version 0.2.0 being used.

None.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "diff": "^4.0.2",
         "dragula": "3.7.2",
         "focus-visible": "^5.2.0",
-        "joomla-ui-custom-elements": "0.1.0 ",
+        "joomla-ui-custom-elements": "0.2.0",
         "jquery": "^3.6.0",
         "jquery-migrate": "^3.3.2",
         "mark.js": "^8.11.1",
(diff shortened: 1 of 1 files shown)


richard67:
As PR #34813 was a release blocker and this PR restores the partly lost change from that PR, this here is a release blocker in my opinion.

richard67:
Thanks!

PR #35066, opened 2021-08-07
[3.x] [WiP] Delete folders on update only if they are empty

richard67:
Pull Request for Issue #34298 .

### Summary of Changes

_Work in Progress (WiP)_

Change function "deleteUnexistingFiles" in the CMS installer script "administrator/components/com_admin/script.php" so that folders are only attempted to be deleted if they are empty when updating the CMS.

### Testing Instructions

_Will be added soon._

_Will be added soon._

_Will be added soon._

None.

### DIFF
--- a/libraries/src/Filesystem/Folder.php
+++ b/libraries/src/Filesystem/Folder.php
@@ -728,4 +728,48 @@ abstract class Folder
 
 		return preg_replace($regex, '', $path);
 	}
+
+	/**
+	 * Checks if a folder is empty.
+	 *
+	 * @param   string  $path  The path of the folder to check.
+	 *
+	 * @return  boolean  True if the folder is a valid empty folder.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function isEmpty($path)
+	{
+		// Check to make sure the path valid and clean
+		$pathObject = new PathWrapper;
+		$path = $pathObject->clean($path);
+
+		// Is the path a folder?
+		if (!is_dir($path))
+		{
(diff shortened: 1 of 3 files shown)


libraries/src/Filesystem/Folder.php
```diff
@@ -728,4 +728,48 @@ public static function makeSafe($path)
...
+	 *
+	 * @param   string  $path  The path of the folder to check.
+	 *
+	 * @return  boolean  True if the folder is a valid empty folder.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function isEmpty($path)
+	{
+		// Check to make sure the path valid and clean
```
brianteeman, line 743:
```suggestion
		// Check to make sure the path is valid and clean
```

libraries/src/Filesystem/Folder.php
```diff
@@ -728,4 +728,48 @@ public static function makeSafe($path)
...
+
+			return false;
+		}
+
+		if (!($handle = @opendir($path)))
+		{
+			return false;
+		}
+
+		// Return false as soon as the first file or subfolder found
```
brianteeman, line 760:
```suggestion
		// Return false as soon as the first file or subfolder is found
```

richard67:
Sorry for having wasted your time.

This PR is incomplete or even wrong as it is, and I don't even think that the files and folder deletion should be changed in CMS version 3, because in opposite to version 4 it **_might_** be (I haven't checked that) that the list of files and folders to be deleted contains only a parent folder but not subfolders and files in some case.

Let's continue discussion in the issue.

PR #35067, opened 2021-08-07
[4.0] [WiP] Delete folders on update only if they are empty

richard67:
Pull Request for Issue #34298 .

### Summary of Changes

_Work in Progress (WiP)_

Change function "deleteUnexistingFiles" in the CMS installer script "administrator/components/com_admin/script.php" so that folders are only attempted to be deleted if they are empty when updating the CMS.

### Testing Instructions

_Will be added soon._

_Will be added soon._

_Will be added soon._

None.

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -7316,13 +7316,15 @@ class JoomlaInstallerScript
 
 		foreach ($files as $file)
 		{
-			if ($fileExists = File::exists(JPATH_ROOT . $file))
+			$filePath = JPATH_ROOT . $file;
+
+			if ($fileExists = File::exists($filePath))
 			{
 				$status['files_exist'][] = $file;
 
 				if ($dryRun === false)
 				{
-					if (File::delete(JPATH_ROOT . $file))
+					if (File::delete($filePath))
 					{
 						$status['files_deleted'][] = $file;
 					}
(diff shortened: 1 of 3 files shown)


administrator/components/com_admin/script.php
```diff
@@ -7336,13 +7338,19 @@ public function deleteUnexistingFiles($dryRun = false, $suppressOutput = false)
...
 		{
-			if ($folderExists = Folder::exists(JPATH_ROOT . $folder))
+			$folderPath = JPATH_ROOT . $folder;
+
+			if ($folderExists = Folder::exists($folderPath))
 			{
 				$status['folders_exist'][] = $folder;
 
-				if ($dryRun === false)
+				if (!(Folder::isEmpty($folderPath)))
```
```suggestion
				if (!Folder::isEmpty($folderPath))
```
Remove obsolete brackets

richard67:
I close this because it contains some mistakes and is not complete yet, and to me it is questionable if we really should go that way.

Just imagine someone browses the Joomla root and then subfolders on Windows with the filesystem explorer or other tools. For every folder which contains an image, a thumbnail file might be created depending on the tool and its settings.

There might be other scenarios where files are created by tools, e.g. if someone uses tools for remote folder synchronization which create hidden files.

These files would block deletion of their parent folders, and these then for their parent folder and so on.

I don't really think that people will check their update log and notice that.

PR #35084, opened 2021-08-10
[4.0] composer update

brianteeman:
Dont really understand why but when I run composer update (after the latest commits by @wilsonge) i get a new and different composer.lock file

From what I can see the difference is that my version includes more funding and support links

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -56,6 +56,10 @@
                 "idna",
                 "php"
             ],
+            "support": {
+                "issues": "https://github.com/algo26-matthias/idna-convert/issues",
+                "source": "https://github.com/algo26-matthias/idna-convert/tree/v3.0.5"
+            },
             "time": "2020-10-05T05:49:30+00:00"
         },
         {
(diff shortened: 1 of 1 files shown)


richard67:
I can see only differences in support and funding except of one URL change from `"url": "https://github.com/joomla-framework/mediawiki-api.git"`to `"url": "git@github.com:joomla-framework/mediawiki-api.git",` which also has a related change a few lines below (`"default-branch": true,`added), and for one other package the "abandoned" property has been removed. But there is not really an update between current 4.0-dev and this PR here. So I think it doesn't really make a difference.

richard67:
But I can confirm I get the same changes as made by this PR here when running "composer update" on a clean, current 4.0-dev branch as of now. Am working on Linux.

richard67:
I have tested this item :white_check_mark: successfully on 1f84b24942243684c3495170142d4fa841b877d7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35084">issues.joomla.org/tracker/joomla-cms/35084</a>.</sub>

richard67:
Maybe @wilsonge uses an older or newer composer version than we do?

brianteeman:
I am using the very latest release
Composer - Dependency Manager for PHP - version 2.1.5

richard67:
2.1.3 here.

wilsonge:
I'm on 2.1.3 too. So I dunno especially if @richard67 is on the same version of me but gets the same as you brian...

Either way looks fine, merging

brianteeman:
cool

PR #35095, opened 2021-08-12
[3.10] - make com_search work on J4

alikon:
### Summary of Changes
make com_search work on J4

### Testing Instructions
apply pr
upgrade to j4

com_search doesn't work

com_search  works in j4

?


brianteeman:
I thought com_search was removed ?

alikon:
it is removed from new installation , but if you upgrade from a j3 site .....

components/com_search/views/search/view.html.php
```diff
@@ -94,7 +94,7 @@ public function display($tpl = null)
 		$lists['searchphrase'] = JHtml::_('select.radiolist', $searchphrases, 'searchphrase', '', 'value', 'text', $state->get('match'));
 
 		// Log the search
-		\Joomla\CMS\Helper\SearchHelper::logSearch($searchWord, 'com_search');
+		SearchHelper::oldlogSearch($searchWord, 'com_search');
```
wilsonge, line 97: This looks like a bad change - your not importing the namespace either. so this will be a invalid class.

wilsonge:
The intention was that people should update to the decoupled com_search here https://github.com/joomla-extensions/search/tree/4.0-dev similar to how we did weblinks.

zero-24:
Yes the only thing we would to do in an ideal world would be to add the update server into 3.10 and make sure it is detected as 3rd party extension at the time of the upgrade checks.

zero-24:
or even better add the update xml into 4.0 as the upgrade works and the new version will be 4.x only anyway.

That would mean 3.10 -> 4.0 -> com_search update like all other 3rd party extensions that are no longer 3.x compatible would go,

components/com_search/views/search/view.html.php
alikon: probably you forgot this remove
https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_admin/script.php#L3857

alikon:
plus there is no information  about com_search nor com_search version j4

Screenshot from 2021-08-12 10-22-18

richard67:
> The intention was that people should update to the decoupled com_search here https://github.com/joomla-extensions/search/tree/4.0-dev similar to how we did weblinks.

@wilsonge We did weblinks? You mean it is ready for J4? If so then I missed something. To me the weblinks repository seems to be abandoned. No PR gets merged there, CI checks broken since months.

zero-24:
Hmm when the code provided here makes com_search work I would say the correct way forward would be to do a PR against: https://github.com/joomla-extensions/search/tree/4.0-dev/src

Once merged there do a release and add the new update server on updating from 3.10.

So after updating to 4.0 the update shows up and the package can be used with the package that works with 4.0

alikon:
> Hmm when the code provided here makes com_search work

that's easy, simply test it

> the correct way forward would be to do a PR against: https://github.com/joomla-extensions/search/tree/4.0-dev/src

there is no need to do this pr there, it's a complete new rewrite in j4 style already

this pr just give us the time to do:

- do a release 
- add the new update server 
- etc

things that hardly can be done until 17

zero-24:
Hi, this PR is the base for the [minimal version of search running on 4.0](https://github.com/joomla-extensions/search/tree/4.0.0-minimal) that we have called and published as 4.0.0-dev here: https://github.com/joomla-extensions/search/releases/tag/4.0.0-dev to be used until a more final package of search has been build.

Thanks @alikon for providing this PR 👍

PR #35097, opened 2021-08-12
Update installation language date on version bump

HLeithner:
Installation language didn't get updated on version bump


zero-24:
Merged

PR #35102, opened 2021-08-12
[4.0] Correct assets name for workflow state -> stage and ID

sanderpotjer:
Correction assets installation data. Asset `name` for sample stage is currently incorrect (`com_content.state.1` instead of `com_content.stage.1`) and default `asset_id` for `#__workflow_stages` should not be related to 0 but the correct asset.

### DIFF
--- a/administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
+++ b/administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
@@ -4,5 +4,5 @@ UPDATE `#__assets`
  WHERE `name` = 'com_content.state.1';
 
 UPDATE `#__workflow_stages`
-   SET `asset_id` = 57
+   SET `asset_id` = (SELECT MAX(`id`) FROM `#__assets` WHERE `name` = 'com_content.stage.1')
  WHERE `asset_id` = 0;
--- a/administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
+++ b/administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
@@ -1,0 +1,8 @@
+-- after 4.0.0 RC6
+UPDATE `#__assets`
+   SET `name` = 'com_content.stage.1'
+ WHERE `name` = 'com_content.state.1';
+
+UPDATE `#__workflow_stages`
+   SET `asset_id` = 57
+ WHERE `asset_id` = 0;
--- a/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
+++ b/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
@@ -1,0 +1,8 @@
+-- after 4.0.0 RC6
+UPDATE "#__assets"
+SET "name" = 'com_content.stage.1'
+WHERE "name" = 'com_content.state.1';
+
+UPDATE "#__workflow_stages"
+SET "asset_id" = 57
+WHERE "asset_id" = 0;
--- a/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
+++ b/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
@@ -4,5 +4,5 @@ SET "name" = 'com_content.stage.1'
 WHERE "name" = 'com_content.state.1';
 
 UPDATE "#__workflow_stages"
-SET "asset_id" = 57
+SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
 WHERE "asset_id" = 0;
--- a/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
+++ b/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
@@ -1,7 +1,7 @@
 -- after 4.0.0 RC6
 UPDATE "#__assets"
-SET "name" = 'com_content.stage.1'
-WHERE "name" = 'com_content.state.1';
+   SET "name" = 'com_content.stage.1'
+ WHERE "name" = 'com_content.state.1';
 
 UPDATE "#__workflow_stages"
    SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
--- a/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
+++ b/administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
@@ -4,5 +4,5 @@ SET "name" = 'com_content.stage.1'
 WHERE "name" = 'com_content.state.1';
 
 UPDATE "#__workflow_stages"
-SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
-WHERE "asset_id" = 0;
+   SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
+ WHERE "asset_id" = 0;
--- a/installation/sql/mysql/base.sql
+++ b/installation/sql/mysql/base.sql
@@ -76,7 +76,7 @@ INSERT INTO `#__assets` (`id`, `parent_id`, `lft`, `rgt`, `level`, `name`, `titl
 (54, 16, 52, 53, 2, 'com_menus.menu.1', 'Main Menu', '{}'),
 (55, 18, 94, 95, 2, 'com_modules.module.87', 'Sample Data', '{}'),
 (56, 8, 20, 37, 2, 'com_content.workflow.1', 'COM_WORKFLOW_BASIC_WORKFLOW', '{}'),
-(57, 56, 21, 22, 3, 'com_content.state.1', 'COM_WORKFLOW_BASIC_STAGE', '{}'),
+(57, 56, 21, 22, 3, 'com_content.stage.1', 'COM_WORKFLOW_BASIC_STAGE', '{}'),
 (58, 56, 23, 24, 3, 'com_content.transition.1', 'Publish', '{}'),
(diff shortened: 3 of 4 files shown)


brianteeman:
This will also need update sql statements

sanderpotjer:
@brianteeman you're welcome. 

Update SQL statements to correct installation data which did not released stable version yet? But statements added anyway.

richard67:
> Update SQL statements to correct installation data which did not released stable version yet? But statements added anyway.

@sanderpotjer Since we are in beta phase we grant updates working between the pre-releases (beta, rc and then to stable) and this requires the right update SQL scripts.

sanderpotjer:
@richard67 thanks for the explanation 👍

administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
```diff
@@ -0,0 +1,8 @@
+-- after 4.0.0 RC6
+UPDATE `#__assets`
+   SET `name` = 'com_content.stage.1'
+ WHERE `name` = 'com_content.state.1';
+
+UPDATE `#__workflow_stages`
+   SET `asset_id` = 57
```
richard67: We cannot use a hard-coded asset ID in update SQL scripts like we can do in the SQL scripts for new installations, because the assets might be different on an updated site.

Therefore you have to get the right asset ID in a subquery by the asset's name, and in case someone changed permissions for workflow stages there might be even 2 assets with name "com_content.stage.1", one coming from the mistake corrected by this PR here with the above update, and another one created by the backend when editing the permissions. So the subquery here should include a "MIN" or "MAX" for the "asset_id" column to return only one value. I suggest "MAX" because that would be the newest.

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
```diff
@@ -0,0 +1,8 @@
+-- after 4.0.0 RC6
+UPDATE "#__assets"
+SET "name" = 'com_content.stage.1'
+WHERE "name" = 'com_content.state.1';
+
+UPDATE "#__workflow_stages"
+SET "asset_id" = 57
```
richard67:
```suggestion
SET "asset_id" = 57
```
See my review comment for the same file for MySQL above: We cannot use hard-coded asset IDs on updated sites.

administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
richard67: So it would be:

```
UPDATE `#__workflow_stages`
   SET `asset_id` = (SELECT MAX(`id`) FROM `#__assets` WHERE `name` = 'com_content.stage.1')
 WHERE `asset_id` = 0;
```

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
richard67: So I suggest:

```
UPDATE "#__workflow_stages"
   SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
 WHERE "asset_id" = 0;
```

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
```diff
@@ -4,5 +4,5 @@ SET "name" = 'com_content.stage.1'
 WHERE "name" = 'com_content.state.1';
 
 UPDATE "#__workflow_stages"
-SET "asset_id" = 57
+SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
```
richard67, line 7:
```suggestion
   SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
```

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
```diff
@@ -0,0 +1,8 @@
+-- after 4.0.0 RC6
+UPDATE "#__assets"
+SET "name" = 'com_content.stage.1'
+WHERE "name" = 'com_content.state.1';
+
+UPDATE "#__workflow_stages"
+SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
+WHERE "asset_id" = 0;
```
richard67:
```suggestion
   SET "asset_id" = (SELECT MAX("id") FROM "#__assets" WHERE "name" = 'com_content.stage.1')
 WHERE "asset_id" = 0;
```

sanderpotjer:
@richard67 thanks for the feedback. Made the requested changes. 

Side note: I suggest to reconsider to support updates between non-stable versions, and recommend people to use a fresh install for stable versions instead 😉

brianteeman:
> Side note: I suggest to reconsider to support updates between non-stable versions, and recommend people to use a fresh install for stable versions instead

Since we made the change it has made a massive difference, for the better, with people testing pull requests. Its slightly more work for the person creating the PR but its been worth it.

chmst:
Transitions an their records in assets are not correct  (fresh installation)

Transitions
grafik

Assets
grafik

administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
```diff
@@ -0,0 +1,8 @@
+-- after 4.0.0 RC6
+UPDATE `#__assets`
+   SET `name` = 'com_content.stage.1'
+ WHERE `name` = 'com_content.state.1';
+
+UPDATE `#__workflow_stages`
+   SET `asset_id` = (SELECT MAX(`id`) FROM `#__assets` WHERE `name` = 'com_content.stage.1')
```
richard67, line 7: I just came to this again and think I've missed something with my previous review:
1. If someone has deleted the asset, the subquery will not return any record, and that will result in an SQL error here. It needs to change the subquery to a join so we don't need the same subquery again in the where clause for checking if there is some result or not.
2. The where clause should be extended so it does not catch all stages with an asset ID of zero but only com_content's basic stage.

richard67:
@sanderpotjer The findings mentioned by @chmst above are not caused by your PR. She knows that. But it could make sense to fix it with your PR, too.

The update SQL could become a bit complicated for that. I'll try to help with that as soon as I know if you intend to fix these findings here, too, or if not. Or if you are too busy or it becomes too complicated for you, I can take over and make a new PR for all. Just let me know what you prefer.

administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
richard67: Another issue with this UPDATE statement: If someone just opens and saves the workflow stage in backend on a current 4.0-dev without this PR applied, a new asset with the right name is created. So we should do this update only if there is no asset with the right name yet, otherwise we will have two of them.

administrator/components/com_admin/sql/updates/mysql/4.0.0-2021-08-13.sql
```diff
@@ -0,0 +1,8 @@
+-- after 4.0.0 RC6
```
richard67, line 1:
```suggestion
```
We don't need these comments here.

Previous update SQL scripts have such because they were combined from several previously present update SQL scripts when we had combined them together in order to reduce their number. But here this is not the case.

administrator/components/com_admin/sql/updates/postgresql/4.0.0-2021-08-13.sql
```diff
@@ -0,0 +1,8 @@
+-- after 4.0.0 RC6
```
richard67, line 1:
```suggestion
```
We don't need these comments here.

Previous update SQL scripts have such because they were combined from several previously present update SQL scripts when we had combined them together in order to reduce their number. But here this is not the case.

richard67:
@sanderpotjer Because it's better to explain with code changes than words what all would be necessary to fix the things mentioned in my review comments and in @chmst 's findings, I've created the draft PR #35113 which would replace this one here. Please let me know soon what you prefer: Shall I make that PR against your branch of this PR, so you will continue with your PR and update testing instructions? Or shall I take over, i.e. finish my draft PR and you close this one in favour of mine? Thanks in advance for your reply.

richard67:
@sanderpotjer Because I haven't got any feedback from you, I've decided to finish my PR #35113 , which replaces this one here and fixes the other issue mentioned in @chmst 's comment. I hope you are not angry. I'd be happy if you could help with testing. Thanks in advance, and thanks for reporting the issue and making this PR which was a good start.

richard67:
Closing in favour of #35113 . Please test. Thanks in advance.

sanderpotjer:
@richard67 I'm not angry. But if I am disappointed? Yes. A bit more time to respond on the weekends would be nice, or at least fork the PR I was working to remain my initial commits (and in that way credits). 

I respect and appreciate your and everyone else's work on Joomla, but this pull request does not motivate me to contribute further.

brianteeman:
@sanderpotjer I am sure that @richard67 acted in all good faith and that you appreciate that with just 48 hours until release some of the normal timelines have to be shortened. Of course it would have helped if you had been testing your extension (and thus using this one contribution as a marketing tool) much earlier and not at the last possible moment. This bug has been present for a very long time.

sanderpotjer:
@brianteeman dear co-founder of Joomla, what a sad comment and preconception.

richard67:
> ... or at least fork the PR I was working to remain my initial commits (and in that way credits).

@sanderpotjer You are right, I should have done it that way. I only can say sorry for my mistake. I'll see if I can fix that.

sanderpotjer:
@richard67 Please don't spend more of your time to fix that, for similar situations in the future it would be nice 😉

richard67:
@sanderpotjer Ok, thanks. It was like Brian said, I was driven by the tight release time schedule and the experience that it is hard and takes some time to get testers for such things and the assumption that you might not be available because of weekend.

PR #35112, opened 2021-08-14
[4.0] Make edit_lock layout reusable for other extensions

joomdonation:
Pull Request for Issue # .

### Summary of Changes
The layout **layouts/joomla/content/icons/edit_lock.php** is used to display lock icon (when the item is being checked out by different user and could not be edited by the current logged in user)

For some reasons, in the code, we hardcode it to **article** and **contact** only, thus prevent other extensions (found this when checking weblinks layout) from using it.

This PR extends the code abit to make layout reusable by other extensions (weblinks in the case) by passing `ariaDescribed`  to display data and use it if provided. That removes the limit for **article** and **contact** only.

### Testing Instructions
Code review only.

Hope it is accepted so that we can re-use this layout in com_weblinks instead of having to implement it.

### DIFF
--- a/layouts/joomla/content/icons/edit_lock.php
+++ b/layouts/joomla/content/icons/edit_lock.php
@@ -11,15 +11,18 @@ defined('_JEXEC') or die;
 
 use Joomla\CMS\Language\Text;
 
-if (isset($displayData['article']))
+if (isset($displayData['ariaDescribed']))
 {
-	$article = $displayData['article'];
+	$aria_described = $displayData['ariaDescribed'];
+}
+elseif (isset($displayData['article']))
+{
+	$article        = $displayData['article'];
 	$aria_described = 'editarticle-' . (int) $article->id;
 }
-
-if (isset($displayData['contact']))
+elseif (isset($displayData['contact']))
 {
-	$contact = $displayData['contact'];
+	$contact        = $displayData['contact'];
 	$aria_described = 'editcontact-' . (int) $contact->id;
 }
 


brianteeman:
There is also the function checkedout which is used in com_menus and probably elsewhere

joomdonation:
Not sure if I missed anywhere but I searched through the whole source code and only see it is being used in two files:

https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_content/src/Service/HTML/Icon.php#L138
https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_contact/src/Service/HTML/Icon.php#L136

brianteeman:
Yes because the function is used in the other places

joomdonation:
@brianteeman Sorry, I don't look much at the layout, so your comment is not clear to me. From what you see, is it safe to make this change or not?

brianteeman:
The question is not can you make the change to the layout and more do you need to?

All other components are using the function eg
https://github.com/joomla/joomla-cms/blob/72e78f448f12e0b30a1a503440c888c969163134/administrator/components/com_banners/tmpl/banners/default.php#L132

joomdonation:
Ah, no. This is used for edit icon in the frontend such as https://github.com/joomla/joomla-cms/blob/4.0-dev/components/com_contact/tmpl/category/default_items.php#L179. It is not the same with backend.

brianteeman:
ah, i see my confusion and mistake

richard67:
I have tested this item :white_check_mark: successfully on eb4e93a5f5ed4160339aefd293ebfb26048ac2c9<br /><br />LGTM<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35112">issues.joomla.org/tracker/joomla-cms/35112</a>.</sub>

Quy:
I have tested this item :white_check_mark: successfully on eb4e93a5f5ed4160339aefd293ebfb26048ac2c9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35112">issues.joomla.org/tracker/joomla-cms/35112</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35112">issues.joomla.org/tracker/joomla-cms/35112</a>.</sub>

wilsonge:
Thanks!

PR #35121, opened 2021-08-15
Add missing class 'Route'

PhilETaylor:
Code review

File is missing class import needed for code on line 52

```php
$this->app->redirect(Route::_('index.php?option=com_joomlaupdate', true));
```


wilsonge:
Thanks!

PR #35128, opened 2021-08-15
[4.0] build script exclusions

brianteeman:
On the basis that other docs and tests folders have been removed from libraries/vendor this adds some that were _missed_

Easier to remove them now then at a later date

From memory there was a security issue in the past cause by shipping a doc/test folder

@wilsonge 

I guess @richard67 this may need something in the update files script ?


wilsonge:
Thanks!

richard67:
> I guess @richard67 this may need something in the update files script ?

Could be. I will check.

richard67:
See PR #35130 .

brianteeman:
thanks all

PR #35139, opened 2021-08-15
[4] Remove default secret

PhilETaylor:
Code review

Security 101: Do not provide guessable defaults for secrets or credentials because [users cannot be trusted to change default credentials and secrets](https://github.com/search?q=FBVtggIk5lAzEU9H&type=code).

Also can be used to fingerprint/google dork.

### DIFF
--- a/installation/configuration.php-dist
+++ b/installation/configuration.php-dist
@@ -52,7 +52,7 @@ class JConfig
 	public $dbsslcipher = '';
 
 	/* Server Settings */
-	public $secret = '';     // Use something very secure
+	public $secret = '';     // Use something very secure. For example on linux the following command `cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w ${1:-16} | head -n 1`
 	public $gzip = false;
 	public $error_reporting = 'default';
 	public $helpurl = 'https://help.joomla.org/proxy?keyref=Help{major}{minor}:{keyref}&lang={langcode}';
--- a/installation/configuration.php-dist
+++ b/installation/configuration.php-dist
@@ -52,7 +52,7 @@ class JConfig
 	public $dbsslcipher = '';
 
 	/* Server Settings */
-	public $secret = 'FBVtggIk5lAzEU9H';     // Change this to something more secure
+	public $secret = '';     // Use something very secure
 	public $gzip = false;
 	public $error_reporting = 'default';
 	public $helpurl = 'https://help.joomla.org/proxy?keyref=Help{major}{minor}:{keyref}&lang={langcode}';


brianteeman:
You really cant fix stupid. There are only 88 instances on github and they also include server passwords and usernames which is far more serious for them than the secret.

Some people deserve to be hacked

PhilETaylor:
regardless of the actual number or the stupidness of people... not providing defaults for otherwise secure credentials is security 101.

RickR2H:
I have tested this item :white_check_mark: successfully on 141ea3684424f1d5c30fbcb433bde61ad06c8157<br /><br />The secret is generated on install, so IMHO the default can be removed.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35139">issues.joomla.org/tracker/joomla-cms/35139</a>.</sub>

Quy:
I have tested this item :white_check_mark: successfully on 141ea3684424f1d5c30fbcb433bde61ad06c8157<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35139">issues.joomla.org/tracker/joomla-cms/35139</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35139">issues.joomla.org/tracker/joomla-cms/35139</a>.</sub>

bembelimen:
So what happens if a user uses this file and forgots to add a secret? Does any place in Joomla! crashes, because there is no secret?

Additional probably we should extend the comment to help the user to define a secret (16 characters, uppercase, lowercase alphanumeric, ...)

installation/configuration.php-dist
```diff
@@ -52,7 +52,7 @@ class JConfig
 	public $dbsslcipher = '';
 
 	/* Server Settings */
-	public $secret = 'FBVtggIk5lAzEU9H';     // Change this to something more secure
+	public $secret = '';     // Use something very secure
```
wilsonge:
```suggestion
	public $secret = '';     // Use something very secure. For example on linux the following command `cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w ${1:-16} | head -n 1`
```

PhilETaylor: Lol that’s very specific advice haha

wilsonge: I googled "generate a alphanumeric random string on linux" and picked the first answer on SO and changed the length (and tested it on one of my linux boxes) :D

wilsonge:
Thanks!

installation/configuration.php-dist
PhilETaylor: I would have gone with providing an example in PHP because We know hundred percent that they will be running PHP, where as we know that a percentage will not be running Linux

wilsonge: My concern was if they can run commands they probably could have fixed the installer file permissions in the first place. I assume you are running this on an external server either way

Issue #34161, opened 2021-05-24
[4] Fatal error can be created calling Factory::createSession


PhilETaylor:
On code review it appears that the deprecated protected method `createSession` in the `\Joomla\CMS\Factory` abstract class is referencing a non existing class `\JSessionHandlerJoomla` and will therefore PHP Fatal error if anyone attempts to call this method.

This method is not used by Joomla and is already marked as deprecated for removal in 5.0 but as such it should remain working until then - and not creating fatal errors.

https://github.com/joomla/joomla-cms/blob/f64e0186e41df3d68fa6b14d185cb4e60593cab9/libraries/src/Factory.php#L630

eopws:
Then the `$sessionHandler` variable is used to get an instance of a Session class
 https://github.com/joomla/joomla-cms/blob/staging/libraries/src/Factory.php#L621

It's done by call to the `getInstance()` method, which has no arguments at all (https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Session/Session.php#L160-L170).

PhilETaylor:
This issue remains today. This class doesnt exist! this is the only reference to "JSessionHandlerJoomla" in the whole codebase. 

@wilsonge please review

wilsonge:
So I guess we just remove `createSession` as it's totally unused? As the session is now injected into the Application. It is protected so technically it's a b/c break to those subclassing Factory. But it's one of those things where practically it's not going to affect anyone. Especially as there's no real way to easily make this work.

PhilETaylor:
PR https://github.com/joomla/joomla-cms/pull/35146

PR #35146, opened 2021-08-16
[4] Remove createSession method

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/34161#issuecomment-899124491

### Summary of Changes

Remove createSession method which cannot be used anyway as it requires a non existing class

### Testing Instructions

Apply PR - test anything that needs a session

Joomla has methods that cause fatal errors when called

Nothing bad happens :)

### DIFF
--- a/libraries/src/Factory.php
+++ b/libraries/src/Factory.php
@@ -597,49 +597,6 @@ abstract class Factory
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


wilsonge:
Thanks!

PR #35150, opened 2021-08-16
Allow empty folder in zip file

BrainforgeUK:
### Summary of Changes

Only throw error if file does not exist.
If folder in zip file is empty then extract does not create the folder.
This caused the library installation to fail.

### Testing Instructions

Install library
[lib_bftest1.zip](https://github.com/joomla/joomla-cms/files/6991587/lib_bftest1.zip)

image

image

None.

### DIFF
--- a/libraries/src/Installer/Installer.php
+++ b/libraries/src/Installer/Installer.php
@@ -1712,7 +1712,7 @@ class Installer extends Adapter
 					// The source file / folder does not exist. Nothing to copy so set an error.
 					Log::add(Text::sprintf('JLIB_INSTALLER_ERROR_NO_FILE', $filesource), Log::WARNING, 'jerror');
 
-					// Okay to ignore missing folder - it maybe an empty folder in the zip file.
+					// Okay to ignore missing folder - it may be an empty folder in the zip file.
 					if ($filetype === 'folder')
 					{
 						continue;
--- a/libraries/src/Installer/Installer.php
+++ b/libraries/src/Installer/Installer.php
@@ -1709,10 +1709,10 @@ class Installer extends Adapter
 
 				if (!file_exists($filesource))
 				{
-					// The source file / folder does not exist. Nothing to copy so set an error
+					// The source file / folder does not exist. Nothing to copy so set an error.
 					Log::add(Text::sprintf('JLIB_INSTALLER_ERROR_NO_FILE', $filesource), Log::WARNING, 'jerror');
 
-					// Okay to ignore missing folder - it maybe an empty folder in the zip file
+					// Okay to ignore missing folder - it maybe an empty folder in the zip file.
 					if ($filetype === 'folder')
 					{
 						continue;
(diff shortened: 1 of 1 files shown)


bembelimen:
Hello @BrainforgeUK thank you for your PR.
I'm a bit unsure if this is a good way to be honest, if e.g. the unzip process has an error and there is then an empty folder which should have files but there aren't any, the installation succeed instead of fail...

brianteeman:
If I remember correctly we have faced the same issue with core and that is why folders that would otherwise be empty eg cache, tmp and logs all have an index.html in there to ensure that the folder is created. I would recommend that you do the same with your library instead of this PR which as @bembelimen points out can introduce other issues.

libraries/src/Installer/Installer.php
```diff
@@ -1709,12 +1709,15 @@ public function copyFiles($files, $overwrite = null)
 
 				if (!file_exists($filesource))
 				{
-					/*
-					 * The source file does not exist.  Nothing to copy so set an error
-					 * and return false.
-					 */
+					// The source file / folder does not exist. Nothing to copy so set an error
```
PhilETaylor:
```suggestion
					// The source file / folder does not exist. Nothing to copy so set an error.
```

libraries/src/Installer/Installer.php
```diff
@@ -1709,12 +1709,15 @@ public function copyFiles($files, $overwrite = null)
...
 					Log::add(Text::sprintf('JLIB_INSTALLER_ERROR_NO_FILE', $filesource), Log::WARNING, 'jerror');
 
+					// Okay to ignore missing folder - it maybe an empty folder in the zip file
```
PhilETaylor:
```suggestion
					// Okay to ignore missing folder - it maybe an empty folder in the zip file.
```

BrainforgeUK: Improved comment.

libraries/src/Installer/Installer.php
```diff
@@ -1709,12 +1709,15 @@ public function copyFiles($files, $overwrite = null)
...
 				if (!file_exists($filesource))
 				{
-					/*
-					 * The source file does not exist.  Nothing to copy so set an error
-					 * and return false.
-					 */
+					// The source file / folder does not exist. Nothing to copy so set an error.
 					Log::add(Text::sprintf('JLIB_INSTALLER_ERROR_NO_FILE', $filesource), Log::WARNING, 'jerror');
 
+					// Okay to ignore missing folder - it maybe an empty folder in the zip file.
```
brianteeman:
```suggestion
					// Okay to ignore missing folder - it may be an empty folder in the zip file.
```

it is a verb phrase in this usecase not an adverb

BrainforgeUK: Improved comment text.

brianteeman:
Empty folders in a zip file are bad. Windows even even removes them when you create zips. My 2c is that we dont need a solution in core for this and developers just have to include an index.html or similar in the folder if they want to include it in the zip

bembelimen:
Thinking about this, there still shouldn't be empty folder ignored when it's there. So I'm closing this for now.

Still thanks for your contribution.

Issue #35043, opened 2021-08-04
J4 not working switch RECORD HITS


korenevskiy:
![image](https://user-images.githubusercontent.com/6898474/128238875-fd884b74-a53b-46c5-b020-ace4ce5abfe4.png)
![image](https://user-images.githubusercontent.com/6898474/128239158-cb814349-0f5c-4a70-9f92-4341000ac344.png)

In the Article settings. The RECORD HITS field is disabled.
But this value is displayed on the site.
The fact is that the disabled RECORD HITS field hides another HITS field.
I believe that it should be like this:
- Either the RECORD HITS field will not hide the HITS field.
- Either disabling the RECORD HITS field will automatically deactivate the HITS field.
![image](https://user-images.githubusercontent.com/6898474/128240189-14b98eb6-6467-415b-af7b-5524bbf04d38.png)

ChristineWk:
> In the Article settings. The RECORD HITS field is disabled.
> But this value is displayed on the site.

It's not displayed in Frontend.

You can set RECORD HITS to Yes and HITS to No. also.

#34211

korenevskiy:
> 
> > In the Article settings. The RECORD HITS field is disabled.
> > But this value is displayed on the site.
> 
> It's not displayed in Frontend.
> 
> You can set RECORD HITS to Yes and HITS to No. also.
> 
> #34211

there are no questions about what you wrote in the article.
The error is when RECORD HITS disabled and HITS enabled is selected.
At the same time, the HITS checkbox is hidden , which means it should also not work.
If RECORDS HITS operates on the principle of disable HITS at all on the site. Then why show hits?

brianteeman:
In this case the showon in the xml is the problem and should be removed so that you can have any combination of show and record hits

korenevskiy:
> In this case the showon in the xml is the problem and should be removed so that you can have any combination of show and record hits

I'm deactivating RECORDS HITS. Then I look at the client's website. And at a loss that the Hits are shown.
After that, I go to the settings, look and see that the hits are turned off. I can't find the setting where hits are enabled at all. Since all visible settings are disabled. But the hits are still being shown.
User experience shows that when you turn off Hits, hits will be shown on the client, this means that the second switch should not disappear, since its function continues to work, and it is no longer visible, hidden.

brianteeman:
Sorry I guess my reply was lost in translation. I agree with you and the way to do it is to change the xml file and delete the line with showon. Will you do a PR

[...]

PR #32920, opened 2021-03-29
[4.0] Article hits (site)

brianteeman:
If you are not recording the hits then there is no point in showing the hits. This PR changes the order of the two fields and makes the show_hits field display only if record_hits is set to yes

### before
image

### after
hits

PR #35154, opened 2021-08-16
[J4] not working switch RECORD HITS

korenevskiy:
image
image

In the Article settings. The RECORD HITS field is disabled.
But this value is displayed on the site.
The fact is that the disabled RECORD HITS field hides another HITS field.
I believe that it should be like this:
- Either the RECORD HITS field will not hide the HITS field.
- Either disabling the RECORD HITS field will automatically deactivate the HITS field.
image

### DIFF
--- a/administrator/components/com_content/config.xml
+++ b/administrator/components/com_content/config.xml
@@ -297,7 +297,6 @@
 			label="JGLOBAL_SHOW_HITS_LABEL"
 			layout="joomla.form.field.radio.switcher"
 			default="1"
-			showon="record_hits:1"
 			>
 			<option value="0">JHIDE</option>
 			<option value="1">JSHOW</option>


korenevskiy:
https://github.com/joomla/joomla-cms/issues/35043

brianteeman:
I have tested this item :white_check_mark: successfully on 9feee70a3e225bbdf3484ca155c44036a986d8f7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35154">issues.joomla.org/tracker/joomla-cms/35154</a>.</sub>

richard67:
@korenevskiy Does this PR solve your issue #35043 ? If yes then please mention the issue in the dexcription of your PR like it should be and like it is suggested by the PR template when creating a new PR:

> Pull Request for Issue #35043 .

And the headings for testing instructions we also don't have without reason in our PR templates.

And finally, please close your issue when having made a PR which claims to solve it.

Thanks in advance.

richard67:
In general we have that problem everywhere where we use showon because we normally don't use js to reset fields to their default value when they are hidden. It always can happen you hide a field by changing an option but that field still has its value. That's a general issue with showon.

brianteeman:
@richard67 that is true but its rarely an issue - here it is

richard67:
> @richard67 that is true but its rarely an issue - here it is

@brianteeman Yes, I agree, and it can be a valid scenario for some site owner to stop recording hits but still show the old count and so have these 2 options independently.

richard67:
@brianteeman What's funny: It was you who implemented it with PR #32920 ;-) But well, things happen.

richard67:
I have tested this item :white_check_mark: successfully on 9feee70a3e225bbdf3484ca155c44036a986d8f7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35154">issues.joomla.org/tracker/joomla-cms/35154</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35154">issues.joomla.org/tracker/joomla-cms/35154</a>.</sub>

korenevskiy:
> In general we have that problem everywhere where we use showon because we normally don't use js to reset fields to their default value when they are hidden. It always can happen you hide a field by changing an option but that field still has its value. That's a general issue with showon.

I think that JS is not needed ever. ShowOn needs to be checked in PHP  as well. Just in PHP conditions, you need to do double checks of 2 fields. The fact is that there may be cases when you do not need to reset hidden fields to default values. For example, a hidden text field will simply store the text. But we will not use it. I think that we have no problems with hidden fields.

bembelimen:
Thx

korenevskiy:
Can the patch be removed?.
Removing the patch will not affect the PR?

Quy:
@korenevskiy Once merged, it is safe to remove the patch.

PR #35168, opened 2021-08-17
Installation cleanup

wilsonge:
### Summary of Changes
Completing installation now deletes the installation directory "automatically" before continuing when not in development mode. When in development mode you continue to get the existing button.

This allows the user to delete installation rather than currently being stuck as the red button is already currently only shown in development mode. Massive shoutout to @zero-24 for last minute finding this one. Would have made for a very fast 4.0.1 otherwise :)

You can't delete the installation directory when the build is "stable"

Screenshot 2021-08-17 at 08 29 50
Screenshot 2021-08-17 at 08 30 08

In development mode the installation directory is not deleted when clicking complete. Only when pressing the big red button. In stable mode it is only when you click either continue button the installation directory is deleted and you are allowed to proceed

1. Go to the installer and change the `Joomla\Version\Version::DEV_STATUS` to be `Stable`. Ensure you can't see the red button but when clicking on complete the installation directory is removed before you are allowed to proceed
2. Go to the installer leaving yourself in dev mode. Ensure the red button is available and on clicking it you have the installation folder deleted. But when clicking on complete the redirect happens with no installation folder removal.

None.

### DIFF
--- a/installation/src/Model/CleanupModel.php
+++ b/installation/src/Model/CleanupModel.php
@@ -38,7 +38,8 @@ class CleanupModel extends BaseInstallationModel
 			$return = File::move(JPATH_ROOT . '/robots.txt.dist', JPATH_ROOT . '/robots.txt');
 		}
 
-		if (function_exists('opcache_reset')) {
+		if (function_exists('opcache_reset'))
+		{
 			\opcache_reset();
 		}
 
--- a/installation/src/Model/CleanupModel.php
+++ b/installation/src/Model/CleanupModel.php
@@ -14,6 +14,9 @@ namespace Joomla\CMS\Installation\Model;
 use Joomla\CMS\Filesystem\File;
 use Joomla\CMS\Filesystem\Folder;
 
+use function clearstatcache;
+use function function_exists;
+
 /**
  * Cleanup model for the Joomla Core Installer.
  *
@@ -37,6 +40,12 @@ class CleanupModel extends BaseInstallationModel
 		{
 			$return = File::move(JPATH_ROOT . '/robots.txt.dist', JPATH_ROOT . '/robots.txt');
 		}
+
+		if (function_exists('opcache_reset')) {
+			\opcache_reset();
+		}
+
+		\clearstatcache(true, JPATH_INSTALLATION . '/index.php');
 
 		return $return;
 	}
--- a/installation/src/Model/CleanupModel.php
+++ b/installation/src/Model/CleanupModel.php
@@ -14,9 +14,6 @@ namespace Joomla\CMS\Installation\Model;
 use Joomla\CMS\Filesystem\File;
 use Joomla\CMS\Filesystem\Folder;
 
-use function clearstatcache;
-use function function_exists;
-
 /**
  * Cleanup model for the Joomla Core Installer.
  *
--- a/installation/template/js/remove.js
+++ b/installation/template/js/remove.js
@@ -51,7 +51,7 @@ completeInstallationOptions.forEach(function(item) {
         });
 
         // In development mode we show the user a pretty button to allow them to choose whether to delete the installation
-        // directory or not. In stable the decision is made forward. Maximum extermination!
+        // directory or not. In stable the decision is made for us. Maximum extermination!
         if ('development' in item.dataset) {
             window.location.href = item.dataset.href;
         } else {
--- a/installation/template/js/remove.js
+++ b/installation/template/js/remove.js
@@ -51,7 +51,7 @@ completeInstallationOptions.forEach(function(item) {
         });
 
         // In development mode we show the user a pretty button to allow them to choose whether to delete the installation
-        // directory or not. In stable the decision is made for us. Maximum extermination!
+        // directory or not. In stable release we always delete the folder. Maximum extermination!
         if ('development' in item.dataset) {
             window.location.href = item.dataset.href;
         } else {
--- a/installation/template/js/remove.js
+++ b/installation/template/js/remove.js
@@ -45,6 +45,11 @@ const completeInstallationOptions = document.querySelectorAll('.complete-install
 
 completeInstallationOptions.forEach(function(item) {
     item.addEventListener('click', function (e) {
+        // Evil voodoo. Once a button is clicked ensure they don't spam click it...
+        completeInstallationOptions.forEach(function(nestedItem) {
+            nestedItem.disabled = true;
+        });
+
         // In development mode we show the user a pretty button to allow them to choose whether to delete the installation
         // directory or not. In stable the decision is made forward. Maximum extermination!
         if ('development' in item.dataset) {
@@ -78,7 +83,11 @@ Joomla.deleteJoomlaInstallationDirectory = function (redirectUrl) {
                 const customInstallation = document.getElementById('customInstallation');
                 customInstallation.parentNode.removeChild(customInstallation);
                 const removeInstallationTab = document.getElementById('removeInstallationTab');
-                removeInstallationTab.parentNode.removeChild(removeInstallationTab);
+
+                // This will only exist in debug mode
+                if (removeInstallationTab) {
+                    removeInstallationTab.parentNode.removeChild(removeInstallationTab);
+                }
 
                 if (redirectUrl) {
                     window.location.href = redirectUrl;
--- a/installation/template/js/remove.js
+++ b/installation/template/js/remove.js
@@ -36,37 +36,59 @@ if (document.getElementById('removeInstallationFolder')) {
 			e.preventDefault();
 			let confirm = window.confirm(Joomla.Text._('INSTL_REMOVE_INST_FOLDER').replace('%s', 'installation'));
 			if (confirm) {
-				Joomla.request({
-					method: "POST",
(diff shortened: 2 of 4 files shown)


webgras:
I have tested this item :white_check_mark: successfully on cbdd6d0870e320e4c9a8dcfd978c4dec0f70311e<br /><br />I downloaded from here: https://ci.joomla.org/artifacts/joomla/joomla-cms/4.0-dev/35168/downloads/46699/
Tested folder in dev.
Then changed the following to act like stable (thanks @zero-24):
libraries\src\Version.php
const EXTRA_VERSION = '';
const DEV_STATUS = 'stable';<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35168">issues.joomla.org/tracker/joomla-cms/35168</a>.</sub>

richard67:
@wilsonge System tests are failing because they expect the button to be shown. It needs to adapt them to the changes here.

installation/template/js/remove.js
```diff
@@ -36,37 +36,68 @@ if (document.getElementById('removeInstallationFolder')) {
...
+			    Joomla.deleteJoomlaInstallationDirectory();
+			}
+		});
+}
+
+const completeInstallationOptions = document.querySelectorAll('.complete-installation');
+
+completeInstallationOptions.forEach(function(item) {
+    item.addEventListener('click', function (e) {
+        // Evil voodoo. Once a button is clicked ensure they don't spam click it...
```
brianteeman:
```suggestion
        // Once a button is clicked ensure they can't click it again...
```

Lets keep it professional ;)

installation/template/js/remove.js
```diff
@@ -36,37 +36,68 @@ if (document.getElementById('removeInstallationFolder')) {
...
+        completeInstallationOptions.forEach(function(nestedItem) {
+            nestedItem.disabled = true;
+        });
+
+        // In development mode we show the user a pretty button to allow them to choose whether to delete the installation
+        // directory or not. In stable the decision is made forward. Maximum extermination!
```
PhilETaylor, line 54: ?

```suggestion
        // directory or not. In stable the decision is made for us. Maximum extermination!
```

brianteeman:
```suggestion
        // directory or not. In a stable release their is no choice. The installation directory is always removed.
```

I think you need some sleep ;)

PhilETaylor: s/their/there

wilsonge: Can I not have one doctor who reference :P But also yes probably 1.5 hours isn't enough :)

wilsonge: also isn't it "there" in this case?

richard67: "Maximum extermination!" ... daleks around? ;-)

PhilETaylor: > also isn't it "there" in this case?

There is in Brians suggestion... which is the wrong use of "their"

installation/tmpl/remove/default.php
```diff
@@ -185,8 +185,14 @@ class="j-install-step-form flex-column mt-5 border rounded d-none"
 				<?php echo HTMLHelper::_('form.token'); ?>
 
 				<div class="form-group j-install-last-step d-grid gap-2">
-					<a class="btn btn-primary w-100" href="<?php echo Uri::root(); ?>" title="<?php echo Text::_('JSITE'); ?>"><span class="icon-eye" aria-hidden="true"></span> <?php echo Text::_('INSTL_COMPLETE_SITE_BTN'); ?></a>
-					<a class="btn btn-primary w-100" href="<?php echo Uri::root(); ?>administrator/" title="<?php echo Text::_('JADMINISTRATOR'); ?>"><span class="icon-lock" aria-hidden="true"></span> <?php echo Text::_('INSTL_COMPLETE_ADMIN_BTN'); ?></a>
+					<button type="button" class="complete-installation btn btn-primary w-100"
+					   data-href="<?php echo Uri::root(); ?>" <?php if ($this->development): ?>data-development<?php endif; ?> title="<?php echo Text::_('JSITE'); ?>">
```
brianteeman:
```suggestion
					   data-href="<?php echo Uri::root(); ?>" <?php if ($this->development): ?>data-development<?php endif; ?>">
```

brianteeman: A title is useless on a button ;)

installation/tmpl/remove/default.php
```diff
@@ -185,8 +185,14 @@ class="j-install-step-form flex-column mt-5 border rounded d-none"
...
+						<span class="icon-eye" aria-hidden="true"></span> <?php echo Text::_('INSTL_COMPLETE_SITE_BTN'); ?>
+					</button>
+					<button type="button" class="complete-installation btn btn-primary w-100"
+					   data-href="<?php echo Uri::root(); ?>administrator/" <?php if ($this->development): ?>data-development<?php endif; ?> title="<?php echo Text::_('JADMINISTRATOR'); ?>">
```
brianteeman:
```suggestion
					   data-href="<?php echo Uri::root(); ?>administrator/" <?php if ($this->development): ?>data-development<?php endif; ?>">
```

installation/template/js/remove.js
richard67: yep:

> In a stable release their is no choice. The installation directory is always removed.

should be "there".

But why not just use:

> In a stable release the installation directory is always removed.

installation/src/Model/CleanupModel.php
```diff
@@ -38,6 +41,12 @@ public function deleteInstallationFolder()
 			$return = File::move(JPATH_ROOT . '/robots.txt.dist', JPATH_ROOT . '/robots.txt');
 		}
 
+		if (function_exists('opcache_reset')) {
+			\opcache_reset();
```
PhilETaylor, line 43: ffs have we learned nothing!?!?!??!

DO NOT DO THIS EVER !!!! 

The correct way https://github.com/joomla/joomla-cms/pull/32915

wilsonge: This doesn't work for our case. we've deleted the folder not a single file. I agree we do need to get #32915 in though.

wilsonge: In theory we can try and just invalidate installation/index.php as that's what we actually test for? But not sure whether it's really a lesser evil.

PhilETaylor: NEVER DO A opcache_reset EVER for all the reasons already stated.

You should be invalidating the opcache on deleting each individual file, the opcache needs to be invalidated BEFORE the file is deleted and not after... if you read the source code of PHP this becomes clear. 

Doing a `opcache_reset` is a complete SERVER WIDE change

wilsonge: OK So concretely @HLeithner on his server gets race conditions without this line. And opcache invalidating that line also doesn't seem to solve it. Unclear right now if opcache just takes a bit longer so solves the race condition or there's some concrete cache being solved. But we'll go with this for 4.0.0 and then we can spend some time and figure out if we're just masking the issue or not for 4.0.1. Again fully agree that we need #32915 - just it doesn't apply to this specific case.

wilsonge:
We'll fix drone this evening once work is done.

PhilETaylor:
Well maybe @HLeithner <https://github.com/HLeithner>  should have tested #32915 before release day!?  The "race condition" is not a race condition, its the default timeout of 2 seconds for opcache and a developer that doesnt understand opcache.

Throwing in opcache_reset seconds before a major release is a major step back!

installation/src/Model/CleanupModel.php
PhilETaylor: Well maybe @HLeithner  should have tested #32915 before release day!?  The "race condition" is not a race condition, its the default timeout of 2 seconds for opcache and a developer that doesnt understand opcache.

Throwing in opcache_reset seconds before a major release is a major step back!

richard67: > In theory we can try and just invalidate installation/index.php as that's what we actually test for? But not sure whether it's really a lesser evil.

Can someone make a new PR to change it that way?

PhilETaylor: And dare I link to Nic's comments about the 2 second opcache timeout  https://github.com/joomla/joomla-cms/pull/32915#issuecomment-887717530

richard67:
@wilsonge I think @PhilETaylor is right here. Should he make a PR to change from "opcache_reset" to something like here for the "installation/index.php" file?

https://github.com/joomla/joomla-cms/pull/32915/files#diff-87944bb26710d00cc11ee21e5a9c9242c5fd957acade50f9eed59b84f87b563bR1644-R1649

installation/src/Model/CleanupModel.php
HLeithner: @PhilETaylor I changed it now to invalidate https://github.com/joomla/joomla-cms/commit/cd2ea50de83d0d72d1bf0dc068e39412e7b8969b

PhilETaylor: Oh well I see this was merged anyway...

Issue #35170, opened 2021-08-17
[4] installation folder delete failure gives no messages


PhilETaylor:
### Steps to reproduce the issue

@wilsonge @HLeithner
last min changes in #35168 got merged - oh well.
If Joomla is unable to delete the installation folder (permissions issues, or opcache_invalidate disabled by server admin

### Expected result

Helpful messages

### Actual result

`chmod 000 installation`

<img width="892" alt="Screenshot 2021-08-17 at 09 31 07" src="https://user-images.githubusercontent.com/400092/129692657-fc135ef4-8e25-4543-b6f1-bd8cb8b1be77.png">

or 

add die() in /installation/index.php

<img width="775" alt="Screenshot 2021-08-17 at 09 41 39" src="https://user-images.githubusercontent.com/400092/129693679-4dea9d2f-32fd-4ebc-a6a7-6fbb1420fe39.png">

Fedik:
please try this one #35173

HLeithner:
Thanks for the report fixed in #35173 and #35171

PR #35173, opened 2021-08-17
[4.0] Fix installation js for messages

Fedik:
Pull Request for Issue #35170 .

### Summary of Changes

Correct call of `Joomla.renderMessage()`

### Testing Instructions
Please follow #35170

### DIFF
--- a/installation/template/js/remove.js
+++ b/installation/template/js/remove.js
@@ -73,11 +73,11 @@ Joomla.deleteJoomlaInstallationDirectory = function (redirectUrl) {
             const successresponse = JSON.parse(response);
             if (successresponse.error === true) {
                 if (successresponse.messages) {
-                    Joomla.renderMessages(successresponse.messages, '#system-message-container');
+                    Joomla.renderMessages(successresponse.messages);
                     Joomla.loadOptions({'csrf.token': successresponse.token});
                 } else {
                     // Stuff went wrong. No error messages. Just panic bail!
-                    Joomla.renderMessages('Unknown error deleting the installation folder.', '#system-message-container');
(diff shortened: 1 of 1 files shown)


installation/template/js/remove.js
```diff
@@ -73,11 +73,11 @@ Joomla.deleteJoomlaInstallationDirectory = function (redirectUrl) {
             const successresponse = JSON.parse(response);
             if (successresponse.error === true) {
                 if (successresponse.messages) {
-                    Joomla.renderMessages(successresponse.messages, '#system-message-container');
+                    Joomla.renderMessages(successresponse.messages);
```
wilsonge, line 76: This needs to have the message in an object too with an error key

Fedik: it should defaulting there:
https://github.com/joomla/joomla-cms/blob/faf9352ba1a5f822cdf73bb4e1dfd9fa39340fc3/build/media_source/system/js/messages.es6.js#L22-L24

That why I removed it here.

Or you got any bug with it?

wilsonge, line 76:
```suggestion
                    Joomla.renderMessages({error: [successresponse.messages]});
```

wilsonge: I mean I *think* this is required the same as we have below? Sorry I wasn't clear above!

wilsonge:
Thanks!

installation/template/js/remove.js
Fedik: hm, here we actually have an object of messages?

PhilETaylor:
Seems to work... 
Screenshot 2021-08-17 at 10 22 48

PR #35187, opened 2021-08-17
[4] Remove undefined class

PhilETaylor:
Code review,

Constructors dont ever return - they construct. 

The comment refers to the non-existing class `AECoreTimer` so thats blatantly wrong, as even the class being constructed is called `AKCoreTimer` (note the K and the E)


wilsonge:
Thanks!

PR #35194, opened 2021-08-17
[4] Correct FQCN for WebAssetManager

PhilETaylor:
Correct FQCN to aid IDE autocompletion and type hint `$wa`


wilsonge:
Thanks!

Issue #35169, opened 2021-08-17
[4] crypto@1.0.1: This package is no longer supported. It's now a built-in Node module


PhilETaylor:
### Steps to reproduce the issue

`npm ci`

### Actual result

```
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
npm WARN deprecated crypto@1.0.1: This package is no longer supported. It's now a built-in Node module. If you've depended on crypto, you should
```

### System information (as much as possible)

```
 npm -v
7.20.6
 node -v
v16.6.2
```

PhilETaylor:
```
merge  <2.1.1
Severity: high
Prototype Pollution - https://npmjs.com/advisories/1666
fix available via `npm audit fix --force`
Will install watch@0.13.0, which is a breaking change
node_modules/merge
  exec-sh  <=0.3.1
  Depends on vulnerable versions of merge
  node_modules/exec-sh
    watch  >=0.14.0
    Depends on vulnerable versions of exec-sh
    node_modules/watch

3 high severity vulnerabilities
```

ghazal:
Issue confirmed.
Same config:
npm -v
7.20.6
node -v
v16.6.2

brianteeman:
This can wait until after the release as it is not include in the actual release

dgrammatiko:
PR: https://github.com/joomla/joomla-cms/pull/35198 (combined with another issue for build tools)

PhilETaylor:
ta

PR #35048, opened 2021-08-05
[4.0] Let build tools always start clean

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Clean all the Joomla core media folders each time that `npm install` is called

### Testing Instructions
Before applying the patch make a backup of your current media folder
Apply the patch
Run `npm install`
Compare the backup folder with the current media

### Documentation Changes Required

@wilsonge this can wait till 4.0 is released

PR #35198, opened 2021-08-17
Build tools clean up/audit/fixes

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/35169

### Summary of Changes

- replace the npm package `watch` with `chokidar`, the latter is well supported and used by Rollup (which also used by the tools)
- Remove unused Babel plugins
- Combine https://github.com/joomla/joomla-cms/pull/35048

### Testing Instructions

Test if `npm install` works correctly 
Test if `npm run watch` works correctly

#35048 had a very good question from @richard67 for the cleanUpFolders:

```
Will someone remember to maintain this hardcoded list if some new extension which has assets will be added or removed in future?
```

The answer is that I can add a little script that extracts these programmatically (run on build.php) so nobody will ever have to do anything). But that will be on another PR...


wilsonge:
Thanks!

PR #35207, opened 2021-08-17
[4] Correctly re-translate the date/time string in the calendar based on language file

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/35204

### Summary of Changes

Correctly re-translate the date/time string in the calendar based on language file

Also set the CORRECT en-GB value for the date/time string in PHP syntax DATE_FORMAT_CALENDAR_DATETIME as "%Y-%m-%d %H:%i:%s" - note this has been copied and pasted to other language packs and EVERY ONE will be broken... (Checked US and AU ) 

### Testing Instructions

Install Joomla 4 with US language set as default
Try to create an article manually specifying a start publishing date and try to save

Screenshot 2021-08-17 at 22 52 52

Article Saves


PhilETaylor:
hmm hang on. Fixed PHP. Broke JS.

HLeithner:
> `%H:%M:%S` is the wrong time... in all the packs I looked at (GB US AU)
> 
> ```
> - DATE_FORMAT_CALENDAR_DATETIME="%Y-%m-%d %H:%M:%S"
> + DATE_FORMAT_CALENDAR_DATETIME="%Y-%m-%d %H:%i:%s"
> ```

does the new calendar uses a different format? The value is 5 years old? https://github.com/joomla/joomla-cms/pull/12102

HLeithner:
Maybe this string is used later in php which it shouldn't?

PhilETaylor:
> Maybe this string is used later in php which it shouldn't?

Which is my exact fix. WHY ON EARTH have two different ways to specify dates....

PhilETaylor:
so basically the bug is that the date/time is in US format coming out of the calendar and we have to convert that to SQL format to save it... in php....

PhilETaylor:
So you are saying you rather keep `%H:%M:%S` just because its been that way for 5 years, and stop reusing it in PHPthe way I proposed (which is the quick 2 line fix to the reported bug, the bug that means any language not using a SQL based date cannot save an article with a publish date......) 

ok back to the drawing board...

PhilETaylor:
replaced by https://github.com/joomla/joomla-cms/pull/35208

particthistle:
* en-GB error may have been in core string.
* en-AU and en-US were then a copy at that specific point in time.
* At the time I translated them, the date string did not have the code highlighting on it to indicate to translators "Don't change this one"

PhilETaylor:
The point is they should be translated so that Americans can have their crazy date format ;-) 

Joomla should Display the American style when using US but internally we still use a standard that is compatible with sql as that's where all the dates end up.

Using M and S in JS instead of i and s just seems insane to me but I'm told it's been that way for 5 years.

My replacement PR to this one allows the US / translated date formats to be visible on calendars at the same time as allowing The data to be saved to sql

PR #35209, opened 2021-08-18
[4.0] Load default language overrides

bembelimen:
### Summary of Changes
On multilingual sites the translations of language strings falls back to the default language (when not in debug mode). But if you have created some language overrides this behaviour does not apply.

### Testing Instructions
- Create a multlingual page
- Install and activate at least one additional language to en-GB
- Create a language override for a not existing constant like: ```TEST_CONSTANT="Test"``` for the english frontend language
- open the index.php of your frontend template and put: ```echo Text::_('TEST_CONSTANT');```
- Switch to the 2nd language

```TEST_CONSTANT``` is outputted.

```Test``` is outputted.

### DIFF
--- a/libraries/src/Language/Language.php
+++ b/libraries/src/Language/Language.php
@@ -193,6 +193,16 @@ class Language
 		$this->metadata = LanguageHelper::getMetadata($this->lang);
 		$this->setDebug($debug);
 
+		/*
+		Let's load the default override once, so we can profit from that, too
+		But make sure, that we don't enforce it on each language file load.
+		So don't put it in $this->override
+		*/
+		if ($lang !== $this->default)
+		{
+			$this->loadLanguage(JPATH_BASE . '/language/overrides/' . $this->default . '.override.ini');
+		}
+
 		$this->override = $this->parse(JPATH_BASE . '/language/overrides/' . $lang . '.override.ini');
 
 		// Look for a language specific localise class
--- a/libraries/src/Language/Language.php
+++ b/libraries/src/Language/Language.php
@@ -198,7 +198,7 @@ class Language
 		But make sure, that we don't enforce it on each language file load.
 		So don't put it in $this->override
 		*/
-		if ($lang !== $this->default)
+		if (!$this->debug && $lang !== $this->default)
 		{
 			$this->loadLanguage(JPATH_BASE . '/language/overrides/' . $this->default . '.override.ini');
 		}
--- a/libraries/src/Language/Language.php
+++ b/libraries/src/Language/Language.php
@@ -194,10 +194,10 @@ class Language
 		$this->setDebug($debug);
 
 		/*
-		* Let's load the default override once, so we can profit from that, too
-		* But make sure, that we don't enforce it on each language file load.
-		* So don't put it in $this->override
-		*/
+		 * Let's load the default override once, so we can profit from that, too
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: libraries/src/Language/Language.php [hunk 1]
```diff
@@ -193,6 +193,16 @@ public function __construct($lang = null, $debug = false)
...
 		$this->setDebug($debug);
 
+		/*
+		Let's load the default override once, so we can profit from that, too
+		But make sure, that we don't enforce it on each language file load.
+		So don't put it in $this->override
+		*/
+		if (!$this->debug && $lang !== $this->default)
+		{
+			$this->loadLanguage(JPATH_BASE . '/language/overrides/' . $this->default . '.override.ini');
```
wilsonge, line 203: Just for my own clarity - we call `loadLanguage` instead of `parse` directly so it can go into `$this->strings`?

bembelimen: Yes, ```parse``` only returns the strings ```loadLanguage``` parses internally and put it in ```$this->strings```. Important is, that we don't do it in ```$this->override``` otherwise we would override again and again but we just handle it like a normal ini language file.

libraries/src/Language/Language.php
```diff
@@ -193,6 +193,16 @@ public function __construct($lang = null, $debug = false)
 		$this->metadata = LanguageHelper::getMetadata($this->lang);
 		$this->setDebug($debug);
 
+		/*
+		Let's load the default override once, so we can profit from that, too
+		But make sure, that we don't enforce it on each language file load.
+		So don't put it in $this->override
```
PhilETaylor: https://developer.joomla.org/coding-standards/inline-code-comments.html 

```
Comment blocks that introduce large sections of code and are more than 2 
lines long should use /* */ (C) style and should use * on each line with the 
same space/tab rules as doc blocks.
```

```suggestion
		* Let's load the default override once, so we can profit from that, too
		* But make sure, that we don't enforce it on each language file load.
		* So don't put it in $this->override
```

maikol-ortigueira:
I have tested this item :white_check_mark: successfully on ee4dd241400675bbab0a9a72ec4cb0a1d339c43b<br /><br />I've tested successfully.

Thank you.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35209">issues.joomla.org/tracker/joomla-cms/35209</a>.</sub>

BertaOctech:
I have tested this item :white_check_mark: successfully on ee4dd241400675bbab0a9a72ec4cb0a1d339c43b<br /><br />Applying the patch had the described effect

Thanks.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35209">issues.joomla.org/tracker/joomla-cms/35209</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35209">issues.joomla.org/tracker/joomla-cms/35209</a>.</sub>

drmenzelit:
Thanks

PR #35221, opened 2021-08-18
[4.0] Add deprecation note for CMSObject

ditsuke:
### Summary of Changes
CMSObject was deprecated in #4910 but it isn't always immediately obvious what to use instead. Depending on the use case, either `stdClass` or `Registry` replaces CMSObject. This PR adds a deprecation note which provides some useful clarity.

### DIFF
--- a/libraries/src/Object/CMSObject.php
+++ b/libraries/src/Object/CMSObject.php
@@ -17,7 +17,7 @@ namespace Joomla\CMS\Object;
  * and an internal error handler.
  *
  * @since       1.7.0
- * @deprecated  4.0.0
+ * @deprecated  4.0.0  Use \StdClass or Joomla\Registry\Registry instead
  */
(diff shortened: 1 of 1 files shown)


libraries/src/Object/CMSObject.php
```diff
@@ -17,7 +17,7 @@
  * and an internal error handler.
  *
  * @since       1.7.0
- * @deprecated  4.0.0
+ * @deprecated  4.0.0  Use \StdClass or Joomla\Registry\Registry instead
```
PhilETaylor:
```suggestion
 * @deprecated  4.0.0  Use \stdClass or \Joomla\Registry\Registry instead.
```

ditsuke: Thank you!

wilsonge:
Thanks!

PR #12414, opened 2016-10-14
[4.0] Adding filter() and postProcessing() to FormField

Hackwar:
## Issues to adress

The `Form` class currently is a god class, with lots of duties and equally a lot of hardcoded behavior, especially when it comes to handling the form data after it has been send from the browser. It contains field specific hardcoded filters, which modify the data and actually makes it very hard for third party developers to provide their own, extended logic to handle the data. It is also not entirely clear (at least to me) how the whole concept out of Form, `FormField` objects and `FormRule` objects is supposed to work.

## (New) expected behavior

When the component receives the data from the form, it sets up the `Form` object with the right form definition and then calls `Form::process($data)`.
`Form::process()` then calls `Form::filter($data)`, which cleans up the data, removing unnecessary spaces, etc., preparing it to be validated with `Form::validate()`. `Form::validate($data)` will either return true or false, based on all fields having a valid content. If `Form::validate()` returns true, `Form:;process()` calls `Form::postProcess($data)`, which allows to post-process the data. 
Each of these methods does not have any specific logic itself, but will load the respective `FormField` object for the given field and value and call `FormField::filter()`, `FormField::validate()` and `FormField::postProcess()` on it. `FormField::filter()` and `FormField::validate()` in the base class allow for `FormRule` and `FormFilter` objects to be used to filter and validate data, but you can also override that functionality in your specific field, for example filtering the input on a calendar field on a localised date format and validating also that after filtering that is indeed the value that we want.
In `FormField::postProcess()`, the aforementioned calendar field would then convert the date from the localised format to a normalised format, like the MySQL date format. Another use would be to move an uploaded file to its final location in a `FormFieldUpload` field.
## Implications for B/C

This change is not completely backwards compatible. While most components using the core MVC classes will not need any modifications, components that override the save() method of their model/controller will most likely need some light tweaking. custom FormField and FormRule classes might need some modifications, but that depends on the respective code.

PR #24417, opened 2019-03-30
Fix to JFormFieldCalendar using different formats

Arpit-24:
Pull Request for Issue [#21148](https://github.com/joomla/joomla-cms/issues/21148) .

### Summary of Changes

Increased colspan of amp/pm selector to make it visible
Changed $this->format to the fixed format of "%Y-%m-%d %H:%M:%S" fixes the issue as the format is already converted and displayed properly using the calendar.js file so no need to change the format twice.

### Testing Instructions

Please refer Issue [#21148](https://github.com/joomla/joomla-cms/issues/21148)

Earlier there was a blank field displayed in the input text field after saving while it was correctly saved in the database.

Screenshot (1263)

The selection now appears properly :

Screenshot (1268)
Screenshot (1269)

PR #35222, opened 2021-08-18
Fix CalendarField::filter to use correct format

Fedik:
Pull Request for Issue #35204 .

### Summary of Changes

Correcting CalendarField::filter to use `DATE_FORMAT_FILTER_DATETIME` for translated date.
Was broken somewhere after #12414

Technical note: there should be 2 string translated:
`DATE_FORMAT_CALENDAR_DATETIME` - used for display (uses strftime format)
`DATE_FORMAT_FILTER_DATETIME` - used for server side filtering  (uses datetime format),

### Testing Instructions
Please follow #35204

none

### DIFF
--- a/libraries/src/Form/Field/CalendarField.php
+++ b/libraries/src/Form/Field/CalendarField.php
@@ -53,7 +53,7 @@ class CalendarField extends FormField
 	/**
 	 * Whether format should be translated
 	 *
-	 * @var    bool
+	 * @var    boolean
 	 * @since  __DEPLOY_VERSION__
 	 */
 	protected $translateFormat;
--- a/libraries/src/Form/Field/CalendarField.php
+++ b/libraries/src/Form/Field/CalendarField.php
@@ -51,12 +51,12 @@ class CalendarField extends FormField
 	protected $format;
 
 	/**
-	 * Whether format should be translated
+	 * The format will be used to filter submitted date and time.
 	 *
-	 * @var    boolean
+	 * @var    string
 	 * @since  __DEPLOY_VERSION__
 	 */
-	protected $translateFormat;
+	protected $filterFormat;
 
 	/**
 	 * The filter.
@@ -113,6 +113,7 @@ class CalendarField extends FormField
 		{
 			case 'maxlength':
 			case 'format':
+			case 'filterFormat':
 			case 'filter':
 			case 'timeformat':
 			case 'todaybutton':
@@ -122,7 +123,6 @@ class CalendarField extends FormField
 			case 'filltable':
 			case 'minyear':
 			case 'maxyear':
-			case 'translateFormat':
 				return $this->$name;
 		}
 
@@ -153,13 +153,11 @@ class CalendarField extends FormField
 			case 'showtime':
 			case 'filltable':
 			case 'format':
+			case 'filterFormat':
 			case 'filter':
 			case 'minyear':
 			case 'maxyear':
 				$this->$name = (string) $value;
-				break;
-			case 'translateFormat':
-				$this->$name = (bool) $value;
 				break;
 
 			default:
@@ -198,11 +196,33 @@ class CalendarField extends FormField
 			$this->singleheader = (string) $this->element['singleheader'] ? (string) $this->element['singleheader'] : 'false';
 			$this->minyear      = \strlen((string) $this->element['minyear']) ? (string) $this->element['minyear'] : null;
 			$this->maxyear      = \strlen((string) $this->element['maxyear']) ? (string) $this->element['maxyear'] : null;
-			$this->translateFormat = (string) $element['translateformat'] && ((string) $element['translateformat']) !== 'false';
 
 			if ($this->maxyear < 0 || $this->minyear > 0)
 			{
 				$this->todaybutton = 'false';
+			}
+
+			$translateFormat = (string) $this->element['translateformat'];
+
+			if ($translateFormat && $translateFormat !== 'false')
+			{
+				$showTime              = (string) $this->element['showtime'];
+
+				$lang  = Factory::getLanguage();
+				$debug = $lang->setDebug(false);
+
+				if ($showTime && $showTime !== 'false')
+				{
+					$this->format       = Text::_('DATE_FORMAT_CALENDAR_DATETIME');
+					$this->filterFormat = Text::_('DATE_FORMAT_FILTER_DATETIME');
+				}
+				else
+				{
+					$this->format       = Text::_('DATE_FORMAT_CALENDAR_DATE');
+					$this->filterFormat = Text::_('DATE_FORMAT_FILTER_DATE');
+				}
+
+				$lang->setDebug($debug);
 			}
 		}
 
@@ -218,27 +238,7 @@ class CalendarField extends FormField
 	 */
 	protected function getInput()
 	{
-		$user = Factory::getUser();
-
-		// Translate the format if requested
-		if ($this->translateFormat)
-		{
-			$showTime = (string) $this->element['showtime'];
-
-			$lang  = Factory::getLanguage();
-			$debug = $lang->setDebug(false);
-
-			if ($showTime && $showTime !== 'false')
-			{
-				$this->format = Text::_('DATE_FORMAT_CALENDAR_DATETIME');
-			}
-			else
-			{
-				$this->format = Text::_('DATE_FORMAT_CALENDAR_DATE');
-			}
-
-			$lang->setDebug($debug);
-		}
+		$user = Factory::getApplication()->getIdentity();
 
 		// If a known filter is given use it.
 		switch (strtoupper($this->filter))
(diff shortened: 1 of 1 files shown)


libraries/src/Form/Field/CalendarField.php
```diff
@@ -44,11 +45,19 @@ class CalendarField extends FormField
...
-	 * @var    integer
+	 * @var    string
 	 * @since  3.2
 	 */
 	protected $format;
 
+	/**
+	 * Whether format should be translated
+	 *
+	 * @var    bool
```
PhilETaylor:
```suggestion
	 * @var    boolean
```

Fedik: to late, I just pushed the same ;)

PhilETaylor:
I have tested this item :white_check_mark: successfully on eceec3598824e1491ec6c729bbc3ea05b6330984<br /><br />tested with offsets as well.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35222">issues.joomla.org/tracker/joomla-cms/35222</a>.</sub>

libraries/src/Form/Field/CalendarField.php
```diff
@@ -185,6 +198,7 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
 			$this->singleheader = (string) $this->element['singleheader'] ? (string) $this->element['singleheader'] : 'false';
 			$this->minyear      = \strlen((string) $this->element['minyear']) ? (string) $this->element['minyear'] : null;
 			$this->maxyear      = \strlen((string) $this->element['maxyear']) ? (string) $this->element['maxyear'] : null;
+			$this->translateFormat = ((string) $element['translateformat']) !== 'false';
```
joomdonation: If I'm not mistaken, the change here is not right.  Before, if a field does not have `translateformat` attribute defined, it won't be processed as translatable.

With this change, if a field does not have this attribute defined, it will be processed as translatable and it will be broken.

Fedik: this is what in joomla 3, unless I missing something

Fedik: hm, I wil change

joomdonation: It is not the same, I think. You can try to check with a field without translateformat attributed defined. In that case:

`$this->translateFormat` will be true.

But the original code to check is:

`if ($translateFormat && $translateFormat != 'false')` equals `false`.

Fedik: I have changed already, please check again ;)

joomdonation: Thanks. Should be fine now (although I have feeling that the line of code a big ugly :D). This form field is widely used, so we should test it carefully. Will try to test it on tomorrow if it is still needed, too late for me here now.

infograf768:
Has this been tested with Jalali calendar (Persian fa-IR, only available for now as 3.x pack)?

Fedik:
I tried, it works there. 
The date value is sent in "gregorian" for every language.

joomdonation:
@Fedik We would get error with Debug Language set to Yes. Basically, in your new code, we will have to process Debug same as how we do in getInput method https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/Field/CalendarField.php#L212-L229

If you don't mind, I want to clean up the code  abit more to avoid repeating code (for calculating format in getInput and filter method):

1. Define a new property filterFormat and remove translateFormat properly
2. Move the logic to process translate format into setup method (so that we won't have to repeat the logic in two separate methods)
3. Few other small cleanup

Here is the result https://github.com/joomla/joomla-cms/compare/4.0-dev...joomdonation:calendar_filter?expand=1 (to me, it is cleaner, but maybe it is just me :D

If you like that change, I made PR https://github.com/Fedik/joomla-cms/pull/9 to your branch.

If not, please handle the Debug language as I mentioned and I will test it.

Fedik:
well, I do not see language debug manipulation in J3, 
https://github.com/joomla/joomla-cms/blob/f521d3a259306748691d85504f1e440e23d0ad6e/libraries/src/Form/Form.php#L1316
and I do not like it :smile: 

But if something I will tell that it is you :wink:

libraries/src/Form/Field/CalendarField.php
```diff
@@ -332,51 +344,44 @@ protected function getLayoutData()
...
+		if ($this->filterFormat)
+		{
+			$value = DateTime::createFromFormat($this->filterFormat, $value)->format('Y-m-d H:i:s');
+		}
+
+		$app = Factory::getApplication();
+
 		// Get the field filter type.
-		$filter = (string) $this->element['filter'];
+		$filter = strtoupper((string) $this->element['filter']);
```
joomdonation: @Fedik Sorry, could you please remove `strtoupper` call here? We call it later on switch statement, so no need to call it here. After that, I'm ready for final testing :).

Fedik: done

joomdonation:
I have tested this item :white_check_mark: successfully on 567cbb0f0c83aa373ccec1c3bfd0cc77048f2687<br /><br />All good now. Thanks !<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35222">issues.joomla.org/tracker/joomla-cms/35222</a>.</sub>

joomdonation:
@PhilETaylor Could you please test it again?

PhilETaylor:
Im going to mark this as successful as it addresses the reported issue

However in testing I was able to break it - with insane values for the format in DATE_FORMAT_CALENDAR_DATETIME

```
DATE_FORMAT_CALENDAR_DATETIME="%m 2- %d     %Y %H:%M:%S"
```

If the date was today (`08-20-2021 04:47:41`) then the above format outputs `08 2- 03     2-2020 11:41:04` which is incorrect.

PhilETaylor:
I have tested this item :white_check_mark: successfully on 567cbb0f0c83aa373ccec1c3bfd0cc77048f2687<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35222">issues.joomla.org/tracker/joomla-cms/35222</a>.</sub>

PhilETaylor:
Another insane check - on the frontend if I use 

```
DATE_FORMAT_CALENDAR_DATETIME="%m-%d     MINE     %Y %H:%M:%S"
```

Then you cannot save - but again I think Im pushing the edges of what we should be supporting here.

joomdonation:
Yes. The language file need to provider correct value for these languages item.

Actually, the error happens if users enter random string into the input instead of select a date/time from the picker, but that could be addressed in separate PR

joomdonation:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35222">issues.joomla.org/tracker/joomla-cms/35222</a>.</sub>

PhilETaylor:
> Actually, the error happens if users enter random string into the input instead of select a date/time from the picker

I tried that and it handled it well, by removing my rubbish and replacing it with a "now" date/time string.

joomdonation:
> I tried that and it handled it well, by removing my rubbish and replacing it with a "now" date/time string.

Ah, Yes. A **happy strange**  that no error throws for that case. I thought there would be exception, happy to see that it does not happen. So all good :)

Fedik:
> However in testing I was able to break it - with insane values for the format in DATE_FORMAT_CALENDAR_DATETIME

The date constants `DATE_FORMAT_CALENDAR_DATETIME` and `DATE_FORMAT_FILTER_DATETIME` must be in sync.
```
DATE_FORMAT_CALENDAR_DATETIME="%m 2- %d     %Y %H:%M:%S"
DATE_FORMAT_FILTER_DATETIME="m 2- d     Y H:i:s"
```
and
```
DATE_FORMAT_CALENDAR_DATETIME="%m-%d     MINE     %Y %H:%M:%S"
DATE_FORMAT_FILTER_DATETIME="m-d     \M\I\N\E     Y H:i:s"
```

This probably a legacy from transition from mootools calendar to current calendar.
It is asking for refactoring, but well, not much easy task.

PhilETaylor:
No worries - like I said, I know I was pushing the extremes and this PR fixes the reported issue.

Bakual:
> This probably a legacy from transition from mootools calendar to current calendar.

Yep, I know for a fact. In J3 there was no way to change that at the time 😄 
If the new calendar also can use the PHP Date format, then a refactoring would be welcome. You could then use the filter format string and we keep the other for B/C until J5.0.

dgrammatiko:
> If the new calendar also can use the PHP Date format, then a refactoring would be welcome. 

@Bakual care to explain a bit further? I might contribute a new calendar based on ecmascript temporal and intl, but would be nice to know other devs preferences

Fedik:
@dgrammatiko it about using 2 diferent format at same time, currently:
`strftime` for calendar js https://www.php.net/manual/en/function.strftime.php for `DATE_FORMAT_CALENDAR_DATETIME`
`datetime` for calendar php https://www.php.net/manual/en/datetime.format.php for `DATE_FORMAT_FILTER_DATETIME`

If both will use `datetime` format, then it could be much simpler.
Or even something from php-intl for correct translations for a dates.

In theory  :wink:

PhilETaylor:
While all these calendar things are still fresh in your mind - maybe you all would like to review this issue from 2019 as well if you have a moment

https://github.com/joomla/joomla-cms/pull/24417

PhilETaylor:
I just had the same issue with the chinese simplified language pack, and this PR fixed that too.

wilsonge:
Thanks!

Issue #35180, opened 2021-08-17
main menu disappears when using "system cache" -> "conservative caching"


BlackBrix:
### Steps to reproduce the issue
- use 4.0 release 2021-08-17
- use default template "Cassiopeia"
- use firefox browser 91.x
- go to admin backend and activate "system cache" -> "conservative caching"
- go to frontend again and reload the page (F5 or CTRL-F5)
- the main menu disappears after some clicks on F5 or CTRL-F5
- ....
- go to admin backend again and deactivate "system cache"
- go to frontend again and reload the page (F5 or CTRL-F5)
- the main menu appears correctly again immediately

### Expected result
- see above

### Actual result
- see above

### System information (as much as possible)
server: --------------------
- "shared" hosting (using Plesk-Onyx)*
- server: Apache*
- sapi_name: cgi-fcgi
- PHP 7.4.22
- mySQL 5.7.33
- PHP-setting: output_buffering: true (I can not change that due to shared hosting)
- *(Plesk-Onyx-Feature: "intelligent nginx proxy" is active for static files)
- see attachment:
[systeminfo-2021-08-17T15 03 56+02 00.txt](https://github.com/joomla/joomla-cms/files/7000149/systeminfo-2021-08-17T15.03.56%2B02.00.txt)

client: -------------------
- Windows 10 professional 64bit
- firefox 91.x

### Additional comments
- "system cache" -> "conservative caching" works perfect on the same server with Joomla 3.10.x

PhilETaylor:
I am unable to replicate this at all.

Tested with output_buffering on/off/4096

Tested with PHP 7.4.22 & 8.09

Even with firefox 91, although the browser is irrelevant for this kind of issue.

brianteeman:
Also see #34568

PhilETaylor:
> Also see #34568

Strange, but cant fix what I cannot replicate

BlackBrix:
@PhilETaylor :
I can give you the backend credentials of my joomla 4.0 installation, 
so you can replicate it by yourself
(you can not damage anything because it's just a test installation of Joomla 4.x)

I'm almost sure that it has to do with this kind of shared hosting ...

richard67:
I don't think it's related, but who knows, so it's worth a try.

@BlackBrix Do you use the htaccess.txt file shipped with Joomla 4? With "use" i mean having copied it to ".htaccess".

If this is the case, edit .htaccess and go to the end. There is following section:

```
[...]

Issue #34568, opened 2021-06-19
[4.0 RC2] When I activate the cache, I lose several modules


Magnytu2:
### Steps to reproduce the problem
Create a site with Joomla 4RC2 with Cassiopeia. Activate the system cache.

### Expected result
No change in Frontend

### Current result
The menus and several modules disappear. Sometimes just the borders of the modules disappear. This is the case for modules in bottom position a or b. https://joomla-4.fr
![Capture 073 - Accueil - joomla-4 fr](https://user-images.githubusercontent.com/26197929/122642312-21178600-d10a-11eb-8906-b0b830f8fcf3.jpg)

### System information (as much as possible)
Joomla 4 RC2 Php 7.4.2

particthistle:
Just encountered similar issue with an install, testing with builds through to J4 RC5-Dev 14 July 2020. Site is PHP 7.4.19 on cpanel.

With cache off, menu modules and other modules behave normally.

With cache on (progressive or conservative) you may get the pages on the site to load with the menus a few times, but then when you return to a page, you get a stripped down page without various modules.

In my case, the modules that disappear are the main menu (Cassiopeia > menu position) and the footer menu (Cassiopeia > footer position).

Setting the menu modules individually to "No caching" appears to restore the functionality, but in most cases menus would be content that should be able to be cached as they're typically not dynamically generated.

Turning on Debug, I noticed that when the problem is occurring, queries drop from 27 to 8, with bulk of them being the queries loading the module and then the menu.

With Cache off:
![image](https://user-images.githubusercontent.com/5515866/125708022-29f721cc-efe2-4705-954e-ea49d709555e.png)

With Cache on:
![image](https://user-images.githubusercontent.com/5515866/125708044-303afd21-a435-432d-b605-2e4f4ede3c45.png)

richard67:
See also #35180 .

@Magnytu2 Can you still reproduce the issue? If so, check my post at the end of issue #35180 . I don't really think it helps, but I'm not sure, so it is worth trying.

Magnytu2:
I still have the problem. If I activate progressive cache, the menu disappears.
My site and now under Joomla! 4.0 today, with the same dysfunction. See here> https://joomla-4.fr

richard67:
@Magnytu2 Then check if my comment here helps: https://github.com/joomla/joomla-cms/issues/35180#issuecomment-900342927

Magnytu2:
The problem is the same after removing the requested rows.

[...]

PR #19416, opened 2018-01-20
[4.0] [RFC] HtmlDocument::countModules improve, allow to count the modules with the content only

Fedik:
Pull Request for Issue # there somewhere in history, hard to find  .

### Issue

`$this->countModules('foobar')` return number of active modules in the position, but do not check whether module has any content, or not (so it will be not visible)

This is one of most annoying things in Joomla! templating to me.
It can happen with 3d modules for display some associated lists (or similar). `HtmlDocument::countModules` still count this module as "exists" even if associated list is empty, and so module will not be displayed.

### The idea, the fix

The idea is to allow to count only the module which actually has a content.
I have added flag `$withContentOnly` to `HtmlDocument::countModules` which allow to do it.

And split `ModuleHelper::renderModule` to 2 part  `ModuleHelper::renderRawModule` (render module content)  and `ModuleHelper::renderModule` (render module Chrome, trigger the events etc). So `HtmlDocument::countModules` will trigger rendering only for module content `ModuleHelper::renderRawModule` and count only the modules which actually have the content.

### Testing Instructions

It a bit tricky.

First make override to mod_custom, call it `empty.php`, and remove all content from it. This will imitate the module without content.

Then create 2 mod_custom modules:
"module1" which use override "empty", at position `foobar`;
"module2" with some random text and without any override, at position `foobar`;

In the index.php of the template add next code:
```php
var_dump($this->countModules('foobar', true)); // Count only modules which has content
var_dump($this->countModules('foobar')); // Count all modules in selected position
```

### Expected result
```
int 1 
int 2 
```

### Actual result
```
int 2
int 2 
```

### Documentation Changes Required

yeap

PR #35225, opened 2021-08-19
Fix caching when HTMLDocument::countModules is used

Fedik:
Pull Request for Issue #35180 and #34568 .

### Summary of Changes
Fix modules caching when `$this->countModules('position', true)` is used in a template.

### Testing Instructions
Please follow #35180

Some modules have broken style

All works 

none

reference #19416

### DIFF
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -693,12 +693,7 @@ class HtmlDocument extends Document
 
 		foreach ($modules as $module)
 		{
-			if (empty($module->contentRendered))
-			{
-				// Preserve content if loaded from cache
-				$module->content = $renderer->render($module, ['contentOnly' => true]);
-				$module->contentRendered = true;
-			}
+			$renderer->render($module, ['contentOnly' => true]);
 
 			if (trim($module->content) !== '')
 			{
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -695,6 +695,8 @@ class HtmlDocument extends Document
 		{
 			if (empty($module->contentRendered))
 			{
+				// Use style "raw", that will render only the content of the module, and precache it, see ModuleHelper::renderRawModule
+				// This style only for internal use, and should not be used within template markup.
 				$renderer->render($module, array('style' => 'raw'));
 			}
 
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -693,7 +693,10 @@ class HtmlDocument extends Document
 
 		foreach ($modules as $module)
 		{
-			$renderer->render($module, ['contentOnly' => true]);
+			if (empty($module->contentRendered))
+			{
+				$renderer->render($module, ['contentOnly' => true]);
+			}
 
 			if (trim($module->content) !== '')
 			{
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -695,9 +695,9 @@ class HtmlDocument extends Document
 		{
 			if (empty($module->contentRendered))
 			{
-				// Use style "raw", that will render only the content of the module, and precache it, see ModuleHelper::renderRawModule
-				// This style only for internal use, and should not be used within template markup.
-				$renderer->render($module, array('style' => 'raw'));
+				// Preserve content if loaded from cache
+				$module->content = $renderer->render($module, ['countOnly' => true]);
+				$module->contentRendered = true;
 			}
 
 			if (trim($module->content) !== '')
--- a/libraries/src/Document/HtmlDocument.php
+++ b/libraries/src/Document/HtmlDocument.php
@@ -696,7 +696,7 @@ class HtmlDocument extends Document
 			if (empty($module->contentRendered))
 			{
 				// Preserve content if loaded from cache
-				$module->content = $renderer->render($module, ['countOnly' => true]);
+				$module->content = $renderer->render($module, ['contentOnly' => true]);
 				$module->contentRendered = true;
 			}
 
--- a/libraries/src/Document/Renderer/Html/ModuleRenderer.php
+++ b/libraries/src/Document/Renderer/Html/ModuleRenderer.php
@@ -92,7 +92,10 @@ class ModuleRenderer extends DocumentRenderer
 			$cacheparams->methodparams = array($module, $attribs);
 			$cacheparams->cachesuffix = $attribs['contentOnly'] ?? false;
 
-			return ModuleHelper::moduleCache($module, $params, $cacheparams);
+			// It need to be done here because the cache controller does not keep reference to the module object
+			$module->content = ModuleHelper::moduleCache($module, $params, $cacheparams);
+
+			return $module->content;
 		}
 
 		return ModuleHelper::renderModule($module, $attribs);
--- a/libraries/src/Document/Renderer/Html/ModuleRenderer.php
+++ b/libraries/src/Document/Renderer/Html/ModuleRenderer.php
@@ -90,7 +90,7 @@ class ModuleRenderer extends DocumentRenderer
 			$cacheparams->class = ModuleHelper::class;
 			$cacheparams->method = 'renderModule';
 			$cacheparams->methodparams = array($module, $attribs);
-			$cacheparams->cachesuffix = $attribs['countOnly'] ?? false;
+			$cacheparams->cachesuffix = $attribs['contentOnly'] ?? false;
 
 			return ModuleHelper::moduleCache($module, $params, $cacheparams);
 		}
--- a/libraries/src/Document/Renderer/Html/ModuleRenderer.php
+++ b/libraries/src/Document/Renderer/Html/ModuleRenderer.php
@@ -84,26 +84,15 @@ class ModuleRenderer extends DocumentRenderer
 
 		if ($params->get('cache', 0) == 1 && Factory::getApplication()->get('caching') >= 1 && $cachemode !== 'id' && $cachemode !== 'safeuri')
 		{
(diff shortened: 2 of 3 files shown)


libraries/src/Helper/ModuleHelper.php
```diff
@@ -165,6 +165,8 @@ public static function renderModule($module, $attribs = array())
 		// Render the module content
 		static::renderRawModule($module, $params, $attribs);
 
+		// Check for a "raw" style: render module content only.
+		// This style only for internal use, and should not be used within template markup.
```
PhilETaylor: CS needs /* */ for multi line comments

libraries/src/Document/Renderer/Html/ModuleRenderer.php
```diff
@@ -84,14 +84,28 @@ public function render($module, $attribs = array(), $content = null)
...
 			$cacheparams->cachemode = $cachemode;
 			$cacheparams->class = ModuleHelper::class;
 			$cacheparams->method = 'renderModule';
 			$cacheparams->methodparams = array($module, $attribs);
+			$cacheparams->cachesuffix = $isRawStyle ? 'style-raw' : '';
+
+			$result = ModuleHelper::moduleCache($module, $params, $cacheparams);
+
+			// For a raw style set the content to the module, so it will not be rendered again next time, see ModuleHelper::renderRawModule
+			// It need to be done here because the cache controller does not keep reference to the module object
```
PhilETaylor: CS needs /* */ for multi line comments

libraries/src/Document/HtmlDocument.php
```diff
@@ -695,6 +695,8 @@ public function countModules(string $positionName, bool $withContentOnly = false
 		{
 			if (empty($module->contentRendered))
 			{
+				// Use style "raw", that will render only the content of the module, and precache it, see ModuleHelper::renderRawModule
+				// This style only for internal use, and should not be used within template markup.
```
PhilETaylor: CS needs /* */ for multi line comments

PhilETaylor:
Testing instructions - MAKE SURE YOU CLEAR YOUR CACHE after applying this PR if you already have a warm PR. I didn't and I still saw a broken chrome until I did clear the cache manually, after that it was fine.

PhilETaylor:
I have tested this item :white_check_mark: successfully on a73cfb2809debc511ced8d0b0fe6945a159fa9ad<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35225">issues.joomla.org/tracker/joomla-cms/35225</a>.</sub>

libraries/src/Helper/ModuleHelper.php
```diff
@@ -218,7 +213,8 @@ public static function renderModule($module, $attribs = array())
 		// If the $module is nulled it will return an empty content, otherwise it will render the module normally.
 		$app->triggerEvent('onRenderModule', array(&$module, &$attribs));
 
-		if ($module === null || !isset($module->content))
+		// Don't apply chromes if we only need the content
+		if ($module === null || !isset($module->content) || !empty($attribs['contentOnly']))
```
Fedik: @HLeithner I think this will make a bug, the `onRenderModule` will be called twice,
first time when countModule, and second time when render chrome style

now I remember why I made early return, originally

libraries/src/Document/HtmlDocument.php
```diff
@@ -695,7 +695,9 @@ public function countModules(string $positionName, bool $withContentOnly = false
 		{
 			if (empty($module->contentRendered))
 			{
-				$renderer->render($module, array('style' => 'raw'));
+				// Preserve content if loaded from cache
+				$module->content = $renderer->render($module, ['contentOnly' => true]);
+				$module->contentRendered = true;
```
Fedik: hmhm, this flag already set by `renderRawModule`, 
it only may need to set when cache enabled, after ModuleHelper::moduleCache

libraries/src/Helper/ModuleHelper.php
brianteeman: and wasnt that the purpose of the original PR to prevent it happening twice?

libraries/src/Document/HtmlDocument.php
```diff
@@ -693,10 +693,7 @@ public function countModules(string $positionName, bool $withContentOnly = false
 
 		foreach ($modules as $module)
 		{
-			if (empty($module->contentRendered))
-			{
-				$renderer->render($module, array('style' => 'raw'));
-			}
+			$renderer->render($module, ['contentOnly' => true]);
```
Fedik: I still would do:
```
if (empty($module->contentRendered))
{
	$renderer->render($module, ['contentOnly' => true]);
}
```
To avoid extra call of renderer

libraries/src/Document/Renderer/Html/ModuleRenderer.php
```diff
@@ -90,8 +90,12 @@ public function render($module, $attribs = array(), $content = null)
 			$cacheparams->class = ModuleHelper::class;
 			$cacheparams->method = 'renderModule';
 			$cacheparams->methodparams = array($module, $attribs);
+			$cacheparams->cachesuffix = $attribs['contentOnly'] ?? false;
 
-			return ModuleHelper::moduleCache($module, $params, $cacheparams);
+			// It need to be done here because the cache controller does not keep reference to the module object
+			$module->content = ModuleHelper::moduleCache($module, $params, $cacheparams);
```
Fedik, line 96: and set `$module->contentRendered = true` here, in the same reason, becasue cach controller does not keep reference

libraries/src/Helper/ModuleHelper.php
HLeithner: Would this be a problem?
I see it as a bigger problem if you don't call it for counting.

think of a plugin that remove the content if the content is '< div >< /div >'.

or a plugin that adds content when a content for a module is empty.

in both cases countModules would be wrong.

on the other side calling the trigger twice of the same module shouldn't be a problem because I can have the same module on multiple positions and the module and plugin would be called on every position if I'm not wrong.

libraries/src/Document/Renderer/Html/ModuleRenderer.php
HLeithner: done

libraries/src/Helper/ModuleHelper.php
HLeithner: reverted

libraries/src/Document/HtmlDocument.php
HLeithner: reverted

Fedik:
Ready for testing again.

@PhilETaylor please re test :)

maikol-ortigueira:
I have tested this item :white_check_mark: successfully on 6fad521d6a0b4c1d2bf9a83af60ca02bd94c77f0<br /><br />I have tested the patch and it works perfectly. I was indeed getting the issue in both Firefox 91.x and Chrome 92.0.4515.131.

I did the test on a local installation with the following configuration:

Server: Devilbox
Joomla version: 4.0.1-dev
PHP version: 7.4.13
Web server: Apache/2.4.5
Database version: Maria DB 10.3.27
Browser: Firefox 91.0 and Google Chrome 92.0.4515.131

Thank you very much.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35225">issues.joomla.org/tracker/joomla-cms/35225</a>.</sub>

richard67:
@PhilETaylor Could you test this one again? Thanks in advance.

richard67:
@Fedik I think this fixes also issue #34568 , right?

PhilETaylor:
I have tested this item :white_check_mark: successfully on 6fad521d6a0b4c1d2bf9a83af60ca02bd94c77f0<br /><br />Seems to work. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35225">issues.joomla.org/tracker/joomla-cms/35225</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35225">issues.joomla.org/tracker/joomla-cms/35225</a>.</sub>

Fedik:
hm, yes, looks like

libraries/src/Helper/ModuleHelper.php
```diff
@@ -165,7 +165,8 @@ public static function renderModule($module, $attribs = array())
 		// Render the module content
 		static::renderRawModule($module, $params, $attribs);
 
-		if (!empty($attribs['style']) && $attribs['style'] === 'raw')
+		// Return early if only the content is required
+		if (!empty($attribs['contentOnly']))
```
wilsonge, line 169: But doesn't this mean we no longer call plugin events if contentOnly is set to true? I don't have a great feel of the code here - but something smells wrong.

HLeithner: That's right but the fear for calling the event twice seems higher...

Fedik: @wilsonge  "contentOnly" supposed to render only module content.
And our Module events called when rendering of module chrome, 
this way it works in old joomla and this true for j4, so here is no b/c.

And as @HLeithner already wrote, if we do `onModuleRender` before this check, then we will have double/tripple event trigger when someone use `HTMLDocuemnt::countModules()`

If need we can introduce another event that triggered from `ModuleHelper::renderRawModule()`, kind of `onAfterModuleContentRender`, but this for future. Not very good to add a new feature in 4.0.1 ;)

Issue #35216, opened 2021-08-18
[4.0.0]  Core LADP Plugin - The option "host" with value null is expected to be of type "string", but is of type "null".


ChristineWk:
### Steps to reproduce the issue
The following was reported by a Forum de user: https://forum.joomla.de/thread/14122-fehlermeldung-der-admin-anmeldung/?postID=91310#post91310

### Expected result

Maybe possibly PR for J 4.0.1
### Actual result

After logging back in to the backend, the following message came up: 

Call Stack:

    The option "host" with value null is expected to be of type "string", but is of type "null".

Call stack
#	Function	Location
1	()	JROOT/libraries/vendor/symfony/options-resolver/OptionsResolver.php:1059
2	Symfony\Component\OptionsResolver\OptionsResolver->offsetGet()	JROOT/libraries/vendor/symfony/options-resolver/OptionsResolver.php:924
3	Symfony\Component\OptionsResolver\OptionsResolver->resolve()	JROOT/libraries/vendor/symfony/ldap/Adapter/AbstractConnection.php:30
4	Symfony\Component\Ldap\Adapter\AbstractConnection->__construct()	JROOT/libraries/vendor/symfony/ldap/Adapter/ExtLdap/Adapter.php:41
5	Symfony\Component\Ldap\Adapter\ExtLdap\Adapter->getConnection()	JROOT/libraries/vendor/symfony/ldap/Ldap.php:40
6	Symfony\Component\Ldap\Ldap->bind()	JROOT/plugins/authentication/ldap/ldap.php:139
7	PlgAuthenticationLdap->onUserAuthenticate()	JROOT/libraries/src/Authentication/Authentication.php:177
8	Joomla\CMS\Authentication\Authentication->authenticate()	JROOT/libraries/src/Application/CMSApplication.php:819
9	Joomla\CMS\Application\CMSApplication->login()	JROOT/libraries/src/Application/AdministratorApplication.php:384
10	Joomla\CMS\Application\AdministratorApplication->login()	JROOT/administrator/components/com_login/src/Controller/DisplayController.php:80
11	Joomla\Component\Login\Administrator\Controller\DisplayController->login()	JROOT/libraries/src/MVC/Controller/BaseController.php:730
12	Joomla\CMS\MVC\Controller\BaseController->execute()	JROOT/libraries/src/Dispatcher/ComponentDispatcher.php:146
[...]

PR #35233, opened 2021-08-19
[4] Dont attempt LDAP connection if no host provided

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/35216

### Summary of Changes

It is possible to publish the LDAP Authentication plugin by just clicking a toggle in the plugin list view.

If you do this you can lock yourself out of your site, because if you try to login then a connection is attempted to LDAP, but cause there is no LDAP server host provided we get a fatal type error deep in Symfony Options Resolver.

### Testing Instructions

Login to Joomla 4 admin
System -> Plugins -> toggle "Authentication - LDAP" 

Fatal Error.

`The option "host" with value null is expected to be of type "string", but is of type "null".`

You can still login with users already in the users table.

### DIFF
--- a/plugins/authentication/ldap/ldap.php
+++ b/plugins/authentication/ldap/ldap.php
@@ -42,7 +42,7 @@ class PlgAuthenticationLdap extends CMSPlugin
 		{
 			return false;
 		}
-		
+
 		// For JLog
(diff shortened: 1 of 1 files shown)


wilsonge:
Thanks!

PR #35234, opened 2021-08-19
Fix error when Event argument is null

alexandreelise:
ACTUAL: when trying to call getAllPropertiesToRender() public method or the getAllRelationsToRender() public method on the OnGetApiFields Event  in a custom system plugin in Joomla! 4. If the fields argument is null or the relations argument is null it throws a TypeError because the returning value of both of theses methods should be array.

SUGGESTION: A quick fix to that is to add a null coalescing operator with an empty array as a fallback like so ?? [] just after getting the argument

EXPECTED: Using an empty array as fallback rather than null because otherwise it throws an TypeError


alexandreelise:
Hi @alikon can you please review this one when you have some time. Take care. I am trying to get involved in core code. Tasting the delicious codebase of Joomla! 4.

wilsonge:
Why is your function for these returning null in the plugin? I think that's the problem.

alexandreelise:
@wilsonge Even without considering my plugin. When debugging the code with XDebug 3 I might have spotted a fishy behaviour leading to a TypeError. Because if we look at the code closely the fields argument and the relations arguments when not set are requiring to be arrays. Hence the TypeError when they are not defined because the default is null when calling getArgument('myarg') without specifiying anything. Just look in the base class you will see it.

alexandreelise:
Thanks @chmst for taking your time to have a look at this pull request. Take care and have a great day Super Joomler.

PR #35243, opened 2021-08-19
Add PHP6|7|8 to file extensions to scan for short tags

PhilETaylor:
@joomla/security @SniperSister 

Keeping our list up to date and inline with the `self::FORBIDDEN_FILE_EXTENSIONS` in the same class

Yes, yes, I know I know, there was never PHP 6 released :)


wilsonge:
Thanks!

SniperSister:
Good catch, thanks @PhilETaylor

PR #35260, opened 2021-08-20
[4] Dont use native constant that might be unavailable.

PhilETaylor:
Code review

`PASSWORD_ARGON2ID` php native constant is only available from PHP 7.3.0 onwards whereas Joomla 4 has to support 7.2.5

PHP may also be compiled without argon support and therefore this constant would not be available.

Just like stated (repeatedly) in [UserHelper](https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/User/UserHelper.php)

https://github.com/joomla/joomla-cms/blob/55b02c923b65dc9016a3f82671bb9bd9177e2e4d/libraries/src/User/UserHelper.php#L46

### DIFF
--- a/libraries/src/Authentication/Password/Argon2idHandler.php
+++ b/libraries/src/Authentication/Password/Argon2idHandler.php
@@ -22,6 +22,9 @@ class Argon2idHandler extends BaseArgon2idHandler implements CheckIfRehashNeeded
 	/**
 	 * Check if the password requires rehashing
 	 *
+	 * Note: PHP's native `PASSWORD_ARGON2ID` constant is not used as PHP may be compiled without this constant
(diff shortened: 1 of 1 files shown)


PhilETaylor:
I see my stalker has returned.

Screenshot 2021-08-20 at 18 43 20

Fedik:
You have to use 
```
defined('BLABLA') ? BLABLA : FALLBACK_BLABLA
```
But I cannot say anything about PASSWORD_ARGON2ID and possible fallbacks

PhilETaylor:
I give up - really I do. Im done. Back to creaming, leeching and sponging of the project and let toxic Teeman rule his project

PR #35261, opened 2021-08-20
[4] Comparison may cause unexpected type coercion

PhilETaylor:
Code review

Comparison document.adminForm.boxchecked.value==0 may cause unexpected type coercion according to phpStorm and JSLint. 

On evaluating `document.adminForm.boxchecked.value` on any list view you can see the value is actually a string when returned, so it makes sense to be evaluated as one


layouts/joomla/toolbar/batch.php
```diff
@@ -23,7 +23,7 @@
 $message = "{'error': [Joomla.Text._('JLIB_HTML_PLEASE_MAKE_A_SELECTION_FROM_THE_LIST')]}";
 $alert = "Joomla.renderMessages(" . $message . ")";
 ?>
-<button<?php echo $id; ?> type="button" data-bs-toggle="modal" onclick="if (document.adminForm.boxchecked.value==0){<?php echo $alert; ?>}else{document.getElementById('collapseModal').open(); return true;}" class="btn btn-primary">
+<button<?php echo $id; ?> type="button" data-bs-toggle="modal" onclick="if (document.adminForm.boxchecked.value==="0"){<?php echo $alert; ?>}else{document.getElementById('collapseModal').open(); return true;}" class="btn btn-primary">
```
joomdonation: Does not work. You need to change it to use single quote instead of double quote. Maybe an extra space before and after === would make the code look nicer.

PhilETaylor: Worked for me else I would not have PRed, no worries - changed as requested.

joomdonation: How did you test it? I don't see where it is used in Joomla 4 core code, so I had to test by applied the change and test it using Joomla 3 batch button. And I'm glad you made a real test instead of just code review :D.

joomdonation:
I have tested this item :white_check_mark: successfully on 67e6a10c4dfbd717bbbbefb265d31a0728de7231<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35261">issues.joomla.org/tracker/joomla-cms/35261</a>.</sub>

layouts/joomla/toolbar/batch.php
PhilETaylor: Code review and by type `document.adminForm.boxchecked.value === '0'` into the JS Console of your browser while on a list page before and after clicking a checkbox of an item in a list view. 

Screenshot 2021-08-22 at 13 18 09

joomdonation: Try to perform a real test and you will see an error on browser console 

**Uncaught SyntaxError: Unexpected end of input**

PhilETaylor: I find that hard to believe. All I did was change `==` to `===` as advised by the lint tools.

PhilETaylor: probably needs all this doing too 

Screenshot 2021-08-22 at 13 24 05

joomdonation: > I find that hard to believe. All I did was change == to === as advised by the lint tools.

Test it yourself and you will see :). Please note that there is double quote surrounded by onclick already, that's the reason we have to change that to single quote. Never mind, we should not waste time to discuss small thing like that.

PhilETaylor: ah that makes sense now - thanks. Im just trying to get phpStorm to stop nagging. The less issues it nags about the more chance we have of seeing real issues that it brings to attention - the whole purpose of using a decent IDE.

PR #35267, opened 2021-08-20
Example php-cs-fixer action

PhilETaylor:
add GitHub action as a test.


PhilETaylor:
PR needs to be open to see the checks.

PR #32896, opened 2021-03-28
[4.0] Redo the com_templates

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

One view to rule them all.
- Removes the 2 different entry layouts for the templates (template, style)
- Introduces the styling from Konshu template
- Fixes all the menus
- Brings the install logic from Konshu

This PR largely is the adoption of the work that was done by @kawshar and his team in their backend template. So all the kudos should go to them, I'm just bringing their code to the current branch (with their permissions)

### Testing Instructions

This is part of a wider rewrite of com_templates, probably has some a11y issues but should be functionally on par with the previous state (actually should be better as everything is one click away)

### Expected result AFTER applying this Pull Request

Screenshot 2021-03-28 at 17 39 31

Screenshot 2021-03-28 at 17 39 48

Screenshot 2021-03-28 at 17 43 45

Install template with one click
Screenshot 2021-03-28 at 19 00 28

Screenshot 2021-03-28 at 19 00 40

### Documentation Changes Required

@brianteeman can you review the a11y and the strings here?

PR #35281, opened 2021-08-21
[4] Move JS file to the build folder

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Change the move to top button to use Intersection observer instead onScroll event (perf)
- move the .js to the build folder and convert it to ES20xx
- use type=module for the script

IntersectionObserver Refs:
- MDN: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API 
- CanIUse: https://caniuse.com/intersectionobserver

### Testing Instructions
Fetch the PR
run `npm install`
Check that the Cassiopeia move to the top button still works as expected

No this is just a rewrite of the script with modern JS, the functionality is the same

@wilsonge the js file although it's in the media folder **IS NOT OVERRIDDABLE** but it's just a step towards converting the Cassiopeia to parent template (eg supporting child templates)


richard67:
I'll care later with another PR for the files deletion on update in script.php if this PR gets merged, so no need to do it here. PR is fine as is regarding that.

dgrammatiko:
@HLeithner care to explain why the thumbs down?

HLeithner:
I think it's wrong to split the template folder across the cms, it's already hard for new developer to create a template and it doesn't makes it easier.

Also I think making core templates a parent template would make us problems at a later point in time. because it means we can never remove it. Beside that we have no b/c promise on templates which makes it useless to create a child of it.

Beside that based on our B/C policy it says all JS functions and classes are covered, I might be wrong but I think you change them in this pr. (don't know why the JS B/C policy is more strict then all others in the cms).

dgrammatiko:
> it's already hard for new developer to create a template 

So it's fine to use ES5 javascript...

> because it means we can never remove it

False assumption. You can't also uninstall a template in use right now. But uninstalling all child tmpls then you can uninstall the parent

> Beside that based on our B/C policy it says all JS functions and classes are covered,

There's NO B/C break here, I just wrote the same thing with modern code...

HLeithner:
don't care if es5 (no idea why we ship it) or es6, it's about the fact that it's hard for a beginner to understand joomla templates.
Why do I have some files in the media folder? I mean everything was in the templates folder the last 16 years and now I have to put something into another folder?! I simple can't copy paste a template without any benefit (or at least I don't see any).

It about we es core does remove the template like we remove j3 templates (which I think is wrong but nobody had the time or the willing to adapt it to j4).

dgrammatiko:
> It about we es core does remove the template like we remove j3 templates

Still, you can check if there are child templates and act accordingly, or do it the Jway and remove everything

>  Why do I have some files in the media folder?

Well, ALL other extensions have their static assets in the media folder. Also, this is about security, locking down everything apart from media/images and couple .php is way easier if ALL static files are in one place. This was explained time and time again...

> I mean everything was in the templates folder the last 16 years and now I have to put something into another folder?

You can still do it the old way, it's not broken...

brianteeman:
Parts of cassiopeia are already in the media folder - the horse has already bolted on that

(not commenting either way about if its good or bad)

brianteeman:
> It about we es core does remove the template like we remove j3 templates (which I think is wrong but nobody had the time or the willing to adapt it to j4).

I don't remember ever seeing an issue about that.

HLeithner:
> Parts of cassiopeia are already in the media folder - the horse has already bolted on that
> 
> (not commenting either way about if its good or bad)

Doesn't matter, he asked me why I gave thumbsdown.
And a error can be reverted. 

I think it's wrong to make the template more completed. Not more not less. It makes it harder for new people and for veterans.

I see no benefit also for security it doesn't matter because you can't looked down the template folder for the same reason none of our folders is locked down.

HLeithner:
> > It about we es core does remove the template like we remove j3 templates (which I think is wrong but nobody had the time or the willing to adapt it to j4).
> 
> I don't remember ever seeing an issue about that.

So you think there are 0 people wondering why there template doesn't exists anymore after j4 upgrade?

brianteeman:
> > > It about we es core does remove the template like we remove j3 templates (which I think is wrong but nobody had the time or the willing to adapt it to j4).
> > 
> > 
> > I don't remember ever seeing an issue about that.
> 
> So you think there are 0 people wondering why there template doesn't exists anymore after j4 upgrade?

No I am simply saying that I don't remember seeing an issue about removing j3 templates OR an issue about converting protostar to J3

RickR2H:
Never mind me asking... But, do we have child templates in J4???<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35281">issues.joomla.org/tracker/joomla-cms/35281</a>.</sub>

dgrammatiko:
> But, do we have child templates in J4??

Partially. The fundamentals are already there since https://github.com/joomla/joomla-cms/pull/30192 but the behaviour is not yet exposed to the UI, although there's already some work done here: https://github.com/joomla/joomla-cms/pull/32896

RickR2H:
@dgrammatiko Thanks for the feedback!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35281">issues.joomla.org/tracker/joomla-cms/35281</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on 2358d61e5eb0d01d4238de1cf0eabd6c1222a92d<br /><br />Tested the patch and it works. Maybe we need documentation changes as the /js folder is now missing in the Cassiopeia template. It now also needs to be created when we want to use the user.js file.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35281">issues.joomla.org/tracker/joomla-cms/35281</a>.</sub>

build/media_source/templates/site/cassiopeia/js/template.es6.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * @package     Joomla.Site
+ * @subpackage  Templates.Cassiopeia
+ * @copyright   (C) 2017 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ * @since       4.0.0
```
PhilETaylor:
```suggestion
 * @since       __DEPLOY_VERSION__
```

dgrammatiko: I'm not quite sure that the JS/SCSS/CSS files are treated by the build tools (definitely not the npm as I wrote them and there's no such thing for `__DEPLOY_VERSION__` as there's no ~versioning~ since tag in the static files)

PhilETaylor: Eitherway, its not since 4.0.0 anyway ;-)

dgrammatiko: Ok so I removed the extra parts to follow the pattern of the rest JS files

dgrammatiko:
> I think it's wrong to make the template more completed. Not more not less. It makes it harder for new people and for veterans.

@HLeithner you are totally wrong here. The only things that changed for supporting Child templates are:
- Couple more lines in the XML (namely: namespace and inheritable)
- The files portion of the XML shouldn't include the css/js/images/other static folders
- the static folders should be inserted in a media section pointing in a path like `templates/{administrator / site}/templateName`
- the static files should be stored in the media folder
- All the overrides of the static files will work as before with only twist that the files should be placed in the media folder (see my previous 2 points)

ALL THESE COULD BE DONE IN THE BUILD STEP OF THE TEMPLATE. I will provide a simple one line command (`npx convert-joomla-template`) that will repackage an existing template to meet these requirements (moving folders to media and update the XML). Although most devs already have their own tools and this won't be a problem.
Also my work on the com_templates really makes all these changes completely transparent to non dev users.

So my question is then: why do you thing that the 15 years old way of doing templates cannot or should not be improved? (especially when the whole point is primary a security concern, templates should not be exposed as entry points in any website, actually Joomla's entry points should be: `/index.php`, `/administrator/index.php`, `/administrator/com_joomlaupdate/restore.php`, `media`, `images`. Everything else should not be exposed, security 101...)

HLeithner:
@dgrammatiko you maybe be right with being more consistent and to remove the direct call to /templates/xxx.

I would really like to see a .htaccess with where only the mentioned core files are accessible (and core folders are forbidden) and can be moved out of of the public folder.

dgrammatiko:
> and can be moved out of of the public folder.

Baby steps...

HLeithner:
> > and can be moved out of of the public folder.
> 
> Baby steps...

It's not about steps it's more about having a plan/big picture.

dgrammatiko:
@HLeithner fwiw here's a tool for automating the conversion of the templates: https://github.com/dgrammatiko/convert-template

templates/cassiopeia/joomla.asset.json
```diff
@@ -53,9 +53,9 @@
       "name": "template.cassiopeia",
       "description": "The file containing the javascript for this template.",
       "type": "script",
-      "uri": "template.js",
+      "uri": "media/templates/site/cassiopeia/js/template.min.js",
```
HLeithner, line 56: I thought this is lookup automatically?

dgrammatiko: It is for the templates that support child templates, at the moment Cassiopeia doesn't support thus the hardcoded path here

dgrammatiko: You can check the https://github.com/joomla/joomla-cms/files/7235154/cassy_1.0.0.zip which is cassiopeia supporting child templates. There the paths are like `template.js`

HLeithner: so this pr mean is on a state were cassiopeia doesn't support child templates? and cassy supports it right? so at some point this will be changed back to template.js ?

dgrammatiko: That's **my** plan to move both templates to the new mode as soon as the UI exposes all the goodness that this offers. Of course, my plan is not necessarily the project plan thus I asked publically for an official response from the PT. It would be counterproductive for me to spend a considerable amount of time (tbh I already have) if something is not aligned with the project's own aims...

HLeithner: I know, discussion is in progress, (at least I said my opinion in the channel)

dgrammatiko: Nice, I know that few people are against the changes (or any changes) but it's for the project's own interest to move forward, namespacing + child templates is a good point to introduce changes (the good thing here is that these are not B/C breaks as the switches are already in place so both worlds could coexist for as long as you want them to)

dgrammatiko:
I'm closing this one as it will be fixed with https://github.com/joomla/joomla-cms/pull/35874

PR #35291, opened 2021-08-22
[4] Cleanup of restore.php - Remove references to non-joomla restorations

PhilETaylor:
Code Review

Remove references to non-joomla restorations

### DIFF
--- a/administrator/components/com_joomlaupdate/restore.php
+++ b/administrator/components/com_joomlaupdate/restore.php
@@ -3688,17 +3688,8 @@ class AKFactory
 			{
 				// In restore.php mode we have to exclude the restoration.php files
 				$moreSkippedFiles     = array(
-					// Akeeba Backup for Joomla!
-					'administrator/components/com_akeeba/restoration.php',
 					// Joomla! Update
 					'administrator/components/com_joomlaupdate/restoration.php',
-					// Akeeba Backup for WordPress
-					'wp-content/plugins/akeebabackupwp/app/restoration.php',
-					'wp-content/plugins/akeebabackupcorewp/app/restoration.php',
-					'wp-content/plugins/akeebabackup/app/restoration.php',
-					'wp-content/plugins/akeebabackupwpcore/app/restoration.php',
-					// Akeeba Solo
-					'app/restoration.php',
 				);
 				$config['skip_files'] = array_merge($config['skip_files'], $moreSkippedFiles);
 			}


administrator/components/com_joomlaupdate/restore.php
```diff
@@ -3688,17 +3688,8 @@ public static function &getUnarchiver($configOverride = null)
 			{
 				// In restore.php mode we have to exclude the restoration.php files
 				$moreSkippedFiles     = array(
-					// Akeeba Backup for Joomla!
-					'administrator/components/com_akeeba/restoration.php',
```
richard67, line 3692: Sure we can also remove this one? It looks as if it could be present if Akeeba is installed. All that restoration stuff is so obscure to me that's I don't dare to guess if it would be safer to leave it in or not.

PhilETaylor: Will you ever distribute `administrator/components/com_akeeba/restoration.php` in the Joomla Upgrade Zip file? No. Therefore you never need to exclude that. 

Remember restore.php is basically a forked version of kickstart from Akeeba - which had to handle restoring Akeeba Backup ZIP and JPA files. There is a LOT in this file that doesnt apply to Joomla. 

If we are ever going to make progress with refactoring it, we first need to clean it up in small stages, so none of us have heart attacks later on :)

dgrammatiko: @PhilETaylor it might be worth rewriting this file, rather than cherry-pick changes. If the rewrite comes with some well-defined tests then there are very few chances that things might be broken. There's code that can be simplified and code that is totally unusable there...

My 2c

PhilETaylor: > it might be worth rewriting this file, rather than cherry-pick changes

Correct. But I dont have the balls to do it all in one go. No one has. Except Nic. Its taken many years to even allow contributors to modify this file ;-) everyone was too scared to. 

Im going to clean it up and then work on it - and add some tests. But its an end to end test that will be the most use, not unit tests. 

There is a lot of code in conditional ifs that can be deleted - your are right.

dgrammatiko: > everyone was too scared to.

Write some unit tests, then there's nothing to worry about (I mean if the tests aren't fooBar).

The process is not even so hard to follow: d/l the zip and then extract it in the root overwriting existing files (ok oversimplified, but essentially that's what is happening)

PhilETaylor: > The process is not even so hard to follow: d/l the zip and then extract it in the root overwriting existing files (ok oversimplified, but essentially that's what is happening)

Correct. But nothing is that simple. 

LMAO... spoken like someone who has never extracted a zip on windows, linux, mac, on crappy web hosts, web hosts with insane configurations, web hosts with disabled functions. web hosts with maximum execution time limits etc... 

There is a reasons kickstart is currently 5000+ lines long :)

PhilETaylor: Im thinking of writing a new version (by cleaning up the existing version) and allowing it to be opt-in/optional so that it can be used for a few live/real upgrade processes before being made the default restore.php process. That way we minimise the risk.

dgrammatiko: > There is a reasons kickstart is currently 5000+ lines long :)

Removing support for JPA and the AES part (which iirc are not used anymore) will take you down 2-3000 locs. Refactoring the existing functions on the file/folder read/write/chmod etc needs tests that should be run on every OS (GitHub actions could do that by the way)

joomdonation:
I have tested this item :white_check_mark: successfully on 12e03381fd75382f8f41a4e26ea2ad59246d8198<br /><br />Makes sense. Also perform real test by updating Joomla nightly build with modified restore.php and everything still works OK.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35291">issues.joomla.org/tracker/joomla-cms/35291</a>.</sub>

administrator/components/com_joomlaupdate/restore.php
richard67: For me my initial question here is resolved, but I won't mark the conversation as resolved so the interesting discussion above will not be hidden.

PhilETaylor: It's long overdue for complete refactoring for sure but there is also much to be said for incremental clean up - we all know that small specific Prs get tested more readily and merged quicker than one huge refactor 

Remember if we get it wrong - we brick Joomla upgrade forever. It's a huge responsibility.

brianteeman: > Remember if we get it wrong - we brick Joomla upgrade forever. It's a huge responsibility.

That's why there is the ability to update the update component separately iirc

PhilETaylor: Yes for when we screw up - like releasing a security fix that screws up the update process on the following core update ... as happened.

richard67:
I have tested this item :white_check_mark: successfully on 12e03381fd75382f8f41a4e26ea2ad59246d8198<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35291">issues.joomla.org/tracker/joomla-cms/35291</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35291">issues.joomla.org/tracker/joomla-cms/35291</a>.</sub>

administrator/components/com_joomlaupdate/restore.php
PhilETaylor: Or when I screw up restore.php ... as has happened today...

PhilETaylor:
Closing as I dont want to break the restore.php for a third time this week :-( 
Screenshot 2021-08-24 at 21 42 52

Issue #35257, opened 2021-08-20
[4] Restored article message issues


PhilETaylor:
### Steps to reproduce the issue

restore an article from the versions model.

### Expected result

1) The "Saved on" date show from `$model->getState('save_date')` to respect my language file date format (any of them, Im not picky ;-)) 
```
DATE_FORMAT_CALENDAR_DATE="%m-%d-%Y"
DATE_FORMAT_CALENDAR_DATETIME="%m-%d-%Y %H:%M:%S"
DATE_FORMAT_FILTER_DATE="m-d-Y"
DATE_FORMAT_FILTER_DATETIME="m-d-Y H:i:s"
DATE_FORMAT_JS1="m-d-y"
DATE_FORMAT_LC="l, F d, Y"
DATE_FORMAT_LC1="l, F d, Y"
DATE_FORMAT_LC2="l, F d, Y H:i"
DATE_FORMAT_LC3="F d, Y"
DATE_FORMAT_LC4="m-d-Y"
DATE_FORMAT_LC5="m-d-Y H:i"
DATE_FORMAT_LC6="m-d-Y H:i:s"
```
2) if there was no `$model->getState('version_note')` to show something like `(No Version Note)` instead of a space that then pushes the period looking odd. 

### Actual result

<img width="584" alt="Screenshot 2021-08-20 at 16 47 16" src="https://user-images.githubusercontent.com/400092/130259354-bc5fd8ff-300a-495d-b643-df44b9a69eec.png">

caused by language string:

```ini
JLIB_APPLICATION_SUCCESS_LOAD_HISTORY="Prior version restored. Saved on %s %s."
```

https://github.com/joomla/joomla-cms/blob/55b02c923b65dc9016a3f82671bb9bd9177e2e4d/libraries/src/Versioning/VersionableControllerTrait.php#L91

```php
$this->setMessage(
Text::sprintf(
  JLIB_APPLICATION_SUCCESS_LOAD_HISTORY', $model->getState('save_date'), $model->getState('version_note')
  )
);
```

brianteeman:
Easy enough to format the date

`HTMLHelper::_('date', $model->getState('save_date'), Text::_('DATE_FORMAT_LC3')),
`

But it would also need to be changed here as well

![image](https://user-images.githubusercontent.com/1296369/130288578-127336fc-2c47-4b35-ba1c-d3d63d043f44.png)

PhilETaylor:
[4] Show better message on restoring from versions #35299

PR #35299, opened 2021-08-22
[4] Show better message on restoring from versions

PhilETaylor:
Its not pretty and I dont like it, but it works.

Pull Request for Issue #35257

### Summary of Changes

Translates date according to `DATE_FORMAT_LC3` at all times
show `No version note` if there was no version note to fulfil a placeholder and not leave a trailing period after a space. 

### Testing Instructions

restore an item from content history

1) The "Saved on" date show from `$model->getState('save_date')` doesnt respect my language file date format

2) if there was no `$model->getState('version_note')` to show something like `(No Version Note)` instead of a space that then pushes the period looking odd. 

Screenshot 2021-08-20 at 16 47 16

Caused by language string:

```ini
JLIB_APPLICATION_SUCCESS_LOAD_HISTORY="Prior version restored. Saved on %s %s."
```

https://github.com/joomla/joomla-cms/blob/55b02c923b65dc9016a3f82671bb9bd9177e2e4d/libraries/src/Versioning/VersionableControllerTrait.php#L91

```php
$this->setMessage(
Text::sprintf(
  JLIB_APPLICATION_SUCCESS_LOAD_HISTORY', $model->getState('save_date'), $model->getState('version_note')
  )
);
```

Screenshot 2021-08-22 at 17 40 59
Screenshot 2021-08-22 at 17 40 55


brianteeman:
~Your screenshot shows why this change to remove time is not a good idea~

PhilETaylor:
> Your screenshot shows why this change to remove time is not a good idea

By using the language string [you yourself chose](https://github.com/joomla/joomla-cms/issues/35257#issuecomment-902931790) from the [list of available translated dates](https://github.com/joomla/joomla-cms/issues/35257#issue-975727511) I provided... I give up.

PhilETaylor:
Now Changed to `DATE_FORMAT_LC2` so that a time is provided. 

Screenshot 2021-08-22 at 18 09 14
Screenshot 2021-08-22 at 18 09 09

Quy:
The Version Note column displays `No version note` even when there is a note.

PhilETaylor:
> The Version Note column displays No version note even when there is a note.

I cannot replicate that sorry.

Screenshot 2021-08-30 at 15 18 47

Screenshot 2021-08-30 at 15 18 54

Quy:
Stupid me entered in `Note` and not `Version Note`. Sorry for the noise.

PR #35310, opened 2021-08-23
[4] Build tools should preserve the original timestamps

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

- adds a flag to preserve the timestamps for folder/files copied from `build/media_source` to `media`
- adds a flag for `mkdir` to use `mode: 0o755`
- adds a flag for `writeFile` to use `mode: 0o644` and also `encoding: 'utf8'`
 
### Testing Instructions

The timestamp flag is documented here: https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy.md

The flags mode and encoding for `writeFile`: https://nodejs.org/api/fs.html#fs_fspromises_writefile_file_data_options

The flags recursive and mode for `mkdir`: https://nodejs.org/api/fs.html#fs_fspromises_mkdir_path_options

Check that ALL files are still generated and with correct timestamps and mode

 
No

### DIFF
--- a/build/build-modules-js/init/common/copy-all-files.es6.js
+++ b/build/build-modules-js/init/common/copy-all-files.es6.js
@@ -15,5 +15,5 @@ const RootPath = process.cwd();
 module.exports.copyAllFiles = async (dirName, name, type) => {
   const folderName = dirName === '/' ? '/' : `/${dirName}`;
   await copy(join(RootPath, `node_modules/${name}/${folderName}`),
-    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`), { preserveTimestamps: true } );
+    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`), { preserveTimestamps: true });
 };
--- a/build/build-modules-js/init/common/copy-all-files.es6.js
+++ b/build/build-modules-js/init/common/copy-all-files.es6.js
@@ -15,5 +15,5 @@ const RootPath = process.cwd();
 module.exports.copyAllFiles = async (dirName, name, type) => {
   const folderName = dirName === '/' ? '/' : `/${dirName}`;
   await copy(join(RootPath, `node_modules/${name}/${folderName}`),
-    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`));
+    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`), { preserveTimestamps: true } );
 };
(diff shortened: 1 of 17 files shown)


build/build-modules-js/init/common/copy-all-files.es6.js
```diff
@@ -15,5 +15,5 @@ const RootPath = process.cwd();
 module.exports.copyAllFiles = async (dirName, name, type) => {
   const folderName = dirName === '/' ? '/' : `/${dirName}`;
   await copy(join(RootPath, `node_modules/${name}/${folderName}`),
-    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`));
+    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`), { preserveTimestamps: true } );
```
richard67:
```suggestion
    join(RootPath, `media/vendor/${name.replace(/.+\//, '')}/${type}`), { preserveTimestamps: true });
```
CS fix for making drone happy.

dgrammatiko: Thanks

richard67:
I have tested this item :white_check_mark: successfully on 567253f470cff14ee602ce30bf00f5c785947abe<br /><br />Code review: According to https://github.com/jprichardson/node-fs-extra/blob/master/docs/copy.md we should use the preserveTimestamps parameter to be sure timestamps of copied files are preserved because the default is false.

Real test: After an "npm ci" wihout this PR applied, all time stamps of files which just have copied from build/media_source to media (or folders below) have changed their time stamp. With the PR applied the timestamps are preserved. Comparing build/media_source and media folder structures with a good comparison tool like e.g. BeyondCompare shows that within a second.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35310">issues.joomla.org/tracker/joomla-cms/35310</a>.</sub>

richard67:
@dgrammatiko Don't forget to adjust title and description to all changes when done with coding ;-)

dgrammatiko:
> @dgrammatiko Don't forget to adjust title and description to all changes when done with coding ;-)

I Will do it in a bit, running some tests right now, so this is not really ready for testing yet

C-Lodder:
Default encoding for `writeFile` is `uft8`, so you shouldn't need to manually define it

dgrammatiko:
> so you shouldn't need to manually define it

It was more or less a find and replace, also the code is not going to be served on a browser so the extra bytes are not really harmful here. Anyways I've switched tasks so @C-Lodder if you pr the changes I will apply them

wilsonge:
Is this ready to test @dgrammatiko ?

dgrammatiko:
> Is this ready to test @dgrammatiko ?

yup

wilsonge:
Thanks!

PR #35292, opened 2021-08-22
[4] Cleanup of restore.php - Remove redundant constructor

PhilETaylor:
Code review

PR #35331, opened 2021-08-24
Fix parent construct

wilsonge:


PhilETaylor:
did I break something?

joomdonation:
Sadly Yes @PhilETaylor

joomdonation:
This PR does not fix the issue. Better revert https://github.com/joomla/joomla-cms/pull/35292

wilsonge:
Can you give me the stack trace?

PhilETaylor:
> Better revert #35292

I'll leave you to work out the issues then. The PR had appropriate tests and only deleted a constructor that contained no code.

wilsonge:
Unfortunately things called the constructor :)

PhilETaylor:
> Unfortunately things called the constructor :)

One thing. not things.

PhilETaylor:
Best pull 4.0.1 quickly from the update channels then ?

PhilETaylor:
looks like there are several places it was called.

PhilETaylor:
Line 799
Line 860
Line 2665

Those all need removing too.

joomdonation:
This one https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_joomlaupdate/restore.php#L799 and https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_joomlaupdate/restore.php#L860

PhilETaylor:
try this one https://github.com/joomla/joomla-cms/pull/35333

PR #35345, opened 2021-08-24
[4] Change update cache timeout to 12h by default

PhilETaylor:
### Summary of Changes

Proposing to increase the update cache timeout from 6 hours to 12 hours. I hope I found all the places this is needed.

Sometime in history, someone decided 6 was a sensible default. 

**As shown in the last week, getting an email every 6 hours about an available Joomla update is forcing people, [even seasoned Joomla people](https://twitter.com/jen4web/status/1429884632154189824), to disable the Update Notification Emails - not really what we want people to do.** 

Further evidence that the value of these notification is diminished by their frequency can be seen [by the number of articles and videos telling people how to disable the emails on their sites ](https://www.google.com/search?q=disable+joomla+update+notifications)and ignore the fact that updates are available. 

There is no need to check for updates every 6 hours - for three of those checks every 24 hours I'm probably asleep. 

This PR proposes to move to 12 hour timeouts for the cache - meaning that: 

 - Joomla potentially will check less for updates
 - Joomla will email update notification emails only once during any persons waking day... I consider a single email a day, and one for me to wake up to, better than one every 6 hours, so when I wake up I get 2 or 3 duplicated emails in my mailbox. 

This PR and the proposal is NOT to change existing sites timeout settings. Therefore there is no data migration planned in this PR, but to have this new setting only going forward (also why no Joomla 3 version of this PR) 

### Testing Instructions

Check for updates. Wait 12 hours, visit your site, see if you get update notification emails. 

potentially 4 emails per day

potentially 2 emails per day
Update sites under less stress/strain
Cheaper CDN payments for extension developers :) 

Probably. 

### Conflicts

Will eventually conflict with typo PR in #35346


brianteeman:
whatever time you set you will steal find people saying it is too often.

also this pr will only apply to new installs

PhilETaylor:
> whatever time you set you will steal find people saying it is too often.

I can't fix stupid but I can propose a more sensible value.

> also this pr will only apply to new installs

As clearly stated in my proposal PR text.

brianteeman:
for me this PR is pointless

simbus82:
For now is a "good" workaroud, but probably the best thing to do is decouple the Update Cache timing, from the Update Notification Emails. Maybe i want to receive these email once per week... but i want only 6 hours for the update cache.

PhilETaylor:
only problem with that is the cache needs updating before you can be sure there is still an update available :)

PhilETaylor:
See: #35368 so not so "pointless"

simbus82:
> only problem with that is the cache needs updating before you can be sure there is still an update available :)

Maybe a cron job can help? Every x hour the update cache will be refreshed, every x hour the notification plugin checks if the update cache is updated (:-P) and then do some actions... like refresh the update cache and send the notification mail.

Or can be do something like what Akeeba Admin Tools does with the "email throttling" for WAF blocked requests. This setting throttle down the frequency of emails sent.

brianteeman:
or simply update your site. thats the best way to stop the email

simbus82:
> or simply update your site. thats the best way to stop the email

A "site"?! We have hundred websites to manage, over 60 active Joomla sites, with dozens of extensions, customizations, integrations, etc. Here we never update an "important" site without going through a staging step.
(luckily softwares like @PhilETaylor's one give us a great help)

I know, we can always do a PR 😄

alikon:
what's wrong in simply disable that plugin if you like or dislike ?

simbus82:
> what's wrong in simply disable that plugin if you like or dislike ?

Because the world is not only black & white ;-)

PR #30056, opened 2020-07-08
[4.0] Replace default symfony error page with a shiny new Joomla one

PhilETaylor:
Closes https://github.com/joomla/joomla-cms/issues/29968

### Summary of Changes
READY FOR TESTING

A new custom fatal error page to replace the default symfony error page.

Joomla 4 now uses  `symfony/error-handler` component for handling fatal errors WHEN DEBUG MODE IS OFF (or in this case, when its not yet been turned on or off by `JConfig`)

**This is NOT a replacement error page or a duplication of `template/system/error.php`. That file will continue to be used for all the reasons its used already.**  It cannot be used when a fatal error occurs on boot up (and doesnt in Joomla 4 before this PR) because it relies on Joomla being booted, working, and relies on the whole stack working.

`fatal-error.html` is displayed when things are REALLY bad and unrecoverable only. Like your database server being offline.

This page is only used during fatal unrecoverable errors, normally at boot time, when it cannot be assumed any of Joomla is working (certainly not a Factory or application running, certainly not a language object available)

# Upstream Dependancy

It relies on [upstream changes I made](https://github.com/symfony/symfony/pull/37492#event-3519270830) in the `symfony/error-handler` component that have been approved and merged, and that have been made specifically to support the ability of Joomla to render such a custom fatal page and Joomla will be ~~one of the very first~~ the first project to customise the Symfony default error page, when using `symfony/error-handler` as a component in another non-symfony stack. 

### Testing Instructions

Install Joomla 4 from 4.0-dev
merge/apply #31743  (running `npm i` or `node build/build.js --build-pages` to be sure)
apply this PR - run `composer self-update --1 && composer update` (with v1 of composer)
edit configuration.php and change class JConfig to be class JConfigBroken
Visit your home page.

### Actual result BEFORE applying this Pull Request

image

### Expected result AFTER applying this Pull Request

Screenshot 2020-12-20 at 21 48 39

### Documentation Changes Required

New screenshots?

This page needs some content https://docs.joomla.org/J4.x:FatalError

PR #35352, opened 2021-08-24
[4] Include custom fatal page so that any post-processing, alerting, can happen as designed

PhilETaylor:
My bad again. Sorry. 

### Summary of Changes

In Joomla 3 fatal errors (like cant connect to the db) are shows like this `Error` - yup thats it, one word in black on a white blank page.

In Joomla 4 we moved to handling fatal errors with the Symfony Error Handler (not my work), and more importantly we (I) [implemented a custom template](https://github.com/joomla/joomla-cms/pull/30056) after changes [were merged upstream in Symfony](https://github.com/symfony/symfony/pull/37492#event-3519270830).

Joomla 4.0.0 was the first product to ever be published with a custom Symfony Error Page 🎉

This gave us the (now almost infamous) red page of death when a fatal error happens on the frontend of Joomla 4.

Screenshot 2021-08-24 at 21 31 56

(Many people are seeing this - when upgrading to Joomla 4 without reading, and with incompatible plugins - thats its job after all! )

One of the principles that was implemented was the introduction of a `fatal-error.custom.php` file that template designers could drop into `templates/system/` which could be used instead of the red screen of death - for the purposes (as documented in `fatal.php`) of:

```
 * Allow overriding Joomla supplied page to prevent changes being wiped on Joomla upgrade.
 * We allow it to be a PHP file so that any post-processing, alerting etc can happen.
```

However, as implemented prior to this PR, any PHP code in the `fatal-error.custom.php` would not actually parse!!!! Doh! as the content of the file was loaded with `file_get_contents` and then `echo` to the response. 

This PR fixes the invalid implementation by correctly including the file, allowing any PHP to parse (This allows the developer to use PHP to call a remote service, display dynamic data, send an email alerting apps, Sentry.io etc)

### Testing Instructions

1) Get a working Joomla 4 site
2) Change your db username to be invalid
3) Visit your home page - see red screen of death
4) create a new `/templates/system/fatal-error.custom.php` file with `<?php echo 'Hello World';` as its content.
5) Visit your home page - see white screen of death, inspect the HTML response and see that `<?php echo 'Hello World';` in unparsed and just output as the response
6) APPLY PR
7) Visit your home page - see `Hello World`

For bonus points, note you can do:

```php
<?php
echo $statusCode;
echo $statusText;
?>
```
and get an output of `HELLO WORLD500Whoops, looks like something went wrong.` this proves that the original status are being passed through for custom template designers to style/place however they like. 

For more bonus points in `configuration.php` set `$debug = true;` and see the Symfony Debug Trace - just to show you what the red page of death is actually "hiding" from your site visitors! 

REMEMBER this "red page of death" is a **FATAL ERROR page, its NOT MEANT to ever be shown in real life. ITS NOT THE ERROR PAGE, its a FATAL error page.** The lowest of the low. Again this is the reason the custom template is NOT in the `template/my_template/` folder but available to be placed in the `templates/system/` folder - because at the FATAL ERROR state  it cannot be assumed that any part of Joomla is available (Eg: a Factory or application or db connection) 

The content of the `fatal-error.custom.php` would be rendered in the source of the response - even if it was PHP.

PHP would not be parsed, just outputted. 

The PHP in `fatal-error.custom.php` is now parsed, triggered, and runs as expected and as originally designed for this feature. 

Custom template developers can go mad and replace the "Red Screen Of Death" with something else...  and can include PHP which is parsed.

Screenshot 2021-08-24 at 21 38 23

None. Joomla 4 is so new I doubt anyone has noticed this is even possible yet :) 

ref: 
@Webdongle https://forum.joomla.org/viewtopic.php?f=810&t=987965&p=3637277#p3637277

### DIFF
--- a/templates/system/fatal.php
+++ b/templates/system/fatal.php
@@ -40,7 +40,9 @@ if (file_exists(__DIR__ . '/fatal-error.html'))
  */
 if (file_exists(__DIR__ . '/fatal-error.custom.php'))
 {
-	$template = file_get_contents(__DIR__ . '/fatal-error.custom.php');
+	require __DIR__ . '/fatal-error.custom.php';
+
+	return;
 }
 
 echo str_replace(


wilsonge:
Thanks!

PhilETaylor:
> @wilsonge wilsonge merged commit b06f7c3 into joomla:4.0-dev  16 minutes ago

With no tests - gulp ;-)

PR #35362, opened 2021-08-25
[4.0] Added clear('order') in _getListCount(Query) of DataBaseModel.php

socke300:
### Introduction

When creating PR #35341, we noticed that the `_getListCount(Query)` of `DataBaseModel.php` did not clear the order of the query in the second if-statement. Because it is not important for a count function in which order the results are placed and could represent a performance problem in the future when using other queries, so we have added the clear function.

### Summary of Changes

In the method `_getListCount(Query)` of `DataBaseModel.php` the order was also cleared in the second if-statement.

### Testing Instructions

1. Install Joomla 4.0
2. Install Sample Data
3. Activate the Debug Tool *System -> Global Config -> Debug System on*
4. Open the Smart Search
5. Search for a word
6. For example, look for the query as in *Actual result BEFORE applying this Pull Request*
7. Repeat the steps with this PR

Search Query contains an **Order BY**.

```sql
SELECT l.link_id, l.object,SUM(m.weight) AS ordering
FROM kaftq_finder_links AS l
INNER JOIN `kaftq_finder_links_terms` AS m ON m.link_id = l.link_id
WHERE `l`.`access` IN (:preparedArray1,:preparedArray2,:preparedArray3) AND l.state = 1 AND l.published = 1 AND (l.publish_start_date IS NULL OR l.publish_start_date <= '2021-08-25 08:24:00') AND (l.publish_end_date IS NULL OR l.publish_end_date >= '2021-08-25 08:24:00') AND m.term_id IN (11)
GROUP BY l.link_id,l.object
HAVING SUM(CASE WHEN m.term_id IN (11) THEN 1 ELSE 0 END) > 0
ORDER BY ordering DESC
```

Search Query no longer contains an **Order BY**.

```sql
SELECT l.link_id, l.object,SUM(m.weight) AS ordering
FROM kaftq_finder_links AS l
INNER JOIN `kaftq_finder_links_terms` AS m ON m.link_id = l.link_id
WHERE `l`.`access` IN (:preparedArray1,:preparedArray2,:preparedArray3) AND l.state = 1 AND l.published = 1 AND (l.publish_start_date IS NULL OR l.publish_start_date <= '2021-08-25 08:23:00') AND (l.publish_end_date IS NULL OR l.publish_end_date >= '2021-08-25 08:23:00') AND m.term_id IN (11)
GROUP BY l.link_id,l.object
HAVING SUM(CASE WHEN m.term_id IN (11) THEN 1 ELSE 0 END) > 0
```

Probably not.

### Contributers
@scout507 @waleet

### DIFF
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -177,9 +177,8 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 		}
 
 		// Otherwise fall back to inefficient way of counting all results.
-
-		// Remove the limit and offset part if it's a DatabaseQuery object
-		// The order part is removed, as it is not needed for counting and could otherwise worsen the performance.
+		
+		// Remove the limit, offset and order parts if it's a DatabaseQuery object
 		if ($query instanceof DatabaseQuery)
 		{
 			$query = clone $query;
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -177,7 +177,7 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 		}
 
 		// Otherwise fall back to inefficient way of counting all results.
-		
+
 		// Remove the limit, offset and order parts if it's a DatabaseQuery object
 		if ($query instanceof DatabaseQuery)
 		{
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -182,7 +182,7 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 		if ($query instanceof DatabaseQuery)
 		{
 			$query = clone $query;
-			$query->clear('limit')->clear('offset');
+			$query->clear('limit')->clear('offset')->clear('order');
 		}
 
 		$this->getDbo()->setQuery($query);
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -179,6 +179,7 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 		// Otherwise fall back to inefficient way of counting all results.
 
 		// Remove the limit and offset part if it's a DatabaseQuery object
+		// The order part is removed, as it is not needed for counting and could otherwise worsen the performance.
 		if ($query instanceof DatabaseQuery)
 		{
 			$query = clone $query;


richard67 (approved):
Makes sense to me.

richard67:
@socke300 Maybe you should also change the comment in line 181 to reflect your change? https://github.com/joomla/joomla-cms/blob/5d2cdf4ea0051724056943660bc9348d5c687ec4/libraries/src/MVC/Model/BaseDatabaseModel.php#L181

socke300:
@richard67 Thanks for the feedback, comment has been added.

richard67:
@socke300 I meant to extend the comment by your change so that the previously available information in that comment is not lost.

I.e. change
`// Remove the limit and offset part if it's a DatabaseQuery object`
to something like
`// Remove the limit, offset and order parts if it's a DatabaseQuery object`

But like you have done it now it only mentions the order part.

socke300:
@richard67  Oh, I get it now, should I change it to ```// Remove the limit, offset and order parts if it's a DatabaseQuery object``` and remove the comment underneath?

richard67:
> 
> 
> @richard67 Oh, I get it now, should I change it to `// Remove the limit, offset and order parts if it's a DatabaseQuery object` and remove the comment underneath?

@socke300 Yes.

libraries/src/MVC/Model/BaseDatabaseModel.php
```diff
@@ -177,12 +177,12 @@ protected function _getListCount($query)
 		}
 
 		// Otherwise fall back to inefficient way of counting all results.
-
-		// Remove the limit and offset part if it's a DatabaseQuery object
+		
+		// Remove the limit, offset and order parts if it's a DatabaseQuery object
```
richard67:
```suggestion

		// Remove the limit, offset and order parts if it's a DatabaseQuery object
```
Please remove the tabs in the empty line 180 (but not the line itself) to satisfy our code style checker.

richard67 (approved):

richard67:
I have tested this item :white_check_mark: successfully on a20207d5e57bc182769c086487c2ccc3acf66d27<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35362">issues.joomla.org/tracker/joomla-cms/35362</a>.</sub>

richard67:
Hint for other testers: There are 2 very similar queries. The first one with limit and ordering, and the second one without limit. This PR here fixes the 2nd one so it also has no ordering. @alikon Could you test this one here?

bembelimen (approved):

alikon:
shouldn't this to be backported to 3.10 ?

alikon:
I have tested this item :white_check_mark: successfully on a20207d5e57bc182769c086487c2ccc3acf66d27<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35362">issues.joomla.org/tracker/joomla-cms/35362</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35362">issues.joomla.org/tracker/joomla-cms/35362</a>.</sub>

alikon:
for 3.10 see #35383

bembelimen:
Thanks

PR #32520, opened 2021-02-26
[4.1] Allow save as copy article in the frontend

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR adds support for **Save As Copy** article from frontend of your site

### Testing Instructions
1. Apply Patch.
2. Login to frontend of your site using an account which has permission to edit an existing article
2. Edit the an existing article. Look at the Bottom of the page, if you are using an account which is having Create Article Permission, you will see a new button **Save As Copy** to allows you to create new article base on that existing article
3. Press Save As Copy button, confirm that the article is successfully created.

PR #35378, opened 2021-08-26
[4.1] Allow Save As Copy Article In The Frontend

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This is a redo of RTC PR https://github.com/joomla/joomla-cms/pull/32520 . When you edit an article from frontend of your site, there will be a new button **Save As Copy** allows you to create a new article from data of that being edited article.

### Testing Instructions
1. Apply Patch.
2. Login to frontend of your site using an account which has permission to edit an existing article (could be super user account)
3. Edit an existing article. Look at the Bottom of the page, if you are using an account which is having **Create** Article Permission, you will see a new button **Save As Copy**.
4. Press **Save As Copy** button, confirm that the article is successfully created.


bembelimen:
Merged it as the other PR had 2 good tests. Thx.

Issue #35360, opened 2021-08-25
joomla4: image upload error message "This file is too large to upload."


djayko:
#### Steps to reproduce the issue
- setup plain joomla4 instance at php 8.0 or 7.3 or 7.4 with mysql 5.7
- keep standard setup values for uploads
- set server memory_limit to -1 (=no limit)
- upload an image via Menu -> Content -> Media 

#### Expected result
- image uploaded

#### Actual result
- error message ""This file is too large to upload."

#### System information (as much as possible)
- shared webspace on Ionos with PHP 8.0.9 mysql 5.7.33
- standard memory set up on serverside is memory_limit=-1 (nolimit)
- php settings from joomla4-system information
Upload Max Filesize | 64M
Post Max Filesize | 64 M
Memory Limit | -1

#### Additional comments
- tested on IONOS - Unlimited Plus - Shared Webhosting (everything is working fine on a joomla 3.10 instance on same webspace)
- tested with different image file types and sizes (from 1KB to 8MB)
- after creating a php.ini or .user.ini with "memory_limit="1024M" it worked well - i also tested lower limits like 32MB, 138MB etc.
- reference discussion: https://forum.joomla.org/viewtopic.php?f=811&t=987786&p=3636942#p3636942

PhilETaylor:
Not a Joomla issue. This is a hosting issue. 

If you override your default webhost settings with sensible sane defaults then it will work, as you have found out. 

Also note that `memory_limit` has no bearing on the ability to upload files. This is a memory limit that PHP can consume when running. 

```
https://forum.joomla.org/viewtopic.php?f=811&t=987786&p=3636942#p3636586

UPDATE:
By accident i figured out something strange.
I just created an empty php.ini file in joomla-root/administrator and this resulted in diffent values for post_max_size and memory_limit.
-> I tried to upload a file and it worked.
```

jwaisner:
Thank you @djayko for reporting this. Based on the information provided, this is not a Joomla issue but a hosting issue. Please work with your hosting provider to resolve.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35360">issues.joomla.org/tracker/joomla-cms/35360</a>.</sub>

djayko:
> Not a Joomla issue. This is a hosting issue.

I know, but joomla doesn't work as expected with setup of memory_limit = -1

[...]

PR #35392, opened 2021-08-27
[4] Remove check for memory_limit when uploading files

PhilETaylor:
Pull Request for Issue #35360

### Summary of Changes

Remove check to `memory_limit` when uploading files. 

### Testing Instructions

try to upload files bigger than `memory_limit`

Specifically from #35360

 - setup plain joomla4 instance at php 8.0 or 7.3 or 7.4 with mysql 5.7
 - keep standard setup values for uploads
 - set server memory_limit to -1 (=no limit)
 - upload an image via Menu -> Content -> Media

File uploads would fail if larger than `memory_limit` or if `ini_get('memory_limit) === -1`

File Uploaded, even larger than `memory_limit` upload without issue. 

@laoneo

### DIFF
--- a/administrator/components/com_media/src/Controller/ApiController.php
+++ b/administrator/components/com_media/src/Controller/ApiController.php
@@ -366,8 +366,7 @@ class ApiController extends BaseController
 
 		if (($params->get('upload_maxsize', 0) > 0 && $serverlength > ($params->get('upload_maxsize', 0) * 1024 * 1024))
 			|| $serverlength > $helper->toBytes(ini_get('upload_max_filesize'))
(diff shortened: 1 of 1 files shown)


brianteeman:
/me confused memory_limit is used in the installer as well

HLeithner:
-I would expect reason to limit it is prevent error messages for code like `$x = file_get_contents()` but that code should shouldn't exists if you expect files lager then 1 mb

chmst:
@PhilETaylor could you please resolve conflicts?

PhilETaylor:
Closing as has no chance or being merged

PR #35396, opened 2021-08-27
[4.1] Add some events when fetching media data

laoneo:
### Summary of Changes
In media manager we have events for the CUD operations but not for the read one. This pr adds the missing events.

During all events the objects can be manipulated, except the getUrl one. There it is possible for subscribers to provide a new url.

@wilsonge code review as there is nothing to test when all system tests are passing.

### Testing Instructions
Open the media manager in the back end.

All is working as expected.

All is working as expected.


administrator/components/com_media/src/Model/ApiModel.php
```diff
@@ -454,7 +460,15 @@ public function getUrl($adapter, $path)
...
 		}
 
-		return $this->getAdapter($adapter)->getUrl($path);
+		$url    = $this->getAdapter($adapter)->getUrl($path);
+		$newUrl = Factory::getApplication()->triggerEvent('onFetchMediaFileUrl', [$url]);
+
+		if ($newUrl && $newUrl !== false)
+		{
+			$url = $newUrl;
+		}
```
joomdonation: This does not look right. `triggerEvent` return array, so the modified $url could become an array, not a string as expected.

If the purpose is allow plugin to manipulate the $url before returning, how about pass it by reference like:

```php
Factory::getApplication()->triggerEvent('onFetchMediaFileUrl', [&$url]);
```

laoneo: In J4 you can't return anymore values, you have to set the value on the passed $event like, `$event['result'] = 'https://mynewurl.com';`
Correct way would be to create a specific event where you can set the url, but it would be a bit of overhead for such a small change.

joomdonation: Maybe I misunderstand your answer but `Factory::getApplication()->triggerEvent` still return array, same as in J3. And isn't your code still return values to `$newUrl`?

laoneo: I'm having an event object and there you can set only the result, check the line here https://github.com/joomla/joomla-cms/blob/4ee06a1fc3222f1c09ad5db37ebfe87deb7d40a8/libraries/src/Application/EventAware.php#L114

dgrammatiko:
I have tested this item :white_check_mark: successfully on e5dc8dd9559de962c5c3261dc13c9a864c197d0f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35396">issues.joomla.org/tracker/joomla-cms/35396</a>.</sub>

administrator/components/com_media/src/Model/ApiModel.php
joomdonation: @laoneo I'm unsure if your answer relates to the code which you change in this PR. Surely your PRs change returned data from string to array if there is a plugin implement the method **onFetchMediaFileUrl**. I tested it by:

- Adding this simple method to plugins/filesystem/local/local.php

```php
public function onFetchMediaFileUrl($url)
{
	return $url;
}
```

- Add simple var_dump($url) to the code you added

- The result is $url is now array instead of string as before.
url

laoneo: I registered the event like here 

https://github.com/joomla/joomla-cms/blob/4.0-dev/plugins/quickicon/joomlaupdate/src/Extension/Joomlaupdate.php#L63

So you get an event object where you can set the result as you can see here

https://github.com/joomla/joomla-cms/blob/4.0-dev/plugins/quickicon/joomlaupdate/src/Extension/Joomlaupdate.php#L143

This is for me the recommend way as events shouldn't have a return value, instead of you should manipulate the event object and set the result. As this is Joomla 4 only we should use the proper way.

Fedik: @laoneo I think @joomdonation is correct here, in the way how you trigger the event you will get an array result.

To achieve what you expect, you need to trigger event in new way: https://github.com/joomla/joomla-cms/blob/a8da66982a39030c0867253fa4116fee1d91f2ad/administrator/modules/mod_quickicon/src/Helper/QuickIconHelper.php#L258-L264

For this probably can use GenericEvent class

Fedik: Okay, sorry, I think my previous comment is wrong, ignore it, you both is right.

The plugin must implement SubscriberInterface, and a legacy plugins still return array:
https://github.com/joomla/joomla-cms/blob/8e2624aea96cb6b392b3368c6007365f964d938b/libraries/src/Plugin/CMSPlugin.php#L293-L296

I wonder what result will be if there will be 2 and more plugins, one is a new with SubscriberInterface and second is a Legacy. Even though legacy plugins should not be used.

laoneo: If it comes from J3, then I'm fine with legacy. But this here is J4 only so we should use the J4 way.
You can download the free version of DPMedia where I have an FTP or Unsplah package in it to test it with multiple adapters. I have not tested it with multiple plugins when one is using the legacy way and the other the new way and it is not the scope of this pr to do some research about this. If it is really an issue when there are plugins using different ways of subscribing and it is an issue, then we should cover this in a separate issue and not here.

Fedik: Yeah, okay if it works with j4 then it good,

We still need some clarification about mixed plugins, maybe add restriction that some plugin group must be only with SubscriberInterface, but you are right it is another issue

laoneo: Actually I did a test as I working on something else. When the plugin implements the SubscriberInterface interface then it uses the new way, otherwise the old one as you can see here 
https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Plugin/CMSPlugin.php#L192

administrator/components/com_media/src/Model/ApiModel.php
```diff
@@ -116,6 +116,8 @@ public function getFile($adapter, $path = '/', $options = [])
 		$file->path    = $adapter . ":" . $file->path;
 		$file->adapter = $adapter;
 
+		Factory::getApplication()->triggerEvent('onFetchMediaFile', [$file]);
```
HLeithner: Please don't use deprecated functions in new code....

administrator/components/com_media/src/Model/ApiModel.php
```diff
@@ -178,8 +180,12 @@ public function getFiles($adapter, $path = '/', $options = [])
 			$file->adapter = $adapter;
 		}
 
+		$files = array_values($files);
+
+		Factory::getApplication()->triggerEvent('onFetchMediaFiles', [$files]);
```
HLeithner: Please don't use deprecated functions in new code....

administrator/components/com_media/src/Model/ApiModel.php
```diff
@@ -454,7 +460,15 @@ public function getUrl($adapter, $path)
 			throw new InvalidPathException;
 		}
 
-		return $this->getAdapter($adapter)->getUrl($path);
+		$url    = $this->getAdapter($adapter)->getUrl($path);
+		$newUrl = Factory::getApplication()->triggerEvent('onFetchMediaFileUrl', [$url]);
```
HLeithner: Please don't use deprecated functions in new code....

HLeithner (changes requested):

administrator/components/com_media/src/Model/ApiModel.php
laoneo: How should it look like then? Make a suggestion.

HLeithner: Depends on the event.
Create a Event like this https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_media/src/Event/MediaProviderEvent.php maybe one can be named that it fits for all events you add here. Or 3 event for each case I think you have a better feeling what's needed, maybe 2 one for files and one for the url?

Then something like this (only a untested draft)
```php
$eventResult = $app->getDispatcher()->dispatch(
	'onFetchMediaFiles',
	AbstractEvent::create(
		'onFetchMediaFiles',
		[
			'eventClass' => 'Joomla\Component\Media\Administrator\Event\MediaFetchFilesEvent',
			'subject' => $this,
			'files' => $files,
		]
	)
);

return current($eventResult->getArgument('files', [$file]));
```

It's only a suggestion, I'm only on mobile to day and only want to give you a starting point before I leave.

laoneo: You don't have an app instance in the model.

nibra: You can just retrieve it like you did before:
```php
```

HLeithner: use `Factory::getApplication()` to get the application.

Other maintainers can help from here.

laoneo: If you do `Factory::getApplication()->->getDispatcher()->dispatch` or `Factory::getApplication()->triggerEvent` is basically the same as it uses static access. Just because there is a deprecated flag on `triggerEvent` and on `getApplication` not, it doesn't mean you should use `getApplication` too easy. If I'm not mistaken, we were once even discussing if `getApplication` should be deprecated as well.
Honestly I expected here an answer to properly inject the Dispatcher and go from there, but yeah :man_shrugging:

nibra: @laneo Agree, the dispatcher (or at least the container) should be injected, but that's not the case ATM. However, the invocation of deprecated methods is by no means acceptable. The static method is therefore the best solution in the current situation.

alikon:
I have tested this item :white_check_mark: successfully on 40c3b812199ea6beeae08f2ede77730ca7736a89<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35396">issues.joomla.org/tracker/joomla-cms/35396</a>.</sub>

dgrammatiko:
I have tested this item :white_check_mark: successfully on 40c3b812199ea6beeae08f2ede77730ca7736a89<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35396">issues.joomla.org/tracker/joomla-cms/35396</a>.</sub>

Fedik:
@laoneo I have thought about that thing with result for `$newUrl`, and I think correct would be to get result from the event instead of dispatcher. I suggest you next:

The event class, that have defined URL get/set:
```php
namespace Joomla\Component\Media\Administrator\Event;

class MediaFileUrlEvent extends AbstractEvent
{
    public function __construct($name, array $arguments = array())
    {
        parent::__construct($name, $arguments);

        // Check for required arguments
        if (empty($arguments['url']) || !is_string($arguments['url']))
        {
            throw new BadMethodCallException("Argument 'url' of event $name is empty or not of the expected type");
        }
    }
    public function getUrl(): string
    {
        return parent::getArgument('url');
    }

    public function setUrl(string $url)
    {
        parent::setArgument('url');
    }
}
```
Triggering the Event:
```php
$url   = $this->getAdapter($adapter)->getUrl($path);
$event = new MediaFileUrlEvent('onFetchMediaFileUrl', ['url' => $url]);

Factory::getApplication()->getDispatcher()->dispatch(
  $event->getName(),
  $event
);

$newUrl = $event->getUrl();
```

Plugin listener:
```php
public function onFetchMediaFileUrl(MediaFileUrlEvent $event)
{
    $event->setUrl('blabla/my/url.jpg');
}
```

As you see it more straightforward ;) 

@nibra @HLeithner please confirm, or maybe I wrong

administrator/components/com_media/src/Model/ApiModel.php
```diff
@@ -116,6 +117,11 @@ public function getFile($adapter, $path = '/', $options = [])
 		$file->path    = $adapter . ":" . $file->path;
 		$file->adapter = $adapter;
 
+		Factory::getApplication()->getDispatcher()->dispatch(
+			'onFetchMediaFile',
+			new Event('onFetchMediaFile', [$file])
```
Fedik: I think here we need to stick to CMS event `Joomla\CMS\Event\GenericEvent` when we do not have an event specific Class, or use `AbstractEvent::create('onFetchMediaFile', [$file])`,
to be consistent

nibra:
Regarding 'get result from the event instead of dispatcher':
This is more a matter of taste. One big source of confusion is that in Joomla, calling a plugin is always called triggering an event, although plugins can be event handlers but also just hooks.

Events are usually just emitted without expecting (or waiting for) results. Other parts of the software or even other systems may or may not react on those events.

Hooks are a technique to inject code - they get some input data and have a return value. Thus they can alter values or provide new data.

In Joomla, we have no distinction between those two approaches. We just have plugins that might or might not return values. In Joomla, we're using events to communicate with the plugins. If a plugin returns a value, it is expected that the value is returned by the `dispatch()` method.

In the present case, both @laneo's and @Fedik's approaches are valid an can be considered 'clean', while @laoneo's is more 'Joomla like'.

administrator/components/com_media/src/Model/ApiModel.php
nibra: It is better to provide a specific event. Generic events should be avoided, because event handlers react on the type of the event.

Fedik:
@nibra thanks for clarification

laoneo:
So boys, I hope now everybody is happy and we can get it merged.

Fedik:
@laoneo I made PR to your branch, 
then it should be good

laoneo:
Merged

laoneo:
Event calls are now the same across all methods. Can be tested now again.

HLeithner:
The event set functions need validators, Like is_array at least for setFiles and is_string for setUrls like in the constructor. Also the parameters must be Immutable (which is already done in the events).

laoneo:
There are no set functions in the new events.

HLeithner:
> There are no set functions in the new events.

in this case it makes no sense to use `return $event->getFiles();` when it's not planned that t he values can be changed by a plugin.

HLeithner:
Re reading your opening post you say
> During all events the objects can be manipulated, except the getUrl one. There it is possible for subscribers to provide a new url.

doesn't that mean the plugins should/can change the event attributes?

laoneo:
You can change by setArgument() which is a public function in the event. It is intended to give the plugin full control over the returned data. Better would be to make a proposal, instead of getting into an argument. All I want is to have these events in MM without over engineering the whole thing.

Fedik:
@laoneo I think @HLeithner means something like there:
https://github.com/joomla/joomla-cms/blob/3a5984d03366f707ef842ee48051af194f68848f/libraries/src/Event/Table/AbstractEvent.php#L44-L61

It does not have `setter` for `subject`, however the Event class will call **protected** `setSubject`, to validate the `subject` if someone will do `$event->setArgument('subject', $foobar);`

So we have validation in both cases, in `__constructor` and `setArgument`

HLeithner:
> You can change by setArgument() which is a public function in the event. It is intended to give the plugin full control over the returned data. Better would be to make a proposal, instead of getting into an argument. All I want is to have these events in MM without over engineering the whole thing.

It's not about over engineering it's that the new event system should be clean and safe, that's the reason for validation.

Fedir already wrote it.

Adding
```
    public function setUrl(string $url)
    {
        if (empty(url))
        {
            throw new BadMethodCallException("Argument 'url' is empty");
        }
        parent::setArgument('url');
    }
```

would fit the needs (at least for the one plugin).

Fedik:
@HLeithner there a caveat, 

```php
public function setUrl(string $url)
{
  if (empty(url))
  {
    throw new BadMethodCallException("Argument 'url' is empty");
  }
  parent::setArgument('url');
}
```

This will lead to infinity loop ;)
The method should be **protected**.

HLeithner:
> @HLeithner there a caveat,
> 
> ```
> public function setUrl(string $url)
> {
>   if (empty(url))
>   {
>     throw new BadMethodCallException("Argument 'url' is empty");
>   }
>   parent::setArgument('url');
> }
> ```
> 
> This will lead to infinity loop ;)
> The method should be **protected**.

haha I copied you example and you are right should be more like

```php
protected function setUrl(string $url)
{
  if (empty(url))
  {
    throw new BadMethodCallException("Argument 'url' is empty");
  }
  return $url;
}
```

Fedik:
yeah, originally I also thought it should work, 
but after couple tests I ended up with infinity loop :)

laoneo:
So instead of using the events arguments feature you want to store the url in an extra variable in the event class? I mean if a plugin wants to set an empty url then this is fine. If a plugin wants to set empty files this is also fine.

HLeithner:
> So instead of using the events arguments feature you want to store the url in an extra variable in the event class? I mean if a plugin wants to set an empty url then this is fine. If a plugin wants to set empty files this is also fine.

It's not an instead, that's how the event system works.

If it's ok that the file(s) is empty then we only have to validate that it is a string or an array (depending of the event).

The event constructor uses the the setArgument() function to save the variables in the $this->arguments array of the the event class.

the setArgument function checks if a set<variablename>() function exists and uses it for validation. if it fails the event can't be created. That's the reason to do extra checks like !empty() in the constructor. In the setUrl() function we would then only validate that the provided parameter is an string.

laoneo:
Also make a pr against my branch with your suggestion.

laoneo:
So changed it to an internal variable, hope @HLeithner is happy now and we can go ahead.

HLeithner:
> So changed it to an internal variable, hope @HLeithner is happy now and we can go ahead.

it has nothing to do to make me happy, it's about having a "bullet proof" api, which is the reason for the new event system. I will create pull request against your branch or commit directly what ever you prefer.

laoneo:
pr please

HLeithner:
created a pull request https://github.com/Digital-Peak/joomla-cms/pull/17 would be good if @Fedik can have a look too.

laoneo:
Merged it. @wilsonge all ok here now

laoneo:
@HLeithner can you have a look on the system tests as they are failing since your changes got merged. Thanks.

HLeithner:
I can replicate the problem locally but don't understand the reason.
It seems that the DELETE XHR request is not send to the server and the vue application is waiting for feedback that will never come because no request has been created. So I think it's unrelated to the PR but happens reliable in drone for this pr...

I can replicate this problem with Joomla 4.0.1, create a folder, click around (optioal), delete it. Sometimes the loading bar doesn't stop and no XHR DELETE request is send. Not sure for the reason. So may be completely unrelated to the current error.

laoneo:
Are you able to fix it?

HLeithner:
> Are you able to fix it?

wasted after of the day yesterday to debug drone but only found out that it fails as soon as line
`		$event = new FetchMediaFileEvent('onFetchMediaFile', ['file' => $file]);`
is executed.

I debug this on my github account to reduce notifications to others https://github.com/HLeithner/joomla-cms/pull/38

last try was successful and I think (while writing this) that the problem is an empty directory.

HLeithner:
@laoneo I think I would the reason the file API also sense directories against the File event
```
  [0]=>
  object(stdClass)#885 (13) {
    ["type"]=>
    string(3) "dir"
    ["name"]=>
    string(11) "test-folder"
    ["path"]=>
    string(34) "local-images:/test-dir/test-folder"
    ["extension"]=>
    string(0) ""
    ["size"]=>
    string(0) ""
    ["mime_type"]=>
    string(9) "directory"
    ["width"]=>
    int(0)
    ["height"]=>
    int(0)
    ["create_date"]=>
    string(25) "2021-11-08T20:09:15+00:00"
    ["create_date_formatted"]=>
    string(16) "2021-11-08 20:09"
    ["modified_date"]=>
    string(25) "2021-11-08T20:09:15+00:00"
    ["modified_date_formatted"]=>
    string(16) "2021-11-08 20:09"
    ["adapter"]=>
    string(12) "local-images"
  }
```

Which fails the file validation, so the question is now how to fix this. Add an additional event type or make the File event more generic. I would expect that a plugin would handle files and directories different.

laoneo:
There is no reason to handle directories different than files. Sending two events makes no sense as the result will be combined again and you loose the sorting when you combine them afterwards. All a developer wants is to have the possibility to interact with the result set before sending it to the client. I would suggest that your validation changes will be reverted and then when this one gets merged you can open a new pr with the validation stuff, so we can move forward with this pr. Otherwise it will never arrive in core and I need it for several reasons.

bembelimen:
I think there is no need to rush this, as we have plenty of time to get it merged till the next release. There are several people supporting this function, so it's very likely that it's getting merged. BUt we should really go for best practise here.

I see two things which has to be done:
- Rename the events (sorry for that) to communicate that there are possible files AND folders (e.g. FetchMediaItemEvent, FetchMediaListEvent, ...? anyone?)
- Add the validation. As @HLeithner did the first validation, probably he will provide the last bits, too?

If both are ready and tested, I'm happy to merge.

laoneo:
>     * Rename the events (sorry for that) to communicate that there are possible files AND folders (e.g. FetchMediaItemEvent, FetchMediaListEvent, ...? anyone?)

The function names are getFile/getFiles, in Media Manager we always talk about files and not items. Adding now a different therm would confuse the developers.

Fedik:
I agree with @laoneo it does not make sense for 2 events. 
`Model::getFile[s]()` handle both folder and files.

hm, only validation for type is https://github.com/joomla/joomla-cms/blob/8c336d45cc8c244e1add33ad99f0f6b04c755480/administrator/components/com_media/src/Event/FetchMediaFileEvent.php#L84-L88

okay, I see, it need to exclude extension validation if type is DIR (or just remove this section). 
Next will fail for folders always:
https://github.com/joomla/joomla-cms/blob/8c336d45cc8c244e1add33ad99f0f6b04c755480/administrator/components/com_media/src/Event/FetchMediaFileEvent.php#L102-L106

HLeithner:
A File is not a Folder, so a plugin developer would expect only a file, I don't know who named the function getFile but it seems it's simple wrong. Continuing the wrong wording make it worse.

Since we have now the time to do it right and don't need to rush out anything we should not introduce new "wrong" names.

Allon and fedir is right, having separate events is not needed since the type is in the object. Solving the validtion should be trivial (I will do this). I would also rename (in a b/c way, even if we have no b/c promise on this) the getFile(s) function to getItem(s) to reflect the real behavior.

laoneo:
But a folder is a file.

HLeithner:
> But a folder is a file.

yes in posix land because everything is a file... but we have separate functions to createFile and createFolder seems not to be the same.

laoneo:
Calm down, I will rename the events. All I want are the events, if the developers will understand then better, I'm fine with this.

administrator/components/com_media/src/Event/FetchMediaFileEvent.php
```diff
@@ -0,0 +1,165 @@
...
+		if (empty($value->path) || !is_string($value->path))
+		{
+				throw new BadMethodCallException("Property 'path' of argument 'file' of event {$this->name} has a wrong value. Valid: non empty string");
+		}
+
+		// A string
+		if (!isset($value->extension) || !is_string($value->extension))
+		{
+				throw new BadMethodCallException("Property 'extension' of argument 'file' of event {$this->name} has a wrong value. Valid: string");
+		}
```
HLeithner: Can you please add the if statement too, then the event should work again.

```suggestion
		if ($value->type === 'file'))
		{
		  // Non empty string
		  if (empty($value->path) || !is_string($value->path))
		  {
				  throw new BadMethodCallException("Property 'path' of argument 'file' of event {$this->name} has a wrong value. Valid: non empty string");
		  }
  
		  // A string
		  if (!isset($value->extension) || !is_string($value->extension))
		  {
				  throw new BadMethodCallException("Property 'extension' of argument 'file' of event {$this->name} has a wrong value. Valid: string");
		  }
		}
```

laoneo:
System test still fails

HLeithner:
> System test still fails

Has also to be fixed in the Files variant of the Event.

laoneo:
Can you do that please.

administrator/components/com_media/src/Event/FetchMediaFilesEvent.php
```diff
@@ -0,0 +1,196 @@
...
+				if (empty($value->path) || !is_string($value->path))
+				{
+						throw new BadMethodCallException("Property 'path' of argument 'file' of event {$this->name} has a wrong value. Valid: non empty string");
+				}
+
+				// A string
+				if (!isset($value->extension) || !is_string($value->extension))
+				{
+						throw new BadMethodCallException("Property 'extension' of argument 'file' of event {$this->name} has a wrong value. Valid: string");
+				}
```
HLeithner:
```suggestion
		if ($value->type === 'file'))
		{
				// Non empty string
				if (empty($value->path) || !is_string($value->path))
				{
						throw new BadMethodCallException("Property 'path' of argument 'file' of event {$this->name} has a wrong value. Valid: non empty string");
				}

				// A string
				if (!isset($value->extension) || !is_string($value->extension))
				{
						throw new BadMethodCallException("Property 'extension' of argument 'file' of event {$this->name} has a wrong value. Valid: string");
				}
		}
```

hopefully it's correct

laoneo:
Still not

laoneo:
You can push directly to the branch if you want. No need to make a suggestion first.

Fedik:
It also need to skip: size, width, height and other non folder thing.
btw, width, height valid only for images, and invalid for pdf and other documents and audio files

HLeithner:
> It also need to skip: size, width, height and other non folder thing. btw, width, height valid only for images, and invalid for pdf and other documents and audio files

if I'm not wrong I only check if it's an integer not if is > 0. What I have check in the original class it sets the values to 0.

Fedik:
```
if (!isset($value->width) || !is_integer($value->width))
```
`isset()` will return false if `$value->width === null` or not exists, 
and I guess for a folder it can be one or another

laoneo:
Events have been renamed. @HLeithner can you please fix your part please.

laoneo:
@HLeithner can you please have a look on the errors?

Fedik:
I have tested this item :white_check_mark: successfully on c42a5824e4ee900a415596b6b2eeea74b98b93cd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35396">issues.joomla.org/tracker/joomla-cms/35396</a>.</sub>

Fedik:
All seems works now, @HLeithner please confirm when you will have a min.

HLeithner:
thanks @laoneo for fixing it your self and simplify the validation. thanks @Fedik for testing.
It looks ok for me. Maintainers have to decide in which branch it should be merged @bembelimen

HusiBT:
I have tested this item :white_check_mark: successfully on c42a5824e4ee900a415596b6b2eeea74b98b93cd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35396">issues.joomla.org/tracker/joomla-cms/35396</a>.</sub>

HLeithner (approved):

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35396">issues.joomla.org/tracker/joomla-cms/35396</a>.</sub>

bembelimen:
I'm pretty sure, this should go into 4.1.

laoneo:
Initially the changes have been small so it made sense at that time to go into 4.0. Now it makes more sense for 4.1. Rebased.

bembelimen:
If you allow maintainer to change your code, I could trigger the sync with the base branch and merge. So for you you have to do it @laoneo , could you please do it?

laoneo:
@bembelimen , the maintainer checkbox works only for personal repos
https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/5

bembelimen:
Ah ok, thx.
Would you mind updating it so I can merge?

laoneo:
Did already

bembelimen:
Thx

PR #35398, opened 2021-08-27
[4][Security] Translation strings should be treated as UNSAFE

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Pass all the translation strings through the sanitizer

### Why?

The translation strings often have HTML markup and many times are used in combination with `innerHTML`. BAD combo!!!
Also, there is a `load` function that could load any arbitrary JSON into the translation string pool (an object basically). Another backdoor for XSS.

Using the existing sanitizer these backdoors are closed.

### Testing Instructions

Apply the PR
Run `npm ci`
Check that the strings everywhere in the backend/frontend are not affected

Yes, there should be a note that inline JS events or any `javascript:` are not allowed anymore and will be discarded.

FWIW I already did a basic search on the languages folders and all strings are **not** affected (eg no funky inline events, etc)

@SniperSister

### DIFF
--- a/build/media_source/system/js/core.es6.js
+++ b/build/media_source/system/js/core.es6.js
@@ -276,7 +276,7 @@ window.Joomla.Modal = window.Joomla.Modal || {
       newDef = newDef === undefined ? newKey : newDef;
       newKey = newKey.toUpperCase();
 
-      return Joomla.Text.strings[newKey] !== undefined ? Joomla.Text.strings[newKey] : newDef;
+      return Joomla.Text.strings[newKey] !== undefined ? Joomla.sanitizeHtml(Joomla.Text.strings[newKey]) : Joomla.sanitizeHtml(newDef);
     },
 
     /**


Fedik:
I am against it, 
the string sanitization should be done only when it need (within the code that use that string), not for all strings.

dgrammatiko:
> the string sanitization should be done only when it need (within the code that use that string), not for all strings.

Technically you're correct, that should be the right thing to do. The catches here are:
- too many files need adjustment (I can't do all that work for the project)
- 3rd PD that use `Joomla.Text` will need to do that themselves and most of them won't...

I'm ok closing this if the project decides to do this on a per-call of `Joomla.Text`

brianteeman:
Please reopen. This protects against naught translations as well.

Fedik:
> too many files need adjustment 

No one said it will be easy :)

For my opinion there is 80% cases where should be used `textContent` instead of `innerHTML`, and  10% should be corrected to use `textContent` and 10% where actually HTML is need to be filtered/sanitized.

Current approach gives overhead for runtime. Maybe it not that big, but still.

It is what I think. 
Will wait what other will say.

dgrammatiko:
> Current approach gives overhead for runtime. Maybe it not that big, but still.

I will have to agree with @Fedik here about the overhead and also it's a B/C break of the API if `Joomla.Text._('WHATEVER')` doesn't return the same string as it did before (it might happen if the sanitiser catches something it doesn't like/understands)

Hopefully someone else will lead the way and start patching the script files...

PR #35407, opened 2021-08-28
4.1 dev : new custom fields storage allowing efficient filtered articles lists 

pulsarinformatique:
Pull Request for Issue # .

### Summary of Changes

Hello

First of all I'm not sure this is the right place to propose the change. Please excuse me if I'm wrong and tell me where I should move my proposal in this case...

I whish we could offer a smart UI to allow non coders to filter the existing joomla list of articles upon the custom fields. We would keep the concept of attaching custom fields to category because
it gives us the notion of 'content types' (content type = category).

To perform fast filtered searches we need to change the way the values of custom fields are stored. instead of the existing vertical storage  in the #__fields_values table:

field_id | item_id | value

we would have a new table for each category: #__categories_fields whith an horizontal storage :

id (=item_id) | #field1# | #field2# | #field3# ...

the first colum (id) would match the id of the article inside this category

to retrieve the values of all the custom fields we would only need one join between #__content and #__categories_fields

The second thing would be to add a admin UI in the joomla list of articles so that we could filter any list upon given values of custom fields. We already have this for tags, let's have a similar thing with custom fields.

I propose we code an interface that will allow to define the field, its value and the way we filter (equal, less than, greater than, %like%, IN ...) 

### Testing Instructions

Let's create an EVENT category
let's create an 'event date' custom field we attach to this category
Let's create two articles (events) in that category with special event dates
let's create a category joomla menu item that displays all the events (= all the articles of this newly created category) with a special filter on the 'event date' custom field

We can't filter on the event date

We can filter on the event date


PhilETaylor:
The problem with this is it is not scalable, you cannot just have unlimited fields in a￼ Database table￼

pulsarinformatique:
yes you may be right but I've been using this very same architecture with some CCK for 8 years or so and we usually have between 5 to 50 fields for each content type without any issue. In some times we have even more
I guess we won't face a real limi in the real life here.

PhilETaylor:
> I guess we won't face a real limi in the real life here.

In your life. Maybe.  

In real life all the large CMS's have gone for the correct db architecture of Entity-Attribute-Value. https://en.wikipedia.org/wiki/Entity–attribute–value_model

This is how Magento allows unlimited attributed per product. 

This is how WordPress allows types, with unlimited fields. 

> we usually have between 5 to 50 fields for each content type without any issue. 

Just one limitation - So tell me, what is the maximum row length in Mysql? MSSQL? A quick google:

> MYSQL: The internal representation of a MySQL table has a maximum row size limit of 65,535 bytes
> MSSQL: So a record size (row size) in SQL server cannot be greater than 8060 bytes

So not unlimited. 

When you start working with large data sets, there are better ways to store the data. CCK's by their very nature, lend themselves to the EAV Model very easily and all the big players use that model.

....and all the big players use that model..... and are still able to "sort" on those fields (albeit not as performant as a SQL Join I admit.)

pulsarinformatique:
ok I guess you are right Phil. I just heard many times we could not propose to filter the joomla articles lists upon the custom fields because of the vertical storage of the CF values.

PhilETaylor:
I've created CCKs... one of the first, called Phil-A-Form, that did exactly what you propose with horizontal (creating a new column per field) and even that was problematic. 

There has been some renewed discussion about CCK and Joomla recently - maybe if everyone can get together and propose something amazing, with the right structure, and with the right migration path of com_content's data on existing sites, to a new Article Type in the CCK, then maybe we might get a CCK into Joomla. Let me link some links here:

PhilETaylor:
https://github.com/joomla/joomla-cms/discussions/19150 ah I see you are in that 
https://github.com/joomla/rfc/pull/26

PhilETaylor:
MAINTAINERS DO NOT MERGE THIS 

read: `wants to merge 47 commits into 4.0-dev from 4.1-dev` very very odd. 

This is not a PR... this is a discussion it seems.

pulsarinformatique:
We have been working with Seblod and others for years and know some of them very very well. My purpose here is not to have a full CCK in the core of Joomla (even if I'd love to but so many people here wrote me so many times it was not possible or too heavy ...)
my only concern here was to add the possibility to to filter the joomla articles lists upon the custom fields.

I whish I could participate in the debate, either for a real embeded CCK or at least this missing feature. Thanks for your kind answers

brianteeman:
@pulsarinformatique something went wrong with your pull request. There is no code from you just a mistaken merge of 4.1 into 4.0

pulsarinformatique:
Yes Brian, sorry, this is what Phil wrote above. I don't know where to write my proposal then :(

PhilETaylor:
Is your proposal just the words/thought of the idea - or do you have concrete code to contribute?

brianteeman:
@pulsarinformatique if it is just words then please write it as requested a thousand times in the discussion https://github.com/joomla/joomla-cms/discussions/19150

If it code then something went wrong with your pull request

pulsarinformatique:
No, I'm sorry this is just the idea of a possible DB structure. I wish I could take part in the debate because I feel I could help. Please tell me where I could participate

brianteeman:
@pulsarinformatique please close this. Please post it as requested in the https://github.com/joomla/joomla-cms/discussions/19150

pulsarinformatique:
ok Brian, with pleasure

Issue #35416, opened 2021-08-29
[jhelp] Glossary Joomla 4


YGomiero:
Joomla 4 :
Do not translate the line in the joomla.ini file 
administrator/language/xx-XX/joomla.ini

JHELP_GLOSSARY="Glossary"

With the current version (J4.0.2), there is an error in the help section of Joomla 4 because the translation term is included in the url of jdoc, necessarily on a page that does not exist.

The problem also occurs on all translation packs.

line 35 : administrator\components\com_admin\tmpl\help\default.php

`<?php echo HTMLHelper::_('link', Help::createUrl('JHELP_GLOSSARY'), Text::_('COM_ADMIN_GLOSSARY'), ['target' => 'helpFrame']); ?>`

HLeithner:
Seems we need a comment here? https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/language/en-GB/joomla.ini#L863 but isn't this true for all other JHELP entries too?

conconnl:
The comment doesn't stop the auto-translation bit. 
Duplicate strings are translated automatically, even when they are Hidden. Just had a whole discussion with Crowdin about it. A comment is just for the translators themselves.

HLeithner:
and can we stop crowdin todo this?

conconnl:
Yes, but that would mean an additional 4700 strings to translate. There are that much string in the CMS which have the same value. 

Another option is to Hide all duplicates with strict comparison. Which will not copy translations if the string constant and value are not the same as one in another file. 
This will of course also increase the amount of translatable strings. 

Another option would be a language file for the JHELP and not have it in Crowdin, that will increase the Core with 1 file.

HLeithner:
looks as the last option would be the best

brianteeman:
Duplicate strings absolutely should not be translated automatically. If they were really duplicates we would have used a global. Just because the english for two strings is the same doesnt mean that it should be translated the same because the context is different.

> Yes, but that would mean an additional 4700 strings to translate. There are that much string in the CMS which have the same value.

That cannot be correct

conconnl:
> Duplicate strings absolutely should not be translated automatically. If they were really duplicates we would have used a global. Just because the english for two strings is the same doesnt mean that it should be translated the same because the context is different.
> 
> > Yes, but that would mean an additional 4700 strings to translate. There are that much string in the CMS which have the same value.
> 
> That cannot be correct

[...]

PR #35418, opened 2021-08-29
[4] - Move JHELP strings too separate files

conconnl:
Pull Request for Issue #35416

### Summary of Changes
The JHELP language strings within Administrator and API are moved to their own files.
These strings should not be translated, while the rest of the string in the file are needed to be translated.
Having them in their own files, makes it possible to ignore the strings completely in the translation flows and decrease the risk of breaking the Help Server within the Administrator. 
Now the help screens are broken in the various languages, because the strings are translated, as they are made available for that purpose, while not needed.

### Testing Instructions
Check if the strings are still loaded in the Joomla Administrator help pages.

The language strings are used within the help pages to buildup the URLs.

The language strings are still used within the help pages to buildup the URLs.


Bakual:
I think we should rather look into removing the need for them in the first place, before we move them around and do nasty workarounds.

conconnl:
I'm not a developer, so I'm not going to play around in that area.
Now we have broken help pages in every other language then en-GB.
This is not a work around if you ask me, this is putting the strings in their own location and not mixed within "Required to be translated strings", which results in working translation packages and not broken help-pages.

Edit: The strings are used in the code.
Example:
line 35 : administrator\components\com_admin\tmpl\help\default.php
`<?php echo HTMLHelper::_('link', Help::createUrl('JHELP_GLOSSARY'), Text::_('COM_ADMIN_GLOSSARY'), ['target' => 'helpFrame']); ?>`

brianteeman:
the api file is not needed. the strings were only there to keep the two joomla.ini in sync. now that they are in their own file you can just delete the new api/joomlahelp.ini

infograf768:
If the only way is to add a specific ini file, it should be loaded whatever the language and totally independent from the language packs.

I would place that help file in `libraries/src/Language/joomlahelp.ini` and not in the en-GB  folders
And always load it directly in language.php

Bakual:
I've started a PR to show how the strings could replaced by hardcoded keys. At least for most of the strings that would work easily.

Opinions welcome. I could at least reduce the strings. Maybe even eliminate all.

conconnl:
We have 3 PRs thanks to @Bakual which fix it in a more Core way.
Therefore closing this one.

Issue #35384, opened 2021-08-26
[4.0][3.10] Pre-Update Checker shows misleading Update Required


OctavianC:
### Steps to reproduce the issue
Install an "outdated" extension that's compatible with the current Joomla! version.

### Expected result
No required updates showing or at least 'Optional Updates' showing.

### Actual result
Update Required
![image](https://user-images.githubusercontent.com/1422378/130936867-b5d0852c-0d74-44e0-839a-31be08deb450.png)

### System information (as much as possible)
Latest 3.10.1 and 4.0.2

### Additional comments
The Pre-Update Checker shows the update as required, even though the version currently installed works fine and is compatible. Very confusing to end-users IMO.
Rant: Honestly the Pre-Update Checker is doing more harm than good at this point. Issue #34599 is closed but points remain valid.

Kostelano:
What is meant by "version is working fine"? How can you know if there are any major changes in the newer version, for example, to improve security?

This page clearly states that there are updates for the installed versions. Isn't it the same information, in a slightly different form, you see on the main page of the control panel?

brianteeman:
![image](https://user-images.githubusercontent.com/1296369/130976594-f11c27f3-baf7-46a7-89cc-0d4e87e3c93e.png)

OctavianC:
> 
> 
> What is meant by "version is working fine"? How can you know if there are any major changes in the newer version, for example, to improve security?

A compatible Joomla! 3.10.x or Joomla! 4.0.x version of the extension, which is a few versions behind the latest version of the extension, but yet still fully compatible.
> 
> This page clearly states that there are updates for the installed versions. Isn't it the same information, in a slightly different form, you see on the main page of the control panel?

The page clearly states that there are REQUIRED updates. How does it assess that? It doesn't, the wording is simply misleading. We've not had this issue before 3.10 as people would always update their Joomla! installation, especially since the _one click_ update mechanism. Now, they are scared because of these messages. They postpone updates, which is detrimental to the CMS's health.

Thank you @brianteeman, I'm aware of that changelog, I've wrote it myself :)

Kostelano:
It took me a while to realize that these are your extensions :).

It seems to me that you should be one of the first to worry if people allow themselves not to update your extensions and then write a ton of support emails asking "I'm broken, what's wrong?" That is why I am still in favor of the fact that the wording is "requires updating", since it is logical to use the latest versions of the product with corrections for the latest version of Joomla.

[...]

PR #35419, opened 2021-08-30
[3.10] Fix misleading Update Required

joomdonation:
Pull Request for Issue #35384.

### Summary of Changes
Currently, the pre-update checker compares the current installed version of an extension with the **latest compatible** version (with the Joomla version the site is going to update to) instead of minimum compatible version and it causes the misdleading update required as described in https://github.com/joomla/joomla-cms/issues/35384 . This PR solves that.

### Testing Instructions
1. Install Joomla 3.10.1
2. Set **Update Channel** to **Joomla Next**.
3. Go to Extensions -> Manage -> Install, click on Options button in the toolbar and set **Minimum Stability** to **Release Candidate** (this step is needed because the extension I provide for testing here is in RC release only. If you use a stable extension for testing, please ignore this step)
4. Install weblinks 3.9.10-rc1 in the attachment
[pkg-weblinks-3.9.0-rc1.zip](https://github.com/joomla/joomla-cms/files/7074571/pkg-weblinks-3.9.0-rc1.zip)

5. Go to Components -> Joomla Update, Joomla should prompt you to update to Joomla 4.0.2 like in the screenshot below

update_to_402

The pre-update checker still marks the extensions as update required although the extension has indicates that the version is compatible with Joomla 4 already. See update element for 3.9.0-rc1 element
```xml
<targetplatform name="joomla" version="((3\.(9|10))|(4\.0))"/>
``` 

update_to_402

No update is required. It's green.
no_update_required

### DIFF
--- a/administrator/components/com_joomlaupdate/models/default.php
+++ b/administrator/components/com_joomlaupdate/models/default.php
@@ -1750,7 +1750,7 @@ ENDDATA;
 
 		$downloadUrl = $update->get('downloadurl');
 
-		return !empty($downloadUrl) && !empty($downloadUrl->_data) ? $update->get('version') : false;
+		return !empty($downloadUrl->_data) ? $update->get('minCompatibleVersion') : false;
 	}
 
 	/**
--- a/libraries/src/Updater/Update.php
+++ b/libraries/src/Updater/Update.php
@@ -66,6 +66,14 @@ class Update extends \JObject
 	protected $version;
 
 	/**
+	 * Update manifest `<version>` element of minimum compatible version
+	 *
+	 * @var    \stdClass
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $minCompatibleVersion;
+
+	/**
 	 * Update manifest `<infourl>` element
 	 *
 	 * @var    string
@@ -411,16 +419,16 @@ class Update extends \JObject
 
 					if ($phpMatch && $stabilityMatch && $dbMatch)
 					{
-						if (isset($this->latest))
-						{
-							if (version_compare($this->currentUpdate->version->_data, $this->latest->version->_data, '>') == 1)
-							{
-								$this->latest = $this->currentUpdate;
-							}
-						}
-						else
+						if (!isset($this->latest)
+							|| version_compare($this->currentUpdate->version->_data, $this->latest->version->_data, '>'))
 						{
 							$this->latest = $this->currentUpdate;
+						}
+
+						if (!isset($this->minCompatibleVersion)
+							|| version_compare($this->currentUpdate->version->_data, $this->minCompatibleVersion->_data, '<'))
+						{
+							$this->minCompatibleVersion = $this->currentUpdate->version;
 						}
 					}
 				}


richard67:
@joomdonation I've tested and noticed 2 things:

1. In both cases without and with this PR, I get no update information available for the weblinks package which I have downloaded via the link provided in the description and installed on a copy of my 3.10.1 homepage. See the red ellipse mark in the below 2 images. **_Update:_** It needs so set minimum stability in the extension installer's option to something lower than "Stable".

2. The search for the oldest compatible version should be limited to the currently installed version and not show versions older than that.

Without the changes of this PR applied, I get:

2021-09-01_j3-test-pr-35419_without-pr

The green frames show where the currently installed version is compatible.

The patchtester version 2.0.1 and not the installed 3.0.0 is shown as compatible with the current CMS 3.10.1 also without the patch of this PR, see the orange frame.

With the changes of this PR applied, I get:

2021-09-01_j3-test-pr-35419_with-pr

The red frames show where the currently installed version would be compatible but a lower version is shown.

For the patchtester nothing has changes, see orange frame.

~~Why I don't get any compatibility info for the weblinks I don't know.~~ **_Update:_** It needs so set minimum stability in the extension installer's option to something lower than "Stable".

richard67:
It seems to me that unfortunately the updater has no information about the currently installed version of an extension and so can't pick up the right update information with version greater or equal to currently installed. I hope so much that I'm wrong with that, otherwise it could become a bigger thing to change.

richard67:
@joomdonation ~~I think the problem comes from the updater's `getUpdateObjectsForSite` function having default value = false for the `$includeCurrent` parameter here https://github.com/joomla/joomla-cms/blob/3.10-dev/libraries/src/Updater/Updater.php#L269 , and so it checks only for versions greater than the installed one here https://github.com/joomla/joomla-cms/blob/3.10-dev/libraries/src/Updater/Updater.php#L289 and here https://github.com/joomla/joomla-cms/blob/3.10-dev/libraries/src/Updater/Updater.php#L361 .~~ Update: No, the `$includeCurrent` parameter is only present for the CMS update, but not for the extensions update here: https://github.com/joomla/joomla-cms/blob/3.10-dev/administrator/components/com_installer/models/update.php#L256 .

richard67:
@joomdonation Why I did not get any compatibility information I know now: I had not changed the minimum stability in the options of the extensions installer, so it was still "stable", and so the rc1 was not accepted..

joomdonation:
@zero-24 I would like to receive your feedback for this PR. It addresses the issue, however, as shown in the the screenshot which Richard posted, the version shows in **Joomla 4 Compatible Version** is now minimum compatible version, so it could be lower version compare to the current installed version.

I think it is OK because if the lower version is compatible with Joomla 4, then of course the installed version (which could be newer version) will also be compatible Joomla 4, and there should not be any problem. However @richard67 thinks that it could be confusing and proposed a ready to use PR https://github.com/joomdonation/joomla-cms/pull/3 to make the system return the compatible version which is greater than or equal current installed version. While it works, it requires more changes and could be a bit harder to maintain the code.

How do you think about it? We can discuss directly on Glip if needed. Thanks !

richard67:
I have an alternative proposal, I only need to complete testing instructions later (need a break now): #35510 .

richard67:
Please check #35510 , it's ready now.

joomdonation:
Closing in favor if https://github.com/joomla/joomla-cms/pull/35510

PR #35453, opened 2021-08-31
[4.0] Persist search in local storage

laoneo:
### Summary of Changes
When searching in the media manager and the page got reloaded, then the search should not be cleared and stored in the ~~localStorage~~ sessionStorage of the browser as we do with other parts in the media manager.

image

### Testing Instructions
Insert "joomla" into the search field in the media manager. And reload the page.

The whole list is shown without the search term.

The page is loaded with the results filtered with the name joomla.

### DIFF
--- a/administrator/components/com_media/resources/scripts/components/toolbar/toolbar.vue
+++ b/administrator/components/com_media/resources/scripts/components/toolbar/toolbar.vue
@@ -32,6 +32,7 @@
         type="text"
         :placeholder="translate('COM_MEDIA_SEARCH')"
         @input="changeSearch"
+        :value="search"
       >
     </div>
     <div class="media-view-icons">
@@ -111,6 +112,9 @@ export default {
       // eslint-disable-next-line max-len
       return (this.$store.getters.getSelectedDirectoryContents.length === this.$store.state.selectedItems.length);
     },
+    search() {
+      return this.$store.state.search;
+    }
   },
   watch: {
     // eslint-disable-next-line
--- a/administrator/components/com_media/resources/scripts/components/toolbar/toolbar.vue
+++ b/administrator/components/com_media/resources/scripts/components/toolbar/toolbar.vue
@@ -31,8 +31,8 @@
         class="form-control"
         type="text"
         :placeholder="translate('COM_MEDIA_SEARCH')"
-        @input="changeSearch"
         :value="search"
+        @input="changeSearch"
       >
     </div>
     <div class="media-view-icons">
@@ -114,7 +114,7 @@ export default {
     },
     search() {
       return this.$store.state.search;
-    }
+    },
   },
   watch: {
     // eslint-disable-next-line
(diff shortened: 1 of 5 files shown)


PhilETaylor:
Excuse me for asking questions without testing myself. 

1) Is the search persistent for a certain folder? I.e if I change folder will the search query remain?

2) Is it a "REAL search" (where it calls back to PHP and trawls the folders beneath this one for files that exist with that name) or just a filter on the current page?

richard67:
@laoneo CS errors reported: https://ci.joomla.org/joomla/joomla-cms/47245/1/21

laoneo:
@PhilETaylor 

1. No, but as you get forwarded to the same folder when the page get reloaded you have the same state.
2. No just filtering.

dgrammatiko:
@laoneo shouldn't it be sessionStorage instead of localStorage? LocalStorage will persist even if a user logs out and re login

laoneo:
Sure sessionStorage, using the same as with the other states.

dgrammatiko:
I have tested this item :white_check_mark: successfully on 8146ea77beef6b522c2f01f23f1673d838a00aac<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35453">issues.joomla.org/tracker/joomla-cms/35453</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 8146ea77beef6b522c2f01f23f1673d838a00aac<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35453">issues.joomla.org/tracker/joomla-cms/35453</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35453">issues.joomla.org/tracker/joomla-cms/35453</a>.</sub>

wilsonge:
I think we need to be a bit more careful than this. In all likihood navigating to a subfolder should clear the search rather than persisting it. although obviously I agree generally speaking that persisting it would be a good thing.

laoneo:
@wilsonge this is not the scope of this pr as you have the same behavior even without the changes here. I can do another one but please lets not mixing things up.

wilsonge:
Sorry so the current behaviour is that the search applies and isn't stored in the input box then?

laoneo:
It is stored, but you have no page reload on folder change. Only the content in the list changes, so the search box keeps the value and search is applied on folder change.

laoneo:
@wilsonge friendly reminder

laoneo:
@wilsonge friendly reminder

wilsonge:
Thanks!

Issue #35094, opened 2021-08-11
[4.0] [Workflow] There are non-localizable strings


Kostelano:
### Steps to reproduce the issue
1. Install the latest version of Joomla 4.
2. Install, for example, Russian (available for installation on the languages page) or some other language of your choice.
3. Set it as the default for the control panel.
4. Enable the use of a workflow in the content settings.
5. Go to the workflow page, then go to Transitions.

Here you can see 3 lines that are not translated as they were added by a hardcoded request.

[administrator\components\com_admin\sql\updates\mysql\4.0.0-2018-05-15.sql](https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_admin/sql/updates/mysql/4.0.0-2018-05-15.sql#L114) (similarly for postgresql)

```
INSERT INTO `#__workflow_transitions` (`id`, `asset_id`, `published`, `ordering`, `workflow_id`, `title`, `description`, `from_stage_id`, `to_stage_id`, `options`, `checked_out_time`, `checked_out`) VALUES
(1, 0, 1, 1, 1, 'Unpublish', '', -1, 1, '{"publishing":"0"}', NULL, 0),
(2, 0, 1, 2, 1, 'Publish', '', -1, 1, '{"publishing":"1"}', NULL, 0),
(3, 0, 1, 3, 1, 'Trash', '', -1, 1, '{"publishing":"-2"}', NULL, 0),
(4, 0, 1, 4, 1, 'Archive', '', -1, 1, '{"publishing":"2"}', NULL, 0),
(5, 0, 1, 5, 1, 'Feature', '', -1, 1, '{"featuring":"1"}', NULL, 0),
(6, 0, 1, 6, 1, 'Unfeature', '', -1, 1, '{"featuring":"0"}', NULL, 0),
(7, 0, 1, 7, 1, 'Publish & Feature', '', -1, 1, '{"publishing":"1","featuring":"1"}', NULL, 0);
```

![Screenshot_1](https://user-images.githubusercontent.com/8440661/129083987-76bd3cbe-0151-4465-a20b-121454a6592e.png)

Some of the strings have been translated, since there have been translation keys for them for a long time.

[...]

Issue #22754, opened 2018-10-21
Update releases should NOT change old sql update files


janmyszkier:
### Steps to reproduce the issue
1. Install Joomla from any year (i.e: 2017)
2. Wait till a new year (2018)
3. Download first release from 2018, check any update file that has copyright in it. I.e:
/administrator/components/com_admin/sql/updates/mysql/3.0.0.sql

### Expected result
New installer files should be provided when changing data. old installers should NOT be changed.

### Actual result
Copyright years is updated within OLD installer file.

### System information (as much as possible)
any system, this is about installer file contents.

### Additional comments
I understand doing `sed` replace inline or simple "Find and replace in the project" is easy and fast. But this is not how data consistency is enforced. 
This also makes very hard to realize what has changed in the update and if it is important or not because of a lot of copyrights changes within the files.

brianteeman:
Not sure how a change would help you now.

Bakual:
> This also makes very hard to realize what has changed in the update and if it is important or not because of a lot of copyrights changes within the files.

Just to give you some heads up. In the release note, you see if the update is important or not. If it's called a security update, it is important - no excuses. If it is a regular update, you can delay the update a bit, but it still absolutely should be installed.
As long as you don't do any core hacks (which you must not do anyway), it shouldn't matter to you what changes were done in the update in detail. You're going to apply the update sooner or later anyway.

So imho, I see no reason to change our past behavior of changing the copyright in all files (SQL files included). If there are other reasons to not change the copyright date, we can evaluate it for sure, but this reason doesn't sound like a good enough argument to me.

janmyszkier:
@brianteeman :
If the process is bad (and here, it is), and there's a way to improve it in the future, it SHOULD be improved. Especially when it saves time. 

@Bakual 
Once the update is not about security BUT it changes the way our core business functionality works (i.e - just stupid example - form handling), then for the business/client it is also important to look into how that thing was changed without browsing through thousands of files that just changed copyright notice along with the rest of the update. Security updates are another type of the required code updates.

brianteeman:
1. I dont believe the process is bad
2. Even if it was changed now it wont help you
3. You should be updating joomla through the joomlaupdate component

mbabker:
If you're talking about copyright statements in the extension manifest caches, honestly those caches shouldn't be in the SQL files anyway as they're automatically (re-)generated at initial install or update.  So that's a non-issue.

[...]

PR #35457, opened 2021-09-01
[4.0] Make transitions translateable

bembelimen:
Pull Request for Issue #35094 .

### Summary of Changes
Transitions are now translateable everywhere.

For example:

Backend article edit view:
grafik

Backend article list view:
grafik

### Testing Instructions
- Make a fresh install of this PR
- Activate the workflow
- Go to the transitions
- See that there are now language keys as title
- Check the different occurrence of the transitions (backend/frontend edit article, list view backend)
- Use own language keys in transitions + language override function.

Transitions are not (everywhere) translated

Transitions are translated everywhere.


Kostelano:
Probably, we need to add new lines including to `api/language/en-GB/joomla.ini`

brianteeman:
Yes please, at least for now, it is helpful if we just keep these files in sync

brianteeman:
I am not a fan of editing previous update sql. They should be immutable.

brianteeman:
Also the update sql you have will fail on an existing site.

/* SQL Error (1062): Duplicate entry '1' for key 'hvu3i_workflow_transitions.PRIMARY' */

richard67:
> I am not a fan of editing previous update sql. They should be immutable.

@brianteeman I thought you meanwhile had understand that "immutable" is wrong due to all my past work on old update SQL files. And here you are again wrong: We can and should change the old update SQL because there is no risk in doing it and it makes the workflow transitions to be translatable also after update from 3.10 to 4.0.

> Also the update sql you have will fail on an existing site.
> /* SQL Error (1062): Duplicate entry '1' for key 'hvu3i_workflow_transitions.PRIMARY' */

Than only happens if you apply the update SQL changed by this PR to a Joomla 4 database. But that will never happen since this SQL will only run when updating a 3.10.

brianteeman:
I stick by my views

richard67:
As usual.

brianteeman:
No point in having opinions if you don't believe in them. To repeat it's not about the technical need but as pointed out previously by another user changes to existing dated files can break complex work flows for server management

bembelimen:
Hello,
thanks for the feedback.

> Also the update sql you have will fail on an existing site.
> 
> /* SQL Error (1062): Duplicate entry '1' for key 'hvu3i_workflow_transitions.PRIMARY' */

My concerns were, that I didn't want to change existing workflows, so I updated only the 3.x => 4.x path, not adding a new file which changes existing data. How did you raise this error? Could you give me please a little guide, so I can fix it?

brianteeman:
If you are updating manually (perhaps you want to keep everything in github) then its reasonable to assume that someone would run the sql updates manually. Thats when you get the error. also see #22754

> My concerns were, that I didn't want to change existing workflows,

So now we have two sites with different data depending on when they were installed

When you now tell a user that workflows are translatable and they just need to create a language override it will only work on one of the two sites.

If you do an update query instead then everyone is on the same data (unless they changed the text already in which case their changed data is untouched)

UPDATE #__workflow_transitions SET title="PUBLISH_AND_FEATURE" WHERE title = "Publish & Feature"

bembelimen:
Thanks @brianteeman I've pushed the update files.

Kostelano:
I have tested this item :white_check_mark: successfully on c091084e4436572a2ad00929de13bee260bc10e7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35457">issues.joomla.org/tracker/joomla-cms/35457</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on c091084e4436572a2ad00929de13bee260bc10e7<br /><br />All overrides work!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35457">issues.joomla.org/tracker/joomla-cms/35457</a>.</sub>

richard67:
@bembelimen This PR has conflicts now due to the names of the SQL update scripts already being used. That comes from the merged PR #35483 . Could you solve that?

richard67:
Conflict solved, Since it doesn't have changed anything on what was relevant for the test, I restore the test results in the issue tracker so they are counted, and will set RTC.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35457">issues.joomla.org/tracker/joomla-cms/35457</a>.</sub>

wilsonge:
Thanks!

Issue #24132, opened 2019-03-08
[4.0]Preview not working for unpublished articles


Puneet-Jain-18:
### Steps to reproduce the issue
go to administrator
Create new article=>save it =>click on preview

### Expected result
article is previewed

### Actual result
![screenshot 11](https://user-images.githubusercontent.com/34083630/54042321-d3eca000-41ef-11e9-82a4-4f73c936742d.png)

### Additional comments
previewing is working correctly for published articles

ReLater:
In title you say "unpublished article", in additional comments "published article".
Could you please show other settings, too. Access level and so on.

Puneet-Jain-18:
PS: I said it was working **correctly** for **published articles**

Here is screeshot for problematic unpublished  article
![screenshot 12](https://user-images.githubusercontent.com/34083630/54047210-7a3ea280-41fc-11e9-845f-939fe5dd5211.png)

ggppdk:
So what is the Expected result if you are not logged ?
- to be redirected to login page ?

And then
if are already logged or you have logged after redirection with an account that cannot preview the content
- then a more helpful message should appear like ?

"You cannot preview this item, please login with a more privileged account"

Puneet-Jain-18:
I think user would not be able to visit this page if he is not logged in with required privileges

simbus82:
@ggppdk  Previewing article must work with or without frontend login, if an user is working in backend. And must work with an published or an unpublished article. There is not much to interpreting...
Otherwise it's not a preview ... That's how it works in the world of content creation.

> **Expected result**
> article is previewed

ggppdk:
@simbus82 

i understand your argument about UX , **yes it would be very nice**

as long as it is not an ugly hack with security concerns,
that also cares to preview with the desired access level

i mean the preview button is a little misleading, which access level do you want to preview ?
public ?
registered ?
other ?

but ok, i understand that most common case of preview is "public" access level ...

- so you would want to preview unpublished content while only having access level public and while you are not logged, and while executing frontend plugins , etc, ok, if this can be done without being hack
- or you get auto-login as the current backend user but your access levels are reduced to public  ?
- or you get auto-login as the current backend user but your access levels do not change ? then there is the case that the backend user is not allowed "site login" but lets handle this case by just forcing a frontend login screen
- other case ? e.g. multiple preview buttons with a drop down , "preview as registered" , "preview as public" ?

**with above i mean**
- **first we need to decide what the preview "thing" will be** 
[...]

PR #35385, opened 2021-08-26
[4.1] Shareable content drafting

Dako390:
### Summary of Changes
Based on the idea of: https://magazine.joomla.org/authors/nikita-dhiman

This PR provides a new entry to the menu called "Draft manager" and the options in "Articles" and "Article" to save one or more articles as drafts.

Draft Manager

In the "Draft manager" you can see all your drafts at once. It has a link to the original article (Link on title) and a link to the draft itself. You are able to manage these selected drafts and have the following options: 

- Select multiple drafts and share them. This option will create a shareable link of the drafted article.
- Select multiple drafts and unshare them. This option will destroy the link and you can't open this link anymore (404 is going to happen).
- Select multiple drafts and delete them. This option will delete the selected drafts.

Additionally there is a new button in "Articles" which allows you to save multiple articles as drafts and you can explicitly save a single article as draft in the "Edit" view.

2021-08-26_5 18 01_PM

2021-08-26_5 17 51_PM

**This PR is mainly an idea of how a "Shareable content drafting" could look like. Not all functionalities are granted.**

### Before Testing
After pulling this PR you need to update your database in "System" --> "Maintenance" --> "Database". If you install Joomla! freshly with this PR everything should be set.

### Testing Instructions
Create some articles as drafts, share them, test these links and delete these drafts.
Best practice at the moment would be to make a copy of the article you would like to draft, edit this copy, save it and then save it as a draft.

### Expected result AFTER applying this Pull Request
You should be able to save articles as drafts and share them. Refer to the text and pictures above to get a feeling how to use the "Draft Manager".

### Documentation Changes Required
Should be documented in the Joomla! Docs as a User Guide eventually.

PR #35488, opened 2021-09-06
[4][com_content] - preview unpublished article from admin

alikon:
Pull Request for Issue #24132 .

### Summary of Changes
check if we are on preview mode

### Testing Instructions
click preview on unpublished article from admin

404

preview works

### DIFF
--- a/components/com_content/src/Model/ArticleModel.php
+++ b/components/com_content/src/Model/ArticleModel.php
@@ -55,6 +55,9 @@ class ArticleModel extends ItemModel
 		$offset = $app->input->getUint('limitstart');
 		$this->setState('list.offset', $offset);
 
+		$preview = $app->input->getInt('preview');
+		$this->setState('article.preview', $preview);
+
 		// Load the parameters.
 		$params = $app->getParams();
 		$this->setState('params', $params);
@@ -206,9 +209,10 @@ class ArticleModel extends ItemModel
 
 				// Filter by published state.
 				$published = $this->getState('filter.published');
-				$archived = $this->getState('filter.archived');
+				$archived  = $this->getState('filter.archived');
+				$preview   = $this->getState('article.preview');
 
-				if (is_numeric($published))
+				if ((is_numeric($published)) && (!$preview))
 				{
 					$query->whereIn($db->quoteName('a.state'), [(int) $published, (int) $archived]);
 				}
@@ -222,10 +226,13 @@ class ArticleModel extends ItemModel
 					throw new \Exception(Text::_('COM_CONTENT_ERROR_ARTICLE_NOT_FOUND'), 404);
 				}
 
-				// Check for published state if filter set.
-				if ((is_numeric($published) || is_numeric($archived)) && ($data->state != $published && $data->state != $archived))
+				// Check for published state if filter set and not preview.
+				if (!$preview)
 				{
-					throw new \Exception(Text::_('COM_CONTENT_ERROR_ARTICLE_NOT_FOUND'), 404);
+					if ((is_numeric($published) || is_numeric($archived)) && ($data->state != $published && $data->state != $archived))
+					{
+						throw new \Exception(Text::_('COM_CONTENT_ERROR_ARTICLE_NOT_FOUND'), 404);
+					}
 				}
 
 				// Convert parameter fields to objects.


PhilETaylor:
Doesnt work for me. 

1) Ensure logged out of frontend
2) Login to admin
3) Ensure logged out of frontend (no mixed use session)
4) Edit an article and unpublished it. Save.
5) Click preview
6) Still see a 404 error on my iframe url http://127.0.0.1:4444/index.php/asdf where asdf is my article title/slug.

However what this PR does do is introduce a huge security issue which allows ANY UNPUBLISHED ARTICLE to be viewed, without login to either the frontend or admin just by appending `?preview=1` to any url (and because you can guess the url of content items because you can increment id numbers, you can now view all unpublished content with no ACL checks at all... 

So yeah... This PR in its current state should be rejected :)

alikon:
open to impovements of course

PhilETaylor:
Im confused why you ever thought this would work? as the preview modal doesnt have code to even add `preview=1` to the url?

This needs some kind of cryptographic hashes as tokens as a preview param, or a better way is to check Admin Sessions from the frontend just before going to the 404 exception.

HLeithner:
funny phil I tested this PR and found out the same ;-) it's a complete security nightmare and by the way the issue is not an issue if you have shared sessions. Then you should have to see the article in frontend even if it's unpublished...

alikon:
https://github.com/joomla/joomla-cms/blob/4ee06a1fc3222f1c09ad5db37ebfe87deb7d40a8/libraries/src/Toolbar/ToolbarHelper.php#L126

HLeithner:
@alikon I close this because it's not needed, if you have a shared session or loggedin in frontend the preview works even for an unpublished article. Or do I miss something?

PhilETaylor:
@alikon well that quoted method still doesnt add `preview=1` 

```
// Add a preview button.
		$bar->appendButton('Popup', $icon, 'Preview', $url . '&task=preview', 640, 480, $bodyHeight, $modalWidth);
```

Im sorry @HLeithner but this is a genuine issue that needs resolving (there are open issues from me about it) conceptually if you are logged into the admin console and there is a preview button it should be useable WITHOUT SHARED SESSIONS because shared sessions IS NOT THE DEFAULT for Joomla.

PhilETaylor:
>  if you are logged into the admin console and there is a preview button it should be useable WITHOUT SHARED SESSIONS because shared sessions IS NOT THE DEFAULT for Joomla.

So basically what you are saying @HLeithner - is that for "some Joomla admin features" to work, you first need to be logged into the frontend as well... 

So that begs the question, why we have split sessions at all... would you force me to be logged into the frontend to change a users password in admin? or to install an extension? No of course not, and there is no reason I should need to be logged into a different application (client id) in order to preview the content Im creating in admin console. 

If the preview button shows - it should be useable, without making someone think. 

Maybe we just show the preview button if the user has a valid frontend (ie. shared) session? and hide the button if they dont? thats better than showing a button and having it broken (404) when pressed?

alikon:
> if you have a shared session or loggedin in frontend the preview works even for an unpublished article. 

and if you have not ?

HLeithner:
So maybe we should disable the preview button on not shared session, adding a magic token to the request is possible but needs more work then just add a parameter.

PhilETaylor:
> So maybe we should disable the preview button on not shared session, 

Thats one solution, but it hides an otherwise nice feature, a new feature of Joomla 4. 

> adding a magic token to the request is possible but needs more work then just add a parameter.

Agreed. 

What do other CMS's and applications do for preview modes? might be worth checking out, I think WP just uses shared session, so thats how they get around it 

Whats the reason we dont use shared session by default?

HLeithner:
@PhilETaylor I don't have a strong opinion if it should be disabled or have a secure way to view it in the frontend without being logged in. If we add a secured parameter we should do this in a generic way so it works for all components out of the box (or with little effort)

HLeithner:
> What do other CMS's and applications do for preview modes? might be worth checking out, I think WP just uses shared session, so thats how they get around it

I think that's also how it's done in typo3 and drupal but don't nail me.

> 
> Whats the reason we dont use shared session by default?

I would expect for legacy reasons? I see no problem activating this per default for new installations. (if it's a problem it shouldn't exists)

PhilETaylor:
> . If we add a secured parameter we should do this in a generic way so it works for all components out of the box (or with little effort)

Well thats what the preview button is currently trying to do by adding `task=preview`... that could become `secureToken=tokenhere...` to make it generic. 

```php
/**
	 * Writes a preview button for a given option (opens a popup window).
	 *
	 * @param   string   $url            The name of the popup file (excluding the file extension)
	 * @param   bool     $updateEditors  Unused
	 * @param   string   $icon           The image to display.
	 * @param   integer  $bodyHeight     The body height of the preview popup
	 * @param   integer  $modalWidth     The modal width of the preview popup
	 *
	 * @return  void
	 *
	 * @since   1.5
	 */
	public static function preview($url = '', $updateEditors = false, $icon = 'preview', $bodyHeight = null, $modalWidth = null)
	{
		$bar = Toolbar::getInstance('toolbar');

		// Add a preview button.
		$bar->appendButton('Popup', $icon, 'Preview', $url . '&task=preview', 640, 480, $bodyHeight, $modalWidth);
	}
```

PhilETaylor:
> I see no problem activating this per default for new installations. (if it's a problem it shouldn't exists)

Well I guess thats a decision for someone with a higher pay grade than me :)

PhilETaylor:
just checking the documentation for the edit article page - https://docs.joomla.org/Help4.x:Articles:_Edit/en and there is not really help there:

"Preview. Opens a modal dialog showing a site view of this article."

I guess as a "quick fix" that could be improved to say that shared sessions needs to be enabled for that to work.

HLeithner:
I would suggest to use the shared session (because it's already proofed to work) and activate it in new installations starting with 4.1. 
I will bring this into PD tomorrow.

And hide the button if shared sessions are disabled. I think we don't have a way to detect if the user is logged in in frontend right?

PhilETaylor:
Im thinking this is probably the best solution. Shared Sessions that is. 

A secure token would have to be time based, because it could not be stored in a session for checking - cause - eh - different sessions :) 

We could also check the popup url for the preview var, and if 1) We are just about to go to the 404 exception, AND 2) We have a preview var in the url AND 3) shared sessions is NOT enabled, THEN we show a helpful "In  order to preview content you need shared sessions enabled (or logged into the frontend)... 

This is then backward compatible and Joomla 4.0.x sites can have that fix while waiting for the 4.1 shared session as default (because a 4.0.1 site upgraded to 4.1.0 will not have shared session... so yeah, this is probably what needs to be done as well.)

PhilETaylor:
Also important to note, that preview function cited by @alikon and referenced by me IS NOT USED for this toolbar button!!! so that `&task=preview` is NOT appended. (tested by deleting that method from the ToolbarHelper.php !!!)

PhilETaylor:
The method that IS used is in libraries/src/Toolbar/CoreButtonsTrait.php

```php
	/**
	 * Writes a preview button for a given option (opens a popup window).
	 *
	 * @param   string  $url        The name of the popup file (excluding the file extension)
	 * @param   string  $text       The text of button.
	 * @param   bool    $newWindow  Whether to open the preview in _blank or just a modal
	 *
	 * @return  PopupButton|LinkButton
	 *
	 * @since   4.0.0
	 */
	public function preview(string $url, string $text = 'JGLOBAL_PREVIEW', $newWindow = false)
	{
		if ($newWindow === true)
		{
			$button = $this->linkButton('link', $text)
				->url($url)
				->attributes(['target' => '_blank'])
				->icon('icon-eye');
		}
		else
		{
			$button = $this->popupButton('preview', $text)
				->url($url)
				->iframeWidth(640)
				->iframeHeight(480)
				->icon('icon-eye');
		}

		return $button;
	}
```

HLeithner:
I'm not sure but I think task=preview was used in earlier versions of joomla? Adding a preview=1 parameter and show a proper error message should be possible, at least we should not check if the article exists at this time, but you say to simple write it to the 404 page so it should be fine.

I tried to edit the wiki page but failed :-(

PhilETaylor:
>  at least we should not check if the article exists at this time, but you say to simple write it to the 404 page so it should be fine.

Correct, as to do it any other way could introduce a security issue where a hacker could enumerate which article ids exist based on the response. 

My suggestion, doing it just before throwing the 404 exception around line 233 - so hacker with preview=1 would see "you need to login to preview, or use shared sessions help text...." (And prefer a HTTP header of 404 so that google doesnt index that url) for ALL ids he tries to access - regardless if the article exists in the db or not.

alikon:
did some of you ever tryied that silly code . and maybe even on debug ? or it's asking too much ?
Screenshot from 2021-09-06 21-08-13

PhilETaylor:
> did some of you ever tryied that silly code . and maybe even on debug ? or it's asking too much ?

Im not sure what point you are trying to make but

1) That is your code in your PR. "article.preview" doesnt show anywhere in Joomal 4
2) I use debug mode = on and error_reporting = development (and php.ini display_errors=on and error_reporting=E_ALL) all the time, every day. 
3) Asking too much? I tested the PR, I actually looked at the PR on the sofa, saw that it was obviously wrong and a security risk, and walked to my mac to test to ensure I was right - and it was a security issue. I tested the PR as presented.

PhilETaylor:
> did some of you ever tryied that silly code

Yes, I set all my test articles to unpublished - logged out of the admin console and the frontend, and appended `?preview=1` to all urls 

http://127.0.0.1:4444/index.php?option=com_content&view=article&id=1&Itemid=102&preview=1
http://127.0.0.1:4444/index.php?option=com_content&view=article&id=2&Itemid=102&preview=1
http://127.0.0.1:4444/index.php?option=com_content&view=article&id=3&Itemid=102&preview=1
http://127.0.0.1:4444/index.php?option=com_content&view=article&id=4&Itemid=102&preview=1

and I could view all my unpublished content - like a hacker would.

HLeithner:
I saw the mail and checked the code and see that it doesn't looked good, so I tested the patch on one of my dev instances to be sure. So yes I tested it and was able to view the unpublished article without being logged in (frontend and backend).

alikon:
my point was only that there is some kind of "preview" info 
...not more nor less ...
looking forward to see a good 1 pr
:smile:

ghost:
@HLeithner 

> I tried to edit the wiki page but failed :-(

Why you failed, find the chunk? If so i can do it and change the text if i get the new text.

HLeithner:
> @Franzwohlkoenig

Welcome back? 🎉🎉🎉🎈🎈

> Why you failed, find the chunk? If so i can do it and change the text if i get the new text.

I already asked Sandra and she explained me how to edit this subset safesubst, hopefully I did it right... but thanks

HLeithner:
@PhilETaylor the feed back from the PD was to not activate shared sessions as default. So at the first set we should add a proper error message and help to user to understand why he or she can't preview the article and have to login to the frontend or have to activate shared sessions.

In the mid/long term a drafting system should be introduced into joomla (maybe as part of the workflow).

PhilETaylor:
> the feed back from the PD was to not activate shared sessions as default. 

Was there a documentable reason for this decision?

> So at the first set we should add a proper error message and help to user to understand why he or she can't preview the article and have to login to the frontend or have to activate shared sessions.

Cool. 

> In the mid/long term a drafting system should be introduced into joomla (maybe as part of the workflow).

 its bad, but this kind of draft is proposed....  https://github.com/joomla/joomla-cms/pull/35385

HLeithner:
It will be in the notes but no motion has been done.

I have seen the pr and talked to the students, sadly at the end of the course. I would have something different in mind but not ready to write about it.

Would you like and have the time to create the preview information 404 pr?

PhilETaylor:
> Would you like and have the time to create the preview information 404 pr?

on the back burner yes, this week Im hoping to get my teeth into the more important CloudFlare issue https://github.com/joomla/joomla-cms/issues/35244

PR #35525, opened 2021-09-09
Propose prompting to upload system information when logging issues.

PhilETaylor:
### Summary of Changes

The title says it all ... Propose prompting to upload system information when logging issues.

~Most~ ~some~ ~A high percentage of the time~ ~occasionally~ its the first thing we ask for from those new to posting issues in GitHub. Lets ask for it up front. 

You are free to ignore the prompt, delete it or leave it, it doesnt show in the final submitted issue

And if you disagree thats ok - why not go the whole hog and spell it out with an issue FORM? https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -11,8 +11,17 @@
 
 
 ### System information (as much as possible)
+<!-- 
+Please download your System information and upload with this GitHub Issue. 
+The download removes any secure/personal data so you can upload it in public here.
+
+Joomla 4: Log in to your Joomla Admin. Go to System in the left menu. 
+Then select "System Information". Click the button "Download as text". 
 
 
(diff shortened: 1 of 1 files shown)


brianteeman:
The _problem_ with that file is that it includes too much information. The form is a good idea though. However you also have to consider that users who post an issue on the issue tracker. Would they see the forms

PhilETaylor:
> The problem with that file is that it includes too much information.

I don't see that as a problem. 

> However you also have to consider that users who post an issue on the issue tracker. 

Its optional. 

> Would they see the forms [sic]

Forms are optional, in beta, and are just one of many different templates... There could be an additional template like the one we have now, so that when you want to post something freehand - you can.

PR #35527, opened 2021-09-10
Skip calling callback cache controller if caching is disabled

Denitz:
### Summary of Changes

No need to render a module via callback cache controller if caching is disabled.

### Testing Instructions

Enable Joomla caching, apply patch, see no changes except slight cached module rendering time improvement for logged users.

The module is rendered or the module data is loaded via callback cache without storing the data, even if caching is disabled.

The callback is executed directly.

No.


Fedik:
This method does not used when cache is disabled
https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Document/Renderer/Html/ModuleRenderer.php#L36

So your changes does not make sense.

Denitz:
@Fedik Enable conservative caching in Joomla, ensure that module setting Caching is enabled, log-in and see

Fedik:
That may happen if the module itself also use cache, because `$params->get('cache') == 1`.

HLeithner:

HLeithner:
@Denitz thanks for this pr but I think @Fedik is right and the module it self could be responsible for this.

I'm closing this for now.

PR #35531, opened 2021-09-11
Allow the translation bot to recompile the htmls

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
Do not exit if `composer install` was not executed and there is a system variable `TRANSLATION_BOT` with a value `YES`

Basically, this allows the translation bot to create the error pages by skipping the composer part

@wilsonge a simple review should be enough here, it solves the issue https://github.com/joomla/joomla-cms/pull/35379#issuecomment-917367450

### DIFF
--- a/build/build-modules-js/init/cleanup-media.es6.js
+++ b/build/build-modules-js/init/cleanup-media.es6.js
@@ -11,7 +11,7 @@ const RootPath = process.cwd();
  *
  * @returns {Promise}
  */
-module.exports.cleanVendors = async (skip = true) => {
+module.exports.cleanVendors = async () => {
   // eslint-disable-next-line no-console
   console.log('Cleanup the Vendor ');
 
--- a/build/build-modules-js/init/cleanup-media.es6.js
+++ b/build/build-modules-js/init/cleanup-media.es6.js
@@ -12,9 +12,9 @@ const RootPath = process.cwd();
  * @returns {Promise}
  */
 module.exports.cleanVendors = async () => {
-  if (process.env.TRANSLATION_BOT === 'YES') {
+  if (process.env.SKIP_COMPOSER_CHECK === 'YES') {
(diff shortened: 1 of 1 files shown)


HLeithner:
looks odd can we add parameter something like --no-vendor-cleanup ? so it's more generic?

build/build-modules-js/init/cleanup-media.es6.js
```diff
@@ -12,6 +12,12 @@ const RootPath = process.cwd();
  * @returns {Promise}
  */
 module.exports.cleanVendors = async () => {
+  if (process.env.TRANSLATION_BOT === 'YES') {
+    // eslint-disable-next-line no-console
+    console.log('Cleanup the Vendor skipped');
```
brianteeman:
```suggestion
    console.log('Cleanup the Vendor folder skipped');
```

dgrammatiko:
> looks odd can we add parameter something like --no-vendor-cleanup ? so it's more generic?

ENV vars can be used for the same result. If we go with the `--no-vendor-cleanup` option in the script we have to add one more procedure in the build.js, basically the ENV way is shorter. That said if you insist I can refactor the build.js to accept `--no-vendor-cleanup` and do the same

HLeithner:
ok then ENV but please change the name ;-)

dgrammatiko:
> ok then ENV but please change the name ;-)

Give me an alternative, I'm not good at naming things

How about `SKIP_COMPOSER_CHECK`?

HLeithner:
Thanks

HLeithner:
@dgrammatiko seems not to work https://github.com/joomla-translation-bot/joomla-cms/runs/3575595533?check_suite_focus=true

dgrammatiko:
@HLeithner https://github.com/joomla-translation-bot/joomla-cms/pull/4

the value is case sensitive

PR #35339, opened 2021-08-24
[4] Fix spinner bg

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/35334

### Summary of Changes

dont hardcode color

remember to `nom run build:js`

### Testing Instructions

slow down the checking of `fetchExtensionCompatibility` method by putting in a `sleep(10);` in the method.

look at the icons - click the tab on and off checking the colour - after 10 seconds see triangle has white bg

### Actual result BEFORE applying this Pull Request

68747470733a2f2f6973737565732e6a6f6f6d6c612e6f72672f75706c6f6164732f312f66303337646639316665316333646132303832343835636466316235393438612e676966

### Expected result AFTER applying this Pull Request

Screenshot 2021-08-24 at 11 04 37
Screenshot 2021-08-24 at 11 04 34
Screenshot 2021-08-24 at 11 06 50

PR #35542, opened 2021-09-12
[4.0] Fix numerics

brianteeman:
This PR fixes the bug reported in #35539 **AND** removes the Filter Numeric Terms option and related code as it is not complete and doesnt do anything at all


administrator/components/com_finder/src/Indexer/Query.php
```diff
@@ -1283,11 +1283,6 @@ protected function processString($input, $lang, $mode)
 					continue;
 				}
 
-				if ($params->get('filter_numerics', 0) && $token->numeric)
```
Quy, line 1286: There are 5 instances without `s` in this file.:

https://github.com/joomla/joomla-cms/blob/32f6c81fc2f1d35bec3853f345a93421d93dba6f/administrator/components/com_finder/src/Indexer/Query.php#L1010

brianteeman: even more proof that this code never worked and was never tested by the original submitter and it should just be removed

alikon:
i'm lost...
what's the relation between #35339 and this ?

brianteeman:
oops wrong PR - updated now

administrator/components/com_finder/src/Indexer/Indexer.php
```diff
@@ -956,7 +950,11 @@ protected function addTokensToDb($tokens, $context = '')
 				++$values;
 			}
 
-			$db->setQuery($query)->execute();
+			// Only execute the query if there are tokens to insert
```
PhilETaylor, line 953: This change seems to be the same as #35540 ?

richard67:
@brianteeman Now as PR #35540 has been merged, I would expect this PR here to show a conflict because it contains the same change as in PR #35540 . But it doesn't show any conflict on GitHub, and it still shows the change from PR #35540 , too. Could you update your branch to the 4.0-dev of the CMS to solve that? Or if you don't have the time, allow me to do that? I'd like that to be solved before I test.

brianteeman:
.

richard67:
@brianteeman Did you close it by purpose, and will you make a new one? Maybe I can try to fix our branch, but the easiest thing would be to make a new one, I think.

Issue #35543, opened 2021-09-13
[4.0] Category custom fields marked as 'subform' act as if they are not


crystalenka:
### Steps to reproduce the issue
1. Create a category custom field (text, editor, or media) marked as required and as subform only. 
2. If text or editor is your field type of choice, set the field to filter Safe HTML, alphanumeric, or telephone.
3. Go to any content category.
4. The field does not render (as expected).
5. Try to save the category.

### Expected result
The category saves as normal, since the field is subform only and not included in a subform assigned to this category.

### Actual result
The category errors on save, asking for the required field.
<img width="468" alt="Screen Shot 2021-09-13 at 11 13 47" src="https://user-images.githubusercontent.com/9141288/133048339-9e6d8ea7-72da-4020-8da4-8d6c0a679319.png">

### System information (as much as possible)
| Info | Detail |
| --- | --- |
| Joomla! Version | Joomla! 4.0.2 Stable [ Furaha ] 24-August-2021 19:54 GMT |
| PHP Version | 7.4.12 | 

### Additional comments
Additionally, jcfields loads the field in the front end even if there is no content and it is not assigned to a subform.
<img width="582" alt="Screen Shot 2021-09-13 at 11 12 04" src="https://user-images.githubusercontent.com/9141288/133048441-b5f8f33f-8b2c-4fa0-b47f-972592a0a7e4.png">

crystalenka:
Sorry, one more detail: (I will update above also)

Set the field to filter Safe HTML, alphanumeric, or telephone. (Every other filter works fine)

nikosdion:
I can confirm the issue and I have figure out why it happens. This is a two part issue. I will make a PR to solve both parts.

## Part I. “Only Use In Subform” doesn't do anything

The new “Only Use In Subform” option in a Custom Field's definition not doing what it's supposed to do.

Supposedly, custom fields with this option enabled would be excluded from everything _except_ defining or rendering a subform custom field.

In reality, this option does nothing useful. It displays a badge in the fields list. Oh, yeah, it also makes the Custom Fields where this option is enabled have a category list of `[-1]`. That's the beginning, middle and end of it. Totally useless.

## Part II. `PlgSystemFields::onContentPrepareForm` doesn't actually work form com_categories

[...]

PR #35476, opened 2021-09-04
[4.0] Allow workflow plugins to define which components they support

bembelimen:
### Summary of Changes
Currently the ```isSupported``` check for workflow plugins let plugins only hook into components. For example if you want to add another workflow plugin you have to write most of the ```isSupported``` method by yourself. With this change plugins cannot only hook into component functions (like core.state) but can also define by themself if they support a component (by setting ```protected $supportedExtensions```).

### Testing Instructions
- Install the following plugin: 
[ploxreplacer.zip](https://github.com/joomla/joomla-cms/files/7109114/ploxreplacer.zip)
- Activate it
- Activate the workflow
- Execute a transition

### Actual result BEFORE applying this Pull Request
Nothing happens, plugin is not executed

### Expected result AFTER applying this Pull Request
A blue message appears: "Works" as the plugin itself could define that it wants to support com_content.article

Also test:
- publish/feature workflow plugins work like before.

PR #33096, opened 2021-04-11
[4.0] Subform UX

joomdonation:
Pull Request for Issue #28389.

### Summary of Changes
Joomla! has a custom field type called **Subfields** and recently, renamed to **Subform** in a different PR. From it's documentation [https://docs.joomla.org/Custom_fields_type:_Subfields](https://docs.joomla.org/Custom_fields_type:_Subfields)

> the aim of it is to provide the user the possibility to create a repeatable custom field for an item with multiple related fields

There are some UX issues with that kind of custom field as described below (see https://github.com/joomla/joomla-cms/issues/28389 for detailed discussion):

- On Fields Management screen, there is no way to know if a custom field is a standard field or it is a subfield (part of a **Subform** custom field type).
- To prevent a custom field from being used both as standard field and as part of a **Subform**, we need to set **Category** of that custom field to **None**. It's hard to know about that.
- There is no filter option to allow seeing custom fields which are being used as part of **Subform**.
- There is no option to limit maximum number of rows which can be added when adding data for a **Subform**. 
- Users won't know that they must add some standard fields before they can add a **Subform**.

This PR tries to solve these issues base on suggestions from @AndySDH. See https://github.com/joomla/joomla-cms/issues/28389#issuecomment-816878826 for details (one of the suggestion is renaming the custom field type from Subfields to Subform but it is handled by different PR) 

1. Add a new setting **Use In Subform** on add/edit custom field screen to indicate that the field will be used in **Subform**.

- When it set to Yes, categories selection will become hidden.
- Category of the field will become **None**.

2. Add a filter to search tool to allow filtering fields belong to Subform (Yes/No). By default, all fields will be show.

filter

3. If you add a **Subform** custom field but forgot to create at least one standard field before , there will be a warning displayed at the top to tell you about that.
warning

4. On **Custom Fields Management** screen, add a small badge next to the field so that we know if it is a field which is used for **SubForm Only**.

subform_only

5. Add a new setting called **Maximum Rows** to control maximum number of rows which you can add while adding data for that **Subform** custom field type.

### Testing Instructions
1. Install Joomla 4 nightly build. Or if you are having an existing Joomla 4 installation, please update that installation to latest Joomla 4 nightly build.
2. Download update package for this PR at https://ci.joomla.org/artifacts/joomla/joomla-cms/4.0-dev/33096/downloads/42296/Joomla_4.0.0-beta8-dev+pr.33096-Development-Update_Package.zip , go to **System -> Update -> Joomla**, upload and install that update package
[...]

PR #35547, opened 2021-09-13
Category custom fields marked as 'subform' act as if they are not

nikosdion:
Pull Request for Issue #35543 .

### Summary of Changes

Category custom fields marked as “Use Only In Subform” were loaded as regular fields in all categories. This is a two–part bug (see Technical Notes at the bottom).

As a result I made the following changes:
* I changed `\Joomla\Component\Fields\Administrator\Helper\FieldsHelper::getFields` to only load non–subform–only fields by default with a new option to return all fields, including the subform–only, and made the necessary changes to SubfieldsField and PlgFieldsSubform.
* I changed PlgSystemFields::onContentPrepareForm to load the missing form data when you are saving a com_content category.

### Testing Instructions

* In the backend of your site go to Content, Fields.
* From the dropdown select “Category” and click on New.
* Create a new field with the following parameters:
   - Title: Test
   - Label: Test
   - Required: Yes. **THIS IS IMPORTANT**
   - Only Use In Subform: Yes. **THIS IS IMPORTANT**
   - Filter: Safe HTML. **THIS IS IMPORTANT**
* Click on Save & Close
* Go to Content, Categories
* Edit _any category_
* Click on Save

You get an error that the Test field is required.

No error; the category is saved correctly.

None. I mean, it would be nice if someone documented the “Only Use In Subform” field as now there's no tooltip and no documentation. What does this do? Nobody knows. Even looking at the code it was not obvious, presumably because it did nothing at all to begin with...

### Backwards compatibility notes

The `\Joomla\Component\Fields\Administrator\Helper\FieldsHelper::getFields` method has changed so that, by default, it does not return ALL fields but only those not marked as “Only Use In Subform”. Is this a b/c break? I would say no, because the intent of this option was to make these fields unavailable anywhere EXCEPT the subform custom field. Fixing a bug so that something works as it should, per the meager code documentation, shouldn't be considered a b/c break.

If it is to be considered a b/c break then the “Only Use In Subform“ feature should be removed from 4.0 as it cannot be fixed otherwise (well, just part II of the fix will kinda–sorta fix it but it's an ancillary artefact at best). Then again, removing a feature — even one that never worked — would be a b/c break.

So I am going to ping @wilsonge and @zero-24 and let them sort out the decision–making process of what constitutes a backward compatibility break in this context.

### Technical notes

This is a two part issue.

#### Part I. “Only Use In Subform” doesn't do anything

The new “Only Use In Subform” option in a Custom Field's definition not doing what it's supposed to do.

Supposedly, custom fields with this option enabled would be excluded from everything _except_ defining or rendering a subform custom field.

In reality, this option does nothing useful. It displays a badge in the fields list. Oh, yeah, it also makes the Custom Fields where this option is enabled have a category list of `[-1]`. That's the beginning, middle and end of it. Totally useless.

I changed `\Joomla\Component\Fields\Administrator\Helper\FieldsHelper::getFields` to only load non–subform–only fields by default. If you want to get all fields, including subform–only, you need to pass `true` to a new fifth parameter. Note that the ONLY use case for retrieving all fields, including subform–only, is the SubfieldsField and the subform field itself. Third–party software SHOULD NOT receive those fields outside of the context of a subform custom field, therefore this is not a b/c break.

#### Part II. `PlgSystemFields::onContentPrepareForm` doesn't actually work for com_categories (when saving)

You will see that the system plugin **ALLEGEDLY** has special handling for editing categories in com_categories. Allegedly, because it works on the edit page but NOT when saving! When saving all fields from all categories are loaded since the plugin runs before the form data is loaded from the request, meaning there is no category filter present.

You see, `$data` is an empty array when you the `onContentPrepareForm` event is fired _when saving the form_. I see that  the choice to not populate the form data when loading the form in `\Joomla\CMS\MVC\Controller\FormController::save` dates back to 2010 when Andrew made a commit about fixing some unspecified issue saving plugins.

The problem is that for com_categories _and only this component_ we actually need to load the form data from the request when saving data. Otherwise PlgSystemFields:: onContentPrepareForm sees that $data is empty and doesn't set the category ID. As a result `FieldsHelper::prepareForm` populates `$assignedCatids` to null, therefore it loads all fields from all categories. If any of these fields is required AND has a validator which doesn't accept NULL values — like the ones @crystalenka mentioned — you get a save error.

There are a few options to fix it. One is to override CategoryController::save. However, this results in severe code duplication which will be a pain in the posterior to manage whenever the base FormController::save method is updated.

The other option is to force CategoryModel::getForm to always load the data. However, this may break b/c; I'm especially concerned about third party plugins which may use the empty $data as a signal when figuring out which fields to add to the category edit view.

The final option is to change PlgSystemFields::onContentPrepareForm itself. Since we're doing special handling for com_categories we might just as well populate `data` from the request if we are saving the form. This is what I plan on doing. To keep things b/c this nasty workaround will ONLY apply to com_content categories, not third party extensions' categories. In fact, custom fields SHOULD only apply to com_content categories — which is not the case now — but that's a different bug for a different day, _n'est ce pas_?

### DIFF
--- a/administrator/components/com_fields/src/Helper/FieldsHelper.php
+++ b/administrator/components/com_fields/src/Helper/FieldsHelper.php
@@ -108,13 +108,13 @@ class FieldsHelper
 			self::$fieldsCache->setState('list.limit', 0);
 		}
 
-		if (!$includeSubformFields)
+		if ($includeSubformFields)
 		{
-			self::$fieldsCache->setState('filter.only_use_in_subform', 0);
+			self::$fieldsCache->setState('filter.only_use_in_subform', '');
 		}
 		else
 		{
-			self::$fieldsCache->setState('filter.only_use_in_subform', '');
+			self::$fieldsCache->setState('filter.only_use_in_subform', 0);
 		}
 
 		if (is_array($item))
--- a/administrator/components/com_fields/src/Helper/FieldsHelper.php
+++ b/administrator/components/com_fields/src/Helper/FieldsHelper.php
@@ -96,7 +96,7 @@ class FieldsHelper
 	 *
 	 * @since   3.7.0
 	 */
-	public static function getFields($context, $item = null, $prepareValue = false, array $valuesToOverride = null)
+	public static function getFields($context, $item = null, $prepareValue = false, array $valuesToOverride = null, bool $includeSubformFields = false)
 	{
 		if (self::$fieldsCache === null)
 		{
@@ -106,6 +106,15 @@ class FieldsHelper
 
 			self::$fieldsCache->setState('filter.state', 1);
 			self::$fieldsCache->setState('list.limit', 0);
+		}
+
+		if (!$includeSubformFields)
+		{
+			self::$fieldsCache->setState('filter.only_use_in_subform', 0);
+		}
+		else
+		{
+			self::$fieldsCache->setState('filter.only_use_in_subform', '');
 		}
 
 		if (is_array($item))
--- a/administrator/components/com_fields/src/Helper/FieldsHelper.php
+++ b/administrator/components/com_fields/src/Helper/FieldsHelper.php
@@ -87,16 +87,20 @@ class FieldsHelper
 	 * The values of the fields can be overridden by an associative array where the keys
 	 * have to be a name and its corresponding value.
 	 *
-	 * @param   string     $context           The context of the content passed to the helper
-	 * @param   \stdClass  $item              item
-	 * @param   int|bool   $prepareValue      (if int is display event): 1 - AfterTitle, 2 - BeforeDisplay, 3 - AfterDisplay, 0 - OFF
-	 * @param   array      $valuesToOverride  The values to override
+	 * @param   string      $context              The context of the content passed to the helper
+	 * @param   null        $item                 The item being edited in the form
+	 * @param   int|bool    $prepareValue         (if int is display event): 1 - AfterTitle, 2 - BeforeDisplay, 3 - AfterDisplay, 0 - OFF
+	 * @param   array|null  $valuesToOverride     The values to override
+	 * @param   bool        $includeSubformFields Should I include fields marked as Only Use In Subform?
 	 *
 	 * @return  array
 	 *
+	 * @throws \Exception
 	 * @since   3.7.0
 	 */
-	public static function getFields($context, $item = null, $prepareValue = false, array $valuesToOverride = null, bool $includeSubformFields = false)
+	public static function getFields(
+		$context, $item = null, $prepareValue = false, array $valuesToOverride = null, bool $includeSubformFields = false
+	)
 	{
 		if (self::$fieldsCache === null)
 		{
--- a/administrator/components/com_fields/src/Field/SubfieldsField.php
+++ b/administrator/components/com_fields/src/Field/SubfieldsField.php
@@ -68,7 +68,7 @@ class SubfieldsField extends ListField
 		// Check whether we have a result for this context yet
 		if (!isset(static::$customFieldsCache[$this->context]))
 		{
-			static::$customFieldsCache[$this->context] = FieldsHelper::getFields($this->context);
+			static::$customFieldsCache[$this->context] = FieldsHelper::getFields($this->context, null, false, null, true);
 		}
 
 		// Iterate over the custom fields for this context
--- a/plugins/fields/subform/subform.php
+++ b/plugins/fields/subform/subform.php
@@ -388,7 +388,7 @@ class PlgFieldsSubform extends FieldsPlugin
 			static::$customFieldsCache = array();
 
 			// Get all custom field instances
-			$customFields = FieldsHelper::getFields('');
+			$customFields = FieldsHelper::getFields('', null, false, null, true);
 
 			foreach ($customFields as $customField)
 			{
--- a/plugins/system/fields/fields.php
+++ b/plugins/system/fields/fields.php
@@ -265,11 +265,7 @@ class PlgSystemFields extends CMSPlugin
 		if (strpos($context, 'com_categories.category') === 0)
 		{
 			$context = str_replace('com_categories.category', '', $context) . '.categories';
-
-			if ($context === 'com_categories.categorycom_content')
-			{
-				$data = $data ?: Factory::getApplication()->input->get('jform', [], 'array');
-			}
+			$data    = $data ?: Factory::getApplication()->input->get('jform', [], 'array');
 
 			// Set the catid on the category to get only the fields which belong to this category
 			if (is_array($data) && array_key_exists('id', $data))
--- a/plugins/system/fields/fields.php
+++ b/plugins/system/fields/fields.php
@@ -266,6 +266,8 @@ class PlgSystemFields extends CMSPlugin
 		{
 			$context = str_replace('com_categories.category', '', $context) . '.categories';
 
+			$data = $data ?: Factory::getApplication()->input->get('jform', [], 'array');
+
 			// Set the catid on the category to get only the fields which belong to this category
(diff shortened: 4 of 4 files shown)


administrator/components/com_fields/src/Helper/FieldsHelper.php
```diff
@@ -108,6 +108,15 @@ public static function getFields($context, $item = null, $prepareValue = false,
...
 		}
 
+		if (!$includeSubformFields)
+		{
+			self::$fieldsCache->setState('filter.only_use_in_subform', 0);
+		}
+		else
+		{
+			self::$fieldsCache->setState('filter.only_use_in_subform', '');
+		}
```
richard67:
```suggestion
		if ($includeSubformFields)
		{
			self::$fieldsCache->setState('filter.only_use_in_subform', '');
		}
		else
		{
			self::$fieldsCache->setState('filter.only_use_in_subform', 0);
		}
```
Couldn't we use positive logic here?

nikosdion: Either way works for me 🤷🏽‍♂️

richard67: It's nicer with positive logic according to school books ;-)

nikosdion: I prefer to have the code that applies the non–default filter first. This makes it far less likely that someone will see it and go “oh, this parameter and if–block are redundant, we are just setting the filter to its default value”, remove it and break stuff because they didn't actually read the code or run a git blame to see WTF this code is there. It may or may not have happened before...

nikosdion: BTW I did merge your suggestion or at least I told GitHub to do so.

richard67: Yes, you merged it. But if you think it was wrong and your original code was better to understand or to read, then change it back. It was just a suggestion, and as a human being I might be wrong with a suggestion sometimes. It's just school books say positive logic shall be preferred because normally that's better readable.

nikosdion: Nah, it's all good. Let's keep it the way you proposed. It's probably more readable for people who are not like me.

Being neurodiverse I read code differently than most people. I don't care about inverted or negative logic. Same for double negatives. They don't trip me over. What trips me over to some extent is if the “interesting“ branch of an if–block is not at the top. It slows me down reading that code. I also recognise that this is NOT how most people read code — I taught other people to code about a decade ago, a very informing and humbling experience for me. So, when it's not code that I am most likely to mostly read and work on myself I am happy to accommodate the majority of people who will be reading and working on it. In the end of the day FOSS is about helping _other people_ with our code, right? 😄

crystalenka:
I have tested this item :white_check_mark: successfully on 5879d5690ec4e91665e75c80e5e195a85c434e51<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35547">issues.joomla.org/tracker/joomla-cms/35547</a>.</sub>

administrator/components/com_fields/src/Helper/FieldsHelper.php
```diff
@@ -108,6 +108,15 @@ public static function getFields($context, $item = null, $prepareValue = false,
 			self::$fieldsCache->setState('list.limit', 0);
 		}
 
+		if ($includeSubformFields)
+		{
+			self::$fieldsCache->setState('filter.only_use_in_subform', '');
```
Fedik, line 117: Should not this be `1` instead of empty `''`? 
https://github.com/joomla/joomla-cms/blob/d151fd31af85b93fc84378c1bd1643d60bd03b47/administrator/components/com_fields/src/Model/FieldsModel.php#L352-L359

The field `only_use_in_subform` itself is a radio button with `0` and `1`

Fedik:
Looks good to me.

The `filter.only_use_in_subform` was add in #33096 , 
@joomdonation can you please have a look, I think  you forgot to add the `set('filter.only_use_in_subform')` for it or?
The Article does not have this bug because the fields always filtered by `filter.assigned_cat_ids`, however for category this filter always empty.

administrator/components/com_fields/src/Helper/FieldsHelper.php
nikosdion: No. If you use 1 it will return **only** the fields set up as only use in subform. Subforms must have access to all fields, no matter if they were marked as only in subform or not. That's why I use an empty string.

Everywhere else we must ONLY return the fields which are NOT subform only. This is why in the other branch of the if-block I set the filter to 0.

Remember that all publish-style fields which accept an integer will filter records where the respective field is this integer. However, they also have an empty state. When the field is an empty string no filtering will be applied.

Think about a published filter field. If it's an empty string you show all records, published or unpublished. If it's 0 you show unpublished records. If it's 1 you show published records. The same thing applies here.

Are we seriously discussing how Joomla has been working since it was called Mambo...?

Fedik:
I have tested this item :white_check_mark: successfully on 5879d5690ec4e91665e75c80e5e195a85c434e51<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35547">issues.joomla.org/tracker/joomla-cms/35547</a>.</sub>

administrator/components/com_fields/src/Helper/FieldsHelper.php
```diff
@@ -96,7 +96,7 @@ public static function extract($contextString, $item = null)
 	 *
 	 * @since   3.7.0
 	 */
-	public static function getFields($context, $item = null, $prepareValue = false, array $valuesToOverride = null)
+	public static function getFields($context, $item = null, $prepareValue = false, array $valuesToOverride = null, bool $includeSubformFields = false)
```
Fedik: Please update doc block. Add description  for a new parameter

nikosdion: Done!

Fedik:
RTC
It had 2 tests, and last change was a comment correction<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35547">issues.joomla.org/tracker/joomla-cms/35547</a>.</sub>

Fedik:
@wilsonge still have to decide when where and whether  merge it ;)

wilsonge:
Joyfully since PHP 8 this counts as a b/c break because method signatures aren't allowed to change if people are subclassing example: https://3v4l.org/iedIW . This is also causing issues with other PRs such as https://github.com/joomla/joomla-cms/pull/35476 (where I stole the example link from).

I think it's fairly unlikely that people are subclassing fieldshelper to be honest so my gut is to merge this. However I'm going to leave this here a few days to see if anyone else has any strong objections or feelings on the matter.

nikosdion:
I'd be surprised if anyone subclasses that. The way it's constructed it's just a magic static helper you can't really modify without hacking core — or making a PR :trollface:

wilsonge:
No complaints imminent. Merged. Thanks!

nikosdion:
Thank you!

richard67:
Is seems there has been something forgotten, see #35924 .

janschoenherr:
Hi, can one of you guys please review this PR: https://github.com/joomla/joomla-cms/pull/35924

AndySDH:
Thank you @nikosdion for making this, a great fix! This also fixed the same issue for the Users component! (as they have no category either, so they ran into the same problem)

PR #35574, opened 2021-09-16
[4.0] Avoid exception on wrong validation code and just redirect to login

bembelimen:
### Summary of Changes
I got some reports from user, that they land on the error page when they try to validate their account. When I check their account they are already verified and activated. The issue is, that some email application pre-fetch urls (and show e.g. previews) which triggers the validation and if then the user clicks, Joomla! throws an exception.

### Testing Instructions
- Set registration to user activation
- Register a new user
- click on the verification link
- click again on the verification link

grafik

User is redirected to the login (with the same message as normal message)


PhilETaylor:
Funny, I'll find the link in a moment, but when I reported this, I was told it was not an issue...

This is the wrong fix though. 

A non-form-token'ed, GET request should not be allowed to validate a users account. 

We would never allow this anywhere else. We would never allow a GET request to perform an action like this. 

Moving the verification to a POST, with the user having to click the link to prefil a form, and then click a form submit button, as a POST is the correct fix for this issue.

PhilETaylor:
Also your screenshot shows this is an issue in Joomla 3 - yet this PR only addresses it in Joomla 4, it would need to be applied to Joomla 3 and then merged up to Joomla 4.

bembelimen:
I see your point, but think the the "common user" expects to just click in the email link and that's it.

The Screenshot is from J! 4.

PhilETaylor:
The "common user" expects Joomla to implement things securely, according to established security best practice and norms.

Regardless of the screenshot, if this is an issue in Joomla 3 also then it should be first addressed there.

PhilETaylor:
I have tested this item :white_check_mark: successfully on 25b18a23958c148d81cbe34018145fe4e37dbf22<br /><br />Marking this as successful as, as it handles the exception with a redirect as designed. 

However, what seems to be out of scope of this PR, and is a security issue @joomla/security is the changing of a security credential (account validation) by GET request, which should be addressed, as nothing should change when making GET requests, which are designed for page retrieval and not updating a user object. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35574">issues.joomla.org/tracker/joomla-cms/35574</a>.</sub>

PhilETaylor:
For the record this was reported to the Joomla security strike team, via their email address, however I did not receive any reply to my report to date￼

jwaisner:
I have tested this item :white_check_mark: successfully on 25b18a23958c148d81cbe34018145fe4e37dbf22<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35574">issues.joomla.org/tracker/joomla-cms/35574</a>.</sub>

jwaisner:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35574">issues.joomla.org/tracker/joomla-cms/35574</a>.</sub>

components/com_users/src/Controller/RegistrationController.php
```diff
@@ -67,7 +67,12 @@ public function activate()
 
 		if (!$userIdToActivate)
 		{
-			throw new \Exception(Text::_('COM_USERS_ACTIVATION_TOKEN_NOT_FOUND'), 403);
+			$loginUrl = 'index.php?option=com_users&view=login';
```
zero-24:
```suggestion
```

components/com_users/src/Controller/RegistrationController.php
```diff
@@ -67,7 +67,12 @@ public function activate()
...
+
```
zero-24:
```suggestion
```

components/com_users/src/Controller/RegistrationController.php
```diff
@@ -67,7 +67,12 @@ public function activate()
...
+			$this->setMessage(Text::_('COM_USERS_ACTIVATION_TOKEN_NOT_FOUND'));
+			$this->setRedirect(Route::_($loginUrl, false));
```
zero-24:
```suggestion
			$this->setRedirect(Route::_('index.php?option=com_users&view=login', false));
```

PhilETaylor:
forked to https://github.com/joomla/joomla-cms/issues/37172

PR #35590, opened 2021-09-17
[4.0.x] Update skipto

brianteeman:
This updates the skipto js to the newly released 4.1.2 - nothing spectacular changed apart from fixing an issue with WAVE tester making a false error report

### Version 4.1.2
Added aria-busy="true" attribute to menu element when SkipTo is initialized and being updated with new menu items to support validators looking for required menu items for the menu role.
Added the optional aria-controls attribute to button element to reference the id of the menu element as defined in the W3C ARIA Authoring practices for menu button pattern.

### Version 4.1.1
Removed aria-describedby from button, since screen readers read the accesskey information.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,10426 @@
 {
   "name": "joomla",
   "version": "4.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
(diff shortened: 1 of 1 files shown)


wilsonge:
Thanks!

brianteeman:
thanks

PR #35611, opened 2021-09-20
Remove the unnecessary space character in $item->text

shur:
Found a problem with the addition of an extra space character. For example, if the article contains only introtext (with several paragraphs) and no fulltext. Then `$item->text` will be 1 longer due to the added space character.

### Testing Instructions
`var_dump($article);`

```
  ["introtext"]=>
  string(780) "<p>***trick.</p>"
  ["fulltext"]=>
  string(0) ""
  ["text"]=>
  string(781) "<p>***trick.</p> "
```

Notice the line length 761 and now the line no longer ends with `</p>`.

### DIFF
--- a/api/components/com_content/src/View/Articles/JsonapiView.php
+++ b/api/components/com_content/src/View/Articles/JsonapiView.php
@@ -189,7 +189,7 @@ class JsonapiView extends BaseApiView
 	 */
 	protected function prepareItem($item)
 	{
-		$item->text = $item->introtext . ' ' . $item->fulltext;
+		$item->text = $item->introtext . $item->fulltext;
 
 		// Process the content plugins.
 		PluginHelper::importPlugin('content');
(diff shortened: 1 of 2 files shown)


PhilETaylor:
I would argue this space is needed when concatenation of the intro and full text takes place else you end up with

…this is the end of the intro textThis is the start of the full text….

shur:
If it looks like this (and most often it does), then I see no problem.
`…this is the end of the intro text</p><p>This is the start of the full text….`

I am currently working on one content plugin that divides an article into parts by the `</p>` tag. And for a very long time, I could not understand where this space character comes from, when it is absent in the database.

While saving introtext and fulltext in the database, we use `trim` several times to remove all unnecessary ... and then add this space.
Hmm, is it really all right here? and it doesn't need to be reconsidered?

PhilETaylor:
You also assume that intro/full text are both HTML - which is not always (but mostly is) the case.

Issue #35570, opened 2021-09-15
[4.0] itemid is overruled when system - sef is enabled


810:
### Steps to reproduce the issue

1) Setup joomla
2) Install Kunena 6.0 nightly build (https://www.kunena.org/download)
3) Enable kunena forum
4) Create a topic
5) Check the item id on the message action reply button.

### Expected result

`http://localhost/j4/index.php?option=com_kunena&view=topic&catid=2&id=37&mesid=42&Itemid=151`

### Actual result

`http://localhost/j4/index.php?option=com_kunena&view=topic&catid=2&id=37&mesid=42&Itemid=101`

provided url on components/com_kunena/template/aurelia/layouts/widget/button/default.php =
 `index.php?option=com_kunena&view=topic&layout=reply&catid=2&id=37&Itemid=151`

### System information (as much as possible)

Nightly joomla 4.0
Nightly Kunena 6.0

Extra info:

Default joomla menu, where the home is set as itemid 101
Kunena menu itemid = 151

brianteeman:
What do you mean by provided url

810:
the input value.  $this->url

brianteeman:
So not hardcoded as you wrote but using the api and being passed through jroute

810:
when i use hardcoded is the same:

`<a href="index.php?option=com_kunena&view=topic&layout=reply&catid=2&id=37&Itemid=151">
	test
</a>`

Result: 
`<a href="/j4/index.php?option=com_kunena&amp;view=topic&amp;layout=reply&amp;catid=2&amp;id=37&amp;Itemid=101">
	test
</a>`

Ruud68:
I have been fighting this (I think) in trying to get one of my components ready for both J3 and J4 based on the new modern router.
J3 and J4 router behaves differently: the code returns a different SEF URL on J3 and on J4. IMO this means that the new / modern router as implemented in J4 is not B/C.

I have drilled it down to the different handling of the Itemid.

My workaround is to do a check on J version and based on that set / unset the Itemid in the preprocess function of the MenuRules.

richard67:
Closing as having a pull request, see #35634 . @810 Please when you make a PR to solving an issue created by you in future, then close that issue when the PR is ready for testing. Thanks in advance.

Hackwar:
The PR #35634 was closed since there was no response there. However we have a new PR #42989 which should help with the issue as well.

PR #18260, opened 2017-10-06
If the link is the same as active menu item then use it

csthomas:
Pull Request for Issue #17719 and #16774

### Summary of Changes
If link Itemid is equal to active menu and link query is equal to active query then use it, do not try to find a better match.
Add unit tests to prove that.

### Testing Instructions
Use tests from  #17719 or #16774 or this below.

1. Turn off Search Engine Friendly URLs in Global Configuration
2. Create menu item (Style 1) to an article, save, then save as copy (Style 2), publish and save and close.
3. Go to front-page and click on the first one (Style 1), check both URLs, it is OK.
4. Go to second (Style 2), and check link at (Style 2), the Itemid in URL point to (Style 1) instead to itself (Style 2).

### Expected result
Active second menu item (Style 2) should have URL to (Style 2) 

### Actual result
Active second menu item (Style 2) use first menu item URL (Style 1) 

### Documentation Changes Required
No

PR #35634, opened 2021-09-21
[4.0] itemid is overruled when system - sef is enabled #35570

810:
Pull Request for Issue #35570 .

### Summary of Changes
Allow itemid from other menu then from the main menu. Now its getting overruled by the default menu item.

### Testing Instructions

Apply patch.

Add hardcode html:

`<a href="index.php?option=com_kunena&view=topic&layout=reply&catid=2&id=37&Itemid=151">
	test
</a>`

`<a href="/j4/index.php?option=com_kunena&amp;view=topic&amp;layout=reply&amp;catid=2&amp;id=37&amp;Itemid=101">
	test
</a>`

`<a href="/j4/index.php?option=com_kunena&amp;view=topic&amp;layout=reply&amp;catid=2&amp;id=37&amp;Itemid=151">
	test
</a>`


PhilETaylor:
The comments being removed were added in https://github.com/joomla/joomla-cms/pull/12020 by @wilsonge and the code is from https://github.com/joomla/joomla-cms/pull/18260 from @csthomas 

It might be useful to get their comments on this change, as it seems the code being removed/changed has history and has a lot of thought put into it previously. 

It would also be worth checking those two PR's above, and ensuring that this PR doesnt reintroduce the bugs those fix.

810:
Any comments? I don't know why you always needs the active menu. Thats only needed, when have a simple website. 

If you have complicated website with multiple menus, then this "feature" (active menu item matches the requested query)  doesn't make any sense for me.

xillibit:
The line before the change in this PR is identical on J! 3.10.x, the issue was in Kunena, i have found a weird change in KunenaRoute which have reverted it which solve the issue. But we still lose the Kunena menu when we are on views which use pagination and when going to anther page

HLeithner:

bembelimen:
So we can't just use any give Itemid.

I agree that we should not throw away valid Itemids but we have to check if they're valid. Could you please improve your PR in that regards?

HLeithner:

HLeithner:

PR #35518, opened 2021-09-09
[4.0] Update Bootstrap to 5.1.1

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Update to 5.1.1
- refactor the root colours to use the custom ones
- remove the empty mixin file and all the calls to it

### Testing Instructions
Apply the patch
Run `npm ci`

Check that ALL 3 templates are ok, not broken **AND** customising the colours still works as expected

### Documentation Changes Required

NO

Some short explanations of what's happening here:
- We override the _root.scss of the bootstrap
- In the variables, we place all the Atum specific variables in an array `$atum-colors`
- The $body-color is removed from the variables as it's already converted to CSS VAR in the bootstrap source https://github.com/twbs/bootstrap/blob/1df098361cac04217d6a464c80e890c4335ecb5c/scss/_root.scss#L47 and we assign it to the Atum custom CSS Var
- The root loops the `$atum-colors` and appends the CSS VARS to the `:root`
- The mixin part was just a cleanup

PR #35639, opened 2021-09-22
[5.2] Adding Color Picker re-re-based

thednp:
Due to some complications, [the previous commit](https://github.com/joomla/joomla-cms/pull/32508) didn't work.

Pull Request for Issue #32494.

Demo page of the color-picker component [right here](https://codepen.io/thednp/pen/yLVzZzW).

### Summary of Changes
* removed most `minicolors` assets
* added `tinycolor` asset
* added the new `joomla-field-color-picker.es6.js` as described in #32494
* updated the `layouts/joomla/form/field/color/advanced.php` file
* updated 3 language files
* updated most assets related json files with the new assets
* updated the build scripts

### Testing Instructions

**JavaScript Testing**
The most important test for me, please report any error in the console along with the following info:
* Machine Operating System + Version
* Browser Name + Version

**Existing Extensions Testing**
* Open an extension that uses a  `<field type="Color"` `control="advanced">` field;
* Check if the `<color-picker>` updates the background of the visible input with the value stored in the database; if not the script is missing, not initialized or something happened, see below;
* If the above fails, check the browser console and check for any `ColorPicker` or `TinyColor` related errors; Report back here any error you encounter; I really hope there won't be;
* If however no error is reported in the browser console, click on the visible input to open the `.color-dropdown` then click anywhere outside of it, the `.color-dropdown` should close;
* Click on the visible input and pick a new color;
* Save the extension settings and check if the value is the one you have previously set in the above step;
* While the `.color-dropdown` element is visible try and resize the window, see if the console pops any error;
* While the window is under 768px wide and under 600px high, click the visible color input field and check where the `.color-dropdown` element is rendered (above or below the visible input), then scroll the window up/down while open to check if the `.color-dropdown` element should be re-positioned where it has the most available space;

**Color Picker Options Testing**
* Open `/templates/cassiopeia/templateDetails.xml`, find the `<fieldset name="advanced">` tag and paste inside the following code:

```html
<field
	name="advanced-color-hex"
	type="Color"
	format="hex"
	keywords="false"
	hint="Type in a valid hexadecimal colour value"
	default="#069"
	label="Text Color" />

<field
	name="advanced-color-rgb"
	type="Color"
	required="true"
	format="rgb"
	default="rgb(0,102,153)"
	label="Primary Color" />

<field
	name="advanced-color-hsl"
	type="Color"
	required="true"
	format="hsl"
	keywords="navy,lime,crimson,chocolate"
	default="hsla(265,100%,50%,.8)"
	label="Secondary Color" />
```

* `<field name="advanced-color-hex">` sets the `keywords="false"` which disables the usage of presets and restricts the color-picker to only use `format="hex"` colors (in this particular case); the reason for this behavior is the need to send SCSS compilers real and valid HEX/RGB(a)/HSL(a) color values to be used for mixins, eg. `darken($primary,15%)` (`darken('inherit',15%)` won't work); this field should use a placeholder;
* `<field name="advanced-color-rgb">` does not set the `keywords` option and will use the default keywords: `transparent`, `curentColor`, `initial` and `inherit`; as a result, the color-picker will display a button on the right side of the visible input to open a color options menu; these "non-color" values are valid CSS values; this field is required;
* `<field name="advanced-color-hsl">` sets the `keywords="navy,crimson.."` and enables the usage of user defined presets based on web safe color names; because none of the above "non-color" keywords is present in the setting value, color-picker will again restrict values to only valid HSL(a) format colors for the same known reasons.

**Keyboard Events Testing**
* While the `.color-dropdown` is displayed, hit the ESC key of your keyboard, it should close/hide the element;
* While viewing a Joomla form page with a `<field type="Color">`, try and navigate the page with your keyboard, when you focus a color-picker it should open its `.color-dropdown` and allow you to navigate inside;
* For the `<field type="Color" keywords="false">` delete the current value in the visible input first then try typing "inherit", "transparent" or "currentColor", it should set `#000000` value;
* For the `<field type="Color" keywords="navy,crimson..">` delete the current value in the visible input first then try typing "inherit", "transparent" or "currentColor", it should set a valid HEX/RGB(a)/HSL(a) value;
* For the `<field type="Color">` which uses no `keywords` option. delete the current value in the visible input first then try typing "inherit", "transparent" or "currentColor", it should set that value;
* Test the visible input(s) of the `.color-dropdown` (where format is either "rgb" or "hsl"), focus one of them and try and change values by using the up/down buttons of the keyboard, it should instantly update the color picker on every change;
* Test the visible input(s) of the `.color-dropdown` (where format is either "rgb" or "hsl"), focus one of them and try and **type** 0-255 for "R", "G" or "B" for the `<field type="Color" format="rgb">` field or 0-359 for "H", 0-100 for "S" or "L" for the `<field type="Color" format="hsl">` , it should update the color picker on every change after a short delay (TinyColor will validate that color first);
* For the `<field type="Color" format="hsl">`, if the "L" (Lightness component) is 100, the "S" (Saturation component) and the "H" (Hue component) both should be locked at 0, then decrease lightness to 99 and this will unlock Saturation control, ultimately if you set Lightness to less than 100 and Saturation to at least 1, this will ultimately unlock Hue control (this will be explained in the documentation).

**Testing the "onchange"**
The implementation is similar with fancy-select, so you as tester, as well as third party devs should have a familiar feeling about color picker. For any of your `<field type="Color" control="advanced">` make sure to also test "onchange":
* when typing values into inputs;
* when you slide the color-picker controls.

**Mobile Testing**
*  Tap the visible input to open the color-picker and check if the visible input, the `.color-dropdown` and the mobile keyboard all are displayed on the mobile device screen; the color-picker will decrease its height by around 25% to fit most mobile devices;
* Try and reproduce all the tests explained above except everything to do with checking the browser console; we're interested in the behavior of the color-picker, it should feel snappy and accurate; it should dismiss the opened `.color-dropdown` popup on taping outside the element itself;

**Accessibility Testing**
Check the labels of all the visible inputs, find any ARIA related issues, report back anything you think it's not good enough in terms of accessibility, ARIA and so on.

The main focus:
* while the `<joomla-field-color-picker>` element is focused, hit `Enter`/`Space` to toggle the color picker visibility
* while closed, try `Tab` key, it should go to the next form field
* while open, try `Tab` key, it should navigate all focusable elements
* while open, try changing the visible pointers/sliders as well as the visible text inputs with arrow keys
* open your screen reader and try all the above with the screen reader scanning all your values and everything ARIA related.

All `type="color"` fields that use `control="advanced"` are outdated in terms of accessibility, usability and jQuery dependency.

All your `type="Color"` fields that use `control="advanced"` are super awesome on any device and any user.

This [documentation section](https://docs.joomla.org/Color_form_field_type) mentions nothing about the control property/setting of the `Color` field, the following is a draft to provide guidance for this Joomla form field type as coded in `libraries\src\Form\Field\ColorField.php` in Joomla 4.0. Some parts of it need to be updated by their developer or some Joomla intern knowledgeable of this particular form field.

***Documentation Draft***

Provides a flexible form field that allow you to set a colour, with controls ranging from simple to advanced.

* **name** (mandatory) is the unique name of the parameter.
* **type** (mandatory) must be "Color".
* **default** (optional) provides a colour value when not set.
* **format** (optional) allows you to select the type of colour format, you can select "rgb", "rgba", "hsl", "hsla" or "hex".
* **control** (optional) allows you to select the type of controls the user need to set a colour, you can select "simple", "slider" or "advanced". The default is "advanced."
* **colors** (optional) allows you to set a list of colours to be used for the "simple" control. Shorthand HEX colors are not supported ("#fff"). Eg. `colors="#ff0000,#0000ff"`.
* **display** (optional) allows you to display a single component for the "slider" control. You can set "hue", "saturation", "light" or "alpha".
* **keywords** (optional) allows you to set color presets for the "advanced" control. The default value is `keywords="initial,inherit,currentColor,transparent"`, however you can disable the feature by using `keywords="false"` which will restrict the color picker to only use valid HEX, RGB(a) or HSL(a) values. If you include any of the default value (eg. "inherit") in the keywords list, it will remove the restriction.
* **label** (mandatory) (translatable) is the descriptive title of the field.
* **description** (optional) (translatable) allow you to provide more details for the form field.

**Note on HSL(a) Format and Advanced Control**

[TinyColor](https://github.com/scttcper/tinycolor) is the script behind the color picker of the "advanced" control. This script will understand that values like `hsl(0, 0%, 100%)` and `hsl(150, 0%, 100%)` are visually identical, it will use the `hsl(0, 0%, 100%)` value. In addition, if the **L** (Lightness component) is 100% then the **S** (Saturation component) and the **H** (Hue component) both should be locked at 0 in the inputs of the color picker; if you decrease Lightness to 99% it will unlock Saturation control, ultimately if you set Lightness to less than 100% and Saturation to at least 1%, this will unlock Hue control.

**Example XML Definition for Simple Control:**
```xml
<field name="backgroundcolor"
	type="Color"
	format="hex"
	default="#ff0000"
	control="simple"
	colors="#ff0000,#008000,#0000ff"
	label="TPL_BEEZ3_FIELD_HEADER_BACKGROUND_COLOR_LABEL"
	description="TPL_BEEZ3_FIELD_HEADER_BACKGROUND_COLOR_DESC" />
```

**Example XML Definition for Slider Control:** (this part need to be updated by its developer)
```xml
<field name="backgroundcolor" 
	type="Color" 
	default="#ff0000" 
	format="hex"
	control="slider"
	display="hue"
	label="TPL_BEEZ3_FIELD_HEADER_BACKGROUND_COLOR_LABEL"
	description="TPL_BEEZ3_FIELD_HEADER_BACKGROUND_COLOR_DESC" />
```

**Example XML Definition for Advanced Control:**
```xml
<field name="backgroundcolor" 
	type="Color" 
	format="rgb"
	control="advanced"
	default="rgb(250,0,0)"
	keywords="initial,inherit,transparent,navy,lime" 
	label="TPL_BEEZ3_FIELD_HEADER_BACKGROUND_COLOR_LABEL"
	description="TPL_BEEZ3_FIELD_HEADER_BACKGROUND_COLOR_DESC" />
```


thednp:
@dgrammatiko this is a completely rebased commit, still DART-SASS issues. As you can see, all changes are all about removing minicolors and replacing it with our color-picker. There is nothing wrong here.

dgrammatiko:
>  still DART-SASS issue

That's weird. The SASS package is the one that is defined in the package.json on the root of the repo and it doesn't throw any errors for the same files. Maybe an app like https://www.sourcetreeapp.com can help you here (figure out the point that your branch is created, normally if this is a new PR that should be the tip of the 4.1-dev.)

thednp:
Probably changing the SCSS sources too much will result in the issue, but I found another problem.

In my local repository, I didn't do the `composer install` action yet and `npm install` will always fail because many scss files from templates import stuff from the **media** folder which doesn't exist, which makes me think the official `4.1-dev` branch is missing something. Something that I didn't have to deal with when I previously commited to `4.0-dev` branch before.

dgrammatiko:
> I didn't do the composer install action yet 

The composer install needs to be done before the npm install. That's expected

thednp:
Something isn't right.

EDIT: I just did a `composer install` then  `npm install` and got the exact same error as the online build testing.
I think something is wrong with the `4.1-dev` build.

EDIT 2: it must be related to what @brianteeman said yesterday, there were some files changed unrelated to my commit.

thednp:
Guys, I can nail down these pesky SCSS issues real quick (`$font-size: $some-var / 2` => `$font-size: $some-var * .5`). I think it's the expected SCSS format in the future and probably SCSS lint will also consider that in the future.

Just let me know.

dgrammatiko:
@thednp please make another PR that will change the division according to SASS docs: https://sass-lang.com/documentation/breaking-changes/slash-div
 
So your example 

```scss
$font-size: $some-var / 2;
```

should be written as:
```scss
$font-size: math.div($some-var, 2);
```

thednp:
That won't work with our `medialize/sass` unfortunately. and Bootstrap devs still count on it, see their [variables.scss](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L376-L383). My solution I posted above is the better one, for now, at least for the front-end side of things, since the back-end SCSS is compiled on build via `node_modules/sass` (dart-sass).

thednp:
@dgrammatiko 5min into the job: `$fa-fw-width:          (20em / 16);` file *joomla-cms\media\vendor\fontawesome-free\scss\_variables.scss* as well as [font-awesome](https://github.com/FortAwesome/Font-Awesome/blob/master/scss/_variables.scss#L11) repository.

Do we fix/resolve 3rd party code quality?

dgrammatiko:
> Do we fix/resolve 3rd party code quality?

Not in this repo. We can either do a PR to their repo or wait till someone else fixes it...

thednp:
We can just override the value with `$fa-fw-width: (20em * .0625);` (width * 1/16).

For instance *joomla-cms\build\media_source\com_media\scss\components\_media-infobar.scss* overrides `$fa-css-prefix:    fa;`

dgrammatiko:
> We can just override the value with $fa-fw-width: (20em * .0625); (width * 1/16)

If you can override it then should be fine, we already overriding the path, etc

thednp:
We may have to override the entire `variables.scss` file because `$fa-fw-width` is not marked with `!default` and the compiler does not like it.

brianteeman:
You can not remove the minicolors css or js as they may be used by extensions and they have not been marked as deprecated

thednp:
So instead of removing minicolors assets and references in the asset json files, we have our color-picker separate?

dgrammatiko:
> So instead of removing minicolors assets and references in the asset json files, we have our color-picker separate?

No that's not what Brian meant. The js+css needs to be in the media/vendor folder but not in the assets.josn. Basically you need to revert some of the changes in the build/settings.json...

package.json
```diff
@@ -31,7 +31,7 @@
     "not ie < 11"
   ],
   "dependencies": {
-    "@claviska/jquery-minicolors": "^2.3.5",
+    "@ctrl/tinycolor": "^3.4.0",
```
thednp, line 35: @dgrammatiko this here is to also include minicolors: yes/no

build/media_source/system/js/fields/color-field-adv-init.es5.js
```diff
@@ -1,38 +0,0 @@
-/**
```
thednp: @dgrammatiko this here is to keep: yes/no

administrator/components/com_admin/script.php
```diff
@@ -6098,6 +6098,9 @@ public function deleteUnexistingFiles($dryRun = false, $suppressOutput = false)
 			'/templates/cassiopeia/css/global/fonts-web_roboto+noto-sans.min.css.gz',
 			'/templates/cassiopeia/scss/global/fonts-web_fira-sans.scss',
 			'/templates/cassiopeia/scss/global/fonts-web_roboto+noto-sans.scss',
+			// From 4.0.x to 4.1.0
```
thednp: @dgrammatiko changes to be reverted: yes/no

build/build-modules-js/init/recreate-media.es6.js
```diff
@@ -19,9 +19,6 @@ module.exports.recreateMediaFolder = async (options) => {
     if (vendor === '@fortawesome/fontawesome-free') {
       return 'vendor/fontawesome-free';
     }
```
thednp: @dgrammatiko to be reverted: yes/no

package.json
dgrammatiko: keep

administrator/components/com_admin/script.php
dgrammatiko: revert

build/build-modules-js/init/recreate-media.es6.js
dgrammatiko: keep

build/build-modules-js/settings.json
```diff
@@ -536,49 +536,6 @@
         "name": "punycode",
         "licenseFilename": "LICENSE-MIT.txt"
       },
```
thednp: @dgrammatiko this here is to be reverted: yes/no

build/media_source/system/joomla.asset.json
```diff
@@ -114,14 +114,26 @@
       "uri": "system/fields/validate.min.js"
     },
     {
-      "name": "field.color-adv",
+      "name": "field.color-picker-es5",
```
thednp, line 117: @dgrammatiko this here is to also include minicolors: yes/no

build/build-modules-js/settings.json
dgrammatiko: should be something like
```json
      "@claviska/jquery-minicolors": {
        "name": "minicolors",
        "js": {
          "jquery.minicolors.js": "js/jquery-minicolors.js",
          "jquery.minicolors.min.js": "js/jquery.minicolors.min.js"
        },
        "css": {
          "jquery.minicolors.css": "css/jquery.minicolors.css"
        },
        "filesExtra": {
          "jquery.minicolors.png": "css/jquery.minicolors.png"
        },
        "licenseFilename": "LICENSE.md"
      },
```

build/media_source/system/joomla.asset.json
dgrammatiko: no this should be your picker now

build/build-modules-js/init/recreate-media.es6.js
dgrammatiko: I mean revert to have the rename thing otherwise it will copy the files in the wrong directory

thednp: you mean revert my change

dgrammatiko: I mean the file should not be changed in this pr.

administrator/components/com_admin/script.php
dgrammatiko: I mean the file should not be changed in this pr.

build/media_source/system/js/fields/color-field-adv-init.es5.js
dgrammatiko: keep this file

package.json
dgrammatiko: deps should have both 
```
    "@claviska/jquery-minicolors": "^2.3.5",
    "@ctrl/tinycolor": "^3.4.0",
```

build/build-modules-js/settings.json
thednp: not the original code? also no jQuery dependency?

thednp:
@dgrammatiko I've changed most things as you advised, please take a minute to check.

dgrammatiko:
@thednp looking good. I will try to do some testing/review this weekend. Thanks

layouts/joomla/form/field/color/advanced.php
```diff
@@ -42,56 +43,87 @@
...
- * @var   array    $control         Is this field checked?
+ * @var   array    $keywords        Option to set color names like "blue" or values like "transparent" or "inherit"
  * @var   string   $dataAttribute   Miscellaneous data attributes preprocessed for HTML output
  * @var   array    $dataAttributes  Miscellaneous data attributes for eg, data-*.
  */
 
-if ($validate !== 'color' && in_array($format, array('rgb', 'rgba'), true))
+// set hsla / rgba to short format
+// both tinycolor and <joomla-field-color-picker> handle alpha automatically
+// an invalid format defaults to 'hex'
```
PhilETaylor: reformat to a multi-line comment with `/** */`

layouts/joomla/form/field/color/advanced.php
```diff
@@ -42,56 +43,87 @@
...
+$format = in_array($format, array('hex', 'rgb', 'hsl', 'rgba', 'hsla'), true) ? str_replace('a', '', $format) : 'hex';
+
+// all formats require color component labels, color names, format and other labels required for accessibility
```
PhilETaylor:
```suggestion
// All formats require color component labels, color names, format and other labels required for accessibility.
```

layouts/joomla/form/field/color/advanced.php
```diff
@@ -42,56 +43,87 @@
...
+$class = !empty($class) ? ' class="' . trim($class) . '"' : '';
+$hint = strlen($hint) ? ' placeholder="' . $this->escape($hint) . '"' : '';
+
+$format = ' format="' . $format . '"';
+$autocomplete = ' autocomplete="' . (!empty($autocomplete) ? $autocomplete : 'off') . '"';
+$spellcheck = ' spellcheck="false"';
+$value = ' value="' . $this->escape($value) . '"';
+
+// The dir="rtl" attribute will set all color picker style
+// color values of any format can never be RTL
```
PhilETaylor: format as multi line comment with `/** */`

thednp:
Thanks @PhilETaylor

dgrammatiko:
@thednp the CI failure will be fixed once https://github.com/joomla/joomla-cms/pull/35518 and https://github.com/joomla/joomla-cms/pull/35666 are merged

thednp:
@dgrammatiko man, I put back minicolors and still no go. I think [the problem](https://ci.joomla.org/joomla/joomla-cms/47668/1/7) is very much related to the [build/media_source/system/scss/fields/joomla-field-color-picker.scss](https://github.com/joomla/joomla-cms/pull/35639/files#diff-9cf887c4de1bdb0cbc9fdd14df0ec0b3d93f0a9824ce3701b5598b48e7bb8996R1-R284).

It may be a problem with the SCSS + JS compile function in the `build` folders.

dgrammatiko:
@thednp has nothing to do with your code, the failure is in bootstrap.scss
Screenshot 2021-09-25 at 12 33 09

FWIW if you apply both the PRs from https://github.com/joomla/joomla-cms/pull/35639#issuecomment-927101230 CI will be fine

thednp:
Alright buddy, what about my fixes in #35648?

Also, any news on the TinyMCE improvements? Anything I can give it a try or input?

dgrammatiko:
> Alright buddy, what about my fixes in #35648?

That is valid (I think I'm not a SCSS person)

> Also, any news on the TinyMCE improvements?

There will be a POC PR today, I'm just trying to introduce all the changes without an B/C breaks or a bazzilion of cryptic options

thednp:
> > Alright buddy, what about my fixes in #35648?
> 
> That is valid (I think I'm not a SCSS person)

To my research, those fixes are absolutely needed in dart-sass 2.0, it would be a waste of somebody's time someday to find and nail them down like I did.
 
> > Also, any news on the TinyMCE improvements?
> 
> There will be a POC PR today, I'm just trying to introduce all the changes without an B/C breaks or a bazzilion of cryptic options

Sweet, please tag me in.

HLeithner:

HLeithner:

HLeithner:

HLeithner:

PR #35644, opened 2021-09-22
[4.0] Centered text on emptystater screen is hard to read (a11y).

chmst:
Pull Request for Issue # .

### Summary of Changes
Text in paragraphs is hard to read if it is centered. This PR aligns textin emptystate layout for better reading. 

Compare empty page screens screens before and after applying the patch. 

Example here: administrator/index.php?option=com_installer&view=discover

text-before-after

### DIFF
--- a/layouts/joomla/content/emptystate.php
+++ b/layouts/joomla/content/emptystate.php
@@ -39,7 +39,7 @@ $btnadd     = $displayData['btnadd'] ?? Text::_($textPrefix . '_EMPTYSTATE_BUTTO
 		<span class="fa-8x mb-4 <?php echo $icon; ?>" aria-hidden="true"></span>
 		<h1 class="display-5 fw-bold"><?php echo $title; ?></h1>
 		<div class="col-lg-6 mx-auto">
-			<p class="lead mb-4">
+			<p class="lead mb-4 text-start">
 				<?php echo $content; ?>
 			</p>
 			<div class="d-grid gap-2 d-sm-flex justify-content-sm-center">


PhilETaylor:
This is NOT the place to be putting paragraphs of text!!! 

This is meant to be a sentence maximum in this place.

PhilETaylor:
Some examples of how other people add additional help... 
94971685-7c173780-04ff-11eb-9587-4ca882d303ff
94969660-0b6e1c00-04fb-11eb-98f5-a4cf69936877
94971318-ae746500-04fe-11eb-9823-ad9e18890943

brianteeman:
i agree that center alignment should not be used for paragraphs.

phil is also correct in saying that this place is only intended to be used for a sentence

i wonder why in german there is such a long block of text and in english just a short sentence

image

PhilETaylor:
Every single google search I have done for "Empty State Design" shows entered text following the title. I failed to find any examples of Empty State Design with paragraphs.

chmst:
Agree that is should be very short text. So will inform german language team, thanks.

PhilETaylor:
I am not saying that empty state cannot have lots of text, but that needs to be designed into the state, and not just placed in the short description under the title which should just be one sentence long ideally.

There are lots of examples of empty state design which are a lot better than ours, and I fully expect iterative improvements to this kind of view over time￼￼ to provide more help and guides for new users

chmst:
The empty text was a great improvment as is. At the moment  there is not much free capacity for beautifer actions.

PR #17482, opened 2017-08-11
Remove testing sample data plugin

Bakual:
As mentioned in Issue #17456 we don't have the testing sample data available in stable releases, so it doesn't make much sense to ship this sample data plugin in a stable release.
We can make it available separately for testers (needs an own repo and some doc I guess).

Maybe we can add another sample data plugin, because the list looks a bit empty with only one sample set :smile: 

@mbabker I just removed the plugin files and the statetement from the initial SQL. If we only support updates from the latest stable, this will work. If you want to support updates from 3.8.0 beta we would have to add a new SQL where the entry is removed and also add the deleted files to the script. Your choice to make :smile: 

### Summary of Changes
Removes the testing sample data plugin

### Testing Instructions
Test that only the blog sample data plugin is left in the module.

### Documentation Changes Required
None

PR #35645, opened 2021-09-22
[4.1-dev] Fixing SCSS issues with division deprecation and math.div()

thednp:
Pull Request for PR #35639 .

# Important
This PR should be merged into [4.0-dev] as well asap.

### Summary of Changes

* The fontawesome icon library haven't marked their `$fa-fw-width` variable with `!default`, see [#17482](https://github.com/FortAwesome/Font-Awesome/issues/17482) which uses `(20em / 16)` expression which is deprecated in dart-sass 2.0 and also [largely reported](https://github.com/FortAwesome/Font-Awesome/issues/17946) in the fontawesome issues.

* Other changes in this PR also address other dart-sass issues, but we don't replace for instance `24 / 16` with `math.div(24, 16)` as recommended by the compiler, but with `24 * 0.0625` (which is the result of 1/16) and the reason for that is the libsass compiler still largely used and this approach is also widely adopted. I suspect scssPhp also might not have `math` implemented.

* See [this PR in fontawesome](https://github.com/FortAwesome/Font-Awesome/pull/17940)
for more info.

### Testing Instructions
Just build and check SASS compiler output.

The compiler just won't do division via the `/` (forward slash) operator
https://ci.joomla.org/joomla/joomla-cms/47599

All scss files compiled successful.

No.


thednp:
@dgrammatiko and any other please test and push this.

brianteeman:
If you want it to be merged into 4.0 then you should do the pr on that branch and not 4.1

From your description it implies that its just for fontawesome but I see in the files that it applies elsewhere. Please confirm and update your initial post and test instructions

thednp:
I need this in 4.1 for #35639.

It is for fontawesome mainly, but also some other minor changes to pass the compiler. As you can see for yourself. So close commit and re-commit to 4.0? and then someone else will merge the change to 4.1?

4.1 branch is important to get that PR solved.

dgrammatiko:
> So close commit and re-commit to 4.0?

You don't have to close and reopen you can rebase it to 4.0-dev, eg click on the edit next to the PR's title and then select the branch
Screenshot 2021-09-22 at 19 38 01

thednp:
I can't do that right now, I'm still working with GH app on the color-picker branch.

PR #35647, opened 2021-09-22
[4.1] Tiny source code syntax highlighting

brianteeman:
This PR 
changes the source code view of tinymce from raw text to a rich syntax highlighted area using codemirror.

It is enabled by default but there is an option in the plugin to disable it - can't imagine why someone would want though.

There are no additional tinymce strings to translate as I have reused existing strings.

As the code plugin is enabled by default in set 0 for super users you can activate the view either from the tools menu or the `<>` icon. If it is not enabled in your configuration then you will need to drag and drop the icon to your toolbar in the tinymce plugin

### Before 
image

### After
image

As you can see you now also have search and replace etc.

### Testing
Apply the pr and run npm ci
AND copy the file `build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.html` to `media/plg_editors_tinymce/js/plugins/highlighter/source.html

### Why this is a draft
I need help building the source.html file with our build scripts. I am guessing it will be built similar to the error pages

Thanks to @woluweb and @dgrammatiko


build/media_source/plg_editors_tinymce/js/plugins/highlighter/plugin.es5.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * plugin.js
+ *
+ * Original code by Arjan Haverkamp
+ * Copyright 2013-2015 Arjan Haverkamp (arjan@webgear.nl)
+ *
+ * Adapted for use in Joomla by Dimitrios Grammatikogiannis
```
richard67, line 7: Dimitrios? I thought it's Dimitris, without the "o".

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
```diff
@@ -0,0 +1,279 @@
+/**
+ * source.js
+ *
+ * Original code by Arjan Haverkamp
+ * Copyright 2013-2015 Arjan Haverkamp (arjan@webgear.nl)
+ *
+ * Adapted for use in Joomla by Dimitrios Grammatikogiannis
```
richard67, line 7: Dimitrios? Not Dimitris?

build/media_source/plg_editors_tinymce/js/plugins/highlighter/plugin.es5.js
brianteeman: https://github.com/dgrammatiko/tinyMCE-with-CodeMirror/blob/d252cf1653f9dc57c149b5e623a3866a66db8ad7/src/plugin.js#L11

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
brianteeman: https://github.com/dgrammatiko/tinyMCE-with-CodeMirror/blob/d252cf1653f9dc57c149b5e623a3866a66db8ad7/src/plugin.js#L11

build/media_source/plg_editors_tinymce/js/plugins/highlighter/plugin.es5.js
richard67: Strange because in GitHub profile it‘s Dimitris, and in Glip, too.

richard67: @dgrammatiko Could you shed a light on this? Maybe both is right depending on transcription from Greece letters?

dgrammatiko: So, turns out the name printed in my id has an `o` but I hardly ever use that version, and honestly, I don't know why I even wrote that comment in the first place. Seems like a failed multitasking...

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
```diff
@@ -0,0 +1,279 @@
...
+
+  // Create legend for keyboard shortcuts for find & replace:
+  const head = parent.document.querySelectorAll('.tox-dialog__footer')[0];
+  const div = parent.document.createElement('div');
+  const td1 = '<td style="font-size:11px;background:#777;color:#fff;padding:0 4px">';
+  const td2 = '<td style="font-size:11px;padding-right:5px">';
+  div.innerHTML = `
+<table cellspacing="0" cellpadding="0" style="border-spacing:4px">
+  <tr>
+    ${td1}${(isMac ? '&#8984;-F' : 'Ctrl-F</td>')}${td2}${tinymce.translate('Start search')}</td>
```
dgrammatiko, line 145: @brianteeman replace `tinymce.translate('Start search')` with a `Joomla.Text._('SOME_STRING')`

This will also need a conditional in the tinymce.php to pass the string in the browser: `Text::script('SOME_STRING')`

Same goes for `find next`, `Find Previous`, `Replace`, `Replace All`

dgrammatiko: > There are no additional tinymce strings to translate as I have reused existing strings.

Ok maybe I'm wrong here

brianteeman: yes you are wrong. it uses tinymce strings not joomla strings. it works fine

dgrammatiko:
> I need help building the source.html file with our build scripts. I am guessing it will be built similar to the error pages

This is getting complicated due to the fact that the output is an HTML page and we don't want the sources in the media folder. I would just concatenate the source script and CSS in the HTML directly (non minified, it's not that much code anyway) and delete the js and css sources.

brianteeman:
hint?

dgrammatiko:
`source.html`
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <script>
// copy the contents of source.es5.js here
    </script>
    <style type="text/css">
// copy the contents of source.css here
    </style>
  </head>
  <body></body>
</html>
```

Delete the extra files

brianteeman:
oh - didnt realise thats what you meant

woluweb:
Txs @dgrammatiko for your plugin and txs @brianteeman for implementing this PR!
This is very exciting.
It will be so handy on a day-to-day basis... and also it will give an extra incentive to use TinyMCE and stick to the core :)

If I may, I have two suggestions:

### 1. making the popup (almost) fullscreen?

At present the size of the popup is (only) 800x550, which is quite small especially to work comfortably with html code.
When testing Dimitris' plugin, I "cheated" by adding the following in Atum's user.css to go (almost) full-screen, since it is a popup anyway:

```
.tox-dialog {
    width: calc(100vw - 100px) !important;
    max-width: unset !important;
    height: calc(100vh - 100px) !important;
    max-height: unset !important;
}
```

But it would probably be nicer to have it natively.

Alternatively, if it is possible to implement the "fullscreen shortcut" (F10) (which works when selecting CodeMirror editor itself, even if there is a positioning issue bc the first lines are hidden behind the Save buttons bar), that could also solve the issue of the popup being "too small".

### 2. using the options of CodeMirror Plugin?

The popup with Syntax Highliting uses CodeMirror, but does not seem to use the Options set in the CodeMirror Plugin (it uses the default I guess).
I am already very happy as it is now, obviously.
But I wanted to mention it just in case it would be easy to apply CodeMirror Plugin Options.

brianteeman:
I am looking at both those points

dgrammatiko:
> I "cheated" by adding the following in Atum's user.css

This has weird implications as it will make ALL the tinyMCE dialogs almost fullscreen. There's another way for this

> but does not seem to use the Options set in the CodeMirror Plugin

Since I didn't wanted to create a Joomla plugin I left this with some defaults but it is possible to have the options exposed in the tinyMCE's configuration. Basically all it's needed is:
- remove this part from the source.js
```js
// CodeMirror settings
 const CMsettings = {
   path: '../../../../vendor/codemirror',
   indentOnInit: true,
// start remove
   config: {
     mode: 'htmlmixed',
     theme: 'default',
     lineNumbers: true,
     lineWrapping: true,
     indentUnit: 2,
     tabSize: 2,
     indentWithTabs: true,
     matchBrackets: true,
     saveCursorPosition: true,
     styleActiveLine: true,
   },
// end remove
   jsFiles: [// Default JS files
     'lib/codemirror.min.js',
     'addon/edit/matchbrackets.min.js',
     'mode/xml/xml.min.js',
     'mode/javascript/javascript.min.js',
     'mode/css/css.min.js',
     'mode/htmlmixed/htmlmixed.min.js',
     'addon/dialog/dialog.min.js',
     'addon/search/searchcursor.min.js',
     'addon/search/search.min.js',
     'addon/selection/active-line.min.js',
   ],
   cssFiles: [// Default CSS files
     'lib/codemirror.css',
     'addon/dialog/dialog.css',
   ],
 };
```
- in the tinyMCE.php pass the options like:
```php
$scriptOptions['codemirror'] = [
  'settings' => [
    'indentOnInit' =>  true, // the true should be $levelParams->get('codemirroe_settings_indentoninit'); which is a toogle in the settings form
  ]
] 
```

brianteeman:
yeah I know ;)
I just wanted to put up a quick draft so I could ask about the html file

dgrammatiko:
About the maximised modal, this should do it:

```js

    const rects = document.body.getBoundingClientRect();

    const config = {
      title: 'HTML source code',
      url: url + '/source.html',
      width: rects.width - 5,
      height: rects.height - 150,
      resizable: true,
      maximizable: true,
      fullScreen: true,
      saveCursorPosition: false,
      buttons: buttonsConfig
    }

```

Result: 
Screenshot 2021-09-23 at 15 27 31

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
```diff
@@ -0,0 +1,279 @@
...
+      case '&': depth++; break;
+    }
+  }
+  return depth;
+}
+
+/**
+ * This function is called by plugin.js, when user clicks 'Ok' button
+ */
+window.submit = () => {
```
Fedik: This is forbidden. 
I know it not your but if doing a global method then be more verbose, eg 
`window.tinymceHighlighterSubmit ...`

dgrammatiko: You're right, this needs to be renamed both here and in the plugin.js

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
```diff
@@ -0,0 +1,279 @@
...
+    inst.isDirty = true;
+  });
+}
+
+/**
+ * Listen for the escape key and close the modal
+ *
+ * @param {Event} evt
+ */
+document.onkeydown = (evt) => {
```
Fedik: and here `document.addEventListener('keydown', (evt) => { .... })` please

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  // Add trailing slash to path
+  if (!/\/$/.test(CMsettings.path)) {
+    CMsettings.path += '/';
+  }
+
+  // Write stylesheets
+  for (let i = 0; i < CMsettings.cssFiles.length; i++) {
+    document.write('<li'+'nk rel="stylesheet" type="text/css" href="' + CMsettings.path + CMsettings.cssFiles[i] + '" />');
```
Fedik: This is also a blast from past :wink: 
All call of `document.write` need to be replaced to use `document.createElement`
Eg:
```javascript
const $link = document.createElement('link');
$link.type = 'text/css';
$link.href = CMsettings.path + CMsettings.cssFiles[i];
document.head.append($link);
```

build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.es5.js
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  // Write JS source files
+  for (let i = 0; i < CMsettings.jsFiles.length; i++) {
+    document.write('<scr'+'ipt type="text/javascript" src="' + CMsettings.path + CMsettings.jsFiles[i] + '"></scr'+'ipt>');
```
Fedik: Example for script:
```javascript
const $script = document.createElement('script');
$script.defer = true;
$script.src = CMsettings.path + CMsettings.jsFiles[i];
document.head.append($script);
```

bembelimen:
Should this really be 4.0 or is 4.1 the target? If later, would be happy to include it.

woluweb:
> Should this really be 4.0 or is 4.1 the target? If later, would be happy to include it.

Well, it is not up to me to say but I hope we won't have to wait until 4.1 for Syntax Highlighting :)

brianteeman:
@fedik please could you review and check the changes you requested

Fedik:
@brianteeman they looks good now :)

plugins/editors/tinymce/tinymce.php
```diff
@@ -519,6 +519,12 @@ public function onDisplay(
 				$plugins[] = $pName;
 			}
 		}
+		// Use CodeMirror in the code view instead of plain text to provide syntax highlighting
+		$sourcecode     = $levelParams->get('highlightPlus', 1);
+		if ($sourcecode)
```
```suggestion
		}

		// Use CodeMirror in the code view instead of plain text to provide syntax highlighting
		$sourcecode = $levelParams->get('highlightPlus', 1);

		if ($sourcecode)
```

dgrammatiko:
@brianteeman check https://github.com/brianteeman/joomla-cms/pull/130

brianteeman:
hound says no ??

dgrammatiko:
Hound was removed many months ago, I have no clue why you still have that around


@brianteeman it should be fine now

brianteeman:
conflicts resolved

chmst:
Tested on a  fresh installation of 4.1. dev on  a new database. Applied Patch, run npm ci.. 

@brianteeman  could you please remove the testing instruction here:
_AND copy the file build/media_source/plg_editors_tinymce/js/plugins/highlighter/source.html to `media/plg_editors_tinymce/js/plugins/highlighter/source.html_

The result:

image


Did I miss something?

dgrammatiko:
@brianteeman check https://github.com/brianteeman/joomla-cms/pull/131

chmst:
Works :) 

image

woluweb:
Txs to all for this PR 
🧡💙❤️💚
@dgrammatiko had mentioned a clean solution above _to have the popup (almost) full-screen (and not like 600x400 like it is by default now)_.
I really think we should go for that bc in practice it won't be very handy if the popup only covers like 1/4 of the screen, especially for Code editing.
This PR gives an extra incentive to have people using TinyMCE -which is excellent- so let's have this popup (almost) full-screen to get the full benefit of it :)

Krshivam25:
I have tested this item :white_check_mark: successfully on e28330107eecf853ef4c874a37ad63215c63b961<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35647">issues.joomla.org/tracker/joomla-cms/35647</a>.</sub>

chmst:
Can we test now or is it better waiting changes in the window?

chmst:
I have tested this item :white_check_mark: successfully on e28330107eecf853ef4c874a37ad63215c63b961<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35647">issues.joomla.org/tracker/joomla-cms/35647</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35647">issues.joomla.org/tracker/joomla-cms/35647</a>.</sub>

chmst:
THX
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35647">issues.joomla.org/tracker/joomla-cms/35647</a>.</sub>

bembelimen:
Very cool feature, thank you!

brianteeman:
thanks

bsakhanov:
For TinyMCE, such features would also be relevant (in the form of plugins, maybe)... For example, along with Quantum (as a media manager), I have to use JCE for projects (a shell for TinyMCE as a component of a visual editor for jumla) only because of a few things: 1) working with the source code, i.e. with the code is convenient — there is a cool search with autocorrect; 2) auto-insertion is clearly worked out directly into the text of the video article, when just a short YouTube url is immediately transformed into a video insertion code with a frame, plus inserts from social networks (Twitter, Facebook, etc.) work similarly.; 3) in the visual editor mode, the text format is displayed there — a paragraph or divas, for example, which is very convenient.

bsakhanov:
For TinyMCE, such features would also be relevant (in the form of plugins, maybe)... For example, along with Quantum (as a media manager), I have to use JCE for projects (a shell for TinyMCE as a component of a visual editor for jumla) only because of a few things: 1) working with the source code, i.e. with the code is convenient — there is a cool search with autocorrect; 2) auto-insertion is clearly worked out directly into the text of the video article, when just a short YouTube url is immediately transformed into a video insertion code with a frame, plus inserts from social networks (Twitter, Facebook, etc.) work similarly.; 3) in the visual editor mode, the text format is displayed there — a paragraph or divas, for example, which is very convenient.

brianteeman:
Not really sure I understand what you are asking for here but afaict everything is already present in joomla 4 and tinymce

PR #35655, opened 2021-09-24
[4.4] Invalid timings in debug plugin

Denitz:
### Summary of Changes

The time and memory displayed in debug bar (right position) incorrectly includes the usage of debug plugin itself: it's always much bigger than the sum of all timings in Profile tab.

We can't measure 100% correctly but at least should get closer: the most recent profile mark should have right border as closer to right as possible.

See that `afterRender` profile mark is not close the the right, hence we have incorrectly logged total time: it includes the time spent in debug plugin itself and we don't see the real state of page generation.

1. Debug plugin refactored to `Joomla\Event\SubscriberInterface`
2. Fixed problem with the time spent for `onAfterRespond` event is not counted.
3. Added `afterRespond` profiler mark.
4. Fixed problem with the time spent in `onAfterDisconnect` event by debug plugin is included in request time.
5. Remove unused `PlgSystemDebug::$totalQueries` variable
6. Change `PlgSystemDebug::sqlShowProfiles` to local variable because it's the only usage.
7. Custom MemoryCollector to display the peak memory usage before debug data collecting routines.
8. Display main memory usage with 3 digits.

### Testing Instructions

Apply patch, see more accurate timings in debug. Profile tab marks are now more proportionally to the 100% width.

See issues in Changes above.

See issues in Changes above.

No.


plugins/system/debug/debug.php
```diff
@@ -1,10 +1,10 @@
 <?php
 /**
- * @package     Joomla.Plugin
- * @subpackage  System.Debug
+ * @package         Joomla.Plugin
+ * @subpackage      System.Debug
  *
  * @copyright   (C) 2006 Open Source Matters, Inc. <https://www.joomla.org>
- * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ * @license         GNU General Public License version 2 or later; see LICENSE.txt
```
brianteeman: please revert these changes

plugins/system/debug/debug.php
```diff
@@ -520,7 +537,7 @@ public function onAfterDisconnect(ConnectionEvent $event)
 	 *
 	 * @return  void
 	 *
-	 * @since   3.1
+	 * @since       3.1
```
brianteeman: please revert this change

plugins/system/debug/src/DataCollector/ProfileCollector.php
```diff
@@ -3,7 +3,7 @@
  * This file is part of the DebugBar package.
  *
  * @copyright  (c) 2013 Maxime Bouroumeau-Fuseau
- * @license    For the full copyright and license information, please view the LICENSE
+ * @license        For the full copyright and license information, please view the LICENSE
```
brianteeman: please revert this change

plugins/system/cache/cache.php
```diff
@@ -140,6 +140,8 @@ public function onAfterRoute()
 
 				Profiler::getInstance('Application')->mark('afterCache');
 				$this->app->triggerEvent('onAfterRespond');
+
+				Profiler::getInstance('Application')->mark('afterRespond');
```
PhilETaylor:
```suggestion
				$profilerApp = Profiler::getInstance('Application');
				$profilerApp->mark('afterCache');
				$this->app->triggerEvent('onAfterRespond');

				$profilerApp->mark('afterRespond');
```

plugins/system/debug/debug.php
```diff
@@ -510,6 +490,44 @@ public function onAfterDisconnect(ConnectionEvent $event)
...
+
+		$this->timeInOnAfterDisconnect = microtime(true) - $startTime;
+	}
+
+	/**
+	 * Method to check if the current user is allowed to see the debug information or not.
+	 *
+	 * @return  boolean  True if access is allowed.
+	 *
+	 * @since   3.0
```
PhilETaylor:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

plugins/system/debug/src/DataCollector/MemoryCollector.php
```diff
@@ -0,0 +1,153 @@
+<?php
+/**
+ * This file is part of the DebugBar package.
+ *
+ * @copyright  (c) 2013 Maxime Bouroumeau-Fuseau
+ * @license    For the full copyright and license information, please view the LICENSE
+ *             file that was distributed with this source code.
```
PhilETaylor: A note should be inserted to say that this file has been forked and modified since released by the upstream project.

plugins/system/debug/src/DataCollector/MemoryCollector.php
```diff
@@ -0,0 +1,153 @@
...
+	/**
+	 * @var float
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	protected $peakUsage = 0;
+
+	/**
+	 * Constructor.
+	 *
```
PhilETaylor: Useless comment 

```suggestion
```

plugins/system/debug/src/DataCollector/MemoryCollector.php
```diff
@@ -0,0 +1,153 @@
...
+	 * @param Registry    $params Parameters.
+	 * @param float       $peakUsage
+	 * @param boolean     $realUsage
```
PhilETaylor: Spaces as per https://developer.joomla.org/coding-standards/docblocks.html

```
variable type (There must be 3 spaces before variable type.)
variable name (There must be 2 spaces after the longest type.)
variable description (There must be 2 spaces after the longest variable name.)
```


```suggestion
	 * @param   Registry  $params Parameters.
	 * @param   float     $peakUsage
	 * @param   boolean   $realUsage
```

plugins/system/debug/debug.php
Denitz: Hm... but `PlgSystemDebug::isAuthorisedDisplayDebug()` really exists since 3.0

plugins/system/debug/src/DataCollector/ProfileCollector.php
```diff
@@ -58,12 +58,18 @@ class ProfileCollector extends AbstractDataCollector
 	 * Constructor.
 	 *
 	 * @param   Registry  $params  Parameters.
+	 * @param   float     $requestStartTime
+	 * @param   float     $requestEndTime
 	 *
 	 * @since 4.0.0
 	 */
-	public function __construct(Registry $params)
+	public function __construct(Registry $params, $requestStartTime = null, $requestEndTime = null)
```
PhilETaylor: Are we allowed to change class signatures in a minor version?

plugins/system/debug/src/DataCollector/ProfileCollector.php
```diff
@@ -82,8 +93,8 @@ public function __construct(Registry $params)
 	 * @param   string|null  $label      Public name
 	 * @param   string|null  $collector  The source of the collector
 	 *
-	 * @since  4.0.0
 	 * @return void
+	 * @since  4.0.0
```
PhilETaylor: https://developer.joomla.org/coding-standards/docblocks.html

```
@return (required, followed by a blank line)
```

```suggestion
	 * @return void
	 *
	 * @since  4.0.0
```

plugins/system/debug/src/DataCollector/MemoryCollector.php
Denitz: @PhilETaylor I copied the desc from ProfileCollector, actually it's the only our collector which has this.
All other collectors have 
```
/**
 * @package     Joomla.Plugin
 * @subpackage  System.Debug
 *
 * @copyright   (C) 2018 Open Source Matters, Inc. <https://www.joomla.org>
 */
```
So, what should we use for all of our collectors?

plugins/system/debug/src/DataCollector/ProfileCollector.php
```diff
@@ -115,9 +126,9 @@ public function hasStartedMeasure($name): bool
 	 * @param   string  $name    Measurement name.
 	 * @param   array   $params  Parameters
 	 *
-	 * @since  4.0.0
-	 * @throws DebugBarException
 	 * @return void
+	 * @throws DebugBarException
+	 * @since  4.0.0
```
PhilETaylor: https://developer.joomla.org/coding-standards/docblocks.html
correct order and blank lines needed.
```
@return (required, followed by a blank line)
@SInCE (required, followed by a blank line if there are additional tags)
@throws (required if method or function arguments throws a specific type of exception)
```
```suggestion
	 * @return void
	 *
	 * @since  4.0.0
	 *
	 * @throws DebugBarException
```

plugins/system/debug/src/DataCollector/ProfileCollector.php
```diff
@@ -148,8 +153,8 @@ public function stopMeasure($name, array $params = [])
 	 * @param   array        $params     Parameters.
 	 * @param   string|null  $collector  A collector.
 	 *
-	 * @since  4.0.0
 	 * @return void
+	 * @since  4.0.0
```
PhilETaylor: apply the ordering and blank lines referenced here https://github.com/joomla/joomla-cms/pull/35655/files#r715486122 throughout

plugins/system/debug/src/DataCollector/MemoryCollector.php
Denitz: @PhilETaylor Thanks, it's always a pain, I set the PHPStorm code style "From Joomla" and it looks like either outdated or incorrect.

plugins/system/debug/debug.php
PhilETaylor: ok cool.

plugins/system/debug/src/DataCollector/ProfileCollector.php
Denitz: @PhilETaylor This is non-breaking change, and it's just a debug plugin :)

PhilETaylor: does that even matter? its a change to the constructor of a class, I might be wrong, but I dont think you are allowed to do that in a minor version in SemVer... Ive not had a coffee yet, I might be wrong on that.

bembelimen: See: https://github.com/joomla/joomla-cms/pull/35476#issuecomment-914301986 so can only happen in 5.0

Denitz: @PhilETaylor @bembelimen I've reverted the constructor of ProfileCollector and just added `setRequestEndTime()` method.

Denitz:
@brianteeman @PhilETaylor Sorry, any news?

libraries/src/Application/CMSApplication.php
```diff
@@ -319,6 +319,9 @@ public function execute()
 
 		// Trigger the onAfterRespond event.
 		$this->getDispatcher()->dispatch('onAfterRespond');
+
+		// Mark afterRespond in the profiler.
+		JDEBUG ? $this->profiler->mark('afterRespond') : null;
```
Fedik: Does it really work?
From what I see, it will not, because Debug plugin doing "data collection" `onAfterRespond`, so this Profiler point will be lost.

plugins/system/debug/debug.php
```diff
@@ -177,7 +171,7 @@ public function __construct(&$subject, $config)
 		$this->isAjax = $this->app->input->get('option') === 'com_ajax'
 			&& $this->app->input->get('plugin') === 'debug' && $this->app->input->get('group') === 'system';
 
-		$this->showLogs = (bool) $this->params->get('logs', true);
+		$this->showLogs = (bool) $this->params->get('logs', false);
```
Fedik: Please revert it, in the form default is `1`, so should be `true` here.

Fedik (changes requested):
Please restore original positions of onAjaxDebug, isAuthorisedDisplayDebug methods in the file.
So it no looks like "new code", and will be more easy to review and see what was changed.

plugins/system/debug/debug.php
```diff
@@ -198,6 +192,26 @@ public function __construct(&$subject, $config)
...
+	{
+		return [
+			'onBeforeCompileHead' => 'onBeforeCompileHead',
+			'onAjaxDebug'         => 'onAjaxDebug',
+			'onAfterDisconnect'   => 'onAfterDisconnect',
+			'onAfterRespond'      => [
+				'onAfterRespond',
+				\Joomla\Event\Priority::MIN,
+			],
+		];
```
Fedik: ~Please also add~
```
ApplicationEvents::AFTER_RESPOND =>[ 'onAfterRespond', Priority::MIN] 
```
~We will be able to collect data after redirects (for POST request ), trigerred by `AbstractWebApplication::redirect`.
I wanted to make a PR, but you already started with `SubscriberInterface`~ :)

Fedik: Well, no, ignore this comment, I will make it separately, later. 
There need to test some more things for it

libraries/src/Application/CMSApplication.php
Denitz: Good catch, I moved `'afterRespond'` mark into the plugin's `onAfterRespond` event which is now executed as last via `PHP_INT_MIN` priority.

plugins/system/debug/debug.php
Denitz: Done.

Denitz:
@Fedik I restored original positions of onAjaxDebug, isAuthorisedDisplayDebug methods in the file.

plugins/system/debug/debug.php
```diff
@@ -244,8 +258,14 @@ public function onBeforeCompileHead()
 	 */
 	public function onAfterRespond()
 	{
+		// Mark afterRespond in the profiler.
+		JDEBUG ? Profiler::getInstance('Application')->mark('afterRespond') : null;
```
Fedik: Better keep it in Application, as you did before.  But in correct order:

```php
// Mark afterRespond in the profiler.
JDEBUG ? $this->profiler->mark('afterRespond') : null;

// Trigger the onAfterRespond event.
$this->getDispatcher()->dispatch('onAfterRespond');
```

So other custom plugins can access to this data.

Denitz: I disagree, `afterRespond` profiling mark should be set after `onAfterRespond` event is executed, otherwise we don't catch the time of plugins subscribed to `onAfterRespond` event.

Debug plugin listens to `onAfterRespond` event last in the queue, hence it's quite correct to count `afterRespond` mark in its handler.

Fedik: We will not get 100% accurate timing for afterRespond anyway.
There may be more onAfterRespond listeners, that will run after debug plugin.

But by placing mark('afterRespond') in our debug plugin we limit other custom debug plugins to do the same, that not right. It should be generic.

Denitz: OK, what is the purpose of `afterRespond` mark in this case? It won't be visible in debug profile timings. I guess it's better to remove it completely in this case.
But, I can apply register_shutdown_function and process the debug code info after the `afterRespond` mark was added in app, do you agree with this solution?

Fedik: > It won't be visible in debug profile timings.

Why? if you do in this order, it wil be in debug plugin:
```php
// Mark afterRespond in the profiler.
JDEBUG ? $this->profiler->mark('afterRespond') : null;

// Trigger the onAfterRespond event.
$this->getDispatcher()->dispatch('onAfterRespond');
```

> I can apply register_shutdown_function 

I do not think that it need

Denitz: A mark before executing `onAfterRespond` event should be `beforeRespond` in this case, but not `afterRespond`.
Imho, as a developer, I am interested in `afterRespond` mark which provide the timing AFTER executing `onAfterRespond` event and delivers the info about possible issues with plugins subscribed on `onAfterRespond` event.

Fedik: I understood what you wanted.
Here we talking about App respond, not really about whole respond, it not possible to track 100% anyway.
Moving  `mark('afterRespond')`  to plugin does not help much, so better keep it consistent on App level.

Denitz: Let's get rid of this `afterRespond` mark in this case because it's useless. Please confirm.

Fedik: yeah, can remove it :)

Denitz: @Fedik Removed.

Fedik (approved):

HLeithner:

plugins/system/debug/debug.php
```diff
@@ -341,7 +343,7 @@ public function onAfterRespond()
         // No debug for Safari and Chrome redirection.
         if (
             strpos($contents, '<html><head><meta http-equiv="refresh" content="0;') === 0
-            && strpos(strtolower($_SERVER['HTTP_USER_AGENT'] ?? ''), 'webkit') !== false
+            && stripos($_SERVER['HTTP_USER_AGENT'] ?? '', 'webkit') !== false
```
laoneo: Guess this change should be reverted to stay on the safe side.

plugins/system/debug/src/DataCollector/MemoryCollector.php
```diff
@@ -0,0 +1,147 @@
...
+use Joomla\Plugin\System\Debug\AbstractDataCollector;
+use Joomla\Registry\Registry;
+
+/**
+ * Collects info about the request duration as well as providing
+ * a way to log duration of any operations
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class MemoryCollector extends AbstractDataCollector
```
laoneo, line 26: Can you not extend from DebugBar\DataCollector\MemoryCollector instead of copying the whole file?

Denitz: A separate collector, what's wrong here?

laoneo:
Can you fix the conflicts here?

Denitz:
@laoneo Please restart drone, smth is wrong with the tests and I can't find any relationship with the code itself.

Denitz:
Sorry, still weird issue with databases in drone

Issue #33529, opened 2021-05-03
[4] PHP 8 named parameters issue - Cannot edit content with 3rd party editor


PhilETaylor:
### Steps to reproduce the issue

Install JCE
Enable JCE in Global Config
Edit an article

### Expected result

Ability to edit an article

### Actual result

PHP 8 has named parameters!!! This I believe is a Joomla PHP 8 issue and not one to blame JCE for! 

Note that the error and stack trace point to a Joomla set of files and not JCE - **so please DONT close this or write it off as a JCE issue without debugging!!!** If it effects JCE, its probably going to effect other editors ... 

<img width="1596" alt="Screenshot 2021-05-04 at 00 18 21" src="https://user-images.githubusercontent.com/400092/116944887-417ca580-ac6e-11eb-9d74-0a47e9929646.png">

### System information (as much as possible)

PHP 8+ its broken

With PHP 7.4 it works perfectly.

<img width="1240" alt="Screenshot 2021-05-04 at 00 30 24" src="https://user-images.githubusercontent.com/400092/116945601-fcf20980-ac6f-11eb-8f82-aef06f231f4d.png">

### Additional comments

@ryandemmer @widgetfactory

ryandemmer:
Thanks @PhilETaylor 

I've just run a little test where I clear the contents of all methods in the plugins/editors/jce.php file, and disabled all other JCE plugins, and the above error still occurs.

This is obviously not caused by any process within JCE.

PhilETaylor:
It kinda is and kinda is not a JCE issue. But its one that Joomla should fix too.

It will only happen on PHP 8 due to the name arguments feature and the way Joomla calls the JCE method with `call_user_func`

its because JCE is missing the `$params` from its method - but equally Joomla should be able to cope with that.

joomdonation:
If JCE has $params parameter in `onDisplay` method, the error would not happen. The code in Joomla has been there for sometime, so we could also that it is something JCE should fix to be compatible with PHP 8 :).

ryandemmer:
Oh I see! An easy fix for me then.

I blame it on the fact that yesterday was a bank holiday here, so today is basically a Monday... ;)

joomdonation:
@ryandemmer Yes, an easy fix for you and I'm sure you won't complain about it :).

ryandemmer:
[...]

PR #35675, opened 2021-09-26
Causing issue in PHP8.0.11 due to BC break in PHP8.0 regarding call_user_func_array and named parameters

alexandreelise:
Pull Request for Issue #35674 .

### Summary of Changes
Fix issue in Captcha causing a Fatal error in PHP8.0 due to B/C break in PHP8.0 regarding call_user_func_array and named properties
https://github.com/joomla/joomla-cms/issues/35674#issue-1007224748
Added array_values around $params to prevent breaking change behaviour when using assoc arrays and named properties in call_user_func_array in PHP 8.0.11

### Testing Instructions
Use PHP8.0.11
Enable captcha for Joomla 4.0.3 core contact form
Go to the contact form

Fatal error: Unknown named property $id

The captcha shows up without error

I don't think so.

### DIFF
--- a/libraries/src/Captcha/Captcha.php
+++ b/libraries/src/Captcha/Captcha.php
@@ -224,7 +224,7 @@ class Captcha implements DispatcherAwareInterface
 	{
 		if (method_exists($this->captcha, $name))
 		{
-			return call_user_func_array(array($this->captcha, $name), $args);
+			return call_user_func_array(array($this->captcha, $name), array_values($args));
 		}
 
 		return null;


PhilETaylor:
related to related to https://github.com/joomla/joomla-cms/issues/33529 which is the same issue.

alexandreelise:
Thanks @PhilETaylor for helping Joomla! being PHP8.0 compliant. When I see that PHP8.1 should come out in november and we need to do that again. Hopefully it won't break that much since it's a minor version  PHP8.1. Take care you all Super Joomlers!

wilsonge (approved):

alexandreelise:
Thanks George @wilsonge . I am sad that I missed your presentation about Joomla! 4 Web Services at JUG NJ I hope it was recorded. Take care Super Joomler.

alexandreelise:
Can someone please test this manually ?  Maybe @alikon @bembelimen @chmst

alikon:
didn't have PHP8.0.11 yet , but it looks good to me

alikon (approved):

alexandreelise:
Thanks @alikon This pull request needs Human Test Results. It's required. Do you someone who can test this manually with patchtester?

joomdonation:
@alexandreelise I'm using PHP 8.0.1 (on Windows), Joomla 4 nightly build and could not see the fatal error as you mentioned. Captcha still displays OK for me. Tested with both the captcha plugins come with Joomla.

alexandreelise:
Hi @joomdonation. Thanks for your feedback.
There might be at least 2 reasons for that You are using a different PHP version, you are using nightly joomla build, I'm using the official 4.0.3 at the time I was writing this. And one last reason that i can think of, is that I didn't mention that I am talking about the hcaptcha plugin not the builtin recaptcha one. But both of these plugins use the same Captcha class where there is the "bug" I found. Anyway can you share the code lines of your install we are talking about in this pr just where the bug is? I suspect we don't have the same setup or that it has been fixed in the nightly build maybe with another PR. Who knows.
Take care Super Joomler. Have a great day.

joomdonation:
@alexandreelise I actually used 8.0.11 (not 8.0.1 as I wrote before, just a typo). I had a quick look at the hcaptcha plugin and the real reason which causes this error is from that plugin. The init method of that plugin is defined as below:

```php
public function onInit()
{
	// If there is no Public Key set, then this plugin is no use, so exit
	if ($this->params->get('publicKey', '') === '')
	{
		throw new \RuntimeException(Text::_('PLG_CAPTCHA_HCAPTCHA_ERROR_NO_PUBLIC_KEY'));
	}

	// Load the JavaScript from hCaptcha
	HTMLHelper::_('script', 'https://hcaptcha.com/1/api.js', ['version' => 'auto', 'relative' => true], ['defer' => 'defer', 'async' => 'async']);

	return true;
}
```

It does not have $id parameter (like our core captcha plugins) as expected by the system. Your PR should solve the issue, but I'm unsure if we should make this change or should the developer of this plugin fixes his code. 

With that said, I will wait for feedback from others before doing real test for your PR.

alexandreelise:
@joomdonation There is really a breaking change in PHP8 and I didn't invent it. Here is the official php.net documentation talking about call_user_func_array  breaking change in PHP8 [highlighted link in the official documentation](https://www.php.net/manual/en/migration80.incompatible.php#:~:text=call_user_func_array()%20array%20keys%20will%20now%20be%20interpreted%20as%20parameter%20names%2C%20instead%20of%20being%20silently%20ignored.)

alexandreelise:
Thanks for clarifying the situation @PhilETaylor indeed you already reported a similar related issue

conseilgouz:
I have tested this item :white_check_mark: successfully on 690e52b7bf47821a7b265b8478c88f7abebb4f01<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35675">issues.joomla.org/tracker/joomla-cms/35675</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35675">issues.joomla.org/tracker/joomla-cms/35675</a>.</sub>

HLeithner:
Thanks

alexandreelise:
Thanks to all super joomlers contributing back to the Joomla! project. Have a great day.

PR #35679, opened 2021-09-27
_getUser invalide name fix

HRIT-Florian:
"fullname" is undefined, correct key is "name"

tested with joomla 4.0.3

### DIFF
--- a/plugins/user/joomla/joomla.php
+++ b/plugins/user/joomla/joomla.php
@@ -421,7 +421,7 @@ class PlgUserJoomla extends CMSPlugin
 		$defaultUserGroup = $params->get('new_usertype', $params->get('guest_usergroup', 1));
 
 		$instance->id = 0;
-		$instance->name = $user['fullname'];
+		$instance->name = $user['name'];
 		$instance->username = $user['username'];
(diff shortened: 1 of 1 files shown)


brianteeman:
I'm a little confused by this as this code hasn't changed in 13 years<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35679">issues.joomla.org/tracker/joomla-cms/35679</a>.</sub>

HRIT-Florian:
@brianteeman so maybe these is now the problem, because the passed array is not more valide with the 13 years old code.
fullname is not defined, tested with joomla 4.0.3

joomdonation:
@Didldu-Florian **fullname** is still the right key. Maybe it is not working for you because you are using a none-core authentication plugin? For core plugins, it returns **fullname** in response object like this https://github.com/joomla/joomla-cms/blob/4.0-dev/plugins/authentication/joomla/joomla.php#L86 , and you won't get that undefined error.

HRIT-Florian:
@joomdonation okay, that's it, you are all right. 
Did closed that PR

PR #34671, opened 2021-06-30
[4.0] Crowdin comments

brianteeman:
This PR updates the comments in language strings to a format that crowdin understands as requested by @conconnl https://github.com/joomla/core-translations/issues/1

Basically comments are moved to the end of the line in the *.ini

Note comments which are not for translators such as a comment used to indicate a group of strings have not been changed

PR #35681, opened 2021-09-27
[4.0] Fix language strings causing false plurals in crowdin

richard67:
Pull Request for https://github.com/infograf768/j4localise/issues/30#issuecomment-925585512 .

### Summary of Changes

Language string constants ending with "_ONE" seem to trigger some kind of plural string generation for translations on Crodwin.

See https://github.com/infograf768/j4localise/issues/30#issuecomment-925585512 .

To avoid this, we should not use language string constants ending with "_ONE" if they are not really plural strings.

Work in Progress (WiP): I have to complete testing instructions.

As soon as ready, I'll remove draft status from this PR.

### Testing Instructions

Hint: It needs to run "npm ci" or "npm run build:js" after having applied the PR.

1. Code review: Check if there are language strings in the CMS core having a key ending with "_ONE".
2. In content history modal of some content having some history (i.e. more than 1 version), use the "Preview" button without having selected any version.
Hint: You can use that button even if it visually appears to be disabled. But that's another issue to be solved with another PR.
3. In the same modal, use the "Compare" button with only one version selected.
4. Create a module of type "Tags - Similar" and check the available options in the "Match Type" drop-down on the first Tab "Module".

1. Strings "COM_CONTENTHISTORY_BUTTON_SELECT_ONE" and "MOD_TAGS_SIMILAR_FIELD_ONE" found.

2.
Screen Shot 2021-09-24 at 07 18 06

3.
Screen Shot 2021-09-24 at 07 18 24

4. Options "All", "Any" and "Half" are available.

1. None.

2. Same as without this PR.

3. Same as without this PR.

4. Same as without this PR.

None.


brianteeman:
You might need to check MOD_TAGS_SIMILAR_FIELD_ONE

richard67:
>You might need to check MOD_TAGS_SIMILAR_FIELD_ONE

@brianteeman Yes, that's the one I wanna check. Do you think I should do them all in this PR, or should I better make separate PRs?

brianteeman:
if you are fixing the same problem then you can do them in the same pr. its a small enough a change

richard67:
@brianteeman Are you familiar with CrowdIn? I'm not. Do you know if it would be ok to change "MOD_TAGS_SIMILAR_FIELD_ONE" to "MOD_TAGS_SIMILAR_FIELD_ANY", or do you have an idea why it hasn't been named like that from the beginning on? Does ending with "_ANY" cause similar issues on CrowdIn? Or can we use that?

brianteeman:
Answering no and no idea to each question - sorry

richard67:
@brianteeman Other question: Do you know if it's still ok to have the deprecation comment above the line with the language string? Or should that be put to the end of the line with the string?

brianteeman:
End of the line please https://github.com/joomla/joomla-cms/pull/34671

infograf768:
> "MOD_TAGS_SIMILAR_FIELD_ONE" to "MOD_TAGS_SIMILAR_FIELD_ANY", or do you have an idea why it hasn't been named like that from the beginning on? Does ending with "_ANY" cause similar issues on CrowdIn? Or can we use that?

Some languages may use `ANY` in their plural forms and crowdin may wonder there.

I suggest: `MOD_TAGS_SIMILAR_FIELD_ONE_TAG`

brianteeman:
Wont the plurals still be generated as they are still in the language file?

richard67:
> Wont the plurals still be generated as they are still in the language file?

@brianteeman I think so. But what else can we do? We're not allowed just to remove the strings, are we?

brianteeman:
The arguments for not removing them was always something I did not agree with. It was so that you could use an old and outdated language file with the very latest version of joomla and/or in case the string was used by an extension which is unlikely to apply in this case.

The problem is that if you dont remove the strings then you aren't solving the problem (are you?)

richard67:
> The problem is that if you dont remove the strings then you aren't solving the problem (are you?)

@brianteeman I solve the problem ... for Joomla 5.0.

@wilsonge Do you know what we shall do? Can we remove these strings?

Bakual:
I would just remove them as we're fixing a bug.
And chances are slim that the string is used by 3rd party.

brianteeman:
I have tested this item :red_circle: unsuccessfully on df3e40b93c848d06cd0f21e04bec9b97c267283c<br /><br />Doesnt fix the problem<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35681">issues.joomla.org/tracker/joomla-cms/35681</a>.</sub>

richard67:
@brianteeman l’ll update this PR tomorrow.

richard67:
I've changed this PR so the bad strings are not deprecated but removed.

@brianteeman Please test again. Thanks in advance.

brianteeman:
I have tested this item :white_check_mark: successfully on 3b133cf3741ccbdbc48e256486ae3f58a1c9a489<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35681">issues.joomla.org/tracker/joomla-cms/35681</a>.</sub>

infograf768:
I have tested this item :white_check_mark: successfully on 3b133cf3741ccbdbc48e256486ae3f58a1c9a489<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35681">issues.joomla.org/tracker/joomla-cms/35681</a>.</sub>

infograf768:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35681">issues.joomla.org/tracker/joomla-cms/35681</a>.</sub>

bembelimen:
Thx

richard67:
Thanks all.

Issue #30361, opened 2020-08-13
[4.0] Notification in Workflow can't be read by not administrator user


paternax:
In the transition part of the Workflow component it is possible to create notifactions for users, which have to take the next step in the workflow. The information about the generation of new notification is sent by email which contains  a link to the backend. The notification system is only accessible for administrators, so that users which have only a user group based on Registered can't read the notification.

There should be an option to send the whole notification per email, so that also not administrator user can read it.

brianteeman:
@paternax I'm not sure what the use case would possibly be for a non-administrative user to receive the email. However I can see that workflows is not using the new mail templates which it probably should and that would resolve the problem. If no one beats me too it I will take a look at this

paternax:
For example  frontend editors who submit new blog posts could be informed about the change of stages.

brianteeman:
ah - i see the problem now.

The email received is as folllows

```
The status of an Article has been changed 

Title: Typography, changed by Brian Teeman, new state: Basic Stage. 

Please log in to http://localhost/joomla-cms/administrator/index.php?option=com_messages&view=message&message_id=5 to read your message.
```


> There should be an option to send the whole notification per email, so that also not administrator user can read it.

The **_problem_** is that there is no extra text to that email. It is complete but the link that a non-admin user cannot use suggests that there is something more that they can not see.

brianteeman:
You can "fix" this for now by creating a language override for COM_MESSAGES_NEW_MESSAGE_BODY

brianteeman:
Please test #35693

richard67:
Closing as having a pull request. Please test #35693 . Thanks in advance.

PR #35693, opened 2021-09-28
[4.1] Log in to read messages

brianteeman:
when you have enabled notifications for workflows then the default message will be similar to
```

The status of an Article has been changed

Title: New feature: Workflows, changed by Brian Teeman, new state: Basic Stage.

Please log in to http://localhost/joomla-cms/administrator/index.php?option=com_messages&view=message&message_id=2 to read your message.
```

The last line implies that if you log in you will be able to read more about this message - and you might not even have the permissions to log in and read. However the online version of the message does not contain anything extra so there is no need to log in at all.

This PR changes the string to remove the information about logging in

```
The status of an Article has been changed

Title: New feature: Workflows, changed by Brian Teeman, new state: Basic Stage.
```

Pull request for #30361


Kostelano:
Yes, but this is a personal message template. And it contains tons of other available tags.

What is the point of using this template for workflows if, in fact, this template has become useless, because breaks another unrelated component...

brianteeman:
99% sure that you are mistaken

chmst:
Looks good for me. The message is only that a transition has been performed and is complete. 

@Kostelano Sorry,  I don't understand. Which template do you mean?

Kostelano:
@chmst I'm talking about changing the email template. At the moment this template has other tags available that I can use in the body of the email. Considering that this template is also used for workflows (as we see in the original problem), it turns out to be a kind of mess - we'd better NOT use some tags, because they will be irrelevant in workflow emails.

Screenshot_1

brianteeman:
The mail template takes its values from sql not the language string iirc. Either way even in that mail template it is a useless string for the same reasons mentioned above

paternax:
The solution sounds good. 
At the moment not possible to test, because the workflow in J.4.0.3. sends no notifications, even not to super user.

brianteeman:
@paternax I was successfully getting emailed notifications  BUT I suspect you hit another problem with notifications using com_messages in that it wont send a notification to you if you are the one that made the change and triggered the notification. Could you please create another super user and then see if they get the notifications

paternax:
Yes you are right, that is exactly the problem. I tested this with two superusers and an explicit user in notifications. The same by using the usergroup superuser only the other superuser, who didn't made execute the transition gets an email.

brianteeman:
@paternax which is logical but perhaps not expected

brianteeman:
rebased to 4.1

Quy:
I have tested this item :white_check_mark: successfully on 56e181c000e5d5a9abec97f64bf013e1fe587224<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35693">issues.joomla.org/tracker/joomla-cms/35693</a>.</sub>

khu5h1:
I have tested this item :white_check_mark: successfully on 56e181c000e5d5a9abec97f64bf013e1fe587224<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35693">issues.joomla.org/tracker/joomla-cms/35693</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35693">issues.joomla.org/tracker/joomla-cms/35693</a>.</sub>

bembelimen:
Thx

brianteeman:
thanks

Issue #35625, opened 2021-09-21
[4.x] CLI command cli/joomla.php - Notice: Undefined index: HTTP_HOST


Kubik-Rubik:
### Steps to reproduce the issue

With the command `php cli/joomla.php list`, you get a list of all available CLI commands.

The script throws a PHP notice if the system plugin "System - Page Cache" is enabled:

**Notice: Undefined index: HTTP_HOST in /xxx/libraries/src/Uri/Uri.php on line 103**

### Expected result

No errors, regardless of what plugins are activated.

### Actual result

See the error above.

### System information (as much as possible)

Joomla! 4.0.3, PHP 7.x & 8.x

brianteeman:
Confirmed

alikon:
please test #35722

anibalsanchez:
From my comment in #35722:

_About #35625, the problem is that the `php cli/joomla.php list` is firing the "System - Page Cache" when it's called from the console. Why would a console application fire the "System - Page Cache" plugin? It shouldn't. By default, a console application doesn't have to interact with the site cache._

The joomla console application calls to `PluginHelper::importPlugin('system');` to initialize all system plugins. So, the problem seems to be that the "System - Page Cache" plugin tries to read the URI when it is executed in the console context. In my opinion, the plugin should do nothing and ignore the execution if it is the console context.

PS: Now commenting in the right issue ;-)

PR #35722, opened 2021-10-02
[4] Set HTTP_HOST from CLI

alikon:
Pull Request for Issue #35625 .

### Summary of Changes
set a default

### Testing Instructions

> With the command php cli/joomla.php list, you get a list of all available CLI commands.
The script throws a PHP notice if the system plugin "System - Page Cache" is enabled:

### before
Notice: Undefined index: HTTP_HOST in /xxx/libraries/src/Uri/Uri.php on line 103

### after 
No Notice

no

### DIFF
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -123,6 +123,9 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 
 		// Register the client name as cli
 		$this->name = 'cli';
+		
+		// Set HTTP_HOST
+		$_SERVER['HTTP_HOST'] = 'localhost';
 
 		$this->setContainer($container);
 		$this->setDispatcher($dispatcher);
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -123,9 +123,6 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 
 		// Register the client name as cli
 		$this->name = 'cli';
-		
-		// Set HTTP_HOST
-		$_SERVER['HTTP_HOST'] = 'localhost';
 
 		$this->setContainer($container);
 		$this->setDispatcher($dispatcher);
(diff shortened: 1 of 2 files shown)


PhilETaylor:
Although this mutes the PHP Notice, it is not the correct fix. Look at Uri.php and you will see the Notice comes when Joomla assumes you are running IIS 

```
/*
					 * Since we do not have REQUEST_URI to work with, we will assume we are
					 * running on IIS and will therefore need to work some magic with the SCRIPT_NAME and
					 * QUERY_STRING environment variables.
					 *
					 * IIS uses the SCRIPT_NAME variable instead of a REQUEST_URI variable... thanks, MS
					 */
```

The correct fix is to make Uri.php CLI Aware and not have it try to generate http urls using invalid data.

alikon:
make more sense now ?

anibalsanchez:
@alikon I don't think that it is a good idea to prepare `Uri` to understand the Application context and change the internal behavior according to the execution.

If an object has to use "isClient", then it is a bad signal and it is better to think twice about what we are doing.

The problem is the **application context mismatch**, why the current Application is calling Uri without sending the right parameter.

As I pointed out in #35818, the problem is that the Content Component boots the Site Application with `$container->get(SiteApplication::class)`. The Content Component doesn't have to force the Site Application.

About #35625, the problem is that the `php cli/joomla.php list` is firing the "System - Page Cache" when it's called from the console. Why would a console application fire the "System - Page Cache" plugin? It shouldn't. By default, a console application doesn't have to interact with the site cache.

ceford:
I have tested this item :white_check_mark: successfully on a510896acb3c20fbfd9a6d244663195084509015<br /><br />With the code and tutorial I wrote yesterday I can confirm this works as described.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35722">issues.joomla.org/tracker/joomla-cms/35722</a>.</sub>

anibalsanchez:
I have tested this item :red_circle: unsuccessfully on a510896acb3c20fbfd9a6d244663195084509015<br /><br />No No No. 

Faking an URL 'localhost' into $_SERVER['HTTP_HOST'] is not the way to solve it.

The problem is the "System - Page Cache" plugin executing in the console context. The plugin must ignore the execution when it is in the console context.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35722">issues.joomla.org/tracker/joomla-cms/35722</a>.</sub>

alikon:
closed as per #35877 merge

Issue #35349, opened 2021-08-24
[4] Label and description not clear in configuration


chmst:
### Steps to reproduce the issue
Global configuration

![grafik](https://user-images.githubusercontent.com/1035262/130655693-adaf5624-8e56-45b9-89b3-60518753f9c9.png)

The description is not clear - what is logged with "no" and what with "yes"?

simbus82:
Usually in the softwares the logging is "fully" activated and then the user is given the choice to disable the logging of something that could be too verbose or not very informative.

Here it is not just a matter of the label and description, but it is an issue of the intended behaviour for those switches.

The most "user understandable" solution, is to have a "master Log Switch" ex. "Logging Enabled", while the second switch should say "Disable logging of deprecated API".

brianteeman:
please test #35733

alikon:
test #35733

PR #34966, opened 2021-07-29
[4.0] Correct description for custom logging

Fedik:
### Summary of Changes
The description for Log Categories is incorrect.
When the field is empty then custom log is disabled:

https://github.com/joomla/joomla-cms/blob/b1f464ca41320b6117bd78d2d1a17607a86bf137/libraries/src/Application/CMSApplication.php#L1400-L1402

### Testing Instructions

Code review

### Documentation Changes Required

If there a doc about it, then it also need to update.

PR #35733, opened 2021-10-03
[4.0] Log everything string

brianteeman:
Change label to "enable logs" which makes more sense and is more consistent with other labels

Pull Request for Issue #35349

image


ChristineWk:
I have tested this item :white_check_mark: successfully on d80fa5123d9e0e5fe4b3940301053b106b121dba<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35733">issues.joomla.org/tracker/joomla-cms/35733</a>.</sub>

Fedik:
It is misleading now.

It is exactly what it say "log everything":
https://github.com/joomla/joomla-cms/blob/0c41f6b92008abd255c2451347024c4f1f96b342/libraries/src/Application/CMSApplication.php#L1405-L1409

User may assume that log is disabled when it Off, but it is not. 
Log also may be enabled if you add any category to "Log Categories": https://github.com/joomla/joomla-cms/blob/0c41f6b92008abd255c2451347024c4f1f96b342/libraries/src/Application/CMSApplication.php#L1411

The fix for "Log Categories" description is there #34966

brianteeman:
So "Custom logging" really should be called "Logging Options" and the fields reworked to use showon?

Fedik:
hm, not really, Custom logging is about logging of specific category(-ies).
When "Log categories" is empty then Custom logging is disabled,
when example you add `jerror,database` then it will log only these categories, no mater if "Log everything" On or Off.

It is 2 different sections, which migrated from debug plugin.

brianteeman:
wow this really is a mess. Can't be fixed with a string

Fedik:
but please test  #34966 that is a fix for misleading description :)

PR #21037, opened 2018-07-10
[4.0] com_finder: Filtering common & numeric words and language fix

Hackwar:
This PR filters common words and numeric terms from the index and the search query and also fixes a multilanguage issue in com_finder.

Searching for ```I, and, or, it,...``` in com_finder will give you lots of results, but might not be really usefull. So it is common for index based searches, to filter such common words both from the index (in order to keep it smaller) and from the search queries in order to only search for the relevant words. With #20781 I provided a way to add language specific sets of common words, this PR provides a switch to filter out these common words and a second one to filter out numeric terms.

I also found an issue with the multi-language capabilities of com_finder. Up till now, com_finder did not differentiate between terms from different languages. As long as they were written the same, they were okay. However with the stemming and the common words per language, we run into issues here. All of a sudden, a common word might still appear in the index for a content item, because it is correctly filtered out in english, but in the french article that has an english citation, it is still included. Thus it all of a sudden reappears and has some (albeit minor) negative effects on the search results. We are also having issues with different words from different languages which have the same stemmed form. Thus this change now does not silently drop identical words from different languages, but instead inserts each duplicate term with its specific language flag.

Please note that this PR has DB updates which need to be applied before testing.

PR #35740, opened 2021-10-03
filter_numeric doesnt exist

brianteeman:
redo of #35542

This PR fixes the bug reported in #35539 AND removes the Filter Numeric Terms option and related code as it is not complete and doesnt do anything at all


Quy:
See #21037. Instead of removing the option, change 5 instances of `filter_numeric` to `filter_numerics`.

Install Blog Sample Data.
Click `Clear Index`.
Click `Index`.
Click `Statistics`. 

Option disabled: 522 terms
Option enabled: 511 terms

PR #35143, opened 2021-08-15
[4.1] [SOC 21] Add a Scheduled Tasks Infrastructure to Joomla

ditsuke:
> ### **Project Repository:** [joomla-projects/soc21_website-cronjob](https://github.com/joomla-projects/soc21_website-cronjob/).

## Introduction

Many web hosting services do not offer access to the native UNIX cron. A popular solution to scheduling tasks with such servers is using a lazy scheduler (or, 'pseudo-cron'). Other major CMS like Drupal and WordPress already offer some form of pseudo-cron functionality.

However, there is currently no standard way to schedule tasks in the Joomla! core. There is also no way to integrate native cron with Joomla! if the server offers access to this functionality.

This PR introduces a Scheduled Tasks infrastructure to Joomla! As of the last update to this description, this infrastructure includes a Scheduler component, a system plugin to run the Scheduler on each page load implemented in the first half of my work under SOC '21. Also included is a test task plugin along with 2 functional job plugins:
- Plugin *Task - Requests* offers a GET Request task routine with the ability to set authorization headers and a timeout.
- Plugin *Task - Site Status* offers task routines to change the site's offline status. This can be very useful with say a Cron-style execution rule, so it can be used to turn a site offline for a specific timeframe.

## Summary of Changes

1. **Scheduler Component:** The component provides the infrastructure for Scheduled Tasks, which includes the MVC to manage tasks, the plugin API necessary for plugins to advertise and run task routines, a helper for execution rules (currently supported: intervals and cron-style expressions), a trait for task plugins, and a Scheduler API class.
2. **Cronjobs System Plugin:** Subscribes to the `onBeforeRender` event, and registers a method to run the Scheduler as a shutdown_function. At most a single scheduled task is executed per request. The task execution is logged to the `joomla_scheduler.php` log file.
3. **Requests Task Plugin:** Offers a task routine for GET Requests with support for auth headers and request timeout.
4. **Site Status Task Plugin:** Offers task routines to change the site's offline status.
5. **Database:**
  - Adds a `#__scheduler_tasks` database table.
  - Adds `com_scheduler` as an `#__assets` entry.
  - Adds `#__extensions` entries for `com_scheduler`, `plg_system_schedulerunner`, `plg_task_demotasks`, `plg_task_requests`, `plg_job_sitestatus`.
6. **System Dashboard:** Adds the Scheduled Tasks Manager under the Manage module.
  

## Planned Work
Right now, the major planned milestones include:
- [ ] Joomla Console commands:
[...]

PR #35056, opened 2021-08-06
[4.0] Fix state filter options in workflows manager

ditsuke:
### Summary of Changes
Fixes the options on display by using the right `optionsFilter` attribute instead of `filter`.

### Testing Instructions
Go to the workflow manager for some extension, for example `yoursite.example/index.php?option=com_workflow&extension=com_content`

### Actual result BEFORE applying this Pull Request
The state filter lists states not supported by com_workflows

image

### Expected result AFTER applying this Pull Request
The state filter only lists states supported by com_workflows (the labels are still not right)

image

### Documentation Changes Required
None

PR #35743, opened 2021-10-04
[5.0] Add alternative field labels (+ a new attribute) for the Status field

ditsuke:
### Summary of Changes
Many extensions currently use custom status fields ([PluginStatusField](https://github.com/ditsuke/joomla-cms/blob/dcon/libraries/src/Form/Field/PluginstatusField.php), [RedirectStatusField](https://github.com/ditsuke/joomla-cms/blob/dcon/libraries/src/Form/Field/RedirectStatusField.php), [TaskStateField](https://github.com/ditsuke/joomla-cms/blob/dcon/administrator/components/com_scheduler/src/Field/TaskStateField.php) (from #35143)), mainly because it makes more contextual sense to use Enabled/Disabled as states at some places instead of Published/Unpublished. 
This also increases scope for inconsistencies which can be seen as an example in the forms for `com_workflows` (reference #35056). 

This PR adds alternative labels the `status` field which can be used with the `alt_labels` field attribute. This can be used in conjunction with `optionsFilter` at places where a list field or a custom field might be needed right now.

For example, if you want to support the enabled, disabled and trashed state for an item, you could use:
```xml
<field 
	name="state"
	type"status" 
	label="JSTATUS"
	optionsFilter="-2, 0, 1"
	alt_labels="true"
/> 
```

Right now, you'd have to contend with a custom field, a list field everywhere or a language override (https://github.com/joomla/joomla-cms/pull/35743#issuecomment-933180203).

### Testing Instructions
You can test that StatusField (field type="status") behaves the same as before in all existing forms.
I can also add the `alt_labels` attribute to the workflow forms and some more places in this PR so those could be tested here and perhaps some redundant field classes removed.

StatusField only supports "publish" type labels.

StatusField supports alternative state labels needed by a number of existing core extensions.

Update to the Form Fields/StatusField documentation on JDocs.


brianteeman:
Could you give an example please of where this would be used

ditsuke:
> 
> 
> Could you give an example please of where this would be used

Workflow forms (I linked a PR for reference), replacing PluginStatusField, RedirectStatusField and TaskStatusField (from Scheduler).

brianteeman:
Personally I have always done this with just a language string

For example (together with your fix for workflows) it will look like this
image

Just by adding two language string overrides to com_workflows.ini
image

ditsuke:
> 
> 
> Personally I have always done this with just a language string
> 
> For example (together with your fix for workflows) it will look like this [image](https://user-images.githubusercontent.com/1296369/135803232-97e0a15b-c52c-437d-b5a5-00303a49cca1.png)
> 
> Just by adding two language string overrides to com_workflows.ini [image](https://user-images.githubusercontent.com/1296369/135803308-a1763584-08ce-4d0c-af44-77ac97fc4b7e.png)

That's a good solution! I think this change should still be worth considering as it might be easier perhaps to use an attribute for the safe effect as its used in quite a few places.

ditsuke:
Rebased to `4.1-dev`

fancyFranci:
I have tested this item :white_check_mark: successfully on 255d84099782b565094e32f646b1f1f8c5abc1ec<br /><br />I tested the alt_labels attribute in several filters like categories and the patch worked. Enabled and Disabled showed up.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35743">issues.joomla.org/tracker/joomla-cms/35743</a>.</sub>

HLeithner:

HLeithner:

brianteeman:
Looking at this again and I realise now the benefits. Together with the optionsFilter this PR would allow us to remove several fields as they would become redundant.

brianteeman:
could someone change the title to say [5.0] instead of 4.1 please

brianteeman:
I have tested this item :white_check_mark: successfully on bd4110616d3d68f34e6257aa1a65be5a3feccd3a<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35743">issues.joomla.org/tracker/joomla-cms/35743</a>.</sub>

HLeithner:
The proper way would be to create an addition field with other default values, then you don't need an extra attribute.

Something like StateField instead of StatusField.

richard67:
Although it has 2 successful human tests I don't set it to RTC because it has a conflict in file "libraries/src/Form/Field/StatusField.php".

And @HLeithner 's comment above suggest a different solution.

HLeithner:
as said own field instead of additonal attribute please. I'm closing this but can be reopend when we get it's own field or a more flexible way to change the label of the options (maybe swapable language constants)

PR #35745, opened 2021-10-04
[4.1] Enable use of snake_case for form entities

ditsuke:
### Summary of Changes
Enables **use of `snake_case`** style for custom fields (and rules and filters) **in form XML** by adding an additional `class_exists()` checks for entity types normalized to CamelCase. Currently, using `snake_case` is not possible unless a class alias is created through the class map.

With this PR, it will be possible to define a custom field class, for example, in pure CamelCase => `ComplexTypeField` and then use `<field ... type"complex_type"/>` to use it in the XML.

This change should be non-breaking as it only adds additional checks without altering existing code.

### Testing Instructions
For custom fields with CamelClass compliant class names (like `WorkflowComponentSections`), you can replace their usage in forms with snake_case.

`snake_case` style in XML cannot be used for custom fields with CamelCase compliant class names.

Using `snake_case` for custom fields like the `WorkflowComponentSections` should be possible. Existing forms should work as expected.

None I'm aware of.

### DIFF
--- a/libraries/src/Form/FormHelper.php
+++ b/libraries/src/Form/FormHelper.php
@@ -206,12 +206,15 @@ class FormHelper
 		// Check if there is a class in the registered namespaces
 		foreach (self::addPrefix($entity) as $prefix)
 		{
-			// Treat underscores as namespace
+			// Normalise type into space separated words
 			$name = Normalise::toSpaceSeparated($type);
+
+			// Treat underscores as namespace
 			$name = str_ireplace(' ', '\\', ucwords($name));
 
 			$subPrefix = '';
 
+			// If type is of form `context.property`, we break-off `context` as sub-prefix
 			if (strpos($name, '.'))
 			{
 				list($subPrefix, $name) = explode('.', $name);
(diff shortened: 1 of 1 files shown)


Fedik:
Fields like `foo_bar` may be in use. They converted to stuff like `JFormFieldFoo_Bar`, not sure about Namespaces.
Here need to be careful with b.c.

ditsuke:
> 
> 
> Fields like `foo_bar` may be in use. They converted to stuff like `JFormFieldFoo_Bar`, not sure about Namespaces. Here need to be careful with b.c.

I think we should be good wrt B/C with this PR as we're only adding another check for matches. There are some `foo_bar` usages made possible right now through aliases in the [classmap](https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/classmap.php#L336). 

I don't know much about the B/C implications but maybe we can move to real camelCase class names and snake_case in forms someday.

brianteeman:
What is the purpose of this PR. There is a codestyle that says it should be camelcase

ditsuke:
> 
> 
> What is the purpose of this PR. There is a codestyle that says it should be camelcase

I'm unaware, is there one that says it should be camelcase in the XML too?

Fedik:
I have checked. 
> Currently, using snake_case is not possible unless a class alias is created through the class map.

`snake_case` already works, in next way:  `type="complex_type"` => `Complex\TypeField`
See for example: https://github.com/joomla/joomla-cms/blob/63d4e65cda6044e02f4b06adae104f56db882a09/components/com_content/tmpl/article/default.xml#L19
And field class: [Modal/ArticleField](https://github.com/joomla/joomla-cms/blob/4.0-dev/administrator/components/com_content/src/Field/Modal/ArticleField.php)

If you want `ComplexTypeField` then you have to use: `type="complexType"` or `type="complex.type"` 

I would not introduce another behavior for it.

I am closing it.
Thanks for your work.

ditsuke:
Fair enough. Thank you for having a look :)

PR #35747, opened 2021-10-04
[4.1] Remove RIPS analysis step

SniperSister:
### Summary of Changes
Remove RIPS analysis step from drone.yml

RIPS as a standalone product has been discontinued and merged into an existing SAAS service, targeting to be a generic static code analysis platform. The "noise to signal" ratio isn't suitable to for us a project, that's why JSST decided to remove the scanning step from the drone.yml without replacement.

### Testing Instructions
Wait for drone to finish.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -181,25 +181,6 @@ steps:
       status:
         - failure
 
-  - name: analysis4x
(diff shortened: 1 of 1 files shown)


HLeithner:
Thanks...

PR #35184, opened 2021-08-17
[6.0] Modification fields header tag, module tag   add custom options in XML

korenevskiy:
I periodically encounter the problem of adding my own unique tags for the module.
But now you can add tags in the new module: LI , and others. For cases of non-standard situations.

```
<fieldset>
  <legend>Title</legend>
 Module content
</fieldset>
```

Example testing
```
<field
	name="header_tag"
	type="headertag"
	label="COM_MODULES_FIELD_HEADER_TAG_LABEL"
	default="h5" >
                    <option value="figcaption">&lt;figcaption></option> 
                    <option value="legend">&lt;legend></option>
                    <option value="label">&lt;label></option> 
                    <option value="summary">&lt;summary></option>   
</field>
<field 
	name="module_tag" 
	type="moduletag" 
	label="COM_MODULES_FIELD_MODULE_TAG_LABEL"
	default="div" >
                    <option value="figure">&lt;figure></option> 
                    <option value="fieldset">&lt;fieldset></option>
                    <option value="li">&lt;li></option> 
                    <option value="details">&lt;details></option>  
</field>
```

For testing purposes
1. Open any XML module file.
2. You need to find a field with the type `headertag` or with the type `moduletag` in XML file, if there are no such fields, then you need to write the new fields as specified in the documentation [Header tag](https://docs.joomla.org/Special:MyLanguage/HeaderTag_form_field_type) , [Module tag](https://docs.joomla.org/ModuleTag_form_field_type) , Considering also that the field must have a closing part `</field>`
3. Add several list options inside this field.
4. Save the XML module file.
5. Open the module form and check in the fields of the header tag and the module tag, new tags should appear in the drop-down list, which were optionally added to the list.

PR #35750, opened 2021-10-04
[J4]Add extenstions   parent::getoptions() to field

korenevskiy:
I have added extenstions   parent::getoptions()  to tags based on the list tag.
So that you can add your own options to the list tags in the XML file.
It is very often necessary to use your own tags in the module.
You also need to use "Custom HTML Tags"
Without this fix, we are hard-wired to standard tags.

```
<field
                    name="module_tag"
                    type="moduletag"
                    default="fieldset" >
                    <option value="super-calendar">Super Calendar</option> 
                    <option value="custom-tabs">Tabs</option>
                    <option value="fieldset">Fieldset</option>
</field>
```

### DIFF
--- a/libraries/src/Form/Field/HeadertagField.php
+++ b/libraries/src/Form/Field/HeadertagField.php
@@ -36,7 +36,7 @@ class HeadertagField extends ListField
 	 */
 	protected function getOptions()
 	{
-		$options = array();
+		$options = parent::getOptions();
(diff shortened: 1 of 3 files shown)


brianteeman:
the image changes should be in a separate pull request

korenevskiy:
> the image changes should be in a separate pull request

OK

Quy:
Duplicate #35184

PR #35755, opened 2021-10-05
[J4] add autofocus in module finder 

korenevskiy:
I added setting in module mod_finder. In default this setting disabled.
Seted autofocus where page loaded, this focus set as default for this field.


brianteeman:
This is a really bad idea. 

https://brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

korenevskiy:
> This is a really bad idea.
> 

At the moment, I am working on a site with accessibility support.
In the document by the link you gave me.
It is written that one must be careful not to disrupt accessibility.
This idea does NOT violate accessibility. Only poor use of this feature will disrupt accessibility.
I'm making a website header. I have placed the search field at the very beginning of the page, so that the sequence of TAB switches is not disrupted.

The Joomla search engine has become very powerful, it's not just a search for matching words in the headlines.


But even so, autofocus in the module is turned off by default.

Do you think it would be convenient for you personally to go to the store and immediately have the search field in focus, similar to how you go to the Google site, bing.com , yahoo.com , duckduckgo.com .

Sites today are becoming so complex that page search is more important than the menu and categories of the site.
.
We can hide this setting in the advanced settings tab.


Let's come to a better solution.
Metaphor: in order to avoid knife attacks on the streets of the city, we can take knives from everyone, including the best chefs.
But still let's think about it

korenevskiy:
@brianteeman 
I've been doing accessibility for a long time.
As we can see, I added the focus setting not for the feedback form, and not for the comments form. I added a focus for site search.
Therefore, the idea of accessibility is very good, and I stick to it.
I stick to it by adding this setting to the search form. Since this autofocus does not apply to feedback and comment forms.

korenevskiy:
@brianteeman
I moved the setting to the advanced module settings.

brianteeman:
I stand by my comments.

rdeutz:
There might be situations where this is not a bad idea. But this will give a bad example on how to use autofocus and used in core could lead others to do the same. It is also needed that you have only one element with autofocus on a page. This is better done with an overwrite or a site specific setting. 

I am closing this for the reasons stated.

korenevskiy:
> But this will give a bad example on how to use autofocus and used in core could lead others to do the same.

It is very similar to the fact that if parents do not tell the child how children are born. That means he will grow up and not find out about it.

PR #35704, opened 2021-09-30
[4.0] Update codemirror

brianteeman:
**29-09-2021: Version 5.63.1:**
Fix an issue with mouse scrolling on Chrome 94 Windows, which made scrolling by wheel move unusably slow.

**20-09-2021: Version 5.63.0:**
Fix scroll position jumping when scrolling a document with very different line heights.
xml mode: Look up HTML element behavior in a case-insensitive way.
vim bindings: Support guu for case-changing.

**20-08-2021: Version 5.62.3:**
Give the editor a translate=no attribute to prevent automatic translation from modifying its content.
Give vim-style cursors a width that matches the character after them.
merge addon: Make buttons keyboard-accessible.
emacs bindings: Fix by-page scrolling keybindings, which were accidentally inverted.

PR #35766, opened 2021-10-05
[4.0] Bootstrap 5.1.2

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Update BS to 5.1.2

### Testing Instructions

Check that all the templates are still ok

PR #35769, opened 2021-10-06
[4] Update codemirror.xml

PhilETaylor:
Following https://github.com/joomla/joomla-cms/pull/35704 whenever I run a [jinstall](https://docs.joomla.org/J4.x:Setting_Up_Your_Local_Environment) Im left with a codemirror.xml version change modified file

Is this wrong? what needs changing? 

@brianteeman

### DIFF
--- a/plugins/editors/codemirror/codemirror.xml
+++ b/plugins/editors/codemirror/codemirror.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <extension type="plugin" group="editors" method="upgrade">
 	<name>plg_editors_codemirror</name>
-	<version>5.63.1</version>
+	<version>5.62.2</version>
 	<creationDate>28 March 2011</creationDate>
 	<author>Marijn Haverbeke</author>
 	<authorEmail>marijnh@gmail.com</authorEmail>


brianteeman:
It was broken by #35518 

image

dgrammatiko:
> It was broken by #35518

Actually, it was broken because someone forgot to update the package.json to the correct version of codemirror.

Basically the package.josn needs an update here not the XML...

https://github.com/joomla/joomla-cms/blob/7e7c4d8e0a422b08a3364aba93955ced1ef8dcdb/package.json#L43

PhilETaylor:
This PR was only showing what was showing as changed so that you could see it in context, I was not proposing this as the fix.

Brian specifically said to somebody else in his Pr that he did not need to update the json file￼, only the lock file so now I'm confused. Haha

￼

brianteeman:
package.json doesn't need to be updated

dgrammatiko:
> package.json doesn't need to be updated

You're wrong here. I always do `npm install` thus I always read the package.json so just updating the lock file won't work for me

dgrammatiko:
@PhilETaylor it's also fixed with https://github.com/joomla/joomla-cms/pull/35766. Actually here you're downgrading which is not what we want

PhilETaylor:
I was not downgrading on purpose ha ha I was simply running the usual command and this was the result

If you think it is fixed and that of a PR then we can close this now that it has been discussed and resolved elsewhere￼￼

PR #35772, opened 2021-10-06
[4.0] pagination lists

brianteeman:
The pagination lists really should be ol (ordered lists) and not ul (unordered lists) as clearly the items in the list are in a specific order ;)

To test apply pr and then check that the pagination in any admin list view and any site list view still works and is displayed the same

### DIFF
--- a/layouts/joomla/pagination/links.php
+++ b/layouts/joomla/pagination/links.php
@@ -63,7 +63,7 @@ if ($currentPage >= $step)
 			<?php endif; ?>
 
 			<?php if ($showPagesLinks) : ?>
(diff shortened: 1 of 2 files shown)


chmst:
I am not in favor of this change. 
The list is displayed in a specific order as every list, like menus, tags and everything else. The sequence of the list elements is controlled via params. But there is no ordering like in a receipe for example. The user can click whatever number first and whatever number second. 
https://www.w3.org/WAI/tutorials/page-structure/content/#ordered-list

brianteeman:
Sorry you're correct. It was late last night.

PR #35776, opened 2021-10-07
[4.1] Invalid caching of WebAssets

Denitz:
### Summary of Changes

1. `Cache::setWorkarounds()` incorrectly returns ALL current assets even if `headerbefore` option exists and a module/view didn't actually add any own assets.

This results in incremental caching of all assets in the cache of each module. Tons of WebAssets events are next executed on module cache load via `Cache::getWorkarounds()` and the load time is seriously affected.

2. Iterating over `$wa->getManagerState()['activeAssets']` does not requires checking the `ASSET_STATE_ACTIVE` state because this array holds only active assets

3. The resulting array of `getHeadData()['assetManager']['assets']` didn't have the asset name as key which prevents the fast cache workarounds comparison

### Testing Instructions

Enable Joomla conservative caching, browse frontend pages as guest

Explore the contents of module cache files, see that they contain all `WebAssetItem` objects existing at the time of caching.
Note the cache file size.

See that cache file size in much smaller now (delete old files first), see that the module cache now only have the assets really added by this module.

No.


libraries/src/Document/HtmlDocument.php
```diff
@@ -172,10 +172,7 @@ public function getHeadData()
 		{
 			foreach ($assetNames as $assetName => $assetState)
 			{
-				if ($assetState === WebAssetManager::ASSET_STATE_ACTIVE)
-				{
-					$waState['assets'][$assetType][] = $wa->getAsset($assetType, $assetName);
-				}
+				$waState['assets'][$assetType][$assetName] = $wa->getAsset($assetType, $assetName);
```
Fedik, line 176: Please revert it, this was made to skip assets with `WebAssetManager::ASSET_STATE_DEPENDENCY`
to reduce amount of cached data.

There a comment above, at line 167.

Fedik: hmhm, if dependency registered at runtime, then better keep it.
well, ignore my previous comment :smile:

Fedik (approved):
looks good to me

libraries/src/Document/HtmlDocument.php
Denitz: `$waState['activeAssets']` holds ONLY assets with `WebAssetManager::ASSET_STATE_ACTIVE` status

Fedik: > holds ONLY assets ASSET_STATE_ACTIVE

nope, not always,
but nevermind

Fedik:
I have tested this item :white_check_mark: successfully on 2c1296222529c01cd260d3e6d77ce5119ccfcdc2<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35776">issues.joomla.org/tracker/joomla-cms/35776</a>.</sub>

Denitz:
Please check https://github.com/joomla/joomla-cms/pull/37139 with better approach.

richard67:
> Please check #37139 with better approach.

@Denitz Does that PR replace this one here? If so, this one here should be closed.

PR #35777, opened 2021-10-07
[4.0] Move com_templates css to separate file

C-Lodder:
### Summary of Changes

`/atum/scss/pages` contains page specific styling, which should be moved to separate CSS files, so that their only loaded on said specific page, rather than globally.

This PR moves the `com_templates` styling to a separate CSS file

### Testing Instructions

1. Run `npm i`
2. Go to `administrator/index.php?option=com_templates&view=styles`
3. Open the Cassiopeia template
4. Go to the `Menu Assignment` tab

The styling should be the same as before, but you should be able to confirm it derives from the separate CSS file:

image

image

### DIFF
--- a/administrator/components/com_templates/tmpl/style/edit_assignment.php
+++ b/administrator/components/com_templates/tmpl/style/edit_assignment.php
@@ -20,7 +20,8 @@ $user      = Factory::getUser();
 
 /** @var Joomla\CMS\WebAsset\WebAssetManager $wa */
 $wa = $this->document->getWebAssetManager();
-$wa->useScript('com_templates.admin-template-toggle-assignment');
+$wa->useScript('com_templates.admin-template-toggle-assignment')
+	->useStyle('com_templates.admin-template-toggle-assignment');
 
 ?>
 <label id="jform_menuselect-lbl" for="jform_menuselect"><?php echo Text::_('JGLOBAL_MENU_SELECTION'); ?></label>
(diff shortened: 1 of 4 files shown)


brianteeman:
will you be moving all the other pages and does it really matter for so few lines?

C-Lodder:
If this is desired, then I suppose I can do the other pages.

> does it really matter for so few lines

It's not so much the line count that matters, but spliting page specific CSS rather than just bundling everything up into a single file. No point in loading unused CSS where possible.

brianteeman:
i dont really see the need myself for such micro-optimisation in the admin but if its done for one "page" it really should be done for all.

dgrammatiko:
> i dont really see the need myself for such micro-optimisation in the admin 

It's a good thing to actually demonstrate best practices. We did this already for vendor (I think I did that) and the reason is cache invalidation, eg if we have template and fontawesome in one CSS file and any of them has some changes the users have to re download both but that's not the case if the files are not bundled (template.css+fa.css ~ 240kb but template.css ~ 180kb and fa.css ~ 80kb).
BTW this will only work if the new web assets are used as the old methods will **always** invalidate **all** the static assets per joomla release.

I'm in favour of this. My 2c

brianteeman:
maybe its just me then that doesn't see the point in the admin template of doing this

simbus82:
Nice! We have some projects where the backend admin is (from a commercial perspective) a "frontend" for many users, so every optimization is welcome!

Issue #35779, opened 2021-10-07
menu link image not showing


greenlizzard24:
### Steps to reproduce the issue
created a menu item, in the tab for link type, I selected a png image, the image displays in the box but does not display in the menu item. 

### Expected result
menu link with an image next to it

### Actual result
Nothing, blank

### System information (as much as possible)
Joomla 4.0.3
PHP 7.4.24
Database Type	mysql
Database Version	10.3.31-MariaDB-log-cll-lve

### Additional comments
I can select the image but the link is showing a rather strange format, where it should be images/myimage.png instead it is showing images/myimage.png#joomlaimage://local-images/myimage.png?width=41&height=17

dgrammatiko:
If anyone wants to fix this one probably they might be interested to also fix all the other places in the core that images are broken. Here's a good list: https://github.com/ttc-freebies/plugin-responsive-images/tree/4.0.0/src/plugins/system/responsiveplgoverrides/html

dgrammatiko:
@greenlizzard24 please test https://github.com/joomla/joomla-cms/pull/35780

chmst:
Closing as there is a PR. 

But - I see all images .png or .jpg as expected. So do cannot confirm the issue. @dgrammatiko what am I missing?

dgrammatiko:
@chmst check my PR

chmst:
I see and I am a big fan of HtmlHelper for images, so looks really good for me. 
But I got the images also before your patch. So do not understand what was the issue.

dgrammatiko:
> But I got the images also before your patch. So do not understand what was the issue.

Maybe the particular issue here is due to some override or something else custom (module, template, etc). But it was a good reminder that images needed some attention...

PR #35780, opened 2021-10-07
[4.0] Fix images

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/35779.

# THIS IS A RELEASE BLOCKER

### Summary of Changes
- Use the helper to clean the URL
- Use the HTMLHelper consistently
- Based on the https://github.com/ttc-freebies/plugin-responsive-images/tree/4.0.0/src/plugins/system/responsiveplgoverrides/html that had solved these before 4.0 but somehow I forgot to do a PR here...

### Testing Instructions

Test that images in all these views are rendering correctly.

Create a new contact and use the image field
Create a menu item contact category and check the front end that the image html is not having the hash part
Create a menu item single contact and check the front end that the image html is not having the hash part

Create an article with both intro and full images
Create a menu item articles category blog and check the front end that the image html is not having the hash part
Also check on the article view (clicking on the title) the image html is not having the hash part 

Create a new newsfeed with the link: `https://www.joomla.org/announcements.feed?height=660&keepthis=632&tb_iframe=true&type=rss`, make sure to set both the first and second images
Create a menu item newsfeed category and check the front end that the image html is not having the hash part
Also check the the actual feed has the first and second image with the hash

Broken

Fixed

Not these particular changes but the process of cleaning the URL from the media field should definitely be documented


components/com_contact/tmpl/category/default_items.php
```diff
@@ -99,7 +99,14 @@ class="inputbox" onchange="document.adminForm.submit();"
 							<a href="<?php echo Route::_(RouteHelper::getContactRoute($item->slug, $item->catid, $item->language)); ?>">
 								<?php if ($this->params->get('show_image_heading')) : ?>
 									<?php if ($item->image) : ?>
-										<?php echo HTMLHelper::_('image', $item->image, '', array('class' => 'contact-thumbnail img-thumbnail')); ?>
+										<?php $imgAttribs = ['class' => 'contact-thumbnail img-thumbnail']; ?>
+										<?php $img = HTMLHelper::cleanImageURL($item->image); ?>
+										<?php if ($img->width > 0 && $img->height > 0) : ?>
+											<?php $imgAttribs['width'] = $img->width; ?>
+											<?php $imgAttribs['height'] = $img->width; ?>
```
Quy:
```suggestion
											<?php $imgAttribs['height'] = $img->height; ?>
```
Please repeat.

components/com_contact/tmpl/contact/default.php
```diff
@@ -97,11 +97,18 @@
...
+					<?php if ($img->width > 0 && $img->height > 0) : ?>
+						<?php $imgAttribs['width'] = $img->width; ?>
+						<?php $imgAttribs['height'] = $img->height; ?>
+						<?php $imgAttribs['loading'] = 'lazy'; ?>
+					<?php endif; ?>
 					<?php echo HTMLHelper::_(
 						'image',
-						$this->item->image,
+						htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
 						htmlspecialchars($this->item->name,  ENT_QUOTES, 'UTF-8'),
```
Quy:
```suggestion
						htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
						htmlspecialchars($this->item->name, ENT_QUOTES, 'UTF-8'),
```

components/com_content/tmpl/category/blog.php
```diff
@@ -54,8 +54,19 @@
...
+				<?php $imgAttribs = []; ?>
+				<?php $img = HTMLHelper::cleanImageURL($this->category->getParams()->get('image')); ?>
+				<?php if ($img->width > 0 && $img->height > 0) : ?>
+					<?php $imgAttribs['width'] = $img->width; ?>
+					<?php $imgAttribs['height'] = $img->height; ?>
+					<?php $imgAttribs['loading'] = 'lazy'; ?>
+				<?php endif; ?>
+				<?php echo HTMLHelper::_(
+					'image',
+					htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
```
Quy:
```suggestion
					htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
```
Please repeat.

components/com_newsfeeds/tmpl/category/default.php
```diff
@@ -34,7 +34,19 @@
...
+				<?php $imgAttribs = []; ?>
+				<?php $img = HTMLHelper::cleanImageURL($this->category->getParams()->get('image')); ?>
+				<?php if ($img->width > 0 && $img->height > 0) : ?>
+					<?php $imgAttribs['width'] = $img->width; ?>
+					<?php $imgAttribs['height'] = $img->height; ?>
+					<?php $imgAttribs['loading'] = 'lazy'; ?>
+				<?php endif; ?>
+				<?php echo HTMLHelper::_(
+					'image',
+					htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
```
PhilETaylor:
```suggestion
					htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +61,47 @@
...
+				<?php endif; ?>
 				<?php if ($images->image_second_caption) : ?>
-					<?php echo 'class="caption" title="' . htmlspecialchars($images->image_second_caption) . '"'; ?>
+					<?php $imgAttribs['class'] = 'caption'; ?>
+					<?php $imgAttribs['title'] = htmlspecialchars($images->image_second_caption, ENT_COMPAT, 'UTF-8'); ?>
 				<?php endif; ?>
-				src="<?php echo htmlspecialchars($images->image_second, ENT_COMPAT, 'UTF-8'); ?>" <?php echo $alt; ?>>
+				<?php echo HTMLHelper::_(
+					'image',
+					htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
```
PhilETaylor:
```suggestion
					htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -98,9 +115,21 @@
...
+				<?php $imgAttribs = []; ?>
+				<?php $img = HTMLHelper::cleanImageURL($this->rssDoc->image->uri); ?>
+				<?php if ($img->width > 0 && $img->height > 0) : ?>
+					<?php $imgAttribs['width'] = $img->width; ?>
+					<?php $imgAttribs['height'] = $img->height; ?>
+					<?php $imgAttribs['loading'] = 'lazy'; ?>
+				<?php endif; ?>
+				<?php echo HTMLHelper::_(
+					'image',
+					htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
```
PhilETaylor:
```suggestion
					htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
```

modules/mod_articles_news/tmpl/_item.php
```diff
@@ -9,25 +9,38 @@
...
+		<?php $imgAttribs = []; ?>
+		<?php $img = HTMLHelper::cleanImageURL($item->imageSrc); ?>
+		<?php if ($img->width > 0 && $img->height > 0) : ?>
+			<?php $imgAttribs['width'] = $img->width; ?>
+			<?php $imgAttribs['height'] = $img->height; ?>
+			<?php $imgAttribs['loading'] = 'lazy'; ?>
+		<?php endif; ?>
+		<?php echo HTMLHelper::_(
+			'image',
+			htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
```
PhilETaylor:
```suggestion
			htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +61,47 @@
...
+				<?php endif; ?>
 				<?php if ($images->image_first_caption) : ?>
-					<?php echo 'class="caption" title="' . htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8') . '"'; ?>
+					<?php $imgAttribs['class'] = 'caption'; ?>
+					<?php $imgAttribs['title'] = htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8'); ?>
 				<?php endif; ?>
-				src="<?php echo htmlspecialchars($images->image_first, ENT_COMPAT, 'UTF-8'); ?>" <?php echo $alt; ?>>
+				<?php echo HTMLHelper::_(
+					'image',
+					htmlspecialchars($img->url,  ENT_QUOTES, 'UTF-8'),
```
Quy:
```suggestion
					htmlspecialchars($img->url, ENT_QUOTES, 'UTF-8'),
```

modules/mod_menu/tmpl/default_component.php
```diff
@@ -45,12 +45,14 @@
 {
 	if ($item->menu_image_css)
 	{
+		$img = HTMLHelper::cleanImageURL($item->menu_image);
 		$image_attributes['class'] = $item->menu_image_css;
-		$linktype = HTMLHelper::_('image', $item->menu_image, $item->title, $image_attributes);
+		$linktype = HTMLHelper::_('image', $img->url, $item->title, $image_attributes);
```
bembelimen: Why do you clean the image twice? As HTMLHelper does the same, too: https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/HTML/HTMLHelper.php#L750-L753

You should also (if you change this place anyway) use ```HTMLHelper::_('image', ...)```

dgrammatiko: Good point, basically I didn't know that the image helper cleans the url. There are still places that we need to do it before in order to add lazyloading. I Will review the code again tomorrow

dgrammatiko: I reverted the changes in the mod_menu as that is already covered. The rest of the layouts still use the `cleanImageURL` to enable lazy loading (if the image has width and height defined)

joomdonation:
Hmm. We are repeating the same code again and again. Is there a way to add a new helper like lazyImage to handle this?

dgrammatiko:
> Is there a way to add a new helper like lazyImage to handle this?

My approach would be to add a new JLayout and not another HTMLHelper thing. The first is very approachable to any front ender the second requires writing a plugin to override it. Basically, the HTMLHelper (s) should be used only for logic/processing and **never** for spitting out HTML.


My 2c

joomdonation:
That should work, too. If we do not have to repeat that same code again and again, and have a way for third party developers to re-use the code, we should be fine.

richard67:
@dgrammatiko Reading the above discussion it's not 100% clear for me: Is this PR ready for testing now, or do we have to wait for some changes?

dgrammatiko:
> Reading the above discussion it's not 100% clear for me: Is this PR ready for testing now, or do we have to wait for some changes?

The changes are valid but also it's valid that at the current state we have a lot of duplicate code. I will add a new layout for rendering images and use it core wide

richard67:
> I will add a new layout for rendering images and use it core wide

@dgrammatiko In this PR here or in another, future one?

dgrammatiko:
@richard67 in this PR
@joomdonation is this any better?

components/com_contact/tmpl/contact/default.php
```diff
@@ -62,7 +63,7 @@
 		<?php $contactLink = RouteHelper::getCategoryRoute($this->item->catid, $this->item->language); ?>
 		<h3>
 			<span class="contact-category"><a href="<?php echo $contactLink; ?>">
-				<?php echo $this->escape($this->item->category_title); ?></a>
+					<?php echo $this->escape($this->item->category_title); ?></a>
```
richard67:
```suggestion
				<?php echo $this->escape($this->item->category_title); ?></a>
```
Revert wrong change on indentation.

components/com_contact/tmpl/contact/default.php
```diff
@@ -121,7 +130,7 @@
 				<?php if ($tparams->get('allow_vcard')) : ?>
 					<?php echo Text::_('COM_CONTACT_DOWNLOAD_INFORMATION_AS'); ?>
 					<a href="<?php echo Route::_('index.php?option=com_contact&amp;view=contact&amp;id=' . $this->item->id . '&amp;format=vcf'); ?>">
-					<?php echo Text::_('COM_CONTACT_VCARD'); ?></a>
+						<?php echo Text::_('COM_CONTACT_VCARD'); ?></a>
```
richard67:
```suggestion
					<?php echo Text::_('COM_CONTACT_VCARD'); ?></a>
```
Revert wrong change on indentation.

joomdonation:
@dgrammatiko I still have a question about the code. To render an image, we are using this code

```php
echo LayoutHelper::render(
		'joomla.html.image',
		[
			'image' => [
				'src' => $item->imageSrc,
				'alt' => $item->imageAlt,
				'attributes' => []
			],
		]
	);
```

I wonder why we need to use 'image' key here? Shouldn't something like below is faster (to write):

```php
echo LayoutHelper::render(
		'joomla.html.image',
		[		
			'src' => $item->imageSrc,
			'alt' => $item->imageAlt,
			'attributes' => []			
		]
	);
```

And for additional attributes like `alt`, `attributes`, can we not pass it if the data is not available?

Not really an issue with code. Just some questions from me.

components/com_contact/tmpl/category/default_items.php
```diff
@@ -99,7 +100,15 @@ class="inputbox" onchange="document.adminForm.submit();"
...
 								<?php if ($this->params->get('show_image_heading')) : ?>
 									<?php if ($item->image) : ?>
-										<?php echo HTMLHelper::_('image', $item->image, '', array('class' => 'contact-thumbnail img-thumbnail')); ?>
+										<?php echo LayoutHelper::render(
+											'joomla.html.image',
+											[
+												'image' => [
+													'src' => $item->image,
+													'alt' => '',
+													'attributes' => ['class' => 'contact-thumbnail img-thumbnail']
```
PhilETaylor:
```suggestion
													'attributes' => ['class' => 'contact-thumbnail img-thumbnail'],
```

components/com_contact/tmpl/contact/default.php
```diff
@@ -97,11 +102,15 @@
...
-						$this->item->image,
-						htmlspecialchars($this->item->name,  ENT_QUOTES, 'UTF-8'),
-						array('itemprop' => 'image')
+					<?php echo LayoutHelper::render(
+						'joomla.html.image',
+						[
+							'image' => [
+								'src' => $this->item->image,
+								'alt' => $this->item->name,
+								'attributes' => ['itemprop' => 'image']
```
PhilETaylor:
```suggestion
								'attributes' => ['itemprop' => 'image'],
```

components/com_content/tmpl/category/blog.php
```diff
@@ -54,8 +55,16 @@
...
 			<?php if ($this->params->get('show_description_image') && $this->category->getParams()->get('image')) : ?>
-				<?php $alt = empty($this->category->getParams()->get('image_alt')) && empty($this->category->getParams()->get('image_alt_empty')) ? '' : 'alt="' . htmlspecialchars($this->category->getParams()->get('image_alt'), ENT_COMPAT, 'UTF-8') . '"'; ?>
-				<img src="<?php echo $this->category->getParams()->get('image'); ?>" <?php echo $alt; ?>>
+				<?php echo LayoutHelper::render(
+					'joomla.html.image',
+					[
+						'image' => [
+							'src' => $this->category->getParams()->get('image'),
+							'alt' => empty($this->category->getParams()->get('image_alt')) && empty($this->category->getParams()->get('image_alt_empty')) ? '' : $this->category->getParams()->get('image_alt'),
+							'attributes' => []
```
PhilETaylor:
```suggestion
							'attributes' => [],
```

components/com_newsfeeds/tmpl/category/default.php
```diff
@@ -34,7 +35,16 @@
...
 		<div class="com-newsfeeds-category__description category-desc">
 			<?php if ($this->params->get('show_description_image') && $this->category->getParams()->get('image')) : ?>
-				<img src="<?php echo $this->category->getParams()->get('image'); ?>">
+				<?php echo LayoutHelper::render(
+					'joomla.html.image',
+					[
+						'image' => [
+							'src' => $this->category->getParams()->get('image'),
+							'alt' => empty($this->category->getParams()->get('image_alt')) && empty($this->category->getParams()->get('image_alt_empty')) ? '' : $this->category->getParams()->get('image_alt'),
+							'attributes' => []
```
PhilETaylor:
```suggestion
							'attributes' => [],
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,43 @@
...
+					<?php $imgAttribs['title'] = htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8'); ?>
 				<?php endif; ?>
-				src="<?php echo htmlspecialchars($images->image_first, ENT_COMPAT, 'UTF-8'); ?>" <?php echo $alt; ?>>
+				<?php echo LayoutHelper::render(
+					'joomla.html.image',
+					[
+						'image' => [
+							'src' => $images->image_first,
+							'alt' => empty($images->image_first_alt) && empty($images->image_first_alt_empty) ? '' : $images->image_first_alt,
+							'attributes' => $imgAttribs
```
PhilETaylor:
```suggestion
							'attributes' => $imgAttribs,
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,43 @@
...
+					<?php $imgAttribs['title'] = htmlspecialchars($images->image_second_caption, ENT_COMPAT, 'UTF-8'); ?>
 				<?php endif; ?>
-				src="<?php echo htmlspecialchars($images->image_second, ENT_COMPAT, 'UTF-8'); ?>" <?php echo $alt; ?>>
+				<?php echo LayoutHelper::render(
+					'joomla.html.image',
+					[
+						'image' => [
+							'src' => $images->image_second,
+							'alt' => empty($images->image_second_alt) && empty($images->image_second_alt_empty) ? '' : $images->image_second_alt,
+							'attributes' => $imgAttribs
```
PhilETaylor:
```suggestion
							'attributes' => $imgAttribs,
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -98,9 +112,18 @@
...
+		<?php if ($this->rssDoc->image && $this->params->get('show_feed_image')) : ?>
 			<div class="com-newsfeeds-newsfeed__feed-image">
-				<img src="<?php echo $this->rssDoc->image->uri; ?>" alt="<?php echo $this->rssDoc->image->title; ?>" />
+				<?php echo LayoutHelper::render(
+					'joomla.html.image',
+					[
+						'image' => [
+							'src' => $this->rssDoc->image->uri,
+							'alt' => $this->rssDoc->image->title,
+							'attributes' => []
```
PhilETaylor:
```suggestion
							'attributes' => [],
```

modules/mod_articles_news/tmpl/_item.php
```diff
@@ -9,25 +9,35 @@
...
 <?php if ($params->get('img_intro_full') !== 'none' && !empty($item->imageSrc)) : ?>
 	<figure class="newsflash-image">
-		<img src="<?php echo $item->imageSrc; ?>" alt="<?php echo $item->imageAlt; ?>">
+		<?php echo LayoutHelper::render(
+			'joomla.html.image',
+			[
+				'image' => [
+					'src' => $item->imageSrc,
+					'alt' => $item->imageAlt,
+					'attributes' => []
```
PhilETaylor:
```suggestion
					'attributes' => [],
```

dgrammatiko:
> I wonder why we need to use 'image' key here? Shouldn't something like below is faster (to write)

Yes having just the attributes for the image element would be easier, the reason I used the key ‘image’ has more to do with future expanding the layout to render a figure with caption and/or picture  with source nested elements (responsive images). Having the values under a named key will make that easier, instead of naming the keys with funny names . That said jlayouts support sublayouts so in the future this could be a sub layout of figure?
That’s a valid question and I have no clue if this is better than using a plain array and deal with the figure/picture using sublayouts

joomdonation:
Without seeing the actual code, I'm unsure if it is good or bad. I think my questions about optional attributes still valid, with or without using image key

> And for additional attributes like alt, attributes, can we not pass it if the data is not available?

Could anyone else looks at the current implementation before we doing real test? @Fedik ?

Fedik:
I agree with @joomdonation https://github.com/joomla/joomla-cms/pull/35780#issuecomment-955207860
Make it simple :wink: 
In the future no one will remember why it made like that.

For `picture` we can check for `source` variable or have another layout, kind of `joomla.html.picture`, when it will be need. 
(In my code I use similar `blabla.html.image` and `blabla.html.picture`)

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('_JEXEC') or die;
+
+use Joomla\CMS\HTML\HTMLHelper;
+
+extract($displayData);
```
Fedik: please add doc block with list of used/possible variables,
as we usually doing for other layouts

dgrammatiko:
@joomdonation @Fedik is this version any better?

Fedik:
yes and no ;) 

Please move `htmlspecialchars` stuff inside image layout, 
there you can use `$this->escape($blabla);`

This way we can be sure that the values is always escaped,

dgrammatiko:
@Fedik are the changes ok now?

layouts/joomla/html/image.php
```diff
@@ -18,12 +18,26 @@
...
+$hasAlt = !empty($displayData['alt']);
 
 if ($img->width > 0 && $img->height > 0) {
   $displayData['width'] = $img->width;
   $displayData['height'] = $img->height;
   $displayData['loading'] = 'lazy';
 }
 
-echo '<img ' . ArrayHelper::toString($displayData) . '>';
+$src = $hasSrc ? htmlspecialchars($displayData['src'], ENT_QUOTES, 'UTF-8') . ' ' : '';
```
Fedik: `$this->escape($displayData['src'])` :wink: 
it was made specially for it

dgrammatiko: done

Fedik:
Looks good to me :)

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,43 @@
...
+$src = $hasSrc ? $this->escape($displayData['src']) . ' ' : '';
+$alt = $hasAlt ? $this->escape($displayData['alt']) . ' ': '';
+
+if ($hasSrc)
+{
+  unset($displayData['src']);
+}
+
+if ($hasAlt) {
+  unset($displayData['alt']);
```
Quy:
```suggestion
$alt = $hasAlt ? $this->escape($displayData['alt']) . ' ' : '';

if ($hasSrc)
{
  unset($displayData['src']);
}

if ($hasAlt)
{
  unset($displayData['alt']);
```

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,44 @@
...
+{
+  unset($displayData['src']);
+}
+
+if ($hasAlt)
+{
+  unset($displayData['alt']);
+}
+
+echo '<img ' . $src . $alt . ArrayHelper::toString($displayData) . '>';
```
joomdonation: It is not working yet. I applied the PR and image won't be displayed. There are few issues with current code:

- We are not using clean url for the src of the image
- We are missing src="". The $src is being echoed directly

I think the code block could be changed to:

```php
$img = HTMLHelper::_('cleanImageURL', $displayData['src']);

$displayData['src'] = $this->escape($img->url);

if (!empty($displayData['alt']))
{
	$displayData['alt'] = $this->escape($displayData['alt']);
}

if ($img->width > 0 && $img->height > 0)
{
	$displayData['width']   = $img->width;
	$displayData['height']  = $img->height;
	$displayData['loading'] = 'lazy';
}

echo '<img '  . ArrayHelper::toString($displayData) . '>';
```

Also, do we need to escape the src of the image? We did not do that before.

dgrammatiko: You're right, my bad. About the src being or not escaped: since this is a core layout that anyone could use I would keep it

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,31 @@
...
 			<div class="com-newsfeeds-newsfeed__first-image img-intro-<?php echo htmlspecialchars($imgfloat, ENT_COMPAT, 'UTF-8'); ?>">
-				<img
+				<?php $imgAttribs = [
+					'src' => $images->image_first,
+					'alt' => empty($images->image_first_alt) && empty($images->image_first_alt_empty) ? '' : $images-> image_first_alt,
+				]; ?>
 				<?php if ($images->image_first_caption) : ?>
-					<?php echo 'class="caption" title="' . htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8') . '"'; ?>
+					<?php $imgAttribs['class'] = 'caption'; ?>
+					<?php $imgAttribs['title'] = htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8'); ?>
```
joomdonation: Sorry, still have question: Why do we use `htmlspecialchars` here for **title** instead of handle it in the layout like **alt** attribute?

dgrammatiko: Although `title` is an acceptable attribute for images https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes and https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes I have no clue if it's a11y correct. We could loop through all the attributes and escape them but it would be tiny slower and also most of the attributes should be fine without escaping (class, id, etc)

joomdonation: No, I think we should not loop through all the attributes and escape them. I just think we should move htmlspecialchars for title to the layout (same as alt) to make the code uses layout simpler.

@Fedik Any feedback about this? If all is OK, I will do final test and report test result.

Fedik: yeah, just make it as you did for `alt`

dgrammatiko: Kinda disagree here. The problem is that instead of title the parent layout should used figure and figcaption elements. The title is not a11y correct in this case and the platform already has the solution with the figure element. So, adding title to the img layout will be the wrong solution to the problem…

Fedik: then just remove it here from attributes, no one will cry for it ;) 
maybe some old legacy thing

dgrammatiko: Yes that’s a legacy thing

dgrammatiko: @Fedik @joomdonation so I swapped the title for a figure/figcaption, let me know if there's something more here

modules/mod_articles_news/tmpl/_item.php
```diff
@@ -9,25 +9,32 @@
 
 defined('_JEXEC') or die;
 
+use Joomla\CMS\HTML\HTMLHelper;
 use Joomla\CMS\Layout\LayoutHelper;
```
Quy:
```suggestion

use Joomla\CMS\Layout\LayoutHelper;
```

drmenzelit:
I have tested this item :white_check_mark: successfully on 4c54bb57be304192fb8fb7c5165b702484ab7cdf<br /><br />Tested with category blog.

Image url before patch:

imagebefore

Image url after patch:

imageafter

In other views (module newsflash or contacts) I don't see changes in the image url.

<hr/><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35780">issues.joomla.org/tracker/joomla-cms/35780</a>.</sub>

Quy:
Please see the `alt` attribute.

```
<img src="/Joomla_4.0.6-dev+pr.35780-Development-Full_Package/images/sampledata/cassiopeia/nasa2-400.jpg" itemprop="image" alt="alt=&quot;&quot;">
```

dgrammatiko:
@Quy it should fine now, thanks

layouts/joomla/content/intro_image.php
```diff
@@ -22,34 +22,27 @@
...
 
-$imgclass  = empty($images->float_intro) ? $params->get('float_intro') : $images->float_intro;
-$extraAttr = '';
-$img       = HTMLHelper::cleanImageURL($images->image_intro);
-$alt       = empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : 'alt="' . htmlspecialchars($images->image_intro_alt, ENT_COMPAT, 'UTF-8') . '"';
+$imgclass   = empty($images->float_intro) ? $params->get('float_intro') : $images->float_intro;
+$img        = HTMLHelper::cleanImageURL($images->image_intro);
+$layoutAttr = [
+	'src'      => $img->url,
+	'alt'      => empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : htmlspecialchars($images->image_intro_alt, ENT_COMPAT, 'UTF-8'),
```
Quy: Remove `htmlspecialchars`?

layouts/joomla/content/full_image.php
```diff
@@ -20,24 +20,23 @@
...
-$imgclass  = empty($images->float_fulltext) ? $params->get('float_fulltext') : $images->float_fulltext;
-$extraAttr = '';
-$img       = HTMLHelper::cleanImageURL($images->image_fulltext);
-$alt       = empty($images->image_fulltext_alt) && empty($images->image_fulltext_alt_empty) ? '' : 'alt="' . htmlspecialchars($images->image_fulltext_alt, ENT_COMPAT, 'UTF-8') . '"';
+$imgclass   = empty($images->float_fulltext) ? $params->get('float_fulltext') : $images->float_fulltext;
+$img        = HTMLHelper::cleanImageURL($images->image_fulltext);
+$layoutAttr = [
+	'src'      => $img->url,
+	'itemprop' => 'image',
+	'alt'      => empty($images->image_fulltext_alt) && empty($images->image_fulltext_alt_empty) ? '' : htmlspecialchars($images->image_fulltext_alt, ENT_COMPAT, 'UTF-8'),
```
Quy: Remove `htmlspecialchars`?

dgrammatiko: Done

layouts/joomla/content/intro_image.php
```diff
@@ -22,34 +22,27 @@
...
+	'alt'      => empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : $images->image_intro_alt,
+];
 
 // Set lazyloading only for images which have width and height attributes
 if ((isset($img->attributes['width']) && (int) $img->attributes['width'] > 0)
 && (isset($img->attributes['height']) && (int) $img->attributes['height'] > 0))
 {
-	$extraAttr = ArrayHelper::toString($img->attributes) . ' loading="lazy"';
+	$layoutAttr = array_merge($layoutAttr, $img->attributes, ['loading' => 'lazy']);
 }
```
Quy: Should this be removed as it will be done in image layout?

uglyeoin:
@dgrammatiko please can you help me with testing instructions?  I'm happy to test but I don't understand how to.  The previous issue you linked to mentions we can see images before hand.  What should I be looking for?  I don't want to accept a test by accident with incorrect testing.

layouts/joomla/content/intro_image.php
dgrammatiko: Yeah, that was double processing for no reason. Fixed to do all that work in the image jLayout

components/com_content/tmpl/category/blog.php
```diff
@@ -41,8 +42,8 @@
 
 	<?php if ($this->params->get('show_category_title', 1)) : ?>
 	<<?php echo $htag; ?>>
-		<?php echo $this->category->title; ?>
-	</<?php echo $htag; ?>>
+ 		<?php echo $this->category->title; ?>
+ 	</<?php echo $htag; ?>>
```
brianteeman:
```suggestion
		<?php echo $this->category->title; ?>
	</<?php echo $htag; ?>>
```

removes extra leading space

components/com_content/tmpl/category/blog.php
```diff
@@ -100,14 +105,14 @@
...
-			</div>
-		<?php endforeach; ?>
+ 			<div class="com-content-category-blog__item blog-item"
+ 				itemprop="blogPost" itemscope itemtype="https://schema.org/BlogPosting">
+ 					<?php
+ 					$this->item = & $item;
+ 					echo $this->loadTemplate('item');
+ 					?>
+ 			</div>
+ 		<?php endforeach; ?>
```
brianteeman:
```suggestion
			<div class="com-content-category-blog__item blog-item"
				itemprop="blogPost" itemscope itemtype="https://schema.org/BlogPosting">
					<?php
					$this->item = & $item;
					echo $this->loadTemplate('item');
					?>
			</div>
		<?php endforeach; ?>
```
remove extra leading space

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -42,7 +43,7 @@
 		<?php if ($this->params->get('display_num')) : ?>
 		<h1 class="<?php echo $direction; ?>">
 			<?php echo $this->escape($this->params->get('page_heading')); ?>
-		</h1>
+ 		</h1>
```
brianteeman:
```suggestion
		</h1>
```

remove extra leading space

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+/**
+ * Layout variables
+ * -----------------
+ * @var   array  $displayData  Array with all the given attributes for the image element.
+ *                             Eg: src,class,alt,width,height,loading,decoding,style,data-*
```
brianteeman:
```suggestion
 *                             Eg: src, class ,alt, width, height, loading, decoding, style, data-*
```

readability

modules/mod_articles_news/tmpl/_item.php
```diff
@@ -17,17 +17,23 @@
...
-	<?php endif; ?>
-	</<?php echo $item_heading; ?>>
-<?php endif; ?>
+ 			<?php echo $item->title; ?>
+ 		</a>
+ 	<?php else : ?>
+ 		<?php echo $item->title; ?>
+ 	<?php endif; ?>
+ 	</<?php echo $item_heading; ?>>
+ <?php endif; ?>
```
brianteeman:
```suggestion
			<?php echo $item->title; ?>
		</a>
	<?php else : ?>
		<?php echo $item->title; ?>
	<?php endif; ?>
	</<?php echo $item_heading; ?>>
<?php endif; ?>
```

remove extra leading space

dgrammatiko:
@uglyeoin I've updated the testing instructions, hopefully it's easy to follow

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+/**
+ * Layout variables
+ * -----------------
+ * @var   array  $displayData  Array with all the given attributes for the image element.
+ *                             Eg: src, class ,alt, width, height, loading, decoding, style, data-*
```
Quy:
```suggestion
 *                             Eg: src, class, alt, width, height, loading, decoding, style, data-*
```

components/com_contact/tmpl/category/default_items.php
```diff
@@ -99,7 +100,13 @@ class="inputbox" onchange="document.adminForm.submit();"
...
 								<?php if ($this->params->get('show_image_heading')) : ?>
 									<?php if ($item->image) : ?>
-										<?php echo HTMLHelper::_('image', $item->image, '', array('class' => 'contact-thumbnail img-thumbnail')); ?>
+										<?php echo LayoutHelper::render(
+											'joomla.html.image',
+											[
+												'src'   => $item->image,
+												'alt'   => '',
+												'class' => 'contact-thumbnail img-thumbnail',
+											]); ?>
```
Quy:
```suggestion
											]
										); ?>
```

layouts/joomla/content/intro_image.php
```diff
@@ -22,36 +21,21 @@
...
-// Set lazyloading only for images which have width and height attributes
-if ((isset($img->attributes['width']) && (int) $img->attributes['width'] > 0)
-&& (isset($img->attributes['height']) && (int) $img->attributes['height'] > 0))
-{
-	$extraAttr = ArrayHelper::toString($img->attributes) . ' loading="lazy"';
-}
+$imgclass   = empty($images->float_intro) ? $params->get('float_intro') : $images->float_intro;
+$layoutAttr = [
+	'src'      => $images->image_intro,
+	'alt'      => empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : $images->image_intro_alt,
```
Quy:
```suggestion
	'src' => $images->image_intro,
	'alt' => empty($images->image_intro_alt) && empty($images->image_intro_alt_empty) ? '' : $images->image_intro_alt,
```

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,42 @@
...
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+/**
+ * Layout variables
+ * -----------------
+ * @var   array  $displayData  Array with all the given attributes for the image element.
+ *                             Eg: src, class, alt, width, height, loading, decoding, style, data-*
+ *                             Note: only the alt attribute is escaped by default!
```
Quy: Add `src`?

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,37 @@
...
+				<figure>
+					<?php echo LayoutHelper::render(
+						'joomla.html.image',
+						[
+							'src' => $images->image_first,
+							'alt' => empty($images->image_first_alt) && empty($images->image_first_alt_empty) ? '' : $images->image_first_alt,
+						]
+					); ?>
+					<?php if ($images->image_first_caption) : ?>
+						<figcaption><?php echo htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8'); ?></figcaption>
```
Quy:
```suggestion
						<figcaption class="caption"><?php echo htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8'); ?></figcaption>
```

layouts/joomla/html/image.php
dgrammatiko: It'salready there: `Eg: src, class, alt, width, height, loading, decoding, style, data-*`

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,37 @@
...
+				<figure>
+					<?php echo LayoutHelper::render(
+						'joomla.html.image',
+						[
+							'src' => $images->image_second,
+							'alt' => empty($images->image_second_alt) && empty($images->image_second_alt_empty) ? '' : $images->image_second_alt,
+						]
+					); ?>
+					<?php if ($images->image_second_caption) : ?>
+						<figcaption><?php echo htmlspecialchars($images->image_second_caption, ENT_COMPAT, 'UTF-8'); ?></figcaption>
```
Quy:
```suggestion
						<figcaption class="caption"><?php echo htmlspecialchars($images->image_second_caption, ENT_COMPAT, 'UTF-8'); ?></figcaption>
```

layouts/joomla/html/image.php
Quy: Regarding being escaped.

dgrammatiko: yeah, good catch

dgrammatiko: Done, thanks

Quy:
For Article Category Blog, no width, height, and loading attributes.

Before PR:
```
<figure class="left item-image">
	<img src="/joomla-cms-4.0-dev/images/sampledata/fruitshop/tamarind.jpg"
			 			 itemprop="image"
			width="375" height="488" loading="lazy"	/>
			<figcaption class="caption">Full Caption</figcaption>
```

After PR:
```
<figure class="left item-image">
	<img src="/joomla-cms-4.0-dev/images/sampledata/fruitshop/tamarind.jpg" itemprop="image" alt="">			<figcaption class="caption">Full Caption</figcaption>
	</figure>
```

dgrammatiko:
> For Article Category Blog, no width, height, and loading attributes.

That was a bug, I thought I knew the API I introduced many months ago but I was wrong. My bad. It should be fine now

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,42 @@
...
+
+if (!empty($displayData['alt']))
+{
+	$displayData['alt'] = $this->escape($displayData['alt']);
+}
+
+if (isset($img->attributes['width']) && isset($img->attributes['height']) && $img->attributes['width'] > 0 && $img->attributes['height'] > 0)
+{
+	$displayData['width']  = $img->attributes['width'];
+	$displayData['height'] = $img->attributes['width'];
```
Quy:
```suggestion
	$displayData['height'] = $img->attributes['height'];
```

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,42 @@
...
+$img = HTMLHelper::_('cleanImageURL', $displayData['src']);
+
+$displayData['src'] = $this->escape($img->url);
+
+if (!empty($displayData['alt']))
+{
+	$displayData['alt'] = $this->escape($displayData['alt']);
+}
+
+if (isset($img->attributes['width']) && isset($img->attributes['height']) && $img->attributes['width'] > 0 && $img->attributes['height'] > 0)
```
Quy: Can it be simplified to this since `cleanImageURL` checks for `width` and `height` values?
`if (!empty($img->attributes['width']) && !empty($img->attributes['height']))`

Quy: or use `isset`?

dgrammatiko: I had it like that but I switched to use isset and the bigger than 0 because svg files report width and height as 0. But I guess the isset is not needed…

Quy: In `cleanImageURL`, `width` and `height` are set only when `> 0`, therefore, no need to check again so `isset` should suffice.

dgrammatiko: Ok done

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,37 @@
...
+				<figure>
+					<?php echo LayoutHelper::render(
+						'joomla.html.image',
+						[
+							'src' => $images->image_first,
+							'alt' => empty($images->image_first_alt) && empty($images->image_first_alt_empty) ? '' : $images->image_first_alt,
+						]
+					); ?>
+					<?php if ($images->image_first_caption) : ?>
+						<figcaption class="caption"><?php echo htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8'); ?></figcaption>
```
Quy: Last one I promise then testing. 🤣 Change to `$this->escape` to be consistent with the same changes in this PR? If yes, there are several instances of this in this file.

dgrammatiko: This is not a JLayout, I used `$this->escape` only inside JLayouts. But I can change it...

Quy: Let's leave it as is. It can be decided/done in a separate PR.

dgrammatiko: I mean I can change it but then I will have to change all the instances in the file (check couple lines before) 😉

Quy: Let's change them for consistency. Thanks!

layouts/joomla/html/image.php
```diff
@@ -0,0 +1,42 @@
...
+$img = HTMLHelper::_('cleanImageURL', $displayData['src']);
+
+$displayData['src'] = $this->escape($img->url);
+
+if (!empty($displayData['alt']))
+{
+	$displayData['alt'] = $this->escape($displayData['alt']);
+}
+
+if (isset($img->attributes) && isset($img->attributes['width']) && isset($img->attributes['height']))
```
Quy:
```suggestion
if (isset($img->attributes['width']) && isset($img->attributes['height']))
```
I think this is unnecessary to check `isset($img->attributes)`.

components/com_newsfeeds/tmpl/newsfeed/default.php
dgrammatiko: Ok done

Quy:
For Contact Categories, on the front end:

``

layouts/joomla/content/category_default.php
```diff
@@ -73,8 +73,13 @@
 		<?php if ($beforeDisplayContent || $afterDisplayContent || $params->get('show_description', 1) || $params->def('show_description_image', 1)) : ?>
 			<div class="category-desc">
 				<?php if ($params->get('show_description_image') && $category->getParams()->get('image')) : ?>
-					<?php $alt = empty($category->getParams()->get('image_alt')) && empty($category->getParams()->get('image_alt_empty')) ? '' : 'alt="' . htmlspecialchars($category->getParams()->get('image_alt'), ENT_COMPAT, 'UTF-8') . '"'; ?>
-					<img src="<?php echo $category->getParams()->get('image'); ?>" <?php echo $alt; ?>>
+					<?php echo LayoutHelper::render(
+						'joomla.html.image',
+						[
+							'src'   => $category->getParams()->get('image'),
+							'alt'   => empty($category->getParams()->get('image_alt')) && empty($category->getParams()->get('image_alt_empty')) ? '' : 'alt="' . $this->escape($category->getParams()->get('image_alt')) . '"',
```
Quy:
```suggestion
							'src' => $category->getParams()->get('image'),
							'alt' => empty($category->getParams()->get('image_alt')) && empty($category->getParams()->get('image_alt_empty')) ? '' : $category->getParams()->get('image_alt'),
```

components/com_newsfeeds/tmpl/newsfeed/default.php
```diff
@@ -61,30 +62,37 @@
...
-					<?php echo 'class="caption" title="' . htmlspecialchars($images->image_first_caption, ENT_COMPAT, 'UTF-8') . '"'; ?>
-				<?php endif; ?>
-				src="<?php echo htmlspecialchars($images->image_first, ENT_COMPAT, 'UTF-8'); ?>" <?php echo $alt; ?>>
+			<div class="com-newsfeeds-newsfeed__first-image img-intro-<?php echo $this->escape($imgfloat); ?>">
+				<figure>
+					<?php echo LayoutHelper::render(
+						'joomla.html.image',
+						[
+							'src' => $images->image_first,
+							'alt' => empty($images->image_first_alt) && empty($images->image_first_alt_empty) ? '' : $images->image_first_alt,
```
Quy:
```suggestion
							'alt' => empty($images->image_first_alt) && empty($images->image_first_alt_empty) ? false : $images->image_first_alt,
```

Quy:
Please resolve conflicts.

Quy:
I have tested this item :white_check_mark: successfully on 4c54bb57be304192fb8fb7c5165b702484ab7cdf<br /><br />Thank you!!!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35780">issues.joomla.org/tracker/joomla-cms/35780</a>.</sub>

dgrammatiko:
@Quy thanks for all the fixes and the tests

alikon:
I have tested this item :white_check_mark: successfully on 4c54bb57be304192fb8fb7c5165b702484ab7cdf<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35780">issues.joomla.org/tracker/joomla-cms/35780</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35780">issues.joomla.org/tracker/joomla-cms/35780</a>.</sub>

dgrammatiko:
@wilsonge

laoneo:
I have tested this item :white_check_mark: successfully on 4c54bb57be304192fb8fb7c5165b702484ab7cdf<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35780">issues.joomla.org/tracker/joomla-cms/35780</a>.</sub>

laoneo:
By the way, tested also with DPMedia and the hash is removed.

dgrammatiko:
> By the way, tested also with DPMedia and the hash is removed.

That was the whole point of this PR: provide one simple JLayout to fix the hash problem universally. Now we need to backport it to 3.x and ask every DEV to render all their images using this layout...

wilsonge:
Thanks!

PR #35781, opened 2021-10-07
[3.10] Remove expired  - DST Root CA X3

wojsmol:
Pull Request for Issue https://github.com/joomla/joomla-cms/pull/33518#issuecomment-937977203

### Summary of Changes
Remove old expired  DST Root CA X3 Let's Encrypt root certificate - see [Old Let’s Encrypt Root Certificate Expiration and OpenSSL 1.0.2](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/)
### Actual result BEFORE applying this Pull Request
HTTPS connections to servers with latests Let's Encrypt certificates from server with openssl version 1.0.2 are NOT working

### Expected result AFTER applying this Pull Request
HTTPS connections to servers with latests Let's Encrypt certificates from server with openssl version 1.0.2 are working

### Documentation Changes Required
N/A

PR #35782, opened 2021-10-07
[4.0] Update composer/ca-bundle to remove expiret - DST Root CA X3

wojsmol:
Pull Request for Issue https://github.com/joomla/joomla-cms/pull/33518#issuecomment-937977203.

### Summary of Changes
Similar to https://github.com/joomla/joomla-cms/pull/35781 in 4.0-dev and up we are using composer/ca-bundle directly

Remove old expired DST Root CA X3 Let's Encrypt root certificate - see [Old Let’s Encrypt Root Certificate Expiration and OpenSSL 1.0.2](https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/)

### Testing Instructions
Make HTTPS request to server using latest Let's Encrypt certificate from server using openssl 1.0.2

HTTPS connections to servers with latests Let's Encrypt certificates from server with openssl version 1.0.2 are NOT working.

HTTPS connections to servers with latests Let's Encrypt certificates from server with openssl version 1.0.2 are working.

N/A

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -187,16 +187,16 @@
         },
         {
             "name": "composer/ca-bundle",
-            "version": "1.2.10",
+            "version": "1.2.11",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "9fdb22c2e97a614657716178093cd1da90a64aa8"
+                "reference": "0b072d51c5a9c6f3412f7ea3ab043d6603cb2582"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/9fdb22c2e97a614657716178093cd1da90a64aa8",
-                "reference": "9fdb22c2e97a614657716178093cd1da90a64aa8",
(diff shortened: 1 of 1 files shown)


composer.lock
```diff
@@ -9739,5 +9739,5 @@
     "platform-overrides": {
         "php": "7.2.5"
     },
-    "plugin-api-version": "2.1.0"
+    "plugin-api-version": "2.0.0"
```
richard67, line 9742: Where does this downgrade come from? Maybe an unrelated change? If so, it should be reverted.

wojsmol: This is auto generated by `Composer version 2.0.13 2021-04-27 13:11:08` - maybe I have older composer version then last person updating dependencies.

wilsonge:
Closing this in favour of the full file update in https://github.com/joomla/joomla-cms/pull/35785/files

wilsonge:
Woops this was the 4.0 PR - my bad - reopening

composer.lock
wilsonge: Probably - worth running a `composer self-update` probably - but doesn't block this PR

wilsonge:
Merging on review. Thanks!

PR #35794, opened 2021-10-11
[4.0] Fix package installer not setting new changelogurl

Ruud68:
Pull Request for Issue # .

### Summary of Changes
When you have a package installed that didn't have a changelogurl set in the manifest file and you upgrade that package with a version that does have a changelogurl set in the manifest, this changelogurl is not stored in that database.
currently changelogurl is only set (and used) on newly installed packages.
This PR fixes this bug and allows for already installed package have a changelogurl when the update package has one

### Testing Instructions
install a package that does NOT have a changelogurl in the manifest file
upgrade that package with a version that DOES have a changelog URL in the manifest file

Changelogurl not added to the table and thus not used in extension manager

Changelogurl correctly added to the extensions table and used in extension manager


joomdonation:
I have tested this item :white_check_mark: successfully on bfd9b0c9f1c6e7ff7a7e99c370f8dce7e22dc7a5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35794">issues.joomla.org/tracker/joomla-cms/35794</a>.</sub>

joomdonation:
@Ruud68 Could you please check also `FileAdapter`? I don't have experience with this feature but looks like we have same issue with that adapter. Would be great if it could be fixed in this PR, too.

Ruud68:
Hi @joomdonation  yes fileadapter and also templateadapter have same bug.
No experience with these two and now way to test them, that is why i left them out of the package adapter change.

will update pr to also (hopefully) fix them, but as said that will be a PR that i cannot test.
BTW I have very low hopes for this PR as probably nobody is able to test it, let alone test it for package, template and file updates :(

joomdonation:
@Ruud68 Thanks. Actually, `TemplateAdapter`  is OK because it is handled in this line of code https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Installer/Adapter/TemplateAdapter.php#L592 . So please revert the change which you made to that adapter. We can even remove this line of code https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Installer/Adapter/TemplateAdapter.php#L583

Ruud68:
update to template adapter changed

libraries/src/Installer/Adapter/FileAdapter.php
```diff
@@ -396,6 +396,9 @@ protected function storeExtension()
 
 			// Update name
 			$this->extension->name = $this->name;
+
```
PhilETaylor: Remove whitespace

```suggestion

```

Ruud68:
Are we good to go for the next release? if not what needs to be done to get this PR in there?

HLeithner:

laoneo:
I have tested this item :white_check_mark: successfully on 054ae224b8aca74997f3a4c6a87e4a6af36082dc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35794">issues.joomla.org/tracker/joomla-cms/35794</a>.</sub>

Issue #35342, opened 2021-08-24
Important but easy improvements to the code of the articles


impressionestudio:
I write below a list of simple but useful improvements to the HTML code of an article in order to expand its usability to the needs of today (CSS grid).

I created a demo article having a title, image, main text, article information and two custom fields before and after the main text.

![Demo article](https://demo.impression-estudio.gr/_issue-tracker/demo-article.jpg "Demo article")

### 1. BETTER PLACEMENT OF THE  CUSTOM FIELDS

The custom fields: 
- must be placed outside the ul/li elements, 
- must be placed on the same level as the main content, the title and the image 
- and must be placed inside divs. 

(The custom fields may contain large texts, images, links, video players, photo galleries etc. So it is not good to be placed in ul/li elements.)

So the code for a custom field should change from: 
```
<ul class="fields-container">
	<li class="field-entry city">
		<span class="field-label ">City: </span>
		<span class="field-value city">Athens</span>
	</li>
</ul>
```

to: 

```
<div class="field-entry city">
	<span class="field-label ">City: </span>
	<span class="field-value city">Athens</span>
</div>
```

The code for the article part with the above changes is:

![Custom fields code after the improvements](https://demo.impression-estudio.gr/_issue-tracker/custom-fields-after.jpg "Custom fields code after the improvement")

### 2. BETTER NAMES TO CLASSES

There is some inconsistency in the names of the CSS classes as it concerns the code of an article. 
The class names:
- should be smaller
- should have a common logic

Specifically: 
1. The class “page-header” should be renamed to “title” since it represents the title field of an article.
2. The class “item-image” should be only “image”. No need to use “item” since we don’t use it anywhere else in the rest class names.
3. The class “com-content-article__body” is big. Maybe only “body” or “content” since it refers to the “Content” field of an article is better.
4. The class “field-entry” should be renamed to “field” or “custom-field”.
5. The class “article-info” should be renamed to “info”. The “article-“ part is redundant/obvious.
6. The class “article-info-term” could change to “info-title”.
7. The class “category-name” witch is part of the article information should be simply “category” since for example the class “published” is not “published-date”.

### 3. ARTICLE INFO

The article info code should be organized better:
1. The labels of the article information like “Written by:” or “Category:” or “Published:” should be place in their own span elements 
2. They should have a specific class name “info-label” similar to “field-label” in the custom fields.
3. The article info values should have a specific class name “info-value” similar to “field-value” in the custom fields.

So the code should change from:

[...]

Issue #15279, opened 2017-04-13
[RFC] [4.0] Standard for naming of classes?


ciar4n:
We could do with a standard for naming of class attributes. Maybe one already exists? 

Understandably naming is currently quite random..

https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_modules/views/module/tmpl/edit_assignment.php#L62
https://github.com/joomla/joomla-cms/blob/staging/modules/mod_languages/tmpl/default.php#L19 https://github.com/joomla/joomla-cms/blob/staging/modules/mod_syndicate/tmpl/default.php#L13 https://github.com/joomla/joomla-cms/blob/staging/modules/mod_tags_popular/tmpl/default.php#L14

Some mainstream standards exist.. most notable is probably BEM... https://en.bem.info/

brianteeman:
I was at a presentation at a php conference on this exact topic last weekend

You can see the slidedeck here https://drive.google.com/file/d/0B7FGMFd-aDPWQlpMaDdna2ExbWM/view

I hope the video will be online in a few days

ciar4n:
There is likely to be some principles of PHP naming that could be used here. I guess HTML classes would have less variables which would allow us more specific naming.

BEM (Block, Element, Modifier) is effective. However I am not sure how we feel about using double underscores and hyphens.

The BEM principle..

```
/* Block component */
.btn {}

/* Element that depends upon the block */ 
.btn__login {}

/* Modifier that changes the style of the block */
.btn_orange {} 
.btn_big {}
```

One of its benefits is it allows everything to have a unique class which makes CSS specificity very flat and low (less nesting). You can also tell how a class can be reused without conflict and its function by the class name (Eg. btn--orange - a modifier only used within any btn).

The flip side is some think it makes for ugly looking code. I'm not suggesting we go to the extreme of BEM but maybe we can borrow from it.

How we name the 'blocks' is something to consider separately. For the highest level blocks we could use the Joomla architecture...

`ExtensionType-Extension-View` (Eg. `com-content-archive`).

I am not sure how sustainable this is throughout all the markup, anyone with a broader knowledge of Joomla architecture could probably answer this. My knowledge here has proven to be limited at best (",)

yvesh:
@ciar4n hmph, not a sure if i am a fan of double hyphens or underscores.. Is it really making things easier? 

[...]

PR #35796, opened 2021-10-11
Important but easy improvements to the code of the article view (1)

impressionestudio:
Mainly they are changes to the names of the classes using the BEM nomenclature.

### DIFF
--- a/components/com_content/tmpl/article/default.php
+++ b/components/com_content/tmpl/article/default.php
@@ -33,10 +33,10 @@ $currentDate       = Factory::getDate()->format('Y-m-d H:i:s');
 $isNotPublishedYet = $this->item->publish_up > $currentDate;
 $isExpired         = !is_null($this->item->publish_down) && $this->item->publish_down < $currentDate;
 ?>
-<div class="com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
+<div class="article <?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
 	<meta itemprop="inLanguage" content="<?php echo ($this->item->language === '*') ? Factory::getApplication()->get('language') : $this->item->language; ?>">
 	<?php if ($this->params->get('show_page_heading')) : ?>
-	<div class="page-header">
+	<div class="article__title">
 		<h1> <?php echo $this->escape($this->params->get('page_heading')); ?> </h1>
 	</div>
 	<?php endif;
@@ -50,7 +50,7 @@ $isExpired         = !is_null($this->item->publish_down) && $this->item->publish
 	|| $params->get('show_hits') || $params->get('show_category') || $params->get('show_parent_category') || $params->get('show_author') || $assocParam; ?>
 
 	<?php if ($params->get('show_title')) : ?>
-	<div class="page-header">
+	<div class="article__title">
 		<<?php echo $htag; ?> itemprop="headline">
 			<?php echo $this->escape($this->item->title); ?>
 		</<?php echo $htag; ?>>
@@ -98,7 +98,7 @@ $isExpired         = !is_null($this->item->publish_down) && $this->item->publish
 	<?php if (isset ($this->item->toc)) :
 		echo $this->item->toc;
 	endif; ?>
-	<div itemprop="articleBody" class="com-content-article__body">
+	<div itemprop="articleBody" class="article__body">
 		<?php echo $this->item->text; ?>
 	</div>
 


brianteeman:
I dont see how changing the class names achieves what you highlight in the original issue

simbus82:
@impressionestudio, can you "merge" all PR in only one so we can test it? And please rename the classes that you have changed: no underscore please, and not change the actual logic of css classes.

Fedik:

impressionestudio:
> I dont see how changing the class names achieves what you highlight in the original issue

Because it is the first time I use GitHub, I thought to do it in two parts. After this first part with the class name changes, I will do the code changes, which is the second and last part of this issue.

impressionestudio:
> Changing class names does not makes sense, you only make hard time for existing templates, and for an overrides

The class names are very inconsistent. So many things are changed in Joomla 4 and the problem is to change the class names? Now, along with all the other changes, is time to make these class names better because we use them all the time. Otherwise, we will end up to have Joomla 5 using some strange class names from Joomla 1.0. :)

Fedik:
> After this first part with the class name changes, I will do the code changes

If it all for one issue, then better do it in one run. So people can see what is going one, and this way it is more easy to make decision.

> We have changed so many things in Joomla 4 and the problem is to change the class names?

I have looked your changes and to me it is a change for the sake of change (maybe only 1-2 places I would half-agree).
Joomla 4 already out, and now you suggest to ruin someone's template, because you do not like class names? ;) 

Also please check my comment in original issue https://github.com/joomla/joomla-cms/issues/35342#issuecomment-940405118

impressionestudio:
> Joomla 4 already out, and now you suggest to ruin someone's template, because you do not like class names? ;)

I believe that it is still early, or not too late for something so important as I see it. I had mentioned this issue in forums (outside GitHub) long before Joomla 4 was released, and I had a few hopes that the developers of Joomla 4 will take in mind these changes. After Joomla 4 was released, I saw that nothing changed and thought to drop it. But later, as I was checking Joomla 4 and seeing so many improvements, I wasn't feeling right to accept this limitation. Because, despite the class name changes that are only for consistency reasons and having pretty CSS code, the code changes will expand the possibilities of the template design.

simbus82:
> > Changing class names does not makes sense, you only make hard time for existing templates, and for an overrides
> 
> The class names are very inconsistent. So many things are changed in Joomla 4 and the problem is to change the class names? Now, along with all the other changes, is time to make these class names better because we use them all the time. Otherwise, we will end up to have Joomla 5 using some strange class names from Joomla 1.0. :)

I can understand this is your first time using github, but if you had used that "naming" for CSS classes in my agency, I would have had to fire you. (obviously I'm exaggerating).
**CSS classes are not written like this, it is wrong, it does not matter Joomla, it is really wrong at the level of basic web development.**
There is no point in hard coding the "cascade" of the class itself. CSS is made to handle the cascade!
image

impressionestudio:
> > > Changing class names does not makes sense, you only make hard time for existing templates, and for an overrides
> > 
> > 
> > The class names are very inconsistent. So many things are changed in Joomla 4 and the problem is to change the class names? Now, along with all the other changes, is time to make these class names better because we use them all the time. Otherwise, we will end up to have Joomla 5 using some strange class names from Joomla 1.0. :)
> 
> I can understand this is your first time using github, but if you had used that "naming" for CSS classes in my agency, I would have had to fire you. (obviously I'm exaggerating). **CSS classes are not written like this, it is wrong, it does not matter Joomla, it is really wrong at the level of basic web development.** There is no point in hard coding the "cascade" of the class itself. CSS is made to handle the cascade! image

The new class names that I suggest are not my original choice. I had to use them because some guys instructed me to follow the BEM naming rules. 

My original class names were single/simple words like "info" and "published" and a few times something more complex like "info-published". If you check my original class names in https://github.com/joomla/joomla-cms/issues/35342 you will see how simple they were.

@Fedik suggested to follow the RFC #15279, so I will read the new naming rules and possibly the class names will be simpler and shorter.

brianteeman:
@simbus82 Glad I dont work for you

simbus82:
> @simbus82 Glad I dont work for you

You can eat quietly. Changing the logic of the classes in this way, without considering everything around it, is not a good practice.
Then we can talk about BEM and the fact that I don't like it for a frontend in a CMS like Joomla. Too verbose and unreadable, i prefer the deep nesting. But I'm old for these things now.

> The new class names that I suggest are not my original choice. I had to use them because some guys instructed me to follow the BEM naming rules.
> 
> My original class names were single/simple words like "info" and "published" and a few times something more complex like "info-published". If you check my original class names in #35342 you will see how simple they were.

@impressionestudio I really liked the august proposal (you can see my thumb up) about the html structure, only not liked the change in the logic of the CSS naming that took place now.

impressionestudio:
@simbus82 I also don't like BEM, but I am willing to do some compromises in order others to accept the general idea.

PR #35804, opened 2021-10-11
Important but easy improvements to the code of the article view (8)

impressionestudio:
Mainly they are changes to the names of the classes.


Fedik:

PR #35807, opened 2021-10-11
Important but easy improvements to the code of the article view (11)

impressionestudio:
Mainly they are changes to the names of the classes.


laoneo:
This classes were added as part of an effort to introduce BEM class structure. You can read more about it here https://github.com/joomla/joomla-cms/issues/15279. I strongly encourage the maintainers to not accept this kind of pr's as it would end up again in a mess with classes.
And it can to some degree also treated as a BC break.

Fedik:

Issue #35814, opened 2021-10-12
[4.0] Invisible recaptcha label


brianteeman:
### Steps to reproduce the issue
Configure and enable the invisible recaptcha plugin

### Expected result
Contact form is displayed with the invisible recaptcha and the label is **not** displayed

### Actual result
Label is displayed
![image](https://user-images.githubusercontent.com/1296369/137026942-4c1febfd-1502-4bc2-9c8b-535682980a1c.png)

### System information (as much as possible)
cassiopeia 
current main branch

C-Lodder:
> libraries/src/Form/Field/CaptchaField.php

Adding `protected $hiddenLabel = true;` to the field doesn't work, for some odd reason.
Only way I managed to hide it, was by adding an empty `protected function getLabel() {}`, but not sure this is the correct approach.

joomdonation:
Change https://github.com/joomla/joomla-cms/blob/4.0-dev/plugins/captcha/recaptcha_invisible/recaptcha_invisible.php#L172 to

```php
$element['hiddenLabel'] = 'true';
```

should solve the issue.

C-Lodder:
PR: https://github.com/joomla/joomla-cms/pull/35828

Ended up removing that method and adding a class variable instead, cause I don't agree with a mix of strings and booleans

joomdonation:
I made PR https://github.com/joomla/joomla-cms/pull/35834 to fix this issue. Please check and give feedback. Closing this issue for now and will re-open later if needed.

PR #35828, opened 2021-10-15
[4.0] Hide invisible recapctah label

C-Lodder:
Pull Request for Issue #35814

### Summary of Changes

Hide the invisible captcha label

### Testing Instructions

1. Setup and enable invisible recaptcha
2. Go to the registration form

The `Captcha` label was visible

The `Captcha` label is hidden


joomdonation:
I haven't tested but I think this will make the label invisible for the other captchas, too (while it should only hide label for invisible recaptcha).

C-Lodder:
Ah yes, right you are.

PR #35840, opened 2021-10-16
Changes to class names of intro_image.php

impressionestudio:
Pull Request for Issue joomla#35342.

### Detailed explanation of the changes:

1. On row 37, the class "item-image" is replaced by "article__image--full" since it represents the full image field of an article. 
The old class name remained for compatibility reasons, separated by the character | to be indicated as deprecated.

2. On row 54, the class "caption" is replaced by "article__caption--full" since it represents the full image caption of an article. 
The old class name remained for compatibility reasons, separated by the character | to be indicated as deprecated.


brianteeman:
please please do not create individual pull requests for every file

PR #19038, opened 2017-12-11
[4.1] com_media modifications for front-end use

stutteringp0et:
This small change allows media fields to be used in user profiles, where a user is not authorized for core.create in com_users - they can be authorized in com_media.  In conjunction with a filesystem plugin, a user can be granted access to com_media a limited area of the filesystem.  This change should not impact anything other than front-end media modals.

### Summary of Changes

allow core.create for com_media to open com_media

### Testing Instructions

Create a user custom field of type media as a regular user (registered group).  Editing your profile, the media field will have no button to open the media manager because the user has no access.

Apply this change, and allow "Regisered" user group "create" permissions.  The user is now given the "Select"

### Expected result
media custom fields usable

### Actual result
after change, users can use media fields

PR #35863, opened 2021-10-19
[4.1] Allow User Media Field to be accessible to Registered users (Reworks #19038)

particthistle:
### New PR to redo #19038
Thanks to @stutteringp0et for original PR.

#### What it does
This small change allows media fields to be used in user profiles, where a user is not authorized for core.create in com_users - they can be authorized in com_media. In conjunction with a filesystem plugin, a user can be granted access to com_media a limited area of the filesystem. This change should not impact anything other than front-end media modals.

Currently the change:
- requires administrator to enable the permissions for Media component: Registered needs to have Create permissions set to Allow
image
- provides the user with the ability to upload an image, but also select ANY media available.

A future addition / feature enhancement would be to limit user to only viewing their own media and limit uploading to just a folder for that user.

#### Summary of Changes
In `plugins/fields/media/media.php` if com_media has core.create permissions, enable the select button.
In `components/com_media/src/Dispatcher/Dispatcher.php` add a check to see if the user has permissions for core.create on com_media

#### Testing Instructions
- Create a user custom field of type media with Registered access level permissions.
- Editing your profile, the media field will have ~no button to open the media manager~ a Select button, but will not load the Media Manager when clicked because the user has no access.

- Apply this change
- Go to options for Media
- Change the permissions for the "Registered" user group, setting "create" permissions to "Allow".
- The user is now given ~the "Select"~ access to the Media Manager when they go to select an image.

#### Before Patch applied
Media manager will not load due to permissions issue
image

#### Expected result
Media manager loads due to Registered Users having core.create permissions set on com_media
image


richard67:
@particthistle PR #19038 was for the 4.1-dev branch and this one here is for 4.0-dev. Is that by purpose?

plugins/fields/media/media.php
```diff
@@ -39,7 +39,12 @@ public function onCustomFieldsPrepareDom($field, DOMElement $parent, Form $form)
 		}
 
 		$fieldNode->setAttribute('type', 'accessiblemedia');
-
+		
+		if (JFactory::getApplication()->getIdentity()->authorise('core.create', 'com_media'))
```
brianteeman:
```suggestion
		if (Factory::getApplication()->getIdentity()->authorise('core.create', 'com_media'))
```

and add at the top of the file
`use Joomla\CMS\Factory;`

bembelimen:
> 
> 
> @particthistle PR #19038 was for the 4.1-dev branch and this one here is for 4.0-dev. Is that by purpose?

Thx, done

particthistle:
4.0 instead of 4.1 was an oversight.

JFactory changed to Factory

particthistle:
With #35886 doing a similar fix by removing code, which of the two PR is correct?

I've not got a moment to see if #35886 allows the User Media Field to be accessible to Registered Users, but will try and check later unless someone else has a moment to test it.

richard67:
@particthistle As far as I could see, the other PR just removed obsolete conditions, and only this here allows the registered user access to the media field, so both PRs are independent. They could have been made together in one, but keeping it separate makes testing safer because only the relevant change is tested, here the new access and in the other PR that removing the obsolete stuff doesn’t break anything.

particthistle:
I've incorporated the changes from #35886 in advance of that being RTC.

The code removed in #35886 still leaves the problem of allowing those with core.create permissions from uploading new media. This PR resolves that based on what was initially added in #19038

Shubhamverma2796:
I have tested this item :white_check_mark: successfully on 20706fc908602d90e1695b36d775f30572df5434<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35863">issues.joomla.org/tracker/joomla-cms/35863</a>.</sub>

pritam825:
I have tested this item :white_check_mark: successfully on 20706fc908602d90e1695b36d775f30572df5434<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35863">issues.joomla.org/tracker/joomla-cms/35863</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35863">issues.joomla.org/tracker/joomla-cms/35863</a>.</sub>

bembelimen:
Thx

Issue #35604, opened 2021-09-19
[4.x] npm updates


brianteeman:
There needs to be a procedure to make sure that all npm packages are up to date before a new joomla release is made.

For example there have been three tinymce releases since the release of Joomla 4.0.0

dgrammatiko:
https://www.whitesourcesoftware.com/free-developer-tools/renovate/ does an excellent job, it's free and will work nicely in the project's context

brianteeman:
Thanks for the suggestion. I dont care what is used as long as something is and we dont end up making a release that contains outdated software

dgrammatiko:
This can only be tackled by people with the right permissions in this repo. I'm not one of them so I can only propose something that I have used in countless projects so far and I know that it works...
Still waiting for the same thing to be enabled in the custom elements repo (although that repo has dependencies only for the tools)

brianteeman:
Closed as no one who can do anything about this shows the remotes sign of any interest

PR #35889, opened 2021-10-24
[4.0] Update tinymce and codemirror

brianteeman:
### TinyMCE 5.10
https://www.tiny.cloud/docs/changelog/#5100-2021-10-11

Resolve three outdated settings that are deprecated and will be removed in 6.0
Show deprecated notices when debug is enabled

### Codemirror 5.63.3
https://github.com/codemirror/CodeMirror/releases/tag/5.63.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "roboto-fontface": "^0.10.0",
     "short-and-sweet": "^1.0.4",
     "skipto": "^4.1.0",
-    "tinymce": "^5.8.2",
+    "tinymce": "^5.10.0",
     "vue": "^3.1.2",
     "vuex": "^4.0.2",
     "vuex-persistedstate": "^4.0.0-beta.3"
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "roboto-fontface": "^0.10.0",
     "short-and-sweet": "^1.0.4",
     "skipto": "^4.1.0",
-    "tinymce": "^5.10.0",
+    "tinymce": "^5.10.2",
     "vue": "^3.1.2",
     "vuex": "^4.0.2",
     "vuex-persistedstate": "^4.0.0-beta.3"
--- a/plugins/editors/tinymce/tinymce.php
+++ b/plugins/editors/tinymce/tinymce.php
@@ -575,6 +575,7 @@ class PlgEditorTinymce extends CMSPlugin
 		$scriptOptions   = array_merge(
 			$scriptOptions,
 			array(
+				'deprecation_warnings' => true,
 				'suffix'   => '.min',
 				'baseURL'  => Uri::root(true) . '/media/vendor/tinymce',
 				'directionality' => $text_direction,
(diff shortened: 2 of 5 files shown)


dgrammatiko:
@brianteeman could you add `'deprecation_warnings' => false` in the default options array as per their docs: https://www.tiny.cloud/docs/release-notes/6.0-upcoming-changes/

I haven't tested but I suspect that there will be a lot of deprecations logged in the browser's console...

EDIT: This `'deprecation_warnings' => JDEBUG? true : false`  might be a better fit now that I'm thinking about it

brianteeman:
not a fan of adding that

dgrammatiko:
> not a fan of adding that

That's fine, by the way, you need to also update the package.json for the TinyMCE as `5.9.2` and `5.10.0` are different patch versions snd NPM will not automatically fetch the latter

brianteeman:
yes it will - at least it did for me

brianteeman:
Sorry I misunderstood you. It was late, I thought you were asking me to ADD the deprecated notices, not to remove them. Updating this PR shortly

brianteeman:
There were only three deprecation notices which I have resolved

gecko_spellchecker
force_p_newlines
inline_styles

plugins/editors/tinymce/tinymce.php
```diff
@@ -575,6 +575,7 @@ public function onDisplay(
 		$scriptOptions   = array_merge(
 			$scriptOptions,
 			array(
+				'deprecation_warnings' => JDEBUG? true : false,
```
richard67:
```suggestion
				'deprecation_warnings' => JDEBUG ? true : false,
```

brianteeman:
took so long it has now been updated to 5.10.1

brianteeman:
Took so long its now been updated to 5.10.2

package.json
```diff
@@ -56,7 +56,7 @@
     "roboto-fontface": "^0.10.0",
     "short-and-sweet": "^1.0.4",
     "skipto": "^4.1.0",
-    "tinymce": "^5.8.2",
+    "tinymce": "^5.10.2",
```
richard67, line 59: Shouldn't line 43 be changed for codemirror, too? https://github.com/joomla/joomla-cms/blob/4047820995e4c4b5ef388d34067dde6d081a8758/package.json#L43

Currently it shows `"codemirror": "^5.63.1",`, but I think it should be `"codemirror": "^5.63.3",`

brianteeman: thats what the lock file is for

brianteeman: but I repeat we should be automating this #35604

richard67: Then in case of this PR the lock file and the package.json don't fit together. If you update the tinymce entry in the package.json you should do it the same way for codemirror, or not?

dgrammatiko: @richard67 it's fine as is (although confusing)
Screenshot 2021-11-21 at 20 43 02

richard67: Fine and confusing is not really the same if it could be made better, i.e. less confusing.

dgrammatiko: Reads as: `technically the definition is fine` still `for some this is confusing (I'm always wondering if ~ or ^ is for patch/minor)`

richard67: It is simply not consistent to do it here in this PR for TinyMCE but not for Codemirror in the packages.json. But I stop to complain.

dgrammatiko: As Brian said earlier someone with the right permissions needs to create a simple json file for renovate and add the project to their bot. Then this won't be a problem again and also the project will get timely PRs for any of the dependencies (node/PHP). Eg: https://github.com/dgrammatiko/site/blob/main/renovate.json

wilsonge:
Thanks!

brianteeman:
finally

thanks

package.json
wojsmol: There is also a dependabot witch is owned by github and supports also github actions updates.

PR #35905, opened 2021-10-26
[J4] Fix case of letters for filename localization plugins

korenevskiy:
If the plugin name has letters in uppercase, then the localization file for the plugin is not clear in which case of letters to save the file name.
In the plugins component, i.e. when the user enters the plugin settings, a localization file is used in the case of letters corresponding to the case of the plugin name with uppercase letters.
But at the same time, the site itself uses only lowercase letters for the localization file name by default.

I am adding the ability of the plugin component to connect both localization files with uppercase (corresponding to the plugin name) and lowercase. This will fix the error of letter case names for plugins and maintain backward compatibility.

Localization in Plugin class use in low case  letters.
https://github.com/joomla/joomla-cms/blob/7b867392ac7764568b800cd0d6f58c768c8340b7/libraries/src/Plugin/CMSPlugin.php#L164
But in component plugins use not low case letters.


brianteeman:
Can you give an example please of a mixed case plugin. All the core plugins are lowercase.

Wouldn';t the same _problem_ exist for modules and components??

korenevskiy:
> Can you give an example please of a mixed case plugin. All the core plugins are lowercase.
> 
> Wouldn';t the same _problem_ exist for modules and components??

It is difficult for me to give an example of a well-known example.
But I created a plugin with a mixed name many years ago.
https://github.com/korenevskiy/plg_jshopping_PlaceBilet
In order not to suffer from the case of letters. I used a crutch.
`JFactory::getApplication()->getLanguage()->load('plg_jshopping_PlaceBilet', __DIR__); `
I had to write such a crutch in different places of the code before.
- In the plugin class code.
- In the code of custom fields for the XML configuration of the plugin.

But now I was trying to figure out why localization doesn't work without crutches.
As seen here
https://github.com/joomla/joomla-cms/blob/7b867392ac7764568b800cd0d6f58c768c8340b7/libraries/src/Plugin/CMSPlugin.php#L164
But now I was trying to figure out why localization doesn't work without crutches.
As you can see, lowercase is used here.
Despite the fact that there is no conversion of plugin names to lowercase letters in the plugins component.

For modules and components, it's hard for me to tell about this problem. This will require a study of the issue.
But the current fix solves this problem for plugins completely.

Maybe I could have done better.
But I need to understand whether it is necessary to convert all plugin names to lowercase at all, or is it better for CMS to support both registers everywhere?

korenevskiy:
This problem was found by me in J 3.10

brianteeman:
My gut feeling is that you shouldnt have used mixed case filenames in your plugin but i could be wrong . Also on a case sensitive operating system the file **name.ext** is a different to **NAME.ext** so this PR is introducing a new problem by using strlower - i think.

I will leave it to more experienced developers

korenevskiy:
You are right that on linux hosting, names in different registers create problems.
I agree with you that I should not have used the plugin name in a different case.
But the problem here is not me. I solve my problem with a crutch.
I am writing here about the CMS problem in cases when someone else uses a different case for names.
This means that either we need to remove the lowercase conversion of characters everywhere, or we need to apply the lowercase conversion of file names everywhere.
Since if lowercase is a prerequisite for names, then there is no point in bringing the plugin base class code to lowercase and it is necessary to remove `$extension = strtolower($extension);`
Or allow lowercase and take it into account everywhere.
In every sense, there should be one rule.

korenevskiy:
### **This is a very important question, as it will allow you to create names in Hungarian Notation**

HLeithner:

bembelimen:
Hello @korenevskiy ,

thank you for your proposal.

Plugin names must always be lowercase and words are separated by underscores. If you want to have anything different displayed, you can use the language string for the plugin name.

PR #35921, opened 2021-10-28
Allow to set a value for SubformField

HLeithner:
Subform Field doesn't allow to set a `value` because the supplied `value` is automatically converted to `string` which is the wrong variable type for Subform Field. The Field needs an `array`.

### Summary of Changes
This PR automatically converts a `json` encoded string into an `array` and allows to supply an `array` directly

### Testing Instructions
1. Create a custom field with a subform
2. Test if the field still works in backend and frontend

Testing the new functionally is not so easy. 
Editing the com_content default layout could be the simplest way.

You need to get the subform field from the form and set a value.

For a successful test I would expect that the simple test is enough.

* Simple test works
* Extended Test doesn't work

* Simple test works
* Extended Test works

### DIFF
--- a/libraries/src/Form/Field/SubformField.php
+++ b/libraries/src/Form/Field/SubformField.php
@@ -181,7 +181,7 @@ class SubformField extends FormField
 				}
 
 				$this->value = $value !== null ? (array) $value : null;
-			break;
+				break;
 
 			default:
 				parent::__set($name, $value);
--- a/libraries/src/Form/Field/SubformField.php
+++ b/libraries/src/Form/Field/SubformField.php
@@ -15,8 +15,6 @@ use Joomla\CMS\Form\Form;
 use Joomla\CMS\Form\FormField;
 use Joomla\Registry\Registry;
 
-use function json_decode;
-
 /**
  * The Field to load the form inside current form
  *
@@ -182,7 +180,7 @@ class SubformField extends FormField
 					$value = json_decode($value, true);
 				}
 
-				$this->value = (array) $value;
+				$this->value = $value;
 			break;
 
 			default:
--- a/libraries/src/Form/Field/SubformField.php
+++ b/libraries/src/Form/Field/SubformField.php
@@ -181,6 +181,7 @@ class SubformField extends FormField
 				}
 
 				$this->value = $value !== null ? (array) $value : null;
+
 				break;
 
 			default:
--- a/libraries/src/Form/Field/SubformField.php
+++ b/libraries/src/Form/Field/SubformField.php
@@ -15,6 +15,8 @@ use Joomla\CMS\Form\Form;
 use Joomla\CMS\Form\FormField;
 use Joomla\Registry\Registry;
 
+use function json_decode;
+
 /**
  * The Field to load the form inside current form
  *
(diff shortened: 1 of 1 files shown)


Fedik:
we have that here also:
https://github.com/joomla/joomla-cms/blob/ff20420f33f1fe19c94a33112488702647b75306/libraries/src/Form/Field/SubformField.php#L216-L220

maybe also change there to avoid duplication?

libraries/src/Form/Field/SubformField.php
```diff
@@ -173,6 +175,16 @@ public function __set($name, $value)
...
 
+			case 'value':
+				// We allow a json encoded string or an array
+				if (is_string($value))
+				{
+					$value = json_decode($value, true);
+				}
+
+				$this->value = (array) $value;
+			break;
```
Quy:
```suggestion
				break;
```

HLeithner:
> we have that here also:
> 
> https://github.com/joomla/joomla-cms/blob/ff20420f33f1fe19c94a33112488702647b75306/libraries/src/Form/Field/SubformField.php#L216-L220
> 
> maybe also change there to avoid duplication?

Not sure what you mean, isn't this the same code?
Maybe the cast to array should be removed by me and the use statement....

Fedik:
hm,  ignore, my comment, sorry for confusing

> Maybe the cast to array should be removed

yea, I think better to remove, or at least check if there any value,
because `$this->value = (array) null;` will produce `[null]`

HLeithner:
@Fedik should be better now

libraries/src/Form/Field/SubformField.php
Quy: See lines 140, 152, 174....

libraries/src/Form/Field/SubformField.php
```diff
@@ -173,6 +173,16 @@ public function __set($name, $value)
...
 
+			case 'value':
+				// We allow a json encoded string or an array
+				if (is_string($value))
+				{
+					$value = json_decode($value, true);
+				}
+
+				$this->value = $value !== null ? (array) $value : null;
+				break;
```
HLeithner, line 185:
```suggestion

				break;
```

Fedik:
I have tested this item :white_check_mark: successfully on c1978427c5d3bebec274f70add83ca6651b23fc3<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35921">issues.joomla.org/tracker/joomla-cms/35921</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on c1978427c5d3bebec274f70add83ca6651b23fc3<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35921">issues.joomla.org/tracker/joomla-cms/35921</a>.</sub>

RickR2H:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35921">issues.joomla.org/tracker/joomla-cms/35921</a>.</sub>

wilsonge:
Thanks!

PR #35911, opened 2021-10-26
[4.1] Enabled to focus in media files

khu5h1:
Pull Request for Issue #35229 .

### Summary of Changes
I have enabled focus on media files by changing their `tabindex` from -1 to 0.

### Testing Instructions
Check out issue #35229 for testing and reproducing the PR.

### Actual result BEFORE applying this Pull Request
Unable to focus on media files while using the tab button.

### Expected result AFTER applying this Pull Request
Enabled to focus in media files using the tab button.

PR #35931, opened 2021-10-29
[4.1] Implement access levels for media manager folder

bembelimen:
### Summary of Changes
This PR allows to define access levels for the different folder entries, adds a level to the folder selection and fixes a bug with subfolders in the local folder management in the media manager.

grafik

grafik

(2nd folder not visible, because user is logged in)

### Testing Instructions
1. Apply PR
2. Go to the "FileSystem - Local" plugin
3. Set some folders + permissions
4. Create users to check the permissions (by e.g. opening the media select for intro images in an article)

No access level option available

Only folders with the correct access level are shown


plugins/filesystem/local/src/Adapter/LocalAdapter.php
```diff
@@ -16,8 +16,8 @@
 use Joomla\CMS\Filesystem\File;
 use Joomla\CMS\Filesystem\Folder;
 use Joomla\CMS\Filesystem\Path;
-use Joomla\CMS\Helper\MediaHelper;
 use Joomla\CMS\HTML\HTMLHelper;
+use Joomla\CMS\Helper\MediaHelper;
```
richard67: Why changed the order? It was right before, "Helper" before "HTML".

bembelimen: Thx, have no idea why my IDE ordered this wrong (and still does...)

brianteeman: its to do with the capitalisation. Some sort orders work AaBbCcDd and others work ABCDabcd

brianteeman:
The subfolders thing used to be there but it was changed - I forget the reason why but I suspect @PhilETaylor will know.

bembelimen:
The problem with the subfolder was/is, that the "recursive" option breaks the dropdown (as it is currently wrong implemented, that's why I added: https://github.com/joomla/joomla-cms/pull/35931/files#diff-b20a03c7bebd0a7b7ae0b5de9486009ec060a89a12542421578ae22251e5d707R162-R167 ) and I'm absolutely not happy with the current solution (but it's the best Joomla! can do beside offering a normal text field).
Also the local adapter had a bug when loading subfolders.

brianteeman:
ACL for this is great but are you sure Access Levels is the correct way? Thats for controlling who can view something. I would have expected it to be using groups which is more fine grained. 

Access Level only decided who can view the folder
Real ACL with groups would also enable control over who can add,delete media in a folder 

### Example 
#### images/logos
anyone can use the images here but they can not edit or delete them without the correct permissions

#### images/marketing
these images can only be accessed by people in the marketing group who can do what they want with them

#### images/public
anyone can see, use and update images here

brianteeman:
If you split this into 2 pr (subdirectories and acl) then they can be addressed separately. The acl is a bigger issue that subdirectories. My 2c

carcam:
I have tested this item :red_circle: unsuccessfully on ed80823c4c7ec90562e163823e5d61ae60dc19dc<br /><br />I am not sure if I understand the scope of this PR properly and I think it's a great improvement but in my opinion, it needs some work. Also I have found some a11y issues:

## Permission issues

1. When you set restrictive permissions to a subfolder but loose permissions to the parent folder, you can access, view and manage content in the subfolder with no issues.

Example of restrictive permissions on a subfolder and loose permissions on the parent

Children can be seen as you have permission for parent

1. When saving conflictive configuration in the plugin, at least if loose permissions are last, they prevail over more restrictive permissions

Conflicting permissions for the same folder

Looser permissions prevail

## A11y issue

1. When you have different folders configured, you cannot access them using the keyboard as tabindex = -1 for all of them. Also subfolders cannot be accessed in the tree area.

Edited to add screenshots<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35931">issues.joomla.org/tracker/joomla-cms/35931</a>.</sub>

bembelimen:
> ACL for this is great but are you sure Access Levels is the correct way?

In theory, ACL permissions are the correct way to do it, but the limitations are, that you need assets for them, which means, that we have to create DB entries for the folders (and files) which is not a good way in general for folders/files here. It's also not really forseen to make things with the toolbar buttons etc. + there is just no interface to manage permission here.

edit: although how it's implemented now, it's not that wrong, it defines, if you "see" an entry.


> ## Permission issues
> 
> When you set restrictive permissions to a subfolder but loose permissions to the parent folder, you can access, view and manage content in the subfolder with no issues.

The entries are "positive inherit", means when you have access to a folder you have always access to all subfolder.

> ## A11y issue
> 
> When you have different folders configured, you cannot access them using the keyboard as tabindex = -1 for all of them. Also subfolders cannot be accessed in the tree area.

https://github.com/joomla/joomla-cms/pull/35911

carcam:
> The entries are "positive inherit", means when you have access to a folder you have always access to all subfolder.

But for all other permissions in Joomla we always do "negative inherit", don't we? That might be confussing. Also I think it might be an issue in some configurations as you usually provide All permissions to /image but you might want to "hide" some subfolders for some users.

plugins/filesystem/local/local.xml
```diff
@@ -40,10 +40,18 @@
...
+							recursive="1"
 							hide_none="true"
 							validate="options"
 						/>
+						<field
+							name="access"
+							type="accesslevel"
+							label="PLG_FILESYSTEM_LOCAL_DIRECTORIES_ACCESS_LABEL"
+						>
+							<option value="">JALL</option>
```
sandewt, line 52: Move aside **>**
```
	>
	<option value="">JALL</option>
	<option value="-1" disabled="disabled">-</option>
```

Issue #35949, opened 2021-10-30
Maintenance: Database : You have an error in your SQL syntax; check the manual that corresponds to your MariaDB


aminweb2:
after upadated joomla 4.0.3 to 4.0.4 , display a error in Maintenance Database:

> You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ' AND `null` = 'NO'' at line 1

### System information (as much as possible)
Joomla 4.0.4
Database Version 	10.3.31-MariaDB-0ubuntu0.20.04.1 
Database Type 	mysql 
PHP Version 	7.4.22 
Web Server 	LiteSpeed

richard67:
@aminweb2 Do you have any 3rd party extensions installed?

aminweb2:
@richard67 of course, but it is not matter. how can find where happen this error?

richard67:
> @richard67 of course, but it is not matter.

@aminweb2 I would not ask if it does not matter. In Joomla 4 the database checker checks also for 3rd party extensions.

> how can find where happen this error?

There is no easy way for you. If you switch on debug in global configuration you will only find out that it fails in the database checker. But you can give it a try. Switching on debug should cause a stack trace being shown. If so, make a screenshot and post here.

And please let me know which 3rd party extensions you have installed and if possible, if they have update SQL scripts.

And if the database page shows not only the error message, make a screenshot and post it here, too.

Thanks in advance.

richard67:
@aminweb2 I think know where the error happens, but I need some information from you to find what causes it.

What I knowe it that the error is caused by either an `ALTER TABLE <tableName> MODIFY <columnName>...` or an `ALTER TABLE <tableName> CHANGE <columnName>...` SQL statement in an update SQL script, with `<tableName>` being the name of the table and v<columnName>` being the name of the column to be modified.

It is very likely not an update SQL script of the core but one of a 3rd party extension.

There is either a syntax error in that statement or there is wrong names quoting used for the table or the column name, or a table or column name is a reserved word in SQL and has no names quoting ("`" on MySQL and MariaDB).

You could help me with finding the exact mistake in that SQL script so that I can try to make a fix for the database checker.

Do you have access to the file system of your Joomla installation? If so, could you do following?
[...]

Issue #35915, opened 2021-10-27
Update -> Database Table Structure Up to Date: NO


Jokerminator:
Joomla 4.0.3 hosted @ Joomla.com (CCP)

### Steps to reproduce the issue
Pre-Update Check for Joomla ‎4.0.4 says "Database Table Structure Up to Date: NO
Go to 'System - Maintenance - Database' and use the 'Update Structure' button."
But the "Update Structure" button is grayed out and the page says "No Matching Results"!

### Expected result
I can click the "Update Structure" button to meet the update requirements.

### Actual result
"Update Structure" button is grayed out, not clickable and the page says "No Matching Results"

### System information (as much as possible)
I created the site 2 days ago @ joomla.com and updated the PHP version to 8.0.1.1.

ghost:
Correction: PHP 8.0.11<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35915">issues.joomla.org/tracker/joomla-cms/35915</a>.</sub>

richard67:
Does the page with the button show a table with rows, one for the CMS core and the possibly more rows for 3rd party extensions?  Each of these rows should have a badge showing if there are problems found or not, and when a badge shows that there are problems, details are shown in a tool tip when hovering the badge.

To fix probles it first needs to select the corresponding row(s) before using the “Fix” button. Each row has a check box at the left side for selecting it, like all the other list view in backend have.

Please report back. if possible with screenshots, how that page looks for you and which problems were found and if the “Fix” button works for you when having selected a row.

if the page does not show any rows, please report back which 3rd party extensions you have installed, because very likely it would be a syntax error in an SQL update script if such an extension which causes the database checker break.

Thanks in advance.

It’s a new feature in Joomla 4 that the database checker checks not only the CMS core but also 3rd party extensions if they have update SQL scripts.

ghost:
I attached a screenshot of what I've seen. There is no fix button.

Looks like cloudaccess.net ( the joomla.com hoster ) is working on
something, because the site is *extremely* slow I get kicked out after
about only one minute of being logged in.

richard67:
@XenoEgger I don't see any screenshot, but as I understand it would not really show much, so that's ok. Let us know here if there is something new.

ghost:
[...]

PR #35951, opened 2021-10-31
[4.0] Fix exception handling in the database checker's change item

richard67:
Pull Request for Issues #35949 and #35915 .

### Summary of Changes

When an SQL statement in an update SQL script of the core or (new in Joomla 4!) a 3rd party extension has a syntax error in an `ALTER TABLE <tableName> MODIFY <columnName>...` or an `ALTER TABLE <tableName> CHANGE <columnName>...` SQL statement, the check query used by the database checker to check that column will fail with an SQL syntax error.

This error happens already when calling `$this->db->setQuery`, but the try block to handle exceptions does not include that, it handles only the `$this->db->loadRowList` call.

This leads to the database checker being broken, see testing instructions or the issue for details.

This PR fixes it by moving the `$this->db->setQuery` inside the try block in the "check" method of the change item.

The same applies when the database checker would try to fix that problem, so it needs the same fix for the "fix" method.

In addition I've fixed the wrong documentation of the "check" method.

### Testing Instructions

1. Install Weblinks 4.0.0. You can download it from here: https://github.com/joomla-extensions/weblinks/releases/download/4.0.0/pkg-weblinks-4.0.0.zip .

2. Open the following update SQL script of the Weblinks component in a text editor:
- administrator/components/com_weblinks/sql/updates/mysql/4.0.0.sql when using a MySQL or MariaDB database
- administrator/components/com_weblinks/sql/updates/postgresql/4.0.0.sql when using a PostgreSQL database

3. In the first SQL statement, remove the ending names quote for the **table** name and change the **column** name so it looks as follows:
- With MySQL or MariaDB database
```
ALTER TABLE `#__weblinks MODIFY `createdXXX` datetime NOT NULL;
```
- With PostgreSQL database
```
ALTER TABLE "#__weblinks ALTER COLUMN "createdXXX" DROP DEFAULT;
```
After having made the change, save the file.

4. Go to the "System" dashboard and check the icon shown for "Maintenance: Database".
Result: The database check failed.
2021-10-31_snap-1

5. Use the link to go to the database checker.
Result: An error alert shows a message about an SQL syntax error, and the database checker doesn't show any rows with the core or extensions, so nothing can be fixed.
Hint: Depending on your database server type and version it might be a different error message which is shown. The important thing is that there is an error message about an SQL error and nothing is shown below it except of the "No matching results" alert.
2021-10-31_snap-2

6. Apply the patch of this PR.

7. Repeat step 4.
Result: The database check shows there is a problem.
2021-10-31_snap-3

8. Repeat step 5.
Result: An error alert shows a message about an SQL syntax error, and the database checker shows rows with the core or extensions. The tool tip shown when moving over the badge with the problem shows which database column in which SQL script has caused the problem.
2021-11-01_snap-1

9. Select the Weblinks component using the check box at the left of the row and use the "Update Structure" button.
Result: No change, but also no uncaught exception like it would happen without the 2nd change in this PR for the "fix" method of the change item.

10. Code review: Check that the changes in the documentation of the "check" method here reflect what the function does.
Note: "Skipped" means that a check item has no check query because it does not change structure, e.g. an INSERT statement, and these will later be reported as "skipped" in the tool tip with the details.

Database checker broken.

Database checker works.

### Other information

For 3.10 we don't need this fix because when having an SQL error in a core update SQL script like tested here with the Weblinks component, the database checker view is not broken on a 3.10.

None for this PR.

But as far as I know our Joomla 4 documentation lacks information about the fact that in Joomla 4 the database schema checker also checks update SQL scripts of 3rd party extensions.

### DIFF
--- a/libraries/src/Schema/ChangeItem.php
+++ b/libraries/src/Schema/ChangeItem.php
@@ -201,7 +201,7 @@ abstract class ChangeItem
 				$this->db->setQuery($this->checkQuery);
 				$rows = $this->db->loadRowList(0);
 			}
-			catch (\Exception $e)
+			catch (\RuntimeException $e)
 			{
 				// Still render the error message from the Exception object
 				Factory::getApplication()->enqueueMessage($e->getMessage(), 'error');
@@ -252,7 +252,7 @@ abstract class ChangeItem
 					$this->rerunStatus = -2;
 				}
 			}
-			catch (\Exception $e)
+			catch (ExecutionFailureException | \RuntimeException $e)
 			{
 				$this->rerunStatus = -2;
 			}
--- a/libraries/src/Schema/ChangeItem.php
+++ b/libraries/src/Schema/ChangeItem.php
@@ -184,9 +184,8 @@ abstract class ChangeItem
 
 	/**
 	 * Runs the check query and checks that 1 row is returned
-	 * If yes, return true, otherwise return false
 	 *
-	 * @return  boolean  true on success, false otherwise
+	 * @return  integer  1 if success, -1 if skipped, -2 if check failed
 	 *
 	 * @since  2.5
 	 */
--- a/libraries/src/Schema/ChangeItem.php
+++ b/libraries/src/Schema/ChangeItem.php
@@ -196,13 +196,12 @@ abstract class ChangeItem
 
 		if ($this->checkQuery)
 		{
-			$this->db->setQuery($this->checkQuery);
-
 			try
 			{
+				$this->db->setQuery($this->checkQuery);
 				$rows = $this->db->loadRowList(0);
 			}
-			catch (\RuntimeException $e)
+			catch (\Exception $e)
 			{
 				// Still render the error message from the Exception object
 				Factory::getApplication()->enqueueMessage($e->getMessage(), 'error');
@@ -238,10 +237,9 @@ abstract class ChangeItem
 			// At this point we have a failed query
 			$query = $this->updateQuery;
 
-			$this->db->setQuery($query);
-
 			try
 			{
+				$this->db->setQuery($query);
 				$this->db->execute();
 
 				if ($this->check())
@@ -254,7 +252,7 @@ abstract class ChangeItem
 					$this->rerunStatus = -2;
 				}
 			}
-			catch (ExecutionFailureException $e)
+			catch (\Exception $e)
 			{
 				$this->rerunStatus = -2;
 			}


richard67:
I think this is a release blocker.

richard67:
Reason for release blocker: In J4 it is much more likely than in J3 that the issue breaks the database checker because in J4 it checks also update SQL scripts of 3rd party extensions.

rjharishabh:
I have tested this item :white_check_mark: successfully on 18815fb0a341eaee79a8bb269da550abcc022662<br /><br />PHP 8.0.6 and MySQLi 10.4.19-MariaDB<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on 18815fb0a341eaee79a8bb269da550abcc022662<br /><br />Patch works as described.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

RickR2H:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

libraries/src/Schema/ChangeItem.php
```diff
@@ -196,13 +196,12 @@ public function check()
...
 		{
-			$this->db->setQuery($this->checkQuery);
-
 			try
 			{
+				$this->db->setQuery($this->checkQuery);
 				$rows = $this->db->loadRowList(0);
 			}
-			catch (\RuntimeException $e)
+			catch (\Exception $e)
```
wilsonge: Why is runtime not enough here? I thought the framework only threw runtimes?

richard67: @wilsonge I thought it's a good idea to catch anything. Here it seems there was expected that the frameworks throws something else: https://github.com/joomla/joomla-cms/pull/35951/files#diff-64a03c6e98e4fa15e90e706da227307887df2b68ecc691e7daa7b56183873eeeL257 .

Shall I change back to `\RuntimeException` here and at the other place change to `ExecutionFailureException | \RuntimeException`?

richard67: Or would at the other place `\RuntimeException` be sufficient because `ExecutionFailureException` extends from that? https://github.com/joomla/joomla-cms/pull/35951/files#diff-64a03c6e98e4fa15e90e706da227307887df2b68ecc691e7daa7b56183873eeeR254-R255

I think we need the `\RuntimeException` for the `$this->db->setQuery` call because it doesn't throw an `ExecutionFailureException`.

richard67: Ok, I've changed it.

richard67:
@rjharishabh @RickR2H Could you test again? Note that there is a new test step 10 for a code review check. Thanks in advance.

richard67:
Back to pending. Needs new tests due to changes.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

rjharishabh:
I have tested this item :white_check_mark: successfully on 25c8959e6e174a55339cdd96ed5fac8ffad00e3f<br /><br />PHP 8.0.6 and MySQLi 10.4.19-MariaDB
XAMPP on Windows 10<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 25c8959e6e174a55339cdd96ed5fac8ffad00e3f<br /><br />postgres<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35951">issues.joomla.org/tracker/joomla-cms/35951</a>.</sub>

richard67:
@wilsonge Better now?

wilsonge:
Yup - thanks!

richard67:
Thanks all.

PR #35952, opened 2021-10-31
[3.10] Fix exception handling in the database checker's change item

richard67:
Pull Request for Issue # .

### Summary of Changes

This the backport of the J4 PR #35951 .

In opposite to 4.0-dev there is not yet any exception handling in the "fix" method of the change item, so this PR here adds that, too, so at the end it's the same as in 4.0-dev.

### Testing Instructions

Will be added soon.

Will be added soon.

Will be added soon.

None.

### DIFF
--- a/libraries/src/Schema/ChangeItem.php
+++ b/libraries/src/Schema/ChangeItem.php
@@ -192,13 +192,12 @@ abstract class ChangeItem
 
 		if ($this->checkQuery)
 		{
-			$this->db->setQuery($this->checkQuery);
-
 			try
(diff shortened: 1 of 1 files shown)


richard67:
Hmm, on 3.10-dev it seems that the database checker is not broken with a syntax error in an update SQL for the core like it is for 4.0-dev with an error in a 3rd party SQL script. That could be due to the changes in 4.0-dev for the checker handling 3rd party extensions.

So here we do not need that fix. Closing.

PR #35986, opened 2021-11-07
refactor(cache): Reorganize the Page Cache with SubscriberInterface

anibalsanchez:
Pull Request for Issue #35722 #35625 #35877. 

*This PR for Joomla 4.1 is the follow-up of the previous issues #35722 #35625 #35877 to solve the same problem using the Joomla 4 `SubscriberInterface`. It is the same function but uses the new plugin event subscriber interface.*

### Summary of Changes

This PR refactorizes the original System Page Cache to work based on J4 `SubscriberInterface`. The plugin is an interesting case to use the `SubscriberInterface`. The plugin is not executed if it doesn't subscribe to any event. The plugin must only be run under these conditions:

- GET Method
- Site App
- Site is Live 
- User is Guest
- No messages in the queue
- The current Page is not included in the Exclude URLs settings

On Joomla 3, all event methods are always called, leading to duplicated checks and inconsistencies on all methods.

The plugin refactorization avoids duplicated code.

### Testing Instructions

1. Test that the page is cached and retrieved in the cases detailed above.
1. Test that the page is not cached in these cases: 

- Not GET Method
- CLI/ Administrator/ Api App
- Site is Offline 
- Registered User
- A message in the queue
- The current Page is included in the Exclude URLs settings

The event methods are called following the Joomla 3 model (`onAfterRoute`, 'onAfterRender', 'onAfterRespond').

The final result of delivering the cached page is the same.

The event methods are called following the Joomla 4 `SubscriberInterface`  (`verifyIfCurrentPageCanBeCached`, 'dumpCachedPage', 'storePage').

The final result of delivering the cached page is the same.

As mentioned @HLeithner before here #35877, there is a need to document how to develop a Joomla 4 plugin using the new `SubscriberInterface` and avoid the previous way to define the events methods.

CC: @alikon @HLeithner @Kubik-Rubik @PhilETaylor @tampe125 @nikosdion @brianteeman @richard67 @ceford @ricardo1709

### DIFF
--- a/plugins/system/cache/cache.php
+++ b/plugins/system/cache/cache.php
@@ -15,13 +15,15 @@ use Joomla\CMS\Plugin\CMSPlugin;
 use Joomla\CMS\Plugin\PluginHelper;
 use Joomla\CMS\Profiler\Profiler;
 use Joomla\CMS\Uri\Uri;
+use Joomla\Event\SubscriberInterface;
+use Joomla\Registry\Registry;
 
 /**
  * Joomla! Page Cache Plugin.
  *
  * @since  1.5
  */
-class PlgSystemCache extends CMSPlugin
+class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 {
 	/**
 	 * Cache instance.
@@ -68,7 +70,27 @@ class PlgSystemCache extends CMSPlugin
 
 		// Instantiate cache with previous options and create the cache key identifier.
 		$this->_cache     = Cache::getInstance('page', $options);
-		$this->_cache_key = Uri::getInstance()->toString();
+		$this->_cache_key = self::simpleCacheKey();
+	}
+
+	/**
+	 * Returns an array of events this subscriber will listen to.
+	 *
+	 * @return  array
+	 *
+	 * @since   4.0.0
+	 */
+	public static function getSubscribedEvents(): array
+	{
+		if (!self::canPageBeCached()) {
+			return [];
+		}
+
+		return [
+			'onAfterRender' => 'verifyIfCurrentPageCanBeCached',
+			'onAfterRoute' => 'dumpCachedPage',
+			'onAfterRespond' => 'storePage',
+		];
 	}
 
 	/**
@@ -96,7 +118,8 @@ class PlgSystemCache extends CMSPlugin
 	}
 
 	/**
-	 * Checks if URL exists in cache, if so dumps it directly and closes.
+	 * onAfterRoute.
+	 * Only for BC, not used on J4
 	 *
 	 * @return  void
 	 *
@@ -104,18 +127,26 @@ class PlgSystemCache extends CMSPlugin
 	 */
 	public function onAfterRoute()
 	{
-		if ($this->app->isClient('administrator') || $this->app->get('offline', '0') || $this->app->getMessageQueue())
-		{
-			return;
-		}
+		$this->dumpCachedPage();
+	}
 
+	/**
+	 * Checks if URL exists in cache, if so dumps it directly and closes.
+	 *
+	 * @return  void
+	 *
+	 * @since   4.1
+	 */
+	public function dumpCachedPage()
+	{
 		// If any pagecache plugins return false for onPageCacheSetCaching, do not use the cache.
 		PluginHelper::importPlugin('pagecache');
 
 		$results = $this->app->triggerEvent('onPageCacheSetCaching');
 		$caching = !in_array(false, $results, true);
 
-		if ($caching && $this->app->getIdentity()->guest && $this->app->input->getMethod() === 'GET')
+		// Double check for BC
+		if ($caching && $this->canPageBeCached())
 		{
 			$this->_cache->setCaching(true);
 		}
@@ -148,8 +179,8 @@ class PlgSystemCache extends CMSPlugin
 	}
 
 	/**
-	 * After Render Event.
-	 * Verify if current page is not excluded from cache.
+	 * onAfterRender.
+	 * Only for BC, not used on J4.
 	 *
 	 * @return   void
 	 *
@@ -157,14 +188,22 @@ class PlgSystemCache extends CMSPlugin
 	 */
 	public function onAfterRender()
 	{
-		if ($this->_cache->getCaching() === false)
-		{
-			return;
-		}
+		$this->verifyIfCurrentPageCanBeCached();
+	}
 
-		// We need to check if user is guest again here, because auto-login plugins have not been fired before the first aid check.
+	/**
+	 * Verify if current page is not excluded from cache.
+	 *
+	 * @return   void
+	 *
+	 * @since   4.1
+	 */
+	public function verifyIfCurrentPageCanBeCached()
+	{
+		// We need to check if user is guest again here,
+		// 	because auto-login plugins have not been fired before the first aid check.
 		// Page is excluded if excluded in plugin settings.
-		if (!$this->app->getIdentity()->guest || $this->app->getMessageQueue() || $this->isExcluded() === true)
+		if (!$this->canPageBeCached())
 		{
 			$this->_cache->setCaching(false);
 
@@ -176,8 +215,8 @@ class PlgSystemCache extends CMSPlugin
 	}
 
 	/**
-	 * After Respond Event.
-	 * Stores page in cache.
+	 * onAfterRespond.
+	 * Only for BC, not used on J4
 	 *
 	 * @return   void
 	 *
@@ -185,6 +224,20 @@ class PlgSystemCache extends CMSPlugin
 	 */
 	public function onAfterRespond()
 	{
+		$this->storePage();
+	}
+
+	/**
+	 * After Respond Event.
+	 * Stores page in cache.
+	 *
+	 * @return   void
+	 *
+	 * @since   1.5
+	 */
+	public function storePage()
+	{
+		// if it can't be cached due to execution conditions (check verifyIfCurrentPageCanBeCached/canPageBeCached)
 		if ($this->_cache->getCaching() === false)
 		{
 			return;
@@ -195,19 +248,83 @@ class PlgSystemCache extends CMSPlugin
 	}
 
 	/**
-	 * Check if the page is excluded from the cache or not.
+	 * Check if the page can be cached according to all conditions.
+	 *
+	 * @return   boolean  True if the page can be cached
+	 *
+	 * @since    4.1
+	 */
+	private static function canPageBeCached(): bool
+	{
+		$app = Factory::getApplication();
+
+		// Only GET requests are cached.
+		if ($app->input->getMethod() !== 'GET')
+		{
+			return false;
+		}
+
+		// Run only when we're on Site Application side
+		if (!$app->isClient('site'))
+		{
+			return false;
+		}
+
+		// If the site is offline, don't do anything
+		if ((bool) $app->get('offline', '0'))
+		{
+			return false;
+		}
+
+		// The user is authenticated
+		if (!$app->getIdentity()->guest) {
+			return false;
+		}
+
+		// If there are messages in the queue, don't cache the page
+		if (!empty($app->getMessageQueue()))
+		{
+			return false;
+		}
+
+		if (self::isExcludedPage()) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * onAfterRoute.
+	 * Only for BC, not used on J4
 	 *
 	 * @return   boolean  True if the page is excluded else false
 	 *
 	 * @since    3.5
 	 */
-	protected function isExcluded()
+	protected static function isExcluded()
 	{
+		return self::isExcludedPage();
+	}
+
+	/**
+	 * Check if the page is excluded from the cache or not.
+	 *
+	 * @return   boolean  True if the page is excluded else false
+	 *
+	 * @since    4.1
+	 */
+	private static function isExcludedPage(): bool
+	{
+		$app = Factory::getApplication();
+		$plugin = PluginHelper::getPlugin('system', 'cache');
+		$params = new Registry($plugin->params);
+
 		// Check if menu items have been excluded.
-		if ($exclusions = $this->params->get('exclude_menu_items', array()))
+		if ($exclusions = $params->get('exclude_menu_items', array()))
 		{
 			// Get the current menu item.
-			$active = $this->app->getMenu()->getActive();
+			$active = $app->getMenu()->getActive();
 
 			if ($active && $active->id && in_array((int) $active->id, (array) $exclusions))
 			{
@@ -216,7 +333,7 @@ class PlgSystemCache extends CMSPlugin
 		}
 
 		// Check if regular expressions are being used.
-		if ($exclusions = $this->params->get('exclude', ''))
+		if ($exclusions = $params->get('exclude', ''))
 		{
 			// Normalize line endings.
 			$exclusions = str_replace(array("\r\n", "\r"), "\n", $exclusions);
@@ -225,7 +342,7 @@ class PlgSystemCache extends CMSPlugin
 			$exclusions = explode("\n", $exclusions);
 
 			// Gets internal URI.
-			$internal_uri	= '/index.php?' . Uri::getInstance()->buildQuery($this->app->getRouter()->getVars());
+			$internalUri = '/index.php?' . Uri::getInstance()->buildQuery($app->getRouter()->getVars());
 
 			// Loop through each pattern.
 			if ($exclusions)
@@ -236,7 +353,7 @@ class PlgSystemCache extends CMSPlugin
 					if ($exclusion !== '')
 					{
 						// Test both external and internal URI
-						if (preg_match('#' . $exclusion . '#i', $this->_cache_key . ' ' . $internal_uri, $match))
+						if (preg_match('#' . $exclusion . '#i', self::simpleCacheKey() . ' ' . $internalUri, $match))
 						{
 							return true;
 						}
@@ -248,8 +365,20 @@ class PlgSystemCache extends CMSPlugin
 		// If any pagecache plugins return true for onPageCacheIsExcluded, exclude.
 		PluginHelper::importPlugin('pagecache');
 
-		$results = $this->app->triggerEvent('onPageCacheIsExcluded');
+		$results = $app->triggerEvent('onPageCacheIsExcluded');
 
 		return in_array(true, $results, true);
+	}
+
+	/**
+	 * Generate the cache key.
+	 *
+	 * @return   string  The cache key
+	 *
+	 * @since    4.1
+	 */
+	private static function simpleCacheKey(): string
+	{
+		return Uri::getInstance()->toString();
 	}
 }
--- a/plugins/system/cache/cache.php
+++ b/plugins/system/cache/cache.php
@@ -121,12 +121,15 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 * onAfterRoute.
 	 * Only for BC, not used on J4
 	 *
-	 * @return  void
+	 * @return  	void
 	 *
-	 * @since   4.0.0
+	 * @since   	4.0.0
+	 * @deprecated  5.0 Use \Joomla\Event\SubscriberInterface instead
 	 */
 	public function onAfterRoute()
 	{
+		@trigger_error('The '.__METHOD__.' method is deprecated since version 4.1. Use SubscriberInterface instead.', E_USER_DEPRECATED);
+
 		$this->dumpCachedPage();
 	}
 
@@ -182,12 +185,15 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 * onAfterRender.
 	 * Only for BC, not used on J4.
 	 *
-	 * @return   void
+	 * @return   	void
 	 *
-	 * @since   3.9.12
+	 * @since   	3.9.12
+	 * @deprecated  5.0 Use \Joomla\Event\SubscriberInterface instead
 	 */
 	public function onAfterRender()
 	{
+		@trigger_error('The '.__METHOD__.' method is deprecated since version 4.1. Use SubscriberInterface instead.', E_USER_DEPRECATED);
+
 		$this->verifyThatCurrentPageCanBeCached();
 	}
 
@@ -217,12 +223,15 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 * onAfterRespond.
 	 * Only for BC, not used on J4
 	 *
-	 * @return   void
+	 * @return		void
 	 *
-	 * @since   1.5
+	 * @since	  	1.5
+	 * @deprecated  5.0 Use \Joomla\Event\SubscriberInterface instead
 	 */
 	public function onAfterRespond()
 	{
+		@trigger_error('The '.__METHOD__.' method is deprecated since version 4.1. Use SubscriberInterface instead.', E_USER_DEPRECATED);
+
 		$this->storePage();
 	}
 
--- a/plugins/system/cache/cache.php
+++ b/plugins/system/cache/cache.php
@@ -78,7 +78,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return  array
 	 *
-	 * @since   4.0.0
+	 * @since   4.1.0
 	 */
 	public static function getSubscribedEvents(): array
 	{
--- a/plugins/system/cache/cache.php
+++ b/plugins/system/cache/cache.php
@@ -78,7 +78,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return  array
 	 *
-	 * @since   4.1.0
+	 * @since   __DEPLOY_VERSION__
 	 */
 	public static function getSubscribedEvents(): array
 	{
@@ -135,7 +135,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return  void
 	 *
-	 * @since   4.1
+	 * @since   __DEPLOY_VERSION__
 	 */
 	public function dumpCachedPage()
 	{
@@ -196,7 +196,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return   void
 	 *
-	 * @since   4.1
+	 * @since   __DEPLOY_VERSION__
 	 */
 	public function verifyIfCurrentPageCanBeCached()
 	{
@@ -252,7 +252,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return   boolean  True if the page can be cached
 	 *
-	 * @since    4.1
+	 * @since    __DEPLOY_VERSION__
 	 */
 	private static function canPageBeCached(): bool
 	{
@@ -312,7 +312,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return   boolean  True if the page is excluded else false
 	 *
-	 * @since    4.1
+	 * @since    __DEPLOY_VERSION__
 	 */
 	private static function isExcludedPage(): bool
 	{
@@ -375,7 +375,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 *
 	 * @return   string  The cache key
 	 *
-	 * @since    4.1
+	 * @since    __DEPLOY_VERSION__
 	 */
 	private static function simpleCacheKey(): string
 	{
--- a/plugins/system/cache/cache.php
+++ b/plugins/system/cache/cache.php
@@ -87,7 +87,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 		}
 
 		return [
-			'onAfterRender' => 'verifyIfCurrentPageCanBeCached',
+			'onAfterRender' => 'verifyThatCurrentPageCanBeCached',
 			'onAfterRoute' => 'dumpCachedPage',
 			'onAfterRespond' => 'storePage',
 		];
@@ -188,21 +188,20 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 */
 	public function onAfterRender()
 	{
-		$this->verifyIfCurrentPageCanBeCached();
+		$this->verifyThatCurrentPageCanBeCached();
 	}
 
 	/**
-	 * Verify if current page is not excluded from cache.
+	 * Verify that current page is not excluded from cache.
 	 *
 	 * @return   void
 	 *
 	 * @since   __DEPLOY_VERSION__
 	 */
-	public function verifyIfCurrentPageCanBeCached()
+	public function verifyThatCurrentPageCanBeCached()
 	{
-		// We need to check if user is guest again here,
-		// 	because auto-login plugins have not been fired before the first aid check.
-		// Page is excluded if excluded in plugin settings.
+		// Check if the user is a guest again because auto-login plugins
+		// 	have not been fired when getSubscribedEvents was called.
 		if (!$this->canPageBeCached())
 		{
 			$this->_cache->setCaching(false);
@@ -237,7 +236,7 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	 */
 	public function storePage()
 	{
-		// if it can't be cached due to execution conditions (check verifyIfCurrentPageCanBeCached/canPageBeCached)
+		// if it can't be cached due to execution conditions (check verifyThatCurrentPageCanBeCached/canPageBeCached)
 		if ($this->_cache->getCaching() === false)
 		{
 			return;
--- a/plugins/system/cache/cache.php
+++ b/plugins/system/cache/cache.php
@@ -9,38 +9,28 @@
 
 defined('_JEXEC') or die;
 
-use Joomla\CMS\Cache\Cache;
 use Joomla\CMS\Factory;
 use Joomla\CMS\Plugin\CMSPlugin;
 use Joomla\CMS\Plugin\PluginHelper;
 use Joomla\CMS\Profiler\Profiler;
 use Joomla\CMS\Uri\Uri;
 use Joomla\Event\SubscriberInterface;
-use Joomla\Registry\Registry;
+use Joomla\Plugin\System\Cache\Helpers\PageCacheKeyGenerator;
+use Joomla\Plugin\System\Cache\Helpers\PageCacheStorage;
+use Joomla\Plugin\System\Cache\Helpers\PageCachingChecker;
+
+// Why is not autolading the <namespace path="src">Joomla\Plugin\System\Cache</namespace>?
+require 'src/Helpers/PageCacheKeyGenerator.php';
+require 'src/Helpers/PageCachingChecker.php';
+require 'src/Helpers/PageCacheStorage.php';
 
 /**
  * Joomla! Page Cache Plugin.
  *
  * @since  1.5
  */
-class PlgSystemCache extends CMSPlugin implements SubscriberInterface
+final class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 {
-	/**
-	 * Cache instance.
-	 *
-	 * @var    \Joomla\CMS\Cache\CacheController
-	 * @since  1.5
-	 */
-	public $_cache;
-
-	/**
-	 * Cache key
-	 *
-	 * @var    string
-	 * @since  3.0
-	 */
-	public $_cache_key;
-
 	/**
 	 * Application object.
 	 *
@@ -50,28 +40,12 @@ class PlgSystemCache extends CMSPlugin implements SubscriberInterface
 	protected $app;
 
 	/**
-	 * Constructor.
+	 * Page Cache Storage
 	 *
-	 * @param   object  &$subject  The object to observe.
-	 * @param   array   $config    An optional associative array of configuration settings.
-	 *
-	 * @since   1.5
+	 * @var    Joomla\Plugin\System\Cache\Helpers\PageCacheStorage
+	 * @since  __DEPLOY_VERSION__
 	 */
-	public function __construct(&$subject, $config)
-	{
-		parent::__construct($subject, $config);
-
-		// Set the cache options.
-		$options = array(
-			'defaultgroup' => 'page',
-			'browsercache' => $this->params->get('browsercache', 0),
-			'caching'      => false,
-		);
-
-		// Instantiate cache with previous options and create the cache key identifier.
-		$this->_cache     = Cache::getInstance('page', $options);
-		$this->_cache_key = self::simpleCacheKey();
-	}
+	protected $cacheStorage;
 
 	/**
 	 * Returns an array of events this subscriber will listen to.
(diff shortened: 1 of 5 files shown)


plugins/system/cache/cache.php
```diff
@@ -96,26 +118,35 @@ protected function getCacheKey()
...
 	/**
-	 * Checks if URL exists in cache, if so dumps it directly and closes.
+	 * onAfterRoute.
+	 * Only for BC, not used on J4
 	 *
 	 * @return  void
 	 *
 	 * @since   4.0.0
 	 */
 	public function onAfterRoute()
```
Kubik-Rubik: Please do not mix the Observer with the new Event approach. If you use `SubscriberInterface` with the `getSubscribedEvents` method, then the classes are not called, and you should remove them entirely. No need to keep them; there is no BC break with this change.

plugins/system/cache/cache.php
```diff
@@ -195,19 +248,83 @@ public function onAfterRespond()
...
 
 	/**
-	 * Check if the page is excluded from the cache or not.
+	 * Check if the page can be cached according to all conditions.
+	 *
+	 * @return   boolean  True if the page can be cached
+	 *
+	 * @since    4.1
+	 */
+	private static function canPageBeCached(): bool
```
Kubik-Rubik: This change might introduce a new bug and must be checked thoroughly. We previously had different checks for the creating and loading processes, and now, you do the same checks in both methods! For instance, the excluded pages have been checked only in the creation process before this change. Please make sure that this transition will not change the behaviour of the plugin!

plugins/system/cache/cache.php
```diff
@@ -68,7 +70,27 @@ public function __construct(&$subject, $config)
...
-		$this->_cache_key = Uri::getInstance()->toString();
+		$this->_cache_key = self::simpleCacheKey();
+	}
+
+	/**
+	 * Returns an array of events this subscriber will listen to.
+	 *
+	 * @return  array
+	 *
+	 * @since   4.0.0
```
brianteeman: ??

plugins/system/cache/cache.php
```diff
@@ -96,26 +118,35 @@ protected function getCacheKey()
...
-		}
+		$this->dumpCachedPage();
+	}
 
+	/**
+	 * Checks if URL exists in cache, if so dumps it directly and closes.
+	 *
+	 * @return  void
+	 *
+	 * @since   4.1
```
brianteeman:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

plugins/system/cache/cache.php
```diff
@@ -148,23 +179,31 @@ public function onAfterRoute()
...
-		if ($this->_cache->getCaching() === false)
-		{
-			return;
-		}
+		$this->verifyIfCurrentPageCanBeCached();
+	}
 
-		// We need to check if user is guest again here, because auto-login plugins have not been fired before the first aid check.
+	/**
+	 * Verify if current page is not excluded from cache.
```
brianteeman:
```suggestion
	 * Verify that current page is not excluded from cache.
```

plugins/system/cache/cache.php
```diff
@@ -148,23 +179,31 @@ public function onAfterRoute()
...
+	 *
+	 * @return   void
+	 *
+	 * @since   4.1
```
brianteeman:
```suggestion
	 * @since   __DEPLOY_VERSION__
```

plugins/system/cache/cache.php
```diff
@@ -148,23 +179,31 @@ public function onAfterRoute()
...
+	 */
+	public function verifyIfCurrentPageCanBeCached()
+	{
+		// We need to check if user is guest again here,
+		// 	because auto-login plugins have not been fired before the first aid check.
```
brianteeman: first aid??

plugins/system/cache/cache.php
```diff
@@ -195,19 +248,83 @@ public function onAfterRespond()
 	}
 
 	/**
-	 * Check if the page is excluded from the cache or not.
+	 * Check if the page can be cached according to all conditions.
+	 *
+	 * @return   boolean  True if the page can be cached
+	 *
+	 * @since    4.1
```
brianteeman:
```suggestion
	 * @since    __DEPLOY_VERSION__
```

plugins/system/cache/cache.php
```diff
@@ -195,19 +248,83 @@ public function onAfterRespond()
...
+	{
+		return self::isExcludedPage();
+	}
+
+	/**
+	 * Check if the page is excluded from the cache or not.
+	 *
+	 * @return   boolean  True if the page is excluded else false
+	 *
+	 * @since    4.1
```
brianteeman:
```suggestion
	 * @since    __DEPLOY_VERSION__
```

plugins/system/cache/cache.php
```diff
@@ -248,8 +365,20 @@ protected function isExcluded()
...
 
 		return in_array(true, $results, true);
 	}
+
+	/**
+	 * Generate the cache key.
+	 *
+	 * @return   string  The cache key
+	 *
+	 * @since    4.1
```
brianteeman:
```suggestion
	 * @since    __DEPLOY_VERSION__
```

richard67:
@anibalsanchez In the testing instructions you write that we shall check that the page is not cached when

> Page is not excluded in the plugin configuration

Can it be that this is a copy paste from the description of the case when the page shall be cached and that the „not“ in that sentence is wrong here for the not cached case?

plugins/system/cache/cache.php
anibalsanchez: The cache plugin has already been released on Joomla 4 with these protected methods. If there is a plugin out there using this plugin as the parent class, the removal of them is going to break it.

plugins/system/cache/cache.php
anibalsanchez: There have been several assorted ways of checking if the page had to be cached. Not really clear what has been going on and it looks more like fixes on top of previous fixes.

The idea is to check every condition in a single method `canPageBeCached` to simplify the previous implementation. In the worst case, it is an abundance of caution.

plugins/system/cache/cache.php
anibalsanchez: Fixed. Txs.

plugins/system/cache/cache.php
anibalsanchez: Improved. txs.

plugins/system/cache/cache.php
anibalsanchez: Fixed Txs.

plugins/system/cache/cache.php
anibalsanchez: Fixed Txs.

plugins/system/cache/cache.php
anibalsanchez: Fixed Txs.

plugins/system/cache/cache.php
anibalsanchez: Fixed Txs.

plugins/system/cache/cache.php
anibalsanchez: It has been originally documented in this way. I haven't changed the original caching code and limited the changes to the execution conditions.

I'm going to re-write it to clarify

plugins/system/cache/cache.php
nikosdion: The nature of that method would make such use extremely unlikely. I would propose adding a deprecated log notice and throw a user warning with `trigger_error()` to dissuade direct use of a method that should never be used directly.

plugins/system/cache/cache.php
nikosdion: `aid` here refers to ID not ‘first aid‘ as in ‘first aid kit‘. I believe it had to do with the code back in Joomla 1.0 or thereabouts when the plugin first appeared.

The point of this check is that the user may be logged in by a plugin event triggered after the plugin is initialized. We no longer do this check on initialisation so the comment can be simplified to say that we are simply checking if the user is a guest and the reason it happens late is that plugins may have logged in the user anytime between plugins being loaded and this event being triggered.

anibalsanchez: Improved, pls, check it again.

plugins/system/cache/cache.php
anibalsanchez: Solved, Txs.

richard67:
@anibalsanchez What I mean In my previous comment is following, see the red mark:

anibalsanchez:
Thanks @richard67 I've updated the testing instruction to clarify the inconsistency. Pls, check it again.

plugins/system/cache/cache.php
anibalsanchez: As suggested, I've added the deprecated tags and `trigger_error()`

richard67:
> Thanks @richard67 I've updated the testing instruction to clarify the inconsistency. Pls, check it again.

@anibalsanchez Looks ok now. Thanks.

laoneo:
For me there is too much static stuff in here. I would just return the events in `getSubscribedEvents` and then do the checks if cache is enabled in the respective function once called.

plugins/system/cache/cache.php
```diff
@@ -15,13 +15,15 @@
...
+use Joomla\Event\SubscriberInterface;
+use Joomla\Registry\Registry;
 
 /**
  * Joomla! Page Cache Plugin.
  *
  * @since  1.5
  */
-class PlgSystemCache extends CMSPlugin
+class PlgSystemCache extends CMSPlugin implements SubscriberInterface
```
nikosdion: Remove the constructor. Do not access `_cache` directly. Use a private getter which instantiated `$this->_cache` if it's currently null.

Remove _cache_key and the associated static method `simpleCacheKey`. It's not used anywhere.

plugins/system/cache/cache.php
```diff
@@ -148,23 +182,33 @@ public function onAfterRoute()
...
+	 * @since   	3.9.12
+	 * @deprecated  5.0 Use \Joomla\Event\SubscriberInterface instead
 	 */
 	public function onAfterRender()
 	{
-		if ($this->_cache->getCaching() === false)
-		{
-			return;
-		}
+		@trigger_error('The '.__METHOD__.' method is deprecated since version 4.1. Use SubscriberInterface instead.', E_USER_DEPRECATED);
```
nikosdion: Actually, this is wrong here. The method is still used. If the page can be cached we set the gzip global option to false (which means that Joomla does not returned GZip'ped content to the browser).

plugins/system/cache/cache.php
```diff
@@ -148,23 +182,33 @@ public function onAfterRoute()
...
 
-		// We need to check if user is guest again here, because auto-login plugins have not been fired before the first aid check.
-		// Page is excluded if excluded in plugin settings.
-		if (!$this->app->getIdentity()->guest || $this->app->getMessageQueue() || $this->isExcluded() === true)
+		$this->verifyThatCurrentPageCanBeCached();
```
nikosdion: The method name is misleading. See comment above about method naming.

plugins/system/cache/cache.php
```diff
@@ -96,26 +118,38 @@ protected function getCacheKey()
...
+	{
 		// If any pagecache plugins return false for onPageCacheSetCaching, do not use the cache.
 		PluginHelper::importPlugin('pagecache');
 
 		$results = $this->app->triggerEvent('onPageCacheSetCaching');
 		$caching = !in_array(false, $results, true);
 
-		if ($caching && $this->app->getIdentity()->guest && $this->app->input->getMethod() === 'GET')
+		// Double check for BC
+		if ($caching && $this->canPageBeCached())
```
nikosdion: Since you are calling `$this->canPageBeCached` remove static from the method you are referencing here.

plugins/system/cache/cache.php
```diff
@@ -68,7 +70,27 @@ public function __construct(&$subject, $config)
...
+	/**
+	 * Returns an array of events this subscriber will listen to.
+	 *
+	 * @return  array
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function getSubscribedEvents(): array
+	{
+		if (!self::canPageBeCached()) {
```
nikosdion: Best not check anything here. This is called when the plugin is loaded, possibly before input and identity are initialized. Besides, you are checking `canPageBeCached` on each method anyway.

plugins/system/cache/cache.php
```diff
@@ -68,7 +70,27 @@ public function __construct(&$subject, $config)
...
+			return [];
+		}
+
+		return [
+			'onAfterRender' => 'verifyThatCurrentPageCanBeCached',
```
nikosdion: This is a fine mess... Since you are implementing on `onAfter...` methods anyway use them instead of the misleadingly named methods in here.

Remember that when you use SubscriberInterface Joomla will only register event handlers you specify in this static method. Method names DO NOT have special meaning but it's a hell of a lot easier for us humans to see that the onAfterRender event is handled by a method called onAfterRender which now accepts a single Event argument.

plugins/system/cache/cache.php
nikosdion: Moreover, all methods referenced in getSubscribedEvents must have the signature `function (Event $event): void`. You do not have them accepting an argument which raises a PHP Warning. In future PHP versions this will be a Fatal Error!

plugins/system/cache/cache.php
```diff
@@ -175,6 +219,22 @@ public function onAfterRender()
...
+	/**
+	 * onAfterRespond.
+	 * Only for BC, not used on J4
+	 *
+	 * @return		void
+	 *
+	 * @since	  	1.5
+	 * @deprecated  5.0 Use \Joomla\Event\SubscriberInterface instead
+	 */
+	public function onAfterRespond()
```
nikosdion: Same as previously mentioned. Use this in `getSubscribedEvents`, make it accept an Event argument.

plugins/system/cache/cache.php
```diff
@@ -183,8 +243,9 @@ public function onAfterRender()
 	 *
 	 * @since   1.5
 	 */
-	public function onAfterRespond()
+	public function storePage()
```
nikosdion: Same as previously mentioned. Fold it into onAfterRespond. Moreover, make sure you call canPageBeCached.

plugins/system/cache/cache.php
```diff
@@ -195,19 +256,83 @@ public function onAfterRespond()
...
+	/**
+	 * onAfterRoute.
+	 * Only for BC, not used on J4
 	 *
 	 * @return   boolean  True if the page is excluded else false
 	 *
 	 * @since    3.5
 	 */
-	protected function isExcluded()
+	protected static function isExcluded()
```
nikosdion: This method neither needs to be static nor exist separately from isExcludedPage. Fold it back in here, make it private and remove static.

plugins/system/cache/cache.php
```diff
@@ -248,8 +373,20 @@ protected function isExcluded()
...
 	}
+
+	/**
+	 * Generate the cache key.
+	 *
+	 * @return   string  The cache key
+	 *
+	 * @since    __DEPLOY_VERSION__
+	 */
+	private static function simpleCacheKey(): string
```
nikosdion: This is not used anywhere. Remove it.

nikosdion (changes requested):
Having had the time to fully review the code I am not happy.

When you use SubscriberInterface the method names have no meaning. The association between event names and method names is what getSubscribedEvents returns. 

Since there is no need to “keep b/c” in an internal plugin class that should not be extended you don't need to keep the old method signatures. Just upgrade them to have the correct signature for Event handles (a singular argument of the type `Event`, returns void) and use their original names in the getSubsribedEvents to make it easier for humans to parse what is used when and why. You are not using the correct signature.

Use the check `canPageBeCached` on each method. The check itself should be a regular private method, not static.

Do not use static methods, there is no need for that. These static methods are probably leftovers from the Mambot days. If we're upgrading the plugin we might just as well do it right.

There is no need for a constructor. Returning the cache object should be done lazily with a private getter. This prevents the overhead of setting up caching when we do not need it.

The goal is to simplify this plugin, not turn it into a puzzle.

anibalsanchez:
Thanks for all your feedback.

Until now, people have been happy with the current plugin and the legacy events implementation. I didn't want to change how the plugin itself has been working, and the original idea has been only to use the `SubscriberInterface` interface. However, I'm happy to go for a full review and clean the plugin code in a way that, in the future, the same practice can be applied to the rest of the plugins.

Before investing more time in the PR, can we confirm that it will be merged if we continue improving the code? Maybe @bembelimen can give us a hint?

HLeithner:
Me personally would like to get rid of all old events in 5.0 and replace them with the new subscriber interface version. Since we don't have a fallback from events written with the subscriber interface to the old "onXXX" functions we would need to introduce new event names.

Even if we don't remove the old event system in 5.0 we should rewrite all events to the subscriber interface including proper parameter validation.

Also I would be more strict with core plugins and mark them as `final`. May be @nikosdion can have a look at the events we  created in https://github.com/joomla/joomla-cms/pull/35396 and give some feedback if it's the supposed way you had in mind.

nikosdion:
@HLeithner Let me address everything.

# Renaming Events: NO!

You do NOT have to rename the events themselves. You just have to remove the legacy plugin support from CMSPlugin added by me and Michael. Its one and only job is to trawl through the public methods, find which ones start with `on` and register them as event handlers through a legacy wrapper. 

When you remove the legacy wrapper old plugins (replying on the legacy wrapper) will stop working; they will load but do nothing at all. New style plugins, implementing the SubscriberInterface, will work just fine.

Renaming the events themselves causes a massive migration issue: it is impossible to write an extension which is simultaneously compatible with Joomla 4 and 5. This means that it's impossible for users to upgrade to Joomla 5 unless they EITHER start from scratch OR remove all third party extensions losing all related data, i.e. start from scratch.

Furthermore, renaming the events makes two decades of knowledge and documentation invalid overnight. That's a good way to dissuade new developers to come to Joomla.

There's really no good reason to rename the events and there are many good reasons to keep the names the same.

# Creating Event classes: YES

That was the intention all along. I had created a few event classes for the Table events as a sampler and Proof of Concept — even more so since the scope of my original PR 6 years ago was to refactor Tags and other behaviors into “orthogonal” plugins, handling specific Events.

What you did for the Media Manager is on track with the original intention.

IMHO all core events such as onAfterInitialise, onContentPrepare (_especially_ this event; each instance of it is ridiculously inconsistent with every other instance!) and so on should have their own Event classes. This will make them more discoverable, self–documenting and also remove any ambiguities about their call arguments.

FWIW even plugins which currently expect a generic Event object will work as the more specific event classes still extend from the base Event class. In other words, introducing specific event classes should not affect b/c. It will just make core events discoverable and self–documenting.

Once the core starts doing that 3PDs can also follow suit if they feel so inclined (FWIW it is on my to–do list for my extensions).

# `final` in core plugin methods: ALTERNATIVE

Make the plugin class itself final, not the methods. Actually, thinking about it, it should be a documented best practice to make plugin classes final. These should not be extended.

The methods should simply be private.

Of course we cannot do any of that until 5.0. In the meantime (4.1 onwards) we can mark the classes as [`@internal`](https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.internal.pkg.html). This phpDocumentor tag indicates that the entire class is not part of the public API and may change drastically in any upcoming version. This means that drastic changes, like what this PR does, could happen in a patch release without violating semantic versioning.

In short, we will be treating core plugins as implementation details.

# Plugins need _more_ refactoring

The core plugins are for the most part legacy plugins. Those which are not, like what this PR here does, is a hybrid. They are not full Joomla 4 compatible extensions, they are legacy plugins implementing the SubscriberInterface.

Fully native plugins would have their own service provider and a different directory structure. This means that come Joomla 5 we can remove the legacy extension support from ExtensionManagerTrait. It will also mean that all core code — including plugins, modules and templates — will be auto–loaded through the PSR-4 autoloader.

# The extensions installed needs improving

There is one things that needs to be addressed, however: installing plugins. Right now, the extensions installer _expects_ the legacy filename `bar.php` for a plugin `plg_foo_bar`. The reason is that the plugin name is retrieved from the XML manifest through the special `<filename plugin="bar">bar.php</filename>` line. This leads to atrocities like what I had to do here: https://github.com/akeeba/sociallogin/tree/feature/joomla4/plugins/system/sociallogin Essentially, I have a useless empty plugin file which is _never loaded by Joomla_ just so that Joomla 4 can install the plugin following Joomla 4 conventions. Modules are the same deal, they need an empty legacy module file which is never loaded (since ExtensionManagerTrait will only use the legacy files if the `services/provider.php` file does not exist).

# While we are at this...

There are other legacy mannerisms in core code, like using `Factory::getUser()` to get the current user. Understandable since the application's Identity may not be populated. I would prefer it if the UserFactoryInterface had a getCurrentUser method to return the current user without having to write a dozen lines of code in my extensions, essentially replicating Factory::getUser.

Generally speaking, the only two valid use cases for using Factory should be getting the DI container or getting the application object (maybe also the mailer object, I can't remember off the top of my head). Everything else is already deprecated and should not be used in core code.

You can copy my entire reply to the Joomla 4 & 5 roadmaps. These are the low hanging fruit to migrate core code to no longer use deprecated stuff and eventually obsolete stuff in Joomla 5 already deprecated in Joomla 4 without causing a migration disaster 😁

anibalsanchez:
I opened the PR to help in the migration of the legacy stuff, but now it seems that we are again talking about the next major version.  **Joomla 5 is on the horizon ... (Nov 2021)**.

nikosdion:
Migrating the legacy stuff is intrinsically linked to the next major version. Joomla 4 deprecated stuff but Joomla 4 itself is using the deprecated stuff. What you do is a necessary step to get rid of deprecated legacy stuff. How much do we change this plugin? It depends on what we want to do with the next major version 😜 

I would personally go with a full Joomla 4 native plugin (with a service provider) so that we don't need to refactor the refactored code in the future and get rid of deprecated legacy in the next major version. If we are OK kicking the can on removing legacy — essentially removing support for how we were writing plugins in 2007 when J5 is released circa, dunno, 2027? — we can just add SubscriberInterface and call it a day.

You see why discussing Joomla 5 long before any code is written for it makes sense so early in the 4.x release cycle? It's all about legacy handling and making sure we don't screw up Joomla in the future.

anibalsanchez:
Okok. I'll give it a second pass to clean the hybrid approach and the plugin code.

HLeithner:
@nikosdion Thanks for your detailed description, it's very helpful. I have still some questions and comments.

> # Renaming Events: NO!
> You do NOT have to rename the events themselves. You just have to remove the legacy plugin support from CMSPlugin added by me and Michael. Its one and only job is to trawl through the public methods, find which ones start with `on` and register them as event handlers through a legacy wrapper.

Oh it seems I missed this while I looked the last time, so of course it doesn't make sense to change the names when we have a working b/c layer.

> Furthermore, renaming the events makes two decades of knowledge and documentation invalid overnight. That's a good way to dissuade new developers to come to Joomla.

Sadly I think the documentation would be outdated anyway and have to be reworked to match the subscriber interface, but of course it's easier to find the right event.

Thanks for pointing this out.

> # Creating Event classes: YES
> That was the intention all along. I had created a few event classes for the Table events as a sampler and Proof of Concept — even more so since the scope of my original PR 6 years ago was to refactor Tags and other behaviors into “orthogonal” plugins, handling specific Events.
>
> What you did for the Media Manager is on track with the original intention.
>
> IMHO all core events such as onAfterInitialise, onContentPrepare (_especially_ this event; each instance of it is ridiculously inconsistent with every other instance!) and so on should have their own Event classes. This will make them more discoverable, self–documenting and also remove any ambiguities about their call arguments.

What you mean exactly with "(_especially_ this event; each instance of it is ridiculously inconsistent with every other instance!)"? When we clean up/convert the event api it would make sense to do it right this time?

> Once the core starts doing that 3PDs can also follow suit if they feel so inclined (FWIW it is on my to–do list for my extensions).

I had a long talk with @bembelimen couple of weeks ago about this topic to convert all events to subscriber events until 4.2 hopefully we get some traction here to achieve this earlier. 

> # `final` in core plugin methods: ALTERNATIVE
> Make the plugin class itself final, not the methods. Actually, thinking about it, it should be a documented best practice to make plugin classes final. These should not be extended.

That's what I mean with my note. Mark things final that should be final, and I think plugins are nothing that what you want to extend. 

> The methods should simply be private.

Maybe I misunderstood this but the event listener functions need to be public.

>
> Of course we cannot do any of that until 5.0. In the meantime (4.1 onwards) we can mark the classes as [`@internal`](https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.internal.pkg.html). This phpDocumentor tag indicates that the entire class is not part of the public API and may change drastically in any upcoming version. This means that drastic changes, like what this PR does, could happen in a patch release without violating semantic versioning.

Actually we have no semver promise on plugins, and I don't see a reason to wait for 5.0 to change them to use the service provider interface and use namespaces. 

> # Plugins need _more_ refactoring
> The core plugins are for the most part legacy plugins. Those which are not, like what this PR here does, is a hybrid. They are not full Joomla 4 compatible extensions, they are legacy plugins implementing the SubscriberInterface.
>
> Fully native plugins would have their own service provider and a different directory structure. This means that come Joomla 5 we can remove the legacy extension support from ExtensionManagerTrait. It will also mean that all core code — including plugins, modules and templates — will be auto–loaded through the PSR-4 autoloader.

Yeap, that means also that all Events have to be changed to their own `EventAbstract` implementation

> # The extensions installed needs improving
> There is one things that needs to be addressed, however: installing plugins. Right now, the extensions installer _expects_ the legacy filename `bar.php` for a plugin `plg_foo_bar`. The reason is that the plugin name is retrieved from the XML manifest through the special `<filename plugin="bar">bar.php</filename>` line. This leads to atrocities like what I had to do here: https://github.com/akeeba/sociallogin/tree/feature/joomla4/plugins/system/sociallogin Essentially, I have a useless empty plugin file which is _never loaded by Joomla_ just so that Joomla 4 can install the plugin following Joomla 4 conventions. Modules are the same deal, they need an empty legacy module file which is never loaded (since ExtensionManagerTrait will only use the legacy files if the `services/provider.php` file does not exist).

Is this hard? It doesn't sound to be or do I miss something?

> # While we are at this...
> There are other legacy mannerisms in core code, like using `Factory::getUser()` to get the current user. Understandable since the application's Identity may not be populated. I would prefer it if the UserFactoryInterface had a getCurrentUser method to return the current user without having to write a dozen lines of code in my extensions, essentially replicating Factory::getUser.

Only had a quick look but shouldn't the `Application::getIdentity()` already be populated when events are triggered? Because it's loaded in the `CMSApplication::getInstance()` function using the (the deprecated) `Factory::getUser()` function...

> Generally speaking, the only two valid use cases for using Factory should be getting the DI container or getting the application object (maybe also the mailer object, I can't remember off the top of my head). Everything else is already deprecated and should not be used in core code.

Personally I would like to see joomla being able to run multiple times in the same process, but I think that will never happen.  

> You can copy my entire reply to the Joomla 4 & 5 roadmaps. These are the low hanging fruit to migrate core code to no longer use deprecated stuff and eventually obsolete stuff in Joomla 5 already deprecated in Joomla 4 without causing a migration disaster 😁

Will do or at least take bigger parts of it. Will see how it fits and who will do it.

One other thing. How should the validation work in the events if you look at the new media event's I added much validation into the setArgument functions because we don't have an interface for the "file-object". Is this a way to go if we don't have an interface and object that can validate the data itself. In this case something like MediafileInterface with getter and setter defined which do the validation?

nikosdion:
> What you mean exactly with "(especially this event; each instance of it is ridiculously inconsistent with every other instance!)"? When we clean up/convert the event api it would make sense to do it right this time?

We use that event everywhere, from populating fields to processing plugin codes (e.g. `{loadposition}`). The former can operate on any object. The latter expect that the object being passed has a `text` property without actually checking even though this is not a given. I think that the core plugins should be consistent in that regard and we should type the content as an object (there are still some legacy 3PD plugins expecting an array which of course breaks in PHP 8).

> I had a long talk with @bembelimen couple of weeks ago about this topic to convert all events to subscriber events until 4.2 hopefully we get some traction here to achieve this earlier.

Sounds very promising!

> Maybe I misunderstood this but the event listener functions need to be public.

You are correct. My brain farted; I was replying past 1 am my time 😅

> Actually we have no semver promise on plugins, and I don't see a reason to wait for 5.0 to change them to use the service provider interface and use namespaces.

Is this mentioned anywhere? Because if it's not there's the implicit promise that plugins follow SemVer like the rest of the CMS and its core extensions. FWIW core components do follow SemVer so it would be at least internally inconsistent to treat ONE kind of core extensions differently than the rest of the CMS... I would recommend at the very least a marketing campaign before the work even starts so that 3PDs are not caught off–guard in the off–chance they are using a core plugin in a weird way — and yes, _I have seen THINGS_...

> Yeap, that means also that all Events have to be changed to their own EventAbstract implementation

Precisely. That's what I meant.

> Is this hard? It doesn't sound to be or do I miss something?

Well, two hurdles. First, how do you convey the name of the plugin/module if not by an attribute to the filename? An attribute to the root `<extension>` element? Second hurdle, what happens if someone tries to use both methods at once e.g. to support Joomla 4.0 and a newer Joomla version? These are neither hard nor easy, these are decisions that need to be made and which will be binding for years to come.

> Only had a quick look but shouldn't the Application::getIdentity() already be populated when events are triggered? Because it's loaded in the CMSApplication::getInstance() function using the (the deprecated) Factory::getUser() function...

Not on CLI. Can't remember exact details, it was definitely after 4.0.1 that I was creating a CLI command that kept failing because the identity was null instead of a guest user.

> Personally I would like to see joomla being able to run multiple times in the same process, but I think that will never happen.

I'd rather limit the probability of a 3PD reenacting Inception on a server...

> One other thing. How should the validation work in the events if you look at the new media event's I added much validation into the setArgument functions because we don't have an interface for the "file-object". Is this a way to go if we don't have an interface and object that can validate the data itself. In this case something like MediafileInterface with getter and setter defined which do the validation?

The setters should do the validation. Also make sure you call the setters from the constructor. This keeps all checks straightforward.

Whether you should implement an interface or not... It depends and needs to be examined on a case by case basis. Unless you have a desperate need for an interface (as opposed to just making sure something is an object which has a specific method) go for it. If you want to leave the event more open–ended, e.g. like what you **need** for onContentPrepare, don't use an interface. I have obviously not used the media manager events myself that's why I am not handing down an opinion on them. I can tell you that most of the basic events for content and core application events shouldn't have their arguments tied to interfaces because it will break a lot of things in ways that cannot be unbroken. Don't go for pedantry over flexibility. Joomla's bread and butter is flexibility.

Yeah, I know, I am a pragmatist, not a purist. Come with the territory of publishing mass distributed software. Those who want purism can write their own apps from scratch using Symfony (I won't even say Laravel; purists will self–combust on contact with Laravel's façades).

HLeithner:
> We use that event everywhere, from populating fields to processing plugin codes (e.g. `{loadposition}`). The former can operate on any object. The latter expect that the object being passed has a `text` property without actually checking even though this is not a given. I think that the core plugins should be consistent in that regard and we should type the content as an object (there are still some legacy 3PD plugins expecting an array which of course breaks in PHP 8).

OK, now I understand what you mean.

> > Actually we have no semver promise on plugins, and I don't see a reason to wait for 5.0 to change them to use the service provider interface and use namespaces.
>
> Is this mentioned anywhere? Because if it's not there's the implicit promise that plugins follow SemVer like the rest of the CMS and its core extensions. FWIW core components do follow SemVer so it would be at least internally inconsistent to treat ONE kind of core extensions differently than the rest of the CMS... I would recommend at the very least a marketing campaign before the work even starts so that 3PDs are not caught off–guard in the off–chance they are using a core plugin in a weird way — and yes, _I have seen THINGS_...

Our b/c Promise is only for the /library folder, point 6.1.1 at https://developer.joomla.org/development-strategy.html#backward_compatibility

> All PHP code in the /libraries folder which is not flagged as private is considered to be part of the Joomla API and subject to backwards compatibility constraints. This may be extended to include other PHP code, such as component classes, in the future.

The fun part of this document is that javscript has a much harder b/c promise then the php part. (6.1.2 All JavaScript functions and classes that are not flagged as private.) 

Of course, allowed to break things doesn't mean we should do this.

> Well, two hurdles. First, how do you convey the name of the plugin/module if not by an attribute to the filename? An attribute to the root `<extension>` element? Second hurdle, what happens if someone tries to use both methods at once e.g. to support Joomla 4.0 and a newer Joomla version? These are neither hard nor easy, these are decisions that need to be made and which will be binding for years to come.

Supporting 4.0 and 5.0 shouldn't be a problem (even 3.x should be possible if the service provider is not loaded). Based on your [https://github.com/akeeba/sociallogin/blob/feature/joomla4/plugins/system/sociallogin/services/provider.php](Plugin) the loader doesn't need to now the plugin name because it the service provider has a defined name. Or we talk about diffrent topics.

> Not on CLI. Can't remember exact details, it was definitely after 4.0.1 that I was creating a CLI command that kept failing because the identity was null instead of a guest user.

Yeap I think I have saw this, the CLI application doesn't load a user, the question is should the cli application has an user, but maybe that's something for it's own discussion.

>
> > Personally I would like to see joomla being able to run multiple times in the same process, but I think that will never happen.
>
> I'd rather limit the probability of a 3PD reenacting Inception on a server...

Yeah it could be hard, I really like reactPHP and with Fibers in php 8.1 I think we can increase performance on this topic but nothing for a short or midterm goal.  

> The setters should do the validation. Also make sure you call the setters from the constructor. This keeps all checks straightforward.
>
> Whether you should implement an interface or not... It depends and needs to be examined on a case by case basis. Unless you have a desperate need for an interface (as opposed to just making sure something is an object which has a specific method) go for it. If you want to leave the event more open–ended, e.g. like what you **need** for onContentPrepare, don't use an interface. I have obviously not used the media manager events myself that's why I am not handing down an opinion on them. I can tell you that most of the basic events for content and core application events shouldn't have their arguments tied to interfaces because it will break a lot of things in ways that cannot be unbroken. Don't go for pedantry over flexibility. Joomla's bread and butter is flexibility.
>
> Yeah, I know, I am a pragmatist, not a purist. Come with the territory of publishing mass distributed software. Those who want purism can write their own apps from scratch using Symfony (I won't even say Laravel; purists will self–combust on contact with Laravel's façades).

Of course, breaking things just for the sake of consistency is a bad idea. Trying to force existing events into a corset also causes more problems than it solves.

Thanks for your feedback.

anibalsanchez:
I have re-organized the plugin according to the comments.

- Legacy methods removed `onAfterRender`, `onAfterRoute`, `onAfterRespond`
- The plugin class is final
- There are three helpers now to separate responsibilities: `PageCacheKeyGenerator`, `PageCacheStorage` and `PageCachingChecker`
- The plugin only has the minimum code to interact with Joomla and the Helpers are focused on a single responsibility and they don't produce side-effects.
- The `PageCacheKeyGenerator` is managed as a singleton in the container (let's use the container ;-))

One pending pain point is that the Helpers are not being autoloaded... even when the namespace is defined in the manifest. I still have to investigate this.

plugins/system/cache/cache.php
anibalsanchez: The deprecated methods are removed now.

plugins/system/cache/cache.php
anibalsanchez: Please, check the new implementation.

plugins/system/cache/cache.php
anibalsanchez: Constructor removed and `_cache` moved to a Helper class.

plugins/system/cache/cache.php
anibalsanchez: `trigger_error` removed.

plugins/system/cache/cache.php
anibalsanchez: Method renamed. Please, check the new method names now.

plugins/system/cache/cache.php
anibalsanchez: Fixed both issues. Check the new Helpers and the signatures updates (BTW, `PlgSystemHttpHeaders` has the same issue, a new issue is coming).

plugins/system/cache/cache.php
anibalsanchez: If we check it here, we avoid the plugin execution completely.

plugins/system/cache/cache.php
anibalsanchez: I hesitate to define the event callbacks in this way:

```php
			'onAfterRender' => 'onAfterRender',
			'onAfterRoute' => 'onAfterRoute',
			'onAfterRespond' => 'onAfterRespond',
```

It is a duplication of information and it looks like the Joomla 3 implementation.

If we assume that each callback has only one action/responsibility, then it is better to associate the event with the right action name.

plugins/system/cache/cache.php
anibalsanchez: Removed after the refactorization, please check it now.

plugins/system/cache/cache.php
anibalsanchez: Moved to the Helper. I kept the `static` to avoid the empty object Helper creation.

plugins/system/cache/cache.php
anibalsanchez: Removed. Txs. Please, check it now.

plugins/system/cache/cache.php
anibalsanchez: Reorganized. Now, it checks if there's a PageCache storage with the body.

Please, check it again.

plugins/system/cache/src/Helpers/PageCacheKeyGenerator.php
```diff
@@ -0,0 +1,61 @@
...
+		$this->key = md5(serialize($parts));
+	}
+
+	/**
+	 *
+	 *
+	 * @return  string
+	 *
+	 * @since    __DEPLOY_VERSION__
+	 */
```
PhilETaylor:
```suggestion
	/**
	 * @return  string
	 *
	 * @since   __DEPLOY_VERSION__
	 */
```

plugins/system/cache/src/Helpers/PageCacheStorage.php
```diff
@@ -0,0 +1,84 @@
...
+
+defined('_JEXEC') or die;
+
+use Joomla\CMS\Cache\Cache;
+use Joomla\CMS\Cache\Controller\PageController;
+
+/**
+ * Joomla! Page Cache Plugin - PageCacheStorage.
+ *
+ * @since  4.1
```
PhilETaylor:
```suggestion
 * @since  __DEPLOY_VERSION__
```

plugins/system/cache/src/Helpers/PageCacheStorage.php
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	/**
+	 * Store
+	 *
+	 * @param   object $body  		The page body to be stored.
+	 * @param   string $cacheKey	Cache Key
+	 *
+	 * @since  __DEPLOY_VERSION__
+	 */
```
PhilETaylor:
```suggestion
	/**
 	 * @param   object $body  	The page body to be stored.
	 * @param   string $cacheKey	Cache Key
	 *
	 * @since   __DEPLOY_VERSION__
	 */
```

plugins/system/cache/src/Helpers/PageCachingChecker.php
```diff
@@ -0,0 +1,140 @@
...
+
+use Joomla\CMS\Factory;
+use Joomla\CMS\Plugin\PluginHelper;
+use Joomla\CMS\Uri\Uri;
+use Joomla\Registry\Registry;
+
+/**
+ * Joomla! Page Cache Plugin - PageCachingChecker.
+ *
+ * @since  4.1
```
PhilETaylor:
```suggestion
 * @since  __DEPLOY_VERSION__
```

plugins/system/cache/src/Helpers/PageCachingChecker.php
```diff
@@ -0,0 +1,140 @@
...
+
+	/**
+	 * Check if the page is excluded from the cache or not.
+	 *
+	 * @param   string $cacheKey	Cache Key
+	 *
+	 * @return   boolean  True if the page is excluded else false
+	 *
+	 * @since    __DEPLOY_VERSION__
+	 */
```
PhilETaylor, line 84: All the docblocks in this PR need revisiting to fix spacing
https://developer.joomla.org/coding-standards/docblocks.html

plugins/system/cache/src/Helpers/PageCachingChecker.php
```diff
@@ -0,0 +1,140 @@
...
+			// Loop through each pattern.
+			if ($exclusions)
+			{
+				foreach ($exclusions as $exclusion)
+				{
+					// Make sure the exclusion has some content
+					if ($exclusion !== '')
+					{
+						// Test both external and internal URI
+						if (preg_match('#' . $exclusion . '#i', $cacheKey . ' ' . $internalUri, $match))
```
PhilETaylor, line 124: here we are FIVE levels down. Code Smell

PhilETaylor (changes requested):

plugins/system/cache/src/Helpers/PageCachingChecker.php
anibalsanchez: Yep. I have to active the Joomla code style ;-) Tomorrow. Txs

plugins/system/cache/src/Helpers/PageCachingChecker.php
anibalsanchez: I will clean it. Txs.

PhilETaylor:
Screenshot 2021-11-11 at 15 17 46

Also need to re-merge :)

plugins/system/cache/cache.php
```diff
@@ -48,208 +41,158 @@ class PlgSystemCache extends CMSPlugin
...
+	 * @since   __DEPLOY_VERSION__
 	 */
-	public function __construct(&$subject, $config)
+	public static function getSubscribedEvents(): array
 	{
-		parent::__construct($subject, $config);
+		$pageCacheKeyGenerator = new PageCacheKeyGenerator(Uri::getInstance());
+		$cacheKey = $pageCacheKeyGenerator->getKey();
+
+		if (!PageCachingChecker::canPageBeCached($cacheKey)) {
```
laoneo, line 63: I don't know how others see this but from my understanding, the events where the plugin is listen to should not be dependent on the environment. I would always return the same events and then do the check in the respective function. It looks for me too much load during initial phase. I would more go for the lazy load approach and do the check in the function itself. Like that you can then get rid of the static helper class and work on the plugin $app instance.

nikosdion: I fully agree with that and that's what I had commented on previously.

anibalsanchez: If the event subscription is not dynamic and the plugin is always instantiated and subscribed to the same methods, then it is the same as on Joomla 3. We are only moving around the same code. It is simpler to keep `onAfterRender`, `onAfterRoute`, and `onAfterRespond` following the old way of subscribing to the events.

If there is a static function `getSubscribedEvents` to subscribe the plugin to events, then it is meant to return an empty array or a set of events. In this case, there are conditions where the event set can be empty.

nikosdion: It's really not the same. Do you want me to convert this simple plugin over the weekend, do a new PR and discuss what was changed and why? I've been doing that for all my plugins, it's not a big deal for me to do that. I just don't have the time during the week.

HLeithner: In my understanding you you should never need to use Factory::getXYZ(); I know the application is not available in every context (also the CMSPlugin loads it from the Factory but should be possible to change to get it from the provided container).

Also I don't see any benefit to split the functions from the plugin it self. The Plugin class has just 250 lines of code splitting this into 3 helper classes just makes it slower but not more readable.

nikosdion: Yeah, I agree with you Harald. The concept is that we keep the plugin mostly the same but:
* Use a service provider, to showcase how true J4 native (non-legacy) plugins are meant to be written
* Use SubscriberInterface vs the legacy method which goes through reflection (it's faster with SubscriberInterface and more explicit; also shows that plugin events will be handling an Event object, not arbitrary argument lists)
* Remove all static crap
* Clean up dead code

I have practically rewritten this in my head, maybe if I have a bit of time tomorrow I'll show you what I mean.

The plugin _will_ be bigger (going through a service provider and SubscriberInterface adds some boilerplate) but since we no longer go through Reflection it's actually faster. Granted, a small plugin like that seems to get more complicated (it doesn't really, it's just boilerplate) but for bigger plugins e.g. the WebAuthn plugin it would make much more sense. It's just that the smallest piece of code is always a good way to start a bigger conversion project ;)

anibalsanchez:
Since you prefer a different way of implementing the plugin, I close the PR. Thanks for all your feedback.

Issue #34377, opened 2021-06-03
[4.0] Articles disappearing if length of text is too long


nozaro33:
### Steps to reproduce the issue

Can use the text here as reference: https://whiletrue.neocities.org/lte.html
Copy and paste twice into Text editor, and save.
Should get error message, but article is still saved to db, yet no longer viewable in Article Manager

### Expected result

Article should either not be saved with error, or should allow longer text

### Actual result

You get an error if it's in non-latin characters.
<img width="984" alt="Screen Shot 2021-06-03 at 20 46 57" src="https://user-images.githubusercontent.com/25219289/120649777-b6cad880-c4af-11eb-966b-afdc2dd30dbb.png">
or you may get this error if it's in English
<img width="1250" alt="Screen Shot 2021-06-03 at 20 53 53" src="https://user-images.githubusercontent.com/25219289/120649833-c4805e00-c4af-11eb-8a9f-9c374e9711a8.png">
(Finder token error will also show when I did the text in Joomla 3.9.27, but you can still see the article in article manager)

Articles will no longer show in the backend, but will still show normally in frontend. This issue persists regardless of language used.

### System information (as much as possible)

Joomla 4 RC 1
PHP 7.4.2
Mysql 5.7.26

### Additional comments
As you can see, ID 4,5,7 are no longer visible in backend. Please note that they still show up fine in frontend.
Note:  Even if you remove the "long text" from phpmyadmin, the article is still not visible from the backend Articles manager

<img width="964" alt="Screen Shot 2021-06-03 at 20 54 27" src="https://user-images.githubusercontent.com/25219289/120650263-2fca3000-c4b0-11eb-80cd-ffed6268ad77.png">
<img width="735" alt="Screen Shot 2021-06-03 at 20 55 49" src="https://user-images.githubusercontent.com/25219289/120650277-32c52080-c4b0-11eb-9324-f388ad184459.png">

nikosdion:
I cannot reproduce this with a. the text you specified; b. 150 paragraphs of Lorem Ipsum which actually end up with more words and total size than your example and c. randomly adding Emojis.

Reading your error messages, it doesn't look like the article text itself is the problem. I've created articles far longer than 75KB (your test text).

[...]

PR #35995, opened 2021-11-09
[4.1] Fixing "#__finder_tokens full" error

Hackwar:
### Summary of Changes
Smart Search/Finder has a long standing issue that the #__finder_tokens table can run full and abort the indexing process. That table is of type "MEMORY", which can run full and there is an option to reduce the number of tokens to store in that table, which unfortunately doesn't work in all cases. The MEMORY table is used since it is a lot faster than a normal InnoDB table and since this is just used for temporary storage, writing speed is the main factor here. Anyway, enough rambling. ;) This change pushes the code of `addTokenToDb` to the `tokenizeToDbShort()` method and properly counts the number of tokens written to DB. This means, that now the option mentioned above works properly.
This removes the `addTokenToDb` method, but since this is a protected method in a class of a component, this would still be in accordance to our backwards compatibility promise.

### Testing Instructions
- Get a very long text and insert it into an article.
- save it.

The indexing fails.

The indexing works.


richard67:
@Hackwar I have tested this PR but it doesn't change anything for me. I get the error with and without the PR applied when trying to save an article with some 100 or 200 times the text from the site linked in the description of issue #34377 , https://whiletrue.neocities.org/lte.html  .

bembelimen:
Just playing around with this thing: The problem of the whole thing is most likely that the number of the ```memory_table_limit``` parameter is too high + the check is on the wrong position.

So first the checks should be enough to have it that way: [patch.zip](https://github.com/joomla/joomla-cms/files/7591586/patch.zip)

And second either set the parameter to a lower number (works for me with 20.000) or take the number from https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_heap_table_size

Krshivam25:
error

administrator/components/com_finder/src/Indexer/Indexer.php
```diff
@@ -951,31 +917,31 @@ protected function addTokensToDb($tokens, $context = '')
...
-					. $db->quote($token->stem) . ', '
+					$query->quote($token->term) . ', '
+					. $query->quote($token->stem) . ', '
 					. (int) $token->common . ', '
 					. (int) $token->phrase . ', '
-					. $db->quote($token->weight) . ', '
+					. $query->quote($token->weight) . ', '
 					. (int) $context . ', '
-					. $db->quote($token->language)
+					. $query->quote($token->language)
```
HLeithner, line 926: What's the reason to make it slower then needed? please use the `quote` function of the database as before, the `quote` function in the Query object is only a proxy, which is an overhead.

bembelimen:
It seems, this will not make it in 4.1.0, so I moved it to 4.1.1 to not forget it, as it should be fixed asap.

Hackwar:
Closing this PR in favour of #36749

Issue #36006, opened 2021-11-10
Installer ComponentAdapter 


JProof:
### Steps to reproduce the issue

During component-development directly in joomla

* Remove from a installable component in the  `<administration> </administration>`
* Install the component by "discover"
* After discover the following screen
![img-2021-11-10-20-38-26](https://user-images.githubusercontent.com/12245927/141182181-a5d580ea-47fb-4e94-aef4-93a669636cc1.png)

### Expected result

Not installed should really not installed

### Actual result

The Component was installed/discovered into the Database
![img-2021-11-10-20-40-28](https://user-images.githubusercontent.com/12245927/141182319-3e1d485f-7e1d-4a5d-b29e-04827ba45a51.png)

The Problem is, that the component namespace does not work because component is HALF installed 

### System information (as much as possible)
Joomla! 4.0.4
every system 

### Additional comments

Breaks here:

![img-2021-11-10-20-46-48](https://user-images.githubusercontent.com/12245927/141183037-a0b9eb99-a712-46e4-8ce2-55b4f60bf3ab.png)

After Breaking final installation tasks will not finished. So the the component-namespace was not "registered" somewhere 

![img-2021-11-10-20-47-17](https://user-images.githubusercontent.com/12245927/141183065-fb28156a-7591-40b1-886d-03fdf9ab2b85.png)

EDIT:
In addition: is it possible or better for development during discover to re-create / handle the namespace-cache generation with the PlgExtensionNamespacemap???

dgrammatiko:
This is a known issue due to some limitations from the J3 code: https://github.com/mbabker/com_empty/blob/4a199f19b2d7ab240b6092057b7edbff666818e9/com_empty/empty.xml#L18-L20

alikon:
funny

> Sad panda

dgrammatiko:
@alikon I think that the check for the administrator in the manifest should be done much earlier and bail out without doing anything more (now it copies files, etc). So, I guess this is still a valid issue

JProof:
Okay!

By the way, the `<administrator><menu>...` also never works on discover, perhaps that is the right time to change this "ugly" behaviours?

dgrammatiko:
> perhaps that is the right time to change this "ugly" behaviours?

You're welcome to contribute some code

alikon:
yes for sure, @JProof please submit a pr


i've always thinked that
`<administration>
	</administration>`

was mandatory

anyway happy to be wrong :smile:

[...]

PR #36014, opened 2021-11-11
[4.0][bugfix] Fix half broken installation

dgrammatiko:
Pull Request for Issue #36006

### Summary of Changes

Change the order of code execution so the installer bails out early without committing any file writes

### Testing Instructions

Check the issue

No

### DIFF
--- a/libraries/src/Installer/Adapter/ComponentAdapter.php
+++ b/libraries/src/Installer/Adapter/ComponentAdapter.php
@@ -797,6 +797,12 @@ class ComponentAdapter extends InstallerAdapter
 	 */
 	protected function setupInstallPaths()
 	{
+		// Make sure that we have an admin element
+		if (!$this->getManifest()->administration)
+		{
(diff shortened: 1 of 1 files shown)


libraries/src/Installer/Adapter/ComponentAdapter.php
```diff
@@ -797,19 +797,21 @@ protected function removeExtensionFiles()
 	 */
 	protected function setupInstallPaths()
 	{
+		// Make sure that we have an admin element
+		if (!$this->getManifest()->administration)
+		{
+			throw new \RuntimeException(Text::_('JLIB_INSTALLER_ERROR_COMP_INSTALL_ADMIN_ELEMENT'));
+
+			return;
```
richard67: Return after throwing an exception? What is that good for?

dgrammatiko: I have no clue, too many beers (?)

dgrammatiko:
I think this PR solves nothing

Issue #30444, opened 2020-08-21
Wrong version on css and js files in template... results in cache headache


hans2103:
### Steps to reproduce the issue

View source on a clean J3 / J4 installation and you will see scripts and stylesheets in the head.
All css and js files are extended with a string of numbers and letters. 
They are meant for cache invalidation... and it should work if they would change when the file is changed. 
They don't change when the file is changed. 

for example in J3
```
<link href="/templates/protostar/css/template.css?dacd9129dca0f2ae311fd26072498696" rel="stylesheet" />
```

This css file is set in templates/protostar/index.php line 41
```
// Add Stylesheets
JHtml::_('stylesheet', 'template.css', array('version' => 'auto', 'relative' => true));
```

The version = auto causes the extension of the filename
`version = joomla-rocks` will cause the following output

```
<link href="/templates/protostar/css/template.css?joomla-rocks" rel="stylesheet" />
```

The link elements are set in https://github.com/joomla/joomla-cms/blob/staging/libraries/src/Document/Renderer/Html/HeadRenderer.php
The so-called `$mediaversion` comes from https://github.com/joomla/cassiopeia/blob/628f6fef933f46909131c389e4a4f064c01fd6ca/libraries/src/Version.php#L245
It is an md5 hash of the long Joomla version together with the secret from configuration.php plus date. 
The mediaversion will be set when not present or when the Joomla version changes or when in DEBUG mode. 

In J4 the link element is set via https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Document/Renderer/Html/StylesRenderer.php
And uses the same `$mediaversion` from libraries/src/Version.php

It would be nice if this can be rewritten to something that looks at the `filemtime` of the loaded asset.

```
class TemplateHelper
{
    /**
     * Load CSS
     */
    static public function loadCss()
    {
        HTMLHelper::_('stylesheet', 'style.css', ['version' => self::getFileMTime(JPATH_THEMES . '/' . self::template(). '/css/style.css'), 'relative' => true]);
    }

    static public function getFileMTime($file)
    {
        if (!file_exists($file) || filesize($file) == 0)
        {
            return null;
        }

        return filemtime($file);
    }
}
```

should the function getMediaVersion.php from libraries/src/version.php be rewritten?

### Expected result

I am expecting that when I change a css / js file the version will change as well. 

### Actual result

When changing a css / js file in the template the version does not change... this causes me headache when certain css changes are not to be seen. :-)

[...]

PR #36018, opened 2021-11-12
[4.1] Do not set a media version in debug mode on assets

laoneo:
Pull Request for comment https://github.com/joomla/joomla-cms/issues/30444#issuecomment-709488213

### Summary of Changes
When the site runs in debug mode, then on every page load a new media version is added the the assets. This is problematic while debugging Javascript in a browser. Like that the browser always assumes that it is a new script and removes the breakpoints (at least this is the case in Firefox). If you do JS debugging, you know what I'm talking about ;-)

This behavior is a relict from the early days of Joomla (#1618) to simplify development. But nowadays every developer knows how to load the page with `Ctrl+F5` to refetch the assets.

CC: @dgrammatiko, @ceford 

### Testing Instructions
Open any page in debug mode.
Open the dev tools of the browser and check the loaded JS and CSS files.

Assets are always loaded with a new media version after the path.

Assets are loaded without any media version.


laoneo:
@phproberto maybe you want to leave feedback as you was the author of the initial pr.

dgrammatiko:
If devs are using correctly the Web Assets this shouldn't be a problem. Let me explain:
- Joomla's own tools (that also devs could use for their own code) has a versioning script: https://github.com/joomla/joomla-cms/blob/4.0-dev/build/build-modules-js/versioning.es6.js. This script will loop all the `joomla.assets.json` files and calculate each asset's version based on the content. The hash/version is only changed if the dev re-executes the versioning script.
- Of course, HTMLHelper is the old guard and uses the media version that is updated per joomla update, so I guess this patch is covering that case.

I can't but just recommend every dev to endorse the web assets and also try to use, or at least be inspired by, the project's own tooling...

libraries/src/Version.php
```diff
@@ -244,6 +244,11 @@ public function generateMediaVersion(): string
 	 */
 	public function getMediaVersion(): string
 	{
+		if (JDEBUG)
+		{
+			return '';
```
dgrammatiko, line 249: Instead of returning an empty string you can MD5 the current day, something like 

```php

return md5(date('Y-m-d'));

```

Edge case: someone is debugging before and after midnight, the only one time that the hash will change per day

HLeithner:
> If devs are using correctly the Web Assets this shouldn't be a problem. Let me explain:
> 
> * Joomla's own tools (that also devs could use for their own code) has a versioning script: https://github.com/joomla/joomla-cms/blob/4.0-dev/build/build-modules-js/versioning.es6.js. This script will loop all the `joomla.assets.json` files and calculate each asset's version based on the content. The hash/version is only changed if the dev re-executes the versioning script.
> * Of course, HTMLHelper is the old guard and uses the media version that is updated per joomla update, so I guess this patch is covering that case.
> 
> I can't but just recommend every dev to endorse the web assets and also try to use, or at least be inspired by, the project's own tooling...

2 problems with this script.
1. it only runs when I want it to run
2. doesn't work if you have to debug a live/site on a customer server.

I know that's not the way to do it but it's reality that this happens.

I would prefer to take do this update on include time when the file is requests (check modify date and content).

It's also not possible in the WA manager to set a special version for a file from the template.

something I liked in dev process to simply add a '?v='.rand() to all css und js assets.

Also I think in debug mode we should also set the "version" string to something less static (like the hash of the content)

sorry allon for the partly off-topic comment

dgrammatiko:
@HLeithner I have to disagree on both problems

> it only runs when I want it to run

The hash derives from the file content. It shouldn't change with some magic PHP functionality per call. So, for me this is the correct way

> doesn't work if you have to debug a live/site on a customer server.

Actually, since PHP is totally unaware of the hash (is passed from the JSON file) you can debug the file as you're pleased. The hash WON'T change (unless you change it). 

> I would prefer to take do this update on include time when the file is requests (check modify date and content).

No, not in runtime, you're adding too much processing and I/O for something that can be way simpler. How simpler, something like: https://invalidate-cache.dgrammatiko.dev . What I mean is that the functionality of that module should be baked in into the template file editor or anywhere that a user could change the contents of a static file. In short, try to do as little as possible on runtime, precalculate and cache everything you can (at least that's my approach to anything server side)

HLeithner:
I only talk of debug mode, not afterwards. Like a option/switch you can give the asset manager `$wa->setDebug(true);` which magically disable caching (add cache busters).

of course I wouldn't do this in normal operation.

dgrammatiko:
> Like a option/switch you can give the asset manager $wa->setDebug(true);

The switch already exists, the `JDEBUG` constant. Changing the logic to not append the media version if the `JDEBUG == true` should be straight forward. That said wouldn't this change be simpler if this https://github.com/joomla/joomla-cms/blob/92ae4172ab4a7e6de6bb2516a9e15ecbd166c678/libraries/src/Document/Renderer/Html/ScriptsRenderer.php#L186-L190 was altered?

HLeithner:
@dgrammatiko actually yes I think changeing the script renderer to use the JDEBUG constant would be better.
@laoneo what do you think?

laoneo:
With my approach you save a DB insert operation ;-)

Fedik:
> When the site runs in debug mode, then on every page load a new media version is added the the assets.

I think it was made to be sure that hash will be new when you **turn off** debug (one of reasons). 
Kind of hash reset thing

laoneo:
As you can see in the original pr, this was not really the intention, more a side effect :-)

libraries/src/Version.php
```diff
@@ -253,7 +258,7 @@ public function getMediaVersion(): string
 			$mediaVersion = LibraryHelper::getParams('joomla')->get('mediaversion', '');
 
 			// Refresh assets in debug mode or when the media version is not set
```
Quy: Update comment?

libraries/src/Version.php
laoneo: Why a date? What is the benefit?

laoneo:
@Quy good spot

libraries/src/Version.php
dgrammatiko: Your code will produce something like: `script.is?`
With this change will produce something like:
`script.is?a324117227635b14` But the version is now stable, won’t change per call

Issue #36016, opened 2021-11-12
Remove hard-limit of 20 to the Module Tags-Popular list size


pwireless:
### Steps to reproduce the issue

Create new Module Tags-Popular. It's not allowed to show more than 20 tags.

### Expected result

Show more than 20 tags. Have a high limit e.g. 500.

alikon:
you can change the default max here https://github.com/joomla/joomla-cms/blob/4.0-dev/modules/mod_tags_popular/mod_tags_popular.xml#L43

brianteeman:
It wouldnt be "popular" tags if there were 500

chmst:
In other modules (polular articles, most read articles) we use type=number and leave it to the user how many itemshe wants to see.

brianteeman:
It has had a maximum of 20 since  Elin introduced the code 9 years ago - just saying

HLeithner:
I see no technical reason for the limit. if you write a pull request to remove this restriction I would support it.

chmst:
Please test #36019

richard67:
@chmst Why reopened? Isn't there a pull request which claims to fix it?

Quy:
@richard67 it was reopened before #36261.

richard67:
Ah … ok.

Closing as having a pull request. Please test #36261 . Thanks in advance.

PR #36019, opened 2021-11-12
[4.1] Tags module limit

chmst:
Pull Request for Issue #36016.

### Summary of Changes
As title says. For both tag modules. 

### Testing Instructions
Use more than 20 tags to yor articles, 
Set the maximum to a number > 20 and make sure that the tags are displayed corrctly in frontend

see #36016

Any number is possible

### DIFF
--- a/modules/mod_tags_popular/mod_tags_popular.xml
+++ b/modules/mod_tags_popular/mod_tags_popular.xml
@@ -39,6 +39,7 @@
 					label="MOD_TAGS_POPULAR_MAX_LABEL"
 					default="5"
 					filter="integer"
+					min="0"
 				/>
 
 				<field
--- a/modules/mod_tags_popular/mod_tags_popular.xml
+++ b/modules/mod_tags_popular/mod_tags_popular.xml
@@ -35,13 +35,10 @@
 
 				<field
 					name="maximum"
-					type="integer"
+					type="number"
 					label="MOD_TAGS_POPULAR_MAX_LABEL"
 					default="5"
 					filter="integer"
-					first="1"
-					last="20"
-					step="1"
 				/>
 
 				<field
--- a/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
+++ b/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
@@ -38,7 +38,7 @@ abstract class TagsPopularHelper
 		$user        = Factory::getUser();
 		$groups      = $user->getAuthorisedViewLevels();
 		$timeframe   = $params->get('timeframe', 'alltime');
-		$maximum     = $params->get('maximum', 5);
+		$maximum     = (int) $params->get('maximum', 5);
 		$order_value = $params->get('order_value', 'title');
 		$nowDate     = Factory::getDate()->toSql();
 		$nullDate    = $db->getNullDate();
@@ -157,7 +157,7 @@ abstract class TagsPopularHelper
 			}
 		}
 
-		if ($params->get('maximum','5') > 0)
+		if ($maximum > 0)
 		{
 			$query->setLimit((int) $params->get('maximum'));
 		}
--- a/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
+++ b/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
@@ -127,7 +127,6 @@ abstract class TagsPopularHelper
 				// Backup bound parameters array of the original query
 				$bounded = $query->getBounded();
 
-				$query->setLimit($maximum);
 				$query->order($db->quoteName('count') . ' DESC');
 				$equery = $db->getQuery(true)
 					->select(
@@ -158,7 +157,11 @@ abstract class TagsPopularHelper
 			}
 		}
 
-		$query->setLimit($maximum, 0);
+		if ($params->get('maximum','5') > 0)
+		{
+			$query->setLimit((int) $params->get('maximum'));
+		}
+
 		$db->setQuery($query);
 
 		try
--- a/modules/mod_tags_similar/mod_tags_similar.xml
+++ b/modules/mod_tags_similar/mod_tags_similar.xml
@@ -29,6 +29,7 @@
 					label="MOD_TAGS_SIMILAR_MAX_LABEL"
 					default="5"
 					filter="integer"
+					min="0"
 				/>
 
 				<field
--- a/modules/mod_tags_similar/mod_tags_similar.xml
+++ b/modules/mod_tags_similar/mod_tags_similar.xml
@@ -25,13 +25,10 @@
 			<fieldset name="basic">
 				<field
 					name="maximum"
-					type="integer"
+					type="number"
 					label="MOD_TAGS_SIMILAR_MAX_LABEL"
 					default="5"
 					filter="integer"
-					first="1"
-					last="20"
-					step="1"
 				/>
 
 				<field
--- a/modules/mod_tags_similar/src/Helper/TagsSimilarHelper.php
+++ b/modules/mod_tags_similar/src/Helper/TagsSimilarHelper.php
@@ -194,7 +194,7 @@ abstract class TagsSimilarHelper
 			$query->order($query->rand());
 		}
 
-		if ($params->get('maximum','5') > 0)
+		if ((int) $params->get('maximum', 5) > 0)
 		{
 			$query->setLimit((int) $params->get('maximum'));
 		}
--- a/modules/mod_tags_similar/src/Helper/TagsSimilarHelper.php
+++ b/modules/mod_tags_similar/src/Helper/TagsSimilarHelper.php
@@ -194,7 +194,11 @@ abstract class TagsSimilarHelper
 			$query->order($query->rand());
 		}
 
-		$query->setLimit((int) $params->get('maximum', 5));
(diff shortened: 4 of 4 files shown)


HLeithner:
if you do it this way you have to modify the php code to allow "0" as no limit option.

Shubhamverma2796:
Before: 
screen shot 2021-11-12 at 14 24 55

Clearly visible that the maximum count is 20

After:

screen shot 2021-11-12 at 14 28 02

Now user can set the value of the maximum count.

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36019">issues.joomla.org/tracker/joomla-cms/36019</a>.</sub>

Shubhamverma2796:
I have tested this item :white_check_mark: successfully on 38eb9e9e927a202c9b8aac40a4f547e40cf2ccd8<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36019">issues.joomla.org/tracker/joomla-cms/36019</a>.</sub>

BertaOctech:
I have tested this item :white_check_mark: successfully on 38eb9e9e927a202c9b8aac40a4f547e40cf2ccd8<br /><br />I followed the test instructions and the patch worked as described.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36019">issues.joomla.org/tracker/joomla-cms/36019</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36019">issues.joomla.org/tracker/joomla-cms/36019</a>.</sub>

brianteeman:
@HLeithner requested valid changes. please remove RTC

alikon:
oops missed that

chmst:
@HLeithner  default is 5 as in other modules. The value 0 is not really useful, of course, as the result then would be empty, but for mysql this is not a  problem. 
If I change it here ( 0 = all), we must change it in all other modules too.

HLeithner:
> @HLeithner default is 5 as in other modules. The value 0 is not really useful, of course, as the result then would be empty, but for mysql this is not a problem. If I change it here ( 0 = all), we must change it in all other modules too.

That's the reason I requested the implementation for this case in php (check for 0 and remove the setlimit call)

chmst:
Made the changes as requested by @HLeithner . Only if the input is > 0, the maximum number is set as limit. Could you please re-test? @BertaOctech  @Shubhamverma2796 ? Thanks!

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36019">issues.joomla.org/tracker/joomla-cms/36019</a>.</sub>

modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
```diff
@@ -158,7 +157,11 @@ public static function getList(&$params)
 			}
 		}
 
-		$query->setLimit($maximum, 0);
+		if ($params->get('maximum','5') > 0)
```
HLeithner:
```suggestion
		if ((int) $params->get('maximum', 5) > 0)
```

modules/mod_tags_similar/src/Helper/TagsSimilarHelper.php
```diff
@@ -194,7 +194,11 @@ public static function getList(&$params)
 			$query->order($query->rand());
 		}
 
-		$query->setLimit((int) $params->get('maximum', 5));
+		if ($params->get('maximum','5') > 0)
```
HLeithner:
```suggestion
		if ((int) $params->get('maximum', 5) > 0)
```

HLeithner (changes requested):

modules/mod_tags_similar/src/Helper/TagsSimilarHelper.php
richard67, line 197:
```suggestion
		if ($params->get('maximum', '5') > 0)
```
Fix PHP CS for drone.

modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
richard67: Why not use $maximum here? It is set at the beginning here https://github.com/joomla/joomla-cms/blob/14166af1d561ef33887d0ce6aebea01b854b52e9/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php#L41 and now unused after these changes.

We could change that line 41 to `$maximum = (int) $params->get('maximum', 5);`, i.e. add the type cast to integer, and then here and below use `if ($maximum > 0)`, `$query->setLimit($maximum);` and so on.

richard67: Or do as suggested by Harald and remove line 41.

HLeithner: of course use $maximum from line 41 (cast it to int first please) and for get this suggestion

Shubhamverma2796:
@chmst I applied the path and here are the results:

Screenshot (272)

As you can clearly see the maximum tag value can be set as negative too.  I think you should set the lower limit to  0 because obviously there is no point in setting -ve values.

modules/mod_tags_popular/mod_tags_popular.xml
```diff
@@ -35,13 +35,10 @@
 
 				<field
 					name="maximum"
-					type="integer"
+					type="number"
 					label="MOD_TAGS_POPULAR_MAX_LABEL"
 					default="5"
 					filter="integer"
```
HLeithner, line 41:
```suggestion
					filter="integer"
					min="0"
```

modules/mod_tags_similar/mod_tags_similar.xml
```diff
@@ -25,13 +25,10 @@
 			<fieldset name="basic">
 				<field
 					name="maximum"
-					type="integer"
+					type="number"
 					label="MOD_TAGS_SIMILAR_MAX_LABEL"
 					default="5"
 					filter="integer"
```
HLeithner, line 31:
```suggestion
					filter="integer"
					min="0"
```

HLeithner (changes requested):
Add a Minimum auf 0 items

modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
richard67: @chmst This one here is not done yet.

chmst: done

chmst:
The min attribute in the number field is not validated. Also with min=0 it is possible to enter every negative value. This is not in scope of this PR.

modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
```diff
@@ -158,7 +157,11 @@ public static function getList(&$params)
 			}
 		}
 
-		$query->setLimit($maximum, 0);
+		if ($maximum > 0)
+		{
+			$query->setLimit((int) $params->get('maximum'));
```
richard67, line 162:
```suggestion
			$query->setLimit($maximum);
```

richard67:
@chmst Why closed?

chmst:
@Shubhamverma2796 would you like to open an issue for the bug you found  in joomla? https://github.com/joomla/joomla-cms/pull/36019#issuecomment-968291979
Maybe you also find a solution ;)

Shubhamverma2796:
@chmst  How can I open an issue because this PR is closed before getting merged, and the hard limit is still 20. Please correct me if I'm wrong.

chmst:
The hard limit is independent from this. You found out, that it is possible to enter a negative value into a field of type="number" , also when the attribute min="0" is set. The field does not validate the input. 

You can open a new issue here 
image

and write a bug report. If you want ;)

Shubhamverma2796:
https://user-images.githubusercontent.com/54266411/142622699-797fae8a-e78c-41d2-9019-63fd18c862c2.mp4

@chmst  In the above video, you can see that I've fixed the issue; whenever the user tries to input any -ve value, the value automatically gets changed to 0. In this way, we can easily prevent the user from entering any negative value. 
The problem is I am new here and not well-versed with the PHP and code structure of Joomla. Therefore, I cannot figure out where(in which file or folder) I should put the function written by me so that everything works fine. If anyone can guide me, then it would be my pleasure. 
Thanks 

[Link to the function](https://jsfiddle.net/4qju780h/)

brianteeman:
@Shubhamverma2796 javascript is not the solution

brianteeman:
> The min attribute in the number field is not validated. Also with min=0 it is possible to enter every negative value. This is not in scope of this PR.

Just add validate="number" and this will ensure that the min=0 is validated

Shubhamverma2796:
@chmst  should I open a new pr for #36016 or will you re-open this pr itself?

PR #36042, opened 2021-11-14
Refactor the page cache plugin

nikosdion:
Pull Request to replace #35986 

Ping @HLeithner @anibalsanchez @laoneo 

### Summary of Changes

Refactored the system/cache plugin into the new Joomla 4 style.

### Discussion

Please test. I did not have the time to do any decent testing before going to bed; I need to wake up early tomorrow to get the kid to school. Sorry.

I am using the SubscriberInterface to register the three plugin events. ~The two events which add the page to the cache are rigged to always run _last_, otherwise the plugin wouldn't work right.~

Each event checks whether the app is in a state that supports serving cached pages or storing pages into the cache:
* Must be the Site application
* Must be a GET request
* Must be a guest user
* No application messages are to be displayed

The first two checks are cached on first use for a faster false return since these conditions don't change throughout the request. The latter two checks are performed each time the first two checks pass because a user may be logged in and application messages may be generated throughout the request lifetime.

I refactored the page exclusion checks to use a bit more modern PHP code.

The cache controller is not created until it's needed. This saves some initialization time when visiting pages which shouldn't be cached or served from the cache as this means the cache controller would never get to be initialized.

The external (possible SEF) URL to the page is not retrieved until it's needed instead of in the constructor to save a few microseconds there too.


plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,315 @@
...
+		/**
+		 * Note that onAfterRender and onAfterRespond must be the last handlers to run for this
+		 * plugin to operate as expected. These handlers put pages into cache. We must make sure
+		 * that a. the page SHOULD be cached and b. we are caching the complete page, as it's
+		 * output to the browser.
+		 */
+		return [
+			'onAfterRoute'   => 'onAfterRoute',
+			'onAfterRender'  => ['onAfterRender', PHP_INT_MAX],
+			'onAfterRespond' => ['onAfterRespond', PHP_INT_MAX],
```
HLeithner: I think we should not fix stupid people. that wouldn't work. If someone thinks his/her plugin should be the last and the core have to be ignored you will loose anyway. So we should setting a good example and use the proper value.

Btw you used the wrong direction of priority. Since it should be the last one which should be executed the priority have to be lowerest in this case `Priority::MIN` (-3), you selected the maximum priority.

plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,315 @@
...
+use Joomla\CMS\Uri\Uri;
+use Joomla\Event\Event;
+use Joomla\Event\SubscriberInterface;
+
+/**
+ * Joomla! Page Cache Plugin.
+ *
+ * @since  1.5
+ */
+class Cache extends CMSPlugin implements SubscriberInterface
```
HLeithner: Do you want to make the plugin `final`?

HLeithner (commented):
Thanks for this PR, I just did a quick look at the code and found some questions^^.

If we merge this we also should find a way that 3rd party plugins can also be installed without the <pluginname>.php but that's another story.

I will test it later this week.

plugins/system/cache/src/Extension/Cache.php
nikosdion: No problem, I'm removing that.

plugins/system/cache/src/Extension/Cache.php
nikosdion: Right! I forgot. And fixed.

nikosdion:
@HLeithner I forgot to mention this. Over the course of last week @laoneo found a way to make the plugins installable without any change in core code! He moved the `plugin` attribute from `<file>` to `<folder>` and Joomla 4.0 handled it beautifully. We confirmed his change on our own company's extensions. That's what I did in the XML manifest.

plugins/system/cache/src/Extension/Cache.php
HLeithner: Thanks but I think we need the priority `\Joomla\Event\Priority::LOW` to solve the issue when multiple plugins are get executed on the 2 events. Then a plugin that really needs/wants to be executed after the cache can still set it self to `\Joomla\Event\Priority::MIN`

HLeithner:
> @HLeithner I forgot to mention this. Over the course of last week @laoneo found a way to make the plugins installable without any change in core code! He moved the `plugin` attribute from `<file>` to `<folder>` and Joomla 4.0 handled it beautifully. We confirmed his change on our own company's extensions. That's what I did in the XML manifest.

perfect one problem less ;-)

plugins/system/cache/src/Extension/Cache.php
nikosdion: OK, I just made the commit to do that.

laoneo:
> moved the `plugin` attribute from `<file>` to `<folder>` and Joomla 4.0 handled it beautifully. We confirmed his change on our own company's extensions. That's what I did in the XML manifest.

Actually you can do this already since 3.4 https://github.com/joomla/joomla-cms/commit/f90ff6033c80df19ff2bdc22400d5c2913c82b47#diff-f1347eb6172ab8188fa384c60a35be15af08dc341072aad5e842e912679234de

plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,316 @@
...
+		// Set the page content from the cache and output it to the browser.
+		$this->app->setBody($data);
+
+		echo $this->app->toString((bool) $this->app->get('gzip'));
+
+		// Mark afterCache in debug and run debug onAfterRespond events, e.g. show Joomla Debug Console if debug is active.
+		if (JDEBUG)
+		{
+			// Create a document instance and load it into the application.
+			$document = Factory::getContainer()
```
laoneo: I would inject the document factory.

plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,316 @@
...
+
+		// Set the cache options.
+		$options = [
+			'defaultgroup' => 'page',
+			'browsercache' => $this->params->get('browsercache', 0),
+			'caching'      => false,
+		];
+
+		// Instantiate cache with previous options.
+		$this->cache = CMSCache::getInstance('page', $options);
```
laoneo: I would inject the cache controller factory and build the cache with that one.

laoneo:
@HLeithner any comments for triggerEvent usage here? Otherwise I give it a test and get it merged. Afterwards we can discus if we want to migrate more plugins as suggested by @nikosdion.

nikosdion:
@laoneo @HLeithner IMHO the triggerEvent helper method shouldn't disappear forever. I have written a number of plugins which do not use it and my takeaway from this is that the alternative is _clunky_. 

What I'd like to do is create a new trait where event creation and parsing the event results are separate methods. Think of calling something like `$resultsArray = $this->getEventResults($this->createEvent('onFoobar', $argumentsArray, CustomEventClass::class));` where the arguments array could be associative (for named arguments) and the CustomEventClass::class could be omitted to use the generic `\Joomla\Event\Event` class.

Otherwise every time we call an event we need to use the same boilerplate code:
```php
$result = $this->getDispatcher()->dispatch($eventName, $new Event($eventName, $argumentsArray));
$actualResult = !isset($result['result']) || \is_null($result['result']) ? [] : $result['result'];
```

I would argue that the boilerplate code looks clunky and inscrutable to newcomers. We don't want to dissuade new developers so a couple method calls may be better than lab clean code. What do you think?

laoneo:
> Otherwise every time we call an event we need to use the same boilerplate code:
> 
> ```
> $result = $this->getDispatcher()->dispatch($eventName, $new Event($eventName, $argumentsArray));
> $actualResult = !isset($result['result']) || \is_null($result['result']) ? [] : $result['result'];
> ```

This is true for J3 events for BC reasons. But for new ones you can use the argument from the event as result as we did in #35396.

In this pr here you can just replace `$this->app->triggerEvent()` with `$this->app->getDispatcher()->dispatch()` as there are no arguments and results to work with.

HLeithner:
Since the events are not new I wouldn't require a proper event object. But we should get rid of all events without it own class as already discussed.

It would be a great time/pr to do this as part of the refactoring.

As allon already said for new event we can guarantee that $result['result'] exists (and of course should).

If you don't like to do it now or in a future pr, I'm fine with it but as it's also an example PR it would be great if can set a milestone how to do it for events so a refer to this single pr is easier then having a couple of PRs to refer to.

nikosdion:
@HLeithner Just to be clear: you'd like me to create Event classes for all the events referenced in this plugin and use them _without_ going through triggerEvent. Say yes and I'll start working on it tonight.

HLeithner:
@nikosdion yes when I'm not wrong we discussed this in another Issue that we want to move all legacy events to the "new" event system. And iirc this is possible in a fully b/c way right?

nikosdion:
I thought it was a 5.0 goal 😛 You are right that it can be done b/c _as long as_ the event handlers parse the event arguments positionally instead of by name i.e. `[$arg1, $arg2, ...., $argN] = $event->getArguments()` **AND** use the generic \Joomla\Event\Event class for their parameter type hint (or use legacy Joomla 3 event handlers). To make things easier, the docblock should also or–type the specific Event subclass we will be expecting in the future.

Throughout the 4.x release cycle this is how event handlers should be written. Moreover, throughout the 4.x release _we need to VERY CLEARLY document these specific event classes, which version they appeared and instruct developers to use them_.

When we start working on 5.0 we can type hint with the correct event subclass and use its named arguments.

PS: About the “clearly document” we can also create a trait which generates the correct event class based on the event name. For example, if Joomla has a core event onFoobarScream which should be conveyed through an event object of the class `\Joomla\Event\Foobar\Scream` our Trait's makeEvent method would return a new `\Joomla\Event\Foobar\Scream` object with the correct arguments. This would be `@deprecated 6.0` and would act as the b/c layer for the events system come Joomla 5.0.

Should I give it a stab? Now that I've worked with events on real world stuff, not just the lab environment I was using back in 2015, I have a much better ‘feel’ for how to do things in a way that doesn't screw the core over in the long term and doesn't screw third party developers over in the short to medium term.

HLeithner:
I would like to remove the old event system in j5.0 already and to have all Events have there correct classes already in 4.x ready and used. Maybe the validation have to be more relaxed with deprecation warnings if wrong values are provided.

About the trait I can't follow you for what it's needed maybe you can outline it a bit.

The plan to start on j5 is with release of 4.1, at this point we will branch out 4.2 and 5.0, we also have a project to migrate to PSR-12 withing 4.1 which is maybe too late... so it's possible that we do this in 4.1.x or 4.2. to reduce the impact on upmerge to 5.0 but I hadn't the time since the j3.10 phpunit and 8.0 comes into my way...

nikosdion:
I was busy doing releases today. Tomorrow I'll write a POC as a console plugin so you can see what the Trait would look like and how we can tackle b/c. It's about 80% there in my head, I need to put it into code and get it to the next 20% (to make sure I don't mess up b/c with J3 plugins running in Joomla 4). Then we can get back on the discussion about this PR and how much we want to do now.

But as a general idea, yeah, I'd agree that if we convert _all_ of the plugin events at least two minor releases before 5.0 we can definitely remove legacy event handling from 5.0. 

Why two minor releases? Because this would allow for smooth updates based on my experience with the 3.x to 4.0 migration 😉 In August 2021 I had to simultaneously support 3.9, 3.10 and 4.0. This allowed clients using my software on Joomla 3.9 to follow a dead simple upgrade procedure: update extensions, backup, upgrade to Joomla 3.10, backup, upgrade to 4.0, done. My limiting factor was what the _penultimate_ version of 3.x (i.e. 3.9) supported. In those versions I couldn't use any code that only appears in 3.10 and 4.0. So, if we're touching something as fundamental as events handling we should most definitely do it with this kind conservative upgrade process in mind.

<rant>

Regarding PSR-12 I am **NOT** a big fan on two accounts:
1. Namespace bundling in use statements. Sure it makes sense if you use Vim to write PHP. Any half–decent IDE handles imports automatically... as long as you don't do bundling. Enforcing bundling makes for busywork **AND** simultaneously makes it harder to find all uses of classes when they are deprecated, moved to a different namespace etc. 
2. Open braces at the end of the line. Yeah, it makes for more compact code and date back to K&R when 80x25 terminals were _expensive_. We now code in 27" monitors (or larger), typically in vertical orientation. Meanwhile, having braces like that makes the code harder to read and leads to stupid bugs. Why?!
3. 80 characters soft limit for line length. Again, it's no longer 1975, we are not limited to 80x25 terminals. 120 characters is more than reasonable, even on a 13" laptop with sidebars open in any decent IDE (phpStorm, Netbeans, Eclipse, ...).

I think Joomla should have its own code style, like it does now, because the objective of code is to a. perform a certain function; b. make it easily readable; c. minimise busywork and d. minimise the possibility of stupid bugs. PSR-12 fails miserably in the latter three. At some point CS people need to understand the historical context of K&R before trying to apply it indiscriminately on all modern code.

</rant>

HLeithner:
> I was busy doing releases today. Tomorrow I'll write a POC as a console plugin so you can see what the Trait would look like and how we can tackle b/c. It's about 80% there in my head, I need to put it into code and get it to the next 20% (to make sure I don't mess up b/c with J3 plugins running in Joomla 4). Then we can get back on the discussion about this PR and how much we want to do now.

Sounds great thanks

> Why two minor releases? Because this would allow for smooth updates based on my experience with the 3.x to 4.0 migration 😉 In August 2021 I had to simultaneously support 3.9, 3.10 and 4.0. This allowed clients using my software on Joomla 3.9 to follow a dead simple upgrade procedure: update extensions, backup, upgrade to Joomla 3.10, backup, upgrade to 4.0, done. My limiting factor was what the _penultimate_ version of 3.x (i.e. 3.9) supported. In those versions I couldn't use any code that only appears in 3.10 and 4.0. So, if we're touching something as fundamental as events handling we should most definitely do it with this kind conservative upgrade process in mind.

I know the pain...

> Regarding PSR-12 I am **NOT** a big fan on two accounts:
> 
> 1. Namespace bundling in use statements. Sure it makes sense if you use Vim to write PHP. Any half–decent IDE handles imports automatically... as long as you don't do bundling. Enforcing bundling makes for busywork **AND** simultaneously makes it harder to find all uses of classes when they are deprecated, moved to a different namespace etc.
> 2. Open braces at the end of the line. Yeah, it makes for more compact code and date back to K&R when 80x25 terminals were _expensive_. We now code in 27" monitors (or larger), typically in vertical orientation. Meanwhile, having braces like that makes the code harder to read and leads to stupid bugs. Why?!
> 3. 80 characters soft limit for line length. Again, it's no longer 1975, we are not limited to 80x25 terminals. 120 characters is more than reasonable, even on a 13" laptop with sidebars open in any decent IDE (phpStorm, Netbeans, Eclipse, ...).
> 
> I think Joomla should have its own code style, like it does now, because the objective of code is to a. perform a certain function; b. make it easily readable; c. minimise busywork and d. minimise the possibility of stupid bugs. PSR-12 fails miserably in the latter three. At some point CS people need to understand the historical context of K&R before trying to apply it indiscriminately on all modern code.

This topic is heavily opinion driven but to be honest we don't have the resources to maintain our code style and it really doesn't make sense to do every thing different then the rest of the php eco system. We can discuss this to at a later time in another topic. Just to your mentioned things.

1. It's not a must to compound namespaces and I don't like either (namespaces at all haha)
2. To be honest earlier in my life I was unable to deal with other code styles as my own (and I hate open braces on it own line ;-) but I learned to life with other codestyles and in the meantime I'm happy if the same codestyle is use by the complete application and not 10 different (or none) codestyle. But yes that's opinionated and thanks good with have braces and don't need to care about the right indention.
3. The good thing is it's only a softlimit and I'm pretty sure we keep our 150 character limited (also because nobody want's to rewrite the codebase and work on a c64)

Anyway that topic can be discussed later and if we need exceptions or stick to psr-12 completely.

chmst:
@nikosdion could you please fix conflicts here?

nikosdion:
I'm not sure why the merge freaked out. Let me see if I can fix it.

nikosdion:
I think GitHub is having a stroke — or trying to give me one. It finds three changed files... which are actually not changed!

* administrator/components/com_joomlaupdate/tmpl/joomlaupdate/noupdate.php I checked manually, it is the same as the file after merging https://github.com/joomla/joomla-cms/pull/36134/files
* api/components/com_installer/src/Controller/ManageController.php it is as merged through https://github.com/joomla/joomla-cms/pull/36376/files
* build/release-checklist.md it is as merged through ca423557ad5435a3fd42242f9a66619e41819c3a

The other files are indeed what is part of my PR.

It's as if GitHub calculates the PR against the repository as it was **YESTERDAY** instead of as it is right now. When I do a diff between my brach, as it is checked out, and `4.1-dev` I only see my four files changed.

The good news is that if you merge the PR _right now_ you will have no adverse effects. As to why GitHub doesn't calculate the PR correctly, beats me!

nikosdion:
@HLeithner @bembelimen Is there any chance this will be included in Joomla 4.1? It's nearly two months this is done and dusted, no feedback, everyone agrees plugins need to be refactored, what's the holdup here?

laoneo:
I think it should go into 4.2 and I'm willing to help out converting the other plugins as well.

nikosdion:
@laoneo Thank you! I can help out with this as well.

laoneo:
@nikosdion can you rebase it to the branch 4.2? I will give it a test then.

api/components/com_installer/src/Controller/ManageController.php
```diff
@@ -59,11 +59,11 @@ public function displayList()
 
 		if (!\is_null($requestBool))
 		{
-			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0', 'STRING');
+			$this->modelState->set('filter.core', ($requestBool === 'true') ? '1' : '0');
 		}
 
-		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status')), 'INT');
-		$this->modelState->set('filter.type', $this->input->get('type', $this->input->get->get('type')), 'STRING');
+		$this->modelState->set('filter.status', $this->input->get('status', $this->input->get->get('status', null, 'INT'), 'INT'));
```
roland-d: @nikosdion Just curious, why not use `getInt` and `getString` here?

nikosdion: This is one of the stupid things GitHub did to my PR when I rebased last time. I have NOT changed the first three files which appear changed. I have only touched the cache plugin. These changes are not in my local repository either. I don't know why it's having a stroke. See https://github.com/joomla/joomla-cms/pull/36042#issuecomment-1026253136

I'll try to see if I can address this by merging 4.2-dev onto my branch. If this doesn't work I'll close this PR and submit a new one made with 4.2-dev as the starting point with just the refactored plugin files included.

roland-d: Understood

nikosdion:
Hallelujah! Merging 4.2-dev onto my branch and changing the base branch of the PR convinced GitHub to get out of its braindead space and only show the real files changed. @roland-d this is ready for actual review :)

nikosdion:
You may also want to change the labels, unless they are assigned automatically.

plugins/system/cache/cache.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<extension type="plugin" group="system" method="upgrade">
+<extension version="4.0.0" type="plugin" group="system" method="upgrade">
```
laoneo: This should be reverted as we removed all in core.

laoneo (approved):

plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,366 @@
...
+
+		// Mark afterCache in debug and run debug onAfterRespond events, e.g. show Joomla Debug Console if debug is active.
+		if (JDEBUG)
+		{
+			// Create a document instance and load it into the application.
+			$document = $this->documentFactory
+				->createDocument($this->app->input->get('format', 'html'));
+			$this->app->loadDocument($document);
+
+			Profiler::getInstance('Application')->mark('afterCache');
```
laoneo: Can this be injected as well? Can be null if not debug mode in the service provider.

nikosdion: No, we cannot inject the document. There is no document when we initialise the plugin. We need to wait onAfterRoute to have a document. That's why the service provider injects the document factory.

plugins/system/cache/cache.xml
nikosdion: Will do

plugins/system/cache/src/Extension/Cache.php
laoneo: I meant the profiler instance

plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,365 @@
...
+	 * throughout the lifetime of the request.
+	 *
+	 * The other two tests MUST NOT be cached because auto–login plugins may fire anytime within
+	 * the application lifetime logging in a user and messages can be generated anytime within the
+	 * application's lifetime.
+	 *
+	 * @return  boolean
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function appStateSupportsCaching(): bool
```
laoneo: Why protected, when the class is final? Same for the others.

plugins/system/cache/src/Extension/Cache.php
nikosdion: Would you please do a PR to my branch with that? I am 95% into understanding what you mean but I am unclear how you'd inject the dependency conditionally.

plugins/system/cache/src/Extension/Cache.php
nikosdion: Fair point. I left the prior visibility. Let me make all protected members private.

plugins/system/cache/src/Extension/Cache.php
laoneo: Somehow github doesn't find your fork to create a pr to. But basically I had something like https://github.com/Digital-Peak/joomla-cms/commit/3fc88cfb59c7079469d0c8f2a6f06c640e73baf5 in mind.

nikosdion: Thank you! Done :)

laoneo:
I have tested this item :white_check_mark: successfully on 52e20e334c5f8b2cafcaf79a2f61daf1bfc68ec3<br /><br />Made some menu items and browsed them on the front. Cache is created and served on next page load. Added a die on one of the default.php after the page is cached. On a reload the page loaded normally.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36042">issues.joomla.org/tracker/joomla-cms/36042</a>.</sub>

roland-d (approved):
@nikosdion Well done, looks good to me.

plugins/system/cache/services/provider.php
```diff
@@ -0,0 +1,46 @@
+<?php
+/**
+ * @package         Joomla.Plugin
```
laoneo: I guess this doc block is somehow screwed, when I compare it with other files.

nikosdion: What do you mean?

brianteeman: other than the copyright line which I guess should be 2022 and indented so that the alignment is on the (c and not the 2 otherwise it seems ok

laoneo:
@nikosdion can you fix the conflicts here? Thanks.

nikosdion:
@laoneo There you go.

I also fixed the docblock and inject the frontend router object; these should be the last two pending items.

laoneo:
The intendation is still not correct as the other docblocks do look like:
image
 while on yours:
image

nikosdion:
The funny thing is that the weird indentation comes from me copying that weird docblock from the original file. Whenever I have time (this week is really difficult) I will reformat the docblocks. Unless you want to do a PR to my branch if you have the time?

plugins/system/cache/services/provider.php
```diff
@@ -0,0 +1,48 @@
...
```
laoneo:
```suggestion
 * @package     Joomla.Plugin
```

plugins/system/cache/services/provider.php
```diff
@@ -0,0 +1,48 @@
...
+ * @subpackage      System.cache
```
laoneo:
```suggestion
 * @subpackage  System.cache
```

plugins/system/cache/services/provider.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
```
laoneo:
```suggestion
 * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

plugins/system/cache/services/provider.php
```diff
@@ -0,0 +1,48 @@
...
+ * @license         GNU General Public License version 2 or later; see LICENSE.txt
```
laoneo:
```suggestion
 * @license     GNU General Public License version 2 or later; see LICENSE.txt
```

laoneo:
Guess now it is correct.

nikosdion:
Thank you

plugins/system/cache/src/Extension/Cache.php
```diff
@@ -0,0 +1,401 @@
...
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct(
+		&$subject,
+		$config,
+		DocumentFactoryInterface $documentFactory,
+		CacheControllerFactoryInterface $cacheControllerFactory,
+		?Profiler $profiler,
+		?SiteRouter $router
```
laoneo, line 111: Guess you have to make this dependency required, otherwise add a check where the router is used if it is null.

nikosdion: No, you cannot make it a required dependency. 

A `system` plugin is **ALWAYS** loaded, even in applications which do not have a router (e.g. the built-in CLI application, custom applications extending WebApplication etc). Making the router a required dependency would cause the plugin initialisation to fail, taking the application down with it, when the application does not provide a router. That was why I was not very enthusiastic about making it a dependency. The only reason I did was to make the code testable.

Beyond unit tests, where does this plugin execute? It will execute in applications extending from WebApplication. **HOWEVER** this plugin will **ONLY EVER** execute its methods if it is absolutely sure we are in the `site` application. It checks for that, remember? That's what we started doing here :) Therefore when the time comes to use the router dependency we are **guaranteed** to have a router.

Now, you will tell me, what if the Site application does not have a router? Not a problem _for this plugin_. This would means that Joomla (or, rather, the impostor application pretending to be Joomla's `site` application but not providing a router object!) is completely broken and it never reaches the execution point where we try to use the missing router dependency. Therefore there's an implicit check. Adding an explicit check is redundant and a sign that the developer didn't understand the application architecture.

This is the kind of subtle reasoning that is paramount to refactoring plugins, especially `system` plugins, in Joomla 4. Do not approach this as a mechanical task with inflexible computer science axioms or you will break userland. You need to stop and think how this is used, where this used and what your change will cause beyond the immediate effects on your target. Most bugs I find in Joomla are because someone was looking at the tree, missing the forest.

laoneo: Then you need to check here if it is null, or not? https://github.com/joomla/joomla-cms/pull/36042/files#diff-99d32ed5de01e06de65236b838410449dc6f1fdcb99c51a5066ac6e5c44e6e95R354

nikosdion: No, I do not. That's the point. `onAfterRender` will only be called **long** after Joomla performs SEF routing. Therefore if the application does not have a SEF router — the only case where our plugin's service provider would send null instead of Router object — it will have failed _before_ calling our plugin's event. As a result we do not need to check if the router is null because, by definition, at the time we are called we are guaranteed to have a valid router object.

laoneo:
Thanks for your patience! Now we can start to convert the rest of the plugins.

nikosdion:
Thank you for merging!

Issue #35808, opened 2021-10-11
Media Manager changes folder and file names to all lowercase


TonyCowling:
### Steps to reproduce the issue
From the administrator interface, use the media manager to 
(i) create a folder with upper case characters in its name, or
(ii) upload an image with upper case characters in its filename.

### Expected result
Names should be created as they have been input.

### Actual result
All uppercase letters are converted to lowercase (both as displayed, and as the actual folders or files are created within the folder <images>).

### System information (as much as possible)
PHP : version 7.4.23.
Build : Joomla! 4.0.3 Stable [ Furaha ] 12-September-2021 10:39 GMT
Browser : Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:93.0) Gecko/20100101 Firefox/93.0
Database : MySQL 5.7.35-cll-lve
Customer : I'm an administrator for the site.
Web Server : Apache (version number not easily accessible).

### Additional comments
The behaviour reported here is different from Joomla 3.9, and from my perspective as a user it is a difference that is not wanted, which is why I am reporting it as a bug.

I note that, if I use the underlying filesystem to change the lowercase back to uppercase, the media manager displays the changed names correctly, so I can not see any technical reason for the Media Manager to make these changes to names.

joomdonation:
Could not re-procedure the issue with 4..0.4

AMurray2016:
I reproduced this.  I made a folder called "CAP_letters" and it changed it to "cap_letters".

Maybe it depends on if you're on Linux or Windows hosting.  It would make sense, to have all lower case so you don't confuse file names.  "This_Image.JPEG" is different to "this_image.jpeg" on Linux.  Windows doesn't care one way or the other.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35808">issues.joomla.org/tracker/joomla-cms/35808</a>.</sub>

AMurray2016:
[...]

PR #28002, opened 2020-02-21
[4.0] Media manager: Transliteration of UTF8 file names when uploading (makeSafe)

infograf768:
Redo of https://github.com/joomla/joomla-cms/pull/16878 for J4 as we do not have the same problems.
https://github.com/joomla/joomla-cms/pull/16878#issuecomment-586713012

### Summary of Changes
Similar to https://github.com/joomla/joomla-cms/pull/27974 but for media files upload

It takes advantage of the PHP extension `intl` **when it is enabled** to use the `transliterator_transliterate()` method, itself using `ICU` library.

The php extension is available since **php 5.4.0**, but may not be enabled on some hosts.
If disabled, former behavior is used which does not accept to upload files with UTF8 names and displays an error. This PR is therefore totally BC.

Using `iconv` and `IGNORE` let's get rid of some prime-characters that can't be transliterated, like the Cyrillic letter `ь`.

### Testing Instructions
Check in System Information => PHP Information that the extension is enabled:
You should get something like this:

Screen Shot 2020-02-18 at 08 08 42

If it is not enabled on your local environment, try to modify your PHP.ini.

Change an image file name to utf8.
I used here `'完 成'".png` to be sure we also get rid of unwanted characters.

Go to Media Manager =>Upload, select the file and click upload.

'完 成'

In the case above I get a pure ascii name wan-cheng.png:

media_utf8

PR #12049, opened 2016-09-16
Improve creation filesystem-safe filenames, by using multiple language transliterations

ggppdk:
Pull Request for Issue #7841
### Summary of Changes
1. Multiple transliterations are tried
2. [RFC] In order to avoid ugly looking empty filenames, current time is used if alls transliteration fail to do a complete job
3. Added new method makeUnique($base_Dir, $file) for finding a unique file name inside a given folder
### Testing Instructions

Upload files that are using characters in SITE or ADMIN default language of your website
- Filenames should be transliterated

Please also do a code review if you can
### Documentation Changes Required

A new optional parameter $language = null is added to the method ... for doing a specific language transliteration

Issue #7841, opened 2015-09-09
JFile::makeSafe creates invalid filenames and prevents upload


osignell:
#### Steps to reproduce the issue
- Create an image with the name "åäö.jpg" or any other filename consisting of ONLY non-english characters.
- upload via Media Manager in the backend.
#### Expected result

File uploaded and stored with a "safe" filename.
#### Actual result

File is not uploaded. An error message about "This file type is not supported." is displayed.
#### System information (as much as possible)

Joomla 3.4.3, Linux, Apache, PHP 5.5
#### Additional comments

The makeSafe function strips out all non-english characters from the filename, leaving only "png" which does not qualify as a valid filename.

Could we do this the same way as in JFilterOutput::stringURLSafe() and use JFactory::getLanguage()->transliterate() on the filename before running the stripping regexp?

brianteeman:
Issue confirmed
Another example is with a file names 1åäö.jpg - this will be uploaded as 1.jpg

ryandemmer:
This needs to be language independent, so it should not require additional processing through a language file.

The regular expression used by JFile::makeSafe needs to use a PCRE modifier to allow UTF-8 characters, eg:

```php
$regex = array('#(\.){2,}#', '#[^a-zA-Z0-9_\.\-~\p{L}\p{N}\s ]#u');
```
however this requires PCRE 5 support.

JCE already handles this quite well I think, so one might consider a variation of:
https://github.com/widgetfactory/jce/blob/master/components/com_jce/editor/libraries/classes/utility.php#L146-L215

Bakual:
I wouldn't make it to complicate.
I would use the `JApplication::stringURLSafe()` method which takes into account the setting from the config if we want to allow unicode alias or not.
If the resulting filename doesn't contain any valid characters anymore, either show an error or just create a filename based on datetime.

ryandemmer:
`JApplication::stringURLSafe()` makes the string lowercase and removes the period character.

mbabker:
It still uses the language system and URL safe strings don't use the same
rules as filesystem safe strings so I wouldn't necessarily go that route.

Bakual:
I don't have an issue if it uses language functions. I'd say there is a high chance the special characters present in the filename are from the same language the user has active.
Also lowercasing the filename isn't necessary a bad thing when it comes to filenames that end up in URLs. It solved a lot of support requests in the case of my extension :smile: 
But it may indeed be the wrong approach for the core.

[...]

PR #36047, opened 2021-11-15
[4.0] Do not force the change of uploaded mm files to lowercase

laoneo:
Pull Request for Issue #35808.

### Summary of Changes
In the media manager all file names are changed to lower case on upload. This behavior was introduce in #28002 and works only when the INTL extension is installed. Not sure if lower case is intended to be added as well, @infograf768 can shed perhaps some light into this.

This pr reverts the lower case check.

The history of this changes can be traces back to 2016:
- https://github.com/joomla/joomla-cms/pull/28002
- https://github.com/joomla/joomla-cms/pull/12049
- https://github.com/joomla/joomla-cms/issues/7841

### Testing Instructions
- Make sure the INTL extension is installed on your host.
- Upload an image with upper case characters in media manager.

Image file name is changed to lowercase.

Image file name keeps upper case characters.

### DIFF
--- a/libraries/src/Filesystem/File.php
+++ b/libraries/src/Filesystem/File.php
@@ -92,7 +92,7 @@ class File
 		if (function_exists('transliterator_transliterate') && function_exists('iconv'))
 		{
 			// Using iconv to ignore characters that can't be transliterated
-			$file = iconv("UTF-8", "ASCII//TRANSLIT//IGNORE", transliterator_transliterate('Any-Latin; Latin-ASCII; Lower()', $file));
+			$file = iconv("UTF-8", "ASCII//TRANSLIT//IGNORE", transliterator_transliterate('Any-Latin; Latin-ASCII', $file));
 		}
 
 		$regex = array('#(\.){2,}#', '#[^A-Za-z0-9\.\_\- ]#', '#^\.#');


brianteeman:
who would have guessed that an untested pull requested would have caused problems 🤦

ceford:
I have tested this item :white_check_mark: successfully on d8e5ecc97ad80ba5869834d396e773629e3715f8<br /><br />Confirmed - case is preserved on upload to media folder.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36047">issues.joomla.org/tracker/joomla-cms/36047</a>.</sub>

richard67:
@laoneo I think the title of this PR is misleading because of a mistake in translation to English commonly made by German native speakers.

"Must not" is **_not_** the opposite of "must" in English language like one would think if not English native speaker, it has a different meaning.

The opposite of "must" is "not need to" or "don't have to", while "must not" has the meaning of "shall not" or "may not" or "are not allowed to".

So "Must not be lower case" doesn't mean that they may me upper or lower case case, it means they **_have to be_** uppercase, and that's wrong.

I suggest to change the title to "MM uploads don't have to be lower case" or "Allow MM uploads to be upper case, too" or something like that.

brianteeman:
From the issue as described I think a more accurate title would be "Do not force the change of uploaded mm files to lowercase"

richard67:
> From the issue as described I think a more accurate title would be "Do not force the change of uploaded mm files to lowercase"

Yes, that would be best.

infograf768:
@laoneo 
> Not sure if lower case is intended to be added as well, @infograf768 can shed perhaps some light into this.

I guess the intention was to normalize results for all languages, including those not using uppercase letters like

Chinese, Korean, Arabic, Japanese, Indic languages (Hindi, Bengali, Tamil, Gujarati, Marathi, etc.), Thai, etc.
See https://www.quora.com/Which-languages-have-no-capitalized-letter for an extensive list of 46 languages (add there their subs) 

Scripts including uppercases are those based on Armenian, Cyrillic, Greek, Latin

Just a matter of choice. There were some comments about making these names url safe, therefore lowercase.

RickR2H:
I have tested this item :white_check_mark: successfully on d8e5ecc97ad80ba5869834d396e773629e3715f8<br /><br />Patch works!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36047">issues.joomla.org/tracker/joomla-cms/36047</a>.</sub>

RickR2H:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36047">issues.joomla.org/tracker/joomla-cms/36047</a>.</sub>

Octopus2:
Can this be integrated please. 
I use fields to make PDF-files available for download. By changing the file names from the media manager to lower case, the result in the frontend looks as if the website operator is not interested in correct spelling. At the moment I can only help myself by uploading the PDF files with the correct filenames via FTP.

richard67:
The is RTC, ready to commit, so it very likely will go into the next release.

Yelling will not speed that up.

Octopus2:
Sorry, I didn't know the workflow here and didn't want to jostle. I just wanted to draw attention to one use case.

wilsonge:
Thanks!

PR #36063, opened 2021-11-16
[4.0] Use an absolute path in image list field

laoneo:
### Summary of Changes
When using an image list custom field. The template needs to have an absolute path. It is a regression from #30784.

We discovered this in DPCalendar when custom fields are rendered for invoices. As I can't offer a test case this one needs to go by code review. It is pretty obvious when you check the `Image::getImageFileProperties` function.

### Testing Instructions
Somehow it needs to render a custom field in the back end which is not possible in core atm.


dgrammatiko:
This Field is still not compatible with external adapters... 😔 Shouldn't it be?

laoneo:
Nope, it is about local images and not media manager.

laoneo:
@wilsonge this here is code review only as I can't find a use case where this can be tested in core.

laoneo:
@wilsonge can you please consider this into the next release? If you check the argument here https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Image/Image.php#L175 then you can clearly see that the full path is needed and not the one relative to the index.php

wilsonge:
Thanks!

Issue #36089, opened 2021-11-22
[4.0] Joomla! Updater has to delete the autoload_psr4.php to prevent Exceptions for new system plugins


bembelimen:
### Steps to reproduce the issue
- Install Joomla! 4.0.x
- Update Joomla! 4.1 with this PR: https://github.com/joomla/joomla-cms/pull/35143 (Using the download option there)

### Expected result
No exception

### Actual result
Exception, that new namespaced classed could not be found.

### Additional comments
This needs an update of the updater in my opinion.

@wilsonge

richard67:
Closing as having a pull request. Please test #36094 . Thanks in advance.

PR #36094, opened 2021-11-23
[4.0] Recreate namespace map on Joomla! update

joomdonation:
Pull Request for Issue #36089.

### Summary of Changes
This PR makes small modification to update Joomla! process so that namespace map is re-generated during Joomla! update. It is needed because when we update to new Joomla version which has new extension added, we need to register namespace of that new extension so that it can work properly.

### Testing Instructions
1. Apply patch from this PR
2. Download the this update package [Joomla_4.1.0-dev+pr.35143-Development-Update_Package.zip](https://github.com/joomla/joomlacms/files/7589180/Joomla_4.1.0-dev%2Bpr.35143-Development-Update_Package.zip). I had to modify original update package for this PR https://github.com/joomla/joomla-cms/pull/35143 (which introduces new extension for Joomla 4.1) to include the change from this PR so that update process runs smooth.
3. Go to System -> Update -> Joomla, upload and update your Joomla 4 installation with the above downloaded package. Make sure you do not see any error with the update process.

Without the change from this PR, updating your site to the update package of this PR https://github.com/joomla/joomla-cms/pull/35143 will cause a fatal error:

> 0 Class 'Joomla\Component\Scheduler\Administrator\Extension\SchedulerComponent' not found

Update success. Namespace for new extension is successfully registered in administrator/cache/autoload_psr4.php.

No.

### DIFF
--- a/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php
+++ b/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php
@@ -661,7 +661,7 @@ ENDDATA;
 		}
 
 		// Re-create namespace map. It is needed when updating to a Joomla! version has new extension added
-		(new \JNamespacePsr4Map())->create();
+		(new \JNamespacePsr4Map)->create();
 
 		$installer->manifest = $manifest;
 
--- a/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php
+++ b/administrator/components/com_joomlaupdate/src/Model/UpdateModel.php
@@ -660,6 +660,9 @@ ENDDATA;
 			return false;
 		}
 
+		// Re-create namespace map. It is needed when updating to a Joomla! version has new extension added
+		(new \JNamespacePsr4Map())->create();
+
 		$installer->manifest = $manifest;
 
 		$installer->setUpgrade(true);


richard67:
Added the release blocker label as inherited from the issue.

bembelimen:
Is a Updater update needed? so if someone updates from 4.0.4 to 4.1 it will still crash?

laoneo:
I would not make a hard reference to the map creator in the model. Instead of listen to the 'onTableAfterStore' event in the `plugins/extension/namespacemap/namespacemap.php` plugin and recreate the map.

joomdonation:
@bembelimen The modified file is extracted before the request is executed, so the new code will be executed, thus we do not need an update for Updater

 @laoneo I don't know how `onTableAfterStore` helps here. If we want to go with plugin approach, I think we can trigger a new event `onJoomlaAfterUpdate` (we triggered `onJoomlaBeforeUpdate` earlier) and add new  method to plugins/extension/namespacemap/namespacemap.php to re-create the map.

laoneo:
@joomdonation that sounds even better. As I made a bad experience adding new events to Joomla, I'm a bit hesitant to add a new ones, that's why I didn't suggest that.

bembelimen:
> @bembelimen The modified file is extracted before the request is executed, so the new code will be executed, thus we do not need an update for Updater

Yes, but what happens, if there is a plugin, which uses the new namespaces on a very early trigger (like onAfterInitialize), is this covered?

joomdonation:
> Yes, but what happens, if there is a plugin, which uses the new namespaces on a very early trigger (like onAfterInitialize), is this covered?

As I understand, that plugin would only run on the next request. In the current request, the new plugin (if added in the update) is not available in database yet, so it is not being executed.

bembelimen:
I have tested this item :white_check_mark: successfully on 1592fc7dc50863a8d32afb05c3870ab704df855f<br /><br />For the test scenario you should not patch the base 4.0 version but only the version you're updating to. (it's then like 4.0.4 => 4.1).

But anyways, changes are working for me.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36094">issues.joomla.org/tracker/joomla-cms/36094</a>.</sub>

khu5h1:
I have tested this item :white_check_mark: successfully on 1592fc7dc50863a8d32afb05c3870ab704df855f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36094">issues.joomla.org/tracker/joomla-cms/36094</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36094">issues.joomla.org/tracker/joomla-cms/36094</a>.</sub>

wilsonge:
I think not having this in a plugin is probably a good idea here - we don't want any weird issues if we ever namespace the plugin in the future.

wilsonge:
Thanks!

Crisje:
hi im having this issue atm and idk what to do, I tried to download the link above but it says not found 
please if anyone could help me 
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36094">issues.joomla.org/tracker/joomla-cms/36094</a>.</sub>

joomdonation:
@Crisje Delete the file administrator/cache/autoload_psr4.php from your Joomla 4 installation should help solving your issue.

hjstrauss:
Thanx **@Crisje** for your tip that worked for me. But why is this still in 4.1.2 code? I did an update from 4.0.3 to 4.1.2 and ran into this error. Kind regards, HaSTe.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36094">issues.joomla.org/tracker/joomla-cms/36094</a>.</sub>

PR #36103, opened 2021-11-24
[4.1][webservices] get bearer token on credentials

alikon:
### Summary of Changes
Add an authorization endpoint

### Testing Instructions
apply patch
create an user that have the API Token
 do a POST `{{base_url}}api/index.php/v1/users/auth`
with this  body
```
{
 "username": "test",
 "password": "123456789012"
}

```

N/A

it returns the Bearer token  so an app can consume the not public webservices endpoints
image

yes

### DIFF
--- a/administrator/components/com_users/src/Model/UsersModel.php
+++ b/administrator/components/com_users/src/Model/UsersModel.php
@@ -602,14 +602,15 @@ class UsersModel extends ListModel
 
 		$db->setQuery($query);
 		$result = $db->loadObject();
-		
+
 		if ($result)
 		{
 			$match = UserHelper::verifyPassword($credentials['password'], $result->password, $result->id);
 
 			if ($match === true)
 			{
-				$user = User::getInstance($result->id);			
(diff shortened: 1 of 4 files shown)


richard67:
@alikon PHP CS: https://ci.joomla.org/joomla/joomla-cms/48630/1/6

wilsonge:
Sorry I'm hard rejecting this - this is a security thing.

It's inherently insecure to be able to grab tokens via a webservice I'm afraid. This is also going to full on conflict with any OAuth services (which are how this stuff tends to manifest in most webservices) - which is something we've documented as something we expect to bubble up in the 3rd party ecosystem rather than in core.

alikon:
where is documented ?

wilsonge:
https://docs.joomla.org/Joomla_Api_Specification

> We have chosen not to implement a full blown oAuth 2 specification into the core however it is intended that this can be achieved with the plugin group (and by disabling the "API Authentication - Joomla Token" plugin and the "User - Joomla Token" plugin)

PR #36128, opened 2021-11-26
[4.1] 'return' is unnecessary as the last statement in a method

PhilETaylor:
code review

### DIFF
--- a/administrator/components/com_scheduler/src/Task/TaskOption.php
+++ b/administrator/components/com_scheduler/src/Task/TaskOption.php
@@ -85,6 +85,7 @@ class TaskOption
 		{
 			return $this->$name;
 		}
-
+		
+		return null;
 	}
 }
--- a/administrator/components/com_scheduler/src/Task/TaskOption.php
+++ b/administrator/components/com_scheduler/src/Task/TaskOption.php
@@ -85,7 +85,7 @@ class TaskOption
 		{
 			return $this->$name;
 		}
-		
+
 		return null;
 	}
 }
--- a/administrator/components/com_scheduler/src/Task/TaskOption.php
+++ b/administrator/components/com_scheduler/src/Task/TaskOption.php
@@ -86,6 +86,5 @@ class TaskOption
 			return $this->$name;
 		}
 
-		return;
 	}
 }


administrator/components/com_scheduler/src/Task/TaskOption.php
```diff
@@ -86,6 +86,5 @@ public function __get(string $name)
 			return $this->$name;
 		}
 
-		return;
```
Fedik, line 89: maybe better would be `return null;`  because it is `__get`

PhilETaylor: No. They both would return null as evidenced here https://3v4l.org/YqP4K

Fedik: it still better to have `return null`, 
check doc block: "return mixed"

HLeithner:
please remove the blank line between the }

administrator/components/com_scheduler/src/Task/TaskOption.php
PhilETaylor: disagree - so does an IDE and Psalm - but hey - have it your way, null added.

PhilETaylor:
> please remove the blank line between the }

No I just put it back in as @Fedik wants to return again

administrator/components/com_scheduler/src/Task/TaskOption.php
Fedik: thanks :)

Fedik:
I have tested this item :white_check_mark: successfully on fab6c5f0ad0432134fc780e09b5bc8ca5ea207ef<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36128">issues.joomla.org/tracker/joomla-cms/36128</a>.</sub>

HLeithner:
now we have 2 tabs instead of an empty line, I think thats wrong ;-)

PhilETaylor:
blame GitHub online editing :(

HLeithner:
thanks

PR #36146, opened 2021-11-28
[4.0] Article Manager RTL correction

brianteeman:
This PR removes a conditional block from the article manager and the featured article manager that was intended to ensure that the display of the category and nested category was correct when the admin is using an rtl language and the category names may be a mix of LTR and RTL.

Not really sure why but
1. this code does not work correctly. the order of the categories is wrong and the << character is wrong
2. there is no need for this alternate block of code for RTL

### Before
image

### After
image

### DIFF
--- a/administrator/components/com_content/tmpl/articles/default.php
+++ b/administrator/components/com_content/tmpl/articles/default.php
@@ -282,46 +282,23 @@ $assoc = Associations::isEnabled();
 													echo ' &#187; ';
 												endif;
 											endif;
-											if (Factory::getLanguage()->isRtl())
-											{
-												if ($canEditCat || $canEditOwnCat) :
-													echo '<a href="' . $CurrentCatUrl . '" title="' . $EditCatTxt . '">';
+											if ($item->category_level != '1') :
+												if ($canEditParCat || $canEditOwnParCat) :
+													echo '<a href="' . $ParentCatUrl . '" title="' . $EditCatTxt . '">';
 												endif;
-												echo $this->escape($item->category_title);
-												if ($canEditCat || $canEditOwnCat) :
+												echo $this->escape($item->parent_category_title);
+												if ($canEditParCat || $canEditOwnParCat) :
 													echo '</a>';
 												endif;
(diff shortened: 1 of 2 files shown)


ceford:
For some reason, with or without the patch applied, with Persian as the admin language I see Persian text embedded in the Category name. also containing a colon:

screen shot 2021-11-29 at 19 00 08<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36146">issues.joomla.org/tracker/joomla-cms/36146</a>.</sub>

brianteeman:
thats the word category in persian

infograf768:
I do not get the same result as the screenshots above. Never gets 2 `«` before or after patch.

Missing instructions: what is the sub category to which is assigned the article in the original description ?

In the sample below, the parent is Kategorie(de-de) and I created a subcategory in which I moved the article.

BEFORE PATCH , RTL

Screenshot 2021-11-30 at 09 01 47


AFTER PATCH
RTL
Screenshot 2021-11-30 at 08 49 30

LTR

Screenshot 2021-11-30 at 08 57 11

Is it what is expected? If yes, patch looks good.

brianteeman:
I am guessing that your categories are
```
Kategorie (de-de)
-  مجموعه whatever (fa-ir)
```

If so then your screenshots show the expected results and a faulty before shot

Not sure what you mean about a second `«`

brianteeman:
@infograf768 please can you respond

brianteeman:
@infograf768 please explain what the problem is as your comment is blocking this being merged

infograf768:
Read again:
`Is it what is expected? If yes, patch looks good.`

I just remarked a wrong description of the issue. You wrote:
`this code does not work correctly. the order of the categories is wrong and the << character is wrong`

instead of ` and the « character is wrong`

brianteeman:
Ah you're playing your silly games again. Never mind I will just close it.

Issue #36179, opened 2021-12-01
[4.0.5] Trying to get property '0' of non-object in tinymce.php


Quy:
### Steps to reproduce the issue
Go to System > Plugins
Edit `Editor - TinyMCE` plugin.
Click `Save & Close` button.

Create or edit an article.
See error.

### Actual result
 > Notice: Trying to get property '0' of non-object in \plugins\editors\tinymce\tinymce.php on line 200

brianteeman:
Is this a completely clean installation?

brianteeman:
I could not replicate this on a completely clean install

chmst:
I cannot replicate the error on the site where I am working on (4.0.5 dev, not empty)

brianteeman:
The reason that I said not empty was that I could replicate it on an install that had moved up to 4.1 and then back to 4.0

Quy:
4.0.4 Clean install OK
4.0.5 Clean install of the nightly build. NOT OK

![36179](https://user-images.githubusercontent.com/368084/144309061-e29342e6-00c1-477a-b168-48bed46160b8.jpg)

dgrammatiko:
`$toolbarParams = $toolbarParamsAll->{$set};` should do

Quy:
Same error. Initially, it is an object. After saving TinyMCE settings, it is an array.

Before:
```
object(stdClass)[1058]
  public '2' => 
    object(stdClass)[1057]
      public 'toolbar1' => 
        array (size=11)
          0 => string 'bold' (length=4)
          1 => string 'underline' (length=9)
          2 => string 'strikethrough' (length=13)
          3 => string '|' (length=1)
          4 => string 'undo' (length=4)
          5 => string 'redo' (length=4)
          6 => string '|' (length=1)
          7 => string 'bullist' (length=7)
          8 => string 'numlist' (length=7)
          9 => string '|' (length=1)
          10 => string 'pastetext' (length=9)
```
After:
```
array (size=3)
  0 => 
    object(stdClass)[1066]
      public 'menu' => 
        array (size=6)
          0 => string 'edit' (length=4)
          1 => string 'insert' (length=6)
          2 => string 'view' (length=4)
          3 => string 'format' (length=6)
          4 => string 'table' (length=5)
          5 => string 'tools' (length=5)
```

dgrammatiko:
Try 
```php
$extraOptionsAll  = (object) $this->params->get('configuration.setoptions', array());
[...]

PR #36186, opened 2021-12-02
[4.0] Fix tinymce plugin configuration

Fedik:
Pull Request for Issue #36179 .

### Summary of Changes

The changes makes sure the value in correct type, and uses correct order.

### Testing Instructions
Please follow #36179 

PHP notice, and configuration loaded incorrectly, eg wrong toolbar for Super admin.

Configuration loaded correctly

no

### DIFF
--- a/plugins/editors/tinymce/tinymce.php
+++ b/plugins/editors/tinymce/tinymce.php
@@ -186,7 +186,7 @@ class PlgEditorTinymce extends CMSPlugin
 		$extraOptionsAll  = (array) $this->params->get('configuration.setoptions', array());
 		$toolbarParamsAll = (array) $this->params->get('configuration.toolbars', array());
 
-		// Sort the array, so the items with lowest access level goes first
+		// Sort the array in reverse, so the items with lowest access level goes first
 		krsort($extraOptionsAll);
 
 		// Get configuration depend from User group
--- a/plugins/editors/tinymce/tinymce.php
+++ b/plugins/editors/tinymce/tinymce.php
@@ -186,7 +186,7 @@ class PlgEditorTinymce extends CMSPlugin
 		$extraOptionsAll  = (array) $this->params->get('configuration.setoptions', array());
 		$toolbarParamsAll = (array) $this->params->get('configuration.toolbars', array());
 
-		// Reverse the array, so the items with lowest access level goes first
+		// Sort the array, so the items with lowest access level goes first
 		krsort($extraOptionsAll);
 
 		// Get configuration depend from User group
--- a/plugins/editors/tinymce/tinymce.php
+++ b/plugins/editors/tinymce/tinymce.php
@@ -183,12 +183,14 @@ class PlgEditorTinymce extends CMSPlugin
 		$levelParams      = new Joomla\Registry\Registry;
 		$extraOptions     = new stdClass;
 		$toolbarParams    = new stdClass;
-		$extraOptionsAll  = $this->params->get('configuration.setoptions', array());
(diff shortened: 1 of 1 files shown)


joomdonation:
If I'm not mistaken, this won't work for old Tiny MCE configuration (for example, site updated from older version 4.0.5 and don't change editor settings after update). In that case, set 2 would be used for Super Users and it's not right.

Fedik:
hmhm, yea, I probably better change to `krsort()` instead of `array_reverse()`.
so it will sort the same for old and new values

joomdonation:
Yes, it works. Or we can just use `ksort` and stop at the first matching set.

```php
// Sort the array, so the items with highest access level goes first
ksort($extraOptionsAll);

// Get configuration depend from User group
foreach ($extraOptionsAll as $set => $val)
{
	$val = (object) $val;
	$val->access = empty($val->access) ? [] : $val->access;

	if (array_intersect($ugroups, $val->access))
	{
		$extraOptions  = $val;
		$toolbarParams = (object) $toolbarParamsAll[$set];
		break;
	}
}
```
Compare to the current code, it has super tiny improvement because it does not have to loop through all the sets (stop at the first matching). Let me know if you want to make change, otherwise, I will record my test result.

Fedik:
~No no, it should not stop on first,~
~Example if you added some group in two sets (accidentally or in purpose), then it should pick last with higher access level.~

Well I would leave it as it is :)

joomdonation:
Not a problem. As I mentioned, it is a super tiny improvement :D.

joomdonation:
I have tested this item :white_check_mark: successfully on 7de6d1876b8b2f9a9b2002a1c4e59bc6f6fb2dcd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36186">issues.joomla.org/tracker/joomla-cms/36186</a>.</sub>

dgrammatiko:
I have tested this item :white_check_mark: successfully on 7de6d1876b8b2f9a9b2002a1c4e59bc6f6fb2dcd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36186">issues.joomla.org/tracker/joomla-cms/36186</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36186">issues.joomla.org/tracker/joomla-cms/36186</a>.</sub>

khu5h1:
I have tested this item :white_check_mark: successfully on 7de6d1876b8b2f9a9b2002a1c4e59bc6f6fb2dcd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36186">issues.joomla.org/tracker/joomla-cms/36186</a>.</sub>

pritam825:
I have tested this item :white_check_mark: successfully on 7de6d1876b8b2f9a9b2002a1c4e59bc6f6fb2dcd<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36186">issues.joomla.org/tracker/joomla-cms/36186</a>.</sub>

wilsonge:
Thanks!

Issue #36104, opened 2021-11-24
Switch show password icon


sakiss:
### Steps to reproduce the issue
While the password is not visible, the show password icon is the `icon-eye` (eye) while when the password is shown is the `icon-eye-slash` (eye with a cross line).
In most applications out there from our phones to google and facebook is the opposite. So atm it is confusing as it's the opposite to how most users expect it to be.

### Expected result

![j-pwd](https://user-images.githubusercontent.com/1838041/143230247-360cd53e-646b-4cdb-ac27-1f11ddc20747.png)

### Actual result

![j-pwd2](https://user-images.githubusercontent.com/1838041/143230270-f4e16f5e-326d-4e05-95a4-7e9df1d18a2a.png)

### System information (as much as possible)
J4

### Additional comments
This happens in all the login forms.

brianteeman:
https://ux.stackexchange.com/questions/123812/is-google-using-the-password-show-hide-toggle-icon-on-their-login-form-correctly

https://ux.stackexchange.com/questions/1318/should-a-toggle-button-show-its-current-state-or-the-state-to-which-it-will-chan

brianteeman:
Those two links _suggest_ to me that it is a subjective view. The important thing is what does the user think. In this specific case I think its obvious that clicking the icon will change a hidden password to a visible password

sakiss:
I decided to open that issue after i got confused more than once with the show button's current state.
Although it is subjective, users got used to a specific implementation that is applied almost everywhere and and they use it intuitively.

brianteeman:
How could you be confused about the current state when the field will either look like

![image](https://user-images.githubusercontent.com/1296369/143239545-bddaa801-9770-40a7-b1fa-27a3653dba30.png)

or

![image](https://user-images.githubusercontent.com/1296369/143239592-e23ed8f6-c0c2-4fc4-8060-f4e01e38af03.png)


However you are correct that the use of an open and closed icon is not used consistently in Joomla.

sakiss:
> How could you be confused about the current state when the field will either look like

The anticipated state does not match the current one and this is confusing.

brianteeman:
> The anticipated state does not match the current one and this is confusing.

thats a common issue throughout joomla

chmst:
@sakiss 

It seems correct for me on login: Buttons are "call-for-action", not information about the state (icon).

A button says for example "read more" or "hide the content" or "klick me" - and something happens.
[...]

PR #36195, opened 2021-12-04
[4.0] Fix show password icon in login forms #36104

AJAY007L:
Pull Request for Issue #36104.

### Summary of Changes
         In short, This PR makes the show password icon to icon-eye-slash, when the password is not visible and the show 
         password icon changes to icon-eye, when the password is visible.

### Testing Instructions
        see login form on
        1. http://localhost/joomla-cms/administrator/ .
        2. http://localhost/joomla-cms/.
        3. http://localhost/joomla-cms/index.php/login (after installing sample data).
        
         
 

image
text wrong

      

image
text2

        None.
### Note
         This PR made changes on 
        1. http://localhost/joomla-cms/administrator/ login form.
        2. http://localhost/joomla-cms/ login form.
        3. http://localhost/joomla-cms/index.php/login login form.


brianteeman:
For me this is a pointless change. From a straw poll of two sites they both use opposite options. If its changed here now then people will be quite right to submit a pr to reverse it again with the same reasoning. its one of those where 50% of the people will always think its wrong

### Twitter
image

### Facebook
image

PR #36244, opened 2021-12-06
[4.1] Scheduled tasks notes

brianteeman:
If a task has a note then it is displayed below the task name inside a small span (just like in all other components)

If a task does not have a note then a small span is still present but it contains a language string (which is empty)

There is no need for this empty span and there is certainly no need for a language string with no content.

## To test
Create a new task with a note
Create another new task without a note
Check the source code and you will see the empty span on the second task

Apply this PR
Refresh and check the source of the page again. This time there is no empty span on the task without a note and the note is still displayed on the other task.


administrator/components/com_scheduler/tmpl/tasks/default.php
```diff
@@ -201,9 +201,7 @@ class="js-draggable" data-url="<?php echo $saveOrderingUrl; ?>" data-direction="
 							<?php endif; ?>
 
 							<span class="small">
```
Quy: Please move inside the if statement to not have an empty span when no note.

brianteeman: Sorry - that was the whole point of this pr - doh

Quy:
I have tested this item :white_check_mark: successfully on 0ce9a11007ee8792e4b0c64e716efd64bd6dae15<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36244">issues.joomla.org/tracker/joomla-cms/36244</a>.</sub>

richard67:
@brianteeman Couldn't the "COM_SCHEDULER_NO_NOTE" string be removed from the language file? As far as I can see it is not used anymore after the change here.

brianteeman:
@richard67 it is removed

richard67:
Oh, how could I miss that.

alikon:
I have tested this item :white_check_mark: successfully on d80709138bde3c67e7cea2476e73977dcdb9d34e<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36244">issues.joomla.org/tracker/joomla-cms/36244</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36244">issues.joomla.org/tracker/joomla-cms/36244</a>.</sub>

bembelimen:
Thx

brianteeman:
thx

PR #36249, opened 2021-12-07
[4] cannot pass null to trim function in php 8.1

alikon:
Pull Request for Issue # .
cannot pass null to trim function in php 8.1

### Summary of Changes

set a default value

### Testing Instructions
look the log

deprecation from Uri.php
trim() passing null to parameter #1 ($string)

no deprecation

### DIFF
--- a/libraries/src/Uri/Uri.php
+++ b/libraries/src/Uri/Uri.php
@@ -142,11 +142,11 @@ class Uri extends \Joomla\Uri\Uri
 			$uri = static::getInstance();
 			$live_site = ($uri->isSsl()) ? str_replace('http://', 'https://', $config->get('live_site')) : $config->get('live_site');
 
-			if (trim($live_site) != '')
+			if (trim($live_site ?? '') != '')
 			{
 				$uri = static::getInstance($live_site);
 				static::$base['prefix'] = $uri->toString(array('scheme', 'host', 'port'));
-				static::$base['path'] = rtrim($uri->toString(array('path')), '/\\');
+				static::$base['path'] = rtrim($uri->toString(array('path') ?? ''), '/\\');
 
 				if (\defined('JPATH_BASE') && \defined('JPATH_ADMINISTRATOR'))
 				{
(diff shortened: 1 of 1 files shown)


libraries/src/Uri/Uri.php
```diff
@@ -142,11 +142,11 @@ public static function base($pathonly = false)
 			$uri = static::getInstance();
 			$live_site = ($uri->isSsl()) ? str_replace('http://', 'https://', $config->get('live_site')) : $config->get('live_site');
 
-			if (trim($live_site) != '')
+			if (trim($live_site ?? '') != '')
```
Fedik, line 145: Instead doing it, do `$config->get('live_site', '')` above.

`$config->get('<key>', '<default value>')`

libraries/src/Uri/Uri.php
```diff
@@ -142,11 +142,11 @@ public static function base($pathonly = false)
...
 			{
 				$uri = static::getInstance($live_site);
 				static::$base['prefix'] = $uri->toString(array('scheme', 'host', 'port'));
-				static::$base['path'] = rtrim($uri->toString(array('path')), '/\\');
+				static::$base['path'] = rtrim($uri->toString(array('path') ?? ''), '/\\');
```
Fedik, line 149: here 2 issue:
first: you added `??` in wrong place,
second: `$uri->toString()` always return a string, it will return empty string if "nothing found", see `AbstractUri::render()`

libraries/src/Uri/Uri.php
```diff
@@ -177,7 +177,7 @@ public static function base($pathonly = false)
 				// Extra cleanup to remove invalid chars in the URL to prevent injections through broken server implementation
 				$script_name = str_replace(array("'", '"', '<', '>'), array('%27', '%22', '%3C', '%3E'), $script_name);
 
-				static::$base['path'] = rtrim(\dirname($script_name), '/\\');
+				static::$base['path'] = rtrim(\dirname($script_name) ?? '', '/\\');
```
Fedik, line 180: `dirname()` already always return a string

libraries/src/Uri/Uri.php
```diff
@@ -201,7 +201,7 @@ public static function root($pathonly = false, $path = null)
 		{
 			$uri = static::getInstance(static::base());
 			static::$root['prefix'] = $uri->toString(array('scheme', 'host', 'port'));
-			static::$root['path'] = rtrim($uri->toString(array('path')), '/\\');
+			static::$root['path'] = rtrim($uri->toString(array('path') ?? ''), '/\\');
```
Fedik, line 204: same as one of above

alikon:
my worst top 5 pr ever
:smile:

PR #36260, opened 2021-12-08
Add systemtests for php 8.1

HLeithner:
See title


HLeithner:
For unknown reason media manager fails...

Fedik:
I think it may be due to ajax request fail, 
I see there many deprecate notices, one of them could break the body of  ajax response

HLeithner:
it also fails with 7.2 (phpmin) (with the new images, not sure if the last update changed anything since the last build)

.drone.yml
```diff
@@ -152,32 +152,62 @@ steps:
     commands:
       - npm run lint:js
 
-  - name: system-tests-mysql
+  - name: phpmin-system-mysql
     depends_on: [ javascript-cs ]
     image: joomlaprojects/docker-images:systemtests
```
wilsonge, line 195: Looks wrong to me that phpmin and phpmax are running on the same docker image. surely needs to be different docker images to have different php versions?

HLeithner: True have to be systentests81

wilsonge:
Updated this branch a few times as I've merged in some of the 8.1 fixes in case.

wilsonge:
The test fail is super weird. It creates a new folder (NOT via the UI - but by a PHP function (which is also used to create the test-dir that all the tests work from - so it must work)). Then it goes to the UI and tries to rename the folder. But the new sub-folder created doesn't show up in the UI - so it can't rename it and so fails.

FWIW I can confirm that this specific test does work in PHP 8.1 too on my local setup :)

wilsonge:
No figured it out. You changed the install directory to be db folder specific but didn't update https://github.com/joomla/joomla-cms/pull/36260/files#diff-d43be3f113ad505ac2e42c4da89933bbd1e25cee518586ecbf04156bcae55144R44 @HLeithner - just needs that overriding for each DB type now

tests/Codeception/_support/Step/Acceptance/Administrator/Media.php
```diff
@@ -18,6 +18,8 @@
 use Page\Acceptance\Administrator\MediaListPage;
 use PHPUnit\Framework\Assert;
 
+use function print_r;
```
Achraf-dev: function is not used

wilsonge:
@HLeithner still a bit ugly - but this works

wilsonge:
Screenshot 2021-12-17 at 00 17 58

The current test fail matches my local setup results for PHP 8.1 - so I'd say this is now "working as expected"

HLeithner:
At least it works. The deprecation should be fixed before we merge this ;-)

wilsonge:
Can you add an allowed fail for the phpmax system tests? I'd like to get this in and then I can start crossing failures out directly with system test proof

HLeithner:
@wilsonge tests are working now, I also added api tests for postgres and it seems it fails because of a bug https://ci.joomla.org/joomla/joomla-cms/49383/1/27

maybe @richard67 or @alikon can have a look?

I would merge it now and fix es problems later and remove the failure ignore when fixed.

alikon:
yes please merge this now 
we can fix postgresql issues later 
if not by these

- #36362
- #36360
- #36355

PhilETaylor:
I have done a round of fixes tonight, in Joomla-cms Joomla-framework/database and in the unit tests... merge all those and most of the tests will run without deprecated messages

PR #31408, opened 2020-11-15
[4.1] Field group site display enhancement

ceford:
Pull Request for Issue #31385 and related to #31396  .

### Summary of Changes

Code to customise site display of Field Groups

### Testing Instructions

Create some article fields that are not in a field group.
Edit an article and populate those fields.

Test the default behaviour.

Create one or more field groups, the example I use is Ingredients and Procedure.
The PR has added some extra fields to the Options tab of the Field Group data entry form:
	Select Show Title = Yes
	Select a render tag.
Add some fields to those groups.
Edit an article and populate the fields

### Actual result BEFORE applying this Pull Request

The default behaviour is to show all fields in a single bulleted list with no 'title'.

### Expected result AFTER applying this Pull Request

The example I use shows a Recipe with Ingredients as a bulleted list and the Procedure as a numbered list, with titles.

image

### Documentation Changes Required

Help screen will need revision

Issue #31385, opened 2020-11-11
[4.0] [com_fields] Site display dl rather than ul?


ceford:
### Steps to reproduce the issue

See the forum discussion: https://forum.joomla.org/viewtopic.php?f=803&t=982946

Does anyone know why custom fields are set as list items in a bulleted list? The preceding fields (by default) are set as definition terms in a definition list. 

### Expected result

Best would be for custom fields to have same styling as standard fields.

### Actual result

Custom fields are bulleted. Standard fields are not.

### Additional comments

It can be done with overrides and styles. BUT the default ought to be best for novices - all the same.

brianteeman:
It is not about how they are styled but what is the meaning of those lists. A definition list and an unordered list have different semantic meanings. In this case the type of list is semantically correct.

ceford:
For completeness here is a screen shot:

![image](https://user-images.githubusercontent.com/1560390/98893151-c7771b00-2499-11eb-81ac-53a4962e4e2a.png)

So which list is semantically correct? Is it the dl or the ul? And what should we tell users, especially novices, who would like to see a single list?

brianteeman:
Each is semantically correct they are separate lists

ceford:
Can you explain why the first list is a dl and the second is a ul? The only reason I can think of is that dl lists do not contain sub-lists while ul lists may. But I have not found a way to generate a sub-list with the available fields. It seems to me that both are lists of attributes of an article. So why different list types?

brianteeman:
UL is just a list with no specific meaning
DL is a description list of (in this case) the article

see
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

ceford:
I know what they are. What I am asking is why one or other was chosen in this instance. And more specifically why different types of list. It looks like it could be different coders personal preferences. It would make more sense (to me) if both lists were dl or ul. And re-coding the ul as dl would be an incredibly simple change.

200MPHMEDIA:
[...]

PR #36265, opened 2021-12-08
Revised field_group with new base

ceford:
Pull Request for Issue #31385 and related to #31396. Revised version of #31408

Summary of Changes

Added code so that field groups can have alternatives to ul layouts and headings.

Testing Instructions

Apply the patch

Create two field groups:
Ingredients
Set Options / Render Tag to ul
Procedure
Set Options / Render Tag to ol

Create fields with Ingredients as Field Group
with Options / Render Options / Label set to Hide (Optional)
Item 1
Item 2
Item 3

Create fields with Procedure as Field Group
with Options / Render Options / Label set to Hide (Optional)
Stage 1
Stage 2
Stage 3

Edit an article
Fill out the Ingredients form Tab
Water
Tea
Milk
Fill out the Procedure form Tab
Boil the water
Pour the water on the tea
Add the milk

Actual result BEFORE applying this Pull Request

A single unordered list

Expected result AFTER applying this Pull Request

image

Documentation Changes Required

Yes - I will do that when merged.


administrator/components/com_fields/forms/group.xml
```diff
@@ -154,6 +154,60 @@
...
+				<option value="div">div</option>
+			</field>
+			
+			<field
+				name="title_class"
+				type="textarea"
+				label="COM_FIELDS_GROUP_TITLE_TAG_CLASS_LABEL"
+				size="40"
+				validate="CssIdentifier"
+				default="mb-0"
```
brianteeman: is it correct to have a default value and if so should it be bs dependent?

administrator/components/com_fields/src/Model/FieldsModel.php
```diff
@@ -156,7 +156,7 @@ protected function getListQuery()
 
 		// Join over the language
 		$query->select('l.title AS language_title, l.image AS language_image')
-			->join('LEFT', $db->quoteName('#__languages') . ' AS l ON l.lang_code = a.language');
+		->join('LEFT', $db->quoteName('#__languages') . ' AS l ON l.lang_code = a.language');
```
richard67: Please revert this unrelated change since it violates code style rules.

administrator/components/com_fields/src/Model/FieldsModel.php
```diff
@@ -168,14 +168,14 @@ protected function getListQuery()
...
-		$query->select('g.title AS group_title, g.access as group_access, g.state AS group_state, g.note as group_note');
+		$query->select('g.title AS group_title, g.access as group_access, g.state AS group_state, g.note as group_note, g.params as group_params');
 		$query->join('LEFT', '#__fields_groups AS g ON g.id = a.group_id');
 
 		// Filter by context
 		if ($context = $this->getState('filter.context'))
 		{
 			$query->where($db->quoteName('a.context') . ' = :context')
-				->bind(':context', $context);
+			->bind(':context', $context);
```
richard67: Please revert this unrelated change since it violates code style rules.

administrator/components/com_fields/src/Model/FieldsModel.php
```diff
@@ -190,7 +190,7 @@ protected function getListQuery()
 			{
 				$access = (int) $access;
 				$query->where($db->quoteName('a.access') . ' = :access')
-					->bind(':access', $access, ParameterType::INTEGER);
+				->bind(':access', $access, ParameterType::INTEGER);
```
richard67: Please revert this unrelated change since it violates code style rules.

administrator/components/com_fields/forms/group.xml
ceford: Good point. But if the default title tag is p then with no styling the gap between the title and list it entitles is a tad too large. So it looks better and does no harm. Could that be covered in the documentation?

image

brianteeman:
I have tested this item :red_circle: unsuccessfully on 27299d65bbff2baca007d61a2267976eec340e76<br /><br />Tested with blog sample data and the front end displays

Warning: Undefined property: stdClass::$group_params in C:\htdocs\joomla-cms\components\com_fields\layouts\fields\render.php on line 77
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36265">issues.joomla.org/tracker/joomla-cms/36265</a>.</sub>

ceford:
Sorry, I am confused - is that extra tab before ->bind supposed to be there or not?

administrator/components/com_fields/forms/group.xml
brianteeman: really foes against the principle of the frontend being framework agnostic

administrator/components/com_fields/forms/group.xml
```diff
@@ -154,6 +154,60 @@
...
+			</field>
+
+			<field
+				name="title_tag"
+				type="list"
+				label="COM_FIELDS_GROUP_TITLE_TAG_LABEL"
+				default="p"
+				showon="show_title:1[AND]render_tag:ul,ol,p,div"
+				>
+				<option value="p">p</option>
```
brianteeman, line 194: why is this limited to p and div?
why cant i use hx or span for example

administrator/components/com_fields/forms/group.xml
```diff
@@ -154,6 +154,60 @@
...
+				<option value="0">JNO</option>
+				<option value="1">JYES</option>
+			</field>
+
+			<field
+				name="title_tag"
+				type="list"
+				label="COM_FIELDS_GROUP_TITLE_TAG_LABEL"
+				default="p"
+				showon="show_title:1[AND]render_tag:ul,ol,p,div"
```
brianteeman: why not on dl?

ceford:
When fixing the white space in FieldsModel.php I inadvertently reverted the line that caused Brian's test to fail.

dgrammatiko:
This is totally the wrong solution. A proper solution is to have Layouts that are selectable at the field group level. Then, since layouts are overridable, people could create a simple override for every case. In sort, please use the existing platform architecture instead of monkey patching things...

My 2c

administrator/components/com_fields/src/Model/FieldsModel.php
richard67:
```suggestion
			->join('LEFT', $db->quoteName('#__languages') . ' AS l ON l.lang_code = a.language');
```
Since this line belongs to the previous line, i.e. they are one statement, this line needs to be intended by one tab more than the previous one.

administrator/components/com_fields/src/Model/FieldsModel.php
richard67:
```suggestion
				->bind(':context', $context);
```
Since this line belongs to the previous line, i.e. they are one statement, this line needs to be intended by one tab more than the previous one.

administrator/components/com_fields/src/Model/FieldsModel.php
richard67:
```suggestion
				->bind(':access', $access, ParameterType::INTEGER);
```
Same as my 2 previous review comments.

You can find the results of the PHP CS check by following the "details" link of the drone check. Then you select the failed test (PHPCS) at the left hand side and see the report at the right hand side: https://ci.joomla.org/joomla/joomla-cms/49101/1/6

richard67:
> Sorry, I am confused - is that extra tab before ->bind supposed to be there or not?

See my new review comments, this time with suggestions.

brianteeman:
The biggest problem with this PR is that by using default values it changes all existing field groups

Quy:
> Sorry, I am confused - is that extra tab before ->bind supposed to be there or not?

In general, please do not include unrelated changes such as coding style in order to not pollute the PR. It should be done in a separate PR. Thanks.

ceford:
I give up! What seemed like a simple problem a year ago, only having bulleted lists available for fields even when the field is an image, remains the status quo. See #31396<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36265">issues.joomla.org/tracker/joomla-cms/36265</a>.</sub>

richard67:
@ceford Please don’t give up. It was not my intention to block you, I only wanted the unrelated code style changes to be reverted so that drone passes. Sorry if I have confused you with that.

Quy:
@ceford Please don't be discouraged and take it the wrong way. My suggestion is to only help you in your contributions which we appreciate very much!!!

dgrammatiko:
@ceford so I checked the existing code and I found some interesting things

- The templates overrides creator correctly lists the layouts for com_fields:

Screenshot 2021-12-08 at 21 46 46

- so an override is just a click
Screenshot 2021-12-08 at 21 47 01

- then anyone can do whatever they want in that file.

Eg my test code on `templates/cassiopeia/html/layouts/com_fields/fields/render.php` was

<details>
<summary>templates/cassiopeia/html/layouts/com_fields/fields/render.php</summary>

```php
<?php
/**
 * @package     Joomla.Site
 * @subpackage  com_fields
 *
 * @copyright   (C) 2016 Open Source Matters, Inc. <https://www.joomla.org>
 */
defined('_JEXEC') or die;

use Joomla\Component\Fields\Administrator\Helper\FieldsHelper;

// Check if we have all the data
if (!array_key_exists('item', $displayData) || !array_key_exists('context', $displayData))
{
	return;
}

// Setting up for display
$item = $displayData['item'];

if (!$item)
{
	return;
}

$context = $displayData['context'];

if (!$context)
{
	return;
}

$parts     = explode('.', $context);
$component = $parts[0];
$fields    = null;

if (array_key_exists('fields', $displayData))
{
	$fields = $displayData['fields'];
}
else
{
	$fields = $item->jcfields ?: FieldsHelper::getFields($context, $item, true);
}

if (empty($fields))
{
	return;
}

$output = array();

foreach ($fields as $field)
{
	// If the value is empty do nothing
	if (!isset($field->value) || trim($field->value) === '')
	{
		continue;
	}

	$class = $field->name . ' ' . $field->params->get('render_class');
	$layout = $field->params->get('layout', 'render');
	$content = FieldsHelper::render($context, 'field.' . $layout, array('field' => $field));

	// If the content is empty do nothing
	if (trim($content) === '')
	{
		continue;
	}

	$output[] = '<div class="field-entry ' . $class . '">' . $content . '</div>';
}

if (empty($output))
{
	return;
}
?>
<div class="fields-container">
	<?php echo implode("\n", $output); ?>
</div>

```

</details>

Which produces: 
Screenshot 2021-12-08 at 21 50 59

Source HTML code:
Screenshot 2021-12-08 at 21 51 09


In short, the proper way already exists...

ceford:
White space problem fixed.

@brianteeman the title tags were those for which I could see a use case. hx would be complicated by the need to get the hierarchy right and almost certainly extra css. Existing field groups use ul, which is why it is the default. dl would take us back to J3!

@dgrammatiko thanks for the extra explanation. In this case I do not see how using a template override can reproduce the Ingredients / Procedure lists in my testing instructions because there is nothing in the Field Group parameters to tell the layout which tag to use.

dgrammatiko:
Just add in the fields group one, one field named layout where you can select one of the available layouts. That’s the solution, adding fields in the field level is not the right approach

brianteeman:
@ceford the same argument about the title tags heirarchy could have been made about modules yet we have the option there. Trust me if you offer one then you can guarantee users will ask why not the other - especially when it is called a title and yet there are no traditional title tags in the list

brianteeman:
@ceford You have identified a very valid issue and we do need the ability to select a layout for a field group.

image

You can then include some standard group layouts eg  ul, dl as you have already created AND the site owner has the ability to create their own layouts for the group

dgrammatiko:
@laoneo can you have a look here. The aim is to have a layout field in the group level

brianteeman:
@dgrammatiko just needs a new field - basically a clone of https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Form/Field/ComponentlayoutField.php

and then add the layouts

laoneo:
> @laoneo can you have a look here. The aim is to have a layout field in the group level

On the front there is no distinction to render the fields per group yet. Groups are only used in forms for different tabs. This would be a new way to render fields per group and not just flat as it is now.

dgrammatiko:
> This would be a new way to render fields per group and not just flat as it is now.

Could you implement it? My objection here is that the fields try to control the container, which should be considered the wrong approach in any well-designed system. Usually, we want to follow the hierarchy, outer -> inner

Anyways there's https://github.com/joomla/joomla-cms/pull/36275 as an alternative

laoneo:
@ceford would it be ok to switch to ##36275 which is more inline with the Joomla4 way to have as less parameters as possible and do more stuff in layouts?

dgrammatiko:
@ceford I don't want to take over your PR therefore you can use the code from my PR and continue here. I just open the other one to showcase a possible solution (obviously the code needs some more refinement)

ceford:
Take over / switch to #36275 - my first reaction is yes, I would be content for someone else to take this forward. On the other hand I guess you could all do without the distraction. I can see that there is a layout for each tag type so end users don't have to create overrides. I applied the patch and my fields disappear! So there is still some work to do.  Give me a day to see how far I get.

ceford:
@laoneo @dgrammatiko 
Can I have some advice please? Look at lines 440 onwards in plugins/system/fields/fields.php - at that point there is a single list of fields and they are all rendered with the same layout as the first item in the list. So if there are two field groups involved the list needs to be split into two lists that will be rendered separately. Seems a bit clunky but I do not see where else the list might be encoded as an array. So at the moment I propose to split the list there. Good/bad proposal?

laoneo:
@ceford would be better you leave your pr in the initial state, so people can decide which  way to go. And help bringing #36275 to a stable state. So we do not have two pr's with the same code. No need to make your pr a copy of the other. So can you revert your last commit?

ceford:
Revised to use code provided by @dgrammatiko - currently waiting for automated tests to complete. Example outcome:

image

In this example I have left in the label of the first item in each group, just to show items with and without labels. The group title is set as h3. The last item, the image, is a field not in a field group. It uses ul by default. If the bullet is not wanted it could be styled out. Or the image could be put in a group. The default ul appearance, which is a change since J3, is what started this PR a year ago.

ceford:
@laoneo I don't know what to make of your last comment! It was my understanding that @dgrammatiko was leaving this to me with the suggestions that I use his code. That is what I have done. Are you suggesting that I revert to my monkey-patched version? For me, that was more satisfying but it would leave dgrammatiko to complete his version - it does not include group titles.

AndySDH:
> > @laoneo can you have a look here. The aim is to have a layout field in the group level
> 
> On the front there is no distinction to render the fields per group yet. Groups are only used in forms for different tabs. This would be a new way to render fields per group and not just flat as it is now.

Not true, you can do {fieldgroup 5} in your editor to render fields from fieldgroup with ID 5.

plugins/system/fields/fields.php
```diff
@@ -439,15 +439,55 @@ private function display($context, $item, $params, $displayType)
...
+				else
+				{
+					$html .= FieldsHelper::render(
+						$context,
+						$field_group[0]->group_layout ? $field_group[0]->group_layout : 'render',
+						[
+							'item'            => $item,
+							'context'         => $context,
+								'fields'          => $field_group,
+						]
```
PhilETaylor: Indentation 

```suggestion
						[
							'item'            => $item,
							'context'         => $context,
							'fields'          => $field_group,
						]
```

plugins/system/fields/fields.php
```diff
@@ -439,15 +439,55 @@ private function display($context, $item, $params, $displayType)
...
+							'context'         => $context,
+							'fields'          => $field_group,
+						]
+					);
+				}
+				else
+				{
+					$html .= FieldsHelper::render(
+						$context,
+						$field_group[0]->group_layout ? $field_group[0]->group_layout : 'render',
```
PhilETaylor:
```suggestion
						$field_group[0]->group_layout ?: 'render',
```

plugins/system/fields/fields.php
```diff
@@ -439,15 +439,55 @@ private function display($context, $item, $params, $displayType)
...
+			{
+				if ($field->group_title === $last_group)
+				{
+					$field_groups[$last_group][] = $field;
+				}
+				else
+				{
+					$last_group = $field->group_title;
+					$field_groups[$last_group][] = $field;
+				}
```
PhilETaylor:
```suggestion
				foreach ($fields as $field)
				{
					if ($field->group_title !== $last_group)
						$last_group = $field->group_title;
					}
			
					$field_groups[$last_group][] = $field;
				}
```

administrator/components/com_fields/src/Model/FieldsModel.php
```diff
@@ -42,24 +42,24 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
 		if (empty($config['filter_fields']))
 		{
 			$config['filter_fields'] = array(
```
PhilETaylor: Please use short array syntax if you are going to make changes to the array.

ceford: I am not sure what you are asking me to do. Could you elaborate a little?

ceford:
I am not sure what to do about the **This branch is out-of-date** message. Should I be fixing that on my clone and then pushing that to my fork? I merged the latest into my clone and then switched to this pr branch. Should I have merged into my local pr branch instead?

administrator/components/com_fields/src/Model/FieldsModel.php
dgrammatiko: Replace `array()` with the sort one `[]`

richard67:
@ceford As long as there are no conflicts shown in addition, you can ignore that branch out-of-date message.

administrator/components/com_fields/src/Model/FieldsModel.php
ceford: Did I change something (or forget to change something) in the constructor config array? None of the other model constructors use the [] syntax.

dgrammatiko: Since you're changing the indentation of the array you can also switch to the brackets to match the current preferred way of writing an array

PhilETaylor: Short array syntax should be used everywhere from Joomla 4 onwards. The maintainers refused to merge a mass update but favour progressive fixing of the short array syntax coding standards over time. If you change a file with arrays in it, then change the whole file to use short array syntax- as that is the correct standard for Joomla 4 (and the PHP world have been using it for many years!) - all new code should also only ever be merged using short array syntax

ceford: Thank you for that clarification. Done as requested.

ceford:
Are the drone errors something I can fix? Is that the result of the out-of-date message?

richard67:
> Are the drone errors something I can fix? Is that the result of the out-of-date message?

@ceford The answer to both questions is: No. I restart drone now to see if it passes this time.

richard67:
@ceford Drone restart did not help .. now I have used the button to update the branch. Let's see if that helps.

Possibly you created your branch at a time (or when it was once for the 4.0-dev branch then time when it was rebased to 4.10-dev) when in the 4.1-dev branch there was an error in an SQL script for new installations? At least the output from the system tests look as if that was the case.

So or so it was not your fault or your mistake.

richard67:
@ceford After having kicked the drone a few times, it passed. Now in case if you need to make some modifications to that PR, pull the changes from remote before working in order to get the updates from the branch update.

HLeithner:

brianteeman:
@crystalenka something for you. We talked about this last week iirc

drmenzelit:
Thank you for your ideas. 

We discussed it in CMS Maintenance Team and we think that the way is to use overrides / alternative layouts.

PR #36275, opened 2021-12-09
[4.1] Expose group layouts for com_fields

dgrammatiko:
Pull Request is an alternative to https://github.com/joomla/joomla-cms/pull/36265.

### Summary of Changes

Add a layout option on the group level

### DIFF
--- a/administrator/components/com_fields/forms/group.xml
+++ b/administrator/components/com_fields/forms/group.xml
@@ -154,6 +154,20 @@
 				<option value="0">JNO</option>
 				<option value="1">JYES</option>
 			</field>
+
+			<field
+				name="layout"
+				type="filelist"
+				label="COM_FIELDS_FIELD_LAYOUT_LABEL"
+				default="render"
+				directory="components/com_fields/layouts/fields"
+				fileFilter="\.php$"
+				hide_none="true"
+				stripext="true"
+				exclude="render\.php"
+			>
+			</field>
+			<option value="render">JDEFAULT</option>
 		</fieldset>
 	</fields>
 </form>
--- a/components/com_fields/layouts/fields/DIV.php
+++ b/components/com_fields/layouts/fields/DIV.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * @package     Joomla.Site
+ * @subpackage  com_fields
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+defined('_JEXEC') or die;
+
+use Joomla\Component\Fields\Administrator\Helper\FieldsHelper;
+
+// Check if we have all the data
+if (!array_key_exists('item', $displayData) || !array_key_exists('context', $displayData))
+{
+	return;
+}
+
+// Setting up for display
+$item = $displayData['item'];
+
+if (!$item)
+{
+	return;
+}
+
+$context = $displayData['context'];
+
+if (!$context)
+{
+	return;
+}
+
+$parts     = explode('.', $context);
+$component = $parts[0];
+$fields    = null;
+
+if (array_key_exists('fields', $displayData))
+{
+	$fields = $displayData['fields'];
+}
+else
+{
+	$fields = $item->jcfields ?: FieldsHelper::getFields($context, $item, true);
+}
+
+if (empty($fields))
+{
+	return;
+}
+
+$output = array();
+
+foreach ($fields as $field)
+{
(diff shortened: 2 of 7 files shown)


administrator/components/com_fields/forms/group.xml
```diff
@@ -154,6 +154,20 @@
 				<option value="0">JNO</option>
 				<option value="1">JYES</option>
 			</field>
+
+			<field
+				name="layout"
+				type="filelist"
+				label="COM_FIELDS_FIELD_LAYOUT_LABEL"
+				default="render"
+				directory="components/com_fields/layouts/fields"
```
laoneo, line 163: What happens with the overrides in the component and templates? Should be covered also somehow.

brianteeman: that's why I suggested a new fieldtype based on componentlayoutlist

dgrammatiko: > that's why I suggested a new fieldtype based on componentlayoutlist

Unfortunately the `componentLayout` is only for `com_xxx/tmpl` if I read correctly the code

> What happens with the overrides in the component and templates? Should be covered also somehow.

 Overrides of these layouts are still possible on the template due to JLayouts
About `overrides in the component` I'm not sure what you mean but the code in the layouts it's literally the same so it shouldn't break anything. If the option `don't render anything` is on then all these are neglected, right?

dgrammatiko:
I will let @ceford to take over here

laoneo:
I would leave this one open here as alternative. The whole discussion in #36265 is about the old code, so we have a clear history.

dgrammatiko:
ok, reopened

administrator/components/com_fields/forms/group.xml
brianteeman: > > that's why I suggested a new fieldtype based on componentlayoutlist

> Unfortunately the componentLayout is only for com_xxx/tmpl if I read correctly the code

ok "inspired by" then. we also have modulelayoutfield which is similar

components/com_fields/layouts/fields/DIV.php
```diff
@@ -0,0 +1,82 @@
...
+{
+	$fields = $item->jcfields ?: FieldsHelper::getFields($context, $item, true);
+}
+
+if (empty($fields))
+{
+	return;
+}
+
+$output = array();
```
PhilETaylor, line 52:
```suggestion
$output = [];
```

components/com_fields/layouts/fields/DL.php
```diff
@@ -0,0 +1,82 @@
...
+{
+	$fields = $item->jcfields ?: FieldsHelper::getFields($context, $item, true);
+}
+
+if (empty($fields))
+{
+	return;
+}
+
+$output = array();
```
PhilETaylor, line 52:
```suggestion
$output = [];
```

components/com_fields/layouts/fields/DL.php
```diff
@@ -0,0 +1,82 @@
...
+	{
+		continue;
+	}
+
+	$class = $field->name . ' ' . $field->params->get('render_class');
+	$layout = $field->params->get('layout', 'render');
+	$content = FieldsHelper::render($context, 'field.' . $layout, array('field' => $field));
+
+	// If the content is empty do nothing
+	if (trim($content) === '')
```
PhilETaylor, line 67: https://3v4l.org/VDRBp

```suggestion
	if (!trim($content))
```

components/com_fields/layouts/fields/OL.php
```diff
@@ -0,0 +1,82 @@
...
+{
+	$fields = $item->jcfields ?: FieldsHelper::getFields($context, $item, true);
+}
+
+if (empty($fields))
+{
+	return;
+}
+
+$output = array();
```
PhilETaylor, line 52:
```suggestion
$output = [];
```

components/com_fields/layouts/fields/OL.php
```diff
@@ -0,0 +1,82 @@
...
+	{
+		continue;
+	}
+
+	$class = $field->name . ' ' . $field->params->get('render_class');
+	$layout = $field->params->get('layout', 'render');
+	$content = FieldsHelper::render($context, 'field.' . $layout, array('field' => $field));
+
+	// If the content is empty do nothing
+	if (trim($content) === '')
```
PhilETaylor, line 67:
```suggestion
	if (!trim($content))
```

dgrammatiko:
Since @ceford already took over and this code was just to showcase the layout approach, this PR has served it's purpose

PR #36282, opened 2021-12-10
[4] php 8.1 deprecation

alikon:
Pull Request for Issue # .

- PHP Deprecated:  Declaration of Joomla\CMS\Date\Date::format($format, $local = false, $translate = true) should be compatible with DateTime::format(string $format): string in ../joomla-cms/libraries/src/Date/Date.php
- PHP Deprecated:  Declaration of Joomla\CMS\Date\Date::setTimezone($tz) should be compatible with DateTime::setTimezone(DateTimeZone $timezone): DateTime in ../joomla-cms/libraries/src/Date/Date.php  

## Summary of Changes
added annotation `#[\ReturnTypeWillChange]`

### DIFF
--- a/libraries/src/Date/Date.php
+++ b/libraries/src/Date/Date.php
@@ -292,7 +292,7 @@ class Date extends \DateTime
 	 *
 	 * @since   1.7.0
 	 */
-	public function format($format, $local = false, $translate = true)
+	public function format($format, $local = false, $translate = true) :string
 	{
 		if ($translate)
 		{
@@ -409,7 +409,7 @@ class Date extends \DateTime
 	 * @since   1.7.0
 	 * @note    This method can't be type hinted due to a PHP bug: https://bugs.php.net/bug.php?id=61483
 	 */
-	public function setTimezone($tz)
+	public function setTimezone($tz) :Date
 	{
 		$this->tz = $tz;
 
--- a/libraries/src/Date/Date.php
+++ b/libraries/src/Date/Date.php
@@ -292,7 +292,7 @@ class Date extends \DateTime
 	 *
 	 * @since   1.7.0
 	 */
-	public function format($format, $local = false, $translate = true) :string
+	public function format($format, $local = false, $translate = true): string
 	{
 		if ($translate)
 		{
--- a/libraries/src/Date/Date.php
+++ b/libraries/src/Date/Date.php
@@ -409,7 +409,7 @@ class Date extends \DateTime
 	 * @since   1.7.0
 	 * @note    This method can't be type hinted due to a PHP bug: https://bugs.php.net/bug.php?id=61483
 	 */
-	public function setTimezone($tz) :Date
+	public function setTimezone($tz): Date
 	{
 		$this->tz = $tz;
 
--- a/libraries/src/Date/Date.php
+++ b/libraries/src/Date/Date.php
@@ -292,7 +292,7 @@ class Date extends \DateTime
 	 *
 	 * @since   1.7.0
 	 */
-        #[\ReturnTypeWillChange]
+	#[\ReturnTypeWillChange]
 	public function format($format, $local = false, $translate = true)
 	{
 		if ($translate)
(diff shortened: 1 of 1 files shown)


Fedik:
Unfortunately it will lead to Fatal error, for everyone who extends our `Date` class.
We need a better idea, how to solve it.

libraries/src/Date/Date.php
```diff
@@ -292,7 +292,7 @@ public function calendar($format, $local = false, $translate = true)
 	 *
 	 * @since   1.7.0
 	 */
-	public function format($format, $local = false, $translate = true)
+	public function format($format, $local = false, $translate = true) :string
```
PhilETaylor: Also lets start off right with the correct code style. The colon comes right after the closing bracket, then a space then the type. As per https://wiki.php.net/rfc/return_types

E.g:
```suggestion
	public function format($format, $local = false, $translate = true): string
```

libraries/src/Date/Date.php
```diff
@@ -409,7 +409,7 @@ public function monthToString($month, $abbr = false)
 	 * @since   1.7.0
 	 * @note    This method can't be type hinted due to a PHP bug: https://bugs.php.net/bug.php?id=61483
 	 */
-	public function setTimezone($tz)
+	public function setTimezone($tz) :Date
```
PhilETaylor:
```suggestion
	public function setTimezone($tz): Date
```

wilsonge:
For J4 we're just going to have to add the annotations. As mentioned this is a b/c break for anything extending these classes. Annotations should work back to PHP 7.2 as well.

libraries/src/Date/Date.php
```diff
@@ -409,7 +409,7 @@ public function monthToString($month, $abbr = false)
 	 * @since   1.7.0
 	 * @note    This method can't be type hinted due to a PHP bug: https://bugs.php.net/bug.php?id=61483
 	 */
-	public function setTimezone($tz)
```
alikon, line 412: did you mean @wilsonge 
```php
#[\ReturnTypeWillChange]
public function setTimezone($tz)
```

wilsonge: yes

wilsonge: obviously it's not ideal. but considering we need to support PHP 7.2+ it's pretty much essential

alikon: ok thanks

libraries/src/Date/Date.php
```diff
@@ -292,7 +292,7 @@ public function calendar($format, $local = false, $translate = true)
 	 *
 	 * @since   1.7.0
 	 */
-	public function format($format, $local = false, $translate = true)
+	public function format($format, $local = false, $translate = true): string
```
PhilETaylor:
```suggestion
        #[\ReturnTypeWillChange]
	public function format($format, $local = false, $translate = true)
```

alikon: yes this too

libraries/src/Date/Date.php
```diff
@@ -292,6 +292,7 @@ public function calendar($format, $local = false, $translate = true)
 	 *
 	 * @since   1.7.0
 	 */
+        #[\ReturnTypeWillChange]
```
PhilETaylor: This needs to be tab indented and not space indented. 
```suggestion
	#[\ReturnTypeWillChange]
```

PhilETaylor: FILE: /********/src/libraries/src/Date/Date.php
45s
48	----------------------------------------------------------------------
45s
49	FOUND 1 ERROR AFFECTING 1 LINE
45s
50	----------------------------------------------------------------------
45s
51	 295 | ERROR | [x] Tabs must be used to indent lines; spaces are not
45s
52	     |       |     allowed
45s
53	     |       |     (Generic.WhiteSpace.DisallowSpaceIndent.SpacesUsed)
45s
54	----------------------------------------------------------------------
45s
55	PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
45s
56	----------------------------------------------------------------------

richard67: > This needs to be tab indented and not space indented.

Yes, that’s the one which breaks the PHP CS check in drone.

wilsonge:
Thanks!

PR #36295, opened 2021-12-12
[4.0] Update npm dependencies

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- Major versions update for few  dependencies
- Fixing the commander B/C breaks
- Removed files from diff and mediaelement as the dist folder doesn't provide them

### Testing Instructions
- Run `npm install` before applying this PR and make a copy of the `media` folder
- Apply this PR, run `npm install` and compare the `media` with the one you stored in the previous step. They should be identical (apart from the missing `diff.min.js` and `css/mejs-controls.png` which were removed upstream)

- Check that the media manager works correctly: Double-clicking on a folder navigates to that folder

Some changes regarding the `new Regex()`:
- Check that editing the colour in the Atum template works as expected
- Check that custom fields change context when selecting a different category
- Check that installing extensions from the JED is working correctly
- Check that Editing a user by changing the username and te password works as expected

- Also check that all all npm commands still work
- And finally confirm that CSS/JS is not broken, front end/ back end

No. @wilsonge @bembelimen after 4.0.5 could you have some people testing this one? Most dependencies are way behind...
Also please consider adding either dependabot or renovate bot so PRs are automatically created on each new update of any dependency


dgrammatiko:
The failure in codeception is a real bug. Try navigating to `administrator/index.php?option=com_media&path=local-images:/banners` the resolved URL once the page is loaded is `administrator/index.php?option=com_media&path=local-images:/banners&path=local-images:/banners` (there are 2 URL parameters named `path`)

package.json
```diff
@@ -31,77 +31,78 @@
...
+    "metismenujs": "^1.3.1",
     "punycode": "^2.1.1",
     "qrcode-generator": "^1.4.4",
     "roboto-fontface": "^0.10.0",
     "short-and-sweet": "^1.0.4",
-    "skipto": "^4.1.0",
+    "skipto": "^4.1.3",
     "tinymce": "^5.10.2",
-    "vue": "^3.1.2",
+    "vue": "next",
```
laoneo: Even when this is correct, I would stay on 3.1. It looks wrong to have here a string while the others are all versions.

build/build-modules-js/settings.json
```diff
@@ -525,9 +524,7 @@
           "build/mediaelement-flash-audio.swf": "js/mediaelement-flash-audio.swf",
           "build/mediaelement-flash-video-hls.swf": "js/mediaelement-flash-video-hls.swf",
           "build/mediaelement-flash-video-mdash.swf": "js/mediaelement-flash-video-mdash.swf",
-          "build/mediaelement-flash-video.swf": "js/mediaelement-flash-video.swf",
```
brianteeman, line 528: why are these removed?

dgrammatiko: Apparently you can't copy a non existing file

dgrammatiko: I stand corrected, only the png file was dropped

wilsonge:
Thanks!

build/build-modules-js/settings.json
richard67: @dgrammatiko The files '/media/vendor/diff/js/diff.min.js' and '/media/vendor/diff/js/diff.min.js.gz' are added to the list of deleted files and folders in script.php when I run my tool because now after this PR here has been merged, there is no minified and gzipped file anymore in the full package, but the unminified and unzipped '/media/vendor/diff/js/diff.js' file is in the package. Could you make a PR to fix that so that file is minified and gzipped again?

richard67:
@dgrammatiko The files '/media/vendor/diff/js/diff.min.js' and '/media/vendor/diff/js/diff.min.js.gz' are added to the list of deleted files and folders in script.php when I run my tool because now after this PR here has been merged, there is no minified and gzipped file anymore in the full package, but the unminified and unzipped '/media/vendor/diff/js/diff.js' file is in the package. Could you make a PR to fix that so that file is minified and gzipped again?

richard67:
Or is it ok as it is and the files '/media/vendor/diff/js/diff.min.js' and '/media/vendor/diff/js/diff.min.js.gz' should be deleted on update?

dgrammatiko:
@richard67 https://github.com/joomla/joomla-cms/pull/36493 should bring back the minified file

PR #36318, opened 2021-12-15
[4.0] Support to set the location when saving a category through the API

laoneo:
### Summary of Changes
This PR adds the possibility to define the location of the category while saving it through the API. A location can be one of the following values:
- before
- after
- first-child
- last-child

If `before` or `after` is used it is advisable to also set the location_reference parameter which defines the sibling the category should be located.

### Testing Instructions
Run the following curl command:
```
curl --location --request POST 'https://[your J4 website]/api/index.php/v1/content/categories' \
--header 'Content-Type: application/json' \
--header 'X-Joomla-Token: [your user token]' \
--data-raw '{
    "title": "test",
    "location": "first-child",
    "published": "1",
    "language": "*"
}'
```

The new category appears as last item in the list in the back end UI.

The new category appears as first item in the list in the back end UI.

The new `location` and `location_reference` arguments need to be documented.

### DIFF
--- a/api/components/com_categories/src/Controller/CategoriesController.php
+++ b/api/components/com_categories/src/Controller/CategoriesController.php
@@ -86,7 +86,15 @@ class CategoriesController extends ApiController
 		/** @var Category $category */
 		$category = $this->getModel('Category')->getTable('Category');
 		$category->load($recordId);
-		$category->setLocation($category->parent_id, $data['location']);
+
+		$reference = $category->parent_id;
+
+		if (!empty($data['location_reference']))
+		{
+			$reference = $data['location_reference'];
+		}
+
+		$category->setLocation($reference, $data['location']);
 		$category->store();
 
 		return $recordId;
--- a/api/components/com_categories/src/Controller/CategoriesController.php
+++ b/api/components/com_categories/src/Controller/CategoriesController.php
@@ -85,13 +85,13 @@ class CategoriesController extends ApiController
 
 		/** @var Category $category */
 		$category = $this->getModel('Category')->getTable('Category');
-		$category->load((int)$recordId);
+		$category->load((int) $recordId);
 
 		$reference = $category->parent_id;
 
 		if (!empty($data['location_reference']))
 		{
-			$reference = (int)$data['location_reference'];
+			$reference = (int) $data['location_reference'];
 		}
 
 		$category->setLocation($reference, $data['location']);
--- a/api/components/com_categories/src/Controller/CategoriesController.php
+++ b/api/components/com_categories/src/Controller/CategoriesController.php
@@ -12,6 +12,7 @@ namespace Joomla\Component\Categories\Api\Controller;
 \defined('_JEXEC') or die;
 
 use Joomla\CMS\MVC\Controller\ApiController;
+use Joomla\CMS\Table\Category;
 
 /**
  * The categories controller
@@ -55,6 +56,40 @@ class CategoriesController extends ApiController
 		$this->input->set('extension', $extension);
 
 		return $data;
+	}
+
+	/**
+	 * Method to save a record.
+	 *
+	 * @param   integer  $recordKey  The primary key of the item (if exists)
+	 *
+	 * @return  integer  The record ID on success, false on failure
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function save($recordKey = null)
+	{
+		$recordId = parent::save($recordKey);
+
+		if (!$recordId)
+		{
+			return $recordId;
+		}
+
+		$data = $this->input->get('data', json_decode($this->input->json->getRaw(), true), 'array');
+
+		if (empty($data['location']))
+		{
+			return $recordId;
+		}
+
+		/** @var Category $category */
+		$category = $this->getModel('Category')->getTable('Category');
+		$category->load($recordId);
+		$category->setLocation($category->parent_id, $data['location']);
+		$category->store();
+
+		return $recordId;
 	}
 
 	/**
--- a/api/components/com_categories/src/Controller/CategoriesController.php
+++ b/api/components/com_categories/src/Controller/CategoriesController.php
@@ -85,13 +85,13 @@ class CategoriesController extends ApiController
 
 		/** @var Category $category */
 		$category = $this->getModel('Category')->getTable('Category');
-		$category->load($recordId);
+		$category->load((int)$recordId);
 
 		$reference = $category->parent_id;
 
 		if (!empty($data['location_reference']))
 		{
-			$reference = $data['location_reference'];
+			$reference = (int)$data['location_reference'];
 		}
 
 		$category->setLocation($reference, $data['location']);


laoneo:
@alikon , @khu5h1 can you guys give this one here a test too? Similar procedure as the other ones. Thanks!!

api/components/com_categories/src/Controller/CategoriesController.php
```diff
@@ -57,6 +58,48 @@ protected function preprocessSaveData(array $data): array
 		return $data;
 	}
 
+	/**
+	 * Method to save a record.
+	 *
+	 * @param   integer  $recordKey  The primary key of the item (if exists)
+	 *
+	 * @return  integer  The record ID on success, false on failure
```
PhilETaylor, line 66: false is not an integer - so is it `integer|bool` or is `false on failure` incorrect ?

api/components/com_categories/src/Controller/CategoriesController.php
```diff
@@ -57,6 +58,48 @@ protected function preprocessSaveData(array $data): array
 		return $data;
 	}
 
+	/**
+	 * Method to save a record.
+	 *
+	 * @param   integer  $recordKey  The primary key of the item (if exists)
```
PhilETaylor, line 64:
```suggestion
	 * @param   integer|null  $recordKey  The primary key of the item (if exists)
```

laoneo: This is copied from the base class https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/MVC/Controller/ApiController.php#L404

api/components/com_categories/src/Controller/CategoriesController.php
laoneo: This is copied from the base class https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/MVC/Controller/ApiController.php#L406

PhilETaylor: So what? It is still incorrect! And needs fixing!

It's time people stop copying and pasting the crap and actually started implementing the correct code instead. Else Joomla will never improve.

api/components/com_categories/src/Controller/CategoriesController.php
PhilETaylor: It is still factually incorrect and needs fixing in this PR before it can be accepted

api/components/com_categories/src/Controller/CategoriesController.php
laoneo: It is not about copy paste, it is about that an overriding method needs a proper docblock in Joomla. This is IMO the failure. In Java world nobody is doing that and you can inherit docs https://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javadoc.html#inheritingcomments.

So it is since ever a common practice to copy the base block in Joomla core, even when it is wrong. I want to have here a statement from maintainers, that an overriding function can have a different docblock that the base function. Before that I do not change it.

PhilETaylor: You should fix the blatantly wrong docblock in the base class now that you are aware it's wrong too! 

I dont care what Java does, it's clearly wrong here and Joomla is full of wrong type hints and it's about time we started putting them all right before we move to a fully typed PHP 8.1 like standard in the future!  

This PR, and all PRs that have factually incorrect code should be rejected. Period.

PhilETaylor: It's also missing the docs for the multiple exceptions that can be thrown #facepalm

laoneo: Then make a pr in the base class and fix the doc block there. I will then adapt. But having different doc blocks for the same function feels wrong to me.

PhilETaylor: In the time you have spent attempting to defend the indefensible you could have fixed your PR and included the base class changes in your PR. #facepalm 

This kind of gatekeeping has to stop. Quality has to improve. I make no excuse for holding the code to high standards.

PhilETaylor: https://github.com/joomla/joomla-cms/pull/36331/files

PhilETaylor: If your docblock is going to be identical then you should be using only this, and nothing more:

```
/**
 * {@inheritdoc}
 */
```

laoneo: Copy the base docblock is what I had the impression is the standard in Joomla because the CS checker doesn't understand {@inheritdoc} as you can see here https://github.com/joomla/joomla-cms/pull/36333. I'm not blindly change here something which was a common practice. This should be done carefully to keep the consistency in the code base.

I'm not gatekeeping or try to put you down. I really appreciate your effort in improving the code documentation as you are very good on that.

PhilETaylor: "consistency" LMAO.. yup lets keep everything "wrong" for "consistency" sake... #facepalm 

If the Code Style checker doesn't understand the correct code style - then that also should be fixed, NOT ignored and "worked around" with invalid and factually incorrect code ...

PhilETaylor: which tool exactly is moaning about `@inheritdoc` ?

alikon:
i guess we should do the same for endpoint POST `{{base_path}}/api/index.php/v1/menus/{app}`

alikon:
I have tested this item :white_check_mark: successfully on b51641dcec33fd929887dc3e38a00730a9f3dd4d<br /><br />CS issue apart<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36318">issues.joomla.org/tracker/joomla-cms/36318</a>.</sub>

api/components/com_categories/src/Controller/CategoriesController.php
laoneo: The code style checker Joomla uses. Run the following command in a terminal in the root of your joomla installation after you have changed to the inherit flag:
`./libraries/vendor/bin/phpcs --extensions=php -p --standard=libraries/vendor/joomla/cms-coding-standards/lib/Joomla-CMS .`

PhilETaylor: looks like a dead end https://github.com/squizlabs/PHP_CodeSniffer/pull/214

laoneo: There is a way
https://github.com/squizlabs/PHP_CodeSniffer/issues/2770#issuecomment-762532979
https://github.com/squizlabs/PHP_CodeSniffer/issues/2770#issuecomment-673133190

laoneo: And you can also use `:man_facepalming:` as emoji to better express your feelings. Hope your head doesn't hurt about so many facepalms you had to do to day :v:

laoneo: Marking this as resolved as fixing doc blocks is out of the scope of this pr.

api/components/com_categories/src/Controller/CategoriesController.php
laoneo: Marking this also as resolved, see discussion above.

laoneo:
@alikon if accepted, then I can do also for menus.

api/components/com_categories/src/Controller/CategoriesController.php
PhilETaylor: Rubbish! You are adding new code, that code should be correct. This PR should be rejected. To accept this PR would show that Joomla doesn't care about code quality as it's blatantly factually incorrect. Docblocks ARE within scope of a PR adding new methods!

HLeithner:
While testing this it seems that api can by-pass the alias naming restrictions is this intended?

```
curl --location --request POST 'https://[domain]/api/index.php/v1/content/categories' \
--header 'Content-Type: application/json' \
--header 'X-Joomla-Token: [key]' \
--data-raw '{"location_reference":2,
    "title": "test before Uncategorised",
    "location": "before",
    "published": "1",
    "language": "*"
}
```
executing this command 2 times ends, in one item before and one (the second call) as last item... I know unrelated to this PR...

What the rest is working find but could we get some variable validation in this function because passing location and location_reference without input validation is wrong.

laoneo:
Validation is done on the model level, so there is no reason to do it on the controller level as well. So if there is something missing on the validation side, it must be fixed in the model. But I casted to int in the controller to be double save.

laoneo:
@HLeithner can you mark your test as successful please.

HLeithner:
I have tested this item :white_check_mark: successfully on 323a846d9af22de9c5452d54952e5d81faab85e1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36318">issues.joomla.org/tracker/joomla-cms/36318</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 323a846d9af22de9c5452d54952e5d81faab85e1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36318">issues.joomla.org/tracker/joomla-cms/36318</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36318">issues.joomla.org/tracker/joomla-cms/36318</a>.</sub>

wilsonge:
Thanks!

PR #36331, opened 2021-12-17
[4.0] Correct docblock

PhilETaylor:
https://github.com/joomla/joomla-cms/pull/36318#discussion_r771104989

### Summary of Changes

Correct docblock to be factually correct so that people can copy and paste the correct docblock in the future instead of making excuses on why they are adding false information to their code.

### DIFF
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -401,7 +401,7 @@ class ApiController extends BaseController
 	/**
 	 * Method to save a record.
 	 *
-	 * @param   integer  $recordKey  The primary key of the item (if exists)
+	 * @param   integer|null  $recordKey  The primary key of the item (if exists)
 	 *
 	 * @return  integer|boolean  The record ID on success, false on failure
 	 *
--- a/libraries/src/MVC/Controller/ApiController.php
+++ b/libraries/src/MVC/Controller/ApiController.php
@@ -403,7 +403,12 @@ class ApiController extends BaseController
 	 *
 	 * @param   integer  $recordKey  The primary key of the item (if exists)
 	 *
-	 * @return  integer  The record ID on success, false on failure
+	 * @return  integer|boolean  The record ID on success, false on failure
+	 *
+	 * @throws InvalidParameterException
+	 * @throws \RuntimeException
+	 * @throws \Exception
+	 * @throws \Joomla\CMS\MVC\Controller\Exception\CheckinCheckout
 	 *
 	 * @since   4.0.0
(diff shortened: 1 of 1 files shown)


libraries/src/MVC/Controller/ApiController.php
```diff
@@ -401,9 +401,14 @@ public function edit()
 	/**
 	 * Method to save a record.
 	 *
-	 * @param   integer  $recordKey  The primary key of the item (if exists)
+	 * @param   integer|null  $recordKey  The primary key of the item (if exists)
 	 *
-	 * @return  integer  The record ID on success, false on failure
+	 * @return  integer|boolean  The record ID on success, false on failure
```
HLeithner, line 406:
```suggestion
	 * @return  mixed  The record ID on success
```

HLeithner, line 406:
```suggestion
	 * @return  integer|string  The record ID on success
```

The correct comment depends on what have been set to the `state` which is usually an integer, but it's also possible (not so common) that it is a string. On the end it's maybe an array? because we support multi column keys? But I don't know how they are saved in the state...

the only thing what could not be is `null` and `boolean` (actually it could but doesn't make sense for a table key).

So in the end we don't know and we don't validate the parameter and can't guaranty that it is a integer.

So my current suggestion is to set it to `mixed`.

PhilETaylor: And this is why Joomla will never improve. "Mixed" really is a cop out generally whenever used and should never be used. :-( it's more an admission that the type is unknown and guessed 

The more and more I attempt to get these right, the more I see that Joomla is never going to be able to cope.

These are docblocks- if we cannot determine types for docblocks we're have absolutely no chance in the future of adding real param and return types in PHP (like modern PHP apps do already)...

HLeithner: I'm completely on your opinion but that matter that the function returns more then an integer.

So we can change this in 5.0. At this time I'm not sure how because the state is a `Registry` and if I'm not wrong we don't have filters...

On the other side the concept of not having an `integer` as `id` is broken on many points in the cms (only checked BaseController) so we also can force it to integer in this function and add return type `integer`.

laoneo: This works for core, but extensions can work differently and especially in the API context where ID's are often UID's, we have to be careful. I guess we have to stop thinking only in database based integer relations. This has nothing to do with a bad architecture, more with a paradigm change where we need to start thinking outside of our closed DB monolith.

Issue #34840, opened 2021-07-19
[4][CLI] - reuse of code from CMS


alikon:
### Steps to reproduce the issue
try to reuse code from CLI, for example notify superuser 

```php
// Push a notification to the site's super users
/** @var MessageModel $messageModel */
$messageModel = $this->app->bootComponent('com_messages')->getMVCFactory()->createModel('Message', 'Administrator');

//.... do something

$_SERVER['HTTP_HOST'] = 'fake.url';
$username= 'fakeuser';

// Load com_message's language
$language = Factory::getLanguage();
$language->load('com_messages', JPATH_ADMINISTRATOR, 'en-GB', false, true);
$language->load('com_messages', JPATH_ADMINISTRATOR, null, true);

$messageModel->notifySuperUsers(
	Text::_('PLG_SYSTEM_PRIVACYCONSENT_NOTIFICATION_USER_PRIVACY_EXPIRED_SUBJECT'),
	Text::sprintf('PLG_SYSTEM_PRIVACYCONSENT_NOTIFICATION_USER_PRIVACY_EXPIRED_MESSAGE', $username)
);
```

### Expected result

able to call `$messageModel->notifySuperUsers`   from CLI

### Actual result
![image](https://user-images.githubusercontent.com/181681/126177635-fb2d1dac-a8bb-4452-850c-b2ca7d5f9916.png)

### System information (as much as possible)
j4

### Additional comments
as a quick dirty fix
i've  just added this method to ConsoleApplication.php https://github.com/joomla/joomla-cms/blob/4.0-dev/libraries/src/Application/ConsoleApplication.php

```php
	public static function getRouter($name = null, array $options = array())
	{
		$app = Factory::getApplication();

		if (!isset($name))
		{
			$name = $app->getName();
		}

		$options['mode'] = $app->get('sef');

		return Router::getInstance($name, $options);
	}
```

is there a better way ?

alikon:
fixed with #36332

PR #36332, opened 2021-12-17
[4.0] Make routing work in console applications

laoneo:
### Summary of Changes
When working on the console, then routing in Joomla doesn't work propperly. Especially the widely used `Route::_()` code is crashing. This pr adds a function to the console class to get the router object and fills the `HTTP_HOST` server variable with the live site parameter.

### Testing Instructions
- Enable SEF in the Joomla configuration
- Create an article
- Create a blog menu item with the category of the article selected
- Add the following code to the file /libraries/src/Console/ListUserCommand.php somewhere in the `doExecute` function:  
  `$this->ioStyle->success(\JRoute::_('index.php?option=com_content&view=article&id=1&catid=8'));`
  Replace the ids 1 and 8 with the ones from your joomla site.
- In the terminal run `php cli/joomla.php user:list`

Error in the terminal `Call to undefined method Joomla\CMS\Application\ConsoleApplication::getRouter()`

You see a green OK text with the SEF route to the article.

### DIFF
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -451,15 +451,9 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 	 */
 	public static function getRouter($name = null, array $options = array())
 	{
-		if (!isset($name))
+		if (empty($name))
 		{
 			throw new InvalidArgumentException('A router name must be set in console application.');
-		}
-
-		// Fallback to site router when console router is requested
-		if (Factory::getApplication()->getName() === $name)
-		{
-			$name = 'site';
 		}
 
 		$options['mode'] = Factory::getApplication()->get('sef');
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -10,6 +10,7 @@ namespace Joomla\CMS\Application;
 
 \defined('JPATH_PLATFORM') or die;
 
+use InvalidArgumentException;
 use Joomla\CMS\Console;
 use Joomla\CMS\Extension\ExtensionManagerTrait;
 use Joomla\CMS\Factory;
@@ -449,12 +450,18 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 	 * @return  Router
 	 *
 	 * @since   __DEPLOY_VERSION__
+	 * @throws  \InvalidArgumentException
 	 */
 	public static function getRouter($name = null, array $options = array())
 	{
-		if (!isset($name) || Factory::getApplication()->getName() === $name)
+		if (!isset($name))
 		{
-			// Use site as default as there is no console router
+			throw new InvalidArgumentException('A router name must be set in console application.');
+		}
+
+		// Fallback to site router when console router is fetched
+		if (Factory::getApplication()->getName() === $name)
+		{
 			$name = 'site';
 		}
 
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -456,7 +456,7 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 			throw new InvalidArgumentException('A router name must be set in console application.');
 		}
 
-		// Fallback to site router when console router is fetched
+		// Fallback to site router when console router is requested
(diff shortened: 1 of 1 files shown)


alikon:
should fix #34840 i guess... i'll test later

laoneo:
@alikon yes it should fix it

alikon:
I have tested this item :white_check_mark: successfully on ef15ffe799e25a3d5cb67f9f8904bc3ece23d46c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36332">issues.joomla.org/tracker/joomla-cms/36332</a>.</sub>

astridx:
I have tested this item :white_check_mark: successfully on ef15ffe799e25a3d5cb67f9f8904bc3ece23d46c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36332">issues.joomla.org/tracker/joomla-cms/36332</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36332">issues.joomla.org/tracker/joomla-cms/36332</a>.</sub>

wilsonge:
Sorry this isn't correct (in my opinion) - I accept some of this is somewhat more subjective than other PR's though. There's kinda two parts here so I'm going to address each separately

First of all the `getRouter` method. Happy to add that to the application so we can use `\Joomla\CMS\Router\Route::link`. However in terms of the default router even though there is no site router - I think it's wrong to load the site router explicitly as the default. I'd rather it errors out by default if it's called with the wrong data rather than be wrongly relied on.

In terms of the setting of the `HTTP_HOST` I'm very sure that overloading server side variables is wrong - plus we still end up with potentially bad behaviour here when `live_site` isn't set (I'm still yet to set that on any of my sites) - and we just end up with that being null and still getting junk URLs out on each of those instances. This needs a proper fix (I'm not 100% sure how) where I guess the application has the ability to inject the base domain into the `Uri::getInstance` method which would override the HTTP_HOST. Possibly by taking advantage of the WebApplication's https://github.com/joomla-framework/application/blob/2.0-dev/src/AbstractWebApplication.php#L1001-L1004 variables and doing some sort of similar config in the console application so we have consistency over both applications. Open to ideas on the best way forward there - but the idea being we need consistency over both web and console and to treat them as equals rather than bodging around the console.

laoneo:
> First of all the `getRouter` method. Happy to add that to the application so we can use `\Joomla\CMS\Router\Route::link`. However in terms of the default router even though there is no site router - I think it's wrong to load the site router explicitly as the default. I'd rather it errors out by default if it's called with the wrong data rather than be wrongly relied on.

I changed the code, so when no name is set that that an exception is thrown. But I left the code in place when a cli router is requested that it uses the site router as this is the case when `Route::_()` is called.

> I'm very sure that overloading server side variables is wrong

There is no other way to get the web site url than from the live site in console applications. But I removed it and now every extension dev has to set it in the console command by itself. Anyway, for me it is important to have routing work in console applications.

wilsonge:
> I changed the code, so when no name is set that that an exception is thrown. But I left the code in place when a cli router is requested that it uses the site router as this is the case when Route::_() is called.

But the entire point of Route::_ has always been it's been application specific. That's why we introduced Router::link to start with in 3.7 so we could do cross-application routing.

laoneo:
Ok, removed the fallback. All I did was to remove some new code since RTC, So @wilsonge can you consider it now ready to merge?

wilsonge:
Thanks!

I think for 4.1 (as technically it's a low impact b/c break) we might also want to consider putting `getRouter` into the `\Joomla\CMS\Application\CMSApplicationInterface` so typehints work nicer - but this is definitely all we can do for the 4.0-dev branch.

laoneo:
Definitely, the whole getRouter stuff needs some love. @bembelimen can you ping me when 4.0-dev is merged into 4.1? So I will do a followup pr.

laoneo:
@wilsonge made a 4.1 pr for the interface changes in #36631. But https://github.com/joomla/joomla-cms/pull/36630 should be merged first.

PR #36354, opened 2021-12-19
[4.1] Add dragula autoscroll

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Add Auto scroll in the draggable lists, eg drag an item and the list will automatically scroll up/down if you reached the bottom/top of the list.

### Testing Instructions

Apply the PR and check the `administrator/index.php?option=com_plugins&view=plugins` to see if the list auto scrolls when the dragged item reaches to the bottom of the page

### DIFF
--- a/build/build-modules-js/settings.json
+++ b/build/build-modules-js/settings.json
@@ -344,14 +344,14 @@
       "dom-autoscroller": {
         "name": "dragula-scroller",
         "js": {
-          "dist/dom-autoscroller.js": "js/dragula-scroller.js",
-          "dist/dom-autoscroller.min.js": "js/dragula-scroller.min.js"
+          "dist/dom-autoscroller.js": "js/dom-autoscroller.js",
+          "dist/dom-autoscroller.min.js": "js/dom-autoscroller.min.js"
         },
         "provideAssets": [
           {
             "name": "dragula-scroller",
             "type": "script",
-            "uri": "dragula-scroller.min.js",
+            "uri": "dom-autoscroller.min.js",
             "attributes": {
               "defer": true
             }
--- a/build/build-modules-js/settings.json
+++ b/build/build-modules-js/settings.json
@@ -341,6 +341,25 @@
         "dependencies": [],
         "licenseFilename": "license"
       },
+      "dom-autoscroller": {
+        "name": "dragula-scroller",
+        "js": {
+          "dist/dom-autoscroller.js": "js/dragula-scroller.js",
+          "dist/dom-autoscroller.min.js": "js/dragula-scroller.min.js"
+        },
+        "provideAssets": [
+          {
+            "name": "dragula-scroller",
+            "type": "script",
+            "uri": "dragula-scroller.min.js",
+            "attributes": {
+              "defer": true
+            }
+          }
+        ],
+        "dependencies": [],
+        "licenseFilename": "LICENSE"
+      },
       "focus-visible": {
         "name": "focus-visible",
         "js": {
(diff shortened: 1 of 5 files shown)


brianteeman:
nice find

build/build-modules-js/settings.json
```diff
@@ -341,6 +341,25 @@
         "dependencies": [],
         "licenseFilename": "license"
       },
+      "dom-autoscroller": {
+        "name": "dragula-scroller",
+        "js": {
+          "dist/dom-autoscroller.js": "js/dragula-scroller.js",
```
brianteeman: ?? is it good practice to rename the script?

dgrammatiko: Sure, the build script is allowing this so the project names things as it finds more appropriate. I mean if the core provides some script then devs should use that version. If they want to roll their own dependencies that's also fine as long as none of their pages doesn't load some scripts from the core and the exact same from their own distribution. In short, this shouldn't be a problem if devs are not mixing the same script from different folders (the naming won't help in such case, they're destined to fail either way)

dgrammatiko: Never mind, I already renamed the files to match the source

PhocaCz:
I have tested this item :red_circle: unsuccessfully on e8773893886b6ca01e42dae8bd874d8ee2268263<br /><br />Hi,

it does work for standard items (e.g. articles) but e.g. on Linux, the scrolling starts only in specific position, very hard to find it - in fact it is very hard to start the autoscroll


https://i.imgur.com/GOqpBcL.gif

In subforms, it does not work entirely.

Jan<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36354">issues.joomla.org/tracker/joomla-cms/36354</a>.</sub>

dgrammatiko:
> In subforms, it does not work entirely.

Subforms are NOT using the same drag and drop script, so that's expected

PhocaCz:
>Subforms are NOT using the same drag and drop script, so that's expected

Ok, as you can see on my GIF, I get problems to sort on my Linux, so will be good, if other can test it in other environments.

brianteeman:
I have tested this item :red_circle: unsuccessfully on e8773893886b6ca01e42dae8bd874d8ee2268263<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36354">issues.joomla.org/tracker/joomla-cms/36354</a>.</sub>

brianteeman:
The drag scroll is very very slow
It is not longer possible to drop except at the last position<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36354">issues.joomla.org/tracker/joomla-cms/36354</a>.</sub>

brianteeman:
Note: i tested the codepen which worked fine

PhocaCz:
> The drag scroll is very very slow

I have the same behaviour on Linux (KDE), it starts to be faster but only in specific position and it is hard to find this position.

PhocaCz:
>let's use the max speed param

For me, there is no change :-( ... it is still slow until I find randomly the place where it starts to work (but as written above, it is really hard to find this place)

dgrammatiko:
I'm putting this in draft mode. Will revisit later

dgrammatiko:
Closing here as the solution is different than the approach in this PR

PR #36360, opened 2021-12-20
[4.0] PHP 8.1 deprecation notices administrator/components

frostmakk:
Pull Request for Issue # .

### Summary of Changes
Patching PHP 8.1 deprecation notices where null is provided instead of 'something'.

### Testing Instructions
Code review or checking logs.

Deprecation notices

No deprecation notices from these files.

Probably not.


alikon:
I have tested this item :white_check_mark: successfully on 9137ca6c1bbf42ccab78789f4da71c06e76e69bc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36360">issues.joomla.org/tracker/joomla-cms/36360</a>.</sub>

administrator/components/com_mails/src/Model/TemplateModel.php
```diff
@@ -114,12 +114,21 @@ public function getForm($data = array(), $loadData = true)
 			$form->removeField('copyto', 'params');
 		}
 
-		if (!trim($params->get('attachment_folder')))
+		if ($params->get('attachment_folder') === null)
```
HLeithner: I would expect that `$params` is a `JRegistry` and the `get` method expects a "default" parameter.

administrator/components/com_menus/tmpl/items/default_batch_body.php
```diff
@@ -31,7 +31,7 @@
 }
 ?>
 <div class="p-3">
-	<?php if (strlen($menuType) && $menuType != '*') : ?>
+	<?php if (strlen($menuType ?? '') && $menuType != '*') : ?>
```
HLeithner: `getUserState` accepts a second parameter for the default value

administrator/components/com_menus/tmpl/items/default_batch_footer.php
```diff
@@ -18,7 +18,7 @@
 <button type="button" class="btn btn-secondary" onclick="document.getElementById('batch-menu-id').value='';document.getElementById('batch-access').value='';document.getElementById('batch-language-id').value=''" data-bs-dismiss="modal">
 	<?php echo Text::_('JCANCEL'); ?>
 </button>
-<?php if ((strlen($menuType) && $menuType != '*' && $clientId == 0) || ($published > 0 && $clientId == 1)): ?>
+<?php if ((strlen($menuType ?? '') && $menuType != '*' && $clientId == 0) || ($published > 0 && $clientId == 1)): ?>
```
HLeithner: same here add a second parameter to the `getUserState` call

HLeithner (changes requested):

administrator/components/com_mails/src/Model/TemplateModel.php
frostmakk: Since the if statement is negated, wouldn't that resolve to false and the program going in the wrong direction?

HLeithner: sure

HLeithner: it's even simpler 
```suggestion
		if (!trim($params->get('attachment_folder', ''))
```
and remove the else block please

frostmakk: I don't get it. Isn't this originally a null trap? 
If the attachment_folder params does not exist it result in a negated false, and the form field is removed.
If we put in a default value, would this not resolve to false, and the form field not being removed?
Or have I stared myselves blind?

wilsonge:
Thanks!

PR #36364, opened 2021-12-21
[4.0] codemirror update

brianteeman:
Updates codemirror to 5.65.0

brace-folding addon: Fix broken folding on lines with both braces and square brackets.
vim bindings: Support g0, g$, g<Arrow>.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2747,9 +2747,9 @@
       }
     },
     "node_modules/codemirror": {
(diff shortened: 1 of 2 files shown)


wilsonge:
If you can fix the conflicts here in the package lock I'll make sure this is merged

brianteeman:
conflict fixed as requested

wilsonge:
Thanks!

PR #36388, opened 2021-12-24
Dynamic hide Table columns (help please)

brianteeman:
This is a fork and continuation of @PhilETaylor , now closed, original pr

### Summary of Changes
Decide for yourself which columns to display for any table. The button shows an updating count of how many columns are shown. You can NOT hide the _title_ column or the row select checkbox as this would make the table useless.

The choice of which column to display is stored in local storage so it is persistent for you in your browser for that site. 
tables2

This should be added to every table and it can also be added by extension developers to their own table by just adding one line to the existing web assets.

```
// add the script
/** @var \Joomla\CMS\WebAsset\WebAssetManager $wa */
$wa = $this->document->getWebAssetManager();
$wa->useScript('table.columns');
```

### Testing Instructions
 For the purposes of this pr it has **only** been applies to com_content articles and com_content featured articles

As there is new js you will need to `run npm ci`

### Known Issue HELP PLEASE
The code from @philetaylor for local storage is not working 100% and I have no idea why. Its as if its being applied to a row instead of a column. See video below. ALL HELP on this part greatly appreciated. The js is almost meaningless to me so I can't fix it myself.

tables3


brianteeman:
@fedik any ideas on the js bug?

Fedik:
Cannot say from code review, need to debug.
Probably something wrong with some querySelector or some index.

Will try to check later.

Fedik:
The answer that the js is a horror, I have made PR with updates.

PhilETaylor:
> The answer that the js is a horror

I have never pretended to be a professional JavaScript developer, and @dgrammatiko provided some of the code for me and I butchered it further.

Fedik:
I do not blame anyone, I just state the reason ;)

PhilETaylor:
As a side note - Joomla is meant to be a content manager, but not up to 50% of the pixels on the screen, are nothing to do with the content of the page... we need to address this eventually... 

oie_oie_glitters

brianteeman:
@PhilETaylor can you please remove that gif as the flashing is triggering

@Fedik Thanks very much - now I just need to update it with the code to preserve the title column.

build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,173 @@
...
+  // Look for table.table-columns-js with id or data-name attribute
+  document.querySelectorAll('table.table-columns-js').forEach(($table) => {
+    const tableName = ($table.dataset.name ? $table.dataset.name : $table.id).toLowerCase();
+
+    // Skip unnamed table
+    if (!tableName) {
+      return;
+    }
+
+    new TableColumns($table, tableName);
```
Fedik, line 172: for eslint add:
```javascript
/* eslint-disable-next-line no-new */
new TableColumns($table, tableName);
```

administrator/components/com_content/tmpl/featured/default.php
```diff
@@ -108,7 +112,7 @@
 						<?php echo Text::_('JGLOBAL_NO_MATCHING_RESULTS'); ?>
 					</div>
 				<?php else : ?>
-					<table class="table itemList" id="articleList">
+					<table class="table itemList table-columns-js" id="articleList" data-name="articles">
```
Fedik, line 115: Does "articles" and "featured" table have the same amount and order of columns?
If not, then better use another "name", or it will lead to confusion.
The "name" used for storage (saving the state).

brianteeman: yes they do.
BUT the original js created the "name" for storage dynamically which I think is much better
AND as there will only ever be one table on the page I dont think it needs to have the table-columns-js class or data-name

It should be enough to just have the webasset and a table imho

Fedik: > AND as there will only ever be one table 

you never know, it for safety :)

> the original js created the "name" for storage

It also for safety, in case if there multiple tables.
Well, you can try to ignore a name attribute, it will try to pick the table ID, 
but I did no checked how they unique across different views

Fedik: btw, old "name" for storage was based on page title, and it have an unknown behavior for different languages and translations

brianteeman: the original code was
```
    pagens() {
      return `joomla-tableoptions-${document.querySelector('.page-title')
        .innerText
        .trim()
        .replace(/\s'\//, '-')
        .toLowerCase()}`;
    },
```

that should be enough shouldnt it?

Fedik: you can try if you like, I do not mind.
add instead of 
```
const tableName = ($table.dataset.name ? $table.dataset.name : $table.id).toLowerCase();
```
this:
```
const tableName = document.querySelector('.page-title').textContent.trim()
        .replace(/\s'\//, '-')
        .toLowerCase();
```

Fedik: hm, no, better like this:
```
const tableName = $table.dataset.name ? $table.dataset.name : document.querySelector('.page-title').textContent.trim()
        .replace(/\s'\//, '-')
        .toLowerCase();
```
so developers will be able to define name if it will be need.

brianteeman: lol - I just wrote that myself and came back to post it for your approval

dgrammatiko: My 2c: `document.head.querySelector('title')`

brianteeman:
Closing here - will submit as a new pr shortly

administrator/components/com_content/tmpl/featured/default.php
Fedik: @dgrammatiko no, that will be different, I also thought about that

PR #36338, opened 2021-12-17
[4.1] Enable Composer plugins - dealerdirect/phpcodesniffer-composer-installer

PhilETaylor:
Composer 2.2 brings with it new security features, one of these is to disable composer plugins for security with `allow-plugins`

Composer 2.2 will be a LTS Version

Use `composer self-update --preview` to try the latest prerelease version to test this PR.
Use `composer self-update --stable` to go back to stable releases..

https://github.com/composer/composer/releases/tag/2.2.0-RC1
https://getcomposer.org/doc/06-config.md#allow-plugins

When installing Joomla 4 using composer 2.2 you will get this:

Screenshot 2021-12-17 at 16 31 31

Composer plugins should be disabled by default, unless there is a compelling reason for them to run. 

Unless this PR is merged, you will be asked each time if you want to run this plugin or not.

PR #36399, opened 2021-12-24
[3] Enable Composer plugins - dealerdirect/phpcodesniffer-composer-installer

PhilETaylor:
Joomla 3.10 version of https://github.com/joomla/joomla-cms/pull/36338 in case we can make this change sooner..


PhilETaylor:
drone failures unrelated to PR

HLeithner:
We don't use dealerdirect/phpcodesniffer-composer-installer in j3 or do I miss something?

PhilETaylor:
4.1 https://github.com/joomla/joomla-cms/pull/36338

PR #36404, opened 2021-12-24
[4] Ensure message is defined

PhilETaylor:
### Summary of Changes

There is an edge case, where this method can be called with a set of ids, that gets pruned, leaving no ids to feature, which in turn rules a code path that leaves the redirect `$message` undefined. 

### Testing Instructions

Code review. Best viewed with no whitespace https://github.com/joomla/joomla-cms/pull/36404/files?diff=unified&w=1

`$message` is undefined. 

return early after identifying that we have no `$ids` to process.

None

### DIFF
--- a/administrator/components/com_content/src/Controller/ArticlesController.php
+++ b/administrator/components/com_content/src/Controller/ArticlesController.php
@@ -86,6 +86,8 @@ class ArticlesController extends AdminController
 		if (empty($ids))
 		{
 			$this->app->enqueueMessage(Text::_('JERROR_NO_ITEMS_SELECTED'), 'error');
+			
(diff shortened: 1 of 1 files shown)


Quy:
I have tested this item :white_check_mark: successfully on ae6d1f1f011222afbdfa19f049df99752b9b6618<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36404">issues.joomla.org/tracker/joomla-cms/36404</a>.</sub>

wilsonge:
Thanks!

PR #36406, opened 2021-12-24
[4.0] Correct return type

PhilETaylor:
Code review

This method does not return a string, like the other `getQuickiconContent` methods in Joomla, this one also returns `void` and should be documented as such.


wilsonge:
Thanks!

PR #36408, opened 2021-12-24
[4.0] Various modern cleanup of ModulesHelper

PhilETaylor:
This started out as Correctly return an array as per the documented return type instead of void,  and then turned into a general code cleanup.

Can be merged on review.


administrator/components/com_modules/src/Helper/ModulesHelper.php
```diff
@@ -160,11 +164,13 @@ public static function getTemplates($clientId = 0, $state = '', $template = '')
...
+	 * @since  1.6
 	 */
-	public static function getModules($clientId)
+	public static function getModules(int $clientId): array
 	{
-		$db    = Factory::getDbo();
+		$db    = Factory::getContainer()->get('DatabaseDriver');
 		$query = $db->getQuery(true)
 			->select('element AS value, name AS text')
 			->from('#__extensions as e')
```
Quy, line 176: Please remove `(int)` in next line.

PhilETaylor: Done.

PR #36409, opened 2021-12-24
[4.0] Fix array to string conversion

PhilETaylor:
Code Review.

At this location `$pks` is an array, and you are trying to `sprintf` it as if it were a string, which would raise a warning/notice (see https://3v4l.org/lnOBl) 

We know its going to be an array at this point as `$pks    = ArrayHelper::toInteger($pks);` returns an array. 

Ive chosen implode, but json_encode would have been just as good I guess. No one has ever seen this error in real life I doubt 

In PHP 8 this would be a `Warning` whereas before it would be a `Notice` and I bet it will become a `Fatal Error` in later PHP versions probably... who knows... 

https://3v4l.org/lnOBl

### DIFF
--- a/libraries/src/Table/Nested.php
+++ b/libraries/src/Table/Nested.php
@@ -969,7 +969,7 @@ class Nested extends Table
 			// Nothing to set publishing state on, return false.
 			else
 			{
-				$e = new \UnexpectedValueException(sprintf('%s::publish(%s, %d, %d) empty.', \get_class($this), $pks, $state, $userId));
+				$e = new \UnexpectedValueException(sprintf('%s::publish(%s, %d, %d) empty.', \get_class($this), implode(',', $pks), $state, $userId));
 				$this->setError($e);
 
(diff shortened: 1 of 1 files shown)


wilsonge:
Obviously this is fine. But can I persuade you to redo this one targeting the 3.10 branch please. As it's going to be relevant there and it will be merged up.

PhilETaylor:
This fix is not needed on Joomla 3 as the code is different and Joomla 3 is using `$pk[0]` and not the whole array of `$pk` like Joomla 4 does. 

The line in Joomla 3 is this:

https://github.com/joomla/joomla-cms/blob/d55ab7f9ddd92b19e519745c92f698ad5505cd9a/libraries/src/Table/Nested.php#L938

wilsonge:
Doh! I failed to read properly. Thanks!

PhilETaylor:
No worries - thanks.

PR #36412, opened 2021-12-24
[4] Return value type is not compatible with declared

PhilETaylor:
Code review

A bunch of return type fixes to bring code in line with docs and docs in line with code. 

And some short array syntax while I was in the same file.


wilsonge:
Thanks!

PR #36417, opened 2021-12-24
[4] Delete lots more code... 

PhilETaylor:
Code review

### DIFF
--- a/libraries/src/Application/SiteApplication.php
+++ b/libraries/src/Application/SiteApplication.php
@@ -145,7 +145,6 @@ final class SiteApplication extends CMSApplication
 
 		// Set up the params
 		$document = $this->getDocument();
-		$router   = static::getRouter();
 		$params   = $this->getParams();
 
 		// Register the document object with Factory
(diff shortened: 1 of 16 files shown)


libraries/src/Application/SiteApplication.php
```diff
@@ -154,10 +153,6 @@ public function dispatch($component = null)
 		switch ($document->getType())
 		{
 			case 'html':
-				// Get language
-				$lang_code = $this->getLanguage()->getTag();
-				$languages = LanguageHelper::getLanguages('lang_code');
```
wilsonge, line 159: I'm scared about merging this one on review alone because this sets up the language caches before the system plugins are booted at the end of this method. Absolutely not disputing the result is unused btw. Just unsure if there's any system plugins that rely on that behaviour.

PhilETaylor: therefore lets delete `$lang_code` and `$languages` local vars as they are not used in this method, leaving the function calls

wilsonge: Good by me. Happy if you add a TODO as well to investigate removal. Just I'd leave it for a separate PR where we have some human tests.

PhilETaylor: Im happy with just

```
// Set up the language
LanguageHelper::getLanguages('lang_code');
```

as the getters in ` $this->getLanguage()->getTag()` are just getting object properties and not running anything special like `LanguageHelper::getLanguages('lang_code');` does.

PhilETaylor: Done.

wilsonge:
Thanks!

PR #36419, opened 2021-12-25
4.1 dev

Magnytu2:
Hello, would it be possible to develop a calendar frontend module for Joomla?
The objective is to natively retrieve the different Joomla dates (modification date, publication date, ...), but also the "date" fields that would be created. So we could easily create events and be able to view them in frontend in the format of a mini calendar.
Capture d’écran 2021-12-25 à 14 58 24


brianteeman:
This is not a pull request. Please close it and create an issue instead

PR #36422, opened 2021-12-26
Simple class names in articles

impressionestudio:

### Summary of Changes
The proposed changes affect the class names that relate to an article. The purpose is to have simple, short and easy to remember class names in order to create more complex article layouts in an easier way  (with the help of a few changes that I will also propose for custom fields).

Specifically:
- When new class names are added, the old and not useful class names remained for compatibility reasons.
- The old and not useful class names are separated with the special character | so the developers will distinguish them and not use them.
- (The special character is allowed according to https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors.)

### Notes:
- Some guys recommended to use BEM for naming the classes and initially I was using BEM in the proposed changes. But I cancelled using BEM because I don't find it useful. Using BEM, the class names would remain complex and very long.
Also, I think that BEM changed a little some of the naming rules and I think that there are some inconsistencies in the documentation. For example, the class "search-form__button_size_m" should be "search-form__button_size-m".

For many years, I create websites having high quality templates from scratch (only with Joomla of course) and I don't use purchased templates. So I have faced many cases and requirements and I believe that the proposed changes make things way much better.

### DIFF
--- a/components/com_content/tmpl/article/default.php
+++ b/components/com_content/tmpl/article/default.php
@@ -98,7 +98,7 @@ $isExpired         = !is_null($this->item->publish_down) && $this->item->publish
 	<?php if (isset ($this->item->toc)) :
 		echo $this->item->toc;
 	endif; ?>
-	<div itemprop="articleBody" class="article | com-content-article__body">
+	<div itemprop="articleBody" class="content | com-content-article__body">
 		<?php echo $this->item->text; ?>
 	</div>
 
--- a/components/com_content/tmpl/article/default.php
+++ b/components/com_content/tmpl/article/default.php
@@ -33,10 +33,10 @@ $currentDate       = Factory::getDate()->format('Y-m-d H:i:s');
 $isNotPublishedYet = $this->item->publish_up > $currentDate;
 $isExpired         = !is_null($this->item->publish_down) && $this->item->publish_down < $currentDate;
 ?>
-<div class="com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
+<div class="article <?php echo $this->pageclass_sfx; ?> | com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
 	<meta itemprop="inLanguage" content="<?php echo ($this->item->language === '*') ? Factory::getApplication()->get('language') : $this->item->language; ?>">
 	<?php if ($this->params->get('show_page_heading')) : ?>
-	<div class="page-header">
+	<div class="title | page-header">
 		<h1> <?php echo $this->escape($this->params->get('page_heading')); ?> </h1>
 	</div>
 	<?php endif;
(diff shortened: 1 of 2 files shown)


components/com_content/tmpl/article/default.php
```diff
@@ -33,10 +33,10 @@
 $isNotPublishedYet = $this->item->publish_up > $currentDate;
 $isExpired         = !is_null($this->item->publish_down) && $this->item->publish_down < $currentDate;
 ?>
-<div class="com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
+<div class="article <?php echo $this->pageclass_sfx; ?> | com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
```
impressionestudio, line 36: The class "com-content-article" is too long. Also, we already know that we are using the article component and article view from the classes "com_content" and "view-article" in the body element.

components/com_content/tmpl/article/default.php
```diff
@@ -33,10 +33,10 @@
...
 	<meta itemprop="inLanguage" content="<?php echo ($this->item->language === '*') ? Factory::getApplication()->get('language') : $this->item->language; ?>">
 	<?php if ($this->params->get('show_page_heading')) : ?>
-	<div class="page-header">
+	<div class="title | page-header">
```
impressionestudio, line 39: The class "page-header" should change to "title" since it represents the title field of an article.

components/com_content/tmpl/article/default.php
```diff
@@ -50,18 +50,18 @@
 	|| $params->get('show_hits') || $params->get('show_category') || $params->get('show_parent_category') || $params->get('show_author') || $assocParam; ?>
 
 	<?php if ($params->get('show_title')) : ?>
-	<div class="page-header">
+	<div class="title | page-header">
```
impressionestudio, line 53: The class "page-header" should change to "title" since it represents the title field of an article.

components/com_content/tmpl/article/default.php
```diff
@@ -50,18 +50,18 @@
...
 		<<?php echo $htag; ?> itemprop="headline">
 			<?php echo $this->escape($this->item->title); ?>
 		</<?php echo $htag; ?>>
 		<?php if ($this->item->state == ContentComponent::CONDITION_UNPUBLISHED) : ?>
-			<span class="badge bg-warning text-light"><?php echo Text::_('JUNPUBLISHED'); ?></span>
+			<span class="badge | bg-warning text-light"><?php echo Text::_('JUNPUBLISHED'); ?></span>
```
impressionestudio, line 58: The classes "bg-warning" and "text-light" shouldn't be there at all. It is an obligation of the template to give styling to an element using only the basic class and not additional classes that are not needed for other templates.

components/com_content/tmpl/article/default.php
```diff
@@ -98,7 +98,7 @@
 	<?php if (isset ($this->item->toc)) :
 		echo $this->item->toc;
 	endif; ?>
-	<div itemprop="articleBody" class="article | com-content-article__body">
+	<div itemprop="articleBody" class="content | com-content-article__body">
```
impressionestudio, line 101: The class "com-content-article__body" can change to "content" since it represents the text in the content tab of an article. Of course, content is also the images of an article and the custom fields, but to keep things simple, let's accept this as a small exception.

brianteeman:
using the two classes will create confusion and not provide the backwards compatibility you expect. The pipe character does not do anything.  Both classes are applied

image

impressionestudio:
I know both classes are applied. That is the purpose. The pipe is there only to separate the new classes from the old ones (or deprecated) without the need of a documentation. The point is for developers to know the meaning of the pipe and to start using the classes before the pipe and ignore the classes after it. Adding many classes does not mean that all will apply a style to the element. The template will decide which one will use. Of course, at the beginning, the already created templates (like Cassiopeia) will keep using the classes after the pipe but as they update they will use the classes before the pipe.

brianteeman:
Why would they even bother. Seems to me that the only reason for this change is to satisfy your own personal design ideas and as had been said by many people before - that is why we have template overrides.

impressionestudio:
Actually, the changes I suggest are not only different class names. There are many cases where classes do not exist at all, or even more, there are no tags for adding the class. For example, in the info block, there are parts that need special css tricks in order to modify. I guess your answer will be still the same: "override the template", but I keep facing these disadvantages so often and I don't feel good to see the new Joomla 4 to continue having these disadvantages. It is like having a Ferrari engine inside a white old car of 80s. Furthermore, the overrides are good for special needs and not for everyday styling and also for not so many files.

All I want is to propose something that I consider good for Joomla. If the rest Joomlers don't agree, then at least I will have tried for what I believe is better for Joomla.

brianteeman:
I dont understand why you dont want to use the tools provided - template overrides?
Nor do I see any advantages of these changes? The only changes proposed here are renaming

impressionestudio:
What I understand from your side is that you don't want some random strangers (like me) to suggest crazy ideas and ruin Joomla. I agree with that. But on the other hand, who would come here to use this Github thing and mess with the code of many files and spend time if at least it was not for serious reason?

brianteeman:
But you dont explain the reasons

impressionestudio:
Here are some reasons:
1. Joomla 4 is a general improvement of Joomla 3. Why an improvement to the HTML code (including some bad class names, some missing classes and some missing tags) not to be included in this improvement?
2. When I create a simple website that doesn't need overriding but simple CSS styling, I don't want to see these inconsistent classes like: com-content-article__body, page-header, item-image, article-info... Sometimes it is article, sometimes it is page, sometimes it is item!
3. I believe that overrides are for special and hard cases and I have done overrides many times with great result and satisfaction. But I don't want to create permanent overrides on maybe 15 files just for using them in a simple website. I would like the new version of Joomla to provide the basic needs.
4. Finally, how can I be sure that during all these years of Joomla 4 life, there will be no code change/upgrade to all these overridden files? I want the override to be an exception, so the other files to keep upgrading.

PR #36431, opened 2021-12-27
[4] composer update maximebf/debugbar to 1.17

alikon:
update maximebf/debugbar to 1.17  
https://github.com/maximebf/php-debugbar/releases/tag/v1.17.3


brianteeman:
if only we had tooling to manage this automaticallty

wilsonge:
Thanks!

PR #36432, opened 2021-12-27
Ensure returned getAuthorisedViewLevels array has only unique values

aDaneInSpain:
I stumbled over this on a site I was debugging. The system sends the following query:

```
SELECT `folder` AS `type`,`element` AS `name`,`params` AS `params`,`extension_id` AS `id`

  FROM #__extensions

  WHERE enabled = 1 
  AND type = 'plugin' 
  AND state IN (0,1) 
  AND access IN (1,1,5)

  ORDER BY ordering
```

Notice the ```AND access IN (1,1,5)``` the extra 1 comes from the fact that this method hardcodes the 1 on line 1027 and then later also finds in the database. This does not really break anything but is just proper data management. For instance if someone wanted to count how many groups have access to something and used this method and counted the entries would get the answer 3 and not 2. It also leads to "wasted" time with developers who are curious where this comes from and spends time hunting down the source.

Pull Request for Issue # .

### Summary of Changes

Added array_unique before returning values

### Testing Instructions

Not totally sure how to test this as I am sure this method is called many many places. But if you enable debug and look at executed queries and find one that starts with:

```
SELECT `folder` AS `type`,`element` AS `name`,`params` AS `params`,`extension_id` AS `id`
```
Then ensure the parenthesis in ACCESS IN statement only has unique values

There could be duplicates

There can not be duplicates

No

### DIFF
--- a/libraries/src/Access/Access.php
+++ b/libraries/src/Access/Access.php
@@ -1044,9 +1044,9 @@ class Access
 					}
 				}
 			}
-			
+	
 			$authorised = array_unique($authorised);
-			
+
 			return $authorised;
 		}
 
@@ -1071,7 +1071,7 @@ class Access
 				}
 			}
 		}
-		
+
 		$authorised = array_unique($authorised);
 
 		return $authorised;
--- a/libraries/src/Access/Access.php
+++ b/libraries/src/Access/Access.php
@@ -1044,7 +1044,9 @@ class Access
 					}
 				}
 			}
-
+			
+			$authorised = array_unique($authorised);
+			
 			return $authorised;
 		}
 
@@ -1069,6 +1071,8 @@ class Access
 				}
 			}
 		}
+		
+		$authorised = array_unique($authorised);
 
 		return $authorised;
 	}
--- a/libraries/src/Access/Access.php
+++ b/libraries/src/Access/Access.php
@@ -1044,7 +1044,7 @@ class Access
 					}
 				}
 			}
-	
+
 			$authorised = array_unique($authorised);
 
 			return $authorised;
--- a/libraries/src/Access/Access.php
+++ b/libraries/src/Access/Access.php
@@ -1045,9 +1045,7 @@ class Access
 				}
 			}
 
-			$authorised = array_unique($authorised);
-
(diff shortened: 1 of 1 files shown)


libraries/src/Access/Access.php
```diff
@@ -1070,6 +1072,8 @@ public static function getAuthorisedViewLevels($userId)
 			}
 		}
 
+		$authorised = array_unique($authorised);
+
```
dgrammatiko: Can you combine these 2 line into one? Like `return array_unique($authorised);`

Same couple lines above

dgrammatiko (changes requested):

libraries/src/Access/Access.php
aDaneInSpain: Sure. I was not a 100% sure if the coding standards prohibited modifying a variable on the return line as some coding standards goes against this.

wilsonge (approved):
Makes sense to me

Quy:
I have tested this item :white_check_mark: successfully on 474b76c809c02bccbdc81627db3cabdc6af8d468<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36432">issues.joomla.org/tracker/joomla-cms/36432</a>.</sub>

dgrammatiko:
I have tested this item :white_check_mark: successfully on 474b76c809c02bccbdc81627db3cabdc6af8d468<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36432">issues.joomla.org/tracker/joomla-cms/36432</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36432">issues.joomla.org/tracker/joomla-cms/36432</a>.</sub>

bembelimen:
Thx

PR #36433, opened 2021-12-27
Add KEY to table structure to prevent filesort

aDaneInSpain:
This is more to open a discussion, I guess. When reviewing performance issues with a site, I noticed that even loading the core language selector triggers a filesort in MariaDB which is not ideal. As there will only ever be a few languages on a site this will most likely never be a huge performance issue, but every little optimization we can do, we should do, right? In order to not have a filesort for the query:

```
SELECT *
  FROM #__languages
  WHERE published=1
  ORDER BY ordering ASC
```
We need to add a composite key for both published and ordering. 

This does speed up loading very minutely and if done throughout the app so that at least there are keys for core functionality would have a real impact (although small) on performance.

### Summary of Changes
Add KEY to table structure to prevent filesort

### DIFF
--- a/installation/sql/mysql/base.sql
+++ b/installation/sql/mysql/base.sql
@@ -412,7 +412,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
   UNIQUE KEY `idx_sef` (`sef`),
   UNIQUE KEY `idx_langcode` (`lang_code`),
   KEY `idx_access` (`access`),
-  KEY `idx_ordering` (`ordering`)
+  KEY `idx_ordering` (`ordering`),
 KEY `idx_published_ordering` (`published`, `ordering`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
 
--- a/installation/sql/mysql/base.sql
+++ b/installation/sql/mysql/base.sql
@@ -413,7 +413,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
   UNIQUE KEY `idx_langcode` (`lang_code`),
   KEY `idx_access` (`access`),
   KEY `idx_ordering` (`ordering`)
-	KEY `idx_published_ordering` (`published`, `ordering`)
+KEY `idx_published_ordering` (`published`, `ordering`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
 
 --
--- a/installation/sql/mysql/base.sql
+++ b/installation/sql/mysql/base.sql
@@ -413,6 +413,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
   UNIQUE KEY `idx_langcode` (`lang_code`),
   KEY `idx_access` (`access`),
   KEY `idx_ordering` (`ordering`)
+	KEY `idx_published_ordering` (`published`, `ordering`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
 
 --


installation/sql/mysql/base.sql
```diff
@@ -413,6 +413,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
   UNIQUE KEY `idx_langcode` (`lang_code`),
   KEY `idx_access` (`access`),
   KEY `idx_ordering` (`ordering`)
+	KEY `idx_published_ordering` (`published`, `ordering`)
```
brianteeman:
```suggestion
KEY `idx_published_ordering` (`published`, `ordering`)
```

codestyle

brianteeman:
Does this change not also have to be made to the postgresql install AND the updates?

aDaneInSpain:
Yes it would need to be made there, and I would imagine in some migration scripts as well to add the keys to already running installations. And I assume there are multiple instances of these in other tables, so it would be a bigger job, that I would be willing to take on. But I would like to hear if there is any input, objections or reasons why it is not already implemented first.

installation/sql/mysql/base.sql
```diff
@@ -413,6 +413,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
   UNIQUE KEY `idx_langcode` (`lang_code`),
   KEY `idx_access` (`access`),
   KEY `idx_ordering` (`ordering`)
+KEY `idx_published_ordering` (`published`, `ordering`)
```
PhilETaylor, line 416:
```suggestion
 KEY `idx_published_ordering` (`published`, `ordering`)
```

PhilETaylor:
This will also need to be added to the upgrade SQL's but I guess @richard67 will be able to comment more on that. 

docs: https://dev.mysql.com/doc/refman/8.0/en/order-by-optimization.html

PhilETaylor:
and if this is needed on this table, maybe its needed on other tables too?

richard67:
It needs to do the same for PostgreSQL, and it needs to add an update script for each MySQL and PostgreSQL.

installation/sql/mysql/base.sql
```diff
@@ -413,6 +413,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
...
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
```
PhilETaylor:
```suggestion
  KEY `idx_access` (`access`),
  KEY `idx_ordering` (`ordering`)
  KEY `idx_published_ordering` (`published`, `ordering`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
```

PhilETaylor:
This also breaks the unit tests... because a comma is missing... 


InstallCest installJoomla mysql fail

installation/sql/mysql/base.sql
```diff
@@ -413,6 +413,7 @@ CREATE TABLE IF NOT EXISTS `#__languages` (
   UNIQUE KEY `idx_langcode` (`lang_code`),
   KEY `idx_access` (`access`),
   KEY `idx_ordering` (`ordering`)
```
PhilETaylor:
```suggestion
  KEY `idx_ordering` (`ordering`),
```

alikon:
i don't think this is needed cause with the low cardinality of these table items, the db-engine will always perform a table scan, so PK index apart, no need for other indexes

aDaneInSpain:
> i don't think this is needed cause with the low cardinality of these table items, the db-engine will always perform a table scan, so PK index apart, no need for other indexes

This is the kind of input I was after. I am not a DB optimization expert. But we have found minute performance improvements by adding additional keys to several tables. Milliseconds, but still. From EXPLAIN SELECT I can see that it is using "file_sort" which is my understanding is always ideal to avoid?

aDaneInSpain:
From what I can read around the web the general guideline is that you should provide indexes for your select queries to improve performance. The only downside is that inserts, deletes and some updates can take slightly longer (as it needs to also update the index). But that should only be an issue if you have a table with frequent i/o operations which #__languages and #__extensions certainly aren't.

PhilETaylor:
While you are correct... in theory. 

the `#__languages` table will only ever be a handful of languages... so the fraction of a millisecond is hardly worth arguing about. 

Now, if the same applies on the content tables, then it would be worth it, as those tables can get very very large.

aDaneInSpain:
I have not reviewed everything yet, but I know it applies to the #__extensions table, for instance. And if it does, minutely help on performance (and shows less "file sort" warnings) in the Joomla debug then I think it is worth fixing throughout. I just want to make sure there is no reason NOT to do it.

aDaneInSpain:
Would still love to hear if anyone has any reasons for **not** to do this. If there are no objections, I will modify the pull request to handle all instances in all core tables and implement installation and migration for both MySQL and PostgreSQL.

aDaneInSpain:
I am going to close this as I don't believe it has much impact on performance and the debugger in J4 does not show the "explain selects", so it does not jump out at you (which is a shame really) so it will not "scare" users as much.

HLeithner:
> the debugger in J4 does not show the "explain selects", so it does not jump out at you (which is a shame really)

@aDaneInSpain Actually you have to activate this functionality in the plugin, then you have also an explain to each query.

aDaneInSpain:
> @aDaneInSpain Actually you have to activate this functionality in the plugin, then you have also an explain to each query.

Oh! Was not aware. Thanks.

PR #36440, opened 2021-12-27
[6.2] Use an injected logger in the Task driver

ditsuke:
### Summary of Changes
A logger is injected into Task at creation, and into routines through ExecutedTaskEvent at runtime.

### Testing Instructions
Tasks should be logged as expected to the scheduler log file "administrator/logs/joomla_scheduler.php".
Logs for individual tasks, if enabled, should be saved to the file as specified in the configuration.

No DI, ugly setup code in TaskPluginTrait etc.

DI, less static calls (still depends on the global Log instance to setup sub-loggers! :/), more succinct.

\-


administrator/components/com_scheduler/src/Scheduler/Scheduler.php
```diff
@@ -255,7 +260,7 @@ public function getTask(array $options = []): ?Task
 			return null;
 		}
 
-		return new Task($task);
+		return new Task($task, Factory::getContainer()->get(LoggerInterface::class));
```
laoneo: Would be nice if you can get here a logger which is injected all the way down from the component class.

ditsuke: I was not able to figure how to do this, since our instantiation chain starts from a plugin. Any leads? 😕

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
```diff
@@ -127,14 +129,17 @@ public function runTask(array $options): ?Task
...
 
-		$options['text_entry_format'] = '{DATE}	{TIME}	{PRIORITY}	{MESSAGE}';
-		$options['text_file']         = 'joomla_scheduler.php';
-		Log::addLogger($options, Log::ALL, $task->logCategory);
+		$loggerOptions = [
+			'text_entry_format' => '{DATE}	{TIME}	{PRIORITY}	{MESSAGE}',
+			'text_file'         => 'joomla_scheduler.php'
+		];
+		/** ! No non-static methods to add a sub-logger to {@see DelegatingPsrLogger} (Task::logger) */
+		Log::addLogger($loggerOptions, Log::ALL, $task->logCategory);
```
nibra: Better inject a pre-configured logger using Scheduler's constructor. Within the component, you shouild just send messages to the logger and not change its configuration. The setup is up to the front controller, not the component itself.

For example, if one wants to set up an automated evaluation of the log, it is necessary to be able to freely determine the log format and the storage of the log data. Your approach, however, prescribes these properties so that the dependency injection is practically worthless.

administrator/components/com_scheduler/src/Traits/TaskPluginTrait.php
```diff
@@ -233,30 +243,16 @@ protected function getRoutineId(Form $form, $data): string
...
-			'warning' => Log::WARNING,
-		];
-
-		if (!$langLoaded)
+		if (empty($this->logger))
 		{
-			$app = $this->app ?? Factory::getApplication();
-			$app->getLanguage()->load('com_scheduler', JPATH_ADMINISTRATOR);
-			$langLoaded = true;
+			throw new \LogicException("Logger has not been setup!");
```
nibra: Logging should always be optional. Issue a notice or warning, if necassary, but don't stop execution.

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
ditsuke: I see the merit in decoupling the configuration! It's totally doable for everything but task-specific logging since the relevant information is only exposed once we get the task record from the database. What would be the best way to do that setup?

https://github.com/ditsuke/joomla-cms/blob/scheduler-logging-di/administrator/components/com_scheduler/src/Task/Task.php#L151-L160

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
laoneo: Do you boot the scheduler component in the plugin?

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
nibra: Why can't you inject the same logger from the Scheduler into the Task?

ditsuke: I can, but it needs some additional configuration specific to the task (logging to the task-specific log file).

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
ditsuke: Right now, its only done in the plugin's `injectLazyJS()` method. The component is otherwise booted within the Scheduler::getTask() method.

administrator/components/com_scheduler/src/Traits/TaskPluginTrait.php
ditsuke: Good idea! Didn't think of that

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
ditsuke: Should we be injecting the model/component instance?

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
nibra: That's a misconception. *If*, *how* and *where* things are logged is totally up to the system admin. If you want (feel the need) to provide a default logging procedure, you can do so by providing a bespoke logger that is instantiated as a default, if no logger is provided be the environment.

HLeithner:

administrator/components/com_scheduler/src/Scheduler/Scheduler.php
laoneo: Yes there you can start with it. In the provider.php file you can inject the logger into the component and then pass it into the scheduler.

laoneo:
@ditsuke can you implement the the suggestions and fix the conflicts?

HLeithner:

HLeithner:

HLeithner:

HLeithner:

Hackwar:
@ditsuke Are you still working on this?

HLeithner:
This pull request has been automatically rebased to 6.0-dev.

PR #36442, opened 2021-12-27
[4.0] Remove reference to old JPrefixed Class + Cleanup (Misc)

PhilETaylor:
Code Review Removing Technical Debt.

Remove reference to old JPrefixed Class + other small cleanups of doc blocks.

Nothing here is controversial and should be a quick test and merge.

These changes are manually made and reviewed, over several hours, ~and are the **very last** J Prefixed classes~ never mind, found more to do haha...


plugins/system/actionlogs/actionlogs.php
```diff
@@ -100,11 +98,11 @@ public function onContentPrepareForm(Form $form, $data)
 		}
 
 		/**
-		 * We only allow users who has Super User permission change this setting for himself or for other users
-		 * who has same Super User permission
+		 * We only allow users who have Super User permission change this setting for himself or for other users
+		 * who have same Super User permission
```
brianteeman:
```suggestion
		 * We only allow users who have Super User permission to change this setting for themselves or for other
		 * users who have the same Super User permission
```

layouts/joomla/form/field/subform/default.php
```diff
@@ -7,14 +7,16 @@
  * @license     GNU General Public License version 2 or later; see LICENSE.txt
  */
 
+use Joomla\CMS\Form\Form;
+
 defined('_JEXEC') or die;
```
Quy:
```suggestion
defined('_JEXEC') or die;

use Joomla\CMS\Form\Form;
```

libraries/src/Plugin/PluginHelper.php
```diff
@@ -261,9 +262,11 @@ protected static function load()
 			return static::$plugins;
 		}
 
-		$levels = Factory::getUser()->getAuthorisedViewLevels();
+		$levels = Factory::getApplication()
+			->getIdentity()
```
wilsonge: Given the open issue on this you created - not sure this is something to introduce to a on review PR?

libraries/src/Updater/UpdateAdapter.php
```diff
@@ -211,7 +211,7 @@ protected function getUpdateSiteName($updateSiteId)
 	 *
 	 * @param   array  $options  The update options, see findUpdate() in children classes
 	 *
-	 * @return  boolean|\JHttpResponse  False if we can't connect to the site, JHttpResponse otherwise
+	 * @return  \Joomla\CMS\Http\Response|bool  False if we can't connect to the site, JHttpResponse otherwise
```
wilsonge:
```suggestion
	 * @return  \Joomla\CMS\Http\Response|bool  False if we can't connect to the site, HTTP Response object otherwise
```

plugins/actionlog/joomla/joomla.php
```diff
@@ -297,7 +299,7 @@ public function onContentChangeState($context, $pks, $value)
 	 * On Saving application configuration logging method
 	 * Method is called when the application config is being saved
 	 *
-	 * @param   JRegistry  $config  JRegistry object with the new config
+	 * @param   \Joomla\Registry\Registry  $config  JRegistry object with the new config
```
wilsonge:
```suggestion
	 * @param   \Joomla\Registry\Registry  $config  Registry object with the new config
```

plugins/system/actionlogs/actionlogs.php
```diff
@@ -100,11 +98,11 @@ public function onContentPrepareForm(Form $form, $data)
...
 
 		/**
-		 * We only allow users who has Super User permission change this setting for himself or for other users
-		 * who has same Super User permission
+		 * We only allow users who have Super User permission to change this setting for themselves or for other
+		 * users who have the same Super User permission
 		 */
 
-		$user = Factory::getUser();
+		$user = Factory::getApplication()->getIdentity();
```
wilsonge: Same here I guess if it returns null then `authorise` won't exist as a method

plugins/content/emailcloak/emailcloak.php
```diff
@@ -33,10 +32,10 @@ class PlgContentEmailcloak extends CMSPlugin
 	 *
 	 * @param   string   $context  The context of the content being passed to the plugin.
 	 * @param   mixed    &$row     An object with a "text" property or the string to be cloaked.
-	 * @param   mixed    &$params  Additional parameters. See {@see PlgContentEmailcloak()}.
+	 * @param   mixed    &$params  Additional parameters.
 	 * @param   integer  $page     Optional page number. Unused. Defaults to zero.
 	 *
-	 * @return  void
+	 * @return  void|bool
```
wilsonge: Can we just remove the bool return when we're the api rather than documenting this as a bool method please

PhilETaylor:
Applied suggestions

wilsonge:
Thanks!

PR #36447, opened 2021-12-27
[4.0] Remove all calls to deprecated Factory::getUser()

PhilETaylor:
Removes all calls to the deprecated `Factory::getUser()` method with the `Factory::getApplication()->getIdentity()` chained method. 

`Factory::getUser()` is slated for removal in Joomla 5

Manually reviewed each of the changes... 

ONLY changed where original was a full call to `Factory::getUser()` and replaced like for like. 

Code review.


PhilETaylor:
Converting back to draft because... when a user is not logged in 

`Factory::getApplication()->getIdentity()` = `null`

whereas 

`Factory::getUser()` = an empty `Joomla\CMS\User\User` object 

!!!

PhilETaylor:
Ok @wilsonge you up for this challenge? 

The back end is fine - its only the frontend having issues, PluginHelper and SiteApplication seem to be the two files that need touching but I cannot seem to find the "right way" to do this. 

It seems that Joomla is highly dependant on getUser returning a empty User class when not logged in... so that `->getAuthorisedLevels()` and other methods can be run on a guest user. 

The deprecation message therefore is wrong as the two methods are not the same. `getUser` loads from `self::getApplication()->getSession()->get('user')` and `getIdentify` just loads from the class property - and is null if the user is logged out. 

Or is the correct "fix" to refactor most of the frontend, and all the 3PD extensions so that they check `getIdentity` and see if its null or not before using it? ... but then you could have ACL on your logged out users - but as your logged out user is NULL and not a blank user object that levels can be checked on... you are screwed... haha. 

Until this is resolved, the use of `Factory::getApplication()->getIdentity()` on the frontend, where users could be logged out, should be minimised...

PhilETaylor:
Closing here - but leaving the branch because you can checkout this PR to see the problem that needs fixing...

PhilETaylor:
Conflicts resolved.

PhilETaylor:
Zero interest and now conflicts again.

PR #36473, opened 2021-12-29
[4][com_finder] - php 8.1 serializable-deprecated

alikon:
https://php.watch/versions/8.1/serializable-deprecated

### Summary of Changes
added magic method

### Testing Instructions
php 8.1
on frontend do a search
check the log

` PHP Deprecated:  The Serializable interface is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)`

no more deprecation

### DIFF
--- a/administrator/components/com_finder/src/Indexer/Result.php
+++ b/administrator/components/com_finder/src/Indexer/Result.php
@@ -555,4 +555,19 @@ class Result implements \Serializable
 			}
 		}
 	}
+
+	/**
+	 * Magic method used for serializing.
+	 */
+	public function __serialize()
+	{
+	}
+
+	/**
+	 * Magic method used for unserializing.
+	 */
+	public function __unserialize($serialized)
+	{
+	}
+
 }


PhilETaylor:
Surely this is missing the removal of the implements, and also the actual implementation of the these functions, not just empty functions? 

`class Result implements \Serializable`

administrator/components/com_finder/src/Indexer/Result.php
```diff
@@ -555,4 +555,19 @@ public function unserialize($serialized)
...
+	{
+	}
+
+	/**
+	 * Magic method used for unserializing.
+	 */
+	public function __unserialize($serialized)
+	{
+	}
+
```
PhilETaylor, line 572: type hints as per https://www.php.net/manual/en/language.oop5.magic.php#object.serialize

```suggestion

	/**
	 * Magic method used for serializing.
	 */
	public function __serialize(): array
	{
	}

	/**
	 * Magic method used for unserializing.
	 */
	public function __unserialize(array $serialized): void
	{
	}

```

richard67:
> Surely this is missing the actual implementation of the these functions?

@PhilETaylor It should be enough if they call the `serialize()` and `unserialize($serialized)` methods defined a bit above, right? Or can they even be aliased somehow?

administrator/components/com_finder/src/Indexer/Result.php
richard67: @PhilETaylor `public function __serialize(): array` ... shouldn't that be `public function __serialize(): string`?

PhilETaylor: Click the link - see the documentation 
I agree it seems odd but hey

PhilETaylor: I’m not at a desk for several hours to run tests

richard67: >  serialize() checks if the class has a function with the magic name __serialize(). If so, that function is executed prior to any serialization. It must construct and return an associative array of key/value pairs that represent the serialized form of the object. If no array is returned a TypeError will be thrown.

Gulp. I am confused.

PhilETaylor: when you call `serialize($object)` you are calling a [`serialize` FUNCTION](https://www.php.net/manual/en/function.serialize.php) not the `$object->serialize` METHOD

when you call `serialize($object)` it "checks if the OBJECT has a METHOD with the magic name __serialize()"

__serialize should then construct and return an associative array of key/value pairs that represent the serialized form of the object. It returns that array of data - ONLY THAT DATA IS SERIALiZED by PHP

PHP will then (as part of the  [`serialize` FUNCTION](https://www.php.net/manual/en/function.serialize.php) ) convert that ARRAY into a STRING.

PhilETaylor: Unit tests that prove this PR is wrong are incoming....

PhilETaylor:
Replacement PR is here https://github.com/joomla/joomla-cms/pull/36482 with unit tests that can be used to prove before PR (in PHP 7) and after PR (in PHP 8.1+)

alikon:
closing in favour of #36482

PR #36482, opened 2021-12-29
[4][com_finder] - php 8.1 serializable-deprecated with unit test

PhilETaylor:
replacement of https://github.com/joomla/joomla-cms/pull/36473 with Unit tests that run before and after with success (although correctly, before this PR the unit test has the deprecation warning.

@richard67 @alikon 

### Summary of Changes

added magic method

### Testing Instructions

php 8.1
on frontend do a search
check the log

PHP Deprecated: The Serializable interface is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)

no more deprecation

### Note

This is our FIRST EVER Unit test for a method in Joomla 4 Extensions it seems... woot woot (also see: https://github.com/joomla/joomla-cms/pull/36483)


PhilETaylor:
Strange but probably @HLeithner was doing, the Drone PHP 8.1 unit tests are being run by PHP 8.0 at the moment... 

However, this new Unit test passes :) 
Screenshot 2021-12-29 at 16 20 30

PhilETaylor:
Drone failure unrelated to PR content.

administrator/components/com_finder/src/Indexer/Result.php
```diff
@@ -466,11 +466,35 @@ public function setLanguage()
...
+	 *
+	 * @since   4.0.0
+	 */
+	public function unserialize($serialized): void
+	{
+		$this->__unserialize(unserialize($serialized));
+	}
+
+	/**
+	 * Magic method used for serializing.
```
alikon, line 487: sorry to be pedantic...
but shouldn't need to be added `@return`

administrator/components/com_finder/src/Indexer/Result.php
```diff
@@ -508,42 +531,42 @@ public function serialize()
...
+		];
 	}
 
 	/**
-	 * Helper function to unserialise the data for this object
-	 *
-	 * @param   string  $serialized  Serialised data to unserialise
-	 *
-	 * @return  void
+	 * Magic method used for unserializing.
```
alikon, line 539: but shouldn't need to be added `@return`  & `@param`

alikon (changes requested):

alikon:
I have tested this item :white_check_mark: successfully on 50e03bd378cfc6c9ad034956238658ce83d3a961<br /><br />on php 7.4 and php 8.1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36482">issues.joomla.org/tracker/joomla-cms/36482</a>.</sub>

administrator/components/com_finder/src/Indexer/Result.php
PhilETaylor: Joomla coding standards- all methods to be documented regardless

HLeithner:
Can you make an update to the branch to retrigger drone (emojis in the description... I think I have to check drone why this happens maybe mb4 on mysql)

PhilETaylor:
wow - welcome to 2022 where emoji's still break things.

PhilETaylor:
Drone seems to have flagged this PR as passed now, even though some of the suites fail for unrelated reasons.

administrator/components/com_finder/src/Indexer/Result.php
PhilETaylor: And to be honest, in a PHP 8+ world, I agree with you! All my static cleanup tools now will remove docblock type hints if the type hint is in the PHP code as a typehinted param, or type hinted return type. But, Joomla is still a decade behind everyone else when it comes to this stuff.

HLeithner:
> Drone seems to have flagged this PR as passed now, even though some of the suites fail for unrelated reasons.

That's ok, all php8+ system tests are allowed to fail and 8.1 unit tests too, also postgres system tests

administrator/components/com_finder/src/Indexer/Result.php
```diff
@@ -508,42 +531,42 @@ public function serialize()
...
+			'sale_price',
+			'start_date',
+			'state',
+			'taxonomy',
+			'title',
+			'type_id',
+			'url',
+			];
+
+		foreach ($properties as $k => $v)
```
wilsonge, line 568: $properties isn't a key value array. Something's not right here

PhilETaylor: Not my code - I just moved it

PhilETaylor: hmm  maybe I did... checking

PhilETaylor: No this code is correct and works (should not have doubted myself, the unit test proves it) 

The $k is the index 0,1,2,3,4,5,6... etc 
The $v is the property name as a string
The $serialized array is INDEXed and not a k=>v array

Therefore it "magically" correctly matches the correct key with the correct property  :)

wilsonge: Ahh yeah i did wonder how the test worked. That makes sense to me

PhilETaylor: I have added a note with 10f7fe1 - basically as long as you serialise in the same order as you deserialise then it works...

wilsonge (approved):

tests/Unit/CMS/administrator/components/com_finder/src/Indexer/ResultTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * @package     Joomla.UnitTest
+ * @subpackage  com_finder
+ *
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+namespace Joomla\Component\Finder\Administrator\Indexer;
```
wilsonge: I don't see how this is every being namespace autoloaded when the code is in this directory? Feels very wrong somehow. Are we sure this test is actually being run?

PhilETaylor: Does phpunit use namespace loading? I think it just iterates all folders looking for *Test.php 

feel free to update the namespace if you think its wrong. Im knee deep in another project right now.

PhilETaylor: So I have looked into this.

PHPUnit is not loading the class using autoloading, its, as I guessed, just scanning for the *Test.php named files. 

The namespace can be literally anything and the test will still load and run as evidenced by running in debug mode you can see the test in yellow here. 

Screenshot 2022-01-09 at 15 30 42

I have changed the namespace anyway to be 

`namespace Joomla\Tests\Unit\Administrator\Components\Finder\Indexer;`

and I have imported the Result class (as its no longer in the same namespace) and the tests still pass

Screenshot 2022-01-09 at 15 30 13

I'll merge these changes and push to the PR now.

PR #36483, opened 2021-12-29
[4] Rename phpunit test folders to align with the apps structure

PhilETaylor:
https://github.com/joomla/joomla-cms/pull/36482 introduces our first Unit test for Joomla 4 extensions methods

Previously the only tests are for the libraries/src/* classes in the /test/Unit/Libraries/CMS folder which makes no sense 

The Unit testing folder structure should reflect the path structure of the code its testing (Best Practice) 

This PR renames the folders - the tests still run perfectly fine.

This will eventually conflict with https://github.com/joomla/joomla-cms/pull/36482 potentially depending on which is merged first, but I'll fix that myself when I see that happen.


PhilETaylor:
Drone failure unrelated to PR content

wilsonge:
Thanks!

PR #36484, opened 2021-12-29
[4] Requesting to be added to code owners

PhilETaylor:
Requesting to be added to code owners for tests/Unit

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -34,12 +34,12 @@ libraries/src/Updater/*  @rdeutz @wilsonge @zero-24
 tests/*  @hackwar
 tests/Codeception/*  @rdeutz
 tests/javascript/*  @rdeutz
-tests/Unit/*  @rdeutz
+tests/Unit/*  @rdeutz @philetaylor
 .appveyor.yml  @rdeutz @hackwar
 .drone.yml  @rdeutz @hackwar
 .hound.yml  @wilsonge @hackwar
 codeception.yml  @rdeutz @hackwar
-phpunit.xml.dist  @rdeutz @hackwar
+phpunit.xml.dist  @rdeutz @hackwar @philetaylor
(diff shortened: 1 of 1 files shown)


PhilETaylor:
Drone failure unrelated to PR content

wilsonge:
Enjoy!

PhilETaylor:
Gulp! haha Thanks.

HLeithner:
What's the reason for your request?

PhilETaylor:
"Code owners are automatically requested for review when someone opens a pull request that modifies code that they own"

Own is a harsh word. Contribute to and have experience of is a better phrase. 

Seriously though - something has to be done about the low quality of code in Joomla - if Joomla is to survive as a Modern PHP Application (rather than a shiny frontend) - and Unit Testing is something I have considerable experience in for huge projects. I guess I just had a moment of pity. 

There are only a few people providing any unit tests at all recently - and certainly there is a lot to do at the Joomla 4+ level (Im not touching 3 unit tests, I know you are refactoring those)

What is your reason for questioning my involvement? 

However:

"The people you choose as code owners must have write permissions for the repository"

Oh well that means this should be reverted ... oh well nice while it lasted..

brianteeman:
it doesnt work if you dont have  the correct permissions

PhilETaylor:
Like I said. 

> "The people you choose as code owners must have write permissions for the repository"
> Oh well that means this should be reverted ... oh well nice while it lasted..

And I don't want write permissions to this repo. Ever.

HLeithner:
@PhilETaylor that's the reason I asked, with out the permissions it doesn't work. But george should know that...

Anyway yes you are right we need more tests, but we shouldn't do this uncoordinated. Doesn't mean your help isn't appreciated, actually it is.

I revert the pr since as you already found out too it doesn't work as expected.

PR #36486, opened 2021-12-29
[4] Autoloader for psr4 for Unit Testing

PhilETaylor:
### Summary of Changes

In developing the first ever Joomla 4 Extension test (https://github.com/joomla/joomla-cms/pull/36482) it became apparent that when running Unit Tests, Auto loading of Joomla CMS classes was not working. 

For the purposes of https://github.com/joomla/joomla-cms/pull/36482 I simply included the classes, but this is unsustainable for the future and autoloader should work. 

This PR cleans up the bootstrap process for Unit Tests, and implements a cut down namespace map for Unit Testing (as we cannot use the /administrator/cache/autoload_psr4.php as it doesnt exist at this point, and to create it requires the full working stack and DB, and unit testing is about testing the smallest part of code at a time... to much overhead to use that map, so we have a much smaller one for Unit testing anyway (one class to load at present!) 

### Testing Instructions

If the unit tests all still work - then this PR works. 

Once merged the Unit test in https://github.com/joomla/joomla-cms/pull/36482 can be tweaked to remove the `setUp` function. 

Unit tests run fine - but autoloading of Joomla CMS namespaces classes fails

Unit tests run fine - and autoloading of Joomla CMS namespaces classes works

None


PhilETaylor:
@wilsonge Im switching this PR to draft until I check it again after the autoloading issue you reported - thanks.

laoneo:
Just for reference, this one works since #https://github.com/joomla/joomla-cms/pull/37592.

laoneo:
And https://github.com/joomla/joomla-cms/pull/37595.

PR #36488, opened 2021-12-29
Revert "[4] Requesting to be added to code owners"

PhilETaylor:
Reverts joomla/joomla-cms#36484


HLeithner:
as discussed in #36484 
thanks

Issue #36478, opened 2021-12-29
[4.0] Missing mail template not handled correctly


BrainforgeUK:
### Steps to reproduce the issue

Try to send mail using a mail template which does not exist.

```
$mailer = new MailTemplate('com_bftest.invalidtemplateid', $language);
$mailer->addRecipient($templateData['emailto'], $templateData['emailtoname']);
$mailer->setReplyTo($templateData['replytoemail'], $templateData['replytoname']);
$mailer->addTemplateData($templateData);
$sent = $mailer->send();
```

### Expected result

Sensible error message saying template does not exist

### Actual result

Load of PHP notices and a 'Message body empty' warning.

### Additional comments

Be better to add a getInstance() static method to MailTemplate which would allow this
```
$mailer = MailTemplate::getInstance('com_bftest.invalidtemplateid', $language);
if (empty($mailer))
{
     ... take some action ...
}
$mailer->addRecipient($templateData['emailto'], $templateData['emailtoname']);
$mailer->setReplyTo($templateData['replytoemail'], $templateData['replytoname']);
$mailer->addTemplateData($templateData);
$sent = $mailer->send();
```
Can do this as a workaround - downside MailTemplate::getTemplate() gets called twice.
```
$test = MailTemplate::getTemplate($templateId, $language);
if (empty($mail))
{
     ... take some action ...
}
$mailer = new MailTemplate('com_bftest.invalidtemplateid', $language);
$mailer->addRecipient($templateData['emailto'], $templateData['emailtoname']);
$mailer->setReplyTo($templateData['replytoemail'], $templateData['replytoname']);
$mailer->addTemplateData($templateData);
$sent = $mailer->send();
```

PhilETaylor:
The problem is that, mails can come from anywhere, Admin, Frontend, CLI etc... So the correct thing would be an Exception, but that is too hard on a "user trying to send a contact form, formatted by a non existing mail template) and equally, in /administrator/ a "error" message can be disabled in the admin console, and for CLI an exception...

however to capture all that, the try/catch block has to be in your code. 

therefore I would just say that `$mailer->send()` should return false, with no other error messages

or just throw an Exception and expect 3PD to deal with catching it?

PhilETaylor:
Simple approach https://github.com/joomla/joomla-cms/pull/36489

richard67:
Closing as having a pull request. Please test #36489 . Thanks in adcance.

PR #36489, opened 2021-12-30
[4.0] If the Mail Template was not found in the db, return false

PhilETaylor:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/36478

### Summary of Changes

If the Mail Template was not found in the db, we cannot send an email, therefore return false before trying to do lots of things based on the template that doesnt exist and avoid error messages

### Testing Instructions

Code review

```php
$mailer = new MailTemplate('com_bftest.invalidtemplateid', 'en-GB');
// ...
$sent = $mailer->send();
```

Load of PHP notices and a 'Message body empty' warning.

```php
// $sent === false
$sent = $mailer->send();
```

None

### DIFF
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -376,7 +376,7 @@ class MailTemplate
 	 * @param   string  $key       Template identifier
 	 * @param   string  $language  Language code of the template
 	 *
-	 * @return  object  An object with the data of the mail
+	 * @return  object|null  An object with the data of the mail, or null if the template not found in the db.
 	 *
 	 * @since   4.0.0
 	 */
--- a/libraries/src/Mail/MailTemplate.php
+++ b/libraries/src/Mail/MailTemplate.php
@@ -28,7 +28,7 @@ use PHPMailer\PHPMailer\Exception as phpmailerException;
 class MailTemplate
 {
 	/**
-	 * Mailer object to send the actual mail..
+	 * Mailer object to send the actual mail.
 	 *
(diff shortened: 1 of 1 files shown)


PhilETaylor:
I would write a Unit test for this but... `ComponentHelper::getParams` /facepalm/ and static calls cannot be mocked :(

wilsonge:
> I would write a Unit test for this but... ComponentHelper::getParams /facepalm/ and static calls cannot be mocked :(

I mean if you really wanna go ham - which I strongly encourage of course :) You should be able to mock the params in the db I think. I'll see if I can dig out an example tomorrow.

libraries/src/Mail/MailTemplate.php
```diff
@@ -190,6 +190,12 @@ public function send()
 
 		$mail = self::getTemplate($this->template_id, $this->language);
 
+		// If the Mail Template was not found in the db, we cannot send an email.
+		if ($mail === null)
```
wilsonge, line 194: Looks good. Can we take the opportunity to document getTemplate as returning `null` as well as object in it's doc block so this makes sense for code reviews in the future

PhilETaylor: Done

PhilETaylor: oh hang on - wrong branch ... grrr

PhilETaylor: Done.

wilsonge:
Thanks!

PR #28549, opened 2020-04-03
[4.0] Joomlaupdate: fix j version in RTL, alpha lang file

infograf768:
### Summary of Changes
As title says.
in RTL, the Joomla version should be displayed as LTR
Alpha ordered the ini file.

### Example Before patch
`administrator/index.php?option=com_joomlaupdate`
updatebefore

### Example After patch

updateafter

PR #36490, opened 2021-12-30
[4.0] RTL Version number in log file

brianteeman:
### Summary of Changes
PR #28549 added a _force ltr_ character before the version number is displayed to prevent issues when using an rtl language.

It should not have been added to this instance as it is only used for the update log file which is always in english and therefore LTR and the special character is not recognised in plain text

### Testing Instructions
code review as its an obvious fix

the logs\joomla_update.php file will contain a line similar to
`2021-12-30T08:12:30+00:00	INFO ::1	update	Update started by user Brian Teeman (608). Old version is &#x200E;4.0.6-dev.`

the logs\joomla_update.php file will contain a line similar to
`2021-12-30T08:12:30+00:00	INFO ::1	update	Update started by user Brian Teeman (608). Old version is 4.0.6-dev.`

none

### DIFF
--- a/administrator/components/com_joomlaupdate/src/Controller/UpdateController.php
+++ b/administrator/components/com_joomlaupdate/src/Controller/UpdateController.php
@@ -44,7 +44,7 @@ class UpdateController extends BaseController
 
 		try
 		{
-			Log::add(Text::sprintf('COM_JOOMLAUPDATE_UPDATE_LOG_START', $user->id, $user->name, '&#x200E;' . \JVERSION), Log::INFO, 'Update');
+			Log::add(Text::sprintf('COM_JOOMLAUPDATE_UPDATE_LOG_START', $user->id, $user->name, \JVERSION), Log::INFO, 'Update');
 		}
 		catch (\RuntimeException $exception)
 		{


richard67:
I agree that no HTML markup (and the unicode LTR mark can be considered as such) shall appear in a plain text log file.

Unfortunately the log is a PHP file, so some people might have some extension which shows it in the backend and so in a browser. Maybe this is how it came that the LTR mark was added.

I think it was wrong from the beginning on to use a PHP file as log file. It should be plain text with extension "log" or "txt".

If using PHP, the content should me marked somehow as text only, or it should produce HTML markup (with which we then could use the LTR marks or better ways again to show the version with LTR direction in the right way).

We should consider to change that for 5.0.

richard67:
I have tested this item :white_check_mark: successfully on 644708f19089d9637e3003c11a9e0dc8d76e8595<br /><br />Code review.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36490">issues.joomla.org/tracker/joomla-cms/36490</a>.</sub>

richard67:
@brianteeman Regarding the RTL display of version numbers at other places than the log (for which I had started to work on the 3.10-dev branch as you maybe remember): If there are some issues remaining after your CSS cleanups, I can make a PR to fix them. I have already prepared most of it. In opposite to my previous attempt I will not use a helper but it will work the same way, wrap it into a span with dir=ltr if necessary. But I think it has been wrong for so long time and we had not many issues with it, so I think there is no need to hurry. Let me know if you disagree.

PhilETaylor:
> It should be plain text with extension "log" or "txt"

it’s a PHP file for security - when accessed through a direct url it doesn’t render the content but dies. (Or should do)

a txt or log file in the webspace would be publicly available to view which could be bad.

richard67:
> > It should be plain text with extension "log" or "txt"
> 
> it’s a PHP file for security - when accessed through a direct url it doesn’t render the content but dies. (Or should do)
> 
> a txt or log file in the webspace would be publicly available to view which could be bad.

@PhilETaylor I know. But that could be solved differently or the PHP log could use HTML markup or something like that.

brianteeman:
changing the file type is beyond the scope of this PR

richard67:
> changing the file type is beyond the scope of this PR

@brianteeman Yes, I know, that's why I have tested the PR with success.

alikon:
I have tested this item :white_check_mark: successfully on 644708f19089d9637e3003c11a9e0dc8d76e8595<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36490">issues.joomla.org/tracker/joomla-cms/36490</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36490">issues.joomla.org/tracker/joomla-cms/36490</a>.</sub>

wilsonge:
Thanks!

PR #36498, opened 2021-12-30
Revert "[4] Rename phpunit test folders to align with the apps structure"

wilsonge:
Reverts joomla/joomla-cms#36483

@PhilETaylor this broke the composer autoloading (apparently drone didn't fail because it still uses composer v1 🤦 ). But if you look at the original PR "composer" section in Drone you can see the warnings for composer v2


PhilETaylor:
Hmmm works on my machine :-) and I use snapshot composer 

I’ll look again as this change is the right thing to do

wilsonge:
For sure - otherwise I wouldn't have merged it :) Just definitely screws with the composer autoload here I think https://github.com/joomla/joomla-cms/blob/4.0-dev/composer.json#L30-L34 - if we go down this route I think we have to re-amend all the namespaces (or change the autoloading method)

PhilETaylor:
I guess what this also shows is that we are using Composer 1 when Composer 2 is the new norm ;-) 

Leave it with me, I'll take another stab tomorrow.

PR #36499, opened 2021-12-31
[5.0] Rework session/application DIC to avoid circular dependencies

wilsonge:
### Summary of Changes
This is a semi-major rework of the application and session packages to avoid the circular dependencies between the two. As an unideal compromise the request object (i.e. JInput) is created as an explicit resource into the container as it's the main common dependency the two rely on. It's not very architectural best practice - but it's how symfony worked until very recently and it's also how laravel still works - so it's not super terrible compared to global statics :)

### Testing Instructions
Test all 5 main applications in Joomla - Console, Site, Admin, API and Installation. Ensure that they are available and work properly. Ensure that session works, by logging in and anything else that uses the session heavily (e.g. mod_logged in the backend).

Obviously with this being a change to two of the lowest level pieces of the stack it's hard to predict how it would manifest beyond the major breaks (exceptions etc). So I'm somewhat relying on people testing carefully!

Someone who understands the finesee's of cookies better than me - I'd also appreciate a review of the cookiePath parts of JoomlaStorage - as before sometimes the cookiePath was an empty string and sometimes '/' and i'm not 100% sure if there's a difference or not. There might be a difference in the administrator directory - depending on exactly the path is determined? (PHP Reference for those brave enough) https://www.php.net/manual/en/function.setcookie.php 

None


richard67:
What is a DIC? I hope it’s not a typo and the K is missing 😄

brianteeman:
Dependency Injection Container

richard67:
Ah, thanks.

richard67:
@wilsonge API tests are failing, and it could be related to your changes because it complains about something with the input filter: https://ci.joomla.org/joomla/joomla-cms/49850/1/28

wilsonge:
Yup that would be me doing stupid things.

libraries/src/Service/Provider/Application.php
```diff
@@ -168,7 +168,7 @@ function (Container $container)
 			->share(
 				'JApplicationApi',
 				function (Container $container) {
-					$app = new ApiApplication($container->get(CMSInput::class), $container->get('config'), null, $container);
+					$app = new ApiApplication(null, $container->get('config'), null, $container);
```
HLeithner: shouldn't this work instead?
`$container->get(CMSInput::class)->json`

Other question why does the API requires the JSON Input?

wilsonge: Yeah that would work too. I just reverted back to what we currently have - less lines of code changed probably a good thing with this PR 😅

In terms of why - because we are receiving data in JSON format (default data structure being json api) we default to using the JInputJson class.

wilsonge: The session for the api is just the runtime storage (same as the cli) - it's not persistent so doesn't take an input object either. So there's no common setting input object required anyhow

wilsonge:
@HLeithner please can you review this and merge so it's in early in the 4.3 lifecycle. the sooner it's in the better given the b/c impact should be 0 as this all happens before even system plugins are booted

HLeithner:
Is there any possible b/c break if you have your own applicaton?

HLeithner:
@wilsonge is this needed for 4.4? I think we are better safe with 5.0, since it's released at the same time it's only a matter of will it break with later php versions.

if ok please rebase on 5.0 then I add it to my personal tasks list

wilsonge:
I really don't mind where you merge it. But I'm not rebasing this for a 4th minor version in a row without some assurances that it's going to get at least tested. Because it's just a waste of my time.

wilsonge:
See linked PR

HLeithner:
Thanks, I merge this for the next Alpha to get more feedback.

nikosdion:
FWIW, even though George calls this a non ideal compromise, it's actually how I have done things in AWF for the same reason. It also makes testing far easier since we can mock the input without having to mock the entire application. So, yay, thank you for doing this!

PR #36503, opened 2021-12-31
[3] Remove reference & in usort callback function

HLeithner:
usort throws an error when callback functions requires a reference.

`JArrayHelper::_sortObjects(): Argument #1 ($a) must be passed by reference, value given`

### Summary of Changes
Removed & from _sortObjects parameter signature

### Testing Instructions
Use JArrayHelper::sortObjects

* error message on php 8+

No error message

### DIFF
--- a/libraries/joomla/utilities/arrayhelper.php
+++ b/libraries/joomla/utilities/arrayhelper.php
@@ -358,15 +358,15 @@ abstract class JArrayHelper
 	/**
 	 * Callback function for sorting an array of objects on a key
 	 *
-	 * @param   array  &$a  An array of objects
-	 * @param   array  &$b  An array of objects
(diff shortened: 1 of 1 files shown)


richard67:
@HLeithner Does it really need a real test here? I think it should be sufficient by review with respect to https://www.php.net/manual/en/function.usort.php showing by value parameters for the callback:

```
 callback
    The comparison function must return an integer less than, equal to, or greater than zero if the first argument is considered to be respectively less than, equal to, or greater than the second.
    callback(mixed $a, mixed $b): int
```

HLeithner:
Don't think so because unit tests are successful on all (old) php versions and I know the tests are working ;-)

Fedik:
danke!

PR #36502, opened 2021-12-31
[4] remove Factory::getDbo() deprecated

alikon:
### Summary of Changes
get DatabaseDriver from container

### Testing Instructions
code review

### Actual result BEFORE applying this Pull Request
@deprecated 5.0 — Load the database service from the dependency injection container

### Expected result AFTER applying this Pull Request
no more `Factory::getDbo(); @deprecated 5.0 
`

PR #35888, opened 2021-10-24
Getdbo patch fixes #35884

basd82:
Made pull request like sugested by Richard Fath

I made this change to 4.0-dev branch becouse they are fixes that are only valid voor 4.x

PR #36511, opened 2021-12-31
[4] Bye bye getDBO() (and array() in the same files) 

PhilETaylor:
Replacement for #35888 and #36502

### Summary of Changes

1) The manual check of every "getDBO" string in the file base, replacing those calls where appropriate to use `Factory::getContainer()->get('DatabaseDriver')`, some could already use the local class property `$this->_db` so those have been refactored, and some were calling getDBO multiple times where not needed, so those have been cleaned up too

2) In the files modified when completing 1 above,  Some return types in docblocks have been updated to reflect the true state of the current code, as they were highlighted in red in my IDE - ONLY in the files modified in 1 above.

3) some minor rearranging and refactoring of code around the getDBO calls to make them flow better, and be "nicer"

4) Manual change of `array()` to `[]` syntax for ALL arrays - ONLY in the files modified in 1 above.

This initial PR is 3.0 hours work. 

### Testing Instructions

Code Review
Test everything multiple times  :-) 

Technical Debt

Less Technical Debt

None.


administrator/components/com_actionlogs/src/Model/ActionlogModel.php
```diff
@@ -44,8 +44,9 @@ class ActionlogModel extends BaseDatabaseModel
 	 */
 	public function addLog($messages, $messageLanguageKey, $context, $userId = null)
 	{
+		/* @var \Joomla\Database\DatabaseDriver $db */
+		$db     = Factory::getContainer()->get('DatabaseDriver');
```
HLeithner: This looks wrong, isn't the container already injected in the model?

I'm on mobile so I can't check

PhilETaylor: changed.

PhilETaylor:
Im aware drone is broken.

PhilETaylor:
found missing underscore - drone should be fine now.

PhilETaylor:
Something broke with installation...

PhilETaylor:
ok the fringe case is the installation app.

When in the installation app the databasedriver in the container is not connected to the db, and the first step of the installation uses its own local database connection var. 

Its impossible to replace the container resource with a connected db, so, as this is a total exception, I have passed through the $db from the local var, to the `LanguageHelper::getInstalledLanguages` call and heavily documented why.

Drone should be happier now.

PhilETaylor:
Phew - build passing again :)

wilsonge:
So this one's a bit tricky for me. Whilst `Factory::getContainer()->get('DatabaseDriver')` is the direct replacement. the intention is obviously where possible to also use the container as well... as container :) and where possible start to inject the database object into the classes (obviously that only works for the non-static methods/classes - in the rest we probably substitute the container retrieval)

PhilETaylor:
This replaces the direct deprecated code with fully supported future code. Another PR can be made in the future to use dependency injection instead. This PR removes the ton of deprecated code which would end up otherwise just getting logged I’m creating huge deprecated log files

PhilETaylor:
Conflicts resolved.

PhilETaylor:
Too complex to merge it seems, and now have conflicts for the second time.

Issue #36533, opened 2022-01-02
[J4] Media Manager UX/Performance Improvements


crystalenka:
### Is your feature request related to a problem? Please describe.

When using the native Joomla 4 media manager to select an image, there are several usability issues as an end user:

- If the images in a folder are very large, or there are many images, it is very slow to load the folder (up to 2 minutes) - even on a very fast fiber internet connection. (I was testing on a site with only 100 images per folder, but some were large images up to 10 MB.)
- While loading a slow folder, you are not able to navigate to another folder until the first one is completely done loading.
- The component view of the media manager seems to cache the images for display so it's a bit faster, but I didn't notice any caching at all in the image selection view.

### Describe the solution you'd like

There are several possible solutions that would help mitigate the slow loading - any or all of them would help:

- Allow images in the media manager to lazy load so it doesn't take up resources loading off-screen images
- Add (user-configurable?) pagination to the media manager in large folders
- Automatically generate thumbnail images for images on upload and use those when picking an image - don't need to load a 3000px image when it's only being displayed at ~200px wide.
- Cache images in the image selection view so once it's loaded it doesn't have to be loaded from scratch again for the next image field.

I don't know how to fix not being able to navigate folders until the first one is done loading.

### Additional context

Before opening this issue I did some looking around at past issues. This has been a hot topic in the past without any resolution for over a year, it would be nice to solve it now. :)

I noticed that the media manager is not loading inline images (with `<img>` tags), but instead it's setting the images as background images of divs. This seems inefficient but I'm not sure of the direct performance impact. This would definitely be a roadblock to adding native lazy loading though.

I would open a PR to try to solve some of these issues but it seems the media manager is written in Vue.js, making it completely beyond my skill set.

brianteeman:
thumbnails would definitely be my preferred solution

richard67:
Ping @dgrammatiko 😄

dgrammatiko:
I had some ideas: https://github.com/joomla-projects/media-manager-improvement/issues/510

richard67:
Related issue: #28392 .

Fedik:
Thumbnails and some kind of "vue smart scroll" (render only visible data) would really improve UX for large amount of files in view.

upd: https://github.com/tangbc/vue-virtual-scroll-list

dgrammatiko:
[...]

Issue #36302, opened 2021-12-13
[4.0]Missing title attribute of image names in Media Manager


PhocaCz:
### Steps to reproduce the issue
Hi,

not sure, if I miss some parameter or option, but I don't see full image names in Media Manager. The question is, why the image names don't include "title" attributes to display full title when hovering.

### Expected result
When I hover over the image title, the title attribute will display full image name

![img](https://i.imgur.com/bwN3XFA.png)

### Actual result
When I hover over the image title, no full image name is displayed
![img](https://i.imgur.com/ZN7zAxB.png)

### System information (as much as possible)
Joomla 4.0.4

Jan

brianteeman:
I can't think of any good reason for the title not to be displayed

brianteeman:
> I can't think of any good reason for the title not to be displayed

except for the fact the image is a background and you are actually hovering over a layer on top of the image. So even if you put the title it will never be displayed

PhocaCz:
> except for the fact the image is a background and you are actually hovering over a layer on top of the image. So even if you put the title it will never be displayed

The "title" attribute is missing on file name text below the image not on image itself. This way, it works even in Joomla 3:

So adding a title attribute to "media-browser-item-info" div should do the trick:
![img](https://i.imgur.com/ay5MbtA.png)

![img](https://i.imgur.com/xSvtoj3.png)

Joomla 3 - full title when hovering file name:
![img](https://i.imgur.com/AEPJoP6.png)

brianteeman:
I didnt think of trying to put the title on the div. Doing that will display the title BUT its not very intuitive as you are much more likely to hover over the image. 

How about this way and then the title is on the image area

![image](https://user-images.githubusercontent.com/1296369/149513381-9c38571c-bf23-4be8-9c9f-1165befa5935.png)

![image](https://user-images.githubusercontent.com/1296369/149513764-4d08ab68-9361-486d-8bc6-d1847fe32e3d.gif)

PhocaCz:
It is important to be able to see the full file name. Is it on the image, or the name, or both? Personally, I'd put it above both the image and the name.

Now it's more a question for the developers of the hover effect over the image. If it doesn't have any negative effect on this effect. If not, then the full name is fine with the hover effect over the image.

Shubhamverma2796:
please test #36754.

richard67:
Closing as having a pull request. Please test #36754 . Thanks in advance.

PR #36550, opened 2022-01-03
[4.2] Switch div to img and use the loading attr. [Media Manager images list]

dgrammatiko:
Pull Request for Issue #36533 [Solves 1/3 reported problems, the thumbs and the streaming/paginated response are still valid... ].

### Summary of Changes
- Add and intersectionObserver so the browser won't try to fetch every image in the list view (only the ones that are visible)

### Testing Instructions
- Apply this PR and run `npm ci` or d/l the package from github
- navigate to different subfolders in the `images` folder
- Open the browser inspector, select the Network tab and then the img tab. As you scroll in the list view more images will appear both in the list view and in the network/img tab. The Network/img should have less images that the total amount of images in the list (assuming the viewport is not revealing all the images and you haven't scrolled to the end of the list)

All images fetched immediately

Images are fetched as they enter the viewport

No

Ping @laoneo @Fedik @crystalenka


laoneo:
Should we not just add the `loading="lazy"` attribute to the image tag as the dimensions are available for the images. We should use the platform as much as possible said once a Joomla JS developer (can't remember his name :smile: ).

crystalenka:
> Should we not just add the `loading="lazy"` attribute to the image tag as the dimensions are available for the images. We should use the platform as much as possible said once a Joomla JS developer (can't remember his name :smile: ).

There are no img tags in the media manager currently - all images are loaded as background images of div elements.

laoneo:
What is the benefit of loading them as background images?

dgrammatiko:
> We should use the platform as much as possible said once a Joomla JS developer (can't remember his name 😄 ).

This sounds obnoxious. Why would you ever want to use the browser's provided APIS instead of building them again and having a gazillion of bugs?

Joke apart the `loading=lazy` is an attribute for the img element and as @crystalenka already mentioned the design is asking for a style="background-image: ..."

> What is the benefit of loading them as background images?

No alt (obviously this is an a11y problem) and it was easier to style/append the edit buttons/checkbox on top (?)

crystalenka:
I have absolutely no idea, it seems inefficient to me but I'm not sure. Just commenting why lazy loading wouldn't work as the media manager works now.

crystalenka:
> No alt (obviously this is an a11y problem) and it was easier to style/append the edit buttons/checkbox on top (?)

How wonderful it would be to manage alt text globally! But this is beyond the scope of this issue/pr, haha. 

I am happy to help style the edit buttons over the top of inline images if that is more efficient/a necessary change.

dgrammatiko:
Ok, switched to the image tag and used the loading attribute. The styling is a bit of but some `object-fit: contain;` or something similar should get us back to the previous state (design-wise). @crystalenka @laoneo 

Screenshot 2022-01-03 at 18 30 34

administrator/components/com_media/resources/scripts/components/browser/items/image.vue
```diff
@@ -75,5 +75,21 @@ export default {
...
+        : `${this.item.thumb_path}`;
+    },
+    width() {
+      return this.item.width;
+    },
+    height() {
+      return this.item.height;
+    },
+    altTag() {
+      return `Image filename: ${this.item.name}`;
```
dgrammatiko: @brianteeman any ideas what would be a good alt tag for the images view in the MM?

crystalenka:
I'll test this and check styling later today or tomorrow. Thank you!!

administrator/components/com_media/resources/scripts/components/browser/items/image.vue
brianteeman: ask the accessibility team

brianteeman: but its currently completely inaccessible so i doubt it matters

administrator/components/com_media/resources/scripts/components/browser/items/image.vue
```diff
@@ -6,10 +6,7 @@
   >
     <div class="media-browser-item-preview">
       <div class="image-background">
-        <div
-          class="image-cropped"
-          :style="{ backgroundImage: getHashedURL }"
-        />
+        <img class="image-cropped" :src="getURL" :alt="altTag" loading="lazy" :width="width" :height="height" >
```
Quy:
```suggestion
        <img class="image-cropped" :src="getURL" :alt="altTag" loading="lazy" :width="width" :height="height">
```

laoneo:
I first test shows the attributes loading, width and height in the image. But they are loaded initially in the network tab on FF and chrome.

Additionally has the image  container not the same dimensions as the folders:
image

dgrammatiko:
> Additionally has the image container not the same dimensions as the folders:

Yes, switching to the image element needs some scss fixes

> I first test shows the attributes loading, width and height in the image. But they are loaded initially in the network tab on FF and chrome.

If you roll to https://github.com/joomla/joomla-cms/pull/36550/commits/87d01f8cfcc2cdf8ba64b640b7bcbb4da9a4ffff you'll see that the intersectionObserver is a better fit here. The reason is that there's something weird with the zoom-in/out functionality that triggers the browser to always consider more images in the viewport than what's actually there. Didn't spend any time figuring out what's triggering this (container or item size, or anything else)

laoneo:
> The reason is that there's something weird with the zoom-in/out functionality that triggers the browser to always consider more images in the viewport than what's actually there. Didn't spend any time figuring out what's triggering this (container or item size, or anything else)

Pretty sure you can fix this :wink:

crystalenka:
I have tested this item :white_check_mark: successfully on d903326dfea192046d2028f942fe67ddc0726f17<br /><br />If you add the following CSS rules to .image-cropped it fixes the styling issue in modern browsers:
```
aspect-ratio: 1/1;
object-fit: contain;
```

Aspect ratio is supported in most recent releases of most modern browsers (https://caniuse.com/?search=aspect-ratio), I'm not sure what J4 is supposed to support but if necessary I can do this a different way. This is just the most future-proofed way to do it.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36550">issues.joomla.org/tracker/joomla-cms/36550</a>.</sub>

administrator/components/com_media/resources/scripts/components/browser/items/image.vue
```diff
@@ -6,10 +6,14 @@
   >
     <div class="media-browser-item-preview">
       <div class="image-background">
-        <div
+        <img
           class="image-cropped"
-          :style="{ backgroundImage: getHashedURL }"
-        />
+          :src="getURL"
+          :alt="altTag"
```
Quy, line 13: Please consider adding `title` attribute as reported in #36302.

brianteeman: adding alt here has no impact as you can not focus on the issue and adding title here will have no impact and will not resolve that issue for the same reason

Quy:
SVG not displaying correctly.

36550-svg

``

infograf768:
@Quy 
Same result as you in Firefox and Chrome
Screenshot 2022-01-12 at 10 53 03

Same error but different display in Safari:

Screenshot 2022-01-12 at 10 51 05

dgrammatiko:
@laoneo will take over this one

laoneo:
Please test #36930 which is the followup on this one.

Issue #36548, opened 2022-01-03
[J4] Improving back end display of custom fields "list" and "SQL"


crystalenka:
### Is your feature request related to a problem? Please describe.
For the custom field types 'list' and 'SQL', if multiple = true, we are shown a list in a box where to select multiple items you have to hold down CMD or CTRL.

<img width="900" alt="" src="https://user-images.githubusercontent.com/9141288/147939288-087c4cf5-1b88-4cf4-9af9-111eaf77cf5c.png">

This is hard for some people to understand and use, and seems terribly old-fashioned (web-wise).

### Describe the solution you'd like
At a minimum, it would be nice if these fields rendered with the core fancy select list when multiple items are possible.

<img width="300" alt="" src="https://user-images.githubusercontent.com/9141288/147940058-63738f3d-adb7-4847-890f-e25efe5c750c.png">

### Additional context
For consistency, it could also be a better user experience to use the fancy select even when selecting one item; the default drop-down is cumbersome when there are many items and it would be nice to be able to type to find a specific item. This is consistent with the category selection in articles and in other places.

<img width="300" alt="" src="https://user-images.githubusercontent.com/9141288/147940120-429aad70-05e3-48d7-a69d-1aa5c3c4d2a8.png">

nikosdion:
@laoneo Wanna take a look at this? This sounds like a reasonable way to implement it, possibly with an option in the field plugin.

brianteeman:
I think there is already an open issue for this.

laoneo:
This would be a new option where the layout can be defined. The choices.js ones need the layout `joomla.form.field.list-fancy-select`.

nikosdion:
If there was an option to define the layout I'd **love** it. I have several use cases on my site and the upcoming version of my Ticket System software where it'd made a hell of a lot more sense than the default layout. Would you have the time to do a PR? I can promise a test from both me and Crystal — each one of us has a different use case where we'd need it.

crystalenka:
> This would be a new option where the layout can be defined. The choices.js ones need the layout `joomla.form.field.list-fancy-select`.

Allowing the layout to be defined would be absolutely utterly MAGICAL and open so many doors for custom fields for advanced integrators. 😮

> I think there is already an open issue for this.

I didn't see any, I did check!

laoneo:
If we call it layout, we get into a collision with #36265. So we need another wording as this is for the layout of the field for form rendering.

crystalenka:
[...]

PR #36551, opened 2022-01-03
[4.1]Add form layout option to custom fields

laoneo:
Pull Request for Issue #36548.

### Summary of Changes
Makes the form layout for custom field configurable. Actually only for the following fields:
- list
- sql
- radio

Any suggestions for language strings?

### Testing Instructions
- Create a list custom field where multiple is set to yes
- Change the form layout to enhanced list in the options tab
- Edit an article

The list is rendered with the default select HTML Element.

The list is rendered with the choices.js script.


brianteeman:
Is this small bug part of this pr?

radio

brianteeman:
Otherwise everything looks great to me

laoneo:
No, this looks like an init issue on the JS part which I didn't touch. But could also not reproduce the issue on my end.

brianteeman:
ok

brianteeman:
I have tested this item :white_check_mark: successfully on 5da3d2a109303a0930d383f5bc932d98df13fb94<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36551">issues.joomla.org/tracker/joomla-cms/36551</a>.</sub>

Shubhamverma2796:
I have tested this item :white_check_mark: successfully on 5da3d2a109303a0930d383f5bc932d98df13fb94<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36551">issues.joomla.org/tracker/joomla-cms/36551</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36551">issues.joomla.org/tracker/joomla-cms/36551</a>.</sub>

crystalenka:
I have tested this item :white_check_mark: successfully on 5da3d2a109303a0930d383f5bc932d98df13fb94<br /><br />Beautiful!!! Thanks so much, and also for including the radio layout!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36551">issues.joomla.org/tracker/joomla-cms/36551</a>.</sub>

crystalenka:
I reproduced part of Brian's issue. It's because the layout expects the values to be in the order of No - Yes; if they are flipped, it doesn't respect that. I'm wondering if there is any way to detect the values of 0/1 or true/false or yes/no and make it a bit 'smarter'? I'm not sure this is something that would prevent merging the PR, but wanted to clarify the bug.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36551">issues.joomla.org/tracker/joomla-cms/36551</a>.</sub>

laoneo:
Thanks for clarification. When the order mathers, then we need to mention this in the docs.

If this should be handled on the code side, then I would do it in another pr.

bembelimen:
Thx

laoneo:
Can you guys also test #36553 which does the same for the color field.

AndySDH:
@laoneo I just ran into this and I noticed this is the wrong placement for this setting.

The "Options" tab is dedicated to global settings that are shared by all field types. It's not the place to add type-specific settings.

Options tied to specific types of fields should be added in the first "General" tab. This is what has always been done for all field types and field settings.

So this new "Form Layout" should be moved under the "General" tab for all the fields in which you added it. For example, under the "Multiple" setting:

image

But: **Do we even need this setting at all?**

I would argue: do we really need a setting? The other "raw" layout is ugly and unusable. I don't think anyone will ever want to use it.

The new one it's just better, and it's similar to what he had in J3.

I personally would just implement the "enhanced select" by default, and get rid of the other layout altogether. Like we had in J3. What do you think?

laoneo:
Please open a new issue, so it can be discussed properly as this pr is closed.

PR #36557, opened 2022-01-04
Impressionestudio patch: simple article classes

impressionestudio:

### Summary of Changes
The proposed changes affect the class names that relate to an article. The purpose is to have simple, short and easy to remember class names in order to create more complex article layouts in an easier way (with the help of a few changes that I will also propose for custom fields).

Specifically:
- When new class names are added, the old and not useful class names remained for compatibility reasons.
- The old and not useful class names are separated with the special character | so the developers will distinguish them and not use them. (The special character is allowed according to https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors.)

### Notes:
- Some guys recommended to use BEM for naming the classes and initially I was using BEM in the proposed changes. But I cancelled using BEM because I don't find it useful. Using BEM, the class names would remain complex and very long.
Also, I think that BEM changed a little some naming rules and I think that there are some inconsistencies in the documentation. For example, the class "search-form__button_size_m" should be "search-form__button_size-m" since the proposed pattern is "block-name__element-name_modifier-name_modifier-value". 
It seems, to me, that BEM has a lot of rules that are not straight. So the final rule I use is: if it is not simple, I do not use it.

For many years, I create websites with high quality templates from scratch (only with Joomla of course) and I don't use purchased templates. So I have faced many cases and requirements and I believe that the proposed changes make things way much better.

(At the bottom of the following closed pull request, you can find a useful chat: https://github.com/joomla/joomla-cms/pull/36422)

### DIFF
--- a/components/com_content/tmpl/article/default.php
+++ b/components/com_content/tmpl/article/default.php
@@ -98,7 +98,7 @@ $isExpired         = !is_null($this->item->publish_down) && $this->item->publish
 	<?php if (isset ($this->item->toc)) :
 		echo $this->item->toc;
 	endif; ?>
-	<div itemprop="articleBody" class="article | com-content-article__body">
+	<div itemprop="articleBody" class="content | com-content-article__body">
 		<?php echo $this->item->text; ?>
 	</div>
 
--- a/components/com_content/tmpl/article/default.php
+++ b/components/com_content/tmpl/article/default.php
@@ -33,10 +33,10 @@ $currentDate       = Factory::getDate()->format('Y-m-d H:i:s');
 $isNotPublishedYet = $this->item->publish_up > $currentDate;
 $isExpired         = !is_null($this->item->publish_down) && $this->item->publish_down < $currentDate;
 ?>
-<div class="com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
+<div class="article <?php echo $this->pageclass_sfx; ?> | com-content-article item-page<?php echo $this->pageclass_sfx; ?>" itemscope itemtype="https://schema.org/Article">
 	<meta itemprop="inLanguage" content="<?php echo ($this->item->language === '*') ? Factory::getApplication()->get('language') : $this->item->language; ?>">
 	<?php if ($this->params->get('show_page_heading')) : ?>
-	<div class="page-header">
+	<div class="title | page-header">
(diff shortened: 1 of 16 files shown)


dgrammatiko:
I don't think that the `|` is working the way you think it is...

brianteeman:
I will stick by my original comments. These changes are just for personal preference. There are many ways to do things and you can't satisfy everyone all the time. Thats why we have template overrides

impressionestudio:
> I don't think that the `|` is working the way you think it is...

I only want to use it as a symbol for separating the new classes from the old ones. According to StackOverflow, it will not break the code and of course it will not be used as a class. It is something like a comment in the classes.

dgrammatiko:
> and of course it will not be used as a class

Yes because it's a hack, that was my point. Personally, I would prefer not to fight with hacks imposed by the core installation but have solid class names that I can base/extend my work on

chmst:
Already said: Best practice and easy to use are template overrides for own class names.

impressionestudio:
The important issue here is not if it is a personal recommendation, but if it is correct. Facing the issue by using template overrides doesn't mean that there is no problem or that there isn't any better solution than template overrides.

I didn't expect you to totally agree, but at least to admit that there is need to do some improvements. Maybe someone else has better ideas to add. 

Did anybody check the code changes? Nobody said what is wrong.
- Why a template needs two classes "article-info text-muted" in order to give styling to the info elements (e.g. published, created etc) of an article?
- Why there isn't a tag (span) to separate (so I can hide, as an example) the title from the actual value in an info element? You expect to create an override for something simple like that?
- Why the article class names sometimes use "article", sometimes use "page", sometimes "item"?

It seems that in Joomla 5 we will create menus, articles and template overrides using voice and with the help of artificial intelligence, but we will still have the same classes: com-content-article__body, page-header, item-image, and article-info together with text-muted of course! :D:D:D

dgrammatiko:
>  that there isn't any better solution than template overrides.

Actually Joomla users should ALWAYS use overrides for tweaking HTML

> but at least to admit that there is need to do some improvements

The project decided to follow the BEM convention, your proposal is against that decision and actually reverts any work already done in the direction of adhering to BEM

bembelimen:
Hi @impressionestudio thank your for your contributions, I really appreciate your work. But still the arguments from the people here are valid. When we change something then to a standard, not adding classes just because they are "short".

So I will close this PR.

PR #36562, opened 2022-01-04
[4.1] Gettting the router should not be static

laoneo:
### Summary of Changes
The `getRouter` function is still static in the application classes, but it is used in core on instances. Example is here https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Router/Route.php#L133.

This pr removes the static call as applications can have different implementations and the router of the application should always be accessed on the instance and not statically.

It became static 13 years ago in https://github.com/joomla/joomla-cms/commit/7175d2f3808a8c17421ec988e2ec144ca7f4e263. Time to revert it as we do not work anymore on static application access, more on the instance itself.

This can be considered a BC break but in my eyes, it is a wrong implementation.

cc: @wilsonge 

### Testing Instructions
Open the front end and back end.

All is working.

All is working.


PhilETaylor:
> This can be considered a BC break but in my eyes, it is a wrong implementation.

Probably not possible in 4.1.x then, needs to be done in 5... 

We dont yet have a branch for 5.0-dev :)

wilsonge:
I agree that it's wrong. But it's also a fairly painful b/c break to resolve :( I can't remember the details or discussions - but I remember having to just end up fixing all the current method calls way back https://github.com/joomla/joomla-cms/pull/3548  - this was just due to a phpunit upgrade. but i'm sure there was a conversation about making it non-static again too somewhere. https://developer.joomla.org/joomlacode-archive/issue-33688.html?highlight=WzMzNjg4XQ==

I assume we'll have strict errors again after this in the places I changed above (if they still exist - it's possible with the new router lots of that stuff was cleaned up...) but even if we fix them we still have to account for 3rd party plugins doing similar things :( Thinking things like SH404 or other such extensions that overload bits of the router.

Overall completely agree it shouldn't be static. It just might need to be a 5.0 fix :(

laoneo:
So we have already strict errors right now. Should we fix it in 4 and then in 5 revert to none static?

libraries/src/Application/SiteApplication.php
```diff
@@ -378,7 +378,7 @@ public function getPathway($name = 'site', $options = array())
 	 *
 	 * @since	3.2
 	 */
-	public static function getRouter($name = 'site', array $options = array())
+	public function getRouter($name = 'site', array $options = array())
```
PhilETaylor:
```suggestion
	public function getRouter($name = 'site', array $options = [])
```

libraries/src/Application/CMSWebApplicationInterface.php
```diff
@@ -53,7 +53,7 @@ public function getMenu($name = null, $options = array());
 	 *
 	 * @since   4.0.0
 	 */
-	public static function getRouter($name = null, array $options = array());
+	public function getRouter($name = null, array $options = array());
```
PhilETaylor:
```suggestion
	public function getRouter($name = null, array $options = []);
```

libraries/src/Application/AdministratorApplication.php
```diff
@@ -199,7 +199,7 @@ protected function doExecute()
 	 *
 	 * @since	3.2
 	 */
-	public static function getRouter($name = 'administrator', array $options = array())
+	public function getRouter($name = 'administrator', array $options = array())
```
PhilETaylor:
```suggestion
	public function getRouter($name = 'administrator', array $options = [])
```

libraries/src/Application/CMSApplication.php
```diff
@@ -597,7 +597,7 @@ public function getPathway()
 	 *
 	 * @since   3.2
 	 */
-	public static function getRouter($name = null, array $options = array())
+	public function getRouter($name = null, array $options = array())
```
PhilETaylor:
```suggestion
	public function getRouter($name = null, array $options = [])
```

laoneo:
Closing in favor of #36688.

PR #36068, opened 2021-11-17
Correct Cache setWorkarounds to take account of multiple script or st…

GeraintEdwards:
…yle declarations being used that mean that we have an array of scripts or stylesheets to be cached and not just a string

Cache setWorkarounds returns an array not a string

Pull Request for Issue # .

### Summary of Changes
Correct Joomla\CMS\Cache Cache setWorkarounds to take account of multiple script or style declarations being used that mean that we have an array of scripts or stylesheets to be cached and not just a string
Joomla\CMS\Cache Cache setWorkarounds returns an array not a string

### Testing Instructions

1. On a Joomla 4-dev site install this module https://www.yoursites.net/testextensions/mod_cachetest.zip
2. Create **2** instances of this module to display on the same page in different module positions - leave caching setting as 'global' in module settings
3. Enable Joomla conservative caching  in Joomla config
4. Enable maximum error reporting in Joomla config#
5. Make sure the Joomla cache is cleared
6. Load the frontend page where modules are displayed
7. Reload the frontend page

### Actual result BEFORE applying this Pull Request

At Step 6 you get a php warning on arguments to mb_strlen and 4 javascript alert messages

See https://www.yoursites.net/testextensions/Screenshot%202021-11-17%20at%2010-05-05%20Home.png

At step 7 you get a different PHP warning from libraries/src/Document/HtmlDocument.php because it expects an array but has been given a script by Cache.php

### Expected result AFTER applying this Pull Request

Page is displayed with no warning messages and module content is delivered from the cache correctly with 4 alerts

PR #36566, opened 2022-01-04
PHP error on PHP 8 when caching is enabled under a specific edge case

nikosdion:
### Summary of Changes

This PR fixes a caching issue which occurs when a module which defines head scripts or styles is being cached after another extension (e.g. template, plugin, or module) has already defined its own, different, head scripts or styles.

This is a type error in Joomla's cache handler which exists in Joomla 3 as well. This PR is for Joomla 4.0.

Tagging @crystalenka (reported the issue to me), @wilsonge and @HLeithner for review.

### Testing Instructions

* Install a new Joomla 4 site. **Make sure you are using PHP 8.0**
* Login to the backend of your site
* Go to System, Install, Extensions
* Install [the Foobar test plugin](https://cdn.dionysopoulos.me/joomla-prs/cache-edge-case/plg_system_foobar.zip)
* Install [the Foobar test module](https://cdn.dionysopoulos.me/joomla-prs/cache-edge-case/mod_foobar.zip)
* Go to System, Manage, Plugins and enable the System - Foobar plugin
* Go to Content, Site Modules, click on New
* Choose the “Foobar Test Module”
* Use the following information
	- Title: Test
	- Position: Main-top [main-top]
* Click on Save & Close
* Go to your site's public frontend

**Checkpoint 1**

* Back in the backend of your site, go to Home Dashboard and click on Global Configuration
* Click the System tab and set System Cache to ON - Progressive caching
* Click on Save & Close

**Checkpoint 2**

**Checkpoint 1**

The site works properly and you can see the not cached Foobar module.

**Checkpoint 2**

Joomla shows an error page (“The requested page can't be found.”) with the following error message:

> If difficulties persist, please contact the website administrator and report the error below.
> 
> 0 mb_strlen(): Argument #1 ($string) must be of type string, array given

**Checkpoint 1**

The site works properly and you can see the not cached Foobar module.

**Checkpoint 2**

The site works properly and you can see the cached Foobar module.

None whatsoever.

### Technical information

The two extensions are dummies which add head styles in the document at different points of the application lifecycle just to trigger the error condition. It should also be possible if you have two **DIFFERENT** module extensions doing the same, or a module and a template, etc.

The Joomla cache handler (`\Joomla\CMS\Cache\Cache`) has a major type bug in the `setWorkarounds` method. When it compares the `script` and `style` keys of the document object's head data it _incorrectly assumes_ that they contain string values. In actual reality they always contain EITHER an empty string OR an array.

Joomla then calls `\Joomla\String\StringHelper::substr()` and `\Joomla\String\StringHelper::strlen()` against the contents of these header keys. If they are non–empty, i.e. the contain _an array_, this causes a type error.

Under PHP 5 and PHP 7 the type mismatch error is a mere PHP warning. You might see it, most likely you won't (because you're likely running your site with error reporting set to None) and caching doesn't work correctly for ‘inexplicable reasons’.

Under PHP 8 type mismatch errors have been upgraded to ErrorExceptions i.e. fatal PHP errors. This is what is caught by Joomla's error handler and why you see an error page instead of the cached module.

My suspicion is that this code was copied from The Olden Days (Joomla 1.x / 2.x and possibly very early 3.x, if memory serves) when Joomla indeed stored a string in the document instead of an array. When the document object was refactored many moons ago this largely undocumented, esoteric piece of code was not updated. The kind of failure it caused was silent under PHP 5 and 7 so any caching issues it caused were attributed to the mythical ‘ghost in the machine’, the cache expired or was reset and everything was good in the world. Come PHP 8 and the root cause, the type mismatch, is now a stop error and suddenly the ghost in the machine grew fangs and bit us in the posterior.

I also suspect without being certain that some Joomla forum posts may have mistakenly blamed an innocent module for this issue, possibly claiming that the plugin is not compatible with PHP 8. Please let me say again that this issue is 100% unrelated to any third or first party extension. It is a bug in Joomla's cache handler. Fun times...

### DIFF
--- a/libraries/src/Cache/Cache.php
+++ b/libraries/src/Cache/Cache.php
@@ -599,14 +599,14 @@ class Cache
 	 *
 	 * @since   1.7.0
 	 */
-	public static function setWorkarounds($data, $options = array())
+	public static function setWorkarounds($data, $options = [])
 	{
-		$loptions = array(
+		$loptions = [
 			'nopathway'  => 0,
 			'nohead'     => 0,
 			'nomodules'  => 0,
 			'modulemode' => 0,
-		);
+		];
 
 		if (isset($options['nopathway']))
 		{
@@ -657,7 +657,7 @@ class Cache
 			if ($loptions['modulemode'] == 1)
 			{
 				$headnow = $document->getHeadData();
-				$unset   = array('title', 'description', 'link', 'links', 'metaTags');
+				$unset   = ['title', 'description', 'link', 'links', 'metaTags'];
 
 				foreach ($unset as $un)
 				{
--- a/libraries/src/Cache/Cache.php
+++ b/libraries/src/Cache/Cache.php
@@ -599,14 +599,14 @@ class Cache
 	 *
 	 * @since   1.7.0
 	 */
-	public static function setWorkarounds($data, $options = [])
+	public static function setWorkarounds($data, $options = array())
 	{
-		$loptions = [
+		$loptions = array(
 			'nopathway'  => 0,
 			'nohead'     => 0,
 			'nomodules'  => 0,
 			'modulemode' => 0,
-		];
+		);
 
 		if (isset($options['nopathway']))
 		{
@@ -684,9 +684,15 @@ class Cache
 						{
 							foreach ($newvalue as $type => $currentScriptStr)
 							{
+								$currentScriptStr = is_array($currentScriptStr)
+									? implode("\n", $currentScriptStr) : $currentScriptStr;
+
 								if (isset($options['headerbefore'][$now][strtolower($type)]))
 								{
 									$oldScriptStr = $options['headerbefore'][$now][strtolower($type)];
+
+									$oldScriptStr = is_array($oldScriptStr)
+										? implode("\n", $oldScriptStr) : $oldScriptStr;
 
 									if ($oldScriptStr != $currentScriptStr)
 									{


libraries/src/Cache/Cache.php
```diff
@@ -599,14 +599,14 @@ public static function getWorkarounds($data, $options = array())
 	 *
 	 * @since   1.7.0
 	 */
-	public static function setWorkarounds($data, $options = [])
+	public static function setWorkarounds($data, $options = array())
 	{
-		$loptions = [
+		$loptions = array(
```
brianteeman: please revert this change

nikosdion: Why? This is valid syntax as of **PHP 5.4**. The minimum requirement for Joomla 4 is **PHP 7.2**, seven minor versions newer. 

I also don't see anything in the code style (and the automated checks) which prevents short array syntax. 

It's 2022. If we can't use syntax introduced in PHP in March 2012, nine years and 10 months ago, we are doing something REALLY WRONG with this project...

brianteeman: Look again. You are removing short syntax [] and replacing it with long array syntax ()

nikosdion: That's weird. It looks like my main branch wasn't properly updated. I changed it manually.

nikosdion: Well, now I converted an old long array syntax to short so yay, at least it's consistent :)

brianteeman: You did say that your brain was hurting

toivo:
I have tested this item :white_check_mark: successfully on a58fedce30bdf404d350020f5eb665d467d6bbdc<br /><br />Tested successfully in 4.0.6-dev of January 4 using PHP 8.0.13 and 8.1.0 in Wampserver 3.2.6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36566">issues.joomla.org/tracker/joomla-cms/36566</a>.</sub>

crystalenka:
I have tested this item :white_check_mark: successfully on a58fedce30bdf404d350020f5eb665d467d6bbdc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36566">issues.joomla.org/tracker/joomla-cms/36566</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36566">issues.joomla.org/tracker/joomla-cms/36566</a>.</sub>

Fedik:
The fix is incorrect, you should not use implode, but compare array instead. 
(Use of Implode will have different result for the same scripts in defferent order).
it nothing with PHP. It because Joomla 4 use array to store each script, while Joomla 3 a string.

And we already have a fix for this #36068 , please confirm if it works for you.

richard67:
Back to pending due to previous comment. Please check.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36566">issues.joomla.org/tracker/joomla-cms/36566</a>.</sub>

nikosdion:
@Fedik Yeah, thanks, I know why it happens (I actually wrote about that in length at the end of the PR).

I don't see how the scripts / styles would ever be in a different order since what this code checks is whether a script / style that wasn't there has been appended to the document's head. Joomla can only append to the document head, it cannot inject in a different order.

In any case, the point is moot. The other PR is more economical code-wise. I prefer that approach instead of mine.

wilsonge:
Merged the other one

Issue #36567, opened 2022-01-04
Assets not loaded after upgrading from 4.0 to 4.1


joeforjoomla:
### Steps to reproduce the issue
1) Install 4.0.5

2) Upgrade to 4.1 package

3) Update the database:
![image](https://user-images.githubusercontent.com/17835460/148115248-57b5df0f-9b5f-40da-9d78-a1be7100a136.png)

### Expected result
All is working

### Actual result
CSS and assets are lost no longer loaded. It seems that they can't be found in the new 'media' folder but only under 'classic' templates folder.
![image](https://user-images.githubusercontent.com/17835460/148115407-4a34cbb3-5368-4af4-83e4-69a20f748226.png)

brianteeman:
confirmed and why I raised https://github.com/joomla/joomla-cms/issues/36423

dgrammatiko:
I have no clue how you tried to upgrade to 4.1 but the 2 methods that I used worked flawlessly:
- providing the 4.1 update zip file
- using a custom update channel: 
![Screenshot 2022-01-05 at 11 39 32](https://user-images.githubusercontent.com/3889375/148205998-96a75470-1f38-456b-a960-bb6770ef7b75.png)

brianteeman:
1. Thats not the beta3
2. did you try database structure fix afterwards

dgrammatiko:
1. Yes, it the nightly, which is newer than the beta 3
2. No because there wasn't any reason to do something manually. (BTW the db structure will **ONLY** fix the structure, will not apply the table content transforms as they are not structure)
![Screenshot 2022-01-05 at 12 03 13](https://user-images.githubusercontent.com/3889375/148207806-f6332426-9fc9-4af4-a0f3-a8ad7f8ab97a.png)

joeforjoomla:
@dgrammatiko 
1. There are no updates for custom update channel
2. To upgrade, i've used the manual zip file and AFTERWARDS updated the database. And assets are not found in the media folder, but STILL searched under the templates folder
![image](https://user-images.githubusercontent.com/17835460/148208754-62bfc12a-4d9e-45bd-b5b5-e4ed910b16cd.png)

dgrammatiko:
The static files exist in the update package: 
![Screenshot 2022-01-05 at 12 18 24](https://user-images.githubusercontent.com/3889375/148209402-2a785e9f-7ff1-4cf0-89c4-1c8a21406abc.png)

[...]

PR #36574, opened 2022-01-05
[4.1] Fix the update failure on Windows

dgrammatiko:
Pull Request for Issue #36567 .

### Summary of Changes
Upgrading from 4.0.x to 4.1 fails due to some FS inconsistencies

@richard67

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -8506,10 +8506,10 @@ class JoomlaInstallerScript
 
 		foreach ($folders as $oldFolder => $newFolder)
 		{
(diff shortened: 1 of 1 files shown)


wilsonge:
Isn't just you need to wrap the function call in realpath for Folder::exists. because realpath will do the path conversion for slashes (see the second example here https://www.php.net/manual/en/function.realpath.php)

wilsonge:
if your using real_path my point is you shouldn't need the str_replace on top of that

dgrammatiko:
> if your using real_path my point is you shouldn't need the str_replace on top of that

I don't have a windows PC so I try things blindly here. Will remove them

brianteeman:
its nothing to do with the file path as expained in the other issue

PR #36576, opened 2022-01-05
[4.1] Update fix 4.1.0-2021-11-20.sql

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Fix empty parameter for mail template.

### Testing Instructions

1. Make a new installation of 4.1 Beta 3 and check in database the parameter modified by this PR here.
Result: The parameter is wrong `'{"tags": ["task_id", "task_title", ""]}'`.
2. Update the 4.1 beta 3 from step 1 to this PR by using either the patched package https://ci.joomla.org/artifacts/joomla/joomla-cms/4.1-dev/36576/downloads/50275/Joomla_4.1.0-dev+pr.36576-Development-Update_Package.zip or the custom update URL https://ci.joomla.org/artifacts/joomla/joomla-cms/4.1-dev/36576/downloads/50275/pr_list.xml , check if it works without SQL errors and check in database if the empty parameter changed by this PR has been fixed.
Result: The parameter is right `'{"tags": ["task_id", "task_title"]}'`.

None.


brianteeman:
Do we still allow updates from beta to stable? If so do we need an update query as well?

richard67:
What shall this PR fix? For sure it’s not an SQL error during update. The changes here don’t fix any error in SQL syntax.

dgrammatiko:
The params column should be a json string. I just saw that one row was inconsistent, that’s all.

richard67:
Ah, i thought it was an attempt to fix the update issue.

I don’t think we need a new update SQL because it will be fixed anyway if someone saves these options, but maybe @bembelimen has a different opinion.

dgrammatiko:
@richard67 on the last file (the jooally plg) you are also including the columns checkout and checkout_time but this file doesn’t. Would that be a prob?

richard67:
No that shouldn’t be a problem since they have default value of NULL, as far as I remember. Can’t check now because off desk.

richard67:
As agreed in our meeting, I have removed the change for the empty string instead of empty json params in the extensions table so this PR fixes only the wrong mail template params.

I have also added a new update SQL to fix that for previous beta versions, and I have changed that in supports.sql for new installations, too.

The change for the extensions table should be done in a separate PR because it needs to do the same fix for other plugins, too.

You can see that here: https://github.com/joomla/joomla-cms/blob/4.1-dev/installation/sql/mysql/base.sql#L245-L379

We are completely inconsistent with that params. It should be fixed all together.

richard67:
@brianteeman Check my previous comment. I think it answers your question. PR will come sooner or later to fix that, too.

brianteeman:
@richard67 thanks - our posts crossed.

richard67:
@brianteeman Another question is why we change the old update SQL at all if we have a new one which fixes that during the update. I know you don't like us touching the old SQL if not necessary, and I know some people thought they have to run it again when it has changed, which caused a mess. The reason why I prefer to change them is that often they are used as templates for making new update SQL in future, so if we fix them we make sure people don't copy wrong stuff. I had discussed that with @bembelimen and he agreed with me, that's why I did not revert that change in the old update script.

I thought I should explain it to you so you know why I do it like I do here.

richard67:
I have tested this item :white_check_mark: successfully on 4aec8688b7aca528850440c959a4d90291c8cc0d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36576">issues.joomla.org/tracker/joomla-cms/36576</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 4aec8688b7aca528850440c959a4d90291c8cc0d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36576">issues.joomla.org/tracker/joomla-cms/36576</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36576">issues.joomla.org/tracker/joomla-cms/36576</a>.</sub>

bembelimen:
Thx

PR #36591, opened 2022-01-06
[4.1] User defined Hide Table Columns

brianteeman:
### Summary of Changes
Decide for yourself which columns to display for any table. The button shows an updating count of how many columns are shown.

The choice of which column to display is stored in local storage so it is persistent for you in your browser for that site. 
tables2

This should be added to every table and it can also be added by extension developers to their own table by just adding one line to the existing web assets.

```
// add the script
/** @var \Joomla\CMS\WebAsset\WebAssetManager $wa */
$wa = $this->document->getWebAssetManager();
$wa->useScript('table.columns');
```

### Testing Instructions
As there is new js you will need to `run npm ci`

This is a fork and continuation of work by @PhilETaylor @dgrammatiko @fedik


build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,178 @@
...
+
+    if (stored) {
+      this.listOfHidden = stored.split(',').map((val) => parseInt(val, 10));
+    }
+  }
+}
+
+if (window.innerWidth > 992) {
+  // Look for dataset name else page-title
+  document.querySelectorAll('table').forEach(($table) => {
```
dgrammatiko:
```suggestion
  [...document.querySelectorAll('table')].forEach(($table) => {
```

dgrammatiko (changes requested):

build/media_source/system/js/table-columns.es6.js
brianteeman: before I merge could you educate me on what the [... does

PhilETaylor:
# Future improvement ideas already suggested:

I know you will say out of scope of this PR, but important not to lose sight, these could easily be part of this PR. 

- Initially hide most things to "clean up" the user interface and allow users to add more clutter
- Overcome the bootstrap limitation of classes for screens less than 1024px wide and remove all responsive classes that are currently hiding things, and allow all columns on mobile devices to be toggled with table options.
- Store the preferences against the user object in the db for browser/computer independency

# Previous comments/feedback

 - The bottom of the table should be chosen for the placement of the Table Options as the page has been "designed" previously and adding them at the top just gets in the way, pushing the most important area of the page down even further

build/media_source/system/js/table-columns.es6.js
dgrammatiko: Called: Spread Operator 

MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Spread syntax (...) allows an iterable such as an array expression or string to be expanded in places where zero or more arguments (for function calls) or elements (for array literals) are expected, or an object expression to be expanded in places where zero or more key-value pairs (for object literals) are expected.

It is also available in PHP 8.1

brianteeman:
Yes thats out of scope

No I'm not moving it to the bottom. It makes no sense to hide it on the bottom of the second or third page scroll

PhilETaylor:
> No I'm not moving it to the bottom. It makes no sense to hide it on the bottom of the second or third page scroll

But you think it makes sense to push the content further down the page... even more than it already is... got it... /facepalm

brianteeman:
your face must be really bruised the number of times that you smash your palm into it

PhilETaylor:
Im sorry that you are not open to feedback and are so closed minded. Once again.

PhilETaylor:
Not implemented on 

http://127.0.0.1:4444/administrator/index.php?option=com_users&view=levels

Screenshot 2022-01-06 at 21 12 56

PhilETaylor:
Also not implemented on User Notes

PhilETaylor:
maybe something to do with this:
Screenshot 2022-01-06 at 21 14 20

Kostelano:
The first thing I noticed was that the function is not available everywhere, but in fact it makes sense to implement it on all tabular pages in the admin panel.

brianteeman:
The js is producing an error when it can't select the text to use for the column header. This happens whenever the header is not a link because for those headers there is no span containing the text

PhilETaylor:
Also "If you hide the cell that contains the scope then you break the tables accessibility. This is usually (always?) the cell that is the link."... yeah, you are now doing that...

PhilETaylor:
>  This happens whenever the header is not a link because for those headers there is no span containing the text

https://github.com/joomla/joomla-cms/pull/32974/files#diff-5ff666260572eb6e2425429e4ffa0b6815227394932a44d65540f215d4aef8f2R94-R99

brianteeman:
oops that got mixed in the js rewrite by @fedik 

@fedik can you look at putting both of those back please

dgrammatiko:
Also @Fedik please remove the `$` from the variables, the jQuery era was many years ago a thing...

PhilETaylor:
> the jQuery era was many years ago a thing...

I feel attacked ;-) Im still writing very pretty jQuery code ;-)

dgrammatiko:
> I feel attacked ;-) Im still writing very pretty jQuery code ;-)

Ok, I can rephrase the `jQuery` to `PHP`, oops...

Fedik:
I will look

> please remove the $ from the variables

Nope, and it nothing to do with jquery, it for code readability. I would suggest you to use it also.
`$` used for variables that holds DOM element. If you see the code and see a variable with $ you know it some kind of DOM reference, holds one of document elements.
Example `table` and `$table` you can easily say that last one is more likely a table element (or an element that represent the table), while for first one you have to search code definition for it, so you losing a time here.

dgrammatiko:
Dude just use tableElement instead the obnoxious $, if what you’re after is readability…

brianteeman:
get a room to argue about $ ;)

brianteeman:
The pages where the button did not appear have now been fixed - thanks @Fedik 
Just need to make the titles non-selectable now

brianteeman:
@fedik @dgrammatiko if you could take a look at putting back the code that made the title checked by default and readonly that would be the final missing piece

dgrammatiko:
@brianteeman it should be a simple conditional, what's the difference for the title elements ( I think they are using `thead`)?

brianteeman:
It's the cell with scope row and th

Fedik:
@brianteeman I have made PR,
but the problem with that logic, it will not work for empty table.
It better to have a way to detect such columns by looking in a table header (extra data- attribute or something).

brianteeman:
Thanks - I have merged the pr and am working on some tweaks. Instead of hiding the column I am going to make it checked and disabled

build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,199 @@
...
+    $button.setAttribute('aria-expanded', 'false');
+
+    const $ul = document.createElement('ul');
+    $ul.setAttribute('class', 'list-unstyled p-2');
+    $ul.setAttribute('id', 'columnList');
+
+    // Collect a list of headers for dropdown
+    this.$headers.forEach(($el, index) => {
+      // Skip the protected columns
+      if (this.protectedCols.indexOf(index) !== -1) return;
```
Fedik: > Instead of hiding the column I am going to make it checked and disabled

Then change this back as it was:
```
if (index === 0) return;
```

build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,199 @@
...
+
+      const $li = document.createElement('li');
+      const $label = document.createElement('label');
+      const $input = document.createElement('input');
+      $input.classList.add('form-check-input', 'me-1');
+      $input.type = 'checkbox';
+      $input.name = 'column';
+      $input.checked = this.listOfHidden.indexOf(index) === -1;
+      $input.value = index;
```
Fedik, line 80: > Instead of hiding the column I am going to make it checked and disabled

And here:
```
$input.disabled = this.protectedCols.indexOf(index) !== -1;
```

brianteeman:
All issues resolved and ready for final tests - please

Kostelano:
I have tested this item :white_check_mark: successfully on cb10bdeecbe4636cadc83554518ba474604e9e04<br /><br />Cool feature. The headings are pinned, everything works well. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36591">issues.joomla.org/tracker/joomla-cms/36591</a>.</sub>

PhilETaylor:
Is there a reason its not implemented in the following tables - surely if its going to be implemented on some, it should be on all.

 - Banners -> Clients
 - Banners -> Tracks
 - Users -> User Notes
 - Smart Search component ?
 - System -> Install -> Languages
 - System -> Clear Cache
 - System -> Database
 - System -> Manage -> Languages
 - System -> Manage -> Redirects
 - System -> Update Sites
 - System -> Mail Templates = Says 3 columns where there are 4

Outside the scope of this PR, would also be nice to implement something for tables in modules on a dashboard.

PhilETaylor:
Are we happy that the storage keys have spaces and colons?

Screenshot 2022-01-09 at 15 09 50
 
Also if I manipulate the value stored, then on next page load I get a JS Error. Probably ok as I doubt hackers will be changing local storage values (unless there is a way to exploit this), but could be handled better.

PhilETaylor:
Ok here is an issue. 

A 3PD writes a component called com_mytags

My main view is a Tags view. My page title is correctly "Tags". I correctly implement tableoptions

My table view has columns in a different order to the Joomla Core Tags component, or a different number of columns. I get a Javascript error as table options tries to manipulate my columns

solution is to prefix the key with something - maybe com_tags - to prevent collisions when used by other 3pd applications.

Fedik:
> If accessing a key throws an exception like this it should be caught and the key removed from local storage, else the user will forever thing that Joomla is broken

Makes sense,
enough just check if the index exists in `toggleColumn(index, force) `:
```javascript
toggleColumn(index, force) {
  if (!this.$headers[index]) return;
...
...
```

Fedik:
> Are we happy that the storage keys have spaces and colons?

As long as it works, but can change to anything.
The suffix for storage key is configurable, with `<table data-name="foobar">`

brianteeman:
checklist as I work through them - thanks @PhilETaylor 
- [x] Banners -> Clients
- [x] Banners -> Tracks
- [x] Users -> User Notes
- [x] Smart Search component ?
- [x] System -> Install -> Languages
- [x] System -> Clear Cache
- [x] System -> Database
- [x] System -> Manage -> Languages
- [x] System -> Manage -> Redirects
- [x] System -> Update Sites

brianteeman:
> System -> Mail Templates = Says 3 columns where there are 4

That's because the table is not correct and I will do a separate PR for that

administrator/components/com_cache/tmpl/cache/default.php
```diff
@@ -16,6 +16,10 @@
 
 HTMLHelper::_('behavior.multiselect');
 
+/** @var \Joomla\CMS\WebAsset\WebAssetManager $wa */
```
Quy: See line 28.

libraries/src/WebAsset/AssetItem/TableColumnsAssetItem.php
```diff
@@ -0,0 +1,40 @@
...
+	 * Method called when asset attached to the Document.
+	 * Used to add the language strings required by the script.
+	 *
+	 * @param   Document  $doc  Active document
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function onAttachCallback(Document $doc)
```
PhilETaylor, line 35:
```suggestion
	public function onAttachCallback(Document $doc): void
```

Kostelano:
Downloaded and installed the PR package from 15.01. There is a problem. Did not check on all pages, but after installing the assembly with this PR, go to the Components - Tags page. Try to select 1-2-3 items (but not all). The action button in the toolbar will be inactive.

Once you have selected all the elements, it will become active.

There is no conflict in the console.

Screenshot_1

**UPD**
This problem is on all such pages.

**UPD**
Open "Columns" and click on the element. See... 

Screenshot_2

brianteeman:
Sadly I can confirm this.

@fedik @dgrammatiko any ideas?

dgrammatiko:
- Option 1 (easy): the input check becomes unhidable as the title
- Option 2: the buttons need to be aware of the changed state of the checkboxes (some kind of event)

brianteeman:
I dont understand either ;)

Fedik:
That happens because buggy multiselect.js

brianteeman:
The issue has been fixed - thanks @fedik

Kostelano:
This feature is not available on a page with blocked elements. There are only 2 columns, of course... But I'd rather be consistent in functionality.

brianteeman:
Not sure I understand what you mean by "blocked elements".

Kostelano:
Sorry. I mean this: 

Screenshot_8

brianteeman:
ok - I didnt enable it on a page like that as it would be pointless.

Kostelano:
Ok, one more clarification on 2 pages - there is also no option to select columns.

**Private messages
Search query analysis**

Quy:
Users > Manage > Permissions ?

brianteeman:
@quy it cannot work on that view because of the way the content can be filtered by component

Quy:
Maybe add a down caret (v) to indicate that it is a dropdown.

brianteeman:
> Maybe add a down caret (v) to indicate that it is a dropdown.

Done.

Hope this makes it into 4.1 - it solves a lot of user issues

build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,204 @@
...
+            this.listOfHidden.splice(ih, 1);
+          }
+        }
+      });
+    }
+
+    // Set up toggle menu
+    this.createControls();
+
+    // Restore sate
```
Quy:
```suggestion
          // Make sure it's not in the list of hidden
          const ih = this.listOfHidden.indexOf(index);
          if (ih !== -1) {
            this.listOfHidden.splice(ih, 1);
          }
        }
      });
    }

    // Set up toggle menu
    this.createControls();

    // Restore state
```

Quy:
> > System -> Mail Templates = Says 3 columns where there are 4
> 
> That's because the table is not correct and I will do a separate PR for that

Can you please check as this is still an issue? Thanks.

build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,204 @@
...
+    $button.setAttribute('aria-haspopup', 'true');
+    $button.setAttribute('aria-expanded', 'false');
+
+    const $ul = document.createElement('ul');
+    $ul.setAttribute('class', 'list-unstyled p-2');
+    $ul.setAttribute('id', 'columnList');
+
+    // Collect a list of headers for dropdown
+    this.$headers.forEach(($el, index) => {
+      // Skip the first column as we dont want to display the checkboxes
```
Quy:
```suggestion
      // Skip the first column as we don't want to display the checkboxes
```

Quy:
Almost there.

mail-templates

Quy:
Please add Workflows.

brianteeman:
@quy I dont know how to fix mail templates - sorry

build/media_source/system/js/table-columns.es6.js
```diff
@@ -0,0 +1,204 @@
...
+    $button.setAttribute('aria-haspopup', 'true');
+    $button.setAttribute('aria-expanded', 'false');
+
+    const $ul = document.createElement('ul');
+    $ul.setAttribute('class', 'list-unstyled p-2');
+    $ul.setAttribute('id', 'columnList');
+
+    // Collect a list of headers for dropdown
+    this.$headers.forEach(($el, index) => {
+      // Skip the first column, unless its a th, as we don't want to display the checkboxes
```
Quy:
```suggestion
      // Skip the first column, unless it's a th, as we don't want to display the checkboxes
```

brianteeman:
Is there any chance of this making it into 4.1 please.

It's a much requested feature and it would be a shame for it to have to wait another 6 months.

Any improvements can be made during the 4.1 lifecycle

Quy:
I have tested this item :white_check_mark: successfully on e9fc22d229177034b1485151dec7a78a32cf2330<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36591">issues.joomla.org/tracker/joomla-cms/36591</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on e9fc22d229177034b1485151dec7a78a32cf2330<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36591">issues.joomla.org/tracker/joomla-cms/36591</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36591">issues.joomla.org/tracker/joomla-cms/36591</a>.</sub>

roland-d:
@brianteeman As Joomla 4.1 is in feature freeze Benjamin asked me to merge it for 4.2. I agree this is a great feature, so have it merged for 4.2.

Thanks everybody for working on this.

brianteeman:
thanks for merging - shame people will have to wait :(

Kostelano:
Please add a language tag to the PR

joeforjoomla:
This script is too much generic if used for third-party extensions. It searches for all tables in a page throwing a JS error if there is no head section or applying the hide columns control to all found tables even if undesired.
It's a pity it can't be used only for the main or specific table in backend. It would be great an addition to exclude or include only tables that really need it.
image

Fedik:
Well, that was expected. But people wanted like that.
You better open an issue with some "test case" description. If you like.

brianteeman:
a table without a head section is invalid so not going to apologise if it doesnt work

brianteeman:
>  It would be great an addition to exclude or include only tables that really need it.

I dont understand - you have to add it to your component to use it. its not automatic

joeforjoomla:
> > It would be great an addition to exclude or include only tables that really need it.
> 
> I dont understand - you have to add it to your component to use it. its not automatic

@brianteeman of course i know.

The problem now, is that this code should skip unnamed tables, but actually it doesn't do this because if a table has no name it fallbacks to the header title:

```
const tableName = $table.dataset.name ? $table.dataset.name : document.querySelector('.page-title').textContent.trim().replace(/[^a-z0-9]/gi, '-').toLowerCase(); // Skip unnamed table

    if (!tableName) {
      return;
    }
```

As a consequence if there are  let's say 3 tables in the backend interface of a component but not all 3 tables used for columns to be hidden, then this feature will apply to all of them.
Furthermore, if a table has an empty `<head></head> `without any `<tr>`, than you have a the javascript error because of this line of code looking at the 'children' attribute of null:

`this.$headers = [].slice.call($table.querySelector('thead tr').children);
`
Then it should be the case to fix the 'children' code to avoid a JS error, or to introduce an option for a configurable JS selector to include or exclude only certain tables.

I hope to have well explained.

brianteeman:
Two choices

1. Don't implement the script in your component
2. Submit a pull request with the required changes

impressionestudio:
Very useful option but...

Why the new dropdown list button for the columns is placed in a new line and not  in the same row and after the dropdown list button for the list limit?

As, for example, the list limit button affects the table, also does this new button for the columns of the table. So they should placed together.

There is space for them to fit in one row and if the left menu is closed then there is more space. I think we need to stop pushing the table too much down.

Also the other filters have a specific ID or class e.g. "list_limit", "list_fullordering". Why the new button for the columns doesn't have?

HSorgYves:
d-xxl-table-cell isn't removed by the JavaScript. Can you please adjust the JS?

PR #27712, opened 2020-01-30
[4.0] Media fields accessibility - second try

astridx:
Pull Request for Issue #27571 .

Please see https://github.com/joomla/joomla-cms/pull/27650 and https://github.com/joomla/joomla-cms/pull/27712#issuecomment-580956349  for test instructions.

In this Pull Request I tried to implement a suggestion from this [comment](https://github.com/joomla/joomla-cms/pull/27650#issuecomment-579773758
) - I use the Subform Field.

Edit: 10th of February
Test Instructions - You do NOT need composer or npm

1. Create a Custom Field of type media and see, that this field did not include an alternative text in front end. I named this custom field `old`.
2. Apply this patch and create a new custom field of type media. See hat all parameters, for example directory, are working. I named this custom field `new`.
3. See, that the custom fields `new` is still shown correctly in front end. ~When you edit an article, the text field for the alternative text is not show for the `old` custom field at the begin. But when you delete the image temporarily the text field appear and you can use it.~
4. Use the new form field regardless of custom field
a) Add 
```
		<field
			name="image2"
			type="accessiblemedia"
			label="COM_CONTACT_FIELD_PARAMS_IMAGE_LABEL"
			directory="banners"
		/>

```
somewhere for example in line 276 of `/administrator/components/com_contact/forms/contact.xml`.
b) If you use my example add

		`<?php echo $this->form->renderField('image2'); ?>`

in `/administrator/components/com_contact/tmpl/contact/edit.php` line 55.
c) See that you can use the field.

PR #36593, opened 2022-01-07
[4.1] Support class attribute in media custom field

laoneo:
Pull Request for pr #27712.

### Summary of Changes
The accessible media form field doesn't support the class property. This pr fixes it.

### Testing Instructions
- Create a media custom field
- In the options of the field set the "Field Class" attribute to `testclass`
- Save the field
- Edit an article
- Inspect the `joomla-field-media` element of the custom field in the dev tools of the browser

No `testclass` is added to the element.

`testclass` is added to the element.


laoneo:
@astridx can you have a look here if all is correct?

alikon:
I have tested this item :white_check_mark: successfully on e4f4e2df48e48f0685814e2cc1131cb08d23b458<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36593">issues.joomla.org/tracker/joomla-cms/36593</a>.</sub>

HusiBT:
I have tested this item :white_check_mark: successfully on 87ab096ea4ed0f219e467fd1f0058782fe41d845<br /><br />Tested successfully.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36593">issues.joomla.org/tracker/joomla-cms/36593</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36593">issues.joomla.org/tracker/joomla-cms/36593</a>.</sub>

wilsonge:
This looks kinda weird because the class is already being set on the input element within the custom form field. Does it really make sense to have it in both places?

laoneo:
Did you actually test it? The class is not set, as it is not passed into the subform group. It is not rendered at all in any of the elements.

wilsonge:
Yes I understand that the FormField layout is clearly correct. I'm asking specifically about the JLayout change as the class is also used here https://github.com/joomla/joomla-cms/pull/36593/files#diff-ffc48fe23d772f6461955e77801148ec01945822ae6574126b6cece7f7bccd26R54 and you're now adding it again on the custom element.

laoneo:
The class is only added to the custom element and nowhere else. The class is delegated to the element from the Formfield but never used there as you can see here https://github.com/joomla/joomla-cms/blob/4.1-dev/layouts/joomla/form/field/media/accessiblemedia.php. So where should then the class being used in your opinion?

laoneo:
Made another test and indeed you are right, the class is used on the input. So it would be wrong to render it two times.

PR #36617, opened 2022-01-09
[4.0] Fix error in article model

richard67:
Pull Request for https://github.com/joomla/joomla-cms/commit/67215a7dd6648b4dcf3731603b57cc375389e276#r63168677 .

### Summary of Changes

Fix error from the above commit.

I have no idea yet if we have more such errors in the 4.0-dev branch.

### Testing Instructions

Code review.

Wrong condition. On PHP 8.1+ it will raise an error due to trim with null argument, and (regardless of PHP version) the read more will be shown when the full text is null.

Right condition. The read more and the full text are only added if the full text is not null and the trimmed full text is not an empty string.

None.


wilsonge:
Oops thanks!

PR #36568, opened 2022-01-05
[4.1] Fix dependency on web application stopping content from booting in CLI mode

wilsonge:
Pull Request for Issue #35818 .

### Summary of Changes
Removes unused constructor argument. Luckily also fixing #35818 for now although there's still some thought process to work through there.

### Testing Instructions
Check loading of icons in the frontend of com_content and com_contact continues to work (both create and edit icons)

### Documentation Changes Required
None

Issue #35818, opened 2021-10-14
ContentComponent boots only on SiteApplication 


anibalsanchez:
### Steps to reproduce the issue

I'm writing a console command that creates articles on Joomla 4. It used to work on Joomla 3 with some hackery, but now I'm trying to write the native Joomla 4 implementation. However, the Content Component method `boot` forces the load of the SiteApplication/WebApplication in the CLI context. This is the offending line that loads the Icon and the whole SiteApplication:

```php
	public function boot(ContainerInterface $container)
	{
...
		$this->getRegistry()->register('contenticon', new Icon($container->get(SiteApplication::class)));
...
	}
```

### Expected result

ContentComponent running on the terminal.

### Actual result

ContentComponent is forcing the load of the WebApplication 

### System information (as much as possible)

PHP 7.4/Joomla 4.0.3

anibalsanchez:
The ContactComponent also has a similar Icon definition.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35818">issues.joomla.org/tracker/joomla-cms/35818</a>.</sub>

anibalsanchez:
As a workaround to save the day:

```php
        $_SERVER['HTTP_HOST'] = 'example.com';
        $_SERVER['SCRIPT_NAME'] = '/not-running-as-a-command-line.html';
```
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35818">issues.joomla.org/tracker/joomla-cms/35818</a>.</sub>

anibalsanchez:
It looks like the Icon definition for the Web should be removed from the `boot` method.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/35818">issues.joomla.org/tracker/joomla-cms/35818</a>.</sub>

alikon:
could you have a look at #35722 ?

anibalsanchez:
From my comment in #35722:

_About #35625, the problem is that the `php cli/joomla.php list` is firing the "System - Page Cache" when it's called from the console. Why would a console application fire the "System - Page Cache" plugin? It shouldn't. By default, a console application doesn't have to interact with the site cache._

[...]

PR #36633, opened 2022-01-10
[4][CLI] boot com_content from console

alikon:
Create a console plugin that boot the com_content  component and run it from terminal

ie. something like 
```php
$content = Factory::getApplication()->bootComponent('com_content')->getMVCFactory();
$model = $content->createModel('Article', 'Administrator',['ignore_request' => true]);
$result = $model->save($article);
```

### Summary of Changes

- ~~added the `getUserState()` method to the ConsoleApplication.php~~
- don't check session from cli
- early return from Webauthn plugin  when from console

### Testing Instructions
- download this silly example  [plg_console_lorem.zip](https://github.com/joomla/joomla-cms/files/7839093/plg_console_lorem.zip)  wich fetch a lorem ipsum from https://loripsum.net and create an article
- the plugin source code is available here https://github.com/alikon/testcom/tree/main/plugins/console/lorem
- install & enable it & set parameters (like category and user)
- run from terminal  `php cli/joomla.php lorem:ipsum`

image

no error in console and article created

### References
#36568, #36332, #35818 

cc: @wilsonge  @laoneo  @anibalsanchez


wilsonge:
Ok so webauthn looks fine. But where else is the user state being used - i don't particularly like having to implement that in the consoleapplication class

alikon:
the main point is we want  the console app like 1st citizen ?
we want the console application able to re-use existing CMS code or not ?

i've shared the plugin console code already  maybe i've made something wrong there ??
https://github.com/alikon/testcom/blob/main/plugins/console/lorem/src/Extension/LoremIpsumCommand.php

laoneo:
If you have `'ignore_request' => true` then the populateState function should not be called. On other places the userstate should not be called.

wilsonge:
I don't have the ability to test right now - I'll give it a test when I get home but @laoneo is right about the theory of how this should work

alikon:
as you can see i already have set  `['ignore_request' => true]`

```php
$content = Factory::getApplication()->bootComponent('com_content')->getMVCFactory();
$model = $content->createModel('Article', 'Administrator',['ignore_request' => true]);
$result = $model->save($article);
```

laoneo:
And where is the user state requested?

alikon:
even then the `setUserState()` is not a method of ConsoleApplication  what i'm missing ?

wilsonge (approved):

alikon:
let's see what kind of thumb i'll collect  cc: @SharkyKZ

alikon:
closed by mistake i suppose

PR #30727, opened 2020-09-22
[4.1] mod_breadcrumbs - convert microdata to JSON-LD

hans2103:
Pull Request for Issue # .

### Summary of Changes

With https://github.com/joomla/joomla-cms/pull/25117 in mind and the possible adaptation of Spatie\SchemaOrg in mind I would like to prepare current breadcrumbs by converting the Microdata into JSON-LD structure. 
But not only with that possible adaptation in mind... The JSON-LD format is the recommended format. 
The HTML code will be cleaner when the data is moved to JSON-LD format
Another benefit of this change is that the current page will always be added to the data. This is not the case in the current situation with _Hide Last Item_ is active. 

### Testing Instructions

- Install Joomla with Sample Data Blog. 
- Note that there will be a  mod_breadcrumbs on every page. example: http://**your-joomla-4-site**/index.php/blog/article
- Inspect element and see the markup of mod_breadcrumbs
- Apply this patch. No `npm run something:something` is required
- Refresh the page and note the changed markup.

### Actual result BEFORE applying this Pull Request

```
<nav role="navigation" aria-label="Breadcrumbs">
    <ol itemscope="" itemtype="https://schema.org/BreadcrumbList" class="mod-breadcrumbs breadcrumb">
        <li class="mod-breadcrumbs__here float-left">You are here: &nbsp;</li>
        <li itemprop="itemListElement" itemscope="" itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item">
            <a itemprop="item" href="/index.php" class="pathway"><span itemprop="name">Home</span></a>
            <meta itemprop="position" content="1">
        </li>
        <li itemprop="itemListElement" itemscope="" itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item">
            <a itemprop="item" href="/index.php/blog" class="pathway"><span itemprop="name">Blog</span></a>
            <meta itemprop="position" content="2">
        </li>
        <li aria-current="page" itemprop="itemListElement" itemscope="" itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item active">
            <span itemprop="name">Article</span>
            <meta itemprop="position" content="3">
        </li>
    </ol>
</nav>

```

### Expected result AFTER applying this Pull Request

```
<nav role="navigation" aria-label="Breadcrumbs">
    <ol class="mod-breadcrumbs breadcrumb">
[...]

PR #36665, opened 2022-01-13
[4.0] Fix Google Structured Data Errors in Breadcrumbs

crystalenka:
### Summary of Changes
This PR fixes structured data errors that made Google flag breadcrumbs as invalid.

When a breadcrumb item was a menu heading, instead of a link, Google would flag it as an invalid breadcrumb because it did not have a link. This is fixed by removing the structured data markup from items without a link.

### Testing Instructions
Create a site menu with a menu heading and subpages, like so:
- Home
- Menu Heading
  - Subpage 1

Create a breadcrumbs module. Ensure the module is published and visible on Subpage 1.

Navigate to subpage 1. 
#### If this site is accessible over the internet:
Copy the URL and paste it into the field here: https://search.google.com/test/rich-results

#### If this site is hosted locally:
View source of Subpage 1 and copy it into the code box on https://search.google.com/test/rich-results
(Alternatively, you can find the generated code for the ul and just copy that into the code box.)

Google detects the breadcrumbs but shows an error: 
Screenshot showing Google Rich Search results error details

When you click on it to see more, this is the problem:
Screenshot showing Google Rich Search results error details

Google does not consider a breadcrumb item as valid unless it has a link. Since menu headings don't have links, we have to remove the structured data from that item.

Google detects the breadcrumbs and marks them as valid:
Screenshot showing Google Rich Search results with valid notification

None as far as I'm aware.

### DIFF
--- a/modules/mod_breadcrumbs/tmpl/default.php
+++ b/modules/mod_breadcrumbs/tmpl/default.php
@@ -57,7 +57,7 @@ use Joomla\CMS\Router\Route;
 
 				<?php 
 				// Render items without links without structured data so Google doesn't flag it as invalid breadcrumbs.
-				else : 
+				else :
 					$breadcrumbItem = '<span>' . $item->name . '</span>'; ?>
 					<li class="mod-breadcrumbs__item breadcrumb-item">
 						<?php echo $breadcrumbItem; ?>
--- a/modules/mod_breadcrumbs/tmpl/default.php
+++ b/modules/mod_breadcrumbs/tmpl/default.php
@@ -48,7 +48,7 @@ use Joomla\CMS\Router\Route;
 		foreach ($list as $key => $item) :
 			if ($key !== $last_item_key) :
 				// Render links with structured data
-				if (!empty($item->link)) : 
+				if (!empty($item->link)) :
 					$breadcrumbItem = '<a itemprop="item" href="' . Route::_($item->link) . '" class="pathway"><span itemprop="name">' . html_entity_decode($item->name, ENT_QUOTES, 'UTF-8') . '</span></a>'; ?>
 					<li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item">
 						<?php echo $breadcrumbItem; ?>
--- a/modules/mod_breadcrumbs/tmpl/default.php
+++ b/modules/mod_breadcrumbs/tmpl/default.php
@@ -47,16 +47,23 @@ use Joomla\CMS\Router\Route;
 		// Generate the trail
 		foreach ($list as $key => $item) :
 			if ($key !== $last_item_key) :
-				if (!empty($item->link)) :
-					$breadcrumbItem = '<a itemprop="item" href="' . Route::_($item->link) . '" class="pathway"><span itemprop="name">' . html_entity_decode($item->name, ENT_QUOTES, 'UTF-8') . '</span></a>';
-				else :
-					$breadcrumbItem = '<span itemprop="name">' . $item->name . '</span>';
-				endif;
-				// Render all but last item - along with separator ?>
-				<li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item"><?php echo $breadcrumbItem; ?>
-					<meta itemprop="position" content="<?php echo $key + 1; ?>">
-				</li>
-			<?php elseif ($show_last) :
+				// Render links with structured data
+				if (!empty($item->link)) : 
+					$breadcrumbItem = '<a itemprop="item" href="' . Route::_($item->link) . '" class="pathway"><span itemprop="name">' . html_entity_decode($item->name, ENT_QUOTES, 'UTF-8') . '</span></a>'; ?>
+					<li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item">
+						<?php echo $breadcrumbItem; ?>
+						<meta itemprop="position" content="<?php echo $key + 1; ?>">
+					</li>
+
+				<?php 
+				// Render items without links without structured data so Google doesn't flag it as invalid breadcrumbs.
+				else : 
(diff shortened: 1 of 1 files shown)


crystalenka:
This may need to be fixed in 4.1 as well with the JSON microdata based on https://github.com/joomla/joomla-cms/pull/30727

brianteeman:
1. Pretty sure the changes here are not correct as they remove the special handling for the last item (might need more coffee before checking 100%)
2. with the JSON microdata it is valid to have the following. Note the item"" for the non-linked item
```

{
    "@context": "https://schema.org",
    "@type": "BreadcrumbList",
    "itemListElement": [
        {
            "@type": "ListItem",
            "position": 1,
            "name": "Home",
            "item": "https://www.example.com"
        },
        {
            "@type": "ListItem",
            "position": 2,
            "name": "Category",
            "item": ""
        },
        {
            "@type": "ListItem",
            "position": 3,
            "name": "item",
            "item": "https://www.example.com/category/page2"
        }
    ]
}
```

crystalenka:
I did not remove the special handling for the last item. It's untouched.

Also, Google does not accept blank urls in 'item'; it marks it as invalid. I tested your code with the link above, this is what it Google says:

Screen Shot 2022-01-13 at 10 34 03

brianteeman:
well that's annoying because https://validator.schema.org/ says its perfectly valid

crystalenka:
I know, Google seems to be applying their own standard for structured data as it applies to rich search results.

brianteeman:
```
FILE: /********/src/modules/mod_breadcrumbs/tmpl/default.php
FOUND 3 ERRORS AFFECTING 3 LINES
 51 | ERROR | [x] Whitespace found at end of line
    |       |     (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)
 58 | ERROR | [x] Whitespace found at end of line
    |       |     (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)
 60 | ERROR | [x] Whitespace found at end of line
    |       |     (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)
```

modules/mod_breadcrumbs/tmpl/default.php
```diff
@@ -47,16 +47,23 @@
...
-				else :
-					$breadcrumbItem = '<span itemprop="name">' . $item->name . '</span>';
-				endif;
-				// Render all but last item - along with separator ?>
-				<li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item"><?php echo $breadcrumbItem; ?>
-					<meta itemprop="position" content="<?php echo $key + 1; ?>">
-				</li>
-			<?php elseif ($show_last) :
+				// Render links with structured data
+				if (!empty($item->link)) :
```
richard67:
```suggestion
				if (!empty($item->link)) :
```

modules/mod_breadcrumbs/tmpl/default.php
```diff
@@ -47,16 +47,23 @@
...
-			<?php elseif ($show_last) :
+				// Render links with structured data
+				if (!empty($item->link)) : 
+					$breadcrumbItem = '<a itemprop="item" href="' . Route::_($item->link) . '" class="pathway"><span itemprop="name">' . html_entity_decode($item->name, ENT_QUOTES, 'UTF-8') . '</span></a>'; ?>
+					<li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item">
+						<?php echo $breadcrumbItem; ?>
+						<meta itemprop="position" content="<?php echo $key + 1; ?>">
+					</li>
+
+				<?php
```
richard67:
```suggestion
				<?php
```

modules/mod_breadcrumbs/tmpl/default.php
```diff
@@ -47,16 +47,23 @@
...
+				if (!empty($item->link)) : 
+					$breadcrumbItem = '<a itemprop="item" href="' . Route::_($item->link) . '" class="pathway"><span itemprop="name">' . html_entity_decode($item->name, ENT_QUOTES, 'UTF-8') . '</span></a>'; ?>
+					<li itemprop="itemListElement" itemscope itemtype="https://schema.org/ListItem" class="mod-breadcrumbs__item breadcrumb-item">
+						<?php echo $breadcrumbItem; ?>
+						<meta itemprop="position" content="<?php echo $key + 1; ?>">
+					</li>
+
+				<?php 
+				// Render items without links without structured data so Google doesn't flag it as invalid breadcrumbs.
+				else :
```
richard67:
```suggestion
				else :
```

richard67:
@crystalenka I've made review suggestions for the code style error which @brianteeman has mentioned in his previous comment. You can add them to a batch of suggestions and apply that with one click, or you can apply each of the suggestions separately with a direct click. I thought I do that to make it more comfortable for you.

crystalenka:
> @crystalenka I've made review suggestions for the code style error which @brianteeman has mentioned in his previous comment. You can add them to a batch of suggestions and apply that with one click, or you can apply each of the suggestions separately with a direct click. I thought I do that to make it more comfortable for you.

Thank you!!

richard67:
@brianteeman Regarding the last item: It is not touched by this PR, and the if condition before that elseif is also not touched, so there should not be any change for that. If my understanding is right about how breadcrumbs and how menu items of type "Heading" or "Separator" work (which both are the types which have no link), it cannot happen that the last item of breadcrumbs belongs to a menu item of these types because to such menu items there is no navigation. So the PR should be ok as it is, no need to add the same logic changes for the last item. Or am I missing something?

brianteeman:
I can confirm that before this PR google structure data test reports an error and after this pr it does not report an error.

I note however that the code for joomla 4.1 in the breadcrumbs is completely different

brianteeman:
I guess this needs to be closed and remade for 4.1 - the code in 4.1 for the breadcrumbs structured data is completely different

crystalenka:
Yeah, I will be taking a look at 4.1 to see if I can fix it there too but hadn't had a chance yet. I'll just close this one.

PR #36673, opened 2022-01-13
[4.0] Return value type is not compatible with declared

PhilETaylor:
Code review

Return value type is not compatible with declared

### DIFF
--- a/administrator/components/com_menus/src/Field/MenuOrderingField.php
+++ b/administrator/components/com_menus/src/Field/MenuOrderingField.php
@@ -48,7 +48,7 @@ class MenuOrderingField extends ListField
 
 		if (!$parent_id)
 		{
-			return false;
+			return [];
 		}
 
 		$db = Factory::getDbo();
--- a/administrator/components/com_menus/src/Field/MenuOrderingField.php
+++ b/administrator/components/com_menus/src/Field/MenuOrderingField.php
@@ -35,7 +35,7 @@ class MenuOrderingField extends ListField
 	 * Method to get the list of siblings in a menu.
 	 * The method requires that parent be set.
 	 *
-	 * @return  array  The field option objects or false if the parent field has not been set
+	 * @return  array|boolean  The field option objects or false if the parent field has not been set
 	 *
 	 * @since   1.7
 	 */
@@ -48,7 +48,7 @@ class MenuOrderingField extends ListField
 
 		if (!$parent_id)
 		{
-			return [];
+			return false;
 		}
 
 		$db = Factory::getDbo();
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -136,7 +136,7 @@ class MenusHelper extends ContentHelper
 	 * @param   array    $languages  Optional array of specify which languages we want to filter
 	 * @param   int      $clientId   Optional client id - viz 0 = site, 1 = administrator, can be NULL for all (used only if menutype not given)
 	 *
-	 * @return  array
+	 * @return  array|bool
 	 *
 	 * @since   1.6
 	 */
--- a/administrator/components/com_menus/src/Helper/MenusHelper.php
+++ b/administrator/components/com_menus/src/Helper/MenusHelper.php
@@ -136,7 +136,7 @@ class MenusHelper extends ContentHelper
 	 * @param   array    $languages  Optional array of specify which languages we want to filter
 	 * @param   int      $clientId   Optional client id - viz 0 = site, 1 = administrator, can be NULL for all (used only if menutype not given)
 	 *
-	 * @return  array|bool
+	 * @return  array|boolean
 	 *
 	 * @since   1.6
 	 */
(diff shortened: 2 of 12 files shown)


administrator/components/com_menus/src/Helper/MenusHelper.php
```diff
@@ -136,7 +136,7 @@ public static function getMenuTypes($clientId = 0)
 	 * @param   array    $languages  Optional array of specify which languages we want to filter
 	 * @param   int      $clientId   Optional client id - viz 0 = site, 1 = administrator, can be NULL for all (used only if menutype not given)
 	 *
-	 * @return  array
+	 * @return  array|bool
```
sandewt: To be consistent: return a `boolean`

 * @return  array|boolean

See also the other changes :wink:

PhilETaylor: That is not consistent and against every change I have ever made. There is no need to use the extended version, bool is correct

PhilETaylor: Hmmm it seems the outdated, ancient Joomla Sniffs ONLY allow `boolean` and not `bool` and therefore I need to change this so that it passes the phpcs tests. 

I'll PR the code standards as that is clearly wrong, and `bool` is a perfectly valid, and preferred type

administrator/components/com_menus/src/Field/MenuOrderingField.php
```diff
@@ -48,7 +48,7 @@ protected function getOptions()
 
 		if (!$parent_id)
 		{
-			return false;
+			return [];
```
Quy: Please update line 38.

PhilETaylor: ta

sandewt:
https://github.com/joomla/joomla-cms/blob/51d953156964ce995a13cc4adc42642a5cc833b2/administrator/components/com_associations/src/Helper/AssociationsHelper.php#L179

should be ```* @return  string```

PhilETaylor:
This is not an exhaustive PR, Joomla is full of these incorrect types! It’s going to take years to correct them all

sandewt:
Obviously, I already suspected that. 
Great that you have made a start. :thumbsup:

components/com_users/src/Controller/ProfileController.php
```diff
@@ -72,7 +72,7 @@ public function edit()
 	/**
 	 * Method to save a user's profile data.
 	 *
-	 * @return  void
+	 * @return  void|boolean
```
sandewt, line 75: * @return  void|boolean
 should be ???
 * @return  boolean 

Note: "Return type in PHPDoc does not match the actual return type" - PhpStorm

[EDIT] What is common?

libraries/src/Document/HtmlDocument.php
```diff
@@ -326,7 +326,7 @@ public function setHeadData($data)
 	 *
 	 * @param   array  $data  The document head data in array form
 	 *
-	 * @return  HtmlDocument|null instance of $this to allow chaining or null for empty input data
+	 * @return  HtmlDocument|void instance of $this to allow chaining or void for empty input data
```
sandewt, line 329: View the previous comment

components/com_users/src/Controller/ProfileController.php
PhilETaylor: void|boolean is correct, if the code gets to the end of the method then it will return void, it will only return false if there is a problem before the end of the method.

libraries/src/Document/HtmlDocument.php
PhilETaylor: `  HtmlDocument|void ` is correct here.. I dont understand what point you are trying to make.

PhilETaylor:
>should be * @return string 

Done.

libraries/src/Document/HtmlDocument.php
sandewt: I mean 'void'

sandewt:
I have tested this item :white_check_mark: successfully on c0ce5e9329111c2b88c4ce1292f892236fb84da6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36673">issues.joomla.org/tracker/joomla-cms/36673</a>.</sub>

Quy:
I have tested this item :white_check_mark: successfully on c0ce5e9329111c2b88c4ce1292f892236fb84da6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36673">issues.joomla.org/tracker/joomla-cms/36673</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36673">issues.joomla.org/tracker/joomla-cms/36673</a>.</sub>

PhilETaylor:
Thanks

pritam825:
I have tested this item :white_check_mark: successfully on c0ce5e9329111c2b88c4ce1292f892236fb84da6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36673">issues.joomla.org/tracker/joomla-cms/36673</a>.</sub>

bembelimen:
Thx

PR #36629, opened 2022-01-10
[4.0] Access the getRouter function statically

laoneo:
Pull Request for pr #36562.

### Summary of Changes
This is a result of the discussion in #36562 to keep static function `getRouter` in 4 and make it none static in 5. So in 4 all calls should be on the static function.

### Testing Instructions
- Enable the language filter plugin
- Enable preview in the global template options
- Click on the preview link in the back end template styles list for the site

### Actual result BEFORE applying this Pull Request
Works

### Expected result AFTER applying this Pull Request
Works

PR #36688, opened 2022-01-14
[4.2] Deprecate getting the router by the app and introduce a service

laoneo:
Pull Request for pr's #36562, #36629, #36631.

### Summary of Changes
As discussed in #36631. It would be better to get the router from a service. Additionally the `getRouter` function in applications do become deprecated.

A second step would be to turn the language plugin into a service provider plugin and inject the router there. The same goes for the finder extension and the applications.

ping @nibra and @wilsonge 

### Testing Instructions
- Create an article
- Create a blog menu item
- Enable the language filter plugin
- Open the front end

All is working.

All is working.


components/com_finder/src/View/Search/HtmlView.php
```diff
@@ -160,7 +161,7 @@ public function display($tpl = null)
 		if (strpos($this->query->input, '"'))
 		{
 			// Get the application router.
-			$router = $app->getRouter();
+			$router = Factory::getContainer()->get(SiteRouter::class);;
```
PhilETaylor:
```suggestion
			$router = Factory::getContainer()->get(SiteRouter::class);
```

libraries/src/Service/Provider/Router.php
```diff
@@ -32,12 +36,32 @@ class ApiRouter implements ServiceProviderInterface
...
+				},
+				true
+			);
+
+		$container->alias('AdministratorRouter', AdministratorRouter::class)
+			->share(
+				AdministratorRouter::class,
+				function (Container $container)
+				{
+					return new AdministratorRouter();
```
PhilETaylor:
```suggestion
					return new AdministratorRouter;
```

libraries/src/Router/Router.php
```diff
@@ -94,8 +94,9 @@ class Router
 	 *
 	 * @return  Router  A Router object.
 	 *
-	 * @since   1.5
-	 * @throws  \RuntimeException
+	 * @since      1.5
+	 * @throws     \RuntimeException
+	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```
PhilETaylor:
```suggestion
	 * @since      1.5
	 *
	 * @throws     \RuntimeException
	 *
	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```

libraries/src/Application/SiteApplication.php
```diff
@@ -378,7 +377,8 @@ public function getPathway($name = 'site', $options = array())
 	 *
 	 * @return	\Joomla\CMS\Router\Router
 	 *
-	 * @since	3.2
+	 * @since      3.2
+	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```
PhilETaylor:
```suggestion
	 * @since      3.2
	 *
	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```

libraries/src/Application/ConsoleApplication.php
```diff
@@ -446,8 +446,9 @@ public function setName(string $name): void
 	 *
 	 * @return  Router
 	 *
-	 * @since   4.0.6
-	 * @throws  \InvalidArgumentException
+	 * @since      4.0.6
+	 * @throws     \InvalidArgumentException
+	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```
PhilETaylor:
```suggestion
	 * @since      4.0.6
	 *
	 * @throws     \InvalidArgumentException
	 *
	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```

libraries/src/Application/CMSWebApplicationInterface.php
```diff
@@ -51,7 +51,8 @@ public function getMenu($name = null, $options = array());
 	 *
 	 * @return  Router
 	 *
-	 * @since   4.0.0
+	 * @since      4.0.0
+	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```
PhilETaylor:
```suggestion
	 * @since      4.0.0
	 *
	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```

libraries/src/Application/CMSApplication.php
```diff
@@ -595,7 +596,8 @@ public function getPathway()
 	 *
 	 * @return  Router
 	 *
-	 * @since   3.2
+	 * @since      3.2
+	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```
PhilETaylor:
```suggestion
	 * @since      3.2
	 *
	 * @deprecated 5.0 Inject the router or load it from the dependency injection container
```

Fedik:
I would not deprecate it, Web APP still should have it, like it have a Document.
Just add ServiceProvider, and if someone need a specific Router he will pick it from there (like in `Route::link()`).

Another thing, introducing SiteRouterInterface looks a bit not-intuitive to me, what purpose of it?
I would expect then AdministratorRouterInterface, ApiRouterInterface, CliRouterInterface etc

joeforjoomla:
> I would not deprecate it, Web APP still should have it, like it have a Document. Just add ServiceProvider, and if someone need a specific Router he will pick it from there (like in `Route::link()`).
> 
I agree. Just like Document, Session, etc for consistency it should not be deprecated. Otherwise everything should be deprecated. App still should have everything as a shortcut to the container.

laoneo:
The app doesn't have a router instance, all it does it forwards the request to Router::getInstance. There is no reason to have a delegating function in the app. So all of this should be deprecated and the router should be injected wherever possible. On a next step the `Route` class should be deprecated and routing should actually go over the router only which is the available in all views.

I Made only the SiteRouter interface for now as this is what actually mathers if we need other's then we can do them on demand.

libraries/src/Application/SiteApplication.php
```diff
@@ -818,8 +821,64 @@ protected function render()
...
+		// It is not possible to inject the SiteRouter as it requires a SiteApplication
+		// and we would end in an infinite loop
+		$result = $this->getContainer()->get(SiteRouter::class)->parse($uri, true);
+
+		$active = $this->getMenu()->getActive();
+
+		if ($active !== null
+			&& $active->type === 'alias'
+			&& $active->getParams()->get('alias_redirect')
+			&& \in_array($this->input->getMethod(), array('GET', 'HEAD'), true))
```
PhilETaylor:
```suggestion
			&& \in_array($this->input->getMethod(), ['GET', 'HEAD'], true))
```

libraries/src/Router/SiteRouterAwareTrait.php
```diff
@@ -0,0 +1,57 @@
...
+	private $router;
+
+	/**
+	 * Get the site router.
+	 *
+	 * @return  SiteRouter
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  \UnexpectedValueException May be thrown if the router has not been set.
+	 */
```
PhilETaylor:
```suggestion
	/**
	 * Get the site router.
	 *
	 * @return  SiteRouter
	 *
	 * @since   __DEPLOY_VERSION__
	 *
	 * @throws  \UnexpectedValueException May be thrown if the router has not been set.
	 */
```

laoneo:
> 
> I agree. Just like Document, Session, etc for consistency it should not be deprecated. Otherwise everything should be deprecated. App still should have everything as a shortcut to the container.

The app should not be a god object with tons of references. Modern apps inject the stuff which are needed. Especially for the router where the router has a reference to the app but not vice versa. And the container should not be abused as a service locator in the application. This is an anti pattern.

HusiBT:
I have tested this item :white_check_mark: successfully on 51cfca41262740158e94e5064354ebd00858d540<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36688">issues.joomla.org/tracker/joomla-cms/36688</a>.</sub>

dgrammatiko:
I have tested this item :white_check_mark: successfully on f6908c0b14ab1eb73354dd201da6d496e6187d87<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36688">issues.joomla.org/tracker/joomla-cms/36688</a>.</sub>

laoneo:
rtc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36688">issues.joomla.org/tracker/joomla-cms/36688</a>.</sub>

roland-d:
@laoneo This requires another test after the new commit.

@BPBlueprint Could you redo your test again?

laoneo:
@roland-d all I did was adding a try/catch block around, nothing more https://github.com/joomla/joomla-cms/pull/36688/commits/f6908c0b14ab1eb73354dd201da6d496e6187d87

roland-d:
Thanks everybody.

joeforjoomla:
Pinging @laoneo 

Actually this PR is not fully b/c and reliable, given than:

`Factory::getContainer()->get('SiteRouter'):`

is not the same than:

` $this->app::getRouter('Site');`

I've noticed that they are 2 separate and different instances of the SiteRouter class.

So given for example that the Joomla core uses the container: Factory::getContainer()->get('SiteRouter'):
but a system plugin still uses $this->app::getRouter('Site');  to attach some build/parse rules:

```
$router->attachParseRule ( array (
		$this,
		'preProcessParseRule'
), \Joomla\CMS\Router\Router::PROCESS_BEFORE );
```
The result will be that the preProcessParseRule function of this plugin will never be called.

This implication has potential drawbacks, not sure how it can be solved other than update the code of all extensions to the new container to work on the same instance of the router class.

laoneo:
There is an alias missing for the old class names. Can you give #37381 test?

Issue #36674, opened 2022-01-14
[4.0]Inconsistency in MediaHelper::canUpload


toivo:
### Steps to reproduce the issue
Code review of libraries/src/Helper/MediaHelper.php lines 243 - 252

```
		// Media file names should never have executable extensions buried in them.
		$executables = array_merge(self::EXECUTABLES, InputFilter::FORBIDDEN_FILE_EXTENSIONS);

		// Remove allowed executables from array
		if (count($allowedExecutables))
		{
			$executable = array_diff($executables, $allowedExecutables);
		}

		$check = array_intersect($filetypes, $executables);
```
Line 249 should be:
```
			$executables = array_diff($executables, $allowedExecutables);
```
### Additional comments
Compare the above to MediaHelper::checkFileExtension in lines 177 to 183:
```
		// Remove allowed executables from array
		if (count($allowedExecutables))
		{
			$executables = array_diff($executables, $allowedExecutables);
		}

		if (in_array($extension, $executables, true))
```

richard67:
> Line 249 should be:
> `$executables = array_diff($executables, $allowedExecutables);`

From reading code I'd say you are right.

akshitrattan:
after going through the entire file I too think that line 249 should be changed
However, when I upload a restricted file type( E.g.- I tried uploading an SVG image, which is a restricted file type), considering this code snippet, I should not get an error, but I still get an error that I can't upload the file type
Can anyone explain to me why this happens?

brianteeman:
@akshitrattan restricted file type and executable files are completely different things

akshitrattan:
> @akshitrattan restricted file type and executable files are completely different things

oh, so as in a file like .exe??

brianteeman:
https://github.com/joomla/joomla-cms/blob/65aaacc28cddfed55804306eb02b407d28884159/libraries/src/Helper/MediaHelper.php#L28-L40

akshitrattan:
> https://github.com/joomla/joomla-cms/blob/65aaacc28cddfed55804306eb02b407d28884159/libraries/src/Helper/MediaHelper.php#L28-L40

thanks a lot, now I'm familiar with the exact meaning of the term

richard67:
Who wants to make a pull request to fix that? I think the main challenge is to find good testing instructions for a real human test. As far as I can see there are no unit tests for the media helper yet, so I think it needs some kind of human test to make sure nothing is broken.

akshitrattan:
[...]

PR #36695, opened 2022-01-15
Updated MediaHelper::canUpload function

akshitrattan:
Pull Request for Issue #36674.

### Summary of Changes
In libraries/src/Helper/MediaHelper
changed 
`if (count($allowedExecutables))
		{
			$executable = array_diff($executables, $allowedExecutables);
		}`

to
`if (count($allowedExecutables))
		{
			$executables = array_diff($executables, $allowedExecutables);
		}`
to allow only some permitted executable file type to be uploaded in the media tab
### Testing Instructions
In the media tab, try uploading executable file types that are not permitted by the cms
for reference,
`/**
	 * A special list of blocked executable extensions, skipping executables that are
	 * typically executable in the webserver context as those are fetched from
	 * Joomla\CMS\Filter\InputFilter
	 *
	 * @var    string[]
	 * @since  4.0.0
	 */
	public const EXECUTABLES = array(
		'js', 'exe', 'dll', 'go', 'ade', 'adp', 'bat', 'chm', 'cmd', 'com', 'cpl', 'hta',
		'ins', 'isp', 'jse', 'lib', 'mde', 'msc', 'msp', 'mst', 'pif', 'scr', 'sct', 'shb',
		'sys', 'vb', 'vbe', 'vbs', 'vxd', 'wsc', 'wsf', 'wsh', 'html', 'htm', 'msi'
	);`
if you are unable to execute/run the files, the test is successful.

No major Changes

### DIFF
--- a/libraries/src/Helper/MediaHelper.php
+++ b/libraries/src/Helper/MediaHelper.php
@@ -246,7 +246,7 @@ class MediaHelper
 		// Remove allowed executables from array
 		if (count($allowedExecutables))
 		{
-			$executable = array_diff($executables, $allowedExecutables);
+			$executables = array_diff($executables, $allowedExecutables);
 		}
 
 		$check = array_intersect($filetypes, $executables);


richard67:
@bembelimen The error exists already in the 4.0-dev branch. Shall we go on with this PR here for 4.1-dev, or shall it be fixed in 4.0-dev and then be merged up into 4.10-dev?

brianteeman:
> if you are unable to do so, the test is successful.

thats not a valid test of the function

akshitrattan:
> > if you are unable to do so, the test is successful.
> 
> thats not a valid test of the function

I editted the comment, could you cross-check??

richard67:
> > > if you are unable to do so, the test is successful.
> > 
> > 
> > thats not a valid test of the function
> 
> I editted the comment, could you cross-check??

@akshitrattan Now it's even more wrong because it is not about executing the uploaded files.

akshitrattan:
> > > > if you are unable to do so, the test is successful.
> > > 
> > > 
> > > thats not a valid test of the function
> > 
> > 
> > I editted the comment, could you cross-check??
> 
> @akshitrattan Now it's even more wrong because it is not about executing the uploaded files.

I'm so sorry, could you pls help me in editting it??

brianteeman:
@richard67 this really should not be discussed on the public tracker. If its a valid issue (I'm not sure it is) then it absolutely must be handled by the security team.

brianteeman:
I could explain more about this but the public tracker is not the place. The code comments should explain enough.

richard67:
@brianteeman Agree, we are already discussing.

richard67:
@brianteeman Regarding security we are safe here because the error makes the allowed executables feature fail so that they are still not allowed.

brianteeman:
1. I can't find any values for  $allowedExecutables
2. From the original PR I would have thought it would have been $allowedExtensions
3. Have you actually verified that a bug even exists

richard67:
> 1. I can't find any values for $allowedExecutables

The `$allowedExecutables` can be passed as optional parameters to the `checkFileExtension` and the `canUpload` functions. Obviously in the CMS core that's not used yet, and unfortunately it seems that this parameter has been forgotten in the doc block of these functions.

> 2. From the original PR I would have thought it would have been $allowedExtensions

@brianteeman No, the $allowedExtensions is something else, and if we do an array diff with those, we will make the check for the executables break.

> 3. Have you actually verified that a bug even exists

Just check the code of the complete file. There is nowhere any use of the `$executable` (singular) variable, so it MUST be a typo.

And compare the usage of the parameter in the 2 mentioned functions.

For a test one could add some PHP code with a call to the `canUpload` function to the index.php of Atum or Cassiopeia with the parameter being used, and verify the results without and with the patch.

wilsonge:
Because security spent some time figuring out what was going on here (and determined it to not be an issue). I've chosen to directly merge this on review for the RC anyhow - because it's clearly a correct typo fix I've agreed with them to merge this for 4.0.6 as I was building the final packages as we speak.

So this has been merged to 4.0-dev with https://github.com/joomla/joomla-cms/commit/4e4f59673dbd7166318e00bef3b7c2687f810bba thankyou @akshitrattan !

brianteeman:
Thank you

toivo:
Thank you everyone.

Issue #36636, opened 2022-01-10
TextField use XMLElement for layout.


korenevskiy:
Each field must be created using XML or using PHP. 
But this text field cannot be created using PHP.
Examples:
1.`<field name="user" type="text"/>`
2. 
```
$textfield = Joomla\CMS\Form\FormHelper::loadFieldType('text');
$textfield->name = 'user';
$textfield->renderField();
```
This not working.
instead of this method, you can use another method 
`$textfield->render($textfield->renderLayout,[]);`
It doesn't work either.
These methods call the text field method `$textfield->getLayoutData()`
Inside this method there is a method call `$textfield->getOptions()`
And inside this method, data is collected from `$textfield->element` which has the type SimpleXMLElement.
This functionality should be called in the `$textfield->Setup()` method. 
But it is not called there, but is called when called ->renderField(); and ->render().
Thus, it is not possible to create fields in PHP.
[		foreach ($this->element->children() as $option)](https://github.com/joomla/joomla-cms/blob/50685f9147ed458be421e544591444e27ca4613b/libraries/src/Form/Field/TextField.php#L258)

Of course, this can be solved with crutches and props by creating a separate pseudo element with the SimpleXMLElement type. or by creating a Layout object separately from the text field object. But this is not normal at all. This is NOT observed in other fields. Moreover, the text field is the base class for many other fields. It is necessary to transfer the call `$textfield->getOptions()` to the method `$textfield->Setup()`

richard67:
Closing as there is a pull request.

PR #36712, opened 2022-01-16
Text field without xml element for layout rendering

korenevskiy:
Each field must be created using XML or using PHP. 
But this text field cannot be created using PHP.
Examples:
1.`<field name="user" type="text"/>`
2. 
```
$textfield = Joomla\CMS\Form\FormHelper::loadFieldType('text');
$textfield->name = 'user';
$textfield->renderField();
```
This not working.
instead of this method, you can use another method 
`$textfield->render($textfield->renderLayout,[]);`
It doesn't work either.
These methods call the text field method `$textfield->getLayoutData()`
Inside this method there is a method call `$textfield->getOptions()`
And inside this method, data is collected from `$textfield->element` which has the type SimpleXMLElement.
This functionality should be called in the `$textfield->Setup()` method. 
But it is not called there, but is called when called `->renderField();` and `->render();`.
Thus, it is not possible to create fields in PHP.
[		foreach ($this->element->children() as $option)](https://github.com/joomla/joomla-cms/blob/50685f9147ed458be421e544591444e27ca4613b/libraries/src/Form/Field/TextField.php#L258)

Of course, this can be solved with crutches and props by creating a separate pseudo element with the `SimpleXMLElement` type. or by creating a Layout object separately from the text field object. But this is not normal at all. This is NOT observed in other fields. Moreover, the text field is the base class for many other fields. It is necessary to transfer the call `$textfield->getOptions()` to the method `$textfield->Setup()`

### DIFF
--- a/libraries/src/Form/Field/TextField.php
+++ b/libraries/src/Form/Field/TextField.php
@@ -50,6 +50,14 @@ class TextField extends FormField
 	protected $inputmode;
 
 	/**
+	 * The field option objects.
+	 *
+	 * @var    array
+	 * @since  3.2
+	 */
+	protected $options;
+
+	/**
 	 * The name of the form field direction (ltr or rtl).
 	 *
 	 * @var    string
@@ -192,6 +200,19 @@ class TextField extends FormField
 
 			$this->addonBefore = (string) $this->element['addonBefore'];
 			$this->addonAfter  = (string) $this->element['addonAfter'];
+			$this->useglobal = (bool) $this->element['useglobal'];
+			
+			foreach ($this->element->children() as $option)
+			{
+				// Only add <option /> elements.
+				if ($option->getName() !== 'option')
+				{
+					continue;
+				}
+
+				// Create a new option element in list options.
+				$this->options[] = ['value' => (string) $option['value'], 'text' => (string) $option];
+			}
 		}
 
 		return $result;
@@ -206,7 +227,7 @@ class TextField extends FormField
 	 */
 	protected function getInput()
 	{
-		if ($this->element['useglobal'])
+		if ($this->useglobal)
 		{
 			$component = Factory::getApplication()->input->getCmd('option');
 
@@ -253,16 +274,15 @@ class TextField extends FormField
 	 */
 	protected function getOptions()
 	{
+		if(empty($this->options))
+		{
+			return [];
+		}
+
 		$options = array();
 
-		foreach ($this->element->children() as $option)
+		foreach ($this->options as $option)
 		{
-			// Only add <option /> elements.
-			if ($option->getName() !== 'option')
-			{
-				continue;
-			}
-
 			// Create a new option object based on the <option /> element.
 			$options[] = HTMLHelper::_(
 				'select.option', (string) $option['value'],
--- a/libraries/src/Form/Field/TextField.php
+++ b/libraries/src/Form/Field/TextField.php
@@ -53,9 +53,9 @@ class TextField extends FormField
 	 * The field option objects.
 	 *
 	 * @var    array
-	 * @since  3.2
+	 * @since  4.0
 	 */
-	protected $options;
+	protected $options = [];
 
 	/**
 	 * The name of the form field direction (ltr or rtl).
@@ -201,7 +201,7 @@ class TextField extends FormField
 			$this->addonBefore = (string) $this->element['addonBefore'];
 			$this->addonAfter  = (string) $this->element['addonAfter'];
 			$this->useglobal = (bool) $this->element['useglobal'];
-			
+
 			foreach ($this->element->children() as $option)
 			{
 				// Only add <option /> elements.
--- a/libraries/src/Form/Field/TextField.php
+++ b/libraries/src/Form/Field/TextField.php
@@ -274,7 +274,7 @@ class TextField extends FormField
 	 */
 	protected function getOptions()
 	{
-		if(empty($this->options))
+		if (empty($this->options))
 		{
 			return [];
 		}
--- a/libraries/src/Form/FormField.php
+++ b/libraries/src/Form/FormField.php
@@ -688,7 +688,7 @@ abstract class FormField
 		$this->layout = !empty($this->element['layout']) ? (string) $this->element['layout'] : $this->layout;
 
 		$this->parentclass = isset($this->element['parentclass']) ? (string) $this->element['parentclass'] : $this->parentclass;
-		
+
 		$this->labelElement = $this->element['label'] ? (string) $this->element['label'] : null;
 
 		// Add required to class list if field is required.
--- a/libraries/src/Form/FormField.php
+++ b/libraries/src/Form/FormField.php
@@ -808,7 +808,7 @@ abstract class FormField
 		}
 
 		// Get the label text from the XML element, defaulting to the element name.
-		$title = $this->element['label'] ? (string) $this->element['label'] : (string) $this->element['name'];
+		$title = $this->label ? (string) $this->label : (string) $this->fieldname;
 		$title = $this->translateLabel ? Text::_($title) : $title;
 
 		return $title;
@@ -831,7 +831,7 @@ abstract class FormField
 		$data = $this->getLayoutData();
 
 		// Forcing the Alias field to display the tip below
-		$position = $this->element['name'] === 'alias' ? ' data-bs-placement="bottom" ' : '';
+		$position = $this->fieldname === 'alias' ? ' data-bs-placement="bottom" ' : '';
 
 		// Here mainly for B/C with old layouts. This can be done in the layouts directly
 		$extraData = array(
@@ -1319,7 +1319,7 @@ abstract class FormField
 	protected function getLayoutData()
 	{
 		// Label preprocess
-		$label = !empty($this->element['label']) ? (string) $this->element['label'] : null;
+		$label = !empty($this->label) ? (string) $this->label : null;
 		$label = $label && $this->translateLabel ? Text::_($label) : $label;
 
 		// Description preprocess
--- a/libraries/src/Form/FormField.php
+++ b/libraries/src/Form/FormField.php
@@ -688,6 +688,8 @@ abstract class FormField
 		$this->layout = !empty($this->element['layout']) ? (string) $this->element['layout'] : $this->layout;
 
 		$this->parentclass = isset($this->element['parentclass']) ? (string) $this->element['parentclass'] : $this->parentclass;
+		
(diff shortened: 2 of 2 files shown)


korenevskiy:
https://github.com/joomla/joomla-cms/issues/36636

korenevskiy:
It's hard to imagine that it was previously impossible to dynamically create fields in PHP. This feature is the main problem for developers.
.
But now this field can be created dynamically in PHP.

Fedik:
There no problem to use any field class on it own. 
Example for text field:
```php
$field = new TextField;
$field->setup('<field name="blabla"/>', 'some value');
echo $field->input;

// Or just:
$field = new TextField;
$field->setup('<field />', 'some value');
echo $field->input;
```

Each field in Joomla is require XML, whole Form API based on it. 
No need a hacking around.

korenevskiy:
But type SimpleXMLElement  of the $element
`	public function setup(\SimpleXMLElement $element, $value, $group = null)`

Fedik:
yeah, well, but the idea is the same:
```php
$field = new TextField;
$field->setup(new SimpleXMLElement('<field />'), 'some value');
echo $field->input;
```

korenevskiy:
> yeah, well, but the idea the same:
> 
> ```
> $field = new TextField;
> $field->setup(new SimpleXMLElement('<field />'), 'some value');
> echo $field->input;
> ```

That's how it works. Is it OK to create an unnecessary object?

I watched how the FIELDS component works. The plugin of this component creates a complete object XML form similar php object . In order to then get a regular object back from it. Thus, it is created  XML object  from a regular object, and then a regular object is created from XML. All fields are subject to transformation. each property is converted to an attribute, and then the attributes are converted back to properties.

Well, now you can see that by creating a field plugin for an article, this object will be converted to XML and vice versa. And the field plugin contains the usual other field. So to create a field in the article. There will be a four conversion of the object to XML and back, and then again from the object to XML and back to load the usual field into the plugin field.

I was creating my own plugin fields for articles. And also created another field. I can't pass data from the plugin to the field. Since it all happens by conversion via XML.
Thanks a lot for the experience. I will use it in the future. But I think that dynamic fields should work intuitively.

korenevskiy:
Well, now you can see that by creating a field plugin for an article, this object will be converted to XML and vice versa. And the field plugin contains the usual other field. So to create a field in the article. There will be a double conversion of the object to XML and back, and then again from the object to XML and back to load the usual field into the plugin field.

korenevskiy:
@Fedik 
How do you color code?

korenevskiy:
@Fedik  @richard67 @brianteeman 
And who has the authority to say whether the Joomla API will support the creation of tabs, a set of fields, and just fields dynamically in the future, without conversion to XML and vice versa? This is very limiting to create multifunctional components based on materials.

richard67:
@korenevskiy What do you want from me? I‘ve only closed your issue because that’s our regular procedure, close an issue when there is a PR which claims to solve it. You should have done that yourself and should have mentioned the issue in the PR description.

brianteeman:
@richard67 he is looking for your approval on this PR

richard67:
> @richard67 he is looking for your approval on this PR

I will not approve that.

korenevskiy:
The discussion did not take place.

richard67:
There is simply nothing to discuss for me because I am the wrong guy. I am not a field expert. When I have something to say because I have some knowledge on it, I will say, but when I have not much knowledge I prefer not to have to say something :-)

korenevskiy:
> There is simply nothing to discuss for me because I am the wrong guy. I am not a field expert. When I have something to say because I have some knowledge on it, I will say, but when I have not much knowledge I prefer not to have to say something :-)


I thought you were not the most important here, but you are the third oldest. I have seen your participation in many places. And what are you a master at?

Fedik:
> Is it OK to create an unnecessary object?

Yes, every Joomla field require to have the XML element.
You can pass any supported data to your field, just make sure you call setup() first.

> How do you color code?

It is MD markup

richard67:
> And what are you a master at?

I have a master degree in Electrical Engineering :-)

korenevskiy:
> I have a master degree in Electrical Engineering :-)

I am also an electronic engineer, and I can collect ZX-Spectrum. 
And in Joomla, what are you an expert in?, 
I'm just sorry that I needlessly distracted you. In the future, I will close the issue before doing PR in the future. I promise.

richard67:
> And in Joomla, what are you an expert in?

I can find bugs by their smell :-)

korenevskiy:
> Yes, every Joomla field require to have the XML element.
> You can pass any supported data to your field, just make sure you call setup() first.

These requirements described by you are the current requirements. Each code must support these requirements. But the requirements can be changed if the requirements complicate the code unnecessarily.

Programming language developers have come up with interfaces to respect typization. But we are in a transitional moment to the world of typization, but we  convert everything to XML and back. I wrote one cool field. The field has the form of a table, each cell is a different field (text, area, select, radio button). There can be any other field in the table cells.

A new table field, can take data from SQL, or just JSON, Rows can be created, Each row is draggable and removable. Then I create a plugin field for the material. Now I need to add the table field to the plugin field. Having written the field path and loaded the typization, I could not pass an array of values to the table. Since the FIELDS plugin converts all fields to XML FORM in order to then create PHP objects based on XML FORM, it's hard. I ran into the problem that everything is done through one place, through XML. 
**This happens at different levels of the materials component. Since we are gradually moving towards typization, maybe we should reconsider whether there are requirements to always contain XML in the field?**
By creating a multi-level field that contains other fields that may contain other fields, this creates problems.

dgrammatiko:
@korenevskiy 

XML => DECLARATIVE (here's a config file XML, json, whatever)
PHP => IMPERATIVE (here's the code that creates the field like @Fedik shared above)

What you're demoing here is the removal of the XML parser step. Is that so much a limiting factor here? If you want to improve the Form/Fields code just `composer install Symphony Form` (whatever the package name) and connect the missing links...

korenevskiy:
> @korenevskiy
> 
> XML => DECLARATIVE (here's a config file XML, json, whatever) PHP => IMPERATIVE (here's the code that creates the field like @Fedik shared above)
> 
> What you're demoing here is the removal of the XML parser step. Is that so much a limiting factor here? If you want to improve the Form/Fields code just `composer install Symphony Form` (whatever the package name) and connect the missing links...



> @korenevskiy
> 
> XML => DECLARATIVE (here's a config file XML, json, whatever) PHP => IMPERATIVE (here's the code that creates the field like @Fedik shared above)
> 
> What you're demoing here is the removal of the XML parser step. Is that so much a limiting factor here? If you want to improve the Form/Fields code just `composer install Symphony Form` (whatever the package name) and connect the missing links...

Declarative and imperative approach, both are good. But, in the problem I described, I showed that multi-level transformations take place back and forth.
The declarative  approach is a way of programming, but it is not a way of storing objects.
If we pay attention to WPF C#. We will see XAML is a form of writing code, but the objects themselves are the same for any method. There declarative objects in the text are converted to C# objects when compiled. In our case, we do not have a single object form, we will not only write declarative code, but we will store it after the program is launched. We convert data back and forth many times. On the other hand, the support of the imperative method is not an opposition of the approach. I'm not asking you to delete the SETUP method. Let's be honest If I create a field in PHP and put XML in it. Then the imperative approach says that the XML I put is automatically the node of the entire XML FORM. But this is not the case, nothing like this happens, creating a separate field and putting a field string in it, this declarative object is not included in the XML form at all and is no longer used. It does not attach to a common XML object. Unless this field was originally in a shared XML object.

And also, all field objects live by themselves separately from XML, even though they are in each other. Let's look at the Register class. It easily loads declarative data.
The penultimate problem is that when rendering, XML is converted back to an object, and then it is converted to HTML. Even from the point of view of declarative objects, we will still have to convert everything to HTML via PHP.
It might be useful to navigate through the XML object model like SQL queries through the XML tree. We have NOT implemented this, but even in this case we should have data isolation.
XML does not stand up to criticism as a data warehouse, it is not ideal for supporting big data storage and fast navigation through this data. If we continue the declarative idea, then we should inherit the JTable class from XML so that the declarativeness is complete, so that the Database data is converted directly to XML. We also need to make the session storage in the form of an XML object.
In WPF C#, declarativeness is typed, there are types and namespaces at the XML level. In order for Joomla to expand into the corporate sector, it is necessary to move towards typing. Storing data in XML is a step back from typization. The whole framework strives for typing, but we will store the data in untyped XML. What's the point?

1.I suggest using XML only when loading data into the SETUP method.
2.Store data in properties similar to JRegistry.
3.Add the ToXML method for the fields.
4.Add a dispatcher to the HTML code of the tabs of the material editor to add your own tabs with PHP code.

Let's not confuse declarative code in a file with declarative objects in RAM. If we stop confusing, we won't shoot ourselves in the foot. The imperative and declarative approaches have advantages and disadvantages, let's take the best out of both, and not replace one with the other by creating new problems by solving others.

Fedik:
I closing it, because it does not fix anything.

korenevskiy:
> I closing it, because it does not fix anything.

Have you personally checked how these PR fixes work? This code works as I described it here:
```
$field = Joomla\CMS\Form\FormHelper::loadFieldType('text', true);
$field->name = 'user';
echo $field->renderField();
```
Why did you close the PR without checking? But even if I had mistakes, I could make new corrections. But you just took and closed the PR.

Fedik:
I already wrote, that code is incorrect, you have to call `setup()` for every field, as long as Joomla Form API use XML.

```php
$field = Joomla\CMS\Form\FormHelper::loadFieldType('text', true);
$field->setup(new SimpleXMLElement('<field />'));
$field->name = 'user';
echo $field->renderField();
```

korenevskiy:
> I already wrote, that code is incorrect, you have to call `setup()` for every field, as long as Joomla Form API use XML.

Okay, I get it. The `Setup()` method is the second constructor of the field. It is not clear why 2 constructors are used in the fields. What is the purpose of using the method executing the second constructor? What is it invented for? The Setup method should logically be optional for loading XML. But if it is mandatory, then why was it invented? Why not immediately pass XML to the field constructor?
PS That's a good question.

Fedik:
because:
```php
$field = Joomla\CMS\Form\FormHelper::loadFieldType('text', true);
```

korenevskiy:
> 

It doesn't explain anything. Such a method could be exactly XML as an example.
an invented example:
```javascript
$field = Joomla\CMS\Form\FormHelper::loadFieldType('text',  new SimpleXMLElement('<field />'), true);
```

This `loadFieldType()` method is a wrapper of the field class. You told me the reason that it's because of the wrapper, but the wrapper is just a short form of a call, it's just a helper. Your statement is similar, that the fire burns because the smoke rises.

PR #36732, opened 2022-01-18
Removed hard limit of 20 to the module tags

Shubhamverma2796:
Pull Request for Issue #36016.

### Summary of Changes
Removed the hard limit for Module Tags-Popular and Module Tags-Similar.

#### For tags popular :
1. Site modules--> New-->Tags-Popular.
2. Set the maximum tags to a number > 20 and make sure that the tags are displayed correctly in the frontend.

#### For tags similar :
1.Site modules--> New-->Tags-similar.
2. Set the maximum items to a number > 20 and make sure that the tags are displayed correctly in the frontend

see #36016 

Any number is possible for both tags-popular and tags-similar.

### DIFF
--- a/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
+++ b/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
@@ -126,6 +126,12 @@ abstract class TagsPopularHelper
 			{
 				// Backup bound parameters array of the original query
 				$bounded = $query->getBounded();
+
+				if ($maximum > 0)
+				{
+					$query->setLimit($maximum);
+				}
+				
 				$query->order($db->quoteName('count') . ' DESC');
 				$equery = $db->getQuery(true)
 					->select(
--- a/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
+++ b/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
@@ -38,7 +38,7 @@ abstract class TagsPopularHelper
 		$user        = Factory::getUser();
 		$groups      = $user->getAuthorisedViewLevels();
 		$timeframe   = $params->get('timeframe', 'alltime');
-		$maximum =     (int) $params->get('maximum', 5);
+		$maximum     = (int) $params->get('maximum', 5);
 		$order_value = $params->get('order_value', 'title');
 		$nowDate     = Factory::getDate()->toSql();
 		$nullDate    = $db->getNullDate();
--- a/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
+++ b/modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
@@ -38,7 +38,7 @@ abstract class TagsPopularHelper
 		$user        = Factory::getUser();
 		$groups      = $user->getAuthorisedViewLevels();
 		$timeframe   = $params->get('timeframe', 'alltime');
-		$maximum     = $params->get('maximum', 5);
+		$maximum =     (int) $params->get('maximum', 5);
 		$order_value = $params->get('order_value', 'title');
 		$nowDate     = Factory::getDate()->toSql();
 		$nullDate    = $db->getNullDate();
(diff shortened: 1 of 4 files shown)


pabloarias:
I have tested this item :white_check_mark: successfully on 5f4e11c0e9e94a4997688c93c38726f258894510<br /><br />Tested with PHP 8.0.10.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36732">issues.joomla.org/tracker/joomla-cms/36732</a>.</sub>

modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
```diff
@@ -126,8 +126,6 @@ public static function getList(&$params)
 			{
 				// Backup bound parameters array of the original query
 				$bounded = $query->getBounded();
-
-				$query->setLimit($maximum);
```
richard67, line 130: Hmm, removing this could have a performance impact because here the limit was applied to the inner query, and so the result set was reduced earlier. Now when this is removed, the limit is only applied at the end for the outer query. Maybe we should not remove it here but do the same change as below for the outer query?

richard67:
@Shubhamverma2796 Could you extend your testing instructions to test also "Tags-Similar"? Currently there is only step 1 for "Tags-Popular", so a quick reader might test this only. Thanks in advance.

modules/mod_tags_popular/src/Helper/TagsPopularHelper.php
Shubhamverma2796: so should I add this line again? Though I'm not sure about the performance impact part.

richard67: Not just add back that line. Wrap it into an `if ($maximum > 0)` like you did below in lines 159 to 162.

It will then result in the same query as before plus the checks which you have added for `$maximum > 0`.

When you remove that here, the query will be different to like it was before.

Shubhamverma2796:
> @Shubhamverma2796 Could you extend your testing instructions to test also "Tags-Similar"? Currently there is only step 1 for "Tags-Popular", so a quick reader might test this only. Thanks in advance.

requested changes done. Please check sir.

richard67:
> requested changes done. Please check sir.

@Shubhamverma2796 Thanks. Looks ok to me. By the way, no need for "sir". We are a community and so some kind of fellows here.

@pabloarias Could you redo your test, and if not done last time then test also the "Tags-Similar" this time? Thanks in advance.

pabloarias:
I've added example tags to more articles and tested Tags-Similar module. It works fine!

Tested with options All, Any and Half in Match Type. Also tested with different orders results.

pabloarias:
I have tested this item :white_check_mark: successfully on efd471fbe43a6735071f773e10ed4902c65e123a<br /><br />I've uninstalled the path, fetched data from Github and installed the path again.

After that I've been playing with order, direction and font size in Cloud Layout.

It seems to work perfectly! Thank you very much!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36732">issues.joomla.org/tracker/joomla-cms/36732</a>.</sub>

BertaOctech:
I have tested this item :white_check_mark: successfully on efd471fbe43a6735071f773e10ed4902c65e123a<br /><br />Hi,
it worked as expected for me, too.
Once applied the patch I was able to show more than 20 tags in the popular tags module and more than 20 items in the similar tags module.
They were displayed correctly.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36732">issues.joomla.org/tracker/joomla-cms/36732</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36732">issues.joomla.org/tracker/joomla-cms/36732</a>.</sub>

Shubhamverma2796:
Thanks to everyone

bembelimen:
Thx

Shubhamverma2796:
Thanks for merging.

Issue #36668, opened 2022-01-13
[4.0] custom fields of type media disappear when switching article to different category


superknutsel:
### Steps to reproduce the issue

- Create a custom field of the type media, assign to all categories
- Create an article and fill in the custom field with an image and save the article
- Change the categorie to a different one and check the tab with the custom field. It should be gone until you hit save. Then the field will be back but empty.

### Expected result
I would expect the fields not to disappear and keep their content.

### Actual result
Now after moving an article to a different categorie your custom fields are empty, probably because they had some sort of "reset".

### System information (as much as possible)
Joomla 4.0.5
PHP 8.0.14
MySQL 8.0.27

drmenzelit:
I can confirm the issue

alikon:
fixed by #36736

Issue #36734, opened 2022-01-18
[J4.x] Media fields in non-repeatable custom fields subform toggle display on save


crystalenka:
Buckle up, folks, this is a weird one.

### Steps to reproduce the issue
1. Create a media custom field. Configuration doesn't matter.
2. Create another custom field, any type but media. Configuration doesn't matter.
3. Create a subform with those two fields. **Set the "Repeatable" toggle to no. (This is important!)**
4. Open an article with the subform field. Fill out the fields, or don't. It doesn't matter.
5. Save the article.

### Expected result

The media field still displays in the subform, and if any value was added to the media field, it remains.

### Actual result

The media field _disappears_.

Before Save:
<img width="1414" alt="Screen Shot 2022-01-18 at 19 29 41" src="https://user-images.githubusercontent.com/9141288/149988210-b23832d3-6a81-402e-ad9d-3b9f3f610636.png">

After Save:
<img width="1414" alt="Screen Shot 2022-01-18 at 19 29 48" src="https://user-images.githubusercontent.com/9141288/149988226-d65bf75a-d946-40ff-b4cb-4f6f0b60e821.png">

Additionally, if you save the article again, the media field will magically reappear (but empty of any value). This happens regardless of what value (if any) you put in the media field. The only time the media field does NOT reappear is if it is the only field in the subform, which is why I included step 2 above.

### System information (as much as possible)

PHP Version: 8.0.14
Joomla! 4.0.6

### Additional comments
It's really fun if you save the article so the subform disappears, go back to the subform configuration, add another media field, and go back to edit the article. The media fields switch on every save and are never shown at the same time. 😶

brianteeman:
confirmed

crystalenka:
Thanks Brian. Felt like I was going crazy for a while!

brianteeman:
that;s why I quickly posted a confirmation. No idea on the solutions (yet)

richard67:
Maybe something for @Fedik ?

richard67:
I wish we always had such precise instructions on how to reproduce an issue.

brianteeman:
Must be related to the subform NOT being repeatable as everything works as expected when it is repeatable

crystalenka:
Yup, that's why I included the important note on that step. :)

brianteeman:
About to start filming but I found the problem Maybe someone else will write the fix while I am filming

The record saved in #__filed_values when it is **not** a repeatable subform 

[...]

PR #36736, opened 2022-01-18
Media fields in non-repeatable subforms are broken

nikosdion:
Pull Request for Issue #36734 .

### Summary of Changes

Addresses the data format discrepancy between what `Joomla\CMS\Form\Field\AccessiblemediaField` expects and what a non-repeatable subform custom field provides.

### Testing Instructions

Per issue #36734:

1. Create a media custom field. Configuration doesn't matter.
2. Create another custom field, any type but media. Configuration doesn't matter.
3. Create a subform with those two fields. Set the "Repeatable" toggle to no. (This is important!)
4. Open an article with the subform field. Fill out the fields, or don't. It doesn't matter.
5. Save the article.

The media field _disappears_ every other save. Any images you set in the field is ignored; the field appears in its default (empty) state when it reappears.

The media field still displays in the subform, and if any value was added to the media field, it remains.

None

### Technical information

This problem happens because Joomla is playing fast and loose with data types.

The subform custom field stores its value as a JSON document. This includes the values of each and every subfield contained in it. It is always decoded to an associative array, including the subfields' data.

The Media custom field's value is meant to be either an object or a string (which can be a JSON-serialised object or a legacy image file path without alt text). That's how the setup() method of the AccessiblemediaField class expects its values.

When combining the two you create the observed problem.

When you load a custom field it decodes the JSON value into an associative array. This means that the internal data of the Media custom field is also decoded as an array. Since the AccessiblemediaField does not accept an array it considers the data invalid, setup() returns false and the field is removed from the form. Therefore its data is removed when you save the content (e.g. article) again. The next page load has no value for this field stored so the field falls back to its default empty state due to the fall-through of the setup() method's opening if-block and displays, albeit empty.

The dead simple solution is to check if the incoming value is an array and type cast it to an object. This satisfies the if-block's conditions and the field displays correctly.

### DIFF
--- a/libraries/src/Form/Field/AccessiblemediaField.php
+++ b/libraries/src/Form/Field/AccessiblemediaField.php
@@ -131,6 +131,19 @@ class AccessiblemediaField extends SubformField
 	public function setup(\SimpleXMLElement $element, $value, $group = null)
 	{
 		/**
+		 * When you have subforms which are not repeatable (i.e. a subform custom field with the
+		 * repeat attribute set to 0) you get an array here. This happens because the subform data,
+		 * including the inner data of the media field, is stored as JSON and decoded as an
+		 * associative array. Since the code below only expects a string (JSON or legacy single
+		 * image file) or an object it thinks there's invalid data which causes the media field to
+		 * not save its data and, on the next page load, not render at all.
+		 *
+		 * Typecasting that array back to the expected object value solves the discrepancy between
+		 * provided data and expected data, solving the issue.
+		 */
+		$value = is_array($value) ? (object) $value : $value;
+
+		/**
 		 * If the value is not a string, it is
 		 * most likely within a custom field of type subform
 		 * and the value is a stdClass with properties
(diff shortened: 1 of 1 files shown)


brianteeman:
I have tested this item :white_check_mark: successfully on 76f60f723b017f50f89a5b9909bab2b52f9f6122<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36736">issues.joomla.org/tracker/joomla-cms/36736</a>.</sub>

crystalenka:
I have tested this item :white_check_mark: successfully on 76f60f723b017f50f89a5b9909bab2b52f9f6122<br /><br />Thank you!!!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36736">issues.joomla.org/tracker/joomla-cms/36736</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36736">issues.joomla.org/tracker/joomla-cms/36736</a>.</sub>

alikon:
fix #36668 also

bembelimen:
Thx

PR #36751, opened 2022-01-20
[4.2] Smart Search: Add optimisations & garbage collection

Hackwar:
### Summary of Changes
There are situations where the index of Smart Search can get poluted with old data (unused terms and taxonomies, already deleted content items) and where the only solution so far has been to clear the reindex and start indexing all over again. This is not something that you really want to do in a live scenario with a large site, so this attempts some different solutions.
First of all, it runs a garbage collection, which by default deletes all entries, which aren't present in the base table of the base component. We then delete all orphaned term and taxonomy mappings as well as all then orphaned taxonomies and terms. In the end, we then execute the already previously present optimisation queries. To trigger all of this, a new dropdown button was added in the backend named "Maintenance", where you can choose between purging the index and doing the optimisation.
Note: The garbage collection has been added as a default to the Adapter base class and there could be situations where this results in a fail. I'm open for ideas here.

### Testing Instructions
1. Apply the PR
2. Click on "Optimise" in the backend
3. Afterwards see, that searching still works properly.

### Notes
1. Since 4.1 is in RC, this will have to go into 4.2. Since there is no 4.2-dev branch yet, I'm pointing this towards 4.1-dev for now.


administrator/components/com_finder/src/Indexer/Taxonomy.php
```diff
@@ -188,7 +188,7 @@ protected static function storeNode($node, $parentId)
 		 * The database did not match the input. This could be because the
 		 * state has changed or because the node does not exist. Let's figure
 		 * out which case is true and deal with it.
-		 * @todo: use factory?
+		 * TODO: use factory?
```
brianteeman: please revert this change

sakiss:
Would it be possible to add that task to the new com_scheduler ?
Removing that responsibility from the admin would be great.

Hackwar:
This task is not something that you should execute regularly.

nadjak77:
I have tested this item :red_circle: unsuccessfully on 7a4097b106a6a7c3dc53b295f6e81d8a4d6d5cd1<br /><br />After applieing the patch I get an error in seachrindex - indes:
There is no "table.columns" asset of a "script" type in the registry. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36751">issues.joomla.org/tracker/joomla-cms/36751</a>.</sub>

Hackwar:
Can you please check if you need to do a composer install and/or npm i? This PR actually does not change anything in that regard, but maybe a previous change introduced a problem here?

webnet-assmann:
I have tested this item :white_check_mark: successfully on 7a4097b106a6a7c3dc53b295f6e81d8a4d6d5cd1<br /><br />I've tested it with the prebuilt package and sample data. Searching was working before and after click on "Optimise".<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36751">issues.joomla.org/tracker/joomla-cms/36751</a>.</sub>

administrator/components/com_finder/src/Controller/IndexController.php
```diff
@@ -37,6 +40,32 @@ public function getModel($name = 'Index', $prefix = 'Administrator', $config = a
...
+	 * @return  boolean  True on success.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function optimise()
+	{
+		$this->checkToken();
+
+		// Optimise the index by first running the garbage collection
+		$app = Factory::getApplication();
```
joomdonation: Shouldn't we use `$this->app` instead of calling `Factory::getApplication();` ?

administrator/components/com_finder/src/Indexer/Adapter.php
```diff
@@ -262,6 +262,38 @@ public function onBuildIndex()
...
+	/**
+	 * Method to remove outdated index entries
+	 *
+	 * @return  integer
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function onFinderGarbageCollection()
+	{
+		$db = Factory::getDbo();
```
joomdonation: Shouldn't we use `$this->db` instead of calling `Factory::getDbo()` ?

joomdonation (commented):
Tested and it worked OK. Just minor things with the code and I will mark my test result.

administrator/components/com_finder/src/Controller/IndexController.php
Hackwar: done

administrator/components/com_finder/src/Indexer/Adapter.php
Hackwar: done

joomdonation:
I have tested this item :white_check_mark: successfully on 49576c6ac01bf13c61cc228501ea51b6b6aa5516<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36751">issues.joomla.org/tracker/joomla-cms/36751</a>.</sub>

joomdonation:
The last change was a minor change, so I restored test result of @webnet-assmann. This PR is now RTC.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36751">issues.joomla.org/tracker/joomla-cms/36751</a>.</sub>

Hackwar:
I'm investigating why the api tests are failing. Please don't merge this right now. I want to check that first.

Hackwar:
I found the issue and fixed it. This could be merged now.

fancyFranci:
Thank you Hannes!

PR #36753, opened 2022-01-20
[5.0] Smart Search: Add debugging features

Hackwar:
### Summary of Changes
In order to gain wider adoption from third party developers, we need to provide them with improved development tools. This PR implements a few features, which become visible when the debug mode is enabled.

#### Display indexed item
First of all we want to be able to see what has been indexed for an item. When you enable debugging, the list of entries in the index will now be linked and clicking on such an entry will give you a view which displays the common data from indexed items, the type specific items, the indexed terms and the associated taxonomies. This allows you to understand if the content was indexed the way you expected it, as well as with the weighing you expected.

#### Debug indexing of one specific item
Again, when debugging is enabled, a new dropdown button in the index view appears with "Index debugging". This button goes to a new view, where you can select a finder plugin and type in an ID to index and it will then run that plugin for that ID. It will then either return what has been prepared to be indexed by the indexer in the result object or whatever output the plugin generated. This should allow a developer to better write a new finder plugin to index their content.

### Testing Instructions
1. Install Joomla and provide some sample data. Please note that this PR requires "npm i" to be run again after it has been applied!
2. Enable debugging mode in global configuration
3. Go to Smart Search and click on an entry in the list to display the data in the index for this item
4. Click on "Index > Index debugging" and select a finder plugin from the list.
5. Type in an ID, for example "60" in case you are using the testing sample data, and click on "Index"

### Notes
1. The dropdown button seems to have a styling error when using a modal button as first entry. The spacing is wrong and I couldn't find out why.
2. The Javascript is most likely "not optimal". I'm happy for corrections.
3. I'm open for better language strings. 
4. Since 4.1 is in RC, this will have to go into 4.2. Since there is no 4.2-dev branch yet, I'm pointing this towards 4.1-dev for now.


administrator/components/com_finder/forms/indexer.xml
```diff
@@ -0,0 +1,20 @@
...
+				required="true"
+		/>
+
+		<field
+				name="id"
+				type="text"
+				label="JGLOBAL_FIELD_ID_LABEL"
+				size="10"
+				required="true"
+		/>
```
brianteeman:
```suggestion
		<field
			name="plugin"
			type="plugins"
			label="COM_FINDER_FIELD_FINDER_PLUGIN_LABEL"
			folder="finder"
			required="true"
		/>

		<field
			name="id"
			type="text"
			label="JGLOBAL_FIELD_ID_LABEL"
			required="true"
		/>
```

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Method to remove outdated index entries
+	 *
+	 * @return  integer
+	 *
+	 * @since   4.1
```
PhilETaylor:
```suggestion
	 * @since   ___DEPLOY_VERSION__
```

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+
+	/**
+	 * Method to index an item.
+	 *
+	 * @param   Result  $item  The item to index as a Result object.
+	 *
+	 * @return  boolean  True on success.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  Exception on database error.
```
PhilETaylor: An abstract function cannot throw an Exception

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+	 */
+	abstract protected function index(Result $item);
+
+	/**
+	 * Method to reindex an item.
+	 *
+	 * @param   integer  $id  The ID of the item to reindex.
```
PhilETaylor:
```suggestion
	 * @param   int  $id  The ID of the item to reindex.
```

Param uses short version, return uses long version

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+		return true;
+	}
+
+	/**
+	 * Method to setup the adapter before indexing.
+	 *
+	 * @return  boolean  True on success, false on failure.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  Exception on database error.
```
PhilETaylor: Abstract method cannot throw an exception

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+	 * Method to get a list of content items to index.
+	 *
+	 * @param   integer         $offset  The list offset.
+	 * @param   integer         $limit   The list limit.
+	 * @param   QueryInterface  $query   A QueryInterface object. [optional]
+	 *
+	 * @return  Result[]  An array of Result objects.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  Exception on database error.
```
PhilETaylor: Are all these "Exception on error" docs saying an exception of type `\Exception` is expected on error, or are they just commentary? If they are the latter they should be removed or correctly explain what type of exceptions are thrown. Throughout this PR.

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+			$item->layout = $this->layout;
+		}
+
+		return $items;
+	}
+
+	/**
+	 * Method to get the SQL query used to retrieve the list of content items.
+	 *
+	 * @param   mixed  $query  A QueryInterface object. [optional]
```
PhilETaylor: If its nullable then the docs should state that correctly as a type. 

```suggestion
	 * @param   mixed|null  $query  A QueryInterface object
```

administrator/components/com_finder/src/Model/IndexerModel.php
```diff
@@ -11,13 +11,38 @@
...
+	 * @param   array    $data      Data for the form.
+	 * @param   boolean  $loadData  True if the form is to load its own data (default case), false if not.
+	 *
+	 * @return  Form
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 *
+	 * @throws \Exception
+	 */
+	public function getForm($data = array(), $loadData = true)
```
PhilETaylor: Use short array syntax throughout

```suggestion
	public function getForm($data = [], $loadData = true)
```

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,104 @@
...
+
+		$db->setQuery($query);
+
+		return $db->loadObjectList();
+	}
+
+	/**
+	 * Get taxonomies associated with a finder link
+	 *
+	 * @return  object[]
```
PhilETaylor:
```suggestion
	 * @return  \stdClass[]
```

administrator/components/com_finder/src/View/Indexer/HtmlView.php
```diff
@@ -20,5 +25,51 @@
  */
 class HtmlView extends BaseHtmlView
 {
+	/**
+	 * @var Form $form
```
PhilETaylor:
```suggestion
	 * @var   Form  $form
```

administrator/components/com_finder/src/View/Indexer/HtmlView.php
```diff
@@ -20,5 +25,51 @@
...
+	/**
+	 * Method to configure the toolbar for this view.
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function addToolbar()
+	{
+		// Get the toolbar object instance
```
PhilETaylor:
```suggestion
```

build/media_source/com_finder/js/debug.es6.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * @copyright  (C) 2018 Open Source Matters, Inc. <https://www.joomla.org>
```
PhilETaylor:
```suggestion
 * @copyright  (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

administrator/components/com_finder/src/Indexer/Debugadapter.php
Hackwar: PHPStorm added this automatically. I will remove all of this.

Quy:
Please change branch. Thanks.

Hackwar:
Hmm, that was a failure...

administrator/components/com_finder/src/Indexer/Debugadapter.php
Hackwar: Ok, no, PHPStorm did not add all of this automatically. These are part of the original Adapter class already. Since I just copied this over, I wouldn't want to change it further than necessary. removing all those unnecessary tags should be done in another PR.

brianteeman:
Please update the markup of the table to follow the accessibility standards used on every table in the admin

.drone.yml
```diff
@@ -57,13 +57,13 @@ steps:
...
       - ./libraries/vendor/bin/phpunit --testsuite Unit
 
-  - name: php81-unit
-    depends_on: [ phpcs ]
-    image: joomlaprojects/docker-images:php8.1
-    failure: ignore
-    commands:
-      - php -v
-      - ./libraries/vendor/bin/phpunit --testsuite Unit
+#  - name: php81-unit
```
brianteeman: is this change intentional?

Hackwar: No, this happened when switching the branch this was based on. I still need to fix this.

administrator/components/com_finder/tmpl/item/default.php
```diff
@@ -0,0 +1,87 @@
...
+			<dt class="col-sm-3"><?php echo $key; ?></dt>
+			<dd class="col-sm-9"><?php echo $value; ?></dd>
+			<?php endforeach; ?>
+		</dl>
+	</fieldset>
+	<fieldset>
+		<caption><?php echo Text::_('COM_FINDER_ITEM_FIELDSET_TERMS_TITLE'); ?></caption>
+		<table class="table">
+			<thead>
+			<tr>
```
brianteeman: Please update this table to use the accessibility standards used in all other tables

.drone.yml
richard67: I've allowed myself to update the branch, so this is fixed.

administrator/components/com_finder/tmpl/item/default.php
Hackwar: I've added the scop="col" attributes. Is it necessary to add an additional caption, when there is a caption for the fieldset right above it?

brianteeman: take a look at any admin table and you will see exactly what is needed

Hackwar: I'm sorry, but I looked at all of this extensively and what I could find was the hidden caption and spans in the head of the tables. However since there is a visible caption above each table and these tables are neither sortable nor filterable (and it also is unnecessary to have this) I'm fine with the way it is now. I'm not really happy to copy this over, creating duplicate content on the page, just for the sake of having this the same everywhere. I'd rather not start a cargo cult here.

If I missed something besides this, please help me out here.

brianteeman: the <caption> element must be the first child of its parent [< table >](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table) element.

administrator/components/com_finder/tmpl/item/default.php
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+defined('_JEXEC') or die;
+
+use Joomla\CMS\Language\Text;
+?>
+<div role="main">
+	<h1 class="mb-3"><?php echo $this->item->title; ?></h1>
+	<fieldset>
+		<caption><?php echo Text::_('COM_FINDER_ITEM_FIELDSET_ITEM_TITLE'); ?></caption>
```
brianteeman: for a fieldset it should be a legend not a caption
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend

administrator/components/com_finder/tmpl/item/default.php
```diff
@@ -0,0 +1,87 @@
...
+		<dl class="row">
+			<?php foreach ($this->item as $key => $value) : ?>
+			<dt class="col-sm-3"><?php echo $key; ?></dt>
+			<dd class="col-sm-9"><?php echo $value; ?></dd>
+			<?php endforeach; ?>
+		</dl>
+	</fieldset>
+	<fieldset>
```
brianteeman, line 25: Is this really a fieldset?

administrator/components/com_finder/tmpl/item/default.php
brianteeman: but none of this on this page should be a fieldset.

The < fieldset > [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element is used to group several controls as well as labels ([<label>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)) within a web form.

administrator/components/com_finder/tmpl/item/default.php
```diff
@@ -0,0 +1,87 @@
...
+				<th scope="col">phrase</th>
+				<th scope="col">weight</th>
+				<th scope="col">links</th>
+				<th scope="col">language</th>
+			</tr>
+			</thead>
+			<tbody>
+			<?php foreach ($this->terms as $term) : ?>
+				<tr>
+					<td><?php echo $term->term_id; ?></td>
```
brianteeman: If one of the cells in a row is the main indicator (which I am guessing is the term) then it should also be a th with a scope =row

richard67:
@Hackwar Do you think you can implement @brianteeman 's suggestions?

Hackwar:
@brianteeman can you have a look if the changes which I made would be acceptable?

brianteeman:
The tables still do not have a caption. This is essential.

Hackwar:
I kept the caption short. Fine like so? Can you mark those conversations then as resolved?

brianteeman:
I need to review it in the browser. Which I will di in the next day or so

brianteeman:
image

width issues

brianteeman:
> I kept the caption short. Fine like so? Can you mark those conversations then as resolved?

i do nnot have the access to mark the 3 a11y issues as resolved but they as far as i am concerned

brianteeman:
image


For toolback please use the same icon and code as seen in com_workflows transitions

image

Hackwar:
> image
> 
> For toolback please use the same icon and code as seen in com_workflows transitions
> 
> image

I don't know how you got that, but my button already looks like the correct one.

Hackwar:
> image
> 
> width issues

fixed.

brianteeman:
this is what I am seeing 
image

image

Hackwar:
It is just ToolbarHelper::back() and that is returning different code here. Maybe you got some other stuff in your codebase interfering there? That is just not what is in the code and what is being returned on my end. And comparing this with our current 4.2-dev, my version and 4.2-dev agree with each other.

brianteeman:
@Hackwar I have verified once again my previous findings by building a completely new site on a new server using the pre-built package for this PR. I will submit a PR to your branch to fix that

brianteeman:
https://github.com/Hackwar/joomla-cms/pull/53

Hackwar:
done

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getItem()
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
+		$query = $db->getQuery(true)
+			->select('*')
+			->from($db->quoteName('#__finder_links', 'l'))
+			->where($db->quoteName('l.link_id') . ' = ' . $link_id);
```
alikon: please use prepared statements

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	public function getTerms()
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
+		$query = $db->getQuery(true)
+			->select('t.*, l.*')
+			->from($db->quoteName('#__finder_links_terms', 'l'))
+			->leftJoin($db->quoteName('#__finder_terms', 't') . ' ON ' . $db->quoteName('t.term_id') . ' = ' . $db->quoteName('l.term_id'))
+			->where($db->quoteName('l.link_id') . ' = ' . $link_id)
```
alikon: please use prepared statements & quotename

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	public function getTaxonomies()
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
+		$query = $db->getQuery(true)
+			->select('t.*, m.*')
+			->from($db->quoteName('#__finder_taxonomy_map', 'm'))
+			->leftJoin($db->quoteName('#__finder_taxonomy', 't') . ' ON ' . $db->quoteName('t.id') . ' = ' . $db->quoteName('m.node_id'))
+			->where($db->quoteName('m.link_id') . ' = ' . $link_id)
```
alikon: ++

alikon (changes requested):

administrator/components/com_finder/src/Model/ItemModel.php
Hackwar: done

administrator/components/com_finder/src/Model/ItemModel.php
Hackwar: done

administrator/components/com_finder/src/Model/ItemModel.php
Hackwar: done

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,107 @@
...
+ * @subpackage  com_finder
+ *
+ * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\Component\Finder\Administrator\Model;
+
+use Joomla\CMS\Factory;
+use Joomla\CMS\MVC\Model\BaseDatabaseModel;
```
Quy, line 14:
```suggestion
use Joomla\CMS\MVC\Model\BaseDatabaseModel;
use Joomla\Database\ParameterType;
```

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,107 @@
...
+	 */
+	public function getItem()
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
+		$query = $db->getQuery(true)
+			->select('*')
+			->from($db->quoteName('#__finder_links', 'l'))
+			->where($db->quoteName('l.link_id') . ' = :link_id')
+			->bind(':link_id', $link_id);
```
Quy:
```suggestion
			->bind(':link_id', $link_id, ParameterType::INTEGER);
```

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,107 @@
...
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
+		$query = $db->getQuery(true)
+			->select('t.*, l.*')
+			->from($db->quoteName('#__finder_links_terms', 'l'))
+			->leftJoin($db->quoteName('#__finder_terms', 't') . ' ON ' . $db->quoteName('t.term_id') . ' = ' . $db->quoteName('l.term_id'))
+			->where($db->quoteName('l.link_id') . ' = :link_id')
+			->order('l.weight')
+			->bind(':link_id', $link_id);
```
Quy:
```suggestion
			->bind(':link_id', $link_id, ParameterType::INTEGER);
```

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,107 @@
...
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
+		$query = $db->getQuery(true)
+			->select('t.*, m.*')
+			->from($db->quoteName('#__finder_taxonomy_map', 'm'))
+			->leftJoin($db->quoteName('#__finder_taxonomy', 't') . ' ON ' . $db->quoteName('t.id') . ' = ' . $db->quoteName('m.node_id'))
+			->where($db->quoteName('m.link_id') . ' = :link_id')
+			->order('t.title')
+			->bind(':link_id', $link_id);
```
Quy:
```suggestion
			->bind(':link_id', $link_id, ParameterType::INTEGER);
```

administrator/components/com_finder/src/Controller/IndexerController.php
```diff
@@ -310,4 +295,120 @@ public static function sendResponse($data = null)
...
+		$state->batchOffset = 0;
+
+		// Update the indexer state.
+		Indexer::setState($state);
+
+		// Start the indexer.
+		try
+		{
+			// Import the finder plugins.
+			require_once JPATH_ADMINISTRATOR . '/components/com_finder/src/Indexer/Debugadapter.php';
```
laoneo: Is this really needed?

administrator/components/com_finder/src/Controller/IndexerController.php
```diff
@@ -310,4 +295,120 @@ public static function sendResponse($data = null)
...
+			$plugin = Factory::getApplication()->bootPlugin($this->app->input->get('plugin'), 'finder');
+			$plugin->setIndexer(new Debugindexer);
+			$plugin->debug($this->app->input->get('id'));
+
+			$output = '';
+
+			// Create list of attributes
+			$output .= '<fieldset><legend>' . Text::_('COM_FINDER_INDEXER_FIELDSET_ATTRIBUTES') . '</legend>';
```
laoneo: This output stuff should be in a layout.

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,979 @@
...
+	 *
+	 * @param   object  $subject  The object to observe.
+	 * @param   array   $config   An array that holds the plugin configuration.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct(&$subject, $config)
+	{
+		// Get the database object.
+		$this->db = Factory::getDbo();
```
laoneo: Please use the databaseaware trait.

administrator/components/com_finder/src/Model/ItemModel.php
```diff
@@ -0,0 +1,108 @@
...
+	 * Get a finder link object
+	 *
+	 * @return  object
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getItem()
+	{
+		$link_id = (int) $this->getState('item.link_id');
+		$db = $this->getDbo();
```
laoneo: Use $this->getDatabase()

administrator/components/com_finder/src/Controller/IndexerController.php
Hackwar: Yes, since we can't override the original class and the debug feature has no business in the original class, this is necessary.

administrator/components/com_finder/src/Controller/IndexerController.php
Hackwar: I disagree here. This is not something that can be used in production anywhere and should never be modified. This is debug code, similar to the debug plugin, which doesn't have layouts either.

administrator/components/com_finder/src/Model/ItemModel.php
Hackwar: done

administrator/components/com_finder/src/Indexer/Debugadapter.php
Hackwar: fixed.

administrator/components/com_finder/src/Indexer/Debugadapter.php
```diff
@@ -0,0 +1,976 @@
...
+use Joomla\Utilities\ArrayHelper;
+
+/**
+ * Prototype debug adapter class for the Finder indexer package.
+ * THIS CLASS IS ONLY TO BE USED FOR DEBUGGING PURPOSES! DON'T
+ * USE IT FOR PRODUCTIVE USE!
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+abstract class Adapter extends CMSPlugin
```
nibra: Should be `DebugAdapter` (and the filename `DebugAdapter.php`) to support PSR-4

administrator/components/com_finder/src/Indexer/Debugindexer.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Joomla\Component\Finder\Administrator\Indexer;
+
+\defined('_JEXEC') or die;
+
+/**
+ * Debugging indexer class for the Finder indexer package.
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class Debugindexer extends Indexer
```
nibra: `DebugIndexer` (`DebugIndexer.php`)

administrator/components/com_finder/src/Indexer/Debugadapter.php
Hackwar: This is NOT supposed to be autoloaded. The decisions for the file- and classname were deliberate. Again, it is debug code, supposed to inject itself into existing code without breaking it, not meant for production, but for people to debug their finder plugins.

nibra: Nevertheless should names follow the naming convention. Both files and classes.

chmst:
I am confused. 

I expected that I can select an item in the overview
then click on the debug action
then come to the checker view where the Id and  type of the plugin are already known from the list view
and there is no need for input (can be disabled)

As it is now, I would not know how to find the Id (of course it is in the slug, but mybe users don't know that.

administrator/components/com_finder/src/Indexer/Debugindexer.php
Hackwar: Done.

administrator/components/com_finder/src/Indexer/Debugadapter.php
Hackwar: I changed this and it should be better now.

administrator/components/com_finder/src/Controller/IndexerController.php
Hackwar: Actually, this is not necessary and I fixed it. Thanks.

administrator/components/com_finder/src/Controller/IndexerController.php
Hackwar: As I wrote, this is just for debugging purposes, thus the output here is okay.

Quy:
Leave ID blank or enter a non-existing ID.

carlitorweb:
@Hackwar can you fix the conflicting file here, to give a test

laoneo:
Thank you for your contribution, unfortunately the 4.4-dev branch doesn't get new features. When #40181 is merged this pr should be rebased to 5.0-dev. Thanks for understanding.

HLeithner:
I would mark all debuging functions/class "@internal" since it's not intended to be used by 3rd party developer.

administrator/components/com_finder/src/Controller/IndexerController.php
```diff
@@ -281,4 +268,111 @@ public static function sendResponse($data = null)
         // Send the JSON response.
         echo json_encode($response);
     }
+
+    /**
+     * Method to call a specific indexing plugin and return debug info
+     *
+     * @return  void
+     *
+     * @since   __DEPLOY_VERSION__
```
HLeithner, line 277:
```suggestion
     * @since   __DEPLOY_VERSION__
     * @internal
```

administrator/components/com_finder/src/Indexer/DebugAdapter.php
```diff
@@ -0,0 +1,951 @@
...
+use Joomla\Database\DatabaseInterface;
+use Joomla\Database\QueryInterface;
+use Joomla\Utilities\ArrayHelper;
+
+/**
+ * Prototype debug adapter class for the Finder indexer package.
+ * THIS CLASS IS ONLY TO BE USED FOR DEBUGGING PURPOSES! DON'T
+ * USE IT FOR PRODUCTIVE USE!
+ *
+ * @since  __DEPLOY_VERSION__
```
HLeithner, line 25:
```suggestion
 * @since  __DEPLOY_VERSION__
 * @internal
```

administrator/components/com_finder/src/Indexer/DebugIndexer.php
```diff
@@ -0,0 +1,43 @@
...
+ * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\Component\Finder\Administrator\Indexer;
+
+/**
+ * Debugging indexer class for the Finder indexer package.
+ *
+ * @since  __DEPLOY_VERSION__
```
HLeithner, line 16:
```suggestion
 * @since  __DEPLOY_VERSION__
 * @internal
```

HLeithner (changes requested):
I would like to merge this in to j5 do we already have documentation?

Hackwar:
> I would like to merge this in to j5 do we already have documentation?

No, I need to write the complete documentation for Smart Search still.

HLeithner:
Thanks

PR #36769, opened 2022-01-21
 [4.0] Improvement in frontend

niharikamahajan02:
### Summary of Changes
Did some changes in TOOLBAR.

SHOWING SCREENSHOT OF ARTICLES SECTION.
image

image

SHOWING SCREENSHOT OF CHANGES DONE IN ARTICLE SECTION( CAN BE VIEWED IN OTHER CONTENT CATEGORIES TOO)
image

image

### DIFF
--- a/administrator/templates/atum/scss/system/searchtools/searchtools.scss
+++ b/administrator/templates/atum/scss/system/searchtools/searchtools.scss
@@ -13,16 +13,17 @@
 }
 
 .js-stools-container-bar {
-  margin-inline-start: auto;
+  margin-inline-start: initial;
 
   .btn-toolbar {
-    justify-content: flex-end;
+    justify-content: flex-start;
 
     > * {
       margin: 5px 0;
 
       + * {
         margin-inline-start: 8px;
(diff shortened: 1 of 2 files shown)


brianteeman:
Why? What is the reason behind these changes?

niharikamahajan02:
There is a wide gap before search tool , so it is better to minimize that so that toolbar starts from leftmost side . The screen would look much better.

brianteeman:
Changing from right aligned to left aligned is a subjective opinion not a fact.

richard67:
> The screen would look much better.

@niharikamahajan02 It's not only a question what looks better (and that's a subjective thing, i.e. matter of taste), it is more important which way you can navigate faster and how you can find things more quickly.

Having the search tool bar right aligned like it is without your PR is in my opinion better. We can find the paging and the ordering drop downs more quickly in my opinion.

So I am not in favour of this PR. I think it should be closed.

niharikamahajan02:
Ok sir , thanks for the feedback . Got the point.

PR #36771, opened 2022-01-21
[4.0]Edits in menu 

niharikamahajan02:
### Summary of Changes
For screen size less than 992px , linked modules and id is now being displayed in menu manager.

image

image

### DIFF
--- a/administrator/components/com_menus/tmpl/menus/default.php
+++ b/administrator/components/com_menus/tmpl/menus/default.php
@@ -80,11 +80,11 @@ $wa->useScript('com_menus.admin-menus');
 									<span class="icon-trash" aria-hidden="true"></span>
 									<span class="d-none d-md-inline"><?php echo Text::_('COM_MENUS_HEADING_TRASHED_ITEMS'); ?></span>
 								</th>
-								<th scope="col" class="w-10 text-center d-none d-lg-table-cell">
+								<th scope="col" class="w-10 text-center d-none d-md-table-cell">
 									<span class="icon-cube" aria-hidden="true"></span>
 									<span class="d-none d-md-inline"><?php echo Text::_('COM_MENUS_HEADING_LINKED_MODULES'); ?></span>
 								</th>
-								<th scope="col" class="w-5 d-none d-lg-table-cell text-center">
+								<th scope="col" class="w-5 d-none d-md-table-cell text-center">
 									<?php echo HTMLHelper::_('searchtools.sort', 'JGRID_HEADING_ID', 'a.id', $listDirn, $listOrder); ?>
 								</th>
 							</tr>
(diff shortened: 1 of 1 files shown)


brianteeman:
You can not display everything at every size. This PR reduces the space for the most important part. This is not an improvement.

image

richard67:
@niharikamahajan02 I agree with @brianteeman here.

On the other hand I can understand your intention for this PR because for you the ID and the linked modules maybe more important.

For other people other columns might be more important.

So we can never make everybody happy and have needed a long time to do it like it is now, that there is a common use for the most users.

There is a pull request by @brianteeman which will, when accepted (what I hope somehow) allow to adjust which columns are shown and which not. This would allow each user to fit his or her needs.

Another way to adjust that for a complete site is the special Joomla thing called overrides 😄 .

So I think we should not waste time and power to adjust these column things here or there now.

Thanks for your understanding, and if you agree you could close this PR.

I hope you won't get demotivated by that. I see you have the skills, and when you focus on solving the issue we already have, it would be a great help and I'm optimistic that there will be a good PR from you which would be accepted.

richard67:
P.S.: I forgot the link to the PR I was referring to. It is #36591 .

Scrabble96:
> There is a pull request by @brianteeman which will, when accepted (what I hope somehow) allow to adjust which columns are shown and which not. This would allow each user to fit his or her needs.

Which PR # ?

richard67:
> > There is a pull request by @brianteeman which will, when accepted (what I hope somehow) allow to adjust which columns are shown and which not. This would allow each user to fit his or her needs.
> 
> Which PR # ?

I posted it one comment later, it is #36591 .

brianteeman:
I just remembered that the pr is only for screens >992px

richard67:
> I just remembered that the pr is only for screens >992px

I see: https://github.com/joomla/joomla-cms/pull/36591/files#diff-71a477ee67f0770e640fb06cc047869b190a36c117bf0141b33ce280dfc48d12R187

But I think that doesn't change much about what was said before about the important title not having enough space with this PR applied.

brianteeman:
Yes I just wanted to post that as information only

niharikamahajan02:
> @niharikamahajan02 I agree with @brianteeman here.
> 
> On the other hand I can understand your intention for this PR because for you the ID and the linked modules maybe more important.
> 
> For other people other columns might be more important.
> 
> So we can never make everybody happy and have needed a long time to do it like it is now, that there is a common use for the most users.
> 
> There is a pull request by @brianteeman which will, when accepted (what I hope somehow) allow to adjust which columns are shown and which not. This would allow each user to fit his or her needs.
> 
> Another way to adjust that for a complete site is the special Joomla thing called overrides 😄 .
> 
> So I think we should not waste time and power to adjust these column things here or there now.
> 
> Thanks for your understanding, and if you agree you could close this PR.
> 
> I hope you won't get demotivated by that. I see you have the skills, and when you focus on solving the issue we already have, it would be a great help and I'm optimistic that there will be a good PR from you which would be accepted.

Ok , no problem sir . I got the point , again thanks for the feedbacks and motivation 😃.

RickR2H:
@niharikamahajan02 please test the views with longer titles. Even without the ID and Hits, there are some layout problems between 767px and 992px. We need to come up with a clever way to fix this. Showing the ID and Hits on smaller screen sizes is going to make the problem even worse. FYI if you need the ID and Hits then you can the beet way is to create a user.css in the admin template to show the required fields at you liking. They are still there but hidden.

niharikamahajan02:
ok @RickR2H sir , I will test that and find a soln.

PR #36776, opened 2022-01-21
[4.1] Do not double lock assets

laoneo:
### Summary of Changes

Should fix the system tests random 500 error.

Problem is when on postgres two times a lock is done then it throws the following error:

_There is already an active transaction_

### Testing Instructions
- Create nested categories
- Create nested menu items
- Delete categories
- Delete menu items

Works.

Works.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1135,16 +1135,16 @@
         },
         {
             "name": "joomla/database",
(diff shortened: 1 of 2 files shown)


richard67 (approved):

bembelimen:
Thx

PR #36798, opened 2022-01-23
[3.10] [PHP 8.1] Fixes Router/Route.php deprecated null to string in strpos()

beat:
Pull Request for Issue # none

### Summary of Changes

Fixes `Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in libraries/src/Router/Route.php on line 124`

### Testing Instructions

Source-code review should be enough.

Otherwise on PHP 8.1 with all errors on and Joomla Debug ON, go to frontend in registration page (of CB if needed)

`Deprecated: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in libraries/src/Router/Route.php on line 124`

Warning disappears.

None.

### DIFF
--- a/libraries/src/Router/Route.php
+++ b/libraries/src/Router/Route.php
@@ -121,7 +121,7 @@ class Route
 	public static function link($client, $url, $xhtml = true, $tls = self::TLS_IGNORE, $absolute = false)
 	{
 		// If we cannot process this $url exit early.
-		if (!is_array($url) && (strpos((string) $url, '&') !== 0) && (strpos((string) $url, 'index.php') !== 0))
+		if (is_null($url) || (!is_array($url) && (strpos($url, '&') !== 0) && (strpos($url, 'index.php') !== 0)))
 		{
 			return $url;
 		}
--- a/libraries/src/Router/Route.php
+++ b/libraries/src/Router/Route.php
@@ -121,7 +121,7 @@ class Route
 	public static function link($client, $url, $xhtml = true, $tls = self::TLS_IGNORE, $absolute = false)
 	{
 		// If we cannot process this $url exit early.
-		if (is_null($url) || (!is_array($url) && (strpos($url, '&') !== 0) && (strpos($url, 'index.php') !== 0)))
+		if (!is_array($url) && (strpos($url, '&') !== 0) && (strpos($url, 'index.php') !== 0))
 		{
 			return $url;
 		}
--- a/libraries/src/Router/Route.php
+++ b/libraries/src/Router/Route.php
@@ -121,7 +121,7 @@ class Route
 	public static function link($client, $url, $xhtml = true, $tls = self::TLS_IGNORE, $absolute = false)
 	{
 		// If we cannot process this $url exit early.
-		if (!is_array($url) && (strpos($url, '&') !== 0) && (strpos($url, 'index.php') !== 0))
+		if (!is_array($url) && (strpos((string) $url, '&') !== 0) && (strpos((string) $url, 'index.php') !== 0))
 		{
 			return $url;
 		}
--- a/modules/mod_breadcrumbs/helper.php
+++ b/modules/mod_breadcrumbs/helper.php
@@ -51,7 +51,7 @@ class ModBreadCrumbsHelper
 		{
 			$crumbs[$i]       = new stdClass;
 			$crumbs[$i]->name = stripslashes(htmlspecialchars($items[$i]->name, ENT_COMPAT, 'UTF-8'));
-			$crumbs[$i]->link = JRoute::_($items[$i]->link);
+			$crumbs[$i]->link = !is_null($items[$i]->link) ? JRoute::_($items[$i]->link) : '';
 		}
 
 		if ($params->get('showHome', 1))


alikon:
I have tested this item :white_check_mark: successfully on 121e5cb6a2ca8a80e0279c7dbfbaf9e1edeebba6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36798">issues.joomla.org/tracker/joomla-cms/36798</a>.</sub>

libraries/src/Router/Route.php
```diff
@@ -121,7 +121,7 @@ public static function _($url, $xhtml = true, $tls = self::TLS_IGNORE, $absolute
 	public static function link($client, $url, $xhtml = true, $tls = self::TLS_IGNORE, $absolute = false)
 	{
 		// If we cannot process this $url exit early.
-		if (!is_array($url) && (strpos($url, '&') !== 0) && (strpos($url, 'index.php') !== 0))
+		if (!is_array($url) && (strpos((string) $url, '&') !== 0) && (strpos((string) $url, 'index.php') !== 0))
```
zero-24:
```suggestion
		if (is_null($url) || (!is_array($url) && (strpos($url, '&') !== 0) && (strpos($url, 'index.php') !== 0)))
```

Passing null in the first place does not sound to be a good thing in the first place right? If anything we should not try to hide that issue by casting anything to string right? Anything that is passed should be a string in the first place already by the signature.

zero-24 (commented):
Do you have a stack trace so we know where the original issue is comming from?

beat:
> Do you have a stack trace so we know where the original issue is comming from?

Hi @zero-24,
Sure! Here it is:

image

zero-24:
Hmm can you run a `var_dump($items);` here: https://github.com/joomla/joomla-cms/blob/3.10-dev/modules/mod_breadcrumbs/helper.php#L45

From my understanding the issue looks like that for some reason the `->link` is not set right?

beat:
> Hmm can you run a `var_dump($items);` here: https://github.com/joomla/joomla-cms/blob/3.10-dev/modules/mod_breadcrumbs/helper.php#L45

Here you go:

```
modules/mod_breadcrumbs/helper.php:46:
array (size=2)
  0 => 
    object(stdClass)[525]
      public 'name' => string 'CB Profile' (length=10)
      public 'link' => string 'index.php?option=com_comprofiler&view=userprofile&Itemid=486' (length=60)
  1 => 
    object(stdClass)[499]
      public 'name' => string 'Sign up' (length=7)
      public 'link' => null
```

> From my understanding the issue looks like that for some reason the `->link` is not set right?

Looks like that's correct. But must the last item in a breadcrumb even have a link ?

zero-24:
> Looks like that's correct. But must the last item in a breadcrumb even have a link ?

Interesting question I would say yes as it should be the current page. If anything it should also not be "null" as thats not a valid link in the first place. Looks like we have a option to show or hide the last item. And there we never display the URL even when its set: https://github.com/joomla/joomla-cms/blob/3.10-dev/modules/mod_breadcrumbs/tmpl/default.php#L63-L71 but when speaking abstract about the pathway I would say it makes sense to have the URL of all items of the pathway right?

We could fix it like
```
$crumbs[$i]->link = !is_null($items[$i]->link) ? JRoute::_($items[$i]->link) : '';
```

That would do the trick here but generally speaking i would say we should try to find the reason why the link is null in the first place.

beat:
> We could fix it like
> 
> ```
> $crumbs[$i]->link = !is_null($items[$i]->link) ? JRoute::_($items[$i]->link) : '';
> ```
> 
> That would do the trick here but generally speaking i would say we should try to find the reason why the link is null in the first place.

Agree. But in 3.10 maintenance mode we also want to be as B/C as possible, and I like your suggested fix. Adapted this PR for it (and tested your suggestion before committing it).

zero-24:
Merging thanks :+1:

brianteeman:
> Looks like that's correct. But must the last item in a breadcrumb even have a link ?

That is an option in the module so please be careful changing this!!!

zero-24:
> > Looks like that's correct. But must the last item in a breadcrumb even have a link ?
> 
> That is an option in the module so please be careful changing this!!!

Yes I saw that but as posted above we never add the link to the breadcrumb only the text if anything. Do i miss anything here?

Also we have not changed anything yet we only not process an link when it comes as "null".

beat:
> > Looks like that's correct. But must the last item in a breadcrumb even have a link ?
> 
> That is an option in the module so please be careful changing this!!!

Thank you for sharing your concern. All is ok, no worries:

Nothing changed. The behavior stays exactly same as before, in the functions as well as in the usage.

In a nutshell, the only difference is that PHP <8.1 was accepting NULL as a string parameter in its strpos() function and considering it as an empty string, and now it generates a deprecation warning in that case. So the fix is to return the same result as before in case of NULL (which was an empty string for the function that was calling the function that used strpos() with null).

ZKL2337:
666

PR #36834, opened 2022-01-24
[4.1] [PHP 8.1] Fix PHP 8.1 deprecation notice for Rules Field

beat:
Pull Request for Issue # none

### Summary of Changes

Fixes `Deprecated: Automatic conversion of false to array is deprecated in /home/beat/www/4.0/libraries/src/Form/Field/RulesField.php on line 182`

In cases where an access.xml file isn't present for an extension, in Joomla 3.10 (any PHP version) as well as in Joomla 4.0 up to PHP 8.0.x, the permissions rules field can be used without issues. With PHP 8.1, when the file is not present, `Access::getActionsFromFile()` on line 171 returns false instead of an empty array. When adding elements, the boolean is converted to array at previous line 181, which raises that deprecation notice.

This PR fixes this, converting to array that `$this->actions` property which should anyway always be of type array.

I have discovered this issue testing permissions of a CB add-on, that uses _com_comprofiler.plugin.pluginname_ as asset name. And obviously, there is no access.xml file for that field.

### Testing Instructions

Source-code review should be enough here.

But to test, just go to permissions of Joomla to check everything works as before.

Works.

Still works.

None.


libraries/src/Form/Field/RulesField.php
```diff
@@ -173,6 +173,10 @@ protected function getInput()
 			"/access/section[@name='" . $section . "']/"
 		);
 
+		if ($this->actions === false) {
```
richard67:
```suggestion
		if ($this->actions === false)
		{
```
Code style.

richard67:
@bembelimen Shouldn't this be better made for the 4.1-dev branch meanwhile?

bembelimen:
Yes, please every new PR against 4.1+

beat:
> Yes, please every new PR against 4.1+

Done! Thank you @bembelimen and @richard67 :+1:

RickR2H:
I have tested this item :white_check_mark: successfully on c661001856f9d8c620a6607a3131c21540b45480<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36834">issues.joomla.org/tracker/joomla-cms/36834</a>.</sub>

ghazal:
I have tested this item :white_check_mark: successfully on c661001856f9d8c620a6607a3131c21540b45480<br /><br />Thanks for your clear explanation.
I encountered this Warning and had no idea where it came from, ergo, correct it.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36834">issues.joomla.org/tracker/joomla-cms/36834</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36834">issues.joomla.org/tracker/joomla-cms/36834</a>.</sub>

bembelimen:
Tbh, I see what you did there, but for me are more the questions:
- why is it ```false``` at all? if we ignore it anyways, probably the initial value should be an empty array?
- could it have more values than ```false```? For example a string? probably a ```!is_array``` check would be better?

beat:
> Tbh, I see what you did there, but for me are more the questions:
> 
>     * why is it `false` at all? if we ignore it anyways, probably the initial value should be an empty array?

The function `Access::getActionsFromFile()` returns `false` when the file doesn't exist, and probably empty array `[]` if it exists but has no records.

I didn't want to change the API, so didn't touch that function.

> 
>     * could it have more values than `false`? For example a string? probably a `!is_array` check would be better?

The function `Access::getActionsFromFile()` is documented to return either an array or boolean `false`, nothing else.

My best guess is that Joomla 5 will want to go into the direction of strict typing, like PHP 8.1 is setting this good design direction, and not add more mixed *return* types. And without adding a different functionality to that function, I don't see it adding strings in the future without breaking the single-concern principle. So the probability that we want to return suddenly a string from that function would look crazy to me. Now if it would be a third-party library in which we have no confidence on design (and be crazy enough to use it) or no control, I would be more prudent and would have done an `!is_array()`.

If you think that Joomla 5 might go the opposite route and extend types (breaking B/C on that function btw) while whole PHP world is going strict-types, and you prefer `!is_array()` instead, it's your decision and I will adapt the PR.

Please let me know, my goal is to fix this PHP 8.1 bug :-)

bembelimen:
Good enough for now, thanks.

blondie63:
Hi, on my site joomla 4.2.8, if i use PHP > 8.0 i see these msgs on frontend:

Deprecated: Automatic conversion of false to array is deprecated in /home/cavi22/webapps/cavinato2022/libraries/src/HTML/HTMLHelper.php on line 790

Deprecated: Automatic conversion of false to array is deprecated in /home/cavi22/webapps/cavinato2022/libraries/src/Document/Document.php on line 505

Some help ?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36834">issues.joomla.org/tracker/joomla-cms/36834</a>.</sub>

PR #36231, opened 2021-12-06
[4.1] Add server timing response header in debug mode

laoneo:
### Summary of Changes
Browsers do support a [server timing](https://www.w3.org/TR/server-timing/#the-server-timing-header-field.) header field which adds information what happens on the server side. This basically gives the developer insight what happens in the "Time to First Byte".
This pr adds the missing header in the debug plugin, as this header should not be added when in production mode because of performance reasons.

Would be nice to get some feedback about spicing up the values of the header as currently I'm just logging them as they are.

The view looks like in the following picture:
image

### Testing Instructions
- When testing in Firefox, make sure you open the page with HTTPS
- Put joomla in debug mode
- In the debug plugin with make sure that the Profile collector is enabled (is by default)
- Open any page
- Open the dev tools and check the Network tab
- Select the first entry and then click on the timings tab of the specific request

### Actual result BEFORE applying this Pull Request
Only the request timing information is shown.

### Expected result AFTER applying this Pull Request
Beside the request timing is the server timing information shown.

PR #36840, opened 2022-01-25
Revert "[4.1] Add server timing response header in debug mode (#36231)"

bembelimen:
This reverts commit 360409f51cd81de6ca302ee3a51e718f26d3fa7b.

### Summary of Changes

Sorry, was to fast in merging. We should for new events only go with the new event system. So probably @laoneo you could just do the same PR with the new event system?

Sorry for the trouble.

### DIFF
--- a/libraries/src/Application/CMSApplication.php
+++ b/libraries/src/Application/CMSApplication.php
@@ -311,14 +311,11 @@ abstract class CMSApplication extends WebApplication implements ContainerAwareIn
 			ExceptionHandler::handleException($event->getError());
 		}
 
-		// Trigger the onBeforeRespond event.
-		$this->getDispatcher()->dispatch('onBeforeRespond');
-
 		// Send the application response.
 		$this->respond();
 
 		// Trigger the onAfterRespond event.
-		$this->getDispatcher()->dispatch('onAfterRespond');
+		$this->triggerEvent('onAfterRespond');
 	}
 
 	/**
--- a/plugins/system/debug/debug.php
+++ b/plugins/system/debug/debug.php
@@ -19,7 +19,6 @@ use Joomla\CMS\Document\HtmlDocument;
 use Joomla\CMS\Log\Log;
 use Joomla\CMS\Log\LogEntry;
 use Joomla\CMS\Plugin\CMSPlugin;
-use Joomla\CMS\Profiler\Profiler;
 use Joomla\CMS\Session\Session;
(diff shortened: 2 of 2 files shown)


Fedik:
Also can just change `onBeforeRespond` to `onAfterRender`, that triggered almost before response.

laoneo:
I do not agree. It uses the dispatcher which is new way. So should an empty event being introduced here or what?

Fedik:
> So should an empty event being introduced here or what?

We have RFC for versioning https://github.com/joomla/rfc/pull/29 (though it still RFC)

> Changes that do not change signatures of functions or methods of the public API and do not add new ones can go
into a patch release.

New Event add a new public API, so it a new feature :wink: 
And new feature should not go in RC.

laoneo:
@Fedik please read the sentence again. This is not a signature change! And it is a about patch releases and does not mention in any way RC.

Fedik:
> This is not a signature change!

Correct, but a new public API.

laoneo:
> Also can just change `onBeforeRespond` to `onAfterRender`, that triggered almost before response.

Actually I do not care which event is used. If there is an existing one, then this pr can change to it and revert the "new" event. But reverting a whole pr which adds some debug functionality is just ridiculous.

laoneo:
> > This is not a signature change!
> 
> Correct, but a new public API.

Then make a pr which uses your suggested event and all is good. I will immediately test it.

HLeithner:
Actually the original PR fixes a missing event (`onBeforeRespond`) which already exists in the abstracted class `WebApplication` (so it's not needed by allon to create the right event him self) but is missing in all abstracted and overridden classes. Beside this the event is also not triggered in the caching plugin and so would not add performance matrix when the page comes from the cache. 

So I revert my opinion on reverting this PR since the event already exists, only the PR doesn't do what would be expected if the page is full cached because of the missing `onBeforeRespond` in the caching plugin.

laoneo:
Keep in mind, this is only something which should work in debug mode and hopefully the caching plugin is not running when debugging.

HLeithner:
> Keep in mind, this is only something which should work in debug mode and hopefully the caching plugin is not running when debugging.

actually it does, looking at the code it also triggers the `onAfterResponse` event only in debugmode:
https://github.com/joomla/joomla-cms/blob/4.1-dev/plugins/system/cache/cache.php#L147-L155

laoneo:
Wouldn't expect that...

Issue #34952, opened 2021-07-28
PHP 8.1 deprecations


HLeithner:
PHP deprecations for 8.1 has been implemented and this features are planed to be removed in 9.0

https://wiki.php.net/rfc/deprecations_php_8_1

Special attention should go to the removal of  `strftime`

If this is relevant for 3.x is open for discussion, since 3.10 will be EOL before php 8.0 is EOL and at this point of time all Joomla Installation should be upgraded to 4.x (I know that this is more a wish then a fact).

For 4.x series I would say starting with 4.1 it's relevant.

PhilETaylor:
https://github.com/joomla/joomla-cms/issues/34851

PhilETaylor:
Joomal 3.10.1
PHP 8.1-RC2

```

Deprecated: Joomla\Input\Input implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary) in /Users/phil/Sites/Joomla_3.10.1-Stable-Full_Package/libraries/vendor/joomla/input/src/Input.php on line 41

Deprecated: Return type of Joomla\Input\Input::count() should either be compatible with Countable::count(): int, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /Users/phil/Sites/Joomla_3.10.1-Stable-Full_Package/libraries/vendor/joomla/input/src/Input.php on line 170

Deprecated: Joomla\CMS\Input\Input implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary) in /Users/phil/Sites/Joomla_3.10.1-Stable-Full_Package/libraries/src/Input/Input.php on line 31

Deprecated: Return type of Joomla\Registry\Registry::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /Users/phil/Sites/Joomla_3.10.1-Stable-Full_Package/libraries/vendor/joomla/registry/src/Registry.php on line 126

[...]

PR #36941, opened 2022-02-04
 PHP 8.1 deprecation: strtolower()

technik-kvz:
Problem like: #34952

This should fix the message:

Deprecated: strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in [..]/BaseController.php on line 712

Pull Request for Issue # .

### Testing Instructions

Check for set and not null before use, fallback empty string.

php 8.1 warning

Be ready for future php versions.


wilsonge:
This should be fixed by always injecting a task into the module - not by allowing null as a valid option as previously documented. Closing this.

Issue #28852, opened 2020-04-28
[4.0] No possibility to add a custom path in media field


coolcat-creations:
### Steps to reproduce the issue
If you want to define an own path in the media field it's not possible. 
It's not possible in the field itself which would increase speed of workflow for administrators and it's also not possible in the media manager to add a custom path, at least I can't find it anywhere...

![grafik](https://user-images.githubusercontent.com/828371/80458589-4e670a80-8931-11ea-9f22-838537f4b876.png)

### Expected result
Input should be allowed

### Actual result
Input is disabled
No possibility to add a path

brianteeman:
> It's not possible in the field itself which would increase speed of workflow for administrators 

its also not possible in j3 (thats an observation not an explanation or excuse)

>  it's also not possible in the media manager to add a custom path,

You can add as many paths as you want now in joomla 4 using the FileSystem - Local plugin (again not saying its correct just explaining how)

coolcat-creations:
> > It's not possible in the field itself which would increase speed of workflow for administrators
> 
> its also not possible in j3 (thats an observation not an explanation or excuse)

In some places it is possible in Joomla 3 (Article Intro Image for example) 
![grafik](https://user-images.githubusercontent.com/828371/80463401-7c038200-8938-11ea-8a76-1ece1294eedb.png)


> 
> > it's also not possible in the media manager to add a custom path,
> 
> You can add as many paths as you want now in joomla 4 using the FileSystem - Local plugin (again not saying its correct just explaining how)

For example to define a custom path in the template style settings in the image field - where can I input the custom path? Can you lead me to it?

brianteeman:
I do not have that option in my joomla 3 article intro image
![image](https://user-images.githubusercontent.com/1296369/80464107-31cad280-8931-11ea-80ee-518656d556c9.png)

coolcat-creations:
Oh sorry it's an enhancement by JCE MM ... I am already so used to it...
It's really useful to be able to add the path...

But still I don't know where in the new MM Modal can you add the custom path I did not find it? The old Media Manager (not JCE) has a path input in the modal...

brianteeman:
> You can add as many paths as you want now in joomla 4 using the **FileSystem - Loca**l plugin (again not saying its correct just explaining how)

ReLater:
> It's really useful to be able to add the path...

and copy/paste paths or change them quickly and so on...

coolcat-creations:
[...]

PR #34177, opened 2021-05-24
[4.0] Figure out the image size for external adapters

dgrammatiko:
Pull Request is supplamentary to https://github.com/joomla/joomla-cms/pull/33724

PR #31398, opened 2020-11-14
[4.0] Add possibility to add a custom path in media field

Razzo1987:
Pull Request for Issue #28852 .

### Summary of Changes
Enable the media field for input custom url 

### Testing Instructions
Enable the patch, try to create an article and set a custom URL in the Intro Image, for example `https://via.placeholder.com/350x150`

### Actual result BEFORE applying this Pull Request

You can not insert custom URL:
immagine

### Expected result AFTER applying this Pull Request
You can insert custom URL and after save the preview is displayed:
immagine

### Documentation Changes Required
NO

PR #36945, opened 2022-02-06
[4.1] Remove readonly on media fields

brianteeman:
PR for #28852 

The field was readonly for the reasons states by @dgrammatiko and I respect those views. But the request for the ability to paste an image url directly into the field by @coolcat-creations and others also makes sense.

As you can see its a super simple PR to achieve this. All the features of the media field are maintained and you can paste an image url directly. However when pasting a url you do not get the data attributes and lazy loading. I believe this to be an acceptable solution.

To test try to use the full article image or a custom media field and make sure you can still select an image and that you can now paste an image url

### DIFF
--- a/layouts/joomla/form/field/media.php
+++ b/layouts/joomla/form/field/media.php
@@ -181,7 +181,7 @@ if (count($doc->getScriptOptions('media-picker')) === 0) {
 		</div>
 	<?php endif; ?>
 	<div class="input-group">
-		<input type="text" name="<?php echo $name; ?>" id="<?php echo $id; ?>" value="<?php echo htmlspecialchars($value, ENT_COMPAT, 'UTF-8'); ?>" readonly="readonly" <?php echo $attr; ?>>
+		<input type="text" name="<?php echo $name; ?>" id="<?php echo $id; ?>" value="<?php echo htmlspecialchars($value, ENT_COMPAT, 'UTF-8'); ?>" <?php echo $attr; ?>>
 		<?php if ($disabled != true) : ?>
 			<button type="button" class="btn btn-success button-select"><?php echo Text::_('JLIB_FORM_BUTTON_SELECT'); ?></button>
 			<button type="button" class="btn btn-danger button-clear"><span class="icon-times" aria-hidden="true"></span><span class="visually-hidden"><?php echo Text::_('JLIB_FORM_BUTTON_CLEAR'); ?></span></button>


coolcat-creations:
Hi @brianteeman thank you so much for creating the PR!

I thought the right solution would be to "validate the URL, get the image adapter, get the image dimensions" I see it coming that people complain that lazy load does not work.

So how can we make a good solution to 1) make the paths "pasteable" 2) have it working like intended by media manager team ?

Or would this be the first step, and maybe in 4.2 we can create a pseudocron which checks the image paths and adds this markup ?

brianteeman:
> I thought the right solution would be to "validate the URL, get the image adapter, get the image dimensions" I see it coming that people complain that lazy load does not work.

That's kind of what Dimitiris was telling you. If you want to paste the image then you cant have the other parts when you do

coolcat-creations:
Yes, maybe I misunderstood him, I thought it was possible to do these checks for pasted paths but that he does not want to do it so he suggested a bad solution over the right solution. Did I misunderstood?

dgrammatiko:
Fwiw the code for getting the size of the image already exists since https://github.com/joomla/joomla-cms/pull/34177 so what is missing is a way to:
- validate that what the user inserted is a valid url
- Extract the media adapter if there is one

just Because it’s easy to remove an attribute it doesn’t mean that the result will be automatically acceptable. For me it’s not, the media field goes from a field that always produces valid urls to accepting any kind of strings…

coolcat-creations:
Why isn't it possible to let the user put in the path and have a button which says "validate and save" or something?

dgrammatiko:
> Why isn't it possible to let the user put in the path and have a button which says "validate and save" or something?

That's not how form fields validation works in HTML. You have an event `onInput` that you can listen for inserted text and act upon it, there's no need for button or any other fancy UI...

coolcat-creations:
So what is the right approach ? Where do we need to change what line?

Razzo1987:
@brianteeman the preview dosn't work with image generated like https://via.placeholder.com/350x150

Razzo1987:
Hi @dgrammatiko ,
I perfectly understand your safety point.
In the PR https://github.com/joomla/joomla-cms/pull/31398 (which is not very different from @brianteeman 's) we had already discussed that it was necessary to reintroduce the functionality of being able to put custom URLs after their validation, but that at the moment NO ONE was working on it.

Today, the situation does not seem to have changed. Here (https://github.com/joomla/joomla-cms/issues/28852#issuecomment-692972999) you were also talking about integration with third-party sources. Another beautiful idea that I don't see it in the roeadmap for Joomla! 4.1 or 4.2.

Personally I think that for now the best way is to allow users to enter the custom URL as they did in Joomla 3, when we have a media manager who does the necessary checks and inserts images from Dropbox, AWS, S3, custom, etc we can again put back the read-only attribute.

dgrammatiko:
I already stated what needs to be done here https://github.com/joomla/joomla-cms/pull/36945#issuecomment-1030826951 but then again I'm not a gatekeeper or some kind of authority here, so if allowing any string in this field value is acceptable then this PR is fine

coolcat-creations:
Yes thanks but for me you could have write it in Japanese and i would not know what to do now :-) Maybe someone knows. I hope.

RickR2H:
I have tested this item :white_check_mark: successfully on 233ad781b21a04c2477d77caba3054ea3c4db968<br /><br />Patch works! Maybe it is a good idea to check if the image link is valid. Can be added in a new PR though.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

kiki-G:
I have tested this item :red_circle: unsuccessfully on 233ad781b21a04c2477d77caba3054ea3c4db968<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

kiki-G:
I have tested this item :white_check_mark: successfully on 233ad781b21a04c2477d77caba3054ea3c4db968<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

kiki-G:
I have tested this item :red_circle: unsuccessfully on 233ad781b21a04c2477d77caba3054ea3c4db968<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

kiki-G:
I have tested this item :white_check_mark: successfully on 233ad781b21a04c2477d77caba3054ea3c4db968<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

kiki-G:
I have tested this item :white_check_mark: successfully on 233ad781b21a04c2477d77caba3054ea3c4db968<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

kiki-G:
I have tested this item :white_check_mark: successfully on 233ad781b21a04c2477d77caba3054ea3c4db968<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

RickR2H:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/36945">issues.joomla.org/tracker/joomla-cms/36945</a>.</sub>

dgrammatiko:
@RickR2H @brianteeman I object here. What this PR is doing is WRONG! ignoring that there's no validation and you can end up with FOOBAR values and 404 in your pages and the fact that this essentially bypasses the lazy loading feature there's something even more hurtful: the images will NOT have width and height attributes meaning that these images will contribute negatively in the Layout shifting metrics and thus hurting SEO!. Just let the people that want this functionality to create a layout override and themselves remove the readonly attribute OR find someone to write the few lines of JS needed so this could not have any MAJOR impact in performance and SEO.

Web Archive 2022 is in the way and afaik Joomla is not helping itself if keeps the last position in the ranking.

My 2c and unsubscribing

coolcat-creations:
It would be really good to have the possibility to paste the path AND have width height - User Experience for selecting images is zero right now.

richard67:
Setting the RLDQ (release leads decision queue) label to get a decision.

dgrammatiko:
> User Experience for selecting images is zero right now.

You're not improving things by just making them worst, do you? I, keep repeating that this COULD be done reasonably with some JS but not just removing the readonly attribute simply because the results are not the same. Either fix it properly or keep the bad UX (according to you) which at least was done to improve SEO and perf...

roland-d:
As for "security" anybody with some wit can remove the readonly attribute using DevTools from their browser. Validation should always happen server side.

coolcat-creations:
> > User Experience for selecting images is zero right now.
> 
> 
> 
> You're not improving things by just making them worst, do you? I, keep repeating that this COULD be done reasonably with some JS but not just removing the readonly attribute simply because the results are not the same. Either fix it properly or keep the bad UX (according to you) which at least was done to improve SEO and perf...

If it's so easy to improve it why won't you help? I can't because I don't know how.

brianteeman:
> If it's so easy to improve it why won't you help? I can't because I don't know how.

i tried

coolcat-creations:
@dgrammatiko how much would it cost to fix it in a proper way?

dgrammatiko:
Please close this and test https://github.com/joomla/joomla-cms/pull/37654

brianteeman:
Happy to

PR #36977, opened 2022-02-09
[4.1] TinyMCE Update 5.10.3

brianteeman:
### Fixed
Alignment would sometimes be removed on parent elements when changing alignment on certain inline nodes, such as images.
The fullscreen plugin would reset the scroll position when exiting fullscreen mode.


bembelimen:
Thx

brianteeman:
thanks

Issue #36978, opened 2022-02-09
[3.10.*] Add a plugin event to change the SQL query to get a list of records


motokraft:
### Is your feature request related to a problem? Please describe.
In the [**ListModel**](https://github.com/joomla/joomla-cms/blob/eb70adcbe43425c557d8b7b05d847819f22b50b4/libraries/src/MVC/Model/ListModel.php) model class, there is a [**getListQuery**](https://github.com/joomla/joomla-cms/blob/eb70adcbe43425c557d8b7b05d847819f22b50b4/libraries/src/MVC/Model/ListModel.php#L213) function, I want to add a plugin event so that I can change the SQL query before execution.

### Describe the solution you'd like
Add a plugin event to the [**getItems**](https://github.com/joomla/joomla-cms/blob/eb70adcbe43425c557d8b7b05d847819f22b50b4/libraries/src/MVC/Model/ListModel.php#L180) function so that you can change the SQL query before execution

### Additional context
None

prakhar3062:
Are you woking on it now

motokraft:
> Are you woking on it now

Yes. I want to make a **pull request**

Quy:
PR #36979

PR #36979, opened 2022-02-09
Added a plugin event to change the SQL query of the data list in the "ListModel" class

motokraft:
Pull Request for Issue #36978.

### Summary of Changes
Add a plugin event to the **ListModel::getItems** function before executing the SQL query

### Testing Instructions
None

It is not possible to change the SQL query before execution

Add plugin event to change SQL query before execution

None

### DIFF
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -13,6 +13,7 @@ defined('JPATH_PLATFORM') or die;
 use Joomla\CMS\MVC\Factory\LegacyFactory;
 use Joomla\CMS\MVC\Factory\MVCFactoryInterface;
 use Joomla\Utilities\ArrayHelper;
+use Joomla\CMS\Plugin\PluginHelper;
 
 /**
  * Base class for a database aware Joomla Model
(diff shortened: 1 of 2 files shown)


Hackwar:
This would qualify as a new feature and there will be no new features in the 3.x branch. Besides that, we had this feature request several times before and they were always denied. Your best bet is to override the model in a plugin and make your changes directly.

motokraft:
> This would qualify as a new feature and there will be no new features in the 3.x branch. Besides that, we had this feature request several times before and they were always denied. Your best bet is to override the model in a plugin and make your changes directly.

Yes, what are you talking about! I did a merge for the first time... Using your version, what kind of shitty code will turn out then! I want it to be simple for the user and fast for the server, and you suggest writing shitty code ....

motokraft:
> Your best bet is to override the model in a plugin and make your changes directly.

I want to do the best so as not to fence any garbage

Hackwar:
You wont get very far if you insult people. I was trying to propose to you something that you can do right now to get what you want, but regardless of that, my first comment still stands:

Joomla 3.x is EOL and only receives bugfixes and security fixes. This here is neither and instead is a new feature. It can NOT be accepted for 3.10. If you really think that this should be implemented, then you would have to do this PR for Joomla 4, specifically for at least Joomla 4.2.

But as I also wrote, this has been proposed several times before and was always rejected.

motokraft:
> ou wont get very far if you insult people.

I didn't mean to offend you in any way, it's just how Google translate goes

Issue #36988, opened 2022-02-10
Joomla 3.10.x - Administrator Mail is sent also to non-admins


coolcat-creations:
### Steps to reproduce the issue
With this setting all Users who are registered, and have the "Receive System Mails" Flag on "Yes" (1) will receive an email notification that a new User registered on the Website

The Email which is sent out contains even the Greeting "Hello Administrator" 

<img width="530" alt="grafik" src="https://user-images.githubusercontent.com/828371/153384644-0152871c-adae-4a1c-be13-94c123aff08c.png">

### Expected result
When I set "Send Mail to Administrators" to yes, I expect that only users with Administration (ACL) rights will receive a mail.

### Actual result
All users who are flagged to receive system emails receive an email, greeting them as Administrator and informed about sensitive data (User Email Address and Name)

### System information (as much as possible)
Joomla 3.10.5

### Additional comments
1) Either the field **Send Mail to Administrators** has to be renamed to Send System Messages to all enabled users (or similar) or 
2) There needs to be an additional check to only send out this information only to the Administrator level.

blueforce:
This is also the case in version 4.x
![Bildschirmfoto 2022-02-10 um 11 32 45](https://user-images.githubusercontent.com/1141234/153389159-f2ef5de6-a57e-4e20-bd30-a90ac4e06159.png)

brianteeman:
>1. Either the field Send Mail to Administrators has to be renamed to Send System Messages to all enabled users (or similar) or
>2. There needs to be an additional check to only send out this information only to the Administrator level.

For me option 1 is just covering up the problem. 

As these emails always/often include links that require access to the admin interface then they should not be sent to a user who does not have the required acl access irrespective of the setting for this checkbox.

Quy:
Closing per PR #36997.

PR #36989, opened 2022-02-10
PR for #36988 Update en-GB.com_users.ini

coolcat-creations:
This PR will change the label and description of the field for Administrator mails for the user settings.

Before: 

grafik

I changed it to: 

COM_USERS_CONFIG_FIELD_MAILTOADMIN_DESC="If set to Yes then a notification mail will be sent to all users who have receive system mails enabled if 'New User Account Activation' is set to 'None' or 'Self'."
COM_USERS_CONFIG_FIELD_MAILTOADMIN_LABEL="Send System-Mail"

Pull Request for Issue #36988 .

### Testing Instructions
Install the patch and see that the labels are changed.

grafik

grafik

Yes it should be clarified that not only Administrators can get the mail


brianteeman:
Wouldn't it be a better fix to prevent system emails being sent to anyone who does not have admin interface access as the content of the system emails regularly contains links which require admin interface access.

For me this is just covering up the real issue and not actually solving it.

coolcat-creations:
Yes, my skills were only able to solve one of the possibilities. But I think also that it would be good to only send this mails to users who have user administration admin rights. I see this as a huge GDPR issue for the website owners when they made this setting by mistake.

coolcat-creations:
Another Possibilty would be to name System Mails to "Receive Administrator Mails"

brianteeman:
> Another Possibilty would be to name System Mails to "Receive Administrator Mails"

That is still just covering up the issue

coolcat-creations:
Maybe we can continue here: https://github.com/joomla/joomla-cms/issues/36988

brianteeman:
Found the bug will do a PR later

coolcat-creations:
You found a bug regarding this?
I think this database query is wrong but what to add to only get the admins that are allowed to controll users?
https://github.com/joomla/joomla-cms/blob/f412e135298643bab6255fe498db69a5c8a91dbb/components/com_users/src/Model/RegistrationModel.php#L155

brianteeman:
please test #36993

PR #37019, opened 2022-02-13
Update README.md for alternate process using Docker for git and build

beat:
This is NOT urgent for 4.1.0 release.

And it looks like it requires some instructions adaptation for Windows.

Pull Request for Issue # none

### Summary of Changes

Adds alternate instructions using Docker Engine instead of installing git, php-cli, composer and Node.js+NPM.

I had issues installing Node.js into a test-environment to run very latest code for 4.1 tests, so I used Docker to run npm it quickly without installing it. Then I thought that an alternate run from sources using just Docker Engine instead of installing git, php-cli, composer and Node.js would be quite cool and modern-looking. While getting the exact docker command-lines can be and was a bit tricky, once they are known, this should run on any operating system.

These alternate instructions to run from sources could lower a bit one of the barriers to entry to contributors.

### Testing Instructions

1. Proofread English of changes (I'm not English-native).
2. Try it out:

- Install Docker Engine (or Docker Desktop) using the method of the URL provided. 
- Try out the alternate instructions, then 
- `ls -al` inside joomla-cms to check that the Joomla folders are there
- `ls -al libraries/vendor/` should show around 50 folders with your own ownership/permissions
- `ls -al media/vendor/debugbar/` should show around 13 files and 2 folders with your own ownership/permissions
- move joomla-cms folder inside your website directory and change folder permissions so that it can run in your enviroment
- try running joomla from that folder, it should provide the installation screen, and run fine.
- Provide feedback of the operating systems you tried and your test results.

Ideally, we need tests in MacOS and Windows too. I tested it successfully on Linux (latest Ubuntu LTS (20.04))

No documented way to install from sources without installing php-cli, composer and Node.js with npm software natively and permanently into your system.

1. Documented way to install from sources without installing any software (appart from Docker Engine) in the README.md.

2. The documented Docker-assisted installation from source works.

None needed, This, README.md, *is* just the documentation.


brianteeman:
> gives the git clone https:// path too for non-comitters

I dont understand this. From my understanding there is no significant difference

beat:
> > gives the git clone https:// path too for non-comitters
> 
> I dont understand this. From my understanding there is no significant difference

Thank you @brianteeman for noting that my explanation proposal in the README.md wasn't clear enough regarding that aspect! I have now updated the explanation as follows:

> (https read access is public, ssh access [requires a Github account with an active SSH key](https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories))

Does that make more sense why it's good to also show the https variant to lower the bar to access (and avoiding people to have to create and confirm a github account, then generate a SSH keypair, upload the public key) ?

The link above explains in details the subtilities of https vs ssh github access.

The https access makes a lot of sense for public access, and that's what I use for the Docker-variant.

The SSH access to joomla-cms is slightly better technically than https, but only useful for joomla-cms committers to avoid entering their username and password at each direct commit to the joomla/joomla-cms repository.

Btw, if we had to show only one git clone command, it would be the https one, as the "pros" know that they can also use ssh and how to use it.

Does that answer your very valid question ? and is the proposed text clear ?

brianteeman:
> The SSH access to joomla-cms is slightly better technically than https, but only useful for joomla-cms committers to avoid entering their username and password at each direct commit to the joomla/joomla-cms repository.

As thats only a very few people who can do that - its overcomplicating the doc

beat:
> > The SSH access to joomla-cms is slightly better technically than https, but only useful for joomla-cms committers to avoid entering their username and password at each direct commit to the joomla/joomla-cms repository.
> 
> As thats only a very few people who can do that - its overcomplicating the doc

Does that mean that you are suggesting to just replace the documented ssh access with simply the public https access ? I'm fine with that and think that would make full sense.

EDIT: I went ahead to just show the https github cloning access, as it makes full sense.

brianteeman:
> No documented way to install from sources without installing 1 Gigabyte of software natively and permanently into your system.

I think you missed something here.  A clean docker install on windows is 2.5gb

image

brianteeman:
The first command says

> docker run -ti --rm --name my-joomla-git --volume "${HOME}":/root --volume "$PWD":/usr/src/app --user $(id -u):$(id -g) --workdir /usr/src/app alpine/git clone https://github.com/joomla/joomla-cms.git

It just produced

> unknown shorthand flag: 'g' in -g)

brianteeman:
Honestly (even if corrected to be working) I think this is for the docs site and not for the readme.

brianteeman:
and did you know there is an official joomla docker image?
https://hub.docker.com/_/joomla
https://github.com/joomla-docker/docker-joomla

beat:
> I think you missed something here. A clean docker install on windows is 2.5gb

On Windows, Docker Engine comes with Docker Desktop by default, so it's much heavier. Forget my size argument, installation convenience is key here.

> The first command says
> 
> > docker run -ti --rm --name my-joomla-git --volume "${HOME}":/root --volume "$PWD":/usr/src/app --user $(id -u):$(id -g) --workdir /usr/src/app alpine/git clone https://github.com/joomla/joomla-cms.git
> 
> It just produced
> 
> > unknown shorthand flag: 'g' in -g)

And it looks like the `--user $(id -u):$(id -g)` part of the commands should be left out on windows.

1-2 GB size being put appart, installing Docker Engine is so much easier than installing Node.js, which comes with tons of libraries and package manager conflicts.

> and did you know there is an official joomla docker image? https://hub.docker.com/_/joomla https://github.com/joomla-docker/docker-joomla

Yes, I know, and I use it for stable releases testing environments. This here is different: It allows you to get a working source installation with a git repository.

brianteeman:
> And it looks like the --user $(id -u):$(id -g) part of the commands should be left out on windows.

OK I will continue testing that

> installing Docker Engine is so much easier than installing Node.js, which comes with tons of libraries and package manager conflicts.

I dont remember that happening at all. But maybe I forgot.

brianteeman:
Sorry I give up. Docker Desktop just says stopped and nothing will start it.

Tried the first command without the user part and got
> docker: error during connect: This error may indicate that the docker daemon is not running.: Post "http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.24/containers/create?name=my-joomla-git": open //./pipe/docker_engine: The system cannot find the file specified.

Tried it from powershell instead of cmd.exe and got 
> docker: invalid reference format

beat:
> Sorry I give up. Docker Desktop just says stopped and nothing will start it.
> 
> Tried the first command without the user part and got
> 
> > docker: error during connect: This error may indicate that the docker daemon is not running.: Post "http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.24/containers/create?name=my-joomla-git": open //./pipe/docker_engine: The system cannot find the file specified.
> 
> Tried it from powershell instead of cmd.exe and got
> 
> > docker: invalid reference format

Obviously, on your system your Docker Engine is not running or not installed properly (e.g. following instructions pointed to in this PR's README.md) to run.

This PR being not urgent compared to 4.1 release tasks, we can postpone your tests. I have separated the git clone change proposal from ssh to https into a separate PR #37021 and removing that part from this PR.

As instructions seem to need adaptations for Windows, marking as Draft for now.

brianteeman:
> Obviously, on your system your Docker Engine is not running or not installed properly (e.g. following instructions pointed to in this PR's README.md) to run.

It is installed according to the instructions from docker on the link you provided. From checking their forums it is a common problem but without a common solution. Guess docker wasnt that easy after all.

HLeithner:

Quy:
Closing as it appears to  be abandoned.

Issue #37047, opened 2022-02-15
please fix wrong label "page class" in menu - irritating


dozentis:
### Steps to reproduce the issue
- open a menu item, type like _list of categories_
- move to tab _Page Display_
- type in your _Page Class_ and save 

### Actual result in HTML source code of page
`<div class="categories-listyourcssclassname list-group">`

### Additional comments
Page Class is a **Suffix** to the class _categories-list_ (like in module/advanced)

## My recommendation
Please change label of **Page Class** in _Menus: Edit Item_ to **Page Class Suffix**

chmst:
Which Joomla version?

brianteeman:
It's a bug!

It was changed to be a page class and not a suffix iirc and in the body tag its a page class 
eg `<body class="site com_content wrapper-static view-category no-layout no-task itemid-119 banana has-sidebar-right">`

But on the category list it is also being used (**_not sure if it is supposed to be as its not in category blog_**)
eg
```
<div class="com-content-category category-list">
<div class="content-categorybanana">
<h1>Category List</h1>
```

brianteeman:
The "bug" is in C:\htdocs\joomla-cms\layouts\joomla\content\category_default.php

dozentis:
Thanks for reply. 
So I see it has already been revised with Joomla 4 (3.x?), so we no longer have a suffix specification like formely in Modules/Advances?

I've checked some menu types.

Category blog  -> **issue**
`<div class="blogmycssclass">`

single article -> no issue
`<div class="article-details mycssclass" itemscope itemtype="https://schema.org/Article">`

Category list -> **issue**
`<div class="category-listmycssclass">`

List all tags  -> **issue**
`<div class="tag-categorymycssclass">`

I'll check some more on my demo site




<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37047">issues.joomla.org/tracker/joomla-cms/37047</a>.</sub>

chmst:
Thanks for reporting! Yes it should not longer be a suffix but a class. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37047">issues.joomla.org/tracker/joomla-cms/37047</a>.</sub>

dozentis:
List all Tags -> **issue**
[...]

PR #37055, opened 2022-02-15
[4.1] page class not suffix

brianteeman:
Pull Request for Issue #37047 .

The option to set a page class on a menu item sets a class on the body tag. But in some places it is also used on the container div of the component. As it is a class and not a suffix now (despite the name which wasnt changed for bc) this PR ensures that it is used as a class and not a suffix. ie there is a space before the page class.

PS not sure if this page class should even be used here so an alternative would be to remove it from these files

### DIFF
--- a/components/com_contact/tmpl/form/edit.php
+++ b/components/com_contact/tmpl/form/edit.php
@@ -22,7 +22,7 @@ $this->tab_name         = 'com-contact-form';
 $this->ignore_fieldsets = array('details', 'item_associations', 'language');
 $this->useCoreUI        = true;
 ?>
-<div class="edit item-page <?php echo $this->pageclass_sfx; ?>">
+<div class="edit item-page<?php echo $this->pageclass_sfx ? ' ' . $this->pageclass_sfx : ''; ?>">
 	<?php if ($this->params->get('show_page_heading')) : ?>
 		<div class="page-header">
 			<h1>
--- a/components/com_contact/tmpl/form/edit.php
+++ b/components/com_contact/tmpl/form/edit.php
@@ -22,7 +22,7 @@ $this->tab_name         = 'com-contact-form';
 $this->ignore_fieldsets = array('details', 'item_associations', 'language');
 $this->useCoreUI        = true;
 ?>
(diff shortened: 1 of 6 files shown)


RickR2H:
IMHO This will add empty spaces in the class names which is better to avoid. Some HTML validator might get picky about it.
Why not use:
```
<?php echo $this->pageclass_sfx ? ' ' . $this->pageclass_sfx : ''; ?>
```

brianteeman:
Quite correct. I guess I was being lazy and copying the existing code from com_wrapper. I will update this pr shortly. Hopefully someone can then decide if the classes are even needed/desired

RickR2H:
I have tested this item :white_check_mark: successfully on 6b547a5ca7a1517f338348a2ec5456454296e490<br /><br />Patch works. Nice catch!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37055">issues.joomla.org/tracker/joomla-cms/37055</a>.</sub>

simbus82:
I have tested this item :white_check_mark: successfully on 6b547a5ca7a1517f338348a2ec5456454296e490<br /><br />

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37055">issues.joomla.org/tracker/joomla-cms/37055</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37055">issues.joomla.org/tracker/joomla-cms/37055</a>.</sub>

bembelimen:
So we will break all templates which relies on the fact, that the page class suffix is attached to the existing classes (beside that it was a brackish decision in the past to put this class anywhere else than the body tag).

I see the point in this PR, but for b/c reasons I can't merge it (and probably for 5.0 we should remove it from inside the extensions as suggested by Brian).

brianteeman:
I fully agree and understand

Issue #37080, opened 2022-02-17
Getting HTTP 502 Bad Gateway in admin Dashboard when DEBUG is ON


mikeazores:
### Steps to reproduce the issue
After update to Joomla 4.1, having DEBUG ON, when going to the Administration Home Dashboard I get a HTTP 502 Bad Gateway.

### Expected result
Display the admin control panel.

### Actual result
Got a HTTP 502 Bad Gateway message.

### System information (as much as possible)
PHP Built On	Linux ae3dedf9ab04 5.4.39-linuxkit #1 SMP Fri May 8 23:03:06 UTC 2020 x86_64
Database Type	mysql
Database Version	10.5.6-MariaDB
Database Collation	utf8_general_ci
Database Connection Collation	utf8mb4_general_ci
Database Connection Encryption	None
Database Server Supports Connection Encryption	No
PHP Version	8.0.14
Web Server	nginx/1.14.2
WebServer to PHP Interface	fpm-fcgi
Joomla! Version	Joomla! 4.1.0 Stable [ Kuamini ] 15-February-2022 09:30 GMT
User Agent	Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.80 Safari/537.36 Edg/98.0.1108.50

=============
PHP Settings
=============
memory_limit: 384M
upload_max_filesize: 100M
post_max_size: 100M
display_errors: true
short_open_tag: true
file_uploads: true
output_buffering: false
open_basedir: 
session.save_path: 
session.auto_start: 0
disable_functions: 
xml: true
zlib: true
zip: true
mbstring: true
gd: true
iconv: true
intl: true
max_input_vars: 1000

=============
Configuration File
=============
offline: false
offline_message: Este sitio está cerrado por tareas de mantenimiento.<br />Por favor, inténtelo nuevamente más tarde.
display_offline_message: 1
offline_image: 
sitename: xxxxxx
editor: tinymce
captcha: 0
list_limit: 20
access: 1
debug: false
debug_lang: false
debug_lang_const: true
dbtype: mysqli
host: xxxxxx
user: xxxxxx
password: xxxxxx
db: xxxxxx
dbprefix: xxxxxx
dbencryption: 0
dbsslverifyservercert: false
dbsslkey: 
dbsslcert: 
dbsslca: 
dbsslcipher: 
force_ssl: 0
live_site: 
secret: xxxxxx
[...]

PR #37087, opened 2022-02-18
[4.1] Set a header limit for server timing

laoneo:
Pull Request for Issue #37080.

### Summary of Changes
Define a header limit for the server response time in debug mode. This is required due some limits in web servers.

The limit is set to 4000 which should fulfill all common web servers. according to https://stackoverflow.com/a/8623061.

### Testing Instructions
- Use nginx web server.
- Create a lot of dashboard modules.
- Enable debug mode in Joomla configuration.

Bad gatweway error as in issue.

Site is rendered properly.


zero-24:
Can we get a backport of this for 3.10 too?

Fedik:
Maybe a better do not add $metric when it too big, because adding "half data" will show incorrect "timing", that may mislead.
Or I missed something?

Fedik:
@zero-24 we do not have it in 3.10, it was a new feature in 4.1 :wink:

zero-24:
Ha ok no backport needed in this case :D

richard67:
@laoneo Does this PR solve issue #37080 so the issue can be closed?

laoneo:
> @laoneo Does this PR solve issue #37080 so the issue can be closed?

At the end of the day it should solve it

laoneo:
> Maybe a better do not add $metric when it too big, because adding "half data" will show incorrect "timing", that may mislead. Or I missed something?

I was thinking the same, perhaps we should then return one entry with a message too much data.

mikeazores:
In reference to #37080 if I lower the strlen value to say 3600 IT WORKS. With the current PR value (4000) I still have problems.

zero-24:
@laoneo do you think it makes sense to make that a option within the plugin with a default value for 4k (Yes I also hate new options)? So on sites such as @mikeazores 's he can lower the value but on other sites we can use the 4k / 8k etc.?

Fedik:
> I was thinking the same, perhaps we should then return one entry with a message too much data.

Or maybe do only 2 point timing: between start-end (when too big).
Sadly we cannot easily exclude timings for modules.

laoneo:
> @laoneo do you think it makes sense to make that a option within the plugin with a default value for 4k (Yes I also hate new options)? So on sites such as @mikeazores 's he can lower the value but on other sites we can use the 4k / 8k etc.?

For me it should work properly out of the box and it is not that critical, that it must always return a value.



> > I was thinking the same, perhaps we should then return one entry with a message too much data.
> 
> Or maybe do only 2 point timing: between start-end (when too big). Sadly we cannot easily exclude timings for modules.

We can, and deliver module rendering as one metric when there is a string like `mod_` in the profile step.

BrainforgeUK:
Sorry, 4000 limit is too large, I had to make it 3500 to get it to work on my server.
See https://github.com/joomla/joomla-cms/issues/37090

The server-wide Nginx settings for my server are 'out of the box'.

Can you making it configurable on the debug plugin parameters?
Default to 2000?
Warning about needing to change nginx settings if you increase it?
Suggest those nginx settings I grabbed earlier as examples?

laoneo:
@BrainforgeUK probably the better approach would be to collect the module ones as one entry. Having an option for such a small feature would be overhead.

laoneo:
The last commit adds a new behavior where the modules are collected together and are printed as one entry. @Fedik what do you think?

It looks then like:

image

Fedik:
Looks good to me.

Additionally still better to keep your previous approach with limit check, just to be safe, around 3K.
Because we cannot guarantee size when any extensions will add own Profiling data.

laoneo:
Added a limit again. Can you guys test it now so we can ship it with the next patch release?

plugins/system/debug/debug.php
```diff
@@ -691,13 +692,28 @@ public function onBeforeRespond(): void
...
 
+				continue;
+			}
+
+			$desc     = str_ireplace('after', '', $mark->label);
 			$name     = preg_replace('/[^\da-z]/i', '', $desc);
-			$metrics .= sprintf('%s;dur=%f;desc="%s:", ', $index . $name, $mark->time, $desc);
+			$metrics .= sprintf('%s;dur=%f;desc="%s", ', $index . $name, $mark->time, $desc);
+
+			// Do not crete too large headers, some web servers don't love them
```
Quy:
```suggestion
			// Do not create too large headers, some web servers don't love them
```

HLeithner:
Can you Plesse add 2 things?
1. make the limit a parameter
2. Add a line that to the metric that the limit got hit.

So it's possible to know that something is missing and that you can adjust it depending on your needs.

laoneo:
Adding  a new parameter for such a small feature makes no sense and the majority will not even understand for what it is.

HLeithner:
The parameter is in the debug plugin, the person configuring the plugin should know what the web does but OK, it's not important I guess. But the information that the data is truncated is important from my point of view.

laoneo:
How would you render that message?

Fedik:
> How would you render that message?

Kind of last row in metrics
```php
$metrics .= sprintf('%s;dur=%f;desc="%s", ', $index . 'Rest of data is gone forever', 'infinity', '');
```
:smiley:

laoneo:
Last commit shows now a truncated message:

image

If you want to test it, change the code `if (strlen($metrics) > 3000)` to `if (strlen($metrics) > 1)`.

plugins/system/debug/debug.php
```diff
@@ -691,13 +692,29 @@ public function onBeforeRespond(): void
 				continue;
 			}
 
-			$desc = $mark->label;
-			$desc = str_ireplace('after', '', $desc);
+			// Collect the module render time
+			if (strpos($mark->label, 'mod_') !== false)
+			{
+				$moduleTime += $mark->time;
```
Quy, line 702:
```suggestion
```

plugins/system/debug/debug.php
```diff
@@ -691,13 +692,29 @@ public function onBeforeRespond(): void
...
+
+			$desc     = str_ireplace('after', '', $mark->label);
 			$name     = preg_replace('/[^\da-z]/i', '', $desc);
-			$metrics .= sprintf('%s;dur=%f;desc="%s:", ', $index . $name, $mark->time, $desc);
+			$metrics .= sprintf('%s;dur=%f;desc="%s", ', $index . $name, $mark->time, $desc);
+
+			// Do not create too large headers, some web servers don't love them
+			if (strlen($metrics) > 3000)
+			{
+				$metrics .= 'System;dur=0;desc="Data truncated to 3000 characters"';
```
Quy:
```suggestion
				$metrics .= 'System;dur=0;desc="Data truncated to 3000 characters", ';
```

Quy:
I have tested this item :white_check_mark: successfully on 9767226fbdde11c0cf3e7ae2f454139fc38e3672<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

laoneo:
@BrainforgeUK and @mikeazores can you test it again as well?

alikon:
I have tested this item :white_check_mark: successfully on 9767226fbdde11c0cf3e7ae2f454139fc38e3672<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

richard67:
Back to pending due to new changes.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

laoneo:
We had the same problem with access when a none admin get logged in on the back end. The recent changes are grouping also access.

HusiBT:
I have tested this item :white_check_mark: successfully on 105dc3d82423f673f1ed72a69444695f9536696b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

HLeithner:
I have tested this item :white_check_mark: successfully on 105dc3d82423f673f1ed72a69444695f9536696b<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37087">issues.joomla.org/tracker/joomla-cms/37087</a>.</sub>

plugins/system/debug/debug.php
```diff
@@ -691,13 +693,38 @@ public function onBeforeRespond(): void
...
-			$desc = $mark->label;
-			$desc = str_ireplace('after', '', $desc);
+			// Collect the module render time
+			if (strpos($mark->label, 'mod_') !== false)
+			{
+				$moduleTime += $mark->time;
+				continue;
+			}
+
+			// Collect the module render time
```
Quy:
```suggestion
			// Collect the access render time
```

bembelimen:
Can't we just put the array in a variable, count and slice out the items which are too much from the middle (https://www.php.net/manual/en/function.array-splice.php) so we have a start/end time?

laoneo:
You just have a time of a task, there is no start/end time. This is not a whaterfall chart.

laoneo:
I'm setting this one as a release blocker as it affects a lot of sites and should definitely be fixed and finally shipped with 4.1.1.

richard67:
@laoneo Well, the last comment was from @bembelimen , so it would be good to know if he is ok with your reply to it,

laoneo:
He didn't reply so my assumption was that he did understand my explanation. I mean the pr is so correct, if he wants to make some adjustments, then he can do that in a followup pr.

richard67:
Sure.

bembelimen:
Thx

PR #37095, opened 2022-02-19
[4.2] Inject the db into the MVC layer

laoneo:
### Summary of Changes
Injects the database instance into the model. So we don't have to rely on the global db anymore. Additionally it deprecates the use of the database in the model constructor and uses the aware interface and trait from the db package.

### Testing Instructions
- Browse around in the back end and front end of Joomla
- Open the back end user list view
- Open the back end modules list view

All works.

All works.


libraries/src/MVC/Model/DatabaseAwareInterface.php
```diff
@@ -0,0 +1,32 @@
...
+	 * Set the database to use.
+	 *
+	 * @param   DatabaseInterface  $db  The database to use.
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function setDbo(DatabaseInterface $db = null): void;
+}
```
nibra: `getDbo()` should also be part of the interface since it is used many places without checking its existence.

laoneo: I can add it, but getDbo is only  used on classes using the aware trait and is never used where it is worked on the interface only.

wilsonge:
This looks wrong to me. The only place that should be database aware is the model. All controller and views database queries should be eventually moved into the model. not be made database aware

joeforjoomla:
> This looks wrong to me. The only place that should be database aware is the model. All controller and views database queries should be eventually moved into the model. not be made database aware

Totally agree with @wilsonge

laoneo:
> This looks wrong to me. The only place that should be database aware is the model. All controller and views database queries should be eventually moved into the model. not be made database aware

I agree too, and the problem is that there are queries which do not actually fit into a model, for example against the extensions table. I mean I can move all of them to the model of the view, but would be nicer to have them just from the beginning on the right place.

libraries/src/MVC/Factory/MVCFactory.php
```diff
@@ -24,9 +26,9 @@
  *
  * @since  3.10.0
  */
-class MVCFactory implements MVCFactoryInterface, FormFactoryAwareInterface, DispatcherAwareInterface
+class MVCFactory implements MVCFactoryInterface, FormFactoryAwareInterface, DispatcherAwareInterface, DatabaseAwareInterface
```
wilsonge: The factory should use the aware interface. This one needs it injected through the constructor

wilsonge:
This looks better. I don't think that factory should be databaseaware. Just the model. But definitely better than what we had before.

laoneo:
True, removed the interface.

HusiBT:
I have tested this item :white_check_mark: successfully on 67cdd831baf20d9432571fc816f8300608e17e72<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37095">issues.joomla.org/tracker/joomla-cms/37095</a>.</sub>

wilsonge (approved):

roland-d:
Thank you everybody

PR #37097, opened 2022-02-19
[4.2] Inject db into the form fields

laoneo:
Pull Request for pr #37095.

### Summary of Changes
Injects the database instance into the forms and form fields.

### Testing Instructions
- In the file /libraries/src/Form/Form.php add the following code after line 1629:  
`echo 'NOT INJECTED!';die;`
- Open the article form.

All is working.

All is working, the form loads and "NOT INJECTED!" is not displayed on the page.


brianteeman:
From the information provided its impossible to test.

laoneo:
Just open the article form, either in the back end and front end. Make sure there is no error. As this change injects the database into the form, every form field is affected and no "Database must be set, this will not be caught anymore in 5.0." message is added to the deprecate log.

laoneo:
Updated the testing instructions.

roland-d:
Thank you.

PR #37107, opened 2022-02-21
Bootstrap collapse asset for sub-categories

Denitz:
### Summary of Changes

Content, Contact and Newsfeeds components don't load Bootstrap collapse asset and expanding sub-categories is broken.

We need to load `bootstrap.collapse` asset like it's currently done for `com_search/tmpl/default-form.php`

### Testing Instructions

Use any non-native template, i.e. Gantry.
See child sub-categories in com_content, com_contact and com_newsfeeds category view, click "plus" icon to expand sub-categories.

Nothing happens.

Sub-categories are expanded.

No.


brianteeman:
This can not be correct. See 

https://github.com/joomla/joomla-cms/blob/6e71102970d6127874ae2b32c01eb650f8bc0fdb/components/com_contact/tmpl/categories/default.php#L19-L22

https://github.com/joomla/joomla-cms/blob/6e71102970d6127874ae2b32c01eb650f8bc0fdb/build/media_source/com_categories/joomla.asset.json#L7-L14

https://github.com/joomla/joomla-cms/blob/4.1-dev/build/media_source/com_categories/js/shared-categories-accordion.es6.js

Denitz:
OK, but where Bootstrap Collapse is loaded?

brianteeman:
it isnt because it is using a different script as shown

joomdonation:
@Denitz As Brian pointed out, we use our own accordion script. If I remember correctly, all you need to do is using the same code which we are having in com_content https://github.com/joomla/joomla-cms/blob/4.1-dev/components/com_content/tmpl/categories/default_items.php#L35-L44 for these components and it will work as expected.

Denitz:
@brianteeman Yes, you are right! Actually
 `components/com_content/tmpl/categories/default_items.php` 
 `components/com_contact/tmpl/categories/default_items.php` 
 `components/com_newsfeeds/tmpl/categories/default_items.php` 
uses `<button id=category-btn-..` with `com_categories.shared-categories-accordion` script.
No changes here.

But 
`components/com_content/tmpl/category/blog_children.php`
`components/com_content/tmpl/category/default_children.php` 
use `<a data-bs-toggle="collapse"...` without `HTMLHelper::_('bootstrap.collapse')` , I added Bootstrap script load.

Or we should re-build these two layouts in `components/com_content/tmpl/category` to use `<button id=category-btn-..` with `com_categories.shared-categories-accordion` script?

drmenzelit:
Hi @Denitz, sorry for taking so long to answer you question. The correct way should be to change the code in the views to remove bootstrap collapse and use Joomla's own accordion script. It would be great if you can rebase to 4.2 and modify the PR correspondly.

PR #37116, opened 2022-02-22
[4.1] Module "Articles - Categories" - Make the category selection optional

Bakual:
The module "Articles - Categories" requires to select a parent category. But that means you can't have the module show *all* categories, you can only show a subset.

I think this is a requirement which is not needed. The code itself already falls back to "root" if no category is selected.

### Summary of Changes
Removing the "required" flag from the parent category field.

### Testing Instructions
Test the module "Articles - Categories" and make especially sure nothing breaks when no category is selected.

You always need to select a category and the module shows only subset of the article categories.

You don't need to select a category and the module can show all categories.

None

### DIFF
--- a/modules/mod_articles_categories/mod_articles_categories.xml
+++ b/modules/mod_articles_categories/mod_articles_categories.xml
@@ -30,7 +30,6 @@
 					extension="com_content"
 					filter="integer"
 					published=""
-					required="true"
 					select="true"
 					new="true"
 					edit="true"


chmst:
I have tested this item :white_check_mark: successfully on b8c2a08274c91e58ea6989c11ed623ea15ee33d1<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37116">issues.joomla.org/tracker/joomla-cms/37116</a>.</sub>

brianteeman:
I think the history on this code needs to be checked carefully. I'm sure there must be a reason. It is possibly related to multilingual

Bakual:
Good point. It looks like it got added just to use the modal_category formfield for J3.7. Before that, the field was not required.
See https://github.com/joomla/joomla-cms/commit/45e14dd1d4a970a718539a7512c7e60135697882

toivo:
I have tested this item :white_check_mark: successfully on b8c2a08274c91e58ea6989c11ed623ea15ee33d1<br /><br />Tested successfully in Joomla 4.1.1-dev of 23 February in Wampserver 3.2.7 using PHP 8.0.15<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37116">issues.joomla.org/tracker/joomla-cms/37116</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37116">issues.joomla.org/tracker/joomla-cms/37116</a>.</sub>

laoneo:
> Good point. It looks like it got added just to use the modal_category formfield for J3.7. Before that, the field was not required. See [45e14dd](https://github.com/joomla/joomla-cms/commit/45e14dd1d4a970a718539a7512c7e60135697882)

Does the category form field still work as expected?

Bakual:
Why should it not work? I didn't change anything in the formfield.

laoneo:
But you are changing back something which was introduced on purpose. If that still works, then all is fine. Just wanted to double check before merging.

brianteeman:
> But you are changing back something which was introduced on purpose.

which is a big red warning flag to me

Bakual:
I honestly doubt that it indeed was changed on purpose. 
To me it looks more like copy-paste, because in other places it indeed has to be (and was before) required.
But here I couldn't find any issues with it being optional. And yep, of course the formfield can handle empty values.

Also, as I already wrote in the PR description, the code of the module already handles the case when no category is selected. I didn't had to change anything there.

bembelimen:
Thx

PR #37120, opened 2022-02-23
[4.4] Optimize scheduler running plugin

Denitz:
### Summary of Changes

`PlgSystemSchedulerunner::injectLazyJS()` checks the presence of due tasks via weird loading of items list with order by title DESC using the admin model:

```
SELECT a.id, a.asset_id, a.title, a.type, a.execution_rules, a.state, a.last_exit_code, a.locked, a.last_execution, a.next_execution, a.times_executed, a.times_failed, a.ordering, a.note
FROM `jos_scheduler_tasks` AS `a`
WHERE `a`.`state` = :state AND `a`.`next_execution` <= :now
ORDER BY `a`.`title` desc
```

Assuming that we are not loading the data but only counting the tasks which should be executed, we should not retrieve the full list of items, but only count them.

We have a second weird query to check if there are any locked tasks, the same invalid behavior occurs again:
```
SELECT a.id, a.asset_id, a.title, a.type, a.execution_rules, a.state, a.last_exit_code, a.locked, a.last_execution, a.next_execution, a.times_executed, a.times_failed, a.ordering, a.note
FROM `#__scheduler_tasks` AS `a`
WHERE `a`.`state` = :state AND `a`.`locked` IS NOT NULL
ORDER BY `a`.`title` desc
```

These two queries can be optimized into single efficient query which loads the number of due tasks + the number of locked tasks:
```
SELECT SUM(CASE WHEN `a`.`next_execution` <= :now THEN 1 ELSE 0 END) AS due_count, SUM(CASE WHEN `a`.`locked` IS NULL THEN 0 ELSE 1 END) AS locked_count
FROM `#__scheduler_tasks` AS `a`
WHERE `a`.`state` = :state
```

Using backend list model is not efficient, better compose the query in the plugin.

The result of single query is checked and the plugin is not adding JS if we don't have due tasks OR we have locked tasks.

### Testing Instructions

Test scheduled tasks

See two queries for `#__scheduler_tasks` table.

See one query for `#__scheduler_tasks` table.
com_scheduler works as before.

No.


plugins/system/schedulerunner/schedulerunner.php
```diff
@@ -120,22 +130,25 @@ public function injectLazyJS(EventInterface $event): void
...
+		$state = 1;
+		$now = Factory::getDate('now', 'UTC')->toSql();
 
-		$items = $model->getItems();
+		$query = $this->db->getQuery(true)
+			->from('#__scheduler_tasks AS a')
+			->where($this->db->quoteName('a.state') . ' = :state')
+			->bind(':state', $state)
+			// Count due tasks
+			->select('SUM(IF(`a`.`next_execution` <= :now, 1, 0)) AS due_count')
```
richard67: The if function is MySQL syntax. It will not work with PostgreSQL.

Denitz: Yep, I switched to CASE

richard67: Yes, that should work.

plugins/system/schedulerunner/schedulerunner.php
```diff
@@ -120,22 +130,25 @@ public function injectLazyJS(EventInterface $event): void
...
-		$items = $model->getItems();
+		$query = $this->db->getQuery(true)
+			->from('#__scheduler_tasks AS a')
+			->where($this->db->quoteName('a.state') . ' = :state')
+			->bind(':state', $state)
+			// Count due tasks
+			->select('SUM(CASE WHEN ' . $this->db->quoteName('a.next_execution') . ' <= :now THEN 1 ELSE 0 END) AS due_count')
+			->bind(':now', $now)
+			// Count locked tasks
+			->select('SUM(CASE WHEN ' . $this->db->quoteName('a.locked') . ' IS NULL THEN 0 ELSE 1 END) AS locked_count');
```
Quy:
```suggestion
			// Count due tasks
			->select('SUM(CASE WHEN ' . $this->db->quoteName('a.next_execution') . ' <= :now THEN 1 ELSE 0 END) AS due_count')
			// Count locked tasks
			->select('SUM(CASE WHEN ' . $this->db->quoteName('a.locked') . ' IS NULL THEN 0 ELSE 1 END) AS locked_count')
			->from($this->$db->quoteName('#__scheduler_tasks', 'a'))
			->where($this->db->quoteName('a.state') . ' = 1')
			->bind(':now', $now);
```

plugins/system/schedulerunner/schedulerunner.php
```diff
@@ -23,6 +23,8 @@
 use Joomla\CMS\User\UserHelper;
 use Joomla\Component\Scheduler\Administrator\Scheduler\Scheduler;
 use Joomla\Component\Scheduler\Administrator\Task\Task;
+use Joomla\Database\DatabaseInterface;
+use Joomla\Database\ParameterType;
```
Quy:
```suggestion
```

plugins/system/schedulerunner/schedulerunner.php
```diff
@@ -120,22 +130,25 @@ public function injectLazyJS(EventInterface $event): void
 			return;
 		}
 
-		// Check if any task is due to decrease the load
-		$model = $this->app->bootComponent('com_scheduler')
-			->getMVCFactory()->createModel('Tasks', 'Administrator', ['ignore_request' => true]);
-
-		$model->setState('filter.state', 1);
-		$model->setState('filter.due', 1);
+		$state = 1;
```
Quy:
```suggestion
```

richard67 (approved):
Looks ok to me by review.

ditsuke:
It was a deliberate decision to avoid any queries that bypass the model. That said I really like this, but could we instead add this as a method in the Scheduler model which could be then accessed through the Scheduler API class? That way we could avoid any localized logic in the plugin and also increase the useful API surface of the Scheduler which could be useful to other extensions. Thank you for working on this!

HLeithner:

laoneo:
Can you fix the conflict?

administrator/components/com_scheduler/src/Model/TasksModel.php
```diff
@@ -437,4 +438,34 @@ protected function populateState($ordering = 'a.title', $direction = 'ASC'): voi
...
+     * Check if we have any enabled due tasks and no locked tasks.
+     *
+     * @param   Date  $time  The next execution time to check against
+     *
+     * @return boolean
+     * @since  __DEPLOY_VERSION__
+     */
+    public function hasDueTasks(Date $time)
+    {
+        $db  = $this->getDbo();
```
laoneo:
```suggestion
        $db  = $this->getDatabase();
```

administrator/components/com_scheduler/src/Model/TasksModel.php
```diff
@@ -437,4 +438,34 @@ protected function populateState($ordering = 'a.title', $direction = 'ASC'): voi
...
+
+    /**
+     * Check if we have any enabled due tasks and no locked tasks.
+     *
+     * @param   Date  $time  The next execution time to check against
+     *
+     * @return boolean
+     * @since  __DEPLOY_VERSION__
+     */
+    public function hasDueTasks(Date $time)
```
laoneo:
```suggestion
    public function hasDueTasks(Date $time): bool
```

MacJoom:
@Denitz - thank you for your work - what about ditsukes comment, moving the query to the model - would be nice.

Denitz:
@MacJoom The query was moved to the model a year ago.

MacJoom:
Sorry i missed that one

Issue #37122, opened 2022-02-23
Module caching breaks WebAssetsManager


nikosdion:
Making the bat-signal to @bembelimen @richard67 

**WARNING!** This has some moderate security implications which can lead to partial Denial of Service on Joomla 4 sites using caching under certain circumstances.

### Steps to reproduce the issue

* Install a new Joomla 4.1 site (I used the latest 4.1-dev as of February 23rd, 2022)
* Do NOT install any sample data
* Components, Contacts, Contacts, Add Your First Contact and create a contact in the Uncategorised group
* Menus, Main Menu, New
	- Menu item type: Contacts, List All Categories in a Contact category Tree
	- Title: Contacts
	- Alias: contacts
* Save & Close
* System, Global Configuration and set the following in the System tab
	- System Cache: ON - Conservative Caching (VERY IMPORTANT! Do not use Progressive, that caching type makes it harder to reproduce the issue as I explain in the very long technical explanation below)
	- Cache Handler: File
	- Platform Specific Caching: No
	- Cache Time (minutes): 60
* Click on Save & Close
* System, Clear Cache, Delete All

Go to the frontend of your site and click through **IN THIS EXACT ORDER**:
* Contacts
* Uncategorised

### Expected result

I can see the contacts

### Actual result

Unsatisfied dependency "com_categories.shared-categories-accordion.es5" for an asset "com_categories.shared-categories-accordion" of type "script"

### System information (as much as possible)

Completely irrelevant. I have reproduced this issue on a plethora of local and live environments with PHP 7.4, 8.0, 8.1 and the latest versions of Safari, Chrome and Firefox. This is an architectural Joomla issue.

### Additional comments

You can confirm this only happens with conservative caching enabled by turning off the cache OR setting it to Progressive and trying the same click through path. You will see that the expected result of showing the contacts does happen. Therefore we are 100% sure this is a caching issue and ONLY with the Conservative caching option.

This is **far** from being an edge case. Real world sites are affected:
* https://www.facebook.com/groups/joomlanospam/permalink/10158167235345997/
* https://www.hikashop.com/forum/5-support-en-francais-french-support-/903954-0-unsatisfied-dependency-com-tags-tag-default-es5.html?fbclid=IwAR3w6gjkV7nYewfWkrrLOZlXmvt_AZmTBh8JutdYhYarxZ6dgHBqS2tJL8w#339322
[...]

PR #37123, opened 2022-02-23
Module caching breaks WebAssetsManager

nikosdion:
Pull Request for Issue #37122.

### Summary of Changes

Modified the `Joomla\CMS\Cache\Cache::setWorkarounds` to do a special treatment of the WebAssetManager assets before they are cached when caching a module.

### Testing Instructions

* Install a new Joomla 4.1 site (I used the latest 4.1-dev as of February 23rd, 2022)
* Do NOT install any sample data
* Components, Contacts, Contacts, Add Your First Contact and create a contact in the Uncategorised group
* Menus, Main Menu, New
	- Menu item type: Contacts, List All Categories in a Contact category Tree
	- Title: Contacts
	- Alias: contacts
* Save & Close
* System, Global Configuration and set the following in the System tab
	- System Cache: ON - Conservative Caching (VERY IMPORTANT! Do not use Progressive, that caching type makes it harder to reproduce the issue as I explain in the very long technical explanation below)
	- Cache Handler: File
	- Platform Specific Caching: No
	- Cache Time (minutes): 60
* Click on Save & Close
* System, Clear Cache, Delete All

Go to the frontend of your site and click through **IN THIS EXACT ORDER**:
* Contacts
* Uncategorised

### Actual result BEFORE applying this Pull Request

Unsatisfied dependency "com_categories.shared-categories-accordion.es5" for an asset "com_categories.shared-categories-accordion" of type "script"

### Expected result AFTER applying this Pull Request

I can see the contacts

### Documentation Changes Required

None.

### Important

Please look at the referenced issue's “Additional comments”. This can be a security issue for some sites.

PR #37139, opened 2022-02-25
Better module cache

Denitz:
### Summary of Changes

Please check https://github.com/joomla/joomla-cms/issues/37122 and https://github.com/joomla/joomla-cms/pull/35776

Now, before rendering the cached module, we empty the document and WebAssetmanager data, hence the module cache contains all data added by a module.

### Testing Instructions

Apply patch, enable Joomla caching, load page with cached modules

Module cache doesn't have own assets if they were also added before module render.

Module cache always have own assets in cache.

No.

### DIFF
--- a/libraries/src/Cache/Cache.php
+++ b/libraries/src/Cache/Cache.php
@@ -656,75 +656,38 @@ class Cache
 		{
 			if ($loptions['modulemode'] == 1)
 			{
-				$headnow = $document->getHeadData();
-				$unset   = array('title', 'description', 'link', 'links', 'metaTags');
+				$headNow = $document->getHeadData();
+				$unset   = ['title', 'description', 'link', 'links', 'metaTags'];
 
-				foreach ($unset as $un)
+				foreach ($unset as $key)
 				{
-					unset($headnow[$un]);
-					unset($options['headerbefore'][$un]);
+					unset($headNow[$key]);
 				}
 
-				$cached['head'] = [];
-
-				// Only store what this module has added
-				foreach ($headnow as $now => $value)
+				// Sanitize empty data
+				foreach (\array_keys($headNow) as $key)
 				{
-					if (isset($options['headerbefore'][$now]))
+					if (!isset($headNow[$key]) || $headNow[$key] === [])
 					{
-						// We have to serialize the content of the arrays because the may contain other arrays which is a notice in PHP 5.4 and newer
-						$nowvalue    = array_map('serialize', $headnow[$now]);
-						$beforevalue = array_map('serialize', $options['headerbefore'][$now]);
-
-						$newvalue = array_diff_assoc($nowvalue, $beforevalue);
-						$newvalue = array_map('unserialize', $newvalue);
-
-						// Special treatment for script and style declarations.
-						if (($now === 'script' || $now === 'style') && \is_array($newvalue) && \is_array($options['headerbefore'][$now]))
-						{
-							foreach ($newvalue as $type => $currentScriptStr)
-							{
-								if (isset($options['headerbefore'][$now][strtolower($type)]))
-								{
-									$oldScriptStr = $options['headerbefore'][$now][strtolower($type)];
-
-									// Save only the appended declaration.
-									if (\is_array($oldScriptStr) && \is_array($currentScriptStr))
-									{
-										$newvalue[strtolower($type)] = array_diff_key($currentScriptStr, $oldScriptStr);
-									}
-									else
-									{
-										$newvalue[strtolower($type)] = StringHelper::substr($currentScriptStr, StringHelper::strlen($oldScriptStr));
-									}
-								}
-							}
-						}
-					}
-					else
-					{
-						$newvalue = $headnow[$now];
-					}
-
-					if (!empty($newvalue))
-					{
-						$cached['head'][$now] = $newvalue;
+						unset($headNow[$key]);
 					}
 				}
+
+				$cached['head'] = $headNow;
 			}
 			else
 			{
 				$cached['head'] = $document->getHeadData();
+
+				// Document MIME encoding
+				$cached['mime_encoding'] = $document->getMimeEncoding();
 			}
 		}
-
-		// Document MIME encoding
-		$cached['mime_encoding'] = $document->getMimeEncoding();
 
 		// Pathway data
 		if ($app->isClient('site') && $loptions['nopathway'] != 1)
 		{
-			$cached['pathway'] = \is_array($data) && isset($data['pathway']) ? $data['pathway'] : $app->getPathway()->getPathway();
+			$cached['pathway'] = $data['pathway'] ?? $app->getPathway()->getPathway();
 		}
 
 		if ($loptions['nomodules'] != 1)
--- a/libraries/src/Cache/Cache.php
+++ b/libraries/src/Cache/Cache.php
@@ -15,7 +15,6 @@ use Joomla\CMS\Cache\Exception\CacheExceptionInterface;
 use Joomla\CMS\Factory;
 use Joomla\CMS\Filesystem\Path;
 use Joomla\CMS\Session\Session;
-use Joomla\String\StringHelper;
 
 /**
  * Joomla! Cache base object
--- a/libraries/src/Cache/Controller/CallbackController.php
+++ b/libraries/src/Cache/Controller/CallbackController.php
@@ -12,6 +12,7 @@ namespace Joomla\CMS\Cache\Controller;
 
 use Joomla\CMS\Cache\Cache;
 use Joomla\CMS\Cache\CacheController;
+use Joomla\CMS\Document\HtmlDocument;
 use Joomla\CMS\Factory;
 
 /**
@@ -36,6 +37,21 @@ class CallbackController extends CacheController
 	 */
 	public function get($callback, $args = array(), $id = false, $wrkarounds = false, $woptions = array())
 	{
+		if (!\is_array($args))
+		{
+			$referenceArgs = !empty($args) ? array(&$args) : array();
+		}
+		else
+		{
+			$referenceArgs = &$args;
+		}
+
+		// Just execute the callback if caching is disabled.
+		if (empty($this->options['caching']))
+		{
+			return \call_user_func_array($callback, $referenceArgs);
+		}
+
 		if (!$id)
 		{
 			// Generate an ID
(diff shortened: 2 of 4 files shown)


brianteeman:
testing this with the same instructions as for testing  #37123 and this also fixes the issue

brianteeman:
I have tested this item :white_check_mark: successfully on 9b1e757addbbc74cca3c62b4cbe9e9177e455eb4<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

alikon:
I have tested this item :white_check_mark: successfully on 9b1e757addbbc74cca3c62b4cbe9e9177e455eb4<br /><br />under the same conditions of #37123<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

libraries/src/Cache/Cache.php
```diff
@@ -656,75 +656,38 @@ public static function setWorkarounds($data, $options = [])
...
-									$oldScriptStr = $options['headerbefore'][$now][strtolower($type)];
-
-									// Save only the appended declaration.
-									if (\is_array($oldScriptStr) && \is_array($currentScriptStr))
-									{
-										$newvalue[strtolower($type)] = array_diff_key($currentScriptStr, $oldScriptStr);
-									}
-									else
-									{
-										$newvalue[strtolower($type)] = StringHelper::substr($currentScriptStr, StringHelper::strlen($oldScriptStr));
```
Quy, line 698: Please delete line 18.

libraries/src/WebAsset/WebAssetManager.php
```diff
@@ -153,6 +153,22 @@ public function getRegistry(): WebAssetRegistry
...
+	 * Clears all collected items.
+	 *
+	 * @return self
+	 *
+	 * @since  __DEPLOY_VERSION__
+	 */
+	public function reset(): WebAssetManagerInterface
+	{
+		$this->activeAssets = [];
+		$this->locked = false;
```
Fedik: danger danger!
`$this->locked` should not be reset under any circumstances. 
It say that last point when was possible to add asset is closed, so developers who try $wa->useScript() in onAfterRender, or onAfterResponse, clearly see that they doing something wrong

libraries/src/Cache/Cache.php
Denitz: Done.

libraries/src/WebAsset/WebAssetManager.php
Denitz: @Fedik I think it's not our case, we are resetting before rendering metas, hence the manager is never locked at this stage.
But logically, `reset()` allows re-use of this manager instance just like it's newly created, hence we need to unlock it since it doesn't have assets anymore.

nikosdion:
I have tested this item :white_check_mark: successfully on 434735e253e4bd40e411a0593dadb8bfb3d49bd7<br /><br />I performed the test instructions on this PR *and* those in #37122 (where I explained why the approach we see in this here PR would be most preferable). SUCCESS IN BOTH CASES! WOOT!

Thank you so very much @Denitz for writing the code! Here's to hoping this will make it into 4.1.1 :)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

nikosdion:
@alikon @brianteeman Would either (of both) of you please redo the test here so it can become RTC?

alikon:
I have tested this item :white_check_mark: successfully on 434735e253e4bd40e411a0593dadb8bfb3d49bd7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

libraries/src/WebAsset/WebAssetManager.php
Fedik: > But logically, reset() allows re-use of this manager

You should not reset locked manager, you should create a new instance instead.
If it already locked that mean all assets is "rendered", and any further changes will be ignored.

The Lock happens here, and you do need to reset it for module renderer.
https://github.com/joomla/joomla-cms/blob/1369453fd8a2a465b0fd02a3117c6746ca782426/libraries/src/Document/Renderer/Html/MetasRenderer.php#L80-L81

Fedik:
Addittionaly, instead of resetting AssetManager, can create temporary instance.
https://github.com/joomla/joomla-cms/blob/1369453fd8a2a465b0fd02a3117c6746ca782426/libraries/src/Document/Document.php#L858-L863

https://github.com/joomla/joomla-cms/blob/1369453fd8a2a465b0fd02a3117c6746ca782426/libraries/src/Document/Document.php#L336-L341

nikosdion:
@Fedik This is bad architecture. The web asset registry and manager should be locked in the container (set once). Otherwise any extension can break the entire frontend of the application by messing around with the WAM which completely beats the purpose of having the WAM in the first place...

Fedik:
Okay, I fine with both way, but `lock` should not be resetable.

nikosdion:
@Fedik It has to be if we want caching to work _correctly_. The other option is overengineering WAM to keep a log of actions (what I had originally proposed in my comments to you). It's far more complicated.

Any other solution would very likely require rethinking WAM but it's way too late for that. The best time to do that would have been around 3 years ago, before the Joomla 4 betas. We're well past that. So we have to work with what we have, even if it's a sad compromise.

Between the two evils, having reset() is the lesser evil. It's something that cannot be refactored out of Joomla without seeing it's used in the caching code which will (or SHOULD) trigger an alert in whoever is refactoring the code, find these issues and find a solution (or leave it alone). Setting a new WAM instance in the container would be refactored out without a second thought (it's such an obvious architectural mistake nobody will think twice) and we'd have a regression.

There _is_ a way to have our pie and eat it too. We can rig the reset() method to only be callable by Cache descendants by using debug_backtrace and making sure one of the callers is a class which is identical to or extends Cache. This would make the method public but inaccessible to unauthorised callers. It should also be marked as `@internal` in the docblock to note that. This is a trick I am using in my security software after I saw a lot of third party software trying to abuse my .htaccess Maker model to add exceptions for themselves without asking the user.

Fedik:
> It has to be if we want caching to work correctly.

I mean you can still use reset() if you like, but if manager is locked then it should not be possible to reset.
Even more, it should throw an exception when someone tries to reset "locked" manager.

Like in useAsset() https://github.com/joomla/joomla-cms/blob/9b1e757addbbc74cca3c62b4cbe9e9177e455eb4/libraries/src/WebAsset/WebAssetManager.php#L265-L270

The reason I wrote a some comments ago

bembelimen:
Any opinions on @Fedik comment?

nikosdion:
@bembelimen You can do that, but we'd have to reset all tests and retest yet again. At this point besides me, a party affected by the Joomla caching bug and who has the technical experience to test patches like that, nobody else would test, this would never be fixed and Joomla would forever remain broken (**and vulnerable!**) when caching is enabled. That would be unacceptable.

Instead, why don't we merge this as-is and @Fedik can make a new PR to add the lock check in reset()? If you ping me I can provide the test for it. Between the three of us and possibly four (if @alikon is still interested in this saga) we can find a second test and merge his PR. This way all of us can have our pie and eat it too. Fair?

bembelimen:
I thought about it and (yeah I hear y'all laughing in context of Joomla) we should not try to merge anything which has that obvious an issue we can resolve.

So I set this to an release blocker to show that I really intent to have this in 4.1.1. but only with the fix.

Additionally the following promise: if the fix is there till the 15th of march I will do everything so we get two proper and valid tests and it will be part in 4.1.1.

If after the 15th I still will try to get tests but can't guaranteed anymore.

Deal?

richard67:
> I thought about it and (yeah I hear y'all laughing in context of Joomla) we should not try to merge anything which has that obvious an issue we can resolve.
> 
> So I set this to an release blocker to show that I really intent to have this in 4.1.1. but only with the fix.
> 
> Additionally the following promise: if the fix is there till the 15th of march I will do everything so we get two proper and valid tests and it will be part in 4.1.1.
> 
> If after the 15th I still will try to get tests but can't guaranteed anymore.
> 
> Deal?

@Denitz Could you implement the changes suggested here https://github.com/joomla/joomla-cms/pull/37139#issuecomment-1056669038 ? Thanks in advance.

Denitz:
@richard67 Changes applied.

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

richard67:
@bembelimen Is is ok now?

@brianteeman @alikon @nikosdion Could you test again? Thanks in advance.

libraries/src/Cache/Controller/CallbackController.php
```diff
@@ -81,37 +97,29 @@ public function get($callback, $args = array(), $id = false, $wrkarounds = false
...
-		}
-
 		if ($locktest->locked === false && $locktest->locklooped === true)
 		{
 			// We can not store data because another process is in the middle of saving
 			return \call_user_func_array($callback, $referenceArgs);
 		}
 
-		$coptions = array();
+		$coptions = array('modulemode' => 0);
```
Quy, line 106:
```suggestion
		$coptions = ['modulemode' => 0];
```

richard67:
I have tested this item :white_check_mark: successfully on c5e26edbfbaeeaefa8368448b62163a51bf93178<br /><br />I've tested with the test instructions on this PR as well as with those in #37122 .<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

richard67:
@brianteeman @alikon @nikosdion One more test needed.

libraries/src/Document/HtmlDocument.php
```diff
@@ -173,10 +173,7 @@ public function getHeadData()
 		{
 			foreach ($assetNames as $assetName => $assetState)
 			{
-				if ($assetState === WebAssetManager::ASSET_STATE_ACTIVE)
```
Quy, line 176: Please delete line 21. 
https://github.com/joomla/joomla-cms/blob/c5e26edbfbaeeaefa8368448b62163a51bf93178/libraries/src/Document/HtmlDocument.php#L21

richard67: @Denitz Could you do that, remove the now obsolete use statement? Thanks in advance.

Denitz: @richard67 Removed.

richard67:
I have tested this item :white_check_mark: successfully on c59a414ddd16276f635e7dcd02ee445d9bf4fc8f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

richard67:
@brianteeman @alikon @nikosdion Am still begging for a 2nd test.

alikon:
I have tested this item :white_check_mark: successfully on c59a414ddd16276f635e7dcd02ee445d9bf4fc8f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37139">issues.joomla.org/tracker/joomla-cms/37139</a>.</sub>

bembelimen:
@nikosdion could you please quick check if you're still happy with the PR, then I would merge! Thanks.

nikosdion:
I’m happy with this. Please merge. Thank you!

bembelimen:
Thank you all!

PR #37022, opened 2022-02-13
[4.1] Use the image JLayout for all the images

dgrammatiko:
Pull Request is a continuation of https://github.com/joomla/joomla-cms/pull/37008, https://github.com/joomla/joomla-cms/pull/36613 and https://github.com/joomla/joomla-cms/pull/35780

### Summary of Changes
- Render images using the jLayout

### Testing Instructions
Check the images render as expected

### Documentation Changes Required

I didn't touch the `mod_banners/tmpl/default.php` but that one should also be patched

PR #37148, opened 2022-02-25
[4.1] Use HTMLHelper instead of layout helper

laoneo:
Pull Request for pr #37022.

### Summary of Changes
This is a followup of #37022 which uses HTMLHelper instead of LayoutHelper.

### Testing Instructions
See #37022.

Image is rendered.

Image is rendered.


dgrammatiko:
May I ask why? 

The only reason that someone should use the `HTMLHelper` is for the relative path (which obviously none of the user uploaded images will never use).
Also, at least from my point of view, all the HTML fragments should use JLayouts as the overriding is way more friendly for front end devs, HTMLHelper actually needs a plugin.

What I'm trying to say is that it was a conscious decision **NOT** to use the HTMLHelper for reasons I stated above, but if the project thinks that the HTMLHelper is the right way, well no comments...


EDIT: ok I'm an idiot, I should have checked all the files..

THIS IS FINE 👍

dgrammatiko:
@laoneo  some more meaningful feedback here

Insert this in the index.php of cassiopeia (somewhere in the body, just so that it would render):

```php
echo HTMLHelper::_('image', 'images/powered_by.png#joomlaImage://local-images/powered_by.png?width=294&height=44', 'Some Alt Text', ['class' => 'background-image']);
```

The image renders correctly, but the passed URL is cleaned from the `#` part which is problematic as it would never get the width/height and also disregards the loading attribute.

Possible solution:

```php
	public static function image($file, $alt, $attribs = null, $relative = false, $returnPath = 0)
	{
		$returnPath = (int) $returnPath;

		if ($returnPath !== -1) {
			$newFile = static::cleanImageURL($file);
			$includes = static::includeRelativeFiles('images', $newFile->url, $relative, false, false);
			$newFile->url = \count($includes) ? $includes[0] : null;
			$file = $newFile->url . '#' . $newFile->adapter . '?width=' . $newFile->attributes['width'] . '&height=' . $newFile->attributes['height'];
		}

		// If only path is required
		if ($returnPath === 1) {
			$newFile = (static::cleanImageURL($file));
			$includes = static::includeRelativeFiles('images', $newFile->url, $relative, false, false);
			$newFile->url = \count($includes) ? $includes[0] : null;
			return $newFile->url;
		}

		// Ensure we have a valid default for concatenating
		if ($attribs === null || $attribs === false) {
			$attribs = [];
		}

		// When it is a string, we need convert it to an array
		if (is_string($attribs)) {
			$attributes = [];

			foreach (explode(' ', $attribs) as $attribute) {
				if (strpos($attribute, '=') === false) {
					$attributes[$attribute] = '';
					continue;
				}

				list($key, $value) = explode('=', $attribute);
				$attributes[$key]  = trim($value, '"');
			}

			$attribs = $attributes;
		}

		$attribs['src'] = $file;
		$attribs['alt'] = $alt;

		return LayoutHelper::render('joomla.html.image', $attribs);
	}
```

But it needs couple of lines changed here:

```php
	public static function cleanImageURL($url)
	{
		$obj = new \stdClass;

		$obj->attributes = [
			'width'  => 0,
			'height' => 0,
			'adapter' => '',
		];

		if (!strpos($url, '?'))
		{
			$obj->url = $url;

			return $obj;
		}

		$mediaUri = new Uri($url);

		// Old image URL format
		if ($mediaUri->hasVar('joomla_image_height'))
		{
			$height = (int) $mediaUri->getVar('joomla_image_height');
			$width  = (int) $mediaUri->getVar('joomla_image_width');

			$mediaUri->delVar('joomla_image_height');
			$mediaUri->delVar('joomla_image_width');
		}
		else
		{
			// New Image URL format
			$fragmentUri  = new Uri($mediaUri->getFragment());
			$width        = (int) $fragmentUri->getVar('width', 0);
			$height       = (int) $fragmentUri->getVar('height', 0);
			$obj->adapter = $fragmentUri->getVar('scheme', '') . '://' . $fragmentUri->getVar('host', '');
		}

		if ($width > 0)
		{
			$obj->attributes['width'] = $width;
		}

		if ($height > 0)
		{
			$obj->attributes['height'] = $height;
		}

		$mediaUri->setFragment('');
		$obj->url = $mediaUri->toString();

		return $obj;
	}
```

laoneo:
I guess we do that in another pr. This here is just a followup from yours. There are a lot of places we can optimize the image rendering.

dgrammatiko:
But you are undoing all the lazy loading and the rest of the attributes in the images here, so in that sense the pr is wrong

dgrammatiko:
I have tested this item :red_circle: unsuccessfully on 8cdc8bdad54bc4459656b247e4e1bee7da6d2300<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37148">issues.joomla.org/tracker/joomla-cms/37148</a>.</sub>



The output is missing the width, height and loading attribute, so this is a performance regression (this PR will increase the CLS [Layout shift])

Before
Screenshot 2022-02-27 at 15 05 15

With this PR
Screenshot 2022-02-27 at 15 02 53

laoneo:
And this works without the pr? Because all this pr does is to use the htmlhelper and then forward to the layout.

dgrammatiko:
@laoneo check my previous comment, the htmlhelper is stripping the url but then the layout doesn’t have the needed data…

laoneo:
Can you check again. It should work now.
For the other testers, please also test with normal urls and with invalid ones.

components/com_finder/tmpl/search/default_result.php
```diff
@@ -74,13 +73,12 @@
...
 		<figure class="<?php echo htmlspecialchars($imageclass, ENT_COMPAT, 'UTF-8'); ?> result__image">
-			<?php $extraAttr = array_merge($extraAttr, ['src' => $this->result->imageUrl, 'alt' => $this->result->imageAlt]); ?>
 			<?php if ($this->params->get('link_image') && $this->result->route) : ?>
 				<a href="<?php echo Route::_($this->result->route); ?>">
-					<?php echo LayoutHelper::render('joomla.html.image', $extraAttr); ?>
+					<?php echo HTMLHelper::_('image', $image->url, $this->result->imageAlt, $extraAttr); ?>
 				</a>
 			<?php else : ?>
-				<?php echo LayoutHelper::render('joomla.html.image', $extraAttr); ?>
+				<?php echo HTMLHelper::_('image', $image->url, $this->result->imageAlt, $extraAttr); ?>
```
Quy:
```
Notice: Undefined variable: image in \components\com_finder\tmpl\search\default_result.php on line 81

Notice: Trying to get property 'url' of non-object in \components\com_finder\tmpl\search\default_result.php on line 81
```

laoneo: Can you try again?

Quy:
There are additional ones to convert.

Quy:
There are additional ones to convert.

laoneo:
This one here changes the ones from pr https://github.com/joomla/joomla-cms/pull/37022. We can do the other ones as followup.

N6REJ:
Using the testing instructions buried in the comments 
```
echo HTMLHelper::_('image', 'images/powered_by.png#joomlaImage://local-images/powered_by.png?width=294&height=44', 'Some Alt Text', ['class' => 'background-image']);
```
the image is full width 
image
after applying the patch the image is it's normal size.
image

laoneo:
Where did you add that code?

Quy:
I have tested this item :white_check_mark: successfully on 141b2e9b3dc251346de70a85b27b245bf15d9bbc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37148">issues.joomla.org/tracker/joomla-cms/37148</a>.</sub>

N6REJ:
> Where did you add that code?

Very beginning of the body.

laoneo:
@N6REJ can you test it with the 4.1 branch or with the stable package?

N6REJ:
I have tested this item :white_check_mark: successfully on 141b2e9b3dc251346de70a85b27b245bf15d9bbc<br /><br />Works as expected.  Rendered code looks much nicer.

Using todays 4.1 pull<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37148">issues.joomla.org/tracker/joomla-cms/37148</a>.</sub>

Pre-patch:
image


Post Patch:
image
code used...
```
<div class="site-grid">
	<?php echo HTMLHelper::_('image', 'images/powered_by.png#joomlaImage://local-images/powered_by.png?width=294&height=44', 'Some Alt Text', ['class' => 'background-image']); ?>
</div>
```

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37148">issues.joomla.org/tracker/joomla-cms/37148</a>.</sub>

bembelimen:
Thx

DelPoint:
@laoneo HTMLHelper_ code in files of com_tags (default.php, list.php etc.) doesn't take into account the back-end settings of "**Alt Text**", "**Caption**" & "**Float**" for the tag's Teaser/Full Image. See below:
image

DelPoint:
@laoneo com_tags/tags/default.php: logical error in echo HTMLHelper::_ code, backend settings are incorrectly implemented (also the original code with LayoutHelper is wrong).
Backend settings:

image

As you can see **Image Description (Alt Text)** and "**No Description**" settings are related.

There are 4 cases possible:
1. **Alt Text** missing + "**No Desc.**" = false (unset/unselected)
   => no output of "alt" property
2. **Alt Text** missing + "**No Desc.**" = true (set/selected/thicked)
   => no output of "alt" property
3. **Alt Text** exist + "**No Desc.**" = false
   => output of "alt" property
4. **Alt Text** exist + "**No Desc.**" = true
   => no output of "alt" property, as in this case the "No Desc." should overwrite the existing "Alt Text"

The above logic is not followed currently here:
https://github.com/Digital-Peak/joomla-cms/blob/141b2e9b3dc251346de70a85b27b245bf15d9bbc/components/com_tags/tmpl/tags/default.php#L26

INMHO this would be the appropriate code:
image

Note that 2nd empty negated, and the 2 "empties" are in logical "OR" relation. That would mirror the above logic for the mentioned 4 possible cases.

Cheers.

laoneo:
Thanks for the report. But I do not think this is related to this pr. Can you make a new pr with your suggestion so we can analyze the issue as this pr is closed. Thanks for understanding.

DelPoint:
No problem, I've expected a similar reply. I've reported it here because I've seen you are heavily involved in this Layout->HTMLHelper conversion thingie, and thought maybe it is a good opportunity to warn you if you just take the old code with LayoutHelper and you simply replace it with the HTMLHelper there are still some inherent issues not resolved... so we have to touch the same file/code again... and I hate to run in circles :)
I'll open a new issue about these problems.

svenbluege:
I have a small issue with the code change. For components which use dynamic links to render images or use the anchor. 

URL with potential problems: ```http://foo.bar/image#2``` or ```http://foo.bar/image?width=200```

HTMLHelper.php:726
    // Get the arguments positions
		$pos1 = strpos($file, '?');
		$pos2 = strpos($file, '#');

		// Check if there are arguments
		if ($pos1 !== false || $pos2 !== false)
		{
			// Get the path only
			$path = substr($file, 0, min($pos1, $pos2));

			// Determine the arguments is mostly the part behind the #
			$arguments = str_replace($path, '', $file);
		}`

This code expects a ? and a # to work correctly. If one of them is not in the $file-string, the $path-substring will be empty. This creates a issue with the URLs above. ```min(a,b)``` returns false if a or b are false.

Issue #37124, opened 2022-02-23
J4 Action 'Update Structure' in Maintenane:Database for plugins sets value of extension version in manifest_cache to empty string


avjoomla:
![screen shot 2022-02-23 at 14 07 50](https://issues.joomla.org/uploads/1/2dd45524177f3dc55b2fdfb42db2152d.jpg)### Steps to reproduce the issue
- install a plugin which has it's own table in the database and uses update -> schemas
- I've created a simple plugin 'Visforms - Limit number of submissions' for this purpose, but I cannot upload the installation here, due to 'file type of zip is not allowed'. I can mail it, if you like to
- go to System >> Maintenance: Database
- Database check for the entry 'Visforms - Limit number of submissions' shows "No Problems"
- Select entry for 'Visforms - Limit number of submissions'
- Click 'Update Structure'
I've attached some screenshots which show the problem

### Expected result
Still 'No Problems' with the data table structure

### Actual result
Result for 'Visforms - Limit number of submissions' shows 'One Problem'

### System information (as much as possible)
Tested with php7.4 and 8.0 on different environments
Clean Joomla 4.0 and 4.1 with no additional extensions except the plugin

### Additional comments
The problem is cause by by the function fixUpdateVersion in administrator/components/com_installer/src/Model/DatabaseModel.php on line 635ff which does not take into account, that the path to a plugin and it's installation manifest xml is more complex and different form components. (Same statement would be correct for modules, by the way)
As a result, the installation manifest file is not found and the value of the 'version' property in the 'manifest_cache' of the plugin in the table #__extensions is actively and wrongly set to an mpty string.

I've tested the following fix, which seems to work fine.

administrator/components/com_installer/src/Model/DatabaseModel.php
After Line 646: Add custom code for plugin, which store there installationXML files in subfolders and under JPATH_SITE
else if ($table->get('type') === 'plugin') {
    $installationXML  = InstallerHelper::getInstallationXML($table->get('element'), $table->get('type'), 0, $table->get('folder'));
    $extensionVersion = (string) $installationXML->version;
}
[...]

PR #37160, opened 2022-03-01
[4.1] Fix the update version fix of the database fix for 3rd party extensions

richard67:
Pull Request for Issue #37124 .

### Summary of Changes

This pull request (PR) fixes the database checker's method for fixing the update version so it finds the manifest XML in all cases and so uses the right extension version and not an empty string when being used for 3rd party extensions.

At other places in the same source file it is already done in the right way, e.g. here
https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_installer/src/Model/DatabaseModel.php#L192-L196
or here
https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_installer/src/Model/DatabaseModel.php#L563-L568

The issue is only relevant for J4 because the functionality of the database checker to check not only the CMS core but also 3rd party extensions is a new J4 feature.

### Testing Instructions

1. Install a plugin which includes an `<update>` tag in its manifest XML and has some update SQL scripts.
You can use the following dummy plugin: https://test5.richard-fath.de/plg_test_pr_37160.zip

2. Go to "System - Manage - Database".

3. Check that there is no problem shown for that plugin.

4. Select the row of the plugin and use the "Fix Structure" button regardless of no problems being shown.
Result: Now there is one problem shown for the plugin.

5. Hover over the badge with the "1 problem" to see the details.
Result: The problem is about not matching versions, of which one is an empty string.
See screenshot in section "Actual result BEFORE applying this Pull Request" below.

6. Apply the patch of this PR.

7. Go back to "System - Manage - Database".
Result: The problem from step 4 is still shown.

8. Select the row of that plugin (check box) and use the "Fix structure" button.
Result: No problem shown.

9. Repeat step 8 a few times to be sure it works well now.

10. Finally test if there is no issue with using the fix button for any other kind of extension which has update SQL scripts. You can use the patchtester extension for this test.

When using the "Fix structure" button of the database checker for a 3rd party plugin or a 3rd party backend module which come with own update SQL scripts so they are shown in the checker, the version in the manifest cache is set to an empty string, so after the fix you have this database problem:

pr-37160_1

When using the "Fix structure" button of the database checker for a 3rd party plugin or a 3rd party backend module which come with own update SQL scripts so they are shown in the checker, the version in the manifest cache is set to the right value from the manifest XML file, so there are no problems shown after using the "Fix Structure" button..

pr-37160_2

None.

### DIFF
--- a/administrator/components/com_installer/src/Model/DatabaseModel.php
+++ b/administrator/components/com_installer/src/Model/DatabaseModel.php
@@ -650,7 +650,7 @@ class DatabaseModel extends InstallerModel
 				$table->get('element'),
 				$table->get('type'),
 				$table->get('client_id'),
-				$table->get('element') === 'plugin' ? $table->get('folder') : null
+				$table->get('type') === 'plugin' ? $table->get('folder') : null
 			);
 			$extensionVersion = (string) $installationXML->version;
 		}
--- a/administrator/components/com_installer/src/Model/DatabaseModel.php
+++ b/administrator/components/com_installer/src/Model/DatabaseModel.php
@@ -646,7 +646,12 @@ class DatabaseModel extends InstallerModel
 		}
 		else
 		{
-			$installationXML  = InstallerHelper::getInstallationXML($table->get('element'), $table->get('type'));
+			$installationXML = InstallerHelper::getInstallationXML(
+					$table->get('element'),
+					$table->get('type'),
+					$table->get('client_id'),
+					$table->get('element') === 'plugin' ? $table->get('folder') : null
+			);
 			$extensionVersion = (string) $installationXML->version;
 		}
 
--- a/administrator/components/com_installer/src/Model/DatabaseModel.php
+++ b/administrator/components/com_installer/src/Model/DatabaseModel.php
@@ -647,10 +647,10 @@ class DatabaseModel extends InstallerModel
 		else
 		{
 			$installationXML = InstallerHelper::getInstallationXML(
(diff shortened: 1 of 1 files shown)


avjoomla:
I have tested this item :white_check_mark: successfully on 9d5ecfe4642b308dc4e50afde97f720a47d7f3a7<br /><br />I tested with your patch and the fix works. Thanks for taking care of this issue.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37160">issues.joomla.org/tracker/joomla-cms/37160</a>.</sub>

khu5h1:
I have tested this item :white_check_mark: successfully on 9d5ecfe4642b308dc4e50afde97f720a47d7f3a7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37160">issues.joomla.org/tracker/joomla-cms/37160</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37160">issues.joomla.org/tracker/joomla-cms/37160</a>.</sub>

bembelimen:
Thx

PR #37174, opened 2022-03-02
[5.2] Add support ID with 0 for value CheckboxField.php

korenevskiy:
This amendment concerns to use this field to pass ID as an argument. If we use this field not in an XML file but in a dynamic table of fields. In each row of the table, we use the Checkbox to select a row number or select an ID in the list. After activating these checkboxes, we mark which of the parameter lists with which ID we have activated.
I believe that the original code implied the use of checkboxes as just a parameter in the static configuration of the settings page, not to select an ID, but simply to select certain settings.
Therefore, such a subtlety as using the value '0' as a value was not taken into account.
My change does not violate compatibility and does not violate the default settings.
And of course it would be stupid on the part of external developers to use the value '0' of the attribute to get an activated checkbox.
Because the value '0' is interpreted as an empty value and we get an active checkbox. In this way, it is stupid to mark activity using unreadable, this is a stupid approach to activate the checkbox.
Therefore, I suggest using the value '0' as the value for ID selection.

``` PHP
$field = JFormHelper::loadFieldType('сheckbox', true);
$field->setup(simplexml_load_string('<field type="check" default="0"  name="id" />', 0);
echo $field->getInput();
```
The result can only be seen in the generated HTML. Because of this limitation of fields, we can't check in any other way, except to look for values in HTML.
Before:
``` HTML
<div class="form-check form-check-inline">
	<input
		type="checkbox"
		name="check"
		id="id"
		value="1"
	>
</div>
```
After:
``` HTML
<div class="form-check form-check-inline">
	<input
		type="checkbox"
		name="check"
		id="id"
		value="0" 
	>
</div>
```
Here you need to search for the string **`value ="0"`** with tests, otherwise no way.
In the Checkbox field, any default value is rendering correctly except for 0.
The default value of 0 is rendering as 1.
This fix can be checked in any module add writing this XML example to the module’s XML configuration file and after check  this value in the F12 panel in the browser .
```XML
<field 
	name="id"
	type="checkbox"  
	label="Test value 0 in Checkbox!!!"  
	default="0" />
```


laoneo:
Thanks for your pull request. Please add a proper test instruction, so we can test it.

korenevskiy:
> Thanks for your pull request. Please add a proper test instruction, so we can test it.

I can't find the Checkbox field test file. I'm new to writing tests. Previously, I wrote only 2 amendments to Joomla tests. Please tell me where I can find the Checkbox field test.

Quy:
Please provide steps to reproduce the issue in order to test the PR.

korenevskiy:
``` PHP
$field = JFormHelper::loadFieldType('сheckbox', true);
$field->setup(simplexml_load_string('<field type="check" default="0"  name="id" />', 0);
echo $field->getInput();
```
The result can only be seen in the generated HTML. Because of this limitation of fields, we can't check in any other way, except to look for values in HTML.
Before:
``` HTML
<div class="form-check form-check-inline">
	<input
		type="checkbox"
		name="check"
		id="id"
		value="1"
	>
</div>
```
After:
``` HTML
<div class="form-check form-check-inline">
	<input
		type="checkbox"
		name="check"
		id="id"
		value="0" 
	>
</div>
```
Here you need to search for the string **`value ="0"`** with tests, otherwise no way.
In the Checkbox field, any default value is rendering correctly except for 0.
The default value of 0 is rendering as 1.

laoneo:
When you have a look at #37349, there you can see some test instructions how it should be. If the tester has to edit some files, write that in the testing instructions. That's not a problem.

korenevskiy:
> When you have a look at #37349, there you can see some test instructions how it should be. If the tester has to edit some files, write that in the testing instructions. That's not a problem.

You are right, I understood my mistake. I described a new correction indicating the perimers on the principle: as it was before, and how it will become in the future.
I added a description of the PR header .

libraries/src/Form/Field/CheckboxField.php
```diff
@@ -138,7 +138,7 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
 	protected function getLayoutData()
 	{
 		$data            = parent::getLayoutData();
-		$data['value']   = $this->default ?: '1';
+		$data['value']	 = $this->default || $this->default === '0' ? $this->default : '1';
```
Denitz: What about  `$data['value']	 = (string)$this->default === '' ? '1' : $this->default;`

korenevskiy: if `$this->default` will be as `FALSE` or `NULL` ?

Denitz: It can't. But even if any field extends CheckboxField, it's stringified in this code.

korenevskiy: @Denitz  This is only possible when the fields are initialized in PHP itself .
I have long wanted to raise a discussion of the big problem of field objects.
Fields in Joomla are very difficult to use without using XML .
The idea of XML is very good, and XML programming needs to be developed.
But the implementation of XML programming in Joomla has now become a curve and has become a big problem when you need to use Joomla as a Framework library. I want to say that if you do writing a program in a strictly declarative way for simple things, it’s good, it’s convenient. But If you need to take one step aside, the developer falls into the trap of restrictions. And therefore, the use of a declarative approach due to restrictions becomes not possible. I have evidence, and examples of what I say. This is a separate topic of discussion for discussion by community participants.
 But I want to say that for a field object you need to be able to set values in PHP imperative, this will allow developers to do what is not yet provided for by a declarative approach. And a very much more is not provided for by the declarative approach.

HLeithner:

libraries/src/Form/Field/CheckboxField.php
```diff
@@ -137,7 +137,7 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
     protected function getLayoutData()
     {
         $data            = parent::getLayoutData();
-        $data['value']   = $this->default ?: '1';
+        $data['value']   = $this->default || $this->default === '0' ? $this->default : '1';
```
bembelimen: I would recommend the following changes

```suggestion
        $data['value']   = strlen($this->default) ? $this->default : '1';
```

bembelimen (dismissed):

libraries/src/Form/Field/CheckboxField.php
korenevskiy: @bembelimen   Your method is not suitable, since `$this->default` may be undefined and have a `NULL` value.

Message in PHP 8.2:
**Deprecated: strlen(): Passing null to parameter # 1 ($string) of type string is deprecated in `libraries/src/Form/Field/CheckboxField.php` on line 141**

Please cancel the change request.

HLeithner:

korenevskiy:
@HLeithner Is it possible to transfer this PR to Joomla 5 ?

Fedik:
This is a bug, and can go to 4.x, just need to be tested

libraries/src/Form/Field/CheckboxField.php
Fedik: Both is wrong :)
This should be okau:
```php
$data['value'] = $this->default !== null && $this->default !== '' ? $this->default : '1';
```

Fedik (approved):

Fedik:
I have tested this item :white_check_mark: successfully on 6fd6fcb03abb2e18f66c8319320d36df405ad2ec<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37174">issues.joomla.org/tracker/joomla-cms/37174</a>.</sub>

libraries/src/Form/Field/CheckboxField.php
```diff
@@ -137,7 +137,7 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
     protected function getLayoutData()
     {
         $data            = parent::getLayoutData();
-        $data['value']   = $this->default ?: '1';
+        $data['value']   = $this->default !== null && $this->default !== '' ? $this->default : '1';
```
HLeithner, line 141: That's unreadable and doesn't explain on the first look why this construct is so complex.

Reparsing it, or adding a comment would make sense

korenevskiy: `// The explicitly set default value is '0' for using index definitions in arrays and tables of form fields.`

Fedik:
I have tested this item :white_check_mark: successfully on 8b05ef3365f0303c0d226e1d9ada114720310805<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37174">issues.joomla.org/tracker/joomla-cms/37174</a>.</sub>

korenevskiy:
@Fedik How do you think if you set the Default field to FALSE, how should it work? And probably think about in which cases it may turn out that you will need to specify the value FALSE.
```XML
'<field type="check" default="false"  name="id" />'
```

Fedik:
HTML supports only string value, so no boolean values. 
You cannot do `<input type="checkbox" value="false">` and expect it will be a boolean, it will be a string `false`. The same for Joomla XML fields.
so only 1 or 0

korenevskiy:
> so only 1 or 0

You are right, this will be enough for all use cases.

korenevskiy:
@bembelimen Please confirm this PR

HLeithner:

korenevskiy:
@HLeithner 
And when will this PR be merged?

HLeithner:

korenevskiy:
Wow.. 2.5 years have passed.

libraries/src/Form/Field/CheckboxField.php
```diff
@@ -137,7 +137,8 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
     protected function getLayoutData()
     {
         $data            = parent::getLayoutData();
-        $data['value']   = $this->default ?: '1';
+        // The explicitly set default value is '0' for using index definitions in arrays and tables of form fields
```
joomdonation: We should change this comment to something more meaningful. Something like Allow using any none empty string as default value for the field.

joomdonation:
```suggestion
        // Allow using any none empty string like '0' as default value for checkbox
```

libraries/src/Form/Field/CheckboxField.php
```diff
@@ -137,7 +137,8 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
     protected function getLayoutData()
     {
         $data            = parent::getLayoutData();
-        $data['value']   = $this->default ?: '1';
+        // Allow using any none empty string like '0' as default value for checkbox
```
richard67:
```suggestion
        // Allow any non-empty string, such as '0,' to be used as the default value for a checkbox
```

libraries/src/Form/Field/CheckboxField.php
```diff
@@ -137,7 +137,8 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
     protected function getLayoutData()
     {
         $data            = parent::getLayoutData();
-        $data['value']   = $this->default ?: '1';
+        // Allow any non-empty string, such as '0,' to be used as the default value for a checkbox
```
richard67:
```suggestion
        // Allow any non-empty string, such as '0', to be used as the default value for a checkbox
```

joomdonation:
I have tested this item :white_check_mark: successfully on 76a3db2041e6dc4464e118b338854da6787fdb04<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37174">issues.joomla.org/tracker/joomla-cms/37174</a>.</sub>

joomdonation:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37174">issues.joomla.org/tracker/joomla-cms/37174</a>.</sub>

Hackwar:
Thank you!

Issue #36457, opened 2021-12-27
[4.1 ScheduledTasks] No way to reset the statistics


PhilETaylor:
### Steps to reproduce the issue

So lets say my task has been constantly failing and I have 3123 Times Failed. I have fixed the problem and want to reset the stats to that I can keep an eye on it, there is no way to do that without deleting the task and creating a new one. 

<img width="635" alt="Screenshot 2021-12-27 at 23 26 04" src="https://user-images.githubusercontent.com/400092/147513369-47d1d964-5acc-4727-af81-be1244dcec21.png">

@ditsuke

pritam825:
![Screenshot (1962)](https://user-images.githubusercontent.com/54195688/151229923-d3cc8984-82ef-4fa8-8909-87c8040d644d.png)
is this the correct position for the reset button?

ditsuke:
> ![Screenshot (1962)](https://user-images.githubusercontent.com/54195688/151229923-d3cc8984-82ef-4fa8-8909-87c8040d644d.png) is this the correct position for the reset button?

@pritam825 it should go under "Execution History". Also I think the button should be inlined into a field (for example the calendar and user buttons besides "Created" an "Created By".

pritam825:
@ditsuke  is this fine?
![Screenshot (1976)](https://user-images.githubusercontent.com/54195688/151701234-9b5848c2-6298-47fd-a5c0-ed4829757a0c.png)

pritam825:
or you are taking about this, here in the input field there will two values, one is yes and another one is no, and default value will be no, when we will click on yes, then it will so a warning message and after clicking "ok" on the warning button, it resets the history 
![Screenshot (1979)](https://user-images.githubusercontent.com/54195688/151702625-554dc1cf-0f1b-45f5-8836-f0881ba23f72.png)

ditsuke:
@pritam825 take a look at this reset button for user API tokens. I believe we should have a similar radio here. You can have a look at the underlying form XML at `plugins/user/token/forms/token.xml`

![image](https://user-images.githubusercontent.com/72784348/151706093-d901a16e-6791-4657-844f-1a6b479c1d59.png)

Quy:
Or add a reset button next to Time Executed  and Times Failed. See Banners component where this is done.

brianteeman:
Definitely this way is the way to go

[...]

PR #37201, opened 2022-03-04
[4.1] com_scheduler reset task counters

brianteeman:
Pull Request for Issue #36457 .

Add reset button to the counter for times executed and times failed

### Testing Instructions
Apply the pr
Create a task and run it a few times (or update the db directly)
Press the reset button and save the task
Check the task again and the counts have been reset to 0

Not possible

### DIFF
--- a/administrator/components/com_scheduler/src/Field/FailedField.php
+++ b/administrator/components/com_scheduler/src/Field/FailedField.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_scheduler
+ *
+ * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\Component\Scheduler\Administrator\Field;
+
+\defined('_JEXEC') or die;
+
+use Joomla\CMS\Form\FormField;
+use Joomla\CMS\Language\Text;
+
+/**
+ * Count failed field.
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class FailedField extends FormField
+{
+	/**
+	 * The form field type.
+	 *
+	 * @var    string
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $type = 'Failed';
+
+	/**
+	 * Method to get the field input markup.
+	 *
+	 * @return  string  The field input markup.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function getInput()
+	{
+		$onclick = ' onclick="document.getElementById(\'' . $this->id . '\').value=\'0\';"';
+
+		return '<div class="input-group"><input class="form-control" type="text" name="' . $this->name . '" id="' . $this->id . '" value="'
+			. htmlspecialchars($this->value, ENT_COMPAT, 'UTF-8') . '" readonly="readonly">'
+			. '<button type="button" class="btn btn-secondary" ' . $onclick . '>'
+			. '<span class="icon-sync" aria-hidden="true"></span> ' . Text::_('JRESET') . '</button></div>';
+	}
+}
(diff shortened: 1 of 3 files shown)


Quiviro:
I have tested this item :white_check_mark: successfully on a997f980a65c8f7e05524238012ddf24c42e9922<br /><br />It works very well, but I feel that the restart action is too fast and permanent. Could it be affordable to add a confirmation request from the user?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37201">issues.joomla.org/tracker/joomla-cms/37201</a>.</sub>

BertaOctech:
I have tested this item :white_check_mark: successfully on a997f980a65c8f7e05524238012ddf24c42e9922<br /><br />The patch behaved as described.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37201">issues.joomla.org/tracker/joomla-cms/37201</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37201">issues.joomla.org/tracker/joomla-cms/37201</a>.</sub>

richard67:
@brianteeman Does it really need 2 new field types? They have the same code except of the value of the $type variable. Would it not work with one new common type?

brianteeman:
@richard67 It would (i think) but then this is just a copy of the same field in com_banners

richard67:
To me that seems all to be one kind of field which should have a generic name like e.g. ResetableCounter or so. But that's maybe out of scope now.

HLeithner:
It would make more sense to create a generic field "ResetField" and abstract "ClicksField" from it and deprecate the field. Then we don't need to use 3 copies of one field.

HLeithner:
actually we don't need to deprecated the clicksfield we can just simple remove it because it's not part of the b/c policy.

brianteeman:
~closing and will make a new fieldtype~

If someone wants to create a new fieldtype for this then go for it.  Personally I dont see the need

administrator/components/com_scheduler/src/Field/FailedField.php
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * Method to get the field input markup.
+	 *
+	 * @return  string  The field input markup.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function getInput()
+	{
+		$onclick = ' onclick="document.getElementById(\'' . $this->id . '\').value=\'0\';"';
```
dgrammatiko, line 41: @brianteeman can we please NOT use inline JS?

brianteeman: Copy & paste from https://github.com/joomla/joomla-cms/blob/660198833277d4ba0e9e2efd23938081e62fa782/administrator/components/com_banners/src/Field/ClicksField.php#L41

dgrammatiko: Yes I know, this is a thing that the JS Group missed mainly because the data attributes came to Joomla fields very late.
What another approach (which is also way fewer lines of code) could look like:
- in the xml add only one file eg type `reset` with 2 extra data fields one for the function name (eg data-on-click="reset" and another for the value eg data-reset-value="0")
- add if a js file for the field:

```js
[...document.querySelectorAll('button[data-on-click]')].forEach((element) => element.addEventListener('click', (e) => e.target.previousElementSibling.value = parseInt(e.target.dataset.resetValue, 10));
```

One file less...

brianteeman: Go for it

brianteeman:
Closed as stated above. 

please re-open #36457 .

PR #37218, opened 2022-03-06
[4.2] Systemtests: Add MariaDB to potentially supported RDBMS

Hackwar:
### Summary of Changes
Right now we officially support MySQL and Postgres, but not explicitely MariaDB. We are at a point in time where we should add that support officially and since MariaDB and MySQL are deviating from each other more and more, we unfortunately need to add checks for the support of MariaDB to our CI system. This adds MariaDB 10.1 and the latest version with minimally supported PHP version to the CI setup.


laoneo:
Why close?

Hackwar:
The PR didn't lead anywhere and it isn't something I plan to follow up on in the next few months. So I'd rather close it.

PR #37220, opened 2022-03-06
[4] Sort JConfig properties A-Z

PhilETaylor:
### Summary of Changes

When saving Joomla Global Configuration, sort the properties of JConfig A-Z so that configuration.php gets written in a human reader friendly way. 

`SORT_STRING | SORT_FLAG_CASE` is fully compatible with PHP versions Joomla 4 is is targeted at

### Testing Instructions

Save Joomla global config - view contents of configuration.php

Semi-random ordering of properties, db crednetials no where near db host for example

A-Z with natural case flagging (so M and m are close) to make it easier for humans to find the string they are interested in quicker. 

None. Fully backward compatible.


toivo:
I have tested this item :white_check_mark: successfully on e21e9b6172ab84d301131d0244f5b4b64550a53b<br /><br />Tested successfully in 4.1.1-dev of 6 March in Wampserver 3.2.7 using PHP 8.0.15<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37220">issues.joomla.org/tracker/joomla-cms/37220</a>.</sub>

brianteeman:
> Semi-random ordering of properties, db crednetials no where near db host for example

I can not confirm this

image

image

image

brianteeman:
alphabetical sorting is only useful if you already know the name of the variable you are looking for

PhilETaylor:
Go and visit real sites and view their configuration file - you will see I’m right. I open hundreds a month. 

If you are manually editing this file, then there is a good chance that you already know why you are opening the file and you already know what you are looking for!

brianteeman:
> Go and visit real sites and view their configuration file - you will see I’m right. I open hundreds a month.

I pasted the results of three real sites.

The order they are stored in is defined here
 https://github.com/joomla/joomla-cms/blob/8a008244323c3e16180cf811961baf75a88cf3e6/installation/src/Model/ConfigurationModel.php#L389

PhilETaylor:
That file is only used where installing Joomla and is then deleted. That is not the code used when saving Joomla global config

PhilETaylor:
The semi random order is because the order is by luck depending on the other items are added to the $data array, which is also then modified further, and then by converting that array to a Registry object and then ultimately casting of the  registry object to string here

https://github.com/joomla/joomla-cms/blob/8a008244323c3e16180cf811961baf75a88cf3e6/administrator/components/com_config/src/Model/ApplicationModel.php#L953

PR #37245, opened 2022-03-10
webp Support for (deprecated) J3 Image Package

stell:
A lot of people will continue to use Joomla 3 for some time, yet they want to use WebP images. 
This is a small adjustment to let extensions like media manager accomplish that. Our extensions are using the J4 WebP support for resizing and stuff, and this would be good to offer WebP for all the J3 people still around.

Pull Request for Issue # .

### Summary of Changes
webp Support in Image.php

### Testing Instructions
```
$fullpath = 'path/to/image.webp';
$fullpath_resized = 'path/to/image_resized.webp';

try {
    $image = new Image($fullpath);
} catch (Exception $e) {
    return false;
}

$properties = $image->getImageFileProperties($fullpath);

switch ($properties->mime) {
    case 'image/webp':
        $imageType = \IMAGETYPE_WEBP;
        break;
    case 'image/png':
        $imageType = \IMAGETYPE_PNG;
        break;
    case 'image/gif':
        $imageType = \IMAGETYPE_GIF;
        break;
    default:
        $imageType = \IMAGETYPE_JPEG;
}

$image = $image->cropResize(1024, 768);
$image->toFile($fullpath_resized, $imageType, ['quality' => 85]);
```

Attempting to load an image of unsupported type WebP

Successful handling of WebP images

### DIFF
--- a/libraries/vendor/joomla/image/src/Image.php
+++ b/libraries/vendor/joomla/image/src/Image.php
@@ -130,6 +130,7 @@ class Image implements LoggerAwareInterface
 			static::$formats[IMAGETYPE_JPEG] = ($info['JPEG Support']) ? true : false;
 			static::$formats[IMAGETYPE_PNG]  = ($info['PNG Support']) ? true : false;
 			static::$formats[IMAGETYPE_GIF]  = ($info['GIF Read Support']) ? true : false;
+			static::$formats[IMAGETYPE_WEBP]  = ($info['WebP Support']) ? true : false;
 		}
 
 		// If the source input is a resource, set it as the image handle.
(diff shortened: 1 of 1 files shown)


brianteeman:
This is not be complete as you would also need to update the options for the media manager to add webp as a legal file extension and image/webp as a legal mime type.

Actually you can use webp right now with J3 by just updating the options - you just dont get a preview image

stell:
It's not about the media manager, but the possibility to use webp inside the J3 image package.

brianteeman:
> This is a small adjustment to let extensions like media manager accomplish that. 

> not about the media manager,

@zero-24 your decision

stell:
Extensions like media manager means extensions like media manager. So not media manager. 
We use the image package to create thumbnails, and it would be good to handle it the same way as in J4 and not outside the workflow.

swaubhik (commented):
you also need to add a switch case for IMAGETYPE_WEBP at line number 1051

Quy:
Please submit to https://github.com/joomla-framework/image/tree/1.x-dev.

stell:
@Quy Is that package an active dependency and what is 2.0 branch?

brianteeman:
@Quy thats why I said it was up to @zero-24 - that library is archived. If they want it to go into j3 then they need to convince @zero-24 to allow a new feature

PR #37252, opened 2022-03-12
[4.2] Fix wrong installation language select default value

bembelimen:
### Summary of Changes
Fix the language select default value when installing a fresh Joomla!

### Testing Instructions
- Use a browser where "en" is not the default language
- Start the installation process of Joomla!

Although the language is discovered correctly, the "English" is displayed as the selected language:

grafik

Correct language is selected:

grafik


brianteeman:
woohoo!!

brianteeman:
Can we use the same technique to do something that @wilsonge and I talked about for the language installation step which is to display the current installation language as the default option

bembelimen:
> Can we use the same technique to do something that @wilsonge and I talked about for the language installation step which is to display the current installation language as the default option

You mean, when I select de-DE, that this language pack is installed by default (when available) or is on top of the list?

brianteeman:
There are good reasons for it not to be installed by default. For example you are installing in german but the site is for a french client.

I just mean that german is either at the top of the list or it stays in the same position but it is preselected as the default in that list. 

Hope that makes sense

bembelimen:
This method is not needed, to pre-select the installation language, we can just use the value of the dropdown:

```
$model   = new SetupModel;
$options = $model->getOptions();

print_r($options['language']);
```

Then we just have to sort it up [here](https://github.com/joomla/joomla-cms/blob/4.1-dev/installation/src/Model/LanguagesModel.php#L124-L134) based on this value.

brianteeman:
To be honest I never really looked to see how to do it. Something I can work on tomorrow

richard67:
Unrelated to this PR, but I've seen it when testing: Have you noticed that there are 2 entries "English (United Kingdom)" in the language selection? One of them should be "English (New Zealand)", but that's only clear when checking the HTML:

j4 2-test-pr-37252_before_2

Is this a known issue and already reported somewhere?

richard67:
I have tested this item :white_check_mark: successfully on cd2e2dbc534c5198c1351212827eae724045795d<br /><br />Works for me. However I'm not sure yet if the failing drone could be related.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37252">issues.joomla.org/tracker/joomla-cms/37252</a>.</sub>

brianteeman:
no its a bug. If you look at  the xml
https://github.com/joomla/joomla-cms/blob/cb07dbda22c8a8a8f4ced200ff9c079fec8a2f27/installation/language/en-NZ/langmetadata.xml#L3-L14

richard67:
> no its a bug. If you look at the xml

@brianteeman Do you know if this has to be fixed here in the repo or on Crowdin?

richard67:
@bembelimen Drone failing again at the API tests at the installation step. Last time it was the system test at the same place, and the time before again the API test. Always at the installation after having selected the language. It seems it is not always failing, but often. Maybe the tests need a little adjustment?

brianteeman:
> > no its a bug. If you look at the xml
> 
> @brianteeman Do you know if this has to be fixed here in the repo or on Crowdin?

sorry dont  know

richard67:
> > > no its a bug. If you look at the xml
> > 
> > 
> > @brianteeman Do you know if this has to be fixed here in the repo or on Crowdin?
> 
> sorry dont know

I've proposed the fix on Crowdin.

brianteeman:
This is almost perfect but not quite - the problem is when the browser has both a language option and a language+country variant as for example in german

image

If you have one of the variants as your default browser language and joomla has that variant then it works perfectly

image

image

if you have one of the variants as your default browser language and joomla does not have any variants  then it works perfectly

image

image

The _problem_ is if you have one of the pure languages which joomla obviously does not have
OR if you have one of the variants and joomla has variants but not that one.

In this case Joomla offers you the last variant in the list
image

image

image

image



Is this solvable? I suspect its not and I am only reporting this just in case. It is definitely better than it was before this pr

infograf768:
I have tested this item :white_check_mark: successfully on cd2e2dbc534c5198c1351212827eae724045795d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37252">issues.joomla.org/tracker/joomla-cms/37252</a>.</sub>

richard67:
@bembelimen I'd set RTC now as it has 2 tests, but there is still the drone failure of which I'm not sure if it is related.

laoneo:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37252">issues.joomla.org/tracker/joomla-cms/37252</a>.</sub>

richard67:
@laoneo Hmm, we still have the constantly failing installation in drone, due to which I was told I should not set RTC.

When @roland-d would do an update of the 4.2-dev branch to the current 4.1-dev branch (so-called "upmerge"), we could see if the fix in the installation language file for the duplicate "English (United Kingdom)" will help with the tests so it's clear they do not fail due to this PR (and beside that would help me with rebasing some other PRs from 4.1-dev to 4.2-dev).

laoneo:
Ok, restarted drone and it fails again. So time to remove RTC again.

richard67:
Back to pending.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37252">issues.joomla.org/tracker/joomla-cms/37252</a>.</sub>

roland-d:
@richard67 Upmerge done

richard67:
@roland-d Thanks. Will update the branch of this PR here now.

richard67:
Unfortunately it seems that it did not help with the installation failing at the first or sometimes the 2nd api test in drone.

chmst:
Re-set RTC, as tests are still valid.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37252">issues.joomla.org/tracker/joomla-cms/37252</a>.</sub>

brianteeman:
RTC but with an unanswered bug ? https://github.com/joomla/joomla-cms/pull/37252#issuecomment-1066089958

bembelimen:
That bug already exists in 3.x (and should be fixed for sure) but is not related to this PR.
As Joomla! looks only in the first two characters (de-DE => de) or some browser only send the first two characters (de) it takes the first "de" entry from the list. And as de-AT < de-DE it's not the Germany German one which is selected.

brianteeman:
> That bug already exists in 3.x (and should be fixed for sure) but is not related to this PR

But prior to this PR the language is never preselected.

roland-d:
@bembelimen 
> As Joomla! looks only in the first two characters (de-DE => de) or some browser only send the first two characters (de) it takes the first "de" entry from the list.

Where is this done?

bembelimen:
> > That bug already exists in 3.x (and should be fixed for sure) but is not related to this PR
> 
> But prior to this PR the language is never preselected.

In 3.x it is, it was broken in 4.0 (which is fixed with this PR again) but the old behaviour is not improved.

brianteeman:
Ok

bembelimen:
> @bembelimen
> 
> > As Joomla! looks only in the first two characters (de-DE => de) or some browser only send the first two characters (de) it takes the first "de" entry from the list.
> 
> Where is this done?

https://github.com/joomla/joomla-cms/blob/7e2179904446d126ed7938ca1bcd04fc75ff53e3/libraries/src/Language/LanguageHelper.php#L79

laoneo:
Is the change in the composer lock file correct?

roland-d:
Removed RTC as the pull request is still pending.

Hackwar:
Yes, the change in composer lock is correct. That pulls the updated Joomla Browser version, which fixes the issue we had with the tests.

installation/src/Form/Field/Installation/LanguageField.php
```diff
@@ -111,4 +103,50 @@ protected function _sortLanguages($a, $b)
...
+		// Get a forced language if it exists.
+		$forced = $app->getLocalise();
+
+		if (!empty($forced['language']))
+		{
+			$native = $forced['language'];
+		}
+
+		// If a language is already set in the session, use this instead
+		$model   = new SetupModel;
```
laoneo, line 142: As this is a field which can be loaded outside of the booted component, it would be better to load here the model through the boot process.

laoneo: Forget it, didn't realize we are in the installation process.

laoneo:
Then this needs to be fixed, before it can be merged.

Hackwar:
What needs to be fixed?

laoneo:
The conflict

richard67:
I've solved the conflicting composer.lock file. The version from this PR was the right one because newer than the one from #37565 .

RTC is still ok, it doesn't need new human tests as all changes after these were only for the automated tests (Joomla browser update) and clean branch updates.

laoneo:
Thanks!

PR #35967, opened 2021-11-04
[4.1] finder show_search_form

brianteeman:
This if statement (which defaults to true) is pointless as the actual param it is checking does not exist so it can only ever be true

Code Review only to confirm that the param doesn't exist

Pull Request for Issue #35692

PR #37254, opened 2022-03-12
[4.2] Add missing finder parameter

bembelimen:
Replacement for: #35967

### Summary of Changes
Add the missing parameter to the menu/config

### Testing Instructions
- Create a menu item for the finder
- Play around with the parameter "Show search form" in the menu and the config

- Parameter not given

- Parameter there and working


brianteeman:
I have tested this item :red_circle: unsuccessfully on b18576c5af07ed7b754447103c01a6fbb79a4811<br /><br />This option makes no sense to me. 
It requires you to have a search module
If there are NO results then the link to "search again" is broken
Even if you then hide the search again link you would need to have the search module published on the search results page

image

Dont you need to update the db as well with the default value - otherwise the Use Global option doesnt work<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37254">issues.joomla.org/tracker/joomla-cms/37254</a>.</sub>

bembelimen:
Agree, needs more than only the parameters. Can be taken over by s.o. if needed

Issue #37241, opened 2022-03-10
Open_basedir warning in libraries/joomla/form/fields/filelist.php getOptions()


durian808:
### Steps to reproduce the issue
Any code that calls getOptions() in libraries/joomla/form/fields/filelist.php, where $this->directory is NOT a relative path, in my example "/images/virtuemart/payment".

### Expected result
No is_dir() warning

### Actual result
Warning: is_dir(): open_basedir restriction in effect. File(/images/virtuemart/payment) is not within the allowed path(s): (/var/www/vhosts/XXXXX.com/:/tmp/) in /var/www/vhosts/XXXX.com/httpdocs/libraries/joomla/form/fields/filelist.php on line 193

### System information (as much as possible)
Joomla 3.10.6
PHP 7.4.27

### Additional comments
Related to #35895 #36788

**Workaround:**

I replaced

```
       $path = $this->directory;

        if (!is_dir($path))
        {
            $path = JPATH_ROOT . '/' . $path;
        }
```

with

`        $path = JPATH_ROOT . '/' . $this->directory;`

dgrammatiko:
>  in my example "/images/virtuemart/payment".

Your path IS NOT relative, it is absolute. If you want to define a relative directory then it should be `images/virtuemart/payment` (missing the first `/`)

Also you SHOULD NEVER, EVER put executable files (eg .php) inside the images or the media folders from a security point of view...

durian808:
@dgrammatiko Thanks for the correction – I meant to say, "is NOT a relative path", now fixed.  The leading "/" is what causes is_dir to throw the warning.  I think the function is meant to handle both cases, leading "/" and no leading "/".

RE:
> Also you SHOULD NEVER, EVER put executable files (eg .php) inside the images or the media folders from a security point of view...

I don't know why you are mentioning this, because it has no bearing on this bug report, correct?

dgrammatiko:
> it has no bearing on this bug report, correct?

Because it's a potential security issue. The folders `images` and `media` are supposed to **only** contain static assets, nothing `.php`...

durian808:
Dunno why the heck you are saying that.  Are you talking about this .php file?

libraries/joomla/form/fields/filelist.php

This is the code that has the bug on line 193.

Again, what you are saying has no bearing on this bug report.

dgrammatiko:
There is no bug. Your code is wrong, remove the first slash

durian808:
[...]

PR #37261, opened 2022-03-13
[4.1] Change a conditional to check on the first character

dgrammatiko:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/37241#issuecomment-1066110623 .

### Summary of Changes

- Change the conditional from checking if a folder exists to determine if it's a relative or absolute path to checking the first character

### Testing Instructions

- Goto administrator/index.php?option=com_languages&view=language&layout=edit&lang_id=1 and check that the selector for the language images still works
- Edit https://github.com/joomla/joomla-cms/blob/cb07dbda22c8a8a8f4ced200ff9c079fec8a2f27/administrator/components/com_languages/forms/language.xml#L47-L59 to something like (provide a folder outside of Joomla's root that has some files)
```xml
<field
	name="image"
	type="filelist"
	label="COM_LANGUAGES_FIELD_IMAGE_LABEL"
	stripext="1"
	directory="/Users/dgrammatiko/Downloads"
	hide_none="1"
	hide_default="1"
	>
	<option value="">JNONE</option>
</field>
```

- Relative path:
Screenshot 2022-03-13 at 16 15 39

- Absolute path: 
Screenshot 2022-03-13 at 16 14 07

### DIFF
--- a/libraries/src/Form/Field/FilelistField.php
+++ b/libraries/src/Form/Field/FilelistField.php
@@ -187,14 +187,22 @@ class FilelistField extends ListField
 	 */
 	protected function getOptions()
 	{
-		$options = [];
-
-		$path = Path::clean(strpos($this->directory, '/') === 0 ? $this->directory : (JPATH_ROOT . '/' . $this->directory));
+		$path = $this->directory;
 
 		if (!is_dir($path))
 		{
-			return $options;
+			if (is_dir(JPATH_ROOT . '/' . $path))
+			{
+				$path = JPATH_ROOT . '/' . $path;
+			}
+			else
+			{
+				return [];
+			}
 		}
+
+		$path    = Path::clean($path);
+		$options = [];
 
 		// Prepend some default options based on field attributes.
 		if (!$this->hideNone)
--- a/libraries/src/Form/Field/FilelistField.php
+++ b/libraries/src/Form/Field/FilelistField.php
@@ -190,7 +190,7 @@ class FilelistField extends ListField
 		$options = [];
 		$path    = $this->directory;
 
-		if (strpos($path, '/') === 0)
+		if (strpos($path, '/') !== 0)
 		{
 			$path = JPATH_ROOT . '/' . $path;
 		}
--- a/libraries/src/Form/Field/FilelistField.php
+++ b/libraries/src/Form/Field/FilelistField.php
@@ -189,7 +189,14 @@ class FilelistField extends ListField
 	{
 		$options = [];
 
-		$path = Path::clean(strpos($this->directory, '/') === 0 ? $this->directory : (JPATH_ROOT . '/' . $this->directory));
+		$path = $this->directory;
+		
+		if (strpos($path, '/') === 0)
+		{
+			$path = JPATH_ROOT . '/' . $path;
+		}
+		
+		$path = Path::clean($path);
 
 		if (!is_dir($path))
 		{
--- a/libraries/src/Form/Field/FilelistField.php
+++ b/libraries/src/Form/Field/FilelistField.php
@@ -187,22 +187,14 @@ class FilelistField extends ListField
 	 */
 	protected function getOptions()
 	{
-		$path = $this->directory;
-
-		if (!is_dir($path))
-		{
-			if (is_dir(JPATH_ROOT . '/' . $path))
-			{
-				$path = JPATH_ROOT . '/' . $path;
-			}
-			else
-			{
-				return [];
-			}
-		}
-
-		$path    = Path::clean($path);
 		$options = [];
+
+ 		$path = Path::clean(strpos($this->directory, '/') === 0 ? $this->directory : (JPATH_ROOT . '/' . $this->directory));
+
+ 		if (!is_dir($path))
+ 		{
+ 			return $options;
+ 		}
 
 		// Prepend some default options based on field attributes.
 		if (!$this->hideNone)
--- a/libraries/src/Form/Field/FilelistField.php
+++ b/libraries/src/Form/Field/FilelistField.php
@@ -188,9 +188,8 @@ class FilelistField extends ListField
 	protected function getOptions()
 	{
 		$options = [];
+		$path    = $this->directory;
 
-		$path = $this->directory;
-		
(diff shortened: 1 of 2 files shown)


richard67:
@dgrammatiko There is a similar field for folder list where we have slightly different code:

https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Form/Field/FolderlistField.php#L188-L202

But from my point of view those 2 fields FilelistField and FolderlistField should behave in the same way regarding the "directory" option.

Could you change it there, too, in the same way as in this PR here?

@HLeithner Or should we do it differently?

dgrammatiko:
@richard67 done, but I think what @HLeithner was suggesting was to not rely on the folder check for deciding if a folder is absolute or relative. Anyways I'll leave that to you...

richard67:
> @richard67 done, but I think what @HLeithner was suggesting was to not rely on the folder check for deciding if a folder is absolute or relative. Anyways I'll leave that to you...

@dgrammatiko Yes, I understood that, so I thought it should be changed in the FolderlistField.php, too. Now you did it vice versa :-)

richard67:
For testing instructions, the folder list field could be tested e.g. here:

https://github.com/joomla/joomla-cms/blob/4.1-dev/plugins/editors/tinymce/forms/setoptions.xml#L22-L29

dgrammatiko:
@richard67 I think we need a quick test if an absolute folder works on windows, otherwise it's obvious that the choice of `is_dir` for the conditional was done to satisfy that OS

richard67:
> @richard67 I think we need a quick test if an absolute folder works on windows, otherwise it's obvious that the choice of `is_dir` for the conditional was done to satisfy that OS

@dgrammatiko Yes, I know. Unfortunately my Windows environment is not ready yet. Will check later tonight.

libraries/src/Form/Field/FilelistField.php
```diff
@@ -187,17 +187,15 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
 	 */
 	protected function getOptions()
 	{
-		$options = array();
+		$options = [];
 
-		$path = $this->directory;
+		$path = Path::clean(strpos($this->directory, '/') === 0 ? $this->directory : (JPATH_ROOT . '/' . $this->directory));
```
HLeithner: no offense but I'm always impressed how unreadable code can be ;-)

```suggestion
		$path = $this->directory;
		
		if (strpos($path, '/') === 0)
		{
			$path = JPATH_ROOT . '/' . $path;
		}
		
		$path = Path::clean($path);
```

libraries/src/Form/Field/FolderlistField.php
```diff
@@ -183,24 +183,15 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
 	 */
 	protected function getOptions()
 	{
-		$options = array();
+		$options = [];
 
-		$path = $this->directory;
+		$path = Path::clean(strpos($this->directory, '/') === 0 ? $this->directory : (JPATH_ROOT . '/' . $this->directory));
```
HLeithner:
```suggestion
		$path = $this->directory;
		
		if (strpos($path, '/') === 0)
		{
			$path = JPATH_ROOT . '/' . $path;
		}
		
		$path = Path::clean($path);
```

HLeithner (commented):
I think that's better readable

libraries/src/Form/Field/FilelistField.php
```diff
@@ -187,17 +187,21 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
...
 	protected function getOptions()
 	{
-		$options = array();
+		$options = [];
+		$path    = $this->directory;
 
-		$path = $this->directory;
-
-		if (!is_dir($path))
+		if (strpos($path, '/') === 0)
```
richard67:
```suggestion
		if (strpos($path, '/') !== 0)
```
We want to prepend JPATH_ROOT if the path does NOT start with a slash. This is the case if strpos return either false (no slash in the path at all) or something greater than zero (some slash but not at the beginning.

libraries/src/Form/Field/FolderlistField.php
```diff
@@ -183,24 +183,21 @@ public function setup(\SimpleXMLElement $element, $value, $group = null)
...
 	protected function getOptions()
 	{
-		$options = array();
+		$options = [];
+		$path    = $this->directory;
 
-		$path = $this->directory;
-
-		if (!is_dir($path))
+		if (strpos($path, '/') === 0)
```
richard67:
```suggestion
		if (strpos($path, '/') !== 0)
```
We want to prepend JPATH_ROOT if the path does NOT start with a slash. This is the case if strpos return either false (no slash in the path at all) or something greater than zero (some slash but not at the beginning.

HLeithner:
yes sorry my fault thanks richard

dgrammatiko:
5 locs, 50+ commits...

richard67:
> 5 locs, 50+ commits...

@dgrammatiko True team work 😛

richard67:
I don't wanna ruin the party, but ... isn't it some kind of a b/c break? In past, if someone did not have any base dir restriction which blocked that, the check e.g. for `/somefolder` existing in the root failed, so we prepended JPATH_ROOT and treated it like a relative folder, and if that existed all was fine from the user's point of view. Now this will not work anymore.

richard67:
@dgrammatiko @HLeithner There could be indeed a problem on Windows when using a drive letter. When the htdocs folder and the folder used in the absolute path are on the same drive C:, an absolute path like "/Users/Richard/Pictures" works. But "C:/Users/Richard/Pictures" doesn't work anymore because that absolute path starts with a drive letter and not a slash.

Without this PR, an absolute path with a drive letter like "C:/Users/Richard/Pictures" works.

brianteeman:
This PR will not work cross platform

dgrammatiko:
> Without this PR, an absolute path with a drive letter like "C:/Users/Richard/Pictures" works.

Well, there we have our answer why the conditional was using `is_dir`
I think this one served it's purpose, so I'm closing

PR #36860, opened 2022-01-26
[4.2] Enable/disable 'Upload&Install' button on the basis of checkbox value

Shubhamverma2796:
Pull Request for Issue #36274 and #35055 .

### Summary of Changes
Added disabled state to the "UploadAndInstall" button.

### Testing Instructions

1.System-->Update-->Joomla.
2. Click on "Update your site by manually uploading the update package".
3. Button and checkbox are disabled initially.
4. Upload a file greater than 40 MB, checkbox and button are still disabled.
5. upload a file less than 40 MB, the checkbox is enabled. Check in the box, the button gets enabled.
6. again upload a file greater than 40 MB, checkbox and button get disabled again.

### Expected result AFTER applying this Pull Request

https://user-images.githubusercontent.com/54266411/151861745-104601f0-9a7c-4ef7-8479-108c6b4d1f2d.mp4

Issue #36151, opened 2021-11-29
[4.0] Quadruple optin for update


StefanSTS:
### Steps to reproduce the issue

In the update view of Joomla 4 in backend you see an update button.

You cannot press the button until you "opt in" with a tick in the input that the extensions are compatible.
After you tick a popup comes up if you really want to do that. 2nd "opt in".
Now I can press the update button. (But it's not the button for update, it's another opt in.)

Now I come to the next screen that shows an update button. (huh?)
I cannot click the update button, I have to tick another input, 4th "opt in".
There is no popup this time. (Wonderful, but why not, it was there in the screen before, shouldn't it be consistent?)
Now after clicking the 4th "opt in" I can finally click on Update and the button is doing what it says it would.

### Expected result
If I see an update button, I like to see the action performed, otherwise I might want to read something like "Next step to Update". But I would prefer it even shorter.

What I actually expect is that the user is not held for a Dummy and gets one "opt in" and can then press the Update button without being hassled with unnecessary steps.

A hint towards the compatibility of extensions and the backup on one page would suffice.
With the tick of the input the "opt in" process should be good and done.
A popup to ask the user again, what he just now said he wants to do, is Kindergarten. 

### System information (as much as possible)
Came to me on updating two Joomla 4.0.3 to 4.0.4. I have to lie about the backup because for the test page I did not do that. So now I am a liar, or I have to do a backup for a test site where I do not even think about doing a backup. Being a liar is an ugly thing for me, btw, I don't like white or grey lies.

If you think this was a pretty long thing to read for a small issue like that, you exactly feel what I feel.

richard67:
We have had so many issues with people updating when their extensions, especially system plugins, were not compatible, that we saw no other way to make people aware.

Maybe we should offer a kind of subscription where people can opt out and update in one step but will be registered in support forums and here so any issues about updating submitted by those people will directly go into the trash can.

Or we find some 5 or 10 new full time contributors who handle this kind of issues.

StefanSTS:
I am not sure what you are saying.
Means because users had problems updating from Joomla 3 to Joomla 4, we have to go the long road for all tiny updates in the Joomla 4 line?
The other comments are not helpful to take this into a serious conversation.

richard67:
I am saying that it was a deliberate decision to have the check boxes and warnings to make sure people have noticed that they shall make a backup before updating and that they shall check their extensions before updating on a productive site.

joomdonation:
[...]

Issue #35583, opened 2021-09-17
J 4.0.x Mark Field as Required (UX)


coolcat-creations:
### What needs to be fixed
When you want to update or upload a Joomla Package you have a checkbox where you confirm that you made a backup before. This field should be marked as obligatory

### Why this should be fixed
Since no one knows that this field is obligatory it seems that the update Button is not working

### How would you fix it
Solution 1) Put a validation message when not ticked
Solution 2) (less effective) add an asterisk to the checkbox and label it as obligatory

### Side Effects expected
no

PhilETaylor:
I would vote for removing the said checkbox all together. It’s just a useless barrier to functional use and 99% of people will just check it and still not have a backup. It’s a useless roadblock.

coolcat-creations:
yes, would be happy with that too.

Quy:
@bembelimen

Quy:
Duplicate #35055

PR #37270, opened 2022-03-14
Changed the "Start Update" button state on the basis of checkbox value

Shubhamverma2796:
### Summary of Changes

Changed button state on the basis of checkbox value.

### Testing Instructions

1. System-->Update-->Joomla.
2. Check that the "Start button" is disabled by default after applying the patch.
3. Click on the checkbox and make sure that the "Start button" is getting enabled/disabled on the basis of checkbox value.

"Start Button" is always active irrespective of the checkbox value.

"Start Button" gets enabled/disabled on the basis of checkbox value.


richard67:
@Shubhamverma2796 Hasn't this been solved already for 4.2 with your merged PR #36860 ???

Quy:
It is a different page https://github.com/joomla/joomla-cms/pull/36860#issuecomment-1058528465. PR #36860 fixed the manual upload package page.

richard67:
And why is the other one for 4.2-dev and this one for 4.1-dev?

richard67:
And why is it in draft status?

richard67:
@Shubhamverma2796 Is it by purpose that this PR is still in draft status? Or is that by mistake? If the latter, you can change it by using the "Ready for review" button at the bottom of your PR. I think people will not spend time for testing as long as it is in draft status.

Shubhamverma2796:
> @Shubhamverma2796 Is it by purpose that this PR is still in draft status? Or is that by mistake? If the latter, you can change it by using the "Ready for review" button at the bottom of your PR. I think people will not spend time for testing as long as it is in draft status.

Actually I opened the pr but later I realised that it is not the right fix so I changed it to draft. I am working on it.

richard67:
Ah, ok, thanks for the feedback.

coolcat-creations:
Thank you for this PR but could we just remove the checkbox as its useless in my opinion. It just creates a barrier and it will not really educate people to backup.

brianteeman:
>  It just creates a barrier 

As intended so that they stop and think

> and it will not really educate people to backup.

Based on ?

coolcat-creations:
Based on comments here: https://github.com/joomla/joomla-cms/issues/35583 and here: https://github.com/joomla/joomla-cms/issues/36151  (you even liked the suggestion to remove the checkbox back then)

Quy:
> Based on comments here: #35583 and here: #36151 (you even liked the suggestion to remove the checkbox back then)

Please test PR #36531.

Shubhamverma2796:
> Thank you for this PR but could we just remove the checkbox as its useless in my opinion. It just creates a barrier and it will not really educate people to backup.

so should I remove the checkbox?

richard67:
> > Thank you for this PR but could we just remove the checkbox as its useless in my opinion. It just creates a barrier and it will not really educate people to backup.
> 
> so should I remove the checkbox?

@Shubhamverma2796 As long as it is just an opinion in a discussion and not supported by a huge number of people, there is no need to remove the check box.

@coolcat-creations Would Nik's PR #36531 be a way? Beside other changes it makes the presence of the check boxes configurable.

coolcat-creations:
@richard67 I have to check - it's a long read , thank you for directing me to this PR :)

Quy:
@Shubhamverma2796 This probably should be rebased for v4.2 as @richard67 stated. Also why not use `disabled` attribute like you did with #36860?

Shubhamverma2796:
Not using disabled attribute because :

1. There is no disabled attribute for anchor tag like this(a href="#" disabled") .    Therefore, in order to use the disabled attribute, I will need to change the anchor tag to the button tag, which will cause problems for other empty states. For example: "Add your first article " will stop working.
2. Even if I do use the button tag and add the disabled attribute, then initially, I need to set disabled=true because initially, the checkbox will be unchecked so that the button must be disabled. Again, this will cause problems for other states, ex: the "Add your first article" button will become disabled making it inaccessible.

 I first used the disabled attribute but later faced these problems therefore added the disabled to the class only. I took all the steps as per my understanding, please correct me if I'm wrong.

build/media_source/com_joomlaupdate/js/default.es6.js
```diff
@@ -53,6 +54,18 @@ Joomla = window.Joomla || {};
     if (uploadButton) {
       uploadButton.addEventListener('click', Joomla.submitbuttonUpload);
     }
+    if (confirmButton && !updateCheck.checked) {
+      confirmButton.className = 'btn btn-primary btn-lg px-4 me-sm-3 emptystate-btnadd disabled';
+    }
+    if (confirmButton && updateCheck) {
+      updateCheck.addEventListener('change', () => {
+        if (updateCheck.checked) {
+          confirmButton.className = 'btn btn-primary btn-lg px-4 me-sm-3 emptystate-btnadd';
```
Quy: Please use `classList.remove` and `classList.add` to remove/add `disabled` rather than hardcoding/repeating these classes.

Quy:
I have tested this item :white_check_mark: successfully on fbf3353c096b5dac3570243f382c9c3978a89048<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37270">issues.joomla.org/tracker/joomla-cms/37270</a>.</sub>

pritam825:
I have tested this item :white_check_mark: successfully on fbf3353c096b5dac3570243f382c9c3978a89048<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37270">issues.joomla.org/tracker/joomla-cms/37270</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37270">issues.joomla.org/tracker/joomla-cms/37270</a>.</sub>

roland-d:
Thanks everybody

Issue #36448, opened 2021-12-27
[5] Deprecated `Factory::getUser()` is not the same as `Factory::getApplication()->getIdentity()`


PhilETaylor:
Ok @wilsonge you up for this challenge? 

when a user is not logged in 

`Factory::getApplication()->getIdentity()` = `null`

whereas 

`Factory::getUser()` = an empty `Joomla\CMS\User\User` object 

!!! 

The back end is fine - its only the frontend having issues, PluginHelper and SiteApplication seem to be the two files that need touching but I cannot seem to find the "right way" to do this. 

It seems that Joomla is highly dependant on getUser returning a empty User class when not logged in... so that `->getAuthorisedLevels()` and other methods can be run on a guest user. 

The deprecation message therefore is wrong as the two methods are not the same. `getUser` loads from `self::getApplication()->getSession()->get('user')` and `getIdentify` just loads from the class property - and is null if the user is logged out. 

Or is the correct "fix" to refactor most of the frontend, and all the 3PD extensions so that they check `getIdentity` and see if its null or not before using it? ... but then you could have ACL on your logged out users - but as your logged out user is NULL and not a blank user object that levels can be checked on... you are screwed... haha. 

Until this is resolved, the use of `Factory::getApplication()->getIdentity()` on the frontend, where users could be logged out, should be minimised...

Installation is also broken (unit tests show that)

_Originally posted by @PhilETaylor in https://github.com/joomla/joomla-cms/issues/36447#issuecomment-1001780350_

codeling:
This is still the case in the released 5.0.0 I presume? Then I guess this should be prominently mentioned in https://manual.joomla.org/migrations/44-50/ ...?

PR #30834, opened 2020-10-01
[4.3] Message after a login / logout frontend

sandewt:
Pull Request for Issue # .

New feature, see #30767

### Summary of Changes

There is a message after a successfully login and logout in the frontend.
### Messages
new_feature_messages_3
new_feature_messages_2
[EDIT] Removed _successfully_ from the text in message.

### Users: Options
new_feature_messages_4
[EDIT] Removed

### Testing Instructions

Backend: Go to Users > Manage > Options
(Frontend: Is not applicable.)

Set the options:
- Frontend Login Message (= Yes, default = No)
- Frontend Logout Message (= Yes, default = No)

Create a menu link:
- Menu Login form
- Menu Logout form

1) Log in and out, in case of a:
- Login form (default)
- Menu Login form
- Menu Logout form

2) Code review (developers);

- Check if the code is correct and secure.

### Actual result BEFORE applying this Pull Request

There is NOT showing a message after a suscessfully login or logout.

### Expected result AFTER applying this Pull Request

- There is showing a message after a suscessfully login. 
- There is showing a message after a suscessfully logout.

### Documentation Changes Required

Yes.

PR #37321, opened 2022-03-19
Refactored Factory::getUser() to ::getApplication()->getIdentity()

pe7er:
Joomla 4.1 uses the deprecated ``Factory::getUser()``. I've changed it in a couple of files to ``Factory::getApplication()->getIdentity()``. 

- Does it make sense to manually replace it like this?
- Or will it be done automatically with a global search & replace?
- Please advice me if I could continue to refactor it like this in other files as well?

### Summary of Changes
This PR changes ``Factory::getUser`` in 5 files to ``Factory::getApplication()->getIdentity``.
In some cases I added Exception to the code block and as name space.

``$user   = Factory::getUser($userId);``
to
``$user   = Factory::getApplication()->getIdentity($userId);``

``$user = Factory::getUser();``
to
``$user   = Factory::getApplication()->getIdentity();``

``if (!Factory::getUser()->authorise('core.admin'))``
to
``if (!Factory::getApplication()->getIdentity()->authorise('core.admin'))``

### Testing Instructions

Please do a code inspection. 
It should not change any functionality and Joomla should continue work correctly.


administrator/components/com_actionlogs/src/Plugin/ActionLogPlugin.php
```diff
@@ -50,18 +51,19 @@ abstract class ActionLogPlugin extends CMSPlugin
...
 	 * This method adds a record to #__action_logs contains (message_language_key, message, date, context, user)
 	 *
-	 * @param   array   $messages            The contents of the messages to be logged
-	 * @param   string  $messageLanguageKey  The language key of the message
-	 * @param   string  $context             The context of the content passed to the plugin
-	 * @param   int     $userId              ID of user perform the action, usually ID of current logged in user
+	 * @param array $messages The contents of the messages to be logged
+	 * @param string $messageLanguageKey The language key of the message
+	 * @param string $context The context of the content passed to the plugin
+	 * @param int $userId ID of user perform the action, usually ID of current logged in user
```
Quy, line 57: Please revert.

laoneo:
Thanks for your pr. Honestly I would like to see a change to introduce a user service which can be injected into the view, to get a user from rather exchanging a static function call with another.

sandewt:
A small addition:

`Factory::getUser()->id`
or
`Factory::getUser()->get('id')`
to
`Factory::getApplication()->getIdentity()->id`

sandewt:
>It should not change any functionality and Joomla should continue work correctly.

Point of interest:
[https://github.com/joomla/joomla-cms/pull/30834#issuecomment-1067956161](https://github.com/joomla/joomla-cms/pull/30834#issuecomment-1067956161) and
[https://github.com/joomla/joomla-cms/pull/30834#issuecomment-1070383253](https://github.com/joomla/joomla-cms/pull/30834#issuecomment-1070383253)

wilsonge:
Generally speaking this is OK. Worth noting that behaviours aren't exactly identical as noted in https://github.com/joomla/joomla-cms/issues/36448 so you can't do a full auto-replace in an IDE.

@laoneo HTMLView's at least will also be hard coupled to the application for the active template (https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/MVC/View/HtmlView.php#L377). So unless we plan to start injecting that whilst obviously the statics aren't ideal - it might be that we inject the application into the view in the constructor going forwards anyhow. Even if we inject the user because there's going to be authentication checks throughout the view probably makes more sense in the constructor than as a service there anyhow.

laoneo:
What about the [UserFactory](https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/User/UserFactoryInterface.php)?

joomdonation:
@pe7er You might want to close this PR because we now use a different approach:

- PR https://github.com/joomla/joomla-cms/pull/37498 introduced a new way to get current user object in views and models class
- The Factory::getUser() replacement on view classes was merged https://github.com/joomla/joomla-cms/pull/38006
- There is also a PR made to replace Factory::getUser() in model classes, too, see https://github.com/joomla/joomla-cms/pull/38090

administrator/components/com_actionlogs/src/Model/ActionlogModel.php
```diff
@@ -44,7 +44,7 @@ class ActionlogModel extends BaseDatabaseModel
 	 */
 	public function addLog($messages, $messageLanguageKey, $context, $userId = null)
 	{
-		$user   = Factory::getUser($userId);
+		$user   = Factory::getApplication()->getIdentity($userId);
```
joomdonation, line 47: Just for information the change here is not right. The getIdentity method does not accept any parameter like you passed here. If we want to replace it, we will need to use UserFactory, something like:

```php
$user = Factory::getContainer()->get(UserFactoryInterface::class)->loadUserById($userId);
```

laoneo:
Agree, this one should be closed as users will be injected into views and models.

HLeithner:

pe7er:
@joomdonation thanks for your explanation and for your PR to solve this in a better way
@laoneo thanks for your PR to solve this in a better way

closing this one...

Issue #36395, opened 2021-12-24
PHP 8 deprecations to handle 


PhilETaylor:
UserHelper.php and Date.php

<img width="1982" alt="Screenshot 2021-12-24 at 17 52 52" src="https://user-images.githubusercontent.com/400092/147367744-21bbbec0-05fb-4a93-828a-a112f1467068.png">

```
  <b>Deprecated</b>:  str_replace(): Passing null to parameter #1 ($search) of type array|string is deprecated in <b>/tests/www/postgres/libraries/src/User/UserHelper.php</b> on line <b>565</b><br />

272	  <b>Deprecated</b>:  DateTime::__construct(): Passing null to parameter #1 ($datetime) of type string is deprecated in <b>/tests/www/postgres/libraries/src/Date/Date.php</b> on line <b>126</b><br />
40s
273	  <br />
```

PR #37329, opened 2022-03-20
[4.1] Use the date function in calendar field

laoneo:
Pull Request for Issue #34952 and takeover from #36395.

### Summary of Changes
The calendar field uses the `strftime` to format a date. This function got deprecated in PHP 8.1 and will be removed in 9. This pr uses the widely accepted `date` function.
There are probably edge cases for format conversion, which are not covered by this pr, but can be added later.

Full credit goes to @PhilETaylor, I was just taking it over and made some small modifications.

### Testing Instructions
- Create a new article in the back end
- Click on the publishing tab

The following message is displayed:
_Deprecated: Function strftime() is deprecated in /libraries/src/Form/Field/CalendarField.php on line 277_

No error message.


tecpromotion:
I have tested this item :white_check_mark: successfully on fca7309d9d51c5e3723f39dd4380e28988115437<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37329">issues.joomla.org/tracker/joomla-cms/37329</a>.</sub>

brianteeman:
Surely this has to go in 4.2

libraries/src/Form/Field/CalendarField.php
```diff
@@ -274,7 +276,8 @@ protected function getInput()
 		{
 			$tz = date_default_timezone_get();
 			date_default_timezone_set('UTC');
-			$this->value = strftime($this->format, strtotime($this->value));
+			$date = DateTimeImmutable::createFromFormat('U', strtotime($this->value));
+			$this->value = $date->format(HTMLHelper::strftimeFormatToDateFormat($this->format));
```
joomdonation, line 280: I will have to check it again, however, in many cases (at least for Calendar Form Field uses in core), we have the `DateFormat` stored in `filterFormat` property already, and to be safe, we do not need to do magic conversion using `strftimeFormatToDateFormat`. Something like below might work better (and more safe)

```php
if ($this->filterFormat)
{
     $this->value = $date->format($this->filterFormat);
}
else 
{
     $this->value = $date->format(HTMLHelper::strftimeFormatToDateFormat($this->format));
}
```

HusiBT:
I have tested this item :white_check_mark: successfully on fca7309d9d51c5e3723f39dd4380e28988115437<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37329">issues.joomla.org/tracker/joomla-cms/37329</a>.</sub>

laoneo:
rtc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37329">issues.joomla.org/tracker/joomla-cms/37329</a>.</sub>

laoneo:
rtc<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37329">issues.joomla.org/tracker/joomla-cms/37329</a>.</sub>

MacJoom:
I have tested this item :white_check_mark: successfully on fca7309d9d51c5e3723f39dd4380e28988115437<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37329">issues.joomla.org/tracker/joomla-cms/37329</a>.</sub>

bembelimen:
> Surely this has to go in 4.2

Yes, or 5.0

laoneo:
If this has to go into 5, then Joomla 4 will never be able to run on PHP 9. If it has to go into 4.2, then I suggest an alternative parameter where extension developers can define a format in normal [PHP date format](https://www.php.net/manual/en/datetime.format.php). Something like `date_format`. So the tag would look like:
```
<field
	name="begin"
	type="calendar"
	label="COM_BANNERS_BEGIN_LABEL"
	hint="COM_BANNERS_BEGIN_HINT"
	format="%Y-%m-%d"
	date_format="Y-m-d"
	filter="user_utc"
/>
```
Extension developers can then still use the normal format parameter to keep backwards compatibility with Joomla 3.

What do you guys think?

brianteeman:
If you go down the route of the extra date_format then you should also update all date fields in joomla so that they are using the new way and nothing in core is using the old and presumably deprecated method.

@nibra See I think I understand now your comments in the RFC

joomdonation:
Actually, we can use a _hidden_ parameter **filterFormat** https://github.com/joomla/joomla-cms/blob/f019ec961fd0df8a88233d8cb500968310dbf462/libraries/src/Form/Field/CalendarField.php#L118 . It is not documented, but used on all of our calendar form fields when **translateformat="true"** , see https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Form/Field/CalendarField.php#L207

So Calendar Form fields from core will still work as how it is. For third party extensions, we can just add **filterFormat** attribute with the right value and it is working OK, too. If the attribute is not available, we can use the code to convert format like how we are doing in this PR (maybe just for PHP >= 8.1.0)

laoneo:
@joomdonation can you make an alternative pr with the `filterFormat`? But please make it compatible with all PHP versions.

joomdonation:
@laoneo If it is OK to make PR without writing unit test, I will give it a try (I don't know how to write unit tests at the moment). Also, I will have to borrow the code of **strftimeFormatToDateFormat** method in this PR.

laoneo:
Sure

laoneo:
CLosing in favor of #37373.

PR #37330, opened 2022-03-20
[4.2] (re) Introduce a body end renderer

### LINKED ISSUE #37297: [4.x] Modal dialogs don't appear if there are two editors on a page in different tabs when using Editor - None/Editor - CodeMirror
1. Set default editor to Editor - None or Editor - CodeMirror
2. Create a custom module
3. In the .xml file, create two separate `<fieldset>` with a `<field type="editor" />` each
4. Open the module in the backend to edit, swap to the second tab and try to open any editor extensions (ie. Media)

### Expected result
The modal dialog appears

### Actual result
The modal backdrop appears, but the modal dialog does not.

### Additional comments
When inspecting the DOM elements, the issue appears to be that the modal dialog elements are placed in the first tab, and since tabs have a `display:none` when inactive, the dialog itself is also hidden.

I tested with TinyMCE, something about how that one is setup/instantiated has it either fix itself or just handle the modal dialogs differently.

dgrammatiko:
Pull Request is the base for many Issues associated with modals. Fixes #37297

### Summary of Changes
- Introduces another renderer
- Adds setter/getter for the enqueued html fragments

#### Backend Toolbars

Try to edit an article and check the source of the document (scroll to the end) [preferably with debug disabled]
Apply the PR
Check again the source, the last entries before < /body > should be some modals
Check that the toolbar buttons versions, preview and a11y checks are all functionals

#### Backend Multiple editors in different tabs #37297

Ensure the global editor is set to 'None'
Edit `administrator/components/com_content/forms/article.xml` and add 

```xml
<field name="xoxo" type="editor" editor="none" buttons="true" />
```

After line 264. Try to check if both editors have functional buttons

Apply the PR and repeat the test

Repeat the test for code mirror

Chunks of HTML can only be echoed sequentially, eg the toolbar modals are echoed immediately after the button 

Screenshot 2022-03-20 at 15 13 24

Chunks of HTML can be enqueued to be echoed at the end of the document

Screenshot 2022-03-20 at 15 12 41

@laoneo this is needed if you want to see any improvement on the modals

@roland-d in order to solve https://github.com/joomla/joomla-cms/issues/37297 and a few more modal related issues this (or something similar) is required


libraries/src/Document/Document.php
```diff
@@ -262,6 +262,14 @@ class Document
 	 */
 	protected $webAssetManager = null;
 
+		/**
```
brianteeman:
```suggestion
	/**
```

dgrammatiko:
There's nothing in their docs: https://www.php.net/manual/en/function.serialize.php
But probably I could do this in another way

alikon:
FYI https://php.watch/versions/8.1/serializable-deprecated

dgrammatiko:
@brianteeman @alikon that article is talking about implementing an own serialised/unserializer and that for 8.1 it needs both the old and the new functions. But I'm not using `implements` here. So is this an issue here?

layouts/joomla/editors/buttons.php
```diff
@@ -16,7 +16,7 @@
 ?>
 <div class="editor-xtd-buttons" role="toolbar" aria-label="<?php echo Text::_('JTOOLBAR'); ?>">
 	<?php if ($buttons) : ?>
-		<?php foreach ($buttons as $button) : ?>
+		<?php foreach (array_map('unserialize', array_unique(array_map('serialize', $buttons))) as $button) : ?>
```
laoneo: Is this really needed like that?

dgrammatiko: The problem is that button get added as the event fires multiple times and the results are accumulating. Probably this is the wrong solution, the right thing is to fix the particular event to return **only** the buttons for the requested call or don't use the event system at all for the buttons...

laoneo: ok

libraries/src/Document/Renderer/Html/BodyendRenderer.php
```diff
@@ -0,0 +1,37 @@
...
+	 *
+	 * @param   string  $name     Not used.
+	 * @param   array   $params   Associative array of values
+	 * @param   string  $content  Not used.
+	 *
+	 * @return  string  The output of the script
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function render($name, $params = array(), $content = null)
```
Quy:
```suggestion
	public function render($name, $params = [], $content = null)
```

plugins/editors/none/none.php
```diff
@@ -91,7 +91,7 @@ public function _displayButtons($name, $buttons, $asset, $author)
 			$buttonsEvent = new Event(
 				'getButtons',
 				[
-					'editor'  => $name,
+					'editor'    => $name,
```
Quy: This is no longer aligned.

dgrammatiko: I know but it's irrelevant to this PR

dgrammatiko:
@SharkyKZ why the thumbs down?

laoneo:
Why did you close?

Issue #37313, opened 2022-03-18
Schedules Tasks


astridx:
### Steps to reproduce the issue

1. Create Demo Task - Sleep 
2. Select `Cron Expression Advanced` as `Execution Rule`
3. Do not select all values in area `Crone match`.

### Expected result

You see a message about what you have to select.

### Actual result

You see 

![m](https://user-images.githubusercontent.com/9974686/159076411-820aa3ec-f089-43b8-b11e-65da0d8a9d02.png)

richard67:
Could be a good first task for a potential GSoC student.

keshavgbpecdelhi:
Can I go ahead with this?

richard67:
> Can I go ahead with this?

 No need for asking. Just do it.

keshavgbpecdelhi:
I was trying to set up this , but the command not found popped up on composer. Need Assistance.

![image](https://user-images.githubusercontent.com/56075233/159124049-36b39429-ab0c-44ad-b373-b805c69a1943.png)

richard67:
@keshavgbpecdelhi Well, you have to install the Composer software first before you can use it. Same later for node.js before you can use npm. It does not come with a standard Webserver software.

keshavgbpecdelhi:
I have done the installation part, can you please tell how to go to the above mentioned page (the path).
I can see this :
![image](https://user-images.githubusercontent.com/56075233/159126717-adc2c825-c875-4985-9b7c-0921bd0e14de.png)

richard67:
@keshavgbpecdelhi The folder shown in your screenshot should be the root of a local webserver or a subfoder of that root. You then just browser the URL for that local web site, e.g. "http://localhost" if in the webserver root, or "http://localhost/somesubfolder" if in a sub folder.

But honestly: This here is an issue tracker for Joomla issues and not a support forum for teaching how to install Joomla.

If you are a potential GSoC 2022 student, please join our channels on Glip (RingCentral) and put your questions there, so we do not pollute this issue with installation questions and other students can benefit from the answers, too.

Thanks in advance.

teekhandekar:
@astridx Can I work on this ?

Quy:
Please test PR #37350.

bembelimen:
Reopened as the PR is on hold.

richard67:
Closing as having a pull request. Please test #37536 . Thanks in advance.

PR #37350, opened 2022-03-22
Schedules tasks

vorayash:
Pull Request for Issue Schedules Tasks #37313 

### Summary of Changes
Changed "joomla-cms\administrator\components\com_scheduler\forms\task.xml" by making cron fields required so that if user tries to save task without checking all the field he/she will get warning.

### Testing Instructions

1. Create Demo Task - Sleep
2. Select Cron Expression Advanced as Execution Rule
3. Do not select all values in area Crone match.

image

image


richard67:
What happens if you chose another method than cron expression in the drop down? The cron expression fields should not be required in this case.

vorayash:
I have removed required from cron expression and have modified FormField.php. Please check.

richard67:
@vorayash Your PR has code style errors. You can see that when going to the bottom of the PR and see the drone CI is failing. With the "Details" link you can get to the details.

2022-03-23_pr-37350_drone_1

When following to that link you come to a page where on the left hand side you can see the failed PHPCS test. By selecting that test, the right hand side shows the log from the test, which tells the details about the code style errors:

2022-03-23_pr-37350_drone_2

Could you fix them? If you need help let me know, then I will propose a code change which you can accept with one mouse click. But I wanted to give you the chance first to fix them yourself.

Thanks in advance.

richard67:
P.S.: You should adjust your text editor or IDE editor so that it shows you white space like spaces and tabulators. This helps to avoid the white space at the end of line errors.

vorayash:
I will surely do it. Thank you sir for your support.

vorayash:
Please check this.

libraries/src/Form/FormField.php
```diff
@@ -1287,6 +1287,21 @@ public function validate($value, $group = null, Registry $input = null)
 			return new \UnexpectedValueException($message);
 		}
 
+		// Get execution rules and make it array
+		$execution_rules = (array) $input->get('execution_rules');
+
+		// Check if execution rule is cron-expression
+		if ($execution_rules['rule-type'] === "cron-expression")
```
```suggestion
		if ($execution_rules['rule-type'] === 'cron-expression')
```

richard67:
I think the testing instructions are not complete. It does not only need to test if the PR does what is required for the task form where it has a field set "execution_rules", it also needs to be tested that nothing is broken in forms which use the FormField class or a class derived from it and do NOT have that field set.

I think in this case we will get PHP errors related to unset array elements, e.g. `$execution_rules['rule-type']` or `$execution_rules['cron-expression']` and so on.

I haven't tested this PR yet, but I am pretty sure it will raise these PHP errors when doing the additional test which I proposed above.

richard67:
Another question is if it is a good idea to place the checks for this fieldset which only exists in that particular form in a class which is base of many other types of fields used in many other types of forms, so the check is performed in 99.99 % of all cases for nothing.

Maybe it should be placed in the component's model instead, or in the table class for the task table (if we have such, I haven't checked)?

richard67:
I have tested this item :red_circle: unsuccessfully on 6b307709c04f36c4d89e1fc7a9e13e07b84e3a56<br /><br />Saving other forms in backend causes PHP Notice "undefined index". The error is just not visible in backend, but when you check your PHP error log file you will see lots of the following entries (one for each field) when saving any form in the backend, e.g. Global Configuration or Article Edit:

> PHP Notice:  Undefined index: rule-type in /home/richard/lamp/public_html/joomla-cms-4.1-dev/libraries/src/Form/FormField.php on line 1294, referer: https://www.joomla-41-dev.vmubu01.vmnet2.local/administrator/index.php?option=com_config

Another question is if it is really clever to check at validation of (almost) every field for the cron related specials.

These checks belong into either the model of the corresponding component https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_scheduler/src/Model/TaskModel.php or in the table class of the task table https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_scheduler/src/Table/TaskTable.php .
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37350">issues.joomla.org/tracker/joomla-cms/37350</a>.</sub>

laoneo:
Catching the rule error must be added to the scheduler component and not the `FormField` class as it should not be aware about any context. Correctly the controller must pass the valid data to the model as the model and the table should not depend on the input either (I know the model has populatestate where we are suffering since ages from it).

vorayash:
Thank you for your guidance. I will solve this issues.

ditsuke:
@vorayash are you still working on this?

vorayash:
Sir I am right now busy with my GSOC task. Is it possible that after completing it I work on this?

bembelimen:
I'm closing this as this is, as stated above, the wrong approach.

If you are available again, please feel free to create a new PR.

vorayash:
👍

vorayash:
Ok sir.

Issue #37312, opened 2022-03-18
Feature request – “Save & Do Not Close” button (Apply) for frontend article editing


oorzaak:
### Steps to reproduce the issue

– Create or edit an article using frontend article editing, in a Joomla 4 core installation.

– Try to save the article without closing the edit view.

### Expected result

When editing an article in the backend, you can choose between a “Save” button and a “Save & Close” button. This adds to a good user experience. It would make sense to have both options in the frontend as well.

### Actual result

In the frontend, you only have one button, which has the label “Save” but in fact acts as “Save & Close”. There is no way to save and stay in the edit view.

### System information (as much as possible)

Joomla 4.1.0

### Additional comments

I had several customers asking for a “Save & do NOT Close” button in the frontend. 

I suppose that this corresponds with the task: article.apply

This also applies to J.3 but of course it would be most effective if this is added to J.4.

BhavyaT-135:
Hey, @oorzaak Can I work on this issue?? 😃

richard67:
> Hey, @oorzaak Can I work on this issue?? 😃

No need to ask. Just do it.

BhavyaT-135:
> > Hey, @oorzaak Can I work on this issue?? 😃
> 
> No need to ask. Just do it.

I followed the steps mentioned above and created a sample article. Although, I didn't have any problem using either the "Save" or the "Save & Close" button. They seem to work fine :)

richard67:
I guess you have tested in backend but the issue here is for frontend editing.

roland-d:
@BhavyaT-135 Could you check the front-end and propose a pull request? I would love to have this in version 4.2.

oorzaak:
> I guess you have tested in backend but the issue here is for frontend editing.

I mean frontend editing, indeed. I'm sure that many people would be happy to have this feature :smile:<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37312">issues.joomla.org/tracker/joomla-cms/37312</a>.</sub>

BhavyaT-135:
> @BhavyaT-135 Could you check the front-end and propose a pull request? I would love to have this in version 4.2.

Sure sir,  I'll give it a shot.

richard67:
Closing as having a pull request. Please test #37362 . Thanks in advance.

PR #37354, opened 2022-03-22
Fix article save2copy not unpublished

kitepascal:
Pull Request for Issue #37351 .

### Testing Instructions
Save as Copy

### Actual result BEFORE applying this Pull Request
Article is published.

### Expected result AFTER applying this Pull Request
Article shoud be unpublished like in J3 before.

### Documentation Changes Required
None.

Fixes #37351

PR #37362, opened 2022-03-24
Add ability to save an article from the frontend without being redirected (#37312)

magnussinger:
Pull Request for Issue #37312.

### Summary of Changes
Added a new button to the article edit form on the frontend. It allows the user to save the article without closing the form. This corresponds with the functionality of the backend.

### Testing Instructions
Edit an article from the frontend, the button will be on the bottom and you can save the article without being redirected

Only the ability to save the article and get redirected to it right afterward

You can choose to only save the article and stay inside the edit form

None


ChristineWk:
I have tested this item :white_check_mark: successfully on 6ffe5654ba237993d1798ffbc2ddb107d18a00c3<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37362">issues.joomla.org/tracker/joomla-cms/37362</a>.</sub>

components/com_content/tmpl/form/edit.php
```diff
@@ -157,10 +157,14 @@
...
 		</fieldset>
 		<div class="mb-2">
-			<button type="button" class="btn btn-primary" data-submit-task="article.save">
+			<button type="button" class="btn btn-primary" data-submit-task="article.apply">
 				<span class="icon-check" aria-hidden="true"></span>
 				<?php echo Text::_('JSAVE'); ?>
 			</button>
+			<button type="button" class="btn btn-primary" data-submit-task="article.save">
+				<span class="icon-check" aria-hidden="true"></span>
+				<?php echo Text::_('JSAVE'); ?> & <?php echo Text::_('JCLOSE'); ?>
```
richard67: We should not hard code the "&" in PHP since it might not be suitable for all languages.

The language string used for the toolbar in backend is this one:
https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/language/en-GB/joomla.ini#L802
> JTOOLBAR_SAVE="Save &amp; Close"

For api we have it also: https://github.com/joomla/joomla-cms/blob/4.1-dev/api/language/en-GB/joomla.ini#L797

But we don't have it for the frontend: https://github.com/joomla/joomla-cms/blob/4.1-dev/language/en-GB/joomla.ini

I don't think we should add it with key "JTOOLBAR_SAVE" because in frontend it is not a toolbar.

In the backend language file we have also JSAVE="Save &amp; Close".

In the frontend language file we have JSAVE="Save".

In the backend and frontend language files we have JAPPLY="Save".

So that's not really consistent but let me say historically grown.

We can do either one of the 2 things now:

1. We change the language string JSAVE to "Save &amp; Close" in the frontend language file so it is consistent with the backend key of the same name, make the "Save" button use the JAPPLY string and the new Save and Close button use the (modified) JSAVE language string. This would result in consistency between the backend and frontend JSAVE language strings and how this string and the JAPPLY  is used. But it means we change an existing language string's value, i.e. change translation. If translation teams of other languages miss that change or are coming too late with it, we might still have both string JSAVE and JAPPLY haveing the same value "Save" in these languages, and so both buttons having the same text.

2. We add a new string with a new key, e.g. JSAVEANDCLOSE or whater, to the frontend language file, and this has value "Save &amp; Close", and the new button uses this string. This has the advantage that there might not be the above confusion. Languages which don't have the new string will show the untranslated text instead.

Personally I think we have to go the 2nd way.

@brianteeman What do you think?

richard67: Other opinions by others are also welcome of course.

magnussinger: I agree with you. I think doing it the 2nd way is better, I can update it if you want

richard67:
@bembelimen Is this a new feature and so has to go into 4.2? Or can it be considered a progressive enhancement so it's ok for 4.1.x (likely 4.1.2)?

components/com_content/tmpl/form/edit.php
richard67: Personally I would wish we could go the 1st way, but I know we can't. So yes, hey ho, let's go for 2.

magnussinger: All right, I'll update it asap

brianteeman: @richard67 in theory the joomla.ini file should be the same across all three applications

richard67: > @richard67 in theory the joomla.ini file should be the same across all three applications

@brianteeman Yes, but we know it isn't. See the above differences for JSAVE. So what do you suggest? Way 1 or way 2?

bembelimen:
4.2 please

richard67:
@MagnusSinger Could you rebase the PR to 4.2-dev? Let us know if we shall help with that.

magnussinger:
@richard67 yes sure

components/com_content/tmpl/form/edit.php
roland-d: @richard67 What I miss is the option to add the language string to the com_content language file instead of the joomla.ini. In that case we could do `COM_CONTENT_ARTICLE_SAVE_CLOSE` or something similar.

richard67: > @richard67 What I miss is the option to add the language string to the com_content language file instead of the joomla.ini. In that case we could do `COM_CONTENT_ARTICLE_SAVE_CLOSE` or something similar.

@roland-d Would also be a way. No idea what's better. You can make a decision as RL.

Bakual: > @richard67 in theory the joomla.ini file should be the same across all three applications

That has never been the case. `lib_joomla.ini` is the one which should be identical.

roland-d:
@MagnusSinger You can add the new language string `JSAVEANDCLOSE` to the `language/en-GB/joomla.ini` file.

components/com_content/tmpl/form/edit.php
nibra: Semantically, 'apply' means save and stay open, 'save' does not inherently say anything about the state expected afterwards, 'save and close' means what it says. The constants should be used throughout the CMS to enforce consistency. So JAPPLY, JSAVE and JSAVE_AND_CLOSE would be the most forward and thus right choice in my opinion.

magnussinger:
@roland-d thanks for the info, I added it

language/en-GB/joomla.ini
```diff
@@ -111,6 +111,7 @@ JREQUIRED="Required"
 JRESET="Reset"
 JSAVE="Save"
 JSAVEASCOPY="Save As Copy"
+JSAVEANDCLOSE="Save & Close"
```
richard67:
```suggestion
JSAVEANDCLOSE="Save & Close"
JSAVEASCOPY="Save As Copy"
```
Alpha order.

components/com_content/src/Controller/ArticleController.php
```diff
@@ -395,6 +395,10 @@ public function save($key = null, $urlVar = 'a_id')
 		{
 			// Redirect to the article page, use the redirect url set from parent controller
 		}
+		elseif ($this->getTask() === 'apply')
+		{
+			// Do not redirect
+		}
```
roland-d: @MagnusSinger We can use an early return here by moving this block above the if statement on line 382. You can do this:
```
if ($this->getTask() === 'apply')
{
	return $result;
}
```

magnussinger: Good idea, implemented it

language/en-GB/joomla.ini
magnussinger: Okay I didn't know this order yet, I edited it

magnussinger:
@roland-d I just tested your idea and it didn't work, I would leave it like I did it

roland-d:
@MagnusSinger Why would it not work? I see no reason why it would not work. Looking further we could improve it better like this:
```
$menuitem = (int) $params->get('redirect_menuitem');

if (in_array($this->getTask(), ['save2copy', 'apply'], true))
{
	return $result;
}

// Check for redirection after submission when creating a new article only
if ($menuitem > 0 && $articleId == 0)
```
This works for me.

magnussinger:
Yes, you are right this works for me too. I edited it

components/com_content/src/Controller/ArticleController.php
```diff
@@ -374,6 +374,11 @@ public function save($key = null, $urlVar = 'a_id')
 		$params   = $app->getParams();
 		$menuitem = (int) $params->get('redirect_menuitem');
 
+		if (\in_array($this->getTask(), ['save2copy', 'apply'], true))
+		{
+			return $result;
+		}
```
Quy: This can be moved after line 371.

roland-d: Actually we can move this after line 369 because the result is already known then. We do not need the `$app` and `$articleId`.

magnussinger: I commited your proposal

richard67:
@ChristineWk Could you test again with the latest changes? Thanks in advance.

ChristineWk:
I have tested this item :white_check_mark: successfully on 2b900dc158c1472f1e137f2a17dd3a9e2a1ba88d<br /><br />Don't know if important as it's not part of this PR:
Also clicked: save as copy. Message: Save failed with the following error: Category is required. Would possibly fit there: #37354
But everything here refers to the frontend. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37362">issues.joomla.org/tracker/joomla-cms/37362</a>.</sub>

RickR2H:
I have tested this item :white_check_mark: successfully on 2b900dc158c1472f1e137f2a17dd3a9e2a1ba88d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37362">issues.joomla.org/tracker/joomla-cms/37362</a>.</sub>

RickR2H:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37362">issues.joomla.org/tracker/joomla-cms/37362</a>.</sub>

fancyFranci:
Thank you Magnus

richard67:
Yes, congratulations @MagnusSinger to your (as far as I know) first merged pull request for Joomla CMS.

magnussinger:
Thank you @fancyFranci and @richard67 and all the others for helping me with this. It was indeed my first pull request, and I'm looking forward to some more :)

PR #37374, opened 2022-03-25
[4.2] Handle extensions when the installer script doesn't return a boolean

laoneo:
Pull Request for comment https://github.com/joomla/joomla-cms/pull/37091#issuecomment-1079101363.

### Summary of Changes
Adds an extra check if the extension installer script returns a boolean in the routines.

### Testing Instructions
Install any extension.

Some which do not return a boolean do fail like acymailing.

All extension can be installed.

### DIFF
--- a/libraries/src/Installer/LegacyInstallerScript.php
+++ b/libraries/src/Installer/LegacyInstallerScript.php
@@ -42,12 +42,7 @@ class LegacyInstallerScript implements InstallerScriptInterface
 	 */
 	public function install(InstallerAdapter $adapter): bool
 	{
-		if (!method_exists($this->installerScript, 'install'))
-		{
-			return true;
-		}
-
-		return (bool) $this->installerScript->install($adapter);
+		return $this->callOnScript('install', [$adapter]);
 	}
 
 	/**
@@ -61,12 +56,7 @@ class LegacyInstallerScript implements InstallerScriptInterface
 	 */
 	public function update(InstallerAdapter $adapter): bool
 	{
-		if (!method_exists($this->installerScript, 'update'))
-		{
-			return true;
-		}
-
-		return (bool) $this->installerScript->update($adapter);
+		return $this->callOnScript('update', [$adapter]);
 	}
 
 	/**
@@ -80,12 +70,7 @@ class LegacyInstallerScript implements InstallerScriptInterface
 	 */
 	public function uninstall(InstallerAdapter $adapter): bool
 	{
-		if (!method_exists($this->installerScript, 'uninstall'))
-		{
-			return true;
-		}
-
-		return (bool) $this->installerScript->uninstall($adapter);
+		return $this->callOnScript('uninstall', [$adapter]);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


joeforjoomla:
I confirm that the PR works and fixes the issue.

laoneo:
Can you mark the test in the tracker? Thanks!

drmenzelit:
Tested with 3 different extensions, installation works now.
Can't use the tracker, no "Test this" button for this PR....

joeforjoomla:
Can't mark as tested on the tracker

libraries/src/Installer/LegacyInstallerScript.php
```diff
@@ -158,17 +133,47 @@ public function __get(string $name)
...
+	}
+
+	/**
+	 * Calls the function with the given name on the internal script with
+	 * some condition checking.
+	 *
+	 * @param   string $name       The name of the function
+	 * @param   array  $arguments  The arguments
+	 *
+	 * @return  bool
```
heelc29, line 158:
```suggestion
	 * @return  boolean
```

laoneo: Boolean is an alias for bool, so I would go with the native type. Good explanation can be found here https://stackoverflow.com/a/44009101/356375

LucasBenjamin112:
[Tips to Write Effective Persuasive Essays'.pdf](https://issues.joomla.org/uploads/1/8ccb1cc46fda184f7f412d67f6e22177.pdf)[Writing the informative essay.pdf](https://issues.joomla.org/uploads/1/5e1aa5cd3bc22c00c9241386ea3bd174.pdf)[List of Debate Topics on History'.pdf](https://issues.joomla.org/uploads/1/8faae23a8d961e3b7014c0736af9c9c2.pdf)[How to Make Your Persuasive Essay a Masterpiece'.pdf](https://issues.joomla.org/uploads/1/4d56c97de454f178d41f98188e15cef7.pdf)[50 best discussion points for undergrads'.pdf](https://issues.joomla.org/uploads/1/da85cde218abba93fe3eb5b0816b9259.pdf)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37374">issues.joomla.org/tracker/joomla-cms/37374</a>.</sub>

libraries/src/Installer/LegacyInstallerScript.php
heelc29: Yeah, your right.
In the rest of the file (and joomla) _boolean_ is used, so it would only be consistent with the rest.
https://github.com/joomla/joomla-cms/blob/b5adc29e6e45b94f3e59e9ba7323fd59593a63b8/libraries/src/Installer/InstallerAdapter.php#L1136

laoneo: We can do a followup pr, converting it to bool

richard67: @laoneo Why not here? It would not require new human tests as it’s just a doc block change.

richard67:
I think it’s a release blocker.

libraries/src/Installer/LegacyInstallerScript.php
richard67: Ah I misunderstood. You want to change all the others to bool.

richard67: @laoneo I suggest you change this one here to "boolean" like @heelc29 suggested so it's consistent with the rest of the file and nearly all of our other doc blocks.

richard67:
The patchtester seems to be one of these extensions, too.

joeforjoomla:
@richard67 almost all extensions out there have an installer function returning null

fancyFranci:
Thank you for fixing that.

PR #37375, opened 2022-03-25
[4.1] Order of elements in tag layout

Kostelano:
### Summary of Changes
Install Joomla 4.1 build.

Create menu items: **Tags - Compact List of Tagged Items** OR **Tags - Tagged Items**. On the "Options" tab for the "Order" parameter, the global variable will be "incorrectly passed".

The error is clearly visible if you use a language in the admin panel that is clearly different from English.

**Before / After**

Screenshot_1

### DIFF
--- a/components/com_tags/tmpl/tag/default.xml
+++ b/components/com_tags/tmpl/tag/default.xml
@@ -121,11 +121,11 @@
 				useglobal="true"
 				validate="options"
 				>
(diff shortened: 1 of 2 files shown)


brianteeman:
I dont think this is the correct fix and its related to #37386 

To confirm this if you go to the tags component and save the options and then go back to the menu item it will correctly display Use Global (Title)

PR #37383, opened 2022-03-26
[4.1] Adapt deleted files list in script.php to PR 37382

richard67:
Pull Request for Issue # .

### Summary of Changes

Add 3.10.7-2022-02-20.sql to deleted files in script.php where it is missing.

When PR #37382 will bee merged into the 3.10-dev branch and that change will be released, we have to delete the update SQL scripts with the right name, too.

So this PR here adds them to the list in script.php.

For mysql it does not need that change because there the file name was right from the beginning in the 3.10-dev branch.

### Testing Instructions

Code review.

Only the "3.10.7-2022-03-18.sql" file for mysql will be deleted after an update from next 3.10.7 to 4.1, the files for postgresql and sqlazure will remain.

Files "3.10.7-2022-03-18.sql" will be deleted after an update from next 3.10.7 to 4.1.

None.


richard67:
Adding the release blocker label because when #37382 is merged into 3.10-dev (and it should because it is also a release blocker), then it needs this PR here for 4.10 and 4.2, otherwise if the file is not deleted in might run under certain conditions by the updater, and that could cause a problem.

bembelimen:
Thx

Issue #37378, opened 2022-03-25
"Save As Copy" moves the original menu item to another menu


ConfidantCommunications:
### Steps to reproduce the issue
1. Create two menus, e.g. "menu 1" and "menu 2". Mine are for 2 languages. Not sure if language matters.
2. Create a menu item in menu 1. Click Save.
3. Edit the first menu item. Change the title and set the menu to the menu 2. I also changed the language. Select "Save as Copy".

### Expected result
The first menu item should remain in menu 1 and the copy should be found in menu 2.

### Actual result
Both the first and its copy end up in menu 2.

### System information (as much as possible)
Joomla 4.1.0

simbus82:
Confirmed! Same problem in 4.1.1-rc1
It's a bug.

brianteeman:
confirmed

anmode:
Yes, I am also able to reproduce it. can anyone guide me on what to do to solve this issue?

ReLater:
I think the error happens here: https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_menus/src/Model/ItemModel.php#L1429

`$pk` is detected wrongly:
`$pk = (!empty($data['id'])) ? $data['id'] : (int) $this->getState('item.id');`

Background:
- The controller sets `$data['id']` explicitly to `0` if we're in a `save2copy` task.
- Therefore `$data['id']` is detected as `empty` in the model and `$pk` falls back to `$this->getState('item.id')` which is the id of the menu item to be copied.

Maybe solution:
`$pk = (!empty($data['id']) || $data['id'] === 0) ? $data['id'] : (int) $this->getState('item.id');`

ReLater:
> Maybe solution:

Not tested enough, but it seems to fix the issue for me.

ReLater:
@anmode 
If you want: Do you like to provide a pr with my solution (or a better one if you think so)?

anmode:
@ReLater I'll test your solution. If it works fine than awesome. I would like to send pr with your solution and i I'll also try to add something.

anmode:
yeah! Fantastic this working correctly. I have tested it. could i generate a pr for it?

ReLater:
> could i generate a pr for it?

Everybody is allowed to provide a pr for Joomla without asking before. You, too ;-)
It will be reviewed and tested then by others. You can make changes if needed.

anmode:
[...]

PR #36377, opened 2021-12-23
[4.1] 'isset' can be replaced with coalesce

PhilETaylor:
Code review

PR #37395, opened 2022-03-28
[4.1] Save As Copy moves the original menu item to another menu fix

anmode:
Pull Request for Issue #37378

### Summary of Changes
$pk is detected wrongly:

1.The controller sets `$data['id']` explicitly to 0 if we're in a save2copy task.
2.Therefore `$data['id']` is detected as empty in the model and $pk falls back to `$this->getState('item.id')` which is the id of the menu item to be copied.

### Testing Instructions
1. Create two menus, e.g. "menu 1" and "menu 2".
2. Create a menu item in menu 1. Click Save.
3. Edit the first menu item. Change the title and set the menu to the menu 2.  Select "Save as Copy"

Both the first and its copy end up in menu 2.

The first menu item should remain in menu 1 and the copy should be found in menu 2.

No Document.


administrator/components/com_menus/src/Model/ItemModel.php
```diff
@@ -1426,7 +1426,7 @@ public function rebuild()
 	 */
 	public function save($data)
 	{
-		$pk         = (!empty($data['id'])) ? $data['id'] : (int) $this->getState('item.id');
+		$pk = (!empty($data['id']) || $data['id'] === 0) ? $data['id'] : (int) $this->getState('item.id');
```
richard67:
```suggestion
		$pk         = (!empty($data['id']) || $data['id'] === 0) ? $data['id'] : (int) $this->getState('item.id');
```
Keep alignment of "=" to the one in the line below like it was before.

richard67: Another problem which I see with the changed code is that before the change, the `empty($data['id'])` did also catch the case when the array element $data['id'] is not set. Now when this is the case, we will run into the 2nd part of the condition, the `$data['id'] === 0` check, and this for sure will raise a PHP warning when the array element is not set.

See the PHP documentation of `empty` here: https://www.php.net/manual/en/function.empty.php

> No warning is generated if the variable does not exist. That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

anmode: sure! I am Adding commit soon. I didn't notice it. thanks :)

anmode: > Another problem which I see with the changed code is that before the change, the `empty($data['id'])` did also catch the case when the array element $data['id'] is not set. Now when this is the case, we will run into the 2nd part of the condition, the `$data['id'] === 0` check, and this for sure will raise a PHP warning when the array element is not set.
> 
> See the PHP documentation of `empty` here: https://www.php.net/manual/en/function.empty.php
> 
> > No warning is generated if the variable does not exist. That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

ok sure! I'll look into it.

richard67: Regarding how to fix the thing with the `$data['id']` I could tell you how to do it, but maybe you want to find it out yourself, so I won't advise unless you ask me for it.

anmode: > Regarding how to fix the thing with the `$data['id']` I could tell you how to do it, but maybe you want to find it out yourself, so I won't advise unless you ask me for it.

If u know then please give some advise. So i can solve it soon. I want to solve it my own but i m busy in completing task assign regarding project 1 Joomla.

richard67: Well I first have to think about if your fix is really right for all scenarios which we have.

anmode: Sure!

chmst:
I have tested this item :white_check_mark: successfully on d27e6fea8f7a0c67d49caa43be8d8a33aa976305<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37395">issues.joomla.org/tracker/joomla-cms/37395</a>.</sub>

richard67:
@anmode I think your testing instructions are not complete. They cover only the test of the issue fixed by this PR. But the fix is made in the save method of the component's model, and this is used by any other kind of "Save", "Save & New", "Save & Close" or whatever may be available in frontend editing and in backend. It needs to test also if these are not broken with the change from this PR.

So I think you should add a final test step "Check that Save & Close, Save & New and Save as Copy in backend still work" or something like that.

administrator/components/com_menus/src/Model/ItemModel.php
richard67: I think you should do it like it is already done in the admin model: Just use `isset($data['id'])` instead of `(!empty($data['id']) || $data['id'] === 0)` like it is done here https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/MVC/Model/AdminModel.php#L1353 .

anmode:
> @anmode I think your testing instructions are not complete. They cover only the test of the issue fixed by this PR. But the fix is made in the save method of the component's model, and this is used by any other kind of "Save", "Save & New", "Save & Close" or whatever may be available in frontend editing and in backend. It needs to test also if these are not broken with the change from this PR.
> 
> So I think you should add a final test step "Check that Save & Close, Save & New and Save as Copy in backend still work" or something like that.

Okay I got the point what you are trying to convey I will look into Save & Close, Save & New and Save as Copy too so this fix valid in those.
If any advice related to code. Please tell.

richard67:
> If any advice related to code. Please tell.

@anmode See my answer in the above review comment https://github.com/joomla/joomla-cms/pull/37395#discussion_r837200120 :

> I think you should do it like it is already done in the admin model: Just use `isset($data['id'])` instead of `(!empty($data['id']) || $data['id'] === 0)` like it is done here https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/MVC/Model/AdminModel.php#L1353 .

As said, with your current code it will fail when `$data['id']` is not set.

anmode:
> > If any advice related to code. Please tell.
> 
> @anmode See my answer in the above review comment [#37395 (comment)](https://github.com/joomla/joomla-cms/pull/37395#discussion_r837200120) :
> 
> > I think you should do it like it is already done in the admin model: Just use `isset($data['id'])` instead of `(!empty($data['id']) || $data['id'] === 0)` like it is done here https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/MVC/Model/AdminModel.php#L1353 .
> 
> As said, with your current code it will fail when `$data['id']` is not set.

Okay! I'll look into it.

richard67:
> Okay! I'll look into it.

What does it need to look into? I've provided all necessary information already.

anmode:
> > Okay! I'll look into it.
> 
> What does it need to look into? I've provided all necessary information already.
😅just doing changes and pushing commit. Thanks :).

vorayash:
I have tested this change and working correctly on machine.

richard67:
@vorayash Could you mark your test result in the issue tracker by going to this page https://issues.joomla.org/tracker/joomla-cms/37395 , using the blue "Test this" button at the top left corner, selecting the test result ("Tested successfully" in this case) and then submit? Thanks in advance.

vorayash:
Ok sir.

vorayash:
I have tested this item :white_check_mark: successfully on e7c60bb44a1ff33dae1d56700de234c5cfea899d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37395">issues.joomla.org/tracker/joomla-cms/37395</a>.</sub>

vorayash:
I have tested this item ✅  successfully on f7a966e<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37395">issues.joomla.org/tracker/joomla-cms/37395</a>.</sub>

administrator/components/com_menus/src/Model/ItemModel.php
```diff
@@ -1426,7 +1426,7 @@ public function rebuild()
 	 */
 	public function save($data)
 	{
-		$pk         = (!empty($data['id'])) ? $data['id'] : (int) $this->getState('item.id');
+		$pk         = isset($data['id']) ? $data['id'] : (int) $this->getState('item.id');
```
Quy, line 1429:
```suggestion
		$pk         = $data['id']) ?? (int) $this->getState('item.id');
```
See #36377

Quy:
I have tested this item :white_check_mark: successfully on e7c60bb44a1ff33dae1d56700de234c5cfea899d<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37395">issues.joomla.org/tracker/joomla-cms/37395</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37395">issues.joomla.org/tracker/joomla-cms/37395</a>.</sub>

Quy:
Thanks

Issue #37012, opened 2022-02-12
[4.x][com_ajax] Ajax call error from the module


arenamax:
Ajax is not available by namespace!

###### Namespace from install xml
```...
   <namespace path="src">Am\Module\Amap</namespace>
   ...
```

###### js
```
Joomla.request({
	url: "index.php?option=com_ajax&module=amap&method=getReloadLayoutParams&format=json",
	method: "POST",
	headers: {
		"Cache-Control" : "no-cache",
		"Content-Type" : "application/json"
	},
	onBefore: function (xhr){
		console.log("onBefore");
	},

	onSuccess: function (response, xhr){
		if (response !== ""){
			let res = JSON.parse(response);
			console.log(res.message);
		}
		console.log("onSuccess");
	},
					
        onError: function(xhr){
		console.log("onError");
	},

	onComplete: function (xhr){
		console.log("onComplete");
	}
});
```

###### AmapHelper.php
path 'modules/mod_amap/src/Helper/AmapHelper.php'
```
<?php
namespace Am\Module\AMAP\Site\Helper;

\defined( '_JEXEC' ) or die;

class AmapHelper
{
        public static function getReloadLayoutParamsAjax()
	{
		$html = array();
		$html['success'] = true;
		$html['message'] = 'Ok';
		$html['messages'] = '';
		$html['data'] = array();
		
		echo json_encode($html);
                exit;
	}
}
```
### Expected result
JSON data

###### Console log
```
onBefore
File mod_amap/helper.php does not exist. // ??? It doesn't exist anyway
onSuccess
onComplete
```

### System information (as much as possible)
Joomla v. 4.1.0-rc4
PHP 7.4.27

wedal:
I confirm the problem.

In /components/com_ajax/ajax.php

`$moduleInstance = $app->bootModule('mod_' . $module, $app->getName());`

no helper class returned for a helper file in a subdirectory /src/Helper.

As a solution can use:
https://github.com/dgrammatiko/invalidate-cache/blob/d19f7f1b6b375dfcc4417d84c8ac3697097a84fd/src/invalidatecache/services/provider.php 

But that's not good, I think.

dgrammatiko:
> But that's not good, I think.

This is the Joomla 4 way which is not 100% B/C because of namespacing...

wedal:
Creating a separate file in a module just to be able to use com_ajax is definitely a problem.

[...]

PR #32633, opened 2021-03-10
[4.0] Helper factory

laoneo:
Pull Request for pr #32291.

### Summary of Changes
All over joomla are helpers manually loaded. Better to load them through a factory from the booted extension.

### Testing Instructions
- Open the dashboard.
- Add the following code to the file administrator/modules/mod_quickicon/src/Helper/QuickIconHelper.php:  
```php
public static function getDataAjax()
{
	$data = ['test'];

	return $data;
}
```
- Open the url /administrator/index.php?option=com_ajax&module=quickIcon&method=getData&format=json

### Actual result BEFORE applying this Pull Request
- Icons are loaded.
- Ajax url shows `{"success":true,"message":null,"messages":null,"data":["test"]}`

### Expected result AFTER applying this Pull Request
- Icons are loaded.
- Ajax url shows `{"success":false,"message":"The file at mod_quickIcon\/helper.php does not exist.","messages":null,"data":null}`

### Documentation Changes Required
Should be documented on the same place where the new extension setup is documented.

PR #37398, opened 2022-03-29
Com ajax fix

Ruud68:
Pull Request for Issue #37012 .

### Summary of Changes
com_ajax tries to only load the module helper from the J3 module helper location

### Testing Instructions
in browser goto: [yourdomain]/index.php?option=com_ajax&module=articles_news&method=test&format=raw

RuntimeException: The file at mod_articles_news/helper.php does not exist.

LogicException: Method testAjax does not exist.

### DIFF
--- a/components/com_ajax/ajax.php
+++ b/components/com_ajax/ajax.php
@@ -50,7 +50,7 @@ if (!$format)
 /*
  * Module support.
  *
- * modFooHelper::getAjax() is called where 'foo' is the value
+ * FooHelper::getAjax() is called where 'foo' is the value
  * of the 'module' variable passed via the URL
  * (i.e. index.php?option=com_ajax&module=foo).
  *
@@ -63,8 +63,6 @@ elseif ($input->get('module'))
 
 	if ($moduleId && $table->load($moduleId) && $table->enabled)
 	{
-		$helperFile = JPATH_BASE . '/modules/mod_' . $module . '/helper.php';
-
 		if (strpos($module, '_'))
 		{
 			$parts = explode('_', $module);
@@ -90,6 +88,13 @@ elseif ($input->get('module'))
 			$class = 'Mod' . ucfirst($module) . 'Helper';
 		}
 
+		$helperFile = JPATH_BASE . '/modules/mod_' . $module . '/src/Helper/' . substr($class, 3) . '.php';
+
+		if (!is_file($helperFile))
+		{
+			$helperFile = JPATH_BASE . '/modules/mod_' . $module . '/helper.php';
+		}
+
 		$method = $input->get('method') ?: 'get';
 
 		$moduleInstance = $app->bootModule('mod_' . $module, $app->getName());
@@ -109,7 +114,7 @@ elseif ($input->get('module'))
 				$basePath = JPATH_BASE;
 				$lang     = Factory::getLanguage();
 				$lang->load('mod_' . $module, $basePath)
-				||  $lang->load('mod_' . $module, $basePath . '/modules/mod_' . $module);
+					||  $lang->load('mod_' . $module, $basePath . '/modules/mod_' . $module);
 
 				try
 				{
@@ -217,7 +222,7 @@ elseif ($input->get('template'))
 				// Load language file for template
 				$lang = Factory::getLanguage();
 				$lang->load('tpl_' . $template, $basePath)
-				||  $lang->load('tpl_' . $template, $basePath . '/templates/' . $template);
+					||  $lang->load('tpl_' . $template, $basePath . '/templates/' . $template);
 
 				try
 				{
@@ -250,13 +255,13 @@ elseif ($input->get('template'))
 // Return the results in the desired format
 switch ($format)
 {
-	// JSONinzed
+		// JSONinzed
 	case 'json':
 		echo new JsonResponse($results, null, false, $input->get('ignoreMessages', true, 'bool'));
 
 		break;
 
-	// Handle as raw format
+		// Handle as raw format
 	default:
 		// Output exception
 		if ($results instanceof Exception)
--- a/components/com_ajax/ajax.php
+++ b/components/com_ajax/ajax.php
@@ -260,13 +260,13 @@ elseif ($input->get('template'))
 // Return the results in the desired format
 switch ($format)
 {
-		// JSONinzed
+	// JSONinzed
 	case 'json':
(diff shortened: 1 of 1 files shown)


growmoreagency71:
[GrowMore](http://growmore.pk/ "Growmore") Institute and Digital Marketing Agency Focusing To Train The Youth With Million Dollar Income Generating Skills. Which Will Help To Lift Pakistan. We Also Provide [Digital Marketing](http://growmore.pk/ "") Services To Local And International Businesses To Grow Their Businesses, Which Automatically Plays A Part In Growth Of Pakistan Economy.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37398">issues.joomla.org/tracker/joomla-cms/37398</a>.</sub>

Ruud68:
@arenamax @wedal are you able to test this PR?

laoneo:
When you are on Joomla 4, then better to let the module implement `Joomla\CMS\Helper\HelperFactoryInterface`. We want to avoid manual class name compiling. These ways are basically deprecated.

Ruud68:
> When you are on Joomla 4, then better to let the module implement `Joomla\CMS\Helper\HelperFactoryInterface`. We want to avoid manual class name compiling. These ways are basically deprecated.

can you point me to this in mod_articles_news as i use that as reference

laoneo:
Does the pr #32633 help?

Ruud68:
@laoneo no it doesn't: for e.g. mod_articles_news in what file do I add the implements Joomla\CMS\Helper\HelperFactoryInterface

laoneo:
You need to do the same as the quick icon module, convert it first to a service provider and then do the helper stuff.

Ruud68:
okay... so that would basically mean that I have to redo all modules that use ajax. #snif (and probably someone has to refactor all Joomla core modules?)

Although 'deprecated' this PR improves at least the deprecated+++ thing where it is trying to load the helper file from the J3 location

Ruud68:
and as a bonus: the module helper file was / is always an abstract class. so not only do you need to refactor the module into a service provider but also refactor the module helper file into not being abstract (as an abstract class cannot be instantiated). and by doing that potentially refactor / rethink everything that uses the helper class.... 
So now I start to understand why the module/helper.php 'workaround' is in there: it is way easier to not use the src/Helper/helper.php file but keep it in the J3 module helper location...

Ruud68:
okay, so thanks @laoneo : refactoring the module into a service provider works.

but as this is not an option for some devs (probably adding to the 'lets wait until J5 is out before refactoring again') this PR would definitely add value for those developers who do not want their J4 module be refactored into a J3 file structure.

So if we get two testers (@arenamax @wedal looking at you :)), then one of the maintainers responsible for this code can decide to either accept or not accept. At least then we have some clarity on the effort to report the issue and the effort to propose a PR for it.

wilsonge:
Ok so what's the gap we're trying to fill here? I'm missing something for sure I think.

If your J3 compatible your going to use the J3 file structure

If your J4 only your going to use namespaces and services.

I'm not seeing what scenario your solving by namespacing but not using services?

bembelimen:
> but as this is not an option for some devs (probably adding to the 'lets wait until J5 is out before refactoring again') this PR would definitely add value for those developers who do not want their J4 module be refactored into a J3 file structure.

That's one of the worst choices a developer can have, nobody should wait for J5, as now the gap is the smallest possible (and tbh if you programmed in a clean way in the past it's not that much of an effort to get a fully compatible J4 component...beside some renaming you mostly are deleting stuff, if you didn't programmed in a clean way, you shouldn't wait either but start now cleanup and refactor).

> 
> then one of the maintainers responsible for this code can decide to either accept or not accept.

I'm not in charge of 4.2 but for me this is a big "No" in terms of merging.

chmst:
After reading the whole discussion and comments of maintainers I close this PR. Thanks all for your contributions and explanations!

Ruud68:
Hi, thanks for the feedback and decision. please do not forget to re-open the issue again that was closed by this pr.
and as an advise: refactor the core joomla modules to the new namespacing / services as well as now only the administrator module quick-icons is correct. That is what started this issue in the beginning for me (using a core joomla module as boilerplate) and probably as well as for the issue starter.

laoneo:
@Ruud68 you are welcome to help to convert the core modules to services.

Ruud68:
Thanks @laoneo for following up! Maybe the maintainers could make a 'todo' of these kind of changes. Then devs could do PR's for them in a controlled way: the dev then has the 'guarantee' that time invested is appreciated / used and not wasted / ignored.

laoneo:
Your pr started some important discussion in the Bug Squad chat. I will do tomorrow a pr where I convert a module for 4.2. It should act then as example for others. I would really appreciate when the outcome of this pr is a start of a process where we convert existing modules together.

Ruud68:
and thanks again @laoneo this is so important.... #communication :)

laoneo:
Can you have a look on #37445. That's how it should be. Ignore the part with the helper factory aware interfaces, jut focus on the module part.

Ruud68:
Good morning @laoneo , will have a look over the weekend. First need to get a good understanding of how this works as there are more changes (like the dispatcher and dropping the mod_modulename.php then I currently have working on my new.module.  I'm am assuming there is no documentation so need to dive into the code for this. Will follow up as soon as possible.
As for the actual work involved: maybe create an issue per module that needs refactoring so we / the devs can 'claim' that. Or use GitHub projects for that? Not sure what is handy ro have an overview and avail double work

laoneo:
That's a good idea, as cards do not allow to assign them, I'v opened a simple issue #37455 where devs can post a comment with the module they are converting.

growmoreagency71:
GrowMore Institute and Digital Marketing Agency Focusing To Train The Youth With Million Dollar Income Generating Skills. Which Will Help To Lift Pakistan. We Also Provide [Digital Marketing](https://growmore.pk/ "") Services To Local And International Businesses To Grow Their Businesses, Which Automatically Plays A Part In Growth Of Pakistan Economy.
<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37398">issues.joomla.org/tracker/joomla-cms/37398</a>.</sub>

PR #37404, opened 2022-03-29
[4.2] Upmerge 2022-03-29

fancyFranci:
Pull Request for Issue # .


brianteeman:
surely the change in `libraries\src\Version.php` is not correct. This is 4.2 so should be 
```

	const MAJOR_VERSION = 4;

	const MINOR_VERSION = 2;

	const PATCH_VERSION = 0;
```

Kostelano:
Please add a language tag

PR #37412, opened 2022-03-29
[4.1] Refactoring Appveyor test setup

Hackwar:
The Appveyor setup was pretty outdated and did rarely any tests. This PR refactors all of this and hopefully improves this for the future. The following changes were done:
- Updated the system image to run off of from Visual Studio 2015 to Visual Studio 2022
- Removed the limitations for branches. We haven't had l10n branches for years...
- Added PHP 8.0 and 8.1 (where 8.1 is allowed to still fail right now)
- Removed the "caching" of the PHP installation. That stuff never worked. We are always starting with a clean image.
- Using built in chocolatey instead of downgrading to an earlier version. The bug that required that was fixed a year ago.
- Removed WinCache setup. First of all, we never ran any tests with WinCache and second of all, WinCache seems to be dead for several years already. I'm going to create PRs to deprecate it for Joomla 5 soon.
- Removed PostgreSQL. Up till now we didn't run any database tests and Postgres didn't work out of the box in my tests. So for now I've removed that.
- Added the Integration tests.

These changes mean that while we run 5 jobs instead of 3, it doesn't take longer than before, maybe it is even quicker. This can be merged when the Appveyor tests pass, no tests by humans needed.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,89 +1,63 @@
+image: Visual Studio 2022
 build: false
 platform:
   - x64
 clone_folder: C:\projects\joomla-cms
 
-branches:
-  except:
-    - /l10n_*/
-
 ## Build matrix for lowest and highest possible targets
 environment:
   PHPBuild: "x64"
-  VC: "vc15"
-  WINCACHE: "2.0.0.8"
   matrix:
   - php_ver_target: 7.2
   - php_ver_target: 7.3
   - php_ver_target: 7.4
+  - php_ver_target: 8.0
+  - php_ver_target: 8.1
+
+matrix:
+  allow_failures:
+    - php_ver_target: 8.1
 
 init:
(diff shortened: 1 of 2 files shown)


HLeithner:
Any Chance to check hash check for composer? For drone we do this at image build time, but for appveyor you maybe have to do it at runtime

Hackwar:
Uh, I'm pretty sure that we don't actually check the hash of composer... I commented that line... 😉 However, against what do you want check that hash? Do you want us to update our docker images, etc. each time a new composer is released?

HLeithner:
> Uh, I'm pretty sure that we don't actually check the hash of composer... I commented that line... 😉 However, against what do you want check that hash? Do you want us to update our docker images, etc. each time a new composer is released?

That's what's happening now, weekly builds including hash check of all (current) docker images

Hackwar:
Yeah, don't force me to do this again... Setting this up and making this work was not fun.

HLeithner:
thanks

PR #37437, opened 2022-03-31
Fix bug use value 1 as TRUE Workflowcondition Field

korenevskiy:
Correcting a gross mistake. Obviously, the value 1 of the field attribute is always used equivalent to `"TRUE"`. But this value is not supported by this attribute. We very often use the values `"1"` or `"0"` to determine the bool value, and to a small extent use other values. Therefore, to support the value of `"1"` to determine true, I add this PR.
 
This field has an important XML attribute `"hide_all"`. The value of this attribute defining the true value is: `"true", "yes"`. But such in these values there is no value of `"1"`. Since these values are the most common value for determining the true value.
```XML
<field type="workflowcondition" name="fieldname" hide_all="yes" />
```
I am adding a PR capability to support the true value for the `"hide_all"` attribute value of `"1"`.

This example will work with PR, and not work without PR.
```XML
<field type="workflowcondition" name="fieldname" hide_all="1" />
```


chmst:
@korenevskiy could youplease  complete your PR with the usual description and  Testing instruction?

korenevskiy:
@chmst  I described the problem in detail and added this description to the top description.

laoneo:
But we need a test instruction, so other people can test your pr if it fixes the issue. See #37521 for an example.

korenevskiy:
> But we need a test instruction, so other people can test your pr if it fixes the issue. See #37521 for an example.

This example will work with PR, and not work without PR.
```XML
<field type="workflowcondition" name="fieldname" hide_all="1" />
```

HLeithner:

Issue #37432, opened 2022-03-31
Error on logout on Joomla 3.10.7 with PostgreSQL 13.3


motokraft:
Выводит ошибку при выходе из аккаунта в Joomla 3.10.7 и PostgreSQL 13.3

(translated with Google Translator: Displays an error when logging out of an account in Joomla 3.10.7 and PostgreSQL 13.3)

`bin2hex(): Argument #1 ($string) must be of type string, resource given
JPATH_ROOT\libraries\joomla\database\driver\pgsql.php:1021`

motokraft:
<html>
<body>


Call stack
--
# | Function | Location
1 | () | JROOT\libraries\joomla\database\driver\pgsql.php:1021
2 | bin2hex() | JROOT\libraries\joomla\database\driver\pgsql.php:1021
3 | JDatabaseDriverPgsql->quoteBinary() | JROOT\libraries\src\User\UserHelper.php:919
4 | Joomla\CMS\User\UserHelper::destroyUserSessions() | JROOT\plugins\user\joomla\joomla.php:359
5 | PlgUserJoomla->onUserLogout() | JROOT\libraries\joomla\event\event.php:70
6 | JEvent->update() | JROOT\libraries\joomla\event\dispatcher.php:160
7 | JEventDispatcher->trigger() | JROOT\libraries\src\Application\BaseApplication.php:108
8 | Joomla\CMS\Application\BaseApplication->triggerEvent() | JROOT\libraries\src\Application\CMSApplication.php:985
9 | Joomla\CMS\Application\CMSApplication->logout() | JROOT\administrator\components\com_login\controller.php:112
10 | LoginController->logout() | JROOT\libraries\src\MVC\Controller\BaseController.php:702
11 | Joomla\CMS\MVC\Controller\BaseController->execute() | JROOT\administrator\components\com_login\login.php:22
12 | require_once() | JROOT\libraries\src\Component\ComponentHelper.php:402
13 | Joomla\CMS\Component\ComponentHelper::executeComponent() | JROOT\libraries\src\Component\ComponentHelper.php:377
14 | Joomla\CMS\Component\ComponentHelper::renderComponent() | JROOT\libraries\src\Application\AdministratorApplication.php:101
15 | Joomla\CMS\Application\AdministratorApplication->dispatch() | JROOT\libraries\src\Application\AdministratorApplication.php:159
[...]

PR #37440, opened 2022-03-31
[3] postgres logout

alikon:
Pull Request for Issue #37432 .

### Testing Instructions
see #37432 

error

logged out without error

### DIFF
--- a/libraries/joomla/database/driver/pgsql.php
+++ b/libraries/joomla/database/driver/pgsql.php
@@ -1018,12 +1018,9 @@ class JDatabaseDriverPgsql extends JDatabaseDriverPdo
 	 */
 	public function quoteBinary($data)
 	{
-		if (is_resource($data))
+		if (is_resource($data) && get_resource_type($data) === 'stream')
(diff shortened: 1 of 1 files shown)


libraries/joomla/database/driver/pgsql.php
```diff
@@ -1018,6 +1018,14 @@ public function updateObject($table, &$object, $key, $nulls = false)
 	 */
 	public function quoteBinary($data)
 	{
+		if (is_resource($data))
+		{
+			if (get_resource_type($data) === 'stream')
+			{
+				$data = stream_get_contents($data, -1, 0);
+			}
+		}
```
richard67:
```suggestion
		if (is_resource($data) && get_resource_type($data) === 'stream')
		{
			$data = stream_get_contents($data, -1, 0);
		}
```
Should work, too. The get_resource_type will only be called if $data is a resource.

richard67:
@alikon Does that need to be fixed in the framework, too?

PR #33182, opened 2021-04-18
[4.0] J4 Modules should require an element field

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes

Since https://github.com/joomla/joomla-cms/pull/32633 J4 (non legacy) Modules don't have a `module.php` file in their root folder
https://github.com/joomla/joomla-cms/blob/2c302641aaad823ac91421a0b11f14e3f8e2fd27/administrator/modules/mod_quickicon/mod_quickicon.xml#L14

and thus it's impossible to install a module that uses namespace, check the existing code: https://github.com/joomla/joomla-cms/blob/2c302641aaad823ac91421a0b11f14e3f8e2fd27/libraries/src/Installer/Adapter/ModuleAdapter.php#L337-L356.

To mitigate this in the easiest possible way I propose to require (for the J4 native modules, not the legacy) one field in the XML: eg `<element>mod_wow</element>`;

### Testing Instructions

You could test this in 2 different ways:
- go to your db and delete in the extension table the `mod_quickicon`
- Then move the `administrator/mopdules/mod_quickicon` to a folder outside of the Joomla folder
- Zip the content of the folder and try to install it (it will fail with a message `file mod_quickicon.php doesn't exist`)
- Edit the XML file removing the line `<filename module="mod_quickicon">mod_quickicon.php</filename>` and adding after line 3 `<element>mod_quickicon</element>` then zip the contents of the folder and try again

Alternative you can download `https://jinvalidate.netlify.app/dist/j4/mod_invalidatecache_0.0.6.zip` and observe that the module installs correctly

### Documentation Changes Required

@laoneo some feedback here?
@wilsonge ?

PR #37445, opened 2022-04-01
[4.2] Introduce helper factory aware interface and convert news module to services

laoneo:
Pull Request for pr #37398.

### Summary of Changes
Introduces a helper factory aware interface and converts the articles news site module to services. For backwards compatibility a new helper class was introduced and the old abstract one got deprecated.

### Testing Instructions
- Create an article
- Create an Articles - Newsflash module

It works.

It works.


dgrammatiko:
@laoneo please add a line in the xml like `<element>mod_article_news</element>` so the module could be installable. Also remove the old file mod_article_news.php as it is moved inside the src folder, etc

laoneo:
> @laoneo please add a line in the xml like `<element>mod_article_news</element>` so the module could be installable.

No module has such an element tag. Why should I add it?

dgrammatiko:
Because https://github.com/joomla/joomla-cms/pull/33182

laoneo:
This should be enough https://github.com/joomla/joomla-cms/pull/37445/commits/13c1cf6b393b479fd1d47450cc6839fd95a45b1c. #33182 is basically obsolete as it would already work without it, see here https://github.com/joomla/joomla-cms/pull/36042#issuecomment-968362963. Just one entry needs a module attribute, doesn't mater which one.

dgrammatiko:
That's an undocumented hack and probably shouldn't be promoted! The docs are clear:
Screenshot 2022-04-01 at 15 33 18

https://docs.joomla.org/Manifest_files

laoneo:
Where do you see that in action the element tag? None of my extensions does have that element and the code which uses any of the file elements is there since Joomla 1.7, which do extensions use since ages. Your pr was for 4, so I guess you added a new behavior. I wouldn't call the original way a hack. Anyway, lets open an issue where we can discuss this in more deep if you want to get that sorted.

dgrammatiko:
I think the project should have one unique way for defining the element especially for the namespaced extensions. When I did that PR (that was before your discovery?) I thought that normalising the ~optional~ `element` field and move it to the required ones would be self explanatory and also if that was to be standardised then the name field could be left free to be any type of string (translatable or not). This was somehow bite me with 4.1 as the templates name field is not explicitly defined correctly, even for the core ones, (should be `tmpl_cassiopeia` and not `cassiopeia`). Anyways either keep adding attributes to a folder or explicitly asking for an element field should be standardised and needs to be noted in the docs. People shouldn't reverse engineer code to figure out how things are working. End of off topic comments

laoneo:
Show me one extension which exists before Joomla 4 was released that uses the element tag.

dgrammatiko:
I can't, apart from the one I edited in that PR but maybe this Unit can: https://github.com/joomla/joomla-cms/blob/7c12dc7b007e0cdc8e28e88f214f355fb8dae054/tests/Unit/Libraries/Cms/Installer/Adapter/ModuleAdapterTest.php#L109-L123

laoneo:
Ok, we leave the pr as it is for now.

modules/mod_articles_news/src/Helper/ArticlesNewsHelper.php
```diff
@@ -35,167 +30,11 @@ abstract class ArticlesNewsHelper
...
-			else
-			{
-				$item->afterDisplayTitle    = '';
-				$item->beforeDisplayContent = '';
-				$item->afterDisplayContent  = '';
-			}
-		}
-
-		return $items;
+		return (new ArticlesNews)->getList($params, Factory::getApplication());
```
HLeithner: We should trigger a deprecation here.

laoneo: As far as I know we only trigger a deprecate message when something in the code gets deprecated, but when a whole function or class gets deprecated, then only a message in the doc block is added.

Ruud68:
in your commit you have created a new ArticleNews.php in the src/Helper directory.
Will that replace the ArticleNewsHelper.php (as that file now only has the deprecated message)?

Asking because if that is the case, com_ajax is broken as it hardcoded looks for a module helperfile ending in 'Helper': so ArticleNewsHelper in this example.

laoneo:
@Ruud68 you are right. Reverted to the original helper file and made it none abstract and added a new getArticles function which is none static. What do you think?

On the other side also like this it still doesn't work with com_ajax, because the function needs to end with "Ajax" https://github.com/joomla/joomla-cms/blob/4.1-dev/components/com_ajax/ajax.php#L99.

Ruud68:
> @Ruud68 you are right. Reverted to the original helper file and made it none abstract and added a new getArticles function which is none static. What do you think?
> 
> On the other side also like this it still doesn't work with com_ajax, because the function needs to end with "Ajax" https://github.com/joomla/joomla-cms/blob/4.1-dev/components/com_ajax/ajax.php#L99.

Cool, I think this will work. As for the Ajax in the function name that is not the issue. The issue was that the helperfile wasn't loaded. When that is loaded the method is called (ending in 'Ajax') so in the helper file you need to have that function otherwise you will get an error.

This is only relevant if you have a module that will service com_ajax requests because they have functionality that relies on that (which the Joomla core modules do not do)

What still 'bugs' me is that a module following this new approach will not work on J4.1 / 4.0.
For me as extension dev that complicates things as with every dependency on a Joomla version the 'market share' will drop :S
just saying :) should not hold back improvements.

laoneo:
If you leave the old module file from J3 and load the helper the same way in the dispatcher as the quickicon module, then you have compatibility down to J3.

laoneo:
@Ruud68 can you give this one here a test?

Ruud68:
@laoneo so just set up a new nightly 4.2 environment (april 5th), pulled this PR and copied the files over the new test environment.
when going to [domain]/index.php?option=com_ajax&module=articles_news&method=test&format=raw

I get the message: RuntimeException: The file at mod_articles_news/helper.php does not exist.

have not tried to troubleshoot this

laoneo:
Did you add the function testAjax to the helper?

Ruud68:
Hi, that is not necessary. com_ajax first tries to load the helper and when it does then it tries to load the method.
The error I got is for loading the helper > so it didn't get to the loading of the method (testAjax)

laoneo:
It does, but when the helper has no such function it tries the old way and when that doesn't succeed, it throws the old error message. You can see it [here](https://github.com/joomla/joomla-cms/blob/4.1-dev/components/com_ajax/ajax.php#L99), it sets the result to `null` and doesn't throw an error message. If you want to add there an error, then a new pr is required.
Important for this pr is that you test the module on the ordinary way.

Ruud68:
sorry about that, you are completely right.
So the module works, the helper works and when adding testAjax com_ajax also works for modules in this way!

Ruud68:
I have tested this item :white_check_mark: successfully on b081abe4e5d28cc6844051d21988713ba9b4e454<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37445">issues.joomla.org/tracker/joomla-cms/37445</a>.</sub>

SumCompanyInc:
I have tested this item :white_check_mark: successfully on 3ee5c3257342675a66c72442857677cb69cd514f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37445">issues.joomla.org/tracker/joomla-cms/37445</a>.</sub>

Ruud68:
I have tested this item :white_check_mark: successfully on 3ee5c3257342675a66c72442857677cb69cd514f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37445">issues.joomla.org/tracker/joomla-cms/37445</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37445">issues.joomla.org/tracker/joomla-cms/37445</a>.</sub>

libraries/src/Helper/HelperFactory.php
```diff
@@ -8,6 +8,9 @@
 
 namespace Joomla\CMS\Helper;
 
+use Joomla\Database\DatabaseAwareInterface;
+use Joomla\Database\DatabaseAwareTrait;
+
 \defined('_JEXEC') or die;
```
Quy:
```suggestion
\defined('_JEXEC') or die;

use Joomla\Database\DatabaseAwareInterface;
use Joomla\Database\DatabaseAwareTrait;
```

roland-d:
Thanks everybody

Issue #36346, opened 2021-12-18
[4.0] saveOrderAjax does not send form data in Joomla 4


PhocaCz:
### Steps to reproduce the issue
Order some items in administration - in article view - in Joomla 4

### Expected result
saveOrderAjax sends all form data like it does in Joomla 3

![img](https://i.imgur.com/51p2C7X.png)

### Actual result
saveOrderAjax does not send all form data like it does in Joomla 3

![img](https://i.imgur.com/EIHm9BV.png)

### System information (as much as possible)
Joomla 4.0.5

brianteeman:
and the problem is?

PhocaCz:
The problem is that you need this data. E.g. when items can have multiple categories, you need to get the information about the current active category which you get in filter[category] array variable.

No idea if this is a problem for article view, but it is problem for another extensions. I don't know if it's intentional to make the API hostile to extensions (and it's not just this issue), or if it's just plain ignorance.

brianteeman:
I couldnt guess that from your screenshot

PhocaCz:
The difference between Joomla 3 and Joomla 4 is that Joomla 3 sends all form data but Joomla 4 only order and cid variable (or another data in table row - I didn't test).

You don't need to see the screenshots, you can test it for yourself. Just order items in Joomla 3 and Joomla 4 and enable e.g. Google Chrome DevTools - Network tab (Fetch/XHR subtab) and see which Form Data items are sent when ordering items.

- In Joomla 3 - all form data
- In Joomla 4 - cid, order data (or another in table row).

That is all. All form data items are used by extensions because they have extended features in comparison to core - like e.g. multiple categories (items can be assigned to more than one category), etc.

brianteeman:
> You don't need to see the screenshots, you can test it for yourself. 

Maybe but from your original post it was meaningless. It's called saveorderajax so why would it do anything other than save the order

PhocaCz:
To save the order, you need data. I don't ask for any other function. The saveorderajax function just does not send the needed data in Joomla 4. (In Joomla 3 everything is all right).

Sorry but I don't understand your answers. This issue is clear:

Joomla 4 - not all form data items are sent
Joomla 3 - all form data items are sent.

The problem here is not article view which is an example to test it, the problem is the saveorderajax feature.

brianteeman:
Why would you need any more data than the order and the id

Fedik:
Not really a bug, but changed behavior.

[...]

PR #26184, opened 2019-09-06
[4.0] Implement time field

VladikB:
Pull Request for Issue #22140  .

### Summary of Changes

I created new files to rectify the missing of input type="time" with a typical time field. Since IE and Safari don't support this input type, on these browsers it goes to input type="text"

### Testing Instructions

a) 1. Go to  administrator\components\com_users\forms. Edit the first field of user.xml file to 

```
name="name"
type="time"
label="COM_USERS_USER_FIELD_NAME_LABEL"
required="true"
validate="time"
size="30"
```
2. Go to Joomla-Backend -> Users ->Manage -> Click on Your User -> try to change it

b) 1. Go to  administrator\components\com_users\forms. Edit the first field of user.xml file to 

```
name="name"
type="time"
label="COM_USERS_USER_FIELD_NAME_LABEL"
required="true"
validate="time"
size="30"
min="08:00"
max="15:00"
```
2. Go to Joomla-Backend -> Users ->Manage -> Click on Your User-> try to change it

c) 1. Go to  administrator\components\com_users\forms. Edit the first field of user.xml file to 

```
name="name"
type="time"
label="COM_USERS_USER_FIELD_NAME_LABEL"
required="true"
validate="time"
size="30"
step="3"
```
2. Go to Joomla-Backend -> Users ->Manage -> Click on Your User-> try to change it

### Expected result

a) The Name field is a time field with hh:mm
b) The Name field is a time field with hh:mm that goes from 08:00 to 15:00
c) The Name field is a time field with hh:mm:ss and seconds can be changed in step of three seconds

### Documentation Changes Required

If this works, I am going to write a documentation since it is a new field type.

PR #37456, opened 2022-04-02
[6.0] New Date and Datetime fields

Fedik:
### Summary of Changes

Add 2 new input `date` (DateField class), and `datetime-local` (DatetimeField class). Fix Time field to support filter.

Links to MDN docs:
`datetime-local`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local
`date`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

### Testing Instructions

Add fields in some XML, example in to mod_custom params:

```xml
<field type="date" name="date" label="Date"/>
<field type="time" name="time" label="Time" filter="SERVER_UTC"/>
<field type="datetime" name="datetime" label="Date Time" filter="SERVER_UTC"/>
```

Edit the fields, make sure the input works and values stored/displayed corectly

Add new fields info to list of fields: https://docs.joomla.org/Standard_form_field_types


dgrammatiko:
Nice one. Not necessarily something for this PR but would be good to have support for these native inputs:
- datetime-local: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local
- month: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/month
- week: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/week

Oh date time is deprecated: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime#browser_compatibility

Fedik:
> Oh date time is deprecated

yes

> datetime-local

This is DateTime here

Unfortunately month and week have  bad support by Browsers, probably for another time :wink: 
Or maybe I make another PR for them, will see.

layouts/joomla/form/field/date.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+ * @package     Joomla.Site
+ * @subpackage  Layout
+ *
+ * @copyright   (C) 2016 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman:
```suggestion
 * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

Its a new file

layouts/joomla/form/field/datetime.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+ * @package     Joomla.Site
+ * @subpackage  Layout
+ *
+ * @copyright   (C) 2016 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman:
```suggestion
 * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

its a new file

libraries/src/Form/Field/DateField.php
```diff
@@ -0,0 +1,159 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  (C) 2009 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman:
```suggestion
 * @copyright  (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

libraries/src/Form/Field/DatetimeField.php
```diff
@@ -0,0 +1,276 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  (C) 2009 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman:
```suggestion
 * @copyright  (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

brianteeman:
Could you update the description please to be datetime not datetimelocal

brianteeman:
What happens with non gregorian calendars?

Fedik:
> What happens with non gregorian calendars?

It only gregorian. It is standart date and datetime-local input, works as described in HTML specs.
No more no less :wink:

Fedik:
> Could you update the description please to be datetime not datetimelocal

Wich description do you mean?
It actualy displaing `datetime-local`, because `datetime` is deprecated (see @dgrammatiko coment)

brianteeman:
image

Fedik:
@brianteeman that is correct.
Xml `<field type="datetime"/>`  (wich is `DatetimeField` class in PHP), will render `<input type="datetime-local">`

brianteeman:
nice and confusing

layouts/joomla/form/field/datetime.php
```diff
@@ -0,0 +1,56 @@
...
+$hint         = strlen($hint)   ? ' placeholder="' . $hint . '"' : '';
+$autocomplete = !$autocomplete ? ' autocomplete="off"' : ' autocomplete="' . $autocomplete . '"';
+$autocomplete = $autocomplete == ' autocomplete="on"' ? '' : $autocomplete;
+$autofocus    = $autofocus ? ' autofocus' : '';
+$pattern      = !empty($pattern) ? ' pattern="' . $pattern . '"' : '';
+$onchange     = !empty($onchange) ? ' onchange="' . $onchange . '"' : '';
+$minAttr      = !empty($min) ? ' min="' . $this->escape($min) . '"' : '';
+$maxAttr      = !empty($max) ? ' max="' . $this->escape($max) . '"' : '';
+
+echo '<input type="datetime-local" name="' . $name . '" id="' . $id . '" value="' . $this->escape($value) . '"'
```
laoneo, line 54: This would increase our min Safari support to 14. Not sure if we want it that new.

laoneo:
I would also deprecate our calendar field in favor of this one.

Fedik:
> I would also deprecate our calendar field in favor of this one.

it not replacement, because it does not support non gregorian calendars

layouts/joomla/form/field/datetime.php
Fedik: It not used in core, just implementation of standart HTML inputs, that will be available as part of Form API,
Who want fallback for old Safari can do a custom field layout with "fallback script"

laoneo: Safari 13 was released two and a half years ago. So I would not call it old. IMO, the core should come up with a decent fallback. Because when a text field is rendered, then the user is allowed to input anything which makes this feature basically useless as nobody is able to use it.

laoneo:
How do we handle gregorian calendars then?

layouts/joomla/form/field/datetime.php
Fedik: I would not bother with it, as I said, who want they can add falback on their own.
It mainly for developers, and they should test their code depend from their target audience.

Fedik:
> How do we handle gregorian calendars then?

I assume you mean "non gregorian".
There a helper JS depend from language, one for jalali https://github.com/joomla/joomla-cms/blob/4.1-dev/build/media_source/system/js/fields/calendar-locales/date/jalali/date-helper.es5.js

layouts/joomla/form/field/datetime.php
laoneo: > I would not bother with it, as I said, who want they can add falback on their own. It mainly for developers, and they should test their code depend from their target audience.

I do not agree with this statement. The core must come up with a proper behavior when the native field is not supported by the browser. I mean when we support from Safari 14 onwords, then this is fine.

laoneo:
Is this loaded then in this field?

layouts/joomla/form/field/datetime.php
laoneo: Got informed, that Apple is offering Safari 15 updates down to iPhone 6s. So I guess we are decently covered here. Didn't know that.

Fedik:
> Is this loaded then in this field?

Nope, that coded specificaly for calendar.js, it cannot be used somwhere else.
Only `calendar` field support "gregorian" and "non gregorian" calendars, that why it cannot be as replacement.

Also see my answer to Brian https://github.com/joomla/joomla-cms/pull/37456#issuecomment-1086619145

laoneo:
If this pr can't be used in core, then I doubt we should add it. Especially when the fields can't be used all the languages we are supporting.

Fedik:
The fields is part of HTML5, that was missed in Form API.

> Especially when the fields can't be used all the languages we are supporting.

They can be used, they translated by Browser.
It offers what HTML specification offers.

Support for different Calendars format it a huge different topic.

brianteeman:
agree with @fedik

attreyee-muk:
Is this issue still open? If yes, can I contribute to it?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37456">issues.joomla.org/tracker/joomla-cms/37456</a>.</sub>

chmst:
@a-muk sure, everyone can contribute :) I'll assign this to you for now.

Fedik:
> If yes, can I contribute to it?

All is needed for this PR is to be tested :wink:

chmst:
so sorry - i have too many tasks.:)  @a-muk yes - now the work means testing

attreyee-muk:
Okay. Thank you @Fedik @chmst

dgrammatiko:
@laoneo could you elaborate on your thumbs down? Shouldn't Joomla implement the full HTML5 specs, and if so why?

laoneo:
> @laoneo could you elaborate on your thumbs down? Shouldn't Joomla implement the full HTML5 specs, and if so why?

You answered it here perfectly https://github.com/joomla/joomla-cms/issues/36346#issuecomment-997437474. The core can never use this field, because we want to support also none gregorian calendars.

dgrammatiko:
> You answered it here perfectly https://github.com/joomla/joomla-cms/issues/36346#issuecomment-997437474. The core can never use this field, because we want to support also none gregorian calendars.

You're mixing different things here. My comment that you are mentioning is about a tables sorting functionality that is totally a Joomla invented/created/coded thing but this PR is about adding some of the W3C platfrom's missing input types. So, yes Joomla shouldn't roll functionality about imaginary use cases (in the case of my comment the dev was going against the Joomla's architecture: items are direct descendants of a single category and asking to extend the functionality to cover their case) but also Joomla should **FULLY** support the HTML5 elements. Also Joomla rolls a few more input types that aren't actually used by the core, so that's not even an argument against, eg:
- https://github.com/joomla/joomla-cms/blob/4.1-dev/layouts/joomla/form/field/color/simple.php
- https://github.com/joomla/joomla-cms/blob/4.1-dev/layouts/joomla/form/field/tel.php
- https://github.com/joomla/joomla-cms/blob/4.1-dev/layouts/joomla/form/field/time.php

laoneo:
> Joomla should FULLY support the HTML5 elements

Why? All I see here is a maintenance burden (even when it is small).

Color is used on many places, tel is used in contact and time in scheduler. Perhaps you should search other examples. Even then, when something can't be used in core, we should not add it.

Perhaps I miss here something, but from a maintainer point of view this looks like something which should be either published as custom field or in the respective extension, but not in core.

brianteeman:
> , tel is used in contact

No its not

https://github.com/joomla/joomla-cms/blob/1ee64271a331317331a86ced18d012342b7b3a86/components/com_contact/forms/form.xml#L169-L171

brianteeman:
Honestly @laoneo its a pathetic argument that you are making. I'm sure we all have better things to do

dgrammatiko:
> but not in core.

Here is were I strongly disagree. For me ALL the HTML5 input fields should be provided from Joomla core for 2 reasons:
- Joomla's byproduct IS HTML5
- the core SHOULD roll those input types so devs cannot use their names for their own fields to prevent collisions (actually every dev should prefixing their input types/names so that won't ever be a thing)


But anyways, that's my point of view. Also one more thought here, if the maintainers of the project feeling intimidated to actually implement the full HTML5 specs then probably they shouldn't be maintainers of a WEB CMS...

brianteeman:
>  time in scheduler. 

Yes it is the only place that uses it today as of 4.1 but it was added to the code base 3 years ago when nothing used it https://github.com/joomla/joomla-cms/pull/26184

Read the closing comments from @wilsonge who actually understands joomla

laoneo:
How should an extension dev use this field then? Tell his customers that his extension is not working with gregorian calendars, but the rest of Joomla does?

If we are discussing on that level, then I'm out of this discussion. Do whatever you guys think is best for Joomla.

brianteeman:
> How should an extension dev use this field then? Tell his customers that his extension is not working with gregorian calendars, but the rest of Joomla does?

You mean non-gregorian calendars.

HLeithner:
I'm in favor for this PR, we can add a comment to the field that the browser support vor non-gregorian doesn't exists (yet) and the developer must decide if he/she needs it or not.

brianteeman:
As long as this is a custom field and core doesnt change to use this field eg for publishing then I 100% agree as stated before.

HLeithner:

TLWebdesign:
I have tested this item :red_circle: unsuccessfully on 00f151da511722d8e2ea3cefc3b4219217429f93<br /><br />Tested this item unsuccessfully on J5.0.3


The time field is not saved. The other two work just fine but what a god awful UI with these fields. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37456">issues.joomla.org/tracker/joomla-cms/37456</a>.</sub>

HLeithner:

degobbis:
I have tested this item :white_check_mark: successfully on 2cd1d7c9762a81a5f2c0fe32d801c0bbeed9ae76<br /><br />It would be nice if the time could also be entered in the DateTime field when opening the calendar, as with the Calendar field.

When the Time field is called up for the first time, only HH:MM is requested. After saving, HH:MM:SS is displayed and I can then also enter the seconds. Here the output should correspond to the input or vice versa.

Otherwise everything works as expected.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37456">issues.joomla.org/tracker/joomla-cms/37456</a>.</sub>

rakibhossainraju:
I have tested this item :white_check_mark: successfully on 2cd1d7c9762a81a5f2c0fe32d801c0bbeed9ae76<br /><br />This works perfectly.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37456">issues.joomla.org/tracker/joomla-cms/37456</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37456">issues.joomla.org/tracker/joomla-cms/37456</a>.</sub>

softforge:
Thank you to all @Fedik and the testers who put the time into getting it done

PR #37457, opened 2022-04-02
[4.2] MediaVersion use Cache instead of Database (reduce 1 db query)

Fedik:
### Summary of Changes

Change getMediaVersion to use Infinity Cache instead of doing 1 useless DB query every time.

Reduce 1 db query for production installation, and 2 db query in debug mode.

### Testing Instructions

Apply patch and make sure MediaVersion added to every script, and all works as before.

Go to "clear cache" page and remove `_media_version` group, then check MediaVersion in to every script again, it should be new now.

None


dgrammatiko:
I have tested this item :white_check_mark: successfully on 7a03b554ff62ac51ac485d5979fc399698e8a798<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37457">issues.joomla.org/tracker/joomla-cms/37457</a>.</sub>


FWIW https://invalidate-cache.dgrammatiko.dev v0.0.10 supports this 4.2 feature

laoneo:
I don't think so that nowadays writing a cache file with our cache setup is faster than a db call. But I like the new possibility to regenerate the media version by clearing the cache.

HLeithner:
having it cached instead in the database is much better, but what I think is that the thing is conceptional wrong isn't? I mean we still add the cms version to external libraries right?

Fedik:
> we still add the cms version to external libraries right?

yes, and no, when WebAsset have a defined version, then it will be used instead of our MediaVersion.
Example this one (from media/vendor/joomla.asset.json):
```json
{
      "name": "awesomplete",
      "type": "script",
      "uri": "vendor/awesomplete/awesomplete.min.js",
      "attributes": {
        "defer": true
      },
      "package": "awesomplete",
      "version": "1.1.5"
 },
```
Will be as `media/vendor/awesomplete/js/awesomplete.min.js?1.1.5`

dgrammatiko:
@HLeithner just to add to what @Fedik said about the static version number:
- external js/css (eg package.json dependencies) get a static version per their actual version used in the package.json
- Most of the Joomla owned js/css get also a static version, a hash of the content of the file with https://github.com/joomla/joomla-cms/blob/4.1-dev/build/build-modules-js/versioning.es6.js

The only exceptions, that still use the media version string are:
- templates
- any script or stylesheet that is added programmatically with WAM (eg without loading a relative joomla.asset.json file)
- all the legacy HTMLHelper::script/stylesheet/etc that have an option for `version`
- the legacy $doc->sddScriptVersion and $doc->addStylesheetVersion methods
- almost anything 3rd PD

The templates in order to use a file content hash based version need some adjustment in the com_templates file editing view (didn't do that with child templates as it was getting to complicated for the testing)
The sporadic files (mainly in the plugins IIRC) just need a joomla.asset.json file and adjusting the way they are imported.

I hope this clears things up

JonasAtZwetschke:
I have tested this item :white_check_mark: successfully on 7a03b554ff62ac51ac485d5979fc399698e8a798<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37457">issues.joomla.org/tracker/joomla-cms/37457</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37457">issues.joomla.org/tracker/joomla-cms/37457</a>.</sub>

laoneo:
Thanks!

PR #37472, opened 2022-04-03
[6.2] Fix multiple rendering calls on custom field

korenevskiy:
When opening a user's page with an article and any custom field. This same field is rendered 2 times. I repeat. One field in the article is rendered 2 times. This issue has been studied. When the field is rendered 2 times, only 1 result is shown on the site, and 2 render result is meaningless. At the same time, All subscriptions of render events called by the dispatcher are also executed 2 times. Thus, it is not possible to write a user data handler without using crutches and props to avoid double processing of the field object.
For example, if you develop a new field with processing data from forms received from the user. Then such data processing needs to be subscribed to the plugin method
```PHP 
PlgFieldsNewForm->onCustomFieldsPrepareField($context, $item, $field)
```
But this method will be called 2 times.
My correction allows you to make 1 render for one field call. While maintaining backward compatibility.

This correction does not matter for static data output. But this amendment is of great importance for the output of dynamic data in user fields of articles. This can be useful when creating a complex graph field with a lot of data in one field. Or a feedback field. But in order for feedback to work in a complex field, it is required to render exactly as many times as it will be displayed on the site.

Rendering occurs when the method is called
```PHP
FieldsHelper::getFields($context, $item);
```

When calling this method, dispatcher subscriptions are called
```PHP
FieldsPlugin->onCustomFieldsBeforePrepareField($context, $item, &$field)
FieldsPlugin->onCustomFieldsPrepareField($context, $item, &$field)
FieldsPlugin->onCustomFieldsAfterPrepareField($context, $item, $field, &$value)
```
Accordingly, for one field, all these methods will be called twice.

However, this happens in the 2 places indicated below
```PHP
PlgSystemFields->display($context, $item, $params, $displayType)
PlgSystemFields->onContentPrepare($context, $item)
```
This fix allows you to use the same render in both calls to these methods.

The text has been supplemented thanks to Richard's comments.
We need to create this plugin in the `/plugins/fields/newmyfield` folder with the class inherited from `Joomla\Component\Fields\Administrator\Plugin\FieldsPlugin`.
In this plugin we create methods `onCustomFieldsPrepareField($context, $item, &$field)`
```PHP
use Joomla\Component\Fields\Administrator\Plugin;
use Joomla\CMS\Factory;
class PlgFieldsNewMyField extends FieldsPlugin {
	public function onCustomFieldsPrepareField($context, $item, &$field){
		if($field->type!='newmyfield')
			return parent::onCustomFieldsPrepareField($context, $item, $field);
		static $counter;
		if(is_null($counter))
			$counter = 0;
		Factory::getApplication()->enqueueMessage('Counter Rendering field:'.$field->type.'  counter:'.++$counter );	

		return parent::onCustomFieldsPrepareField($context, $item, $field);
	}
}
```
Thanks to this test, it is clear that the new amendment calls this method once. Without new corrections, this method was previously called 2 times for one field. You can also place a counter for the test in the rendering file.
`/plugins/fields/newmyfield/tmpl/newmyfield.php`  - this is the layout file for the field.
inside this file we will place a similar counter with a message call.
`Factory::getApplication()->enqueueMessage('Counter Rendering field:'.$field->type.'  counter:'.++$counter );`

in the message call, you need to specify $counter. Since if the messages are completely identical, then 1 message will be displayed instead of 2x. In order to accurately see the number of message calls, we must specify a different text in the message each time. The sequence number in the message will change the text. and the message will be displayed to the number of calls.

Should I blow up all the plugin methods in a row to test the number of calls? But if it is required, we can prescribe a similar example of code for methods.
```PHP
FieldsPlugin->onCustomFieldsBeforePrepareField($context, $item, &$field) 
FieldsPlugin->onCustomFieldsAfterPrepareField($context, $item, $field, &$value)
```
But I think one method is enough for the test.

Also at the end I added a plugin event call once  without multiple calls in to FOREACH.
```PHP
Factory::getApplication()->triggerEvent('onCustomFieldsContentPrepare', [$context, $item, &$item->jcfields]);
```
This will also allow you to change the value of the fields from the beginning of the list after passing all the results.


richard67:
I don’t see any testing instructions in the description of this PR.

korenevskiy:
Hello Richard. I have completed the description

korenevskiy:
I have little experience in writing tests. But it will take me a long time and I will need to study the test classes. Perhaps I will consult here. But so far I have managed to find the error and issue a correction.

richard67:
I did not mean to write unit test or something like that. I meant to write testing instructions for human testers in the description of this PR. There were the headings for these tests in the description template which you get when you create a new PR, but you have removed them. Do you think we had them there for nothing?

korenevskiy:
>  Do you think we had them there for nothing?

You are right, I am aware of the seriousness of these names. I am aware of the importance of the project for many participants. I was tired, and since I don't know the language well, I'm writing a description through a translator. I was wrong, I will definitely improve in the future.
I have completed the description

administrator/components/com_fields/src/Helper/FieldsHelper.php
```diff
@@ -178,7 +178,7 @@ function ($f)
 
 			$fieldValues = self::$fieldCache->getFieldValues($fieldIds, $item->id);
 
-			$new = array();
+			$item->jcfields = array();
```
laoneo: Attaching here the fields should not be done. The object should not come back differently than it was passed to the function.

korenevskiy: I'll think about how to do it.

korenevskiy:
And what does the **`PBF`** hash tag mean?

richard67:
> And what does the **`PBF`** hash tag mean?

Pizza, Bugs and Fun. At Joomls day USA there was such an event, and we flagged a bunch of PRs with that label so people test them. The label will be removed in a few days.

korenevskiy:
And how do I create another PR on the `FieldsHelper.php` file?
But the new PR is not related to the topic of this PR. But since the current PR will change the code of this `FieldsHelper.php` file and make a shift in the number of lines. That new PR may conflict with the current PR, due to the fact that the amendments affect the same code.

richard67:
@korenevskiy Normally you create a new branch for each PR, based on the target branch of the upstream repository (= this here)

richard67:
P.S. In case of this PR you have done it right and have created a separate branch. For the new PR just make another branch based on your 4.1-dev branch and not of the one from this PR here.

HLeithner:

korenevskiy:
@laoneo Please check.

korenevskiy:
@HLeithner Is it possible to transfer this PR to Joomla 5 ?

HLeithner:

korenevskiy:
@laoneo  
Please explain what is wrong with this PR

Septdir:
It is not correct to do this in the dislpay plugin method.
In fact, an object without jcfields can get there;
Either incorrect for a given context or parameters. For example, they were changed before the trigger was called.

HLeithner:

Hackwar:
@korenevskiy Can you please fix the conflict?

HLeithner:

plugins/system/fields/src/Extension/Fields.php
```diff
@@ -421,11 +422,18 @@ private function display($context, $item, $params, $displayType)
...
+
         if ($fields) {
             foreach ($fields as $key => $field) {
                 $fieldDisplayType = $field->params->get('display', '2');
 
                 if ($fieldDisplayType == $displayType) {
+                    /*
+                    * Event allow plugins to modify the output of the field before it is display
+                    */
+                    Factory::getApplication()->triggerEvent('onCustomFieldsBeforeDisplay', [$context, $item, &$field, $displayType, &$params]);
```
richard67:
```suggestion
                    $this->getApplication()->triggerEvent('onCustomFieldsBeforeDisplay', [$context, $item, &$field, $displayType, &$params]);
```

plugins/system/fields/src/Extension/Fields.php
```diff
@@ -493,6 +501,13 @@ public function onContentPrepare(Content\ContentPrepareEvent $event)
         foreach ($fields as $key => $field) {
             $item->jcfields[$field->id] = $field;
         }
+
+        PluginHelper::importPlugin('fields');
+
+        /*
+        * Event allow plugins to modify the the fields it is with content prepare
+        */
+        Factory::getApplication()->triggerEvent('onCustomFieldsContentPrepare', [$context, $item, &$item->jcfields]);
```
richard67:
```suggestion
        $this->getApplication()->triggerEvent('onCustomFieldsContentPrepare', [$context, $item, &$item->jcfields]);
```
Hmm, seems I mossed one.

korenevskiy:
I checked how the custom fields plugin works in Joomla 6. When displaying an article field for a client page, the layout is still called and rendered 2 times.
Thus, this PR is still relevant and necessary.

PR #37477, opened 2022-04-04
updated index.html file

hrushi2003:
Pull Request for Issue # .


Fedik:
Thanks for you try, but this changes not needed, an existing index.htmk is fine as it is.

richard67:
@hrushi2003 These index.html files which we have in some otherwise empty folders are just placeholders needed for the zip method to build the packages so these folders are included in the package. Otherwise, if empty, they would not be in the package. That's the only use of these index.html files so they do not need your changes.

PR #37489, opened 2022-04-05
[4.x] com_templates when using pdo

brianteeman:
Pull Request for Issue #37487 .

### Testing Instructions
on a joomla install using mysql pdo (eg in configuration.php it says `public $dbtype = 'mysql';` not mysqli
open the template manager

### Actual result BEFORE applying this Pull Request
the path is fubar

### Expected result AFTER applying this Pull Request
the path is correct and no error messages when using the template manager

Issue #37487, opened 2022-04-05
[4.1] PDO - Wrong path of template files in backend


Sandra97:
Hello,

After installing a site on a server with PDO, I noticed that the path of my frontend template files is changed and wrong.   

### Expected result
/templates/nameofmytemplate

### Actual result
/**administrator**/templates/nameofmytemplate

### System information (as much as possible)
Joomla 4.1.2
Database 10.3.34-MariaDB-0ubuntu0.20.04.1 
PHP 7.4.3

Any idea how to fix this? (couldn't find something bout that on GH)

richard67:
@Sandra97 How have you installed the site? Using the normal Joomla full installation package? Or anything special, like e.g. a quickstart installation from the hosting provider or template provider? I ask because I've seen the same issue reported at several places (forum, social media) for quickstart installations.

richard67:
P.S.: The workaround is to use MySQLi instead of Mysql (PDO) if you have both available.

Sandra97:
The site works as expected on my server with MySQLi. I re-installed it on the production server of the client using kickstart from Akeeba. Unfortunately, they have on PDO on their server

toivo:
Confirmed:  the path to the front end template is displayed incorrectly when 'MySQL (PDO)' has been selected as Database Type, shown as 'mysql' in the file configuration.php.  

Change it to 'MySQLi' and the path is displayed correctly.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37487">issues.joomla.org/tracker/joomla-cms/37487</a>.</sub>

richard67:
@toivo What do you mean with "displayed"? Displayed where?

Sandra97:
He means that if you edit configuration.php by replacing `public $dbtype = 'mysql'; `by `public $dbtype = 'mysqli';` then administrator is removed and the frontend template path is then correct

richard67:
> He means that if you edit configuration.php by replacing `public $dbtype = 'mysql'; `by `public $dbtype = 'mysqli';` then administrator is removed and the frontend template path is then correct

I understood that, I just don't understand what the "displayed" means. Are is the template folder of the frontend really located at the wrong place, so after changing the database type it needs to make a new installation to get it at the right place? Or is it just shown wrong somewhere in backend so after the change of the db type all is fine, without reinstalling?

[...]

PR #37497, opened 2022-04-06
[4.1] Cast client_id in template to integer to support mysql PDO

bembelimen:
Pull Request for Issue #37487 .

### Summary of Changes
See #37489 for the details


administrator/components/com_templates/src/Model/TemplateModel.php
```diff
@@ -681,6 +681,9 @@ public function &getTemplate()
 			{
 				$this->template = $result;
 
+				// Client ID is on PDO mySQL not an iteger, so enforce here
```
richard67:
```suggestion
				// Client ID is on PDO mySQL not an integer, so enforce here
```

laoneo:
Is this also needed on the templateS model?

bembelimen:
> Is this also needed on the templateS model?

The usage I found was once in the model itself where it was used only as key and in the default.php where it was already casted. So I would say "no", but in general we have this problem on many places I guess (with other fields)

brianteeman:
> // Client ID is on PDO mySQL not an integer, so enforce here

This doesnt make sense. Not sure exactly what you are trying to say. Guessing at

// Client ID is not an integer, so enforce here

brianteeman:
> in general we have this problem on many places I guess (with other fields)

Shouldnt they be addressed with this PR - at least the ones relating to the template

bembelimen:
> > in general we have this problem on many places I guess (with other fields)
> 
> Shouldnt they be addressed with this PR - at least the ones relating to the template

Do you see any spontanious? It was more a general remark. But I'm happy to fix it, if there are some very obvious ones.

alikon:
this is a funny one...
i'm not able to replicate it...just tested on a normal cfg  :smile: 
even if i'm on mysql pdo 
anyway 
1. the executed query is 
 ```sql
SELECT `extension_id`,`client_id`,`element`,`name`,`manifest_cache`
FROM `#__extensions`
WHERE `extension_id` = :pk AND `type` = 'template'
 ```

2. the client_id field is declared as _tinyint_ so cannot be **string** 
https://github.com/joomla/joomla-cms/blob/126ed6a1d4d4661d839096ca10bb2102486ae0dd/installation/sql/mysql/base.sql#L126

oh wait ...probably i'm not on MariaDB  ? but 
debug tell me this
debug

so it should be digged a bit more deep

toivo:
I have tested this item :white_check_mark: successfully on 8892cea242bd555f5a8f88cabd8b39526288fc94<br /><br />Tested successfully in Joomla 4.1.3-dev of 21 April in Wampserver 3.2.7 using MySQL 8.0.27 and PHP 8.0.15.

This PR fixes the same issue with MySQL (PDO) selection breaking the path to the template files as the already closed #37489 did - https://github.com/joomla/joomla-cms/pull/37489.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37497">issues.joomla.org/tracker/joomla-cms/37497</a>.</sub>

Sandra97:
I have tested this item :white_check_mark: successfully on 8892cea242bd555f5a8f88cabd8b39526288fc94<br /><br />Tested on the site on which I faced the issue (4.1.2). Path is now correct. Thanks<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37497">issues.joomla.org/tracker/joomla-cms/37497</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37497">issues.joomla.org/tracker/joomla-cms/37497</a>.</sub>

alikon:
still a funny one, my observations... are simply still ignored
i give up

richard67:
@alikon Well I was able to reproduce the issue. And in past I have observed that depending on the driver and the database kind and Version columns with integer data types had been converted to strings.

alikon:
i doubt still, anyway....
probably mine are not valid...fine
go forward

alikon:
mine reply was not in synch with your last one ....probably

alikon:
still stand anyway... simply ignore my comments mine are lastly ... killed from someone's thumbs down --- even if I declare as  a DIRTY fix
Note to myself: stop doing things
:unicorn:

richard67:
?

alikon:
Can we have a chat in 15 min?


Il gio 21 apr 2022, 21:36 Richard Fath ***@***.***> ha
scritto:

richard67:
On Glip? If necessary, ok

alikon:
thanks @richard67  to listen

richard67:
We should maybe dig deeper into what causes different behavior of different versions of MySQL and MariaDB, but for the mean time this PR as a quick and maybe dirty fix is ok, I think.

alikon:
you know i'm the man for a quick & dirty fix
...in some tables --- are called workaround---

alikon (commented):
should  be digger more deep

laoneo:
So what is here the exact issue that it is considered a dirty hack?

richard67:
> So what is here the exact issue that it is considered a dirty hack?

@laoneo Maybe the wrong words, but as long as we don't know why we get an integer or a string for an integer data type column depending on database type and version, it leaves me with the feeling that we fix the symptom and not the root cause. But that's just a gut feeling and I may be wrong.

laoneo:
Valid concern.

HLeithner:
The problem with the MySQL drivers are that it depends if PDO, MySQLi (MySQL) if it typecasts correctly and if we use prepared statements or not. it has been fixed (in a b/c incompatible way) in php 8.1 https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.pdo.mysql 

if we fix this here we should add a deprecation notice (or removal notice with J5, where the minimum version will be php 8.1). I'm surprise that this is the only problem found yet for this behavior.

richard67:
I think @HLeithner 's [comment above](https://github.com/joomla/joomla-cms/pull/37497#issuecomment-1109482093) and the [link it refers to](https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.pdo.mysql) explain what happens.

For the mean time, until we have fixed the root cause, which we cleanly could do with J5 as mentioned by Harald, we will need such fixes like provided here.

Therefore I will merge this PR now.

richard67:
Thanks.

PR #37498, opened 2022-04-06
[4.2] Add interfaces for the objects who need a user

laoneo:
### Summary of Changes
As many classes need a user to check permissions or get an email address from it, there is no common way to get a current user instance. This pr introduces a new interface and trait to get/set a current user on an object. The [identity aware trait](https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Application/IdentityAware.php) offers a similar functionality but is intended to be used on applications only. But I'm open here for suggestions if we should reuse the identity trait or go with the approach in this pr.

Also I'm not sure about the name of the new interface or trait, also here open for suggestions.

This pr is mainly needed as @wilsonge doesn't want to inject the application into the models or views. If the app would be available, then we can just load the identity from it.

### Testing Instructions
In the back end, open the articles list.

Works.

Works.


libraries/src/User/CurrentUserTrait.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * The current user object.
+	 *
+	 * @var    User
+	 * @since  __DEPLOY_VERSION__
+	 */
+	private $currentUser;
+
+	/**
+	 * Get the application identity.
```
brianteeman: application ?? Is this a copy paste error?

libraries/src/User/CurrentUserTrait.php
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * @return  User
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getCurrentUser()
+	{
+		if (!$this->currentUser)
+		{
+			@trigger_error(sprintf('The current user must be set on %s. This will not be catched in 5.0.', __CLASS__), E_USER_DEPRECATED);
```
brianteeman: Not english. But sorry I dont understannd it enough to correct it

laoneo: When a user is not set that this will not fallback to the app identity, instead of it returns null.

richard67: Past participle of „catch“ is „caught“, not „catched“.

libraries/src/User/CurrentUserTrait.php
laoneo: Yups

brianteeman:
will you be replacing getIdentity everywhere?

laoneo:
Not really as this is different than getIdentity in terms of logic. GetIdentity is basically the identity of the application which comes from the authentication process. While this here can be anything where the view or model should work with. But as I said, I'm open here for suggestions as the similarities are very obvious.

brianteeman:
Just when I thought I understood the reason for this PR

rdeutz:
Why not make a trait to get the application object and the go from there?

laoneo:
Don't think so it would be a good decision to pass around the application in whole core. Like that there is only a dependency to a simple user and not the app. It brings the benefit to be usable on places where no application is available.

libraries/src/MVC/Controller/BaseController.php
```diff
@@ -619,7 +625,14 @@ protected function createView($name, $prefix = '', $type = '', $config = array()
 	{
 		$config['paths'] = $this->paths['view'];
 
-		return $this->factory->createView($name, $prefix, $type, $config);
+		$view = $this->factory->createView($name, $prefix, $type, $config);
+
+		if ($view instanceof CurrentUserInterface && $this->app->getIdentity())
+		{
+			$view->setCurrentUser($this->app->getIdentity());
+		}
```
roland-d, line 633: @laoneo I am thinking this should be done in the MVCFactory. We already have these functions there:
```
$this->setFormFactoryOnObject($view);
$this->setDispatcherOnObject($view);
$this->setRouterOnObject($view);
```
We can add an `$this->setUserOnObject($view);`. This way we have all these calls in the same place.

laoneo: When you create a model or view in the MVCFactory, then you don't have an app object.

roland-d: We do a `Factory::getApplication()` on line 182. Can't that be used to get the identity?

laoneo: That0s deprecated code

laoneo: https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/MVC/Factory/MVCFactory.php#L172-L183

roland-d: Yes, I saw after your previous comment and already resolved the conversation.

joomdonation:
Here, the current user object will only be available in model if the model is created by our normal MVC workflow (users access to the component via web browser). What if model is created by by MVCFactory ? In this case, `getCurrentUser()` would return null and we will get error (fatal error or at least logical error). Am I right?

laoneo:
Nope https://github.com/joomla/joomla-cms/pull/37498/files#diff-cd69155c336e480348cacd600fe8300407ed0219ce97ed7f83324fef100d311dR41

joomdonation:
That's not current user, but an empty user object?

laoneo:
After reading your comment a couple of times I think I do understand what you mean now. So when an extension is loading a model like `$app->bootComponent('content')->getMVCFactory()->createModel('Article', 'Administrator')` then the user is not the identity from the global application object. Is this what you mean? In this case I would probably fallback to `Factory::getApplication()->getIdentity()` [here](https://github.com/joomla/joomla-cms/pull/37498/files#diff-cd69155c336e480348cacd600fe8300407ed0219ce97ed7f83324fef100d311dR41) instead of an empty user and deprecate that usage for 5.

joomdonation:
Yes, that's what I wanted to say.

roland-d:
@laoneo We have a codestyle failure:
```
FILE: /********/src/libraries/src/User/CurrentUserTrait.php

FOUND 1 ERROR AFFECTING 1 LINE

 43 | ERROR | [x] No space found after comma in argument list

    |       |     (Generic.Functions.FunctionCallArgumentSpacing.NoSpaceAfterComma)
```

roland-d:
@joomdonation Is everything OK for you now on this PR?

joomdonation:
@roland-d The logical bug I concerned before is fixed with latest change, so it is fine now. The architecture change in this PR, right or wrong, I'm not qualify enough to say, sorry. So it's up to you to decide :D .

roland-d:
Thanks everybody

PR #33390, opened 2021-04-28
[4.0] Removing GUI parts of FTP layer

Hackwar:
This removes the GUI parts of the FTP layer. This PR is the counterpart to #31857. The PLT needs to decide if it wants to either merge this or the other PR.

This PR is not entirely complete. There is still FTP code in the joomlaupdate component and the FTP layer has not been removed from the libraries folder.

Issue #35380, opened 2021-08-26
[4.0] isSite isAdmin etc


brianteeman:
Going to repeat what I have said many times before that completely removing these is breaking a lot of sites upgrading to 4.

- Yes I know its been marked as deprecated for a long time (but not with a deprecated log)

- Yes I know its documented on the backwards compatibility page (what user ever reads that)

- Yes I know the pre-update check reports on extensions not marked as j4 compatible (but with so many extensions not reporting anything - for many different reasons)

Is it really so hard to put something in place so at least this code doesnt break a site. **This is just being stubborn and is just like when Andrew said just do a search and replace and add legacy to everything and who cares that people just stopped using joomla- lets not repeat the same mistake**

References https://github.com/joomla/joomla-cms/issues/30913 https://github.com/joomla/joomla-cms/pull/22492

OctavianC:
My 2 cents: users don't really care why some functions are deprecated yet they are the majority that's affected by these changes. We should stop removing functions that have been available for years, especially when there's so little overhead to them (isSite() + isAdmin() were just proxy functions). This is why WordPress is so popular, I can literally throw in an old plugin with the latest version of WordPress and it will still work. As a developer I don't like their code at all, but then again, users don't see code, they just use the CMS, and theirs doesn't crash.

Also, lately we've been flooded by messages such as "is your extension compatible with 3.10 / 4.0" (even though they are marked as such). Users have lost the faith in updates and are scared to do them, rightfully so, given that any new version shift breaks something. The Pre-Update Checker is right now our most time consuming 'job', eating up support time for issues that are non-existent. But I digress...

wilsonge:
The reason we did this is because people tended to use the methods as an if/else where thing went if site else admin (or vice versa). But 80% of these things basically assumed a html page and these just broke cli especially but also to a slightly lesser extend webservices.

Removing doesn't necessarily solve the problem but it basically is developer awareness because as you said Brian they don't read the docs!

sozzled:
Yes, to Brian.  A lot of J! 4 websites are breaking because [mostly 3P] system plugins still have isClient() and isAdmin() coded.  And we've known that this was going to happen for a very long time.  There's ample evidence on [The Joomla Forum](https://forumjoomla.org) (for those who visit this site) where people are discovering that their J! 4 websites are capsizing and no amount of "gentle encouragement" seems to be having any great effect.

[...]

PR #37518, opened 2022-04-08
[4.2] Deprecate the CMS filesystem package in favor of the framework

laoneo:
### Summary of Changes
This pr deprecates the CMS filesystem package in favor of the framework one. Replacing the references of the CMS filesystem library should be done in followup pr's as a search and replace doesn't really work on all classes and functions.

Followup of #33390.

Only code review.

### DIFF
--- a/libraries/src/Client/ClientHelper.php
+++ b/libraries/src/Client/ClientHelper.php
@@ -16,6 +16,8 @@ use Joomla\CMS\Factory;
  * Client helper class
  *
  * @since  1.7.0
+ *
+ * @deprecated  5.0 No replacement
  */
 class ClientHelper
 {
--- a/libraries/src/Client/FtpClient.php
+++ b/libraries/src/Client/FtpClient.php
@@ -85,6 +85,8 @@ if (!\defined('FTP_NATIVE'))
  * FTP client class
  *
  * @since  1.5
+ *
+ * @deprecated  5.0 Use the filesystem package from the framework
  */
 class FtpClient
 {
--- a/libraries/src/Factory.php
+++ b/libraries/src/Factory.php
@@ -805,6 +805,7 @@ abstract class Factory
 	 *
 	 * @see     Stream
 	 * @since   1.7.0
+	 * @deprecated  5.0 Use the native PHP streams or build one from the framework package
 	 */
 	public static function getStream($usePrefix = true, $useNetwork = true, $userAgentSuffix = 'Joomla', $maskUserAgent = false)
 	{
--- a/libraries/src/Filesystem/File.php
+++ b/libraries/src/Filesystem/File.php
@@ -21,6 +21,8 @@ use Joomla\CMS\Log\Log;
  * A File handling class
  *
  * @since  1.7.0
+ *
+ * @deprecated  5.0 Use the filesystem package from the framework
  */
 class File
 {
--- a/libraries/src/Filesystem/File.php
+++ b/libraries/src/Filesystem/File.php
@@ -40,6 +40,8 @@ class File
 	 * @return  string  The file extension
 	 *
 	 * @since   1.7.0
+	 *
+	 * @deprecated  5.0 Use native pathinfo
 	 */
 	public static function getExt($file)
 	{
@@ -665,6 +667,8 @@ class File
 	 * @return  boolean  True if path is a file
 	 *
 	 * @since   1.7.0
+	 *
+	 * @deprecated  5.0 Use native is_file instead
 	 */
 	public static function exists($file)
 	{
--- a/libraries/src/Filesystem/FilesystemHelper.php
+++ b/libraries/src/Filesystem/FilesystemHelper.php
@@ -16,6 +16,8 @@ namespace Joomla\CMS\Filesystem;
  * Holds support functions for the filesystem, particularly the stream
  *
  * @since  1.7.0
+ *
+ * @deprecated  5.0 Use the filesystem package from the framework
  */
 class FilesystemHelper
 {
--- a/libraries/src/Filesystem/Folder.php
+++ b/libraries/src/Filesystem/Folder.php
@@ -480,6 +480,8 @@ abstract class Folder
 	 * @return  boolean  True if path is a folder
 	 *
 	 * @since   1.7.0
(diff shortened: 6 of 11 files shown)


joomdonation:
The two packages are not the same, so if we deprecate the CMS package like this, it will cause trouble for developer using the methods from CMS package which are not available in the Framework package. For example, when I look at the `File` class for the methods which I am using in my extensions, there are already difference:

- `getExt`, `exists` methods are not available in the framework package.
- `upload` method in the CMS package has different signature (thus different logic) with the `upload `method in framework package.

Just name of a few by looking at it quickly. What will we have to do with these differences/changes?

laoneo:
That's why I wrote, we can't just search and replace. You have to check it cases by case and the same must be done by the extension developers. That's why we deprecate it now so all have a enough time to migrate. For the functions which do not exist in the framework package we must add a special deprecate note. If there is a different signature, then I guess the developer can figure out by himself how to do the transition.

Fedik:
I agree with @joomdonation. Also much of that "difference" used by CMS itself. 
Deprecate without a proper replacemen is a bad idea.

laoneo:
What has no replacement? You have to become specific, such general statements doesn't bring any value. There is for every deprecation a replacement (except for the helper) in this pr and when individual functions are removed, then we can add a specific deprecate message.

Fedik:
Maaan, that is your PR you should have been checked what consequences it can have for CMS :wink: 

Here is my list (I did not checked sub-classes):
---
**File**:
getExt :x: 
append :x: 
exists :x: 
**upload** - have a criticall difference, and cannot be replaced directly
invalidateFileCache - non criticall difference
canFlushFileCache  :x: 

**Folder**:
exists :x:

**Helper**:
fileUploadMaxSize :x:
parseSize :x:
parseSizeUnit :x:

**Patcher**:
reset :x:

**Path**:
removeRoot :x:

**Stream**:
filesize :x:

**ClientHelper** class not exists.

---

What I meant, that it should be first replaced in CMS, to make sure all works correctly, so we can mark it as deprecated safely.
Not like: we mark it as deprecated and later figure out what to do.

Addittionaly, because it is static classes, each method should trigger `E_USER_DEPRECATED`:
```php
@trigger_error(
	sprintf('%s is deprecated. Use Joomla\Filesystem\Foo::bar() instead.', __METHOD__),
	E_USER_DEPRECATED
);
```
For non static classes it should be enought to trigger it once in class constructor.

laoneo:
The only reason we left the CMS filesystem package in core for 4 was because of the FTP layer. We did it many times that we deprecated stuff and then replaced it afterwards, see JFactory functions or *::getInstance(). As I wrote in the description, when this is merged, then we need to replace it with the framework package on a case by case basis to see nothing breaks. What I want, is to have one way to use the filesystem and not duplicated stuff.

brianteeman:
> The only reason we left the CMS filesystem package in core for 4 was because of the FTP layer.

Really? The report by @fedik suggests otherwise

> As I wrote in the description, when this is merged, then we need to replace it with the framework package on a case by case basis to see nothing breaks. 

As per @nibra previous comments about deprecations this PR should also replace all the usage in the core of the CMS filesystem layer with the framework library. The core should not be using anything that is deprecated.

Fedik:
> The only reason we left the CMS filesystem package in core for 4 was because of the FTP layer.

Yeah, that probably was a general reason. However if look in deep there a "little more details". 

I more concerning about `File::upload()`, this also kind of "security changes".
A rest can be handled easily somehow (but better "before" than "after").

laoneo:
> > The only reason we left the CMS filesystem package in core for 4 was because of the FTP layer.
> 
> Really? The report by @Fedik suggests otherwise
> 
> > As I wrote in the description, when this is merged, then we need to replace it with the framework package on a case by case basis to see nothing breaks.
> 
> As per @nibra previous comments about deprecations this PR should also replace all the usage in the core of the CMS filesystem layer with the framework library. The core should not be using anything that is deprecated.

I general I do agree when single functions are deprecated, but here we deprecate a full namespace. If I would replace every occurrence in all the 120 files then the pr is not testable. So I suggest to deprecate it first and then replace it step by step. I really don't understand why we have again to discus this back and forth instead of just going forward.

brianteeman:
> I really don't understand why we have again to discus this back and forth instead of just going forward

Where was the previous discussion and decision

Bakual:
We should not deprecate things we still use ourself. That's imho a bad practice we used in the past.
Replace the usage and then deprecate is the better approach. Otherwise we have deprecation warnings in the code all over the place which is bad.

Fedik:
> If I would replace every occurrence in all the 120 files then the pr is not testable. So I suggest to deprecate it first and then replace it step by step.

I thought about it again :)
You can (if you want) do step by step in parallel PR(s), replace methods that does **not use** FTP, like:
```
File::getExt
File::stripExt
File::makeSafe
File::invalidateFileCache
File::canFlushFileCache
File::exists

Folder::exists
Folder::files
Folder::folders
Folder::listFolderTree
Folder::makeSafe

etc.
```
Because we cannot drop FTP on the middle. Then we can merge this PR to deprecate a rest.
But with exception, we cannot deprecate `File::upload` because there no direct replacement, and so `File` class also cannot deprecate, only methods in it.

And still: every public static method should trigger `E_USER_DEPRECATED`, even when whole class is deprecated (because it is static classes).
```php
@trigger_error(
	sprintf('%s is deprecated. Use Joomla\Filesystem\Foo::bar() instead.', __METHOD__),
	E_USER_DEPRECATED
);
```
With info what to use instead.
Or we end up with stuff like this #35380 again.

That I think will be better way to go.

laoneo:
The problem is that pr's get tested very slowly, especially from me. So when I want to do all these pr's it will take ages and when I do afterwards the deprecation, then we arrive at Joomla 7 with the filesystem deprecated. So better to do it now, soe extension developers can clearly see that they shouldn't use the CMS filesystem package anymore and we have enough time to do the replacement.

Fedik:
@laoneo if you will decide to make such PRs, ping me, I try to test,
Ideally a couple small ones: one for Files, one for Folder so on.

PR #37530, opened 2022-04-11
[4.2] Sampledata testing message

heelc29:
### Summary of Changes

check sampledata type before execute `onAjaxSampledataApplyStep9` in testing plugin

### Testing Instructions

install a sampledata plugin with more than 9 steps

Message `Testing Sample Data has been installed!` appears

Message `Testing Sample Data has been installed!` doesn't appears

none


brianteeman:
obviously this is not correct

heelc29:
> obviously this is not correct

@brianteeman What do you mean? The PR or that the function is called during execution of other sampledata plugins?

brianteeman:
think about it

brianteeman:
1. this php file is for that specific set of sample data which only has 9 sets
2.  the number of steps is also defined on line 87
3. This plugin is neveer shipped with core

heelc29:
I know is that the "Testing Sample Data" is not included in any release (production sites)

I'm not sure if I described it correctly above. I don't want to run the testing plugin, I want to run another plugin (with more than 9 steps in test environment)


This is the result:
image


Because `com_ajax` doesn't trigger any special plugin, just the function name `onAjaxSampledataApplyStep9`
https://github.com/joomla/joomla-cms/blob/bcd54b6c14a0441458a606b31903f60745c6c7d7/components/com_ajax/ajax.php#L158


Therefore, every plugin has to check the requested type (and session token) before execution
https://github.com/joomla/joomla-cms/blob/bcd54b6c14a0441458a606b31903f60745c6c7d7/plugins/sampledata/blog/blog.php#L95-L98


PS: The message of the last step is never shown because the finished message appears at the same moment

image
https://github.com/joomla/joomla-cms/blob/bcd54b6c14a0441458a606b31903f60745c6c7d7/build/media_source/mod_sampledata/js/sampledata-process.es6.js#L87-L104

brianteeman:
>  I don't want to run the testing plugin, I want to run another plugin (with more than 9 steps in test environment)

I know.  So why are you not writing your plugin correctly. editing this plugin makes no sense (plus as already pointed out the change is wrong anyway)

heelc29:
> So why are you not writing your plugin correctly

I don't know exactly what you mean by that (all is core)? I can't change anything on "my" plugin so that the message from plg_sampledata_testing doesn't appear, right? (even if this only happens in the test environment)

> pointed out the change is wrong anyway

Why (I can only understand point 3)? The other plugins also have this condition in the last step.
https://github.com/joomla/joomla-cms/blob/bcd54b6c14a0441458a606b31903f60745c6c7d7/plugins/sampledata/blog/blog.php#L1845-L1850
https://github.com/joomla/joomla-cms/blob/bcd54b6c14a0441458a606b31903f60745c6c7d7/plugins/sampledata/multilang/multilang.php#L466-L471

brianteeman:
>  I can't change anything on "my" plugin so that the message from plg_sampledata_testing doesn't appear, right?

Wrong. Your plugin should be completely indpendent

heelc29:
> Wrong. Your plugin should be completely indpendent

I think its independent ... (contains only sleep)

Uploded to: https://github.com/heelc29/plg_sampledata_joomla

brianteeman:
I repeat again and for the last time. Your own sample data plugin has nothing to do with the sample data plugin in core that this pr is editing

heelc29:
> Your own sample data plugin has nothing to do with the sample data plugin in core

Yes, of course, I never said that.

For the moment if no one use any 3rd party sampledata plugins with more then 9 steps (in test environment) the situation does not occur. 

If (someone in future 😉) removing some steps of testing plugin or adding some steps to multilang plugin (currently 8 steps) this bug will occur and why should it not fixed now.

If this is not necessary, we can close the PR.

HLeithner:

brianteeman:
**This is still completely wrong.** This file is ONLY used by the testing plugin. it is not used by any other plugin in core and it will not be used by any plugin someone adds. If in the future more steps are added to the multilang sample data then it doesnt matter what change is made in this file - it will have zero effect. Please just think about it.

PR #37534, opened 2022-04-11
[4.2] Installation updateuserids

brianteeman:
### Summary of Changes
during the installation we run various queries to update default data with the superuseerid created during the install

### Testing Instructions
Clean install from the generated full package
Check that you never get an error message 
image

Running queries on tables that have no data 

No change to the generated install

none


richard67:
In 3.10 the removed entries might be needed because one can chose to install sample data at the end of a new installation. But on 4.1 these entries seem to be not needed because the tables are empty after a new installation. So by review this PR seems to be right.

wilsonge (approved):

installation/src/Model/ConfigurationModel.php
```diff
@@ -295,17 +295,8 @@ protected function updateUserIds($db)
 
 		// Update all core tables created_by fields of the tables with the random user id.
```
Quy:
```suggestion
		// Update core tables created_by fields of the tables with the random user id.
```
Maybe?

brianteeman:
rebased to 4,2 now that there will be no further 4.1 releases

wilsonge:
This is clearly correct given the change in sample data in J4 and given it's not getting tests choosing to merge this on review.

brianteeman:
Thanks

PR #37536, opened 2022-04-11
[5.4] Task Scheduler require custom cron execution rules to be filled out

magnussinger:
Pull Request for Issue #37313

### Summary of Changes

When you create a new scheduled task and select custom execution rules, and you don't specify a execution time, you can't save it now

### Testing Instructions

1. Create Demo Task - Sleep
Select Cron Expression Advanced as Execution Rule
Do not select all values in area Cron match
Try to save, Joomla will show you that you still have empty fields
2. Change execution rule to something else than Cron Expression Advanced and check if that is still working as before.

1. You were able to click save, but then ran into an error

1. You see that there are required fields left and you cannot save
2. Other execution rules than Cron Expression Advanced are still working.

None


richard67:
@MagnusSinger My comment from another, closed PR for the same issue also applies here, see https://github.com/joomla/joomla-cms/pull/37350#issuecomment-1075529830 .

Also the testing instructions should include a test that nothingness broken for other kinds of execution rules that the advanced cron expression .

magnussinger:
@richard67 thanks for the hint, I'll have a look at it

bembelimen:
See: https://github.com/joomla/joomla-cms/pull/37537#issuecomment-1096275706

magnussinger:
@bembelimen @richard67 I tried to control it now via JS, please let me know if this is okay

build/media_source/com_scheduler/js/admin-view-config-helper.es6.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * @copyright   (C) 2021 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman:
```suggestion
 * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

its a new file

richard67:
@dgrammatiko Could you do a quick review on the JS in this PR? To me it looks good, but I am not really a JS expert. Thanks in advance.

richard67:
@MagnusSinger Could you extend your testing instructions by a test if other execution rules than advanced cron expressions don't require the cron parameters (i.e. mainly a test for your last change)? Thanks in advance. In general it's important that testing instructions don't cover only the test for the fixed issue but also the test that other behaviour of the changed code is not broken.

bembelimen:
We need for sure also a server side validation for this (e.g. in the models validate method)

build/media_source/com_scheduler/js/admin-view-config-helper.es6.js
```diff
@@ -0,0 +1,75 @@
+/**
```
dgrammatiko, line 1: can we replace the JS with this:

```js
/**
 * @copyright   (C) 2023 Open Source Matters, Inc. <https://www.joomla.org>
 */

if (!window.Joomla) {
  throw new Error('Joomla API was not properly initialised!');
}

const ruleTypesElement = document.querySelector('#jform_execution_rules_rule_type');

const elements = [
  '#jform_execution_rules_cron_expression_minutes',
  '#jform_execution_rules_cron_expression_hours',
  '#jform_execution_rules_cron_expression_days_month',
  '#jform_execution_rules_cron_expression_months',
  '#jform_execution_rules_cron_expression_days_week'
];

function toggleRequired(process = 'add') {
  if (!['add', 'remove'].contains(process)) {
    return;
  }

  elements.forEach((elementSelector) => {
    const element = document.querySelector(elementSelector);
    if (!element) {
      return;
    }
    const label = document.querySelector(`${elementSelector}-lbl`);
    if (process === 'add') {
      element.setAttribute('required', '');
      element.classList.add('required');
      if (label) {
        label.classList.add('required');
      }
    } else {
      element.removeAttribute('required');
      element.classList.remove('required');
      if (label) {
        label.classList.remove('required');
      }
    }

  });
}

const handleChange = (event) => event.target.value === 'cron-expression' ? toggleRequired('add') : toggleRequired('remove');

if (ruleTypesElement) {
  handleChange({target: ruleTypesElement});
  
  ruleTypesElement.addEventListener('change', handleChange);
}

```

Thanks

dgrammatiko (dismissed):

build/media_source/com_scheduler/js/admin-view-config-helper.es6.js
magnussinger: Sure, I'll have a look at it and change it soon

HLeithner:

build/media_source/com_scheduler/js/admin-view-config-helper.es6.js
magnussinger: Okay, so I reviewed your code and added it, thanks!

HLeithner:

build/media_source/com_scheduler/js/admin-view-config-helper.es6.js
```diff
@@ -0,0 +1,54 @@
...
+const elements = [
+  '#jform_execution_rules_cron_expression_minutes',
+  '#jform_execution_rules_cron_expression_hours',
+  '#jform_execution_rules_cron_expression_days_month',
+  '#jform_execution_rules_cron_expression_months',
+  '#jform_execution_rules_cron_expression_days_week'
+];
+
+function toggleRequired(process = 'add') {
+  if (!['add', 'remove'].contains(process)) {
```
n3t: Shouldn't be be includes instead of contains?

magnussinger: Yes I think you're right

Jan-at-tetronik:
This issue still persists even in version 5.2. The "Cron Expression (Advanced)" option remains virtually unusable that way.

HLeithner:
This pull request has been automatically rebased to 5.4-dev.

build/media_source/com_scheduler/js/admin-view-config-helper.es6.js
richard67:
```suggestion
  if (!['add', 'remove'].includes(process)) {
```
Add suggestion by @n3t .

richard67:
@astridx As the issue #37313 which this PR claims to solve was reported by you: Can you still reproduce the issue on Joomla 5.4? It seems that has been fixed meanwhile so this PR here would not be needed anymore.

Issue #37545, opened 2022-04-12
Iframe Menu Item -> Advanced tab Auto Add set to "No" is ignored


bayareajenn:
### Steps to reproduce the issue
1. Login to the backend of Joomla
2. Add a new Menu Item with the Menu Item Type Wrapper -> Iframe
3. Give it any Title - doesn't matter
3. Click the Advanced tab
4. Change the Auto Add field from Yes to No
5. Click the Details tab
6. Enter a URL without http or https into the URL field (doesn't matter what it is, let's say google.com)
7. Click Save
8. The URL now has http:// added even though the parameter has been set to "No" for Auto Add

### Expected result
That no http or https is added to the URL field when the Auto Add field is set to no.

### Actual result
http is added when it shouldn't be

### System information (as much as possible)
J 4.1.2
PHP 8.1

Quy:
Duplicate #33235?

chmst:
The parameter is for the frontend, it is not used in backend. 
The filter="url" in the url field writes a http:// as prefix.

chmst:
@Quy not a duplicate but similiar. I made a PR - but need more opinions if this has side effects.

chmst:
Closing as there is a PR. Please test https://github.com/joomla/joomla-cms/pull/37547

Issue #37544, opened 2022-04-12
Iframe Menu Item -> Scroll Bar Parameters tab missing Scroll Bars No/Yes/Auto Field


bayareajenn:
### Steps to reproduce the issue
1. Login to the backend of Joomla 4.1.2
2. Add a new menu item "Wrapper -> Iframe"
3. Click on Scroll Bar Parameters tab
There are two fields displayed; Width and Height. There should be a third field "Scroll Bars" with the options of No/Yes/Auto. The help screen shows this field.![screen shot 2022-04-12 at 17 16 54](https://issues.joomla.org/uploads/1/53e51d691e3290797d8588c11858016d.png)

### Expected result
To see and be able to use the Scroll Bars field.

### Actual result
There is no Scroll Bars field. 

### System information (as much as possible)
Joomla 4.1.2
PHP 8.1

richard67:
I think the parameter has been removed with this PR: #32913 . @bayareajenn Please check and if that attribute is really deprecated then close this issue here as expected behavior.

The help pages often are not up to date with the latest code changes so they cannot server as a reference.

bayareajenn:
Thanks @richard67 
Sadly, we use this in our custom extension in an integration. So I guess I'll have to come up with some work around which is unfortunate. In working on getting our extension ready for J4, I found this bug which is not a bug but rather someone thinking no one uses it. Bummer.

richard67:
@bayareajenn As said, the scrolling attribute is deprecated. See e.g. here: https://stackoverflow.com/questions/5048793/the-scrolling-attribute-on-the-iframe-element-is-obsolete-use-css-instead . Maybe you can achieve the same by adding a css class?

bayareajenn:
@richard67 I'm sure I'll figure something out. I can give a go to a CSS class. Thank you for helping me figure this out and for your time and service.

Issue #33235, opened 2021-04-22
[4.0] Menu item type wrapper setback.


frostmakk:
After merging of #32342 it is no longer possible to use relative paths in the URL field of the Iframe wrapper.
This is a major deterioration of the wrappers usefulness.

Nightly Build 4.0.0-beta8-dev-Development-Full_Package.zip Thursday, 22 April 2021 02:00:57 UTC

brianteeman:
What is your use case for wrapping a link on your own site inside itself?

frostmakk:
To show something not made with Joomla inside an Iframe in Joomla.

brianteeman:
So why not use the real url?

frostmakk:
That has to be changed when the site is moved to localhost or anywhere else?
That is not good coding practice.

reburg67:
In the Iframe Wrapper module it is still possible to set a relative URL without using `http://`.
But it is not possible in an Iframe Wrapper Menu Item. I see no logic in that.
Although I set the option "Auto Add" to false, `http://` is automatically added. Therefore I cannot use the same site on my production server and on the test environment. Please change this behaviour.

chmst:
Closing as there is a PR. Please test https://github.com/joomla/joomla-cms/pull/37547

Skrol29:
Local links (in iFrame and other) are useful when Joomla is part of a packaged solution.
So I confirm this was a problem for us too.

Glad you have fixed this problem !
Thank you,<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/33235">issues.joomla.org/tracker/joomla-cms/33235</a>.</sub>

PR #32342, opened 2021-02-07
[4.0] Validate URL for Iframe Wrapper menu item type

joomdonation:
Pull Request for Issue #32245 (part).

### Summary of Changes
This small PR change field type of the url field type to url, also add validate="url" to make sure a valid URL needs to be saved before the menu item can be saved.

### Testing Instructions
1. Create a menu item of Iframe Wrapper menu item type
2. Before patch: You can enter anything into URL
3. After patch: Only valid URL is accepted

PR #37547, opened 2022-04-12
Remove filter="url" for wrapper iframe

chmst:
Pull Request for Issue #37545 .

### Summary of Changes
Remove filter="url" for wrapper iframe url.

### Testing Instructions
see issue #37545

Url without http(s):// gets prefix http:// automatically

Url without prefix is possible.

Make sure that the url then is correct in frontend.


brianteeman:
I have tested this item :red_circle: unsuccessfully on 106b730ca42c25ba7851f6af2e8a568e6aa2eed7<br /><br />The autoadd functionality is now never available<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

chmst:
Removing also validate - enables relative urls as mentioned in https://github.com/joomla/joomla-cms/issues/33235

chmst:
The autoadd function works only in frontend, not in backend. You can see the code in components\com_wrapper\src\View\Wrapper\HtmlView.php. In my opinion this is wrong in the view but this is not in scope here

brianteeman:
I have tested this item :white_check_mark: successfully on 106b730ca42c25ba7851f6af2e8a568e6aa2eed7<br /><br />sorry - misunderstood<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

Quy:
This reverts issue #32245 and PR #32342.

bayareajenn:
@Quy  is correct. It reverts those other issue/PRs. I can never find this stuff so thank you for finding it.
And thank you @chmst for writing the PR. Whilst I'd love to have it back to the way it was, if it isn't going to, we can write it into our custom extension. 
To me there's no point is having an option of "No" if it doesn't actually do what it says. We use a shortcode to a different server in that field in our custom extension. In working on getting our J3 extension ready for J4 is when I found this "bug" which doesn't appear to be a bug but rather other people deciding no one uses it in a different way. Bummer.

So someone let me know if I should test and then the other PR would be reverted or if we close this and I deal with it a different way, k? 

Thanks for testing @brianteeman

chmst:
Thanks @Quy. If this enforced http//: is intended, then the the code in components\com_wrapper\src\View\Wrapper\HtmlView.php is wrong. I never use this and so have no opinion

bayareajenn:
I have tested this item :white_check_mark: successfully on 106b730ca42c25ba7851f6af2e8a568e6aa2eed7<br /><br />Works great and how it did before #32245 and PR #32342.

Let's see what happens now.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

zero-24:
hmm i dont think we should remove the validation here. Better we should create a filter that also allows relative URLs where the question should be asked first why should we iframe an internal URL in the first place?

And by the intial report the 'issue' is that the protocol has been added? why is that an issue isnt adding the protocol not something thats a good thing?

maybe I'm missing something here?

bayareajenn:
@zero-24 At my company, we aren't using an internal URL. We're using a shortcode off to a different server. Our current extension adds the https so in the menu item we kept it turned off. We can't go adding http(s) twice now, can we?! So two choices: the option is added back into Joomla or my company changes our extension to NOT have the http(s) to make up for the change and maybe not be able to use the shortcode anymore since I don't think we can have Joomla detect the "s" on a shortcode.

Simply said, it was something we used. I didn't know it had disappeared in J4. I thought it was a bug.

The fact that the setting/parameter Auto Add can be set to "No" but doesn't obey "No" should be addressed though. Either allow it to be turned off for reals, or remove the Auto Add parameter completely since it doesn't work. It adds http when it is set to No.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

zero-24:
> We're using a shortcode off to a different server. 

What is a shortcode in this context? In the past it was just a "you can add here anything" field now we make sure its an URL including the protocol.

frostmakk:
From the iframe wrapper menu's help screen:

image

The question is not why should we allow a relative path. The question is why did we remove a valid and documented option that people are using?

zero-24:
> The question is not why should we allow a relative path. The question is why did we remove a valid and documented option that people are using?

I still have the question why we need to use relative paths in the first place but when its documented that way an filter that allows relative paths should be created and we should not just remove the current url filter

zero-24:
In the best case there should just be an option within the current URL filter like "allowRelative=true" and in case its not set its not allowing relative URLs while this option will be set here ;)

frostmakk:
Why not simply switch the validation filter on/off with the existing Auto Add option button ?

zero-24:
> Why not simply switch the validation filter on/off with the existing Auto Add option button ?

That would mean the validation rule has to check the option of an specific backend component which is not a good idea and auto Add has nothing to do with the validation but is adding stuff the the when the user does not add it.

bayareajenn:
> What is a shortcode in this context? In the past it was just a "you can add here anything" field now we make sure its an URL including the protocol.

We have a lot of sites. Instead of having to put the server name in each site (different for every instance), we use a shortcode - imagine it's something like `_SERVER_` and it pulls from the parameters of the custom extension which does have the server name spelled out. It makes it easier for our people who build the sites (globally).

This is the second thing to be removed from the iframe menu item but the documentation says it exists https://github.com/joomla/joomla-cms/issues/37544.

In the case of this issue, I can work around it if I have to but the fact remains, there's an Auto Add option toggle that should allow me to turn it off and I can't. So is it a bug? Or is it just other people deciding what's good for me to put in the URL field? OR should the Auto Add parameter also be removed since it's a pointless toggle?

Pallieguy:
We have custom scripts/tools that are developed on a development environment before being moved to production, using a relative path allows us to copy and paste everything between our production and development environments instead of manually setting them.

While not a lot of work, it's unnecessary work and one more thing for someone to miss when trying to code/debug

RickR2H:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

zero-24:
@RickR2H please take off the RTC label as there is still an open discussion how to solve this issue.

brianteeman:
and please put the RLDQ tag back

RickR2H:
Back to pending. @bembelimen can you decide what to do?<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

bayareajenn:
@bembelimen and @zero-24 any updates on this one?

mpradel:
I have tested this item :white_check_mark: successfully on 106b730ca42c25ba7851f6af2e8a568e6aa2eed7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

HLeithner:

HLeithner:

brianteeman:
This has been waitinf ro a release lead to make a decision for over 1 year :(

TLWebdesign:
I've tested this and noticed that when no schema was added to the url it still added a single slash in front of the url in the generated html. When i looked at the dom it looked like this: src="/domain.nl" while the url field only contains domain.nl<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37547">issues.joomla.org/tracker/joomla-cms/37547</a>.</sub>

MacJoom:
> I've tested this and noticed that when no schema was added to the url it still added a single slash in front of the url in the generated html. When i looked at the dom it looked like this: src="/domain.nl" while the url field only contains domain.nl
> 
@TLWebdesign - thanks for testing - is this a problem within this PR or another issue?

PR #37565, opened 2022-04-15
Updating joomla-projects/joomla-browser

Hackwar:
What the title says. This should bring a few improvements for automated tests in the future.


richard67:
@Hackwar What's the reason to base it to 4.2-dev and not 4.1-dev?

Hackwar:
No special reason. It doesn't make a difference really if we do this in 4.1 or 4.2 and since 4.2 will be released soon, this would be perfectly fine for me.

laoneo:
As tests are passing, I'm merging this. Thanks!

PR #37583, opened 2022-04-19
[4.2] Indicate if category not published

brianteeman:
Articles that are in a category which is unpublished or trashed are not visible on the site but there is no way you can tell this from the article manager

### Summary of Changes
1. Adds an indicator if the category is unpublished or trashed. 
2. Changes the published icon to a disabled icon and adds a tooltip to the icon

### Testing Instructions
Create a catgory called test and create 4 featured articles in that category. 
Unpublish or Trash the category

Check the list of articles and the list of featured articles

No indication that the article can not be displayed

Text as shown in the image below

image


rdeutz:
Looks like a good idea, I think this is something we should add and can be added to the next minor version.

rdeutz:
one addition: maybe better to solve this in the PublishedButton class. Then you only have to fix it one time and not in all views. Could be something like "Published but category unpublished or trashed"

brianteeman:
@rdeutz That was my first thought but then it gets complicated by workflows

brianteeman:
@rdeutz took another look but I dont see how to do it in the class as it is not aware of the category

brianteeman:
@rdeutz do you have any pointers as to how to solve it in the PublishedButton. If not I will continue with this approach.

rdeutz:
I looked at the class and (beside Workflow, I don't see how this effects the buttom but maybe I missed something) it should be doable. I will check over the weekend and let you know.

brianteeman:
@rdeutz thanks

rdeutz:
This should do the trick: https://github.com/rdeutz/joomla-cms/commit/847c041c76ff8c913d03fe32c3a310dff11be46a

brianteeman:
> one addition: maybe better to solve this in the PublishedButton class. Then you only have to fix it one time and not in all views. Could be something like "Published but category unpublished or trashed"

as even your code requires a chang in the view I decided to combin the two

BertaOctech:
I have tested this item :white_check_mark: successfully on d3bd8857dd1d25175afc0e85e25590c1d4054b81<br /><br />The patch workes exactly as described<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37583">issues.joomla.org/tracker/joomla-cms/37583</a>.</sub>

pabloarias:
I have tested this item :white_check_mark: successfully on d3bd8857dd1d25175afc0e85e25590c1d4054b81<br /><br />Tested in Joomla 4.1.3-dev with PHP 8.1.

Expected behaviour when category is unpublished or trashed.

It works in Featured Articles list too.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37583">issues.joomla.org/tracker/joomla-cms/37583</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37583">issues.joomla.org/tracker/joomla-cms/37583</a>.</sub>

bembelimen:
I really like this PR, but it will not make it in 4.1, but for sure in 4.2!

Thanks for the improvement.

bembelimen:
JFYI: the "too much files" will be fixed on Tuesday, then there will be the next upmerge from 4.1 (which is already in your branch) after the relese of J! 4.2 Alpha 2. Then we're back to normal (no additional human tests needed).

brianteeman:
No problem

roland-d:
Thanks everybody

PR #37587, opened 2022-04-20
[4.1] Remove php version to speed up the test

rdeutz:
### Summary of Changes
Remove multiple PHP versions from the appveyor testing. 

We only need to make sure we don't trash something that it realated to Windows and this has not su mch to do with the PHP Version so testing only one PHP Version should do the job.


wilsonge (approved):
Fine. Although we might be better using PHP 8.0 so the tests run that bit faster?

rdeutz:
Can be merged on review @Hackwar @bembelimen

PR #37590, opened 2022-04-20
[4.x] Article Manager RTL correction

brianteeman:
This PR removes a conditional block from the article manager and the featured article manager that was intended to ensure that the display of the category and nested category was correct when the admin is using an rtl language and the category names may be a mix of LTR and RTL.

Not really sure why but
1. this code does not work correctly. the order of the categories is wrong and the « character is wrong
2. there is no need for this alternate block of code for RTL

### Before
image

### After
image


Quy:
Also, `$catSeparator` should not be in the loop so it is only executed once.

brianteeman:
image

Weird - I can't replicate your issue

Quy:
Please install Testing Sample Data. Check the 2 articles referenced in the above screenshot. It is fine in Persian before the PR.

brianteeman:
Replicated th problem - it is coming from when there are more than 2 categories in the tree and it is truncated. Will look into a fix later tonight

PR #37592, opened 2022-04-21
[4.2] Load the actionlog params from model instead of the helper

laoneo:
### Summary of Changes
Move the logic to fetch the log content type params of the action log component to a model. Models should do the database calls as they have access to it. Like that it is possible to inject the database object and it is not necessary to do a deprecated `Factory::getDbo()` call. This helps to write proper unit tests which improves the sustainability. Having the code tested with unit tests ensures that errors in further changes will be detected and potential BC breaks can be eliminated. This is only a small step but we need to head into that direction when we want to have a system where we can write easy system tests without a bunch of preparation setup.

Additionally it makes a small modification to the extension namespace mapper so it can be loaded during the bootstrap process of the unit test suite, similar to the libraries.

### Testing Instructions
- Make sure the Joomla action log plugin is loaded
- Save an article

An entry is written to the user action log that an article is saved.

An entry is written to the user action log that an article is saved.


laoneo:
@brianteeman this is my idea how it should be when interacting with a component. The helpers should be replaced with models.

libraries/namespacemap.php
```diff
@@ -142,7 +141,7 @@ protected function writeNamespaceFile($elements)
 		 */
 		$error_reporting = error_reporting(0);
 
-		if (!File::write($this->file, implode("\n", $content)))
+		if (!file_put_contents($this->file, implode("\n", $content)))
```
Hackwar: I fully agree with this change, however I think I would feel better if this were in its own PR. I feel kinda itchy when we lump this into such an unrelated change.

plugins/actionlog/joomla/joomla.php
```diff
@@ -98,7 +98,8 @@ public function onContentAfterSave($context, $article, $isNew): void
 			$context = $this->contextAliases[$context];
 		}
 
-		$params = ActionlogsHelper::getLogContentTypeParams($context);
+		$params = $this->app->bootComponent('actionlogs')->getMVCFactory()
+			->createModel('ActionlogConfig', 'Administrator')->getLogContentTypeParams($context);
```
Hackwar: Did we ever do a test on how much of a performance loss this is? The previous code is just one function call, this one here would be 4 in this line alone, let alone all the indirect calls we have inside of those methods.

laoneo: If you talk about performance, then the whole code must be moved to the plugin as it is used only there. Additionally there should be an index on the type_alias field. Doing a couple of more PHP function calls, doesn't decrease performance, the bottlenecks are on other areas.

administrator/components/com_actionlogs/src/Model/ActionlogConfigModel.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Model to interact with the action log configuration.
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class ActionlogConfigModel extends BaseDatabaseModel
+{
+	/**
+	 * Get parameters to be
```
brianteeman: to be what?

brianteeman:
> @brianteeman this is my idea how it should be when interacting with a component. The helpers should be replaced with models.

Without any explanation of why :(

nibra:
Looks like a step towards better architecture to me. However, I would get the model in the constructor so that the (relatively) complex call only occurs once.

nibra:
> Without any explanation of why :(

That's not necessarily required, as there are a few obvious reasons.
a) Only models should ever interact with the persistence layer
b) Whenever you're tempted to name a class 'Helper', you're doing wrong

libraries/namespacemap.php
laoneo: Here we go https://github.com/joomla/joomla-cms/pull/37595. Still leaving this in here till 37595 is merged.

laoneo:
> Looks like a step towards better architecture to me. However, I would get the model in the constructor so that the (relatively) complex call only occurs once.

The drawback is that the model is then always loaded even when no function is called.

brianteeman:
> That's not necessarily required, as there are a few obvious reasons.

obvious to you but not everyone. 

Remember that there is an entire ecosphere of extension deevlopers whose only knowledge of php etc is from the joomla core. Whenever there is an architecture change it should be fully documented and explained as a reference for them AND for future contributors to the core.

For example if naming a class "helper" is doing it wrong why are there any instances of this? If it is so obviously wrong then that code should have never been meerged. Without any explanation then you are keeping secret the reason why and only an inner clique will ever be able to contribute and no one else will bee able to learn

laoneo:
I'v updated the description, is it more clear now?

libraries/namespacemap.php
Hackwar: Thanks

nibra:
> The drawback is that the model is then always loaded even when no function is called.

Then the whole call should be moved to a private method `getConfigParams()` or the like.

nibra:
> Remember that there is an entire ecosphere of extension deevlopers whose only knowledge of php etc is from the joomla core. Whenever there is an architecture change it should be fully documented and explained as a reference for them AND for future contributors to the core.

Yes, agree. Actually that's one of the reasons for creating the RFC process (which needs to get more attention, but that's another story)

nibra:
> if naming a class "helper" is doing it wrong why are there any instances of this? If it is so obviously wrong then that code should have never been meerged.

Not everyone gives clean architecture the love it deserves. Helper' would be a reasonable name if one could say what one purpose a helper serves - after all, each class should serve only one purpose. But if it does, a more descriptive name for that class can always be found. Therefore, the name 'Helper' is a so-called code-smell.

plugins/actionlog/joomla/joomla.php
```diff
@@ -1175,4 +1171,25 @@ public function onJoomlaAfterUpdate($oldVersion = null)
 		);
 		$this->addLog(array($message), 'PLG_ACTIONLOG_JOOMLA_USER_UPDATE', $context, $user->id);
 	}
+
+	/**
+	 * Does return the action log params for the given context.
```
brianteeman:
```suggestion
	 * Returns the action log params for the given context.
```

plugins/actionlog/joomla/joomla.php
```diff
@@ -1175,4 +1171,25 @@ public function onJoomlaAfterUpdate($oldVersion = null)
...
+	 *
+	 * @param   string  $context  The context of the content
```
brianteeman:
```suggestion
	 * @param   string  $context  The context of the action log
```

Not 100% sure but I think this is what you meant

libraries/namespacemap.php
```diff
@@ -8,9 +8,9 @@
 
 defined('_JEXEC') or die;
 
-use Joomla\CMS\Filesystem\File;
-use Joomla\CMS\Filesystem\Folder;
 use Joomla\CMS\Log\Log;
+use Joomla\Filesystem\File;
```
HLeithner: Same as in the other PR you we have exceptions now.

laoneo: This here was just a copy of the original pr. Got reverted with the last update.

tests/Unit/Components/Actionlogs/ActionlogConfigModelTest.php
```diff
@@ -0,0 +1,62 @@
...
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\Tests\Unit\Components\Actionlogs;
+
+use Joomla\CMS\MVC\Factory\MVCFactoryInterface;
+use Joomla\Component\Actionlogs\Administrator\Model\ActionlogConfigModel;
+use Joomla\Database\DatabaseInterface;
+use Joomla\Tests\Unit\UnitTestCase;
+use stdClass;
```
HLeithner: We don't import stdclass

roland-d: Why not?

laoneo: I really don't care

roland-d:
Thank you

PR #37595, opened 2022-04-21
[4.2] Namespacemap dependency cleanup

laoneo:
Pull Request for comment https://github.com/joomla/joomla-cms/pull/37592#discussion_r854835453.

### Summary of Changes
~Removes~ Replaces the CMS/Filesystem dependencies in the classmap generator.

### Testing Instructions
- Remove the file /administrator/cache/autoload_psr4.php
- Open the back end

The removed file is regenerated and the site loads as expected.

The removed file is regenerated and the site loads as expected.


laoneo:
@Fedik here you have a filesystem pr.

wilsonge (approved):

libraries/namespacemap.php
```diff
@@ -8,8 +8,7 @@
 
 defined('_JEXEC') or die;
 
-use Joomla\CMS\Filesystem\File;
-use Joomla\CMS\Filesystem\Folder;
+use Joomla\Filesystem\Folder;
 use Joomla\CMS\Log\Log;
```
Quy:
```suggestion
use Joomla\CMS\Log\Log;
use Joomla\Filesystem\Folder;
```

Quy:
I have tested this item :white_check_mark: successfully on 7d8ecc3cb1b795d9f6d6acf8571f9019bcdfd8f9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37595">issues.joomla.org/tracker/joomla-cms/37595</a>.</sub>

Shubhamverma2796:
I have tested this item :white_check_mark: successfully on 7d8ecc3cb1b795d9f6d6acf8571f9019bcdfd8f9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37595">issues.joomla.org/tracker/joomla-cms/37595</a>.</sub>

richard67:
RTC <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37595">issues.joomla.org/tracker/joomla-cms/37595</a>.</sub>

HLeithner:
Please don't blindly replace File::write with file_put_contents thanks. There are multiple reasons to keep it as it is.

richard67:
@HLeithner It seems there are different opinions because @wilsonge approved it. Am setting the RLDQ label to flag it for further discussion/decision.

laoneo:
@HLeithner please elaborate why this should not work. It was not a blind replacement. I'v even tested it when the cache folder doesn't exist, it loads the map into the ram thanks to #33263. The site crashes afterwards because the cachecontroller can't find the folder. So we are pretty safe here to move to native PHP.

HLeithner:
FIle::write does much more then just write a file to disk. it creates the folder (ok should exists) but most important it invalidates the opcache.

Never the less on the roadmap on j5 we have the point to use asynchronous functions where it make sense and where it's possible. Don't know yet if it's useful at this point but at least should be considered.

brianteeman:
this is a great example of why low level changes need much better test instructions and descriptions

richard67:
Back to pending due to reasons stated above. I should have remembered the opcache thing when setting RTC.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37595">issues.joomla.org/tracker/joomla-cms/37595</a>.</sub>

laoneo:
> this is a great example of why low level changes need much better test instructions and descriptions

And this is a great example how to not participate in an open source project as it doesn't bring any value and pisses of contributors. Instead of propose a suggestion what needs to be changed

> but most important it invalidates the opcache

My opcache settings are as follows, so what do I need to change to make the pr fail when opcache kicks in as it worked on my end?
image

richard67:
> My opcache settings are as follows, so what do I need to change to make the pr fail when opcache kicks in as it worked on my end?

@laoneo You could try with the settings mentioned in the description of PR #36878 :

> Use a server with OPcache enabled and the settings `opcache.validate_timestamps=1` and `opcache.revalidate_freq=300`

Your screenshot doesn't show the values for these settings.

laoneo:
Now I can reproduce the opcache issue. Reverted to use the framework package write function, so we don't need the full CSM stack.

So I guess we can do another round of testing as it does the same as before, just using the framework package.

richard67:
~~Now the only remaining change is the alpha ordering of the use statements. I think this can be tested by review.~~

~~@laoneo Could you update the description of the PR accordingly? Thanks in advance.~~

Silly me ... was confused by the almost same use statements. But I see it changes from CMS to framework package.

libraries/namespacemap.php
```diff
@@ -8,9 +8,9 @@
 
 defined('_JEXEC') or die;
 
-use Joomla\CMS\Filesystem\File;
-use Joomla\CMS\Filesystem\Folder;
 use Joomla\CMS\Log\Log;
+use Joomla\Filesystem\File;
```
HLeithner, line 12:
```suggestion
```

It's not used in this file

laoneo: It is, as I reverted the code file_put_contents with File:write from the filesystem package.

laoneo: All this pr does now is to use the classes from the filesystem framework package and not core.

HLeithner: it wasn't used before, at least I can't find any usage in this file

laoneo: https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/namespacemap.php#L145

HLeithner: Good then I'm blind sorry core the noise.

HLeithner:
The framework throws exceptions which was a return false before. this have to be caught. Both File::write and Folder:folders need to be in a try catch block now.

laoneo:
Catching now the exceptions.

HLeithner:
thanks

Issue #37575, opened 2022-04-19
Does not specify the error


pearlpriyanka:
### Steps to reproduce the issue
Try to upload an extension in the "upload package file" link will cause an error when you upload a folder in this link, But the error caused is not specified it just shows a red colored error bar without any content.

### Expected result
The error should be specified in the red colored error bar saying "Package FIles not found" when the user tries to upload any other file.
![Screenshot (26)](https://user-images.githubusercontent.com/62989282/163921717-426ed016-bb70-4023-8d03-b6fa6a8ddc04.png)

### Actual result
But the actual result is it does not specify any error, it just shows a red bar with an error symbol.
![Screenshot (25)](https://user-images.githubusercontent.com/62989282/163921751-622e26a1-2ddb-4a72-a6bb-d7417fd4eda1.png)

### Additional comments
The user should upload the extension under the "install from folder" link to avoid this issue but the error should be specified for new users to be able to work efficiently.

brianteeman:
Confirmed - empty error message

brianteeman:
I believe it is duee to a javascript error

POST http://localhost/joomla-cms/administrator/index.php?option=com_installer&task=install.ajax_upload net::ERR_ACCESS_DENIED
Joomla.request @ core.js?e718547c598935f1b696b04a8a218a74:730
(anonymous) @ packageinstaller.js?e718547c598935f1b696b04a8a218a74:151

brianteeman:
please test #38227

alikon:
closing as we have a pr to test #38277

richard67:
> closing as we have a pr to test #38277

No, it's #38227 . Please test.

PR #37601, opened 2022-04-22
[4.2] Improving javascript validation of package installer

vorayash:
Pull Request for Issue #37575.

### Summary of Changes
Added validation for non zip files in 'build\media_source\plg_installer_packageinstaller\js\packageinstaller.es6.js'

### Testing Instructions
Try to upload an extension in the "upload package file".

The error was showing just red colored error bar without any content.
image

image


laoneo:
I would more ask why is there no message in the alert. I think it would be better to serve a proper message from the server why the installation failed. Did you figure out where this empty error comes from?

build/media_source/plg_installer_packageinstaller/js/packageinstaller.es6.js
```diff
@@ -125,6 +125,13 @@ Joomla = window.Joomla || {};
         return;
       }
 
+      if (file.name.split('.').pop() !== "zip") {
```
dgrammatiko: You shouldn't really rely on the extension (for security reasons) on any file check in the web. A better approach is to check the mime type, something like: `file.type !== 'application/zip'`. In sort the extension should NEVER be used to sniff the type of file

dgrammatiko (changes requested):

vorayash:
> I would more ask why is there no message in the alert. I think it would be better to serve a proper message from the server why the installation failed. Did you figure out where this empty error comes from?

When we send folder as data, xhr readystate changes to 4 and status to 0. It means request is failed. When request fails, value of xhr.statusText changes into empty string(""). 

image

xhr.statusText is a string so it will satisfy below-pointed condition and empty message will be shown.

image

I tried xhr request on my local machine and found that when we send folder as data, request fails. So I think sending folder is not allowed in xmlhttprequest.

build/media_source/plg_installer_packageinstaller/js/packageinstaller.es6.js
vorayash: 👍

laoneo:
So it means the request never arrives on the server?

vorayash:
I think so. Can you please confirm that?

Fedik:
It should be fixed the empty message, instead. 
Server already doing validation. as @laoneo pointed.

PR #17328, opened 2017-07-28
[4.0] Create a service registry for JHtml

mbabker:
### Summary of Changes

The manner in which services are integrated into JHtml are very sporadic and not the friendliest thing.  The `register` and `unregister` API maps specific keys to single callbacks and integrating services from components requires registering filesystem paths to scan for specific files.  We can do better than this.

This PR does two things.

First, it introduces a new service registry.  The registry allows developers to register a service key to either a PHP class name (static method calls, essentially what the system does now) or a class object (removing the requirement to either manually register callbacks through `JHtml::register()` or to use purely static methods).

Second, this PR deprecates the existing logic for registering things.  Manual lookup paths, registering (and unregistering) callbacks, and support for a three-piece key will be removed with Joomla 5.0 if this is accepted in favor of registering everything to the service registry.  The three-piece key is what allows developers to replace the `JHtml` class prefix with a prefix of their choosing, with this registry it really won't matter what they name their classes as the key will no longer have to directly map to a class name.  Path lookups are being removed as by the time we get to 5.0 extension architecture should really be at a point where it won't be needed.

The installation application's `JHtml` helper is updated to demonstrate this new methodology.

### Testing Instructions

Register a custom service to the registry by calling `JHtml::getServiceRegistry()->register($key, $handler);` where `$key` is the service key and `$handler` is either a PHP class name or a class object.  Once registered, calls to `JHtml::_($key . '.method');` should use your service.

### Documentation Changes Required

The deprecation and new manner of working should be documented.

PR #37611, opened 2022-04-22
[4.2] Inject the database into the finder classes

laoneo:
### Summary of Changes
Ongoing effort to inject the database into components and get rid of the deprecated Factory::getDbo(); calls. This pr injects the database in the finder component instead of using the factory to get it from.

### Testing Instructions
- Make sure the smart search content plugin is enabled
- Edit an article and add a description text with the word hello
- Save the article
- Go to the url /administrator/index.php?option=com_finder&view=index and hit the Index button
- Create a search module for the front
- Search for "hello"

All works.

All works.


RickR2H:
I have tested this item :white_check_mark: successfully on 3911c153ba163b49c8b9d4b0d01985b609ed16a9<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37611">issues.joomla.org/tracker/joomla-cms/37611</a>.</sub>

pabloarias:
Sorry, not able to test it.

In Joomla 4.2.0-alpha2-dev with PHP 8.0 and 8.1: Error before apply the path, when trying to create Smart Search module: An error has occurred. 0 Attempt to assign property "params" on null.

In Joomla 4.1.3-dev with PHP 8.1: if I apply the path I obtain an error in frontend: ClassNotFoundError. Details:

#### 

Symfony\Component\ErrorHandler\Error\ClassNotFoundError:
Attempted to load trait "DatabaseAwareTrait" from namespace "Joomla\Database".
Did you forget a "use" statement for another namespace?

at /shared/httpd/Joomla_4.1.3-dev/htdocs/administrator/components/com_finder/src/Service/HTML/Finder.php:26

#####

Thank you and best regards<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37611">issues.joomla.org/tracker/joomla-cms/37611</a>.</sub>

RickR2H:
@pabloarias Make sure you are on the 4.2 branch!

administrator/components/com_finder/src/Indexer/Adapter.php
```diff
@@ -135,9 +135,6 @@ abstract class Adapter extends CMSPlugin
 	 */
 	public function __construct(&$subject, $config)
 	{
-		// Get the database object.
-		$this->db = Factory::getDbo();
-
```
Hackwar, line 140: Where is $this->db set? This object is used in most finder plugins. Looks to me like a b/c break right now.

administrator/components/com_finder/src/Service/HTML/Filter.php
```diff
@@ -39,7 +42,7 @@ class Filter
 	 */
 	public function slider($options = array())
 	{
-		$db     = Factory::getDbo();
+		$db     = $this->getDatabase();
```
Hackwar, line 45: this whole class should actually be marked as abstract and the methods as static. These are used with HtmlHelper and I don't see where an instance of this class would be generated, so that we would have an object context and setDatabase() could have been set.

administrator/components/com_finder/src/Service/HTML/Finder.php
```diff
@@ -34,7 +37,7 @@ class Finder
 	public function typeslist()
 	{
 		// Load the finder types.
-		$db = Factory::getDbo();
+		$db    = $this->getDatabase();
```
Hackwar, line 40: This has the same issue as the other class. This class should actually be abstract and static and we wouldn't have an object context.

modules/mod_finder/src/Helper/FinderHelper.php
```diff
@@ -86,6 +87,9 @@ public static function getQuery($params)
 		$options['filters'] = ArrayHelper::toInteger($options['filters']);
 
 		// Instantiate a query object.
-		return new Query($options);
+		$query = new Query($options);
+		$query->setDatabase(Factory::getContainer()->get(DatabaseInterface::class));
```
Hackwar: Why aren't you adding the db object in line 90?

administrator/components/com_finder/src/Indexer/Adapter.php
laoneo: The database is automatically set in the plugin constructor here https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Plugin/CMSPlugin.php#L132.

modules/mod_finder/src/Helper/FinderHelper.php
laoneo: Because I used the aware trait.

administrator/components/com_finder/src/Service/HTML/Finder.php
laoneo: We want to go away from static facade classes. Better to inject the dependency.

modules/mod_finder/src/Helper/FinderHelper.php
Hackwar: Yes, but you added the db object to the contructor of the Query class and I'm wondering why you aren't using that then. That means that you still execute the Factory::getDbo() in the constructor of the Query right now.

administrator/components/com_finder/src/Service/HTML/Finder.php
Hackwar: I understand that, but that doesn't change the fact that this is not in an object context. All methods in this (and the other) class are used exclusively in a static context. The Finder class is never instantiated. Especially not where it is used in com_finder.

Hackwar: To clarify, I'm not against the idea behind this, but this would require A LOT more changes than just this. We are talking basically about a rewrite of this part of com_finder.

modules/mod_finder/src/Helper/FinderHelper.php
laoneo: Yes, sure, you are right, changed it

administrator/components/com_finder/src/Service/HTML/Finder.php
laoneo: Ah wait, this HTML services should not be abstract and static. This is correct like that. Was a change for Joomla 4.

administrator/components/com_finder/src/Service/HTML/Filter.php
laoneo: See my comment below. This is correct like that.

administrator/components/com_finder/src/Service/HTML/Finder.php
laoneo: https://github.com/joomla/joomla-cms/pull/17328

administrator/components/com_finder/src/Indexer/Indexer.php
```diff
@@ -112,13 +113,19 @@ class Indexer
...
 	{
-		$this->db = Factory::getDbo();
+		if ($db === null)
+		{
+			@trigger_error(sprintf('Database will be mandatory in 5.0.'), E_USER_DEPRECATED);
+			$db = Factory::getContainer()->get(DatabaseInterface::class);
+		}
 
-		$db = $this->db;
+		$this->db = $db;
```
roland-d, line 128: @laoneo Why not use `setDatabase($db)` here like you do in the `administrator/components/com_finder/src/Indexer/Query.php` file?

laoneo: Because of BC

roland-d:
Thanks everybody.

PR #37622, opened 2022-04-23
[4.2] Inject the database into console commands

laoneo:
### Summary of Changes
Ongoing effort to inject the database into components and get rid of the deprecated Factory::getDbo(); calls.  This pr injects the database into the console commands.

### Testing Instructions
Run the following commands on the console:
- `php cli/joomla user:add`
- `php cli/joomla user:addtogroup`
- `php cli/joomla user:delete`
- `php cli/joomla user:list`
- `php cli/joomla user:removefromgroup`
- `php cli/joomla extension:remove 0`
- `php cli/joomla extension:list`
- `php cli/joomla extension:discover`
- `php cli/joomla extension:discover:install`

All commands do work except the extension command. It throws an error that the extension with id 0 is not found.

All commands do work except the extension command. It throws an error that the extension with id 0 is not found.


roland-d:
Thank you

PR #37624, opened 2022-04-23
[4.2] The deprecated db functions do have to be the same as the trait

laoneo:
Pull Request for pr #37095.

### Summary of Changes
The function in the deprecated `Joomla\CMS\MVC\Model\DatabaseAwareTrait` must have the same signature as the functions in the `BaseDatabaseModel`. Additionally it ensures, when a model is using that trait and extending the `BaseDatabaseModel` that the database is available as the functions in the trait are overloading the ones from the class. Actually this is not needed as the `BaseDatabaseModel` is using in 4.1 the MVC trait and in 4.2 the DB package trait.

### Testing Instructions
Install the free version of akeeba backup and open the control panel of it in the back end.

It crashes with the following error:
_Compile Error: Declaration of Joomla\CMS\MVC\Model\DatabaseAwareTrait::getDbo() must be compatible with Joomla\CMS\MVC\Model\BaseDatabaseModel::getDbo(): Joomla\Database\DatabaseInterface_

It works normally.

### DIFF
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -96,6 +96,9 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 		{
 			@trigger_error(sprintf('Database is not available in constructor in 5.0.'), E_USER_DEPRECATED);
 			$this->setDatabase($db);
+
+			// Is needed, when models use the deprecated MVC DatabaseAwareTrait, as the trait is overriding the local functions
+			$this->setDbo($db);
 		}
 
 		// Set the default view search path
@@ -364,7 +367,7 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 	 *
 	 * @deprecated  5.0 Use getDatabase() instead
 	 */
-	public function getDbo(): DatabaseInterface
+	public function getDbo()
 	{
 		try
 		{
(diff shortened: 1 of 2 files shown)


laoneo:
@nikosdion any reason you explicitly load the MVC DatabaseAwareTrait in the control panel and update model?

nikosdion:
@laoneo The use of the DatabaseAwareTrait in the ControlpanelModel is a leftover from an older version of the code. That model had started as extending from BaseModel and implementing the DatabaseAwareTrait. It was converted to extend from BaseDatabaseModel (even though it really doesn't _quite_ fit into what BaseDatabaseModel is supposed to do) and I forgot to remove the usage of the trait.

The UpdatesModel does not use the trait, though, it simply extends from BaseDatabaseModel.

Perhaps you meant the UpgradeModel which extends from BaseModel and uses DatabaseAwareTrait? This model is loaded from the installation script in the postflight event. At this point I may not be able to pass an MVCFactoryInterface (if it's a clean installation, not an update) but I definitely need access to Joomla's database. I could either pepper my model with Factory::getContainer()->get('DatabaseDriver') or use the DatabaseAwareTrait and set the DB object once. I chose the latter because it reduces the number of places I will have to change something when inevitably a future version of Joomla changes the way we access the DB driver object. Sure, it took 15 years to go from JFactory::getDbo() to Factory::getContainer()->get('DatabaseDriver') but here I am, having had to do that change in hundreds of places across dozens of extensions 😉 Then again, the UpgradeModel doesn't match your PR so I am not sure you really meant that model.

laoneo:
Thanks for the explanation.

roland-d:
Thanks guys.

PR #37625, opened 2022-04-23
[4.2] Inject the database into the installer

laoneo:
### Summary of Changes
Standardizes the database access in the installer library. This is a preparation to load the installer through a service provider.

### Testing Instructions
- Install Joomla
- Discover extension
- Install any kind of extension, for example the [free version of DPCalendar](https://joomla.digital-peak.com/download/dpcalendar) which contains a package, component, modules and plugins
- Uninstall these extensions

All is working as before.

All is working as before.


installation/src/Application/InstallationApplication.php
```diff
@@ -311,11 +311,11 @@ private function executeController()
 			list($controllerName, $task) = explode('.', $task, 2);
 		}
 
-		// Compile the class name
-		$class = 'Joomla\\CMS\\Installation\\Controller\\' . ucfirst($controllerName) . 'Controller';
+		$factory = new MVCFactory('Joomla\\CMS', $this);
```
roland-d: @laoneo The `MVCFactory` class only takes 1 argument.

roland-d (dismissed):

installation/src/Application/InstallationApplication.php
laoneo: good catch

roland-d:
Thank you

PR #37628, opened 2022-04-23
[4.3] Inject the database into workflow

laoneo:
### Summary of Changes
Ongoing effort to inject the database into components and get rid of the deprecated Factory::getDbo(); calls. As workflow offers already the possibility to inject the database, this pr changes the usages where a workflow is instantiated to inject the dependencies. Additionally it deprecates the usage of null dependencies and removes some dead code.

### Testing Instructions
- Enable workflow in the articles manager
- Open the article form

All works.

All works.


libraries/src/MVC/Model/WorkflowBehaviorTrait.php
```diff
@@ -77,7 +78,17 @@ public function setUpWorkflow($extension)
...
+		{
+			$db = $this->getDatabase();
+		}
+		else
+		{
+			@trigger_error('In Joomla 5 will the getDatabase function be mandatory.', E_USER_DEPRECATED);
+			$db = Factory::getContainer()->get(DatabaseDriver::class);
+		}
+
+		$this->workflow = new Workflow($extension, Factory::getApplication(), $db);
```
roland-d: @laoneo PhpStorm gives me the warning `Expected parameter of type '\Joomla\CMS\Application\CMSApplication|void', '\Joomla\CMS\Application\CMSApplicationInterface' provided ` Is it something that could potentially cause a problem?

roland-d (changes requested):

libraries/src/MVC/Model/WorkflowBehaviorTrait.php
laoneo: Not really, we try to do as much as possible with the interface. So this should be fine as the CMS app is implementing the interface.

roland-d (approved):

laoneo:
Doc article is here https://github.com/joomla/Manual/pull/79

obuisard:
Thank you Allon @laoneo !

PR #37670, opened 2022-04-26
[4.1] Layout for Codemirror

Kostelano:
Pull Request for Issue # .

### Summary of Changes
Plugins, Editor - Codemirror. Perhaps the only parameter where we still used checkboxes.

**Before / After**

Screenshot_1


brianteeman:
My 2c is that in this use case the checkboxes are better. We really only use the select when the list is long or it is dynamically created

Issue #37663, opened 2022-04-26
[4.x] Testing Sample Data plugin should be removed


brianteeman:
I know it is not shipped with any release but people do use it for testing.

The problem is that as well as being outdated it is also broken in many ways which means that it cant really be used for testing. I suspect assets tables but can't be certain without spending a lot of time etc

Examples.
Menu items cannot be copied to another menu. The batch copy appears to be successful but the menu items are not displayed
The deep items menu can never be expanded.

I propose that it is removed until someone has the time to build a completely new set. Testing with broken data is worse than not testing

brianteeman:
I dont mind writing a new testing set but it will take a while and in the meantime if the plugin is not removed people will waste time on bugs that dont exist (as I have done for the last 4 hours)

chmst:
Agree with removing the plugin and all the old images we keep for it in the image folder. 

Writing a new set of data is a huge work - I tried but gave up. If it is rewritten we should separate plugin and data (also for blog sample). Maybe this way: 
https://github.com/dgrammatiko/sloth-pkg/tree/main/plg_sampledata could be <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37663">issues.joomla.org/tracker/joomla-cms/37663</a>.</sub>

PR #37674, opened 2022-04-26
[4.1] fix parks image

N6REJ:
Pull Request for Issue # .

### Summary of Changes

Parks image doesn't display properly due to improper relative path.

### Testing Instructions
Install sample testing data.
Click on "parks" on the menu at the top
Notice that the image does not display.
Verify the url for the image is "http://j4.local/index.php/images/sampledata/parks/banner_cradle.jpg" ( your uri will be different of course )
view that the image displays.
verify url for the image is "http://j4.local/images/sampledata/parks/banner_cradle.jpg"

image

image

None


brianteeman:
see #37663

Quy:
My local testing is in a subdirectory `http://localhost/joomla-cms-4.1-dev/` and not in the root so this PR will not be a correct fix.
Enable `Use URL Rewriting` and it will not be an issue.

brianteeman:
I have tested this item :red_circle: unsuccessfully on cccd072351a21f91f6021b62cbd035f9637131da<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37674">issues.joomla.org/tracker/joomla-cms/37674</a>.</sub>

N6REJ:
> My local testing is in a subdirectory `http://localhost/joomla-cms-4.1-dev/` and not in the root so this PR will not be a correct fix. Enable `Use URL Rewriting` and it will not be an issue.

Mine is a sub as well...
www/j4/.....

We had this exact same issue several years ago when the router was changed.

N6REJ:
> I have tested this item 🔴 unsuccessfully on [cccd072](https://github.com/joomla/joomla-cms/commit/cccd072351a21f91f6021b62cbd035f9637131da)
> 
> This comment was created with the [J!Tracker Application](https://github.com/joomla/jissues) at [issues.joomla.org/tracker/joomla-cms/37674](https://issues.joomla.org/tracker/joomla-cms/37674).

How was it unsuccessful, please show the failure.  Just because you don't like the fix doesn't mean it doesn't work.

brianteeman:
Before PR
image

image

After PR
image

image

brianteeman:
this really should be closed

N6REJ:
No, it shouldn't.  
The "default" dataset returns 
image
which is incorrect.
"http://j4.local/index.php/images/sampledata/parks/banner_cradle.jpg"
image
is off yet is clearly being left on even with .htaccess and url rewriting going on.

IF infact we do remove the testing sample dataset then that is a different issue.

brianteeman:
1. Add suffix to URL is nothing to do with this
2. I already showed that this PR does not work

N6REJ:
your test doesn't even make sense... if the image showed pre-fix why would I change it?  and how would adding a "/" change the url so dramatically.  I'll retest yet again with a clean ( again ) install.

N6REJ:
sample "test" data option has been removed from 4.1-x  So I'm closing this

brianteeman:
sample "test" data has NEVER EVER been part of any joomla release

PR #37673, opened 2022-04-26
Refactored WebAuthn with Windows Hello support

nikosdion:
### Summary of Changes

* Refactored as a native Joomla 4 plugin, using a service provider and implementing SubscriberInterface 😎  
* Removed the ugly `Joomla` helper class, replacing it with native code 💪🏽 
* Now using the (previously undocumented) WebAuthn library's `WebAuth\Server` object. This adds Windows Hello support **without** having to update to a new major version of the third party WebAuthn library 🥳 
* We no longer pass URLs as data arguments; we can figure them out using `Joomla.getOptions` 🪄 
* Fixed the breakage introduced in https://github.com/joomla/joomla-cms/pull/37464 🩹 

### Testing Instructions

Please remember to run `npm ci` after applying the PR — the JavaScript has changed.

Please remember to use HTTPS with a certificate trusted by your computer; WebAuthn doesn't work on plain HTTP.

Please use a relatively recent (2019 onwards) build of Chrome, Edge, Firefox etc.

Go to your user profile in the backend of the site.

Click on the ‘W3C Web Authentication (WebAuthn)’ tab.

On a Windows computer **without** any hardware authenticator attached click on Add New Authenticator.

### Actual result BEFORE applying this Pull Request

The browser asks you to plug in an authenticator.

### Expected result AFTER applying this Pull Request

You can enter your PIN / show your face / use a fingerprint scanner to register Windows Hello as an authenticator.

### Further testing

Delete the authenticator and try adding it again in the user profile page in the frontend of the site. It should still work.

Make sure that in the frontend you can delete an authenticator you added in the backend.

Make sure that in the backend you can delete an authenticator you added in the frontend.

Please make sure you can add more than one authenticators. **IMPORTANT!** You cannot add the same authenticator twice (in the past you could; it was a bug that went unnoticed). You can only test this if you have more than one authenticators, e.g. Windows Hello, a FIDO or FIDO2 hardware authenticator, an Android phone and so on.

Please make sure that you can edit the name of the authenticator. This was broken in #37464.

Please make sure you can log into the front- and backend of the site.

[...]

PR #37464, opened 2022-04-03
[4.1] webauthn table accessibility

brianteeman:
This pr adds the required table caption and column/row scope to the table.

There are no visual changes and brings this table into line with all other admin tables.

As webauthn requires https this PR cannot be tested on a site without https.

image

Issue #37485, opened 2022-04-04
[4.x] Language string sprintf


brianteeman:
There are a lot of lazy uses of %s in language strings passed through sprintf. For example this string was just merged

`COM_TEMPLATES_COPY_CHILD_TEMPLATE_STYLES="%s, copy of %s"`

The problem here is that both instances of %s are not the same

`			$styleName = Text::sprintf('COM_TEMPLATES_COPY_CHILD_TEMPLATE_STYLES', ucfirst($template->element . '_' . $newName), $style->title);
`

This string should have been "%1$s, copy of %2$s" otherwise it is not possible to translate the string correctly where the language uses a different order of parts.

@tecpromotion

nikosdion:
I disagree on principle.

The positional format of sprint arguments, e.g. `%1$s`, must **ONLY** be used when a. an argument is used more than once; or b. an argument is no longer used but the code is not updated for legacy reasons (e.g. supporting old language files); or c. the code passes the arguments in a certain order but the language string author needs them in another.

The canonical language of the Joomla CMS is “English (United Kingdom)”. Therefore the **developers** need to provide language files in this language and code which passes the arguments in the correct order for this argument. 

When a translator needs to transpose the elements of the string _in their own language's strings_ they can do so by employing the aforementioned format.

Example in English (United Kingdom) and Greek (Greece).

```php
$englishString = "%d %s thingamajics are already present in the %s container.";
$greekString = "O περιέκτης %3$s έχει ήδη %1$d ματζαφλάρια του τύπου %2$s."
```

As a translator I chose to rephrase this to make it read better in Greek, putting the counters where they need to be. I could have also gone for the direct but uglier translation:

```php
$greekString = "%d ματζαφλάρια του τύπου %s βρίσκονται ήδη στον περιέκτη %s."
```

So, no, Brian. Not including the position arguments in the canonical string DOES NOT make it impossible to translate. It is still absolutely and perfectly possible. The translator CAN use the positional arguments wherever they need to. Please read https://www.php.net/sprintf

brianteeman:
This convention is already used throughout joomla (with just a few exceptions hence this issue).

PR #37675, opened 2022-04-27
Refactored WebAuthn with Windows Hello support

nikosdion:
### Summary of Changes

* Refactored as a native Joomla 4 plugin, using a service provider and implementing SubscriberInterface
* Removed the ugly `Joomla` helper class, replacing it with native code
* Now using the (previously undocumented) WebAuthn library's `WebAuth\Server` object. This adds Windows Hello support **without** having to update to a new major version of the third party WebAuthn library
* We no longer pass URLs as data arguments; we can figure them out using `Joomla.getOptions`
* Fixed the breakage introduced in https://github.com/joomla/joomla-cms/pull/37464

The PR replaces #37673

### Testing Instructions

Please remember to run `npm ci` after applying the PR — the JavaScript has changed.

Please remember to use HTTPS with a certificate trusted by your computer; WebAuthn doesn't work on plain HTTP.

Please use a relatively recent (2019 onwards) build of Chrome, Edge, Firefox etc.

Go to your user profile in the backend of the site.

Click on the ‘W3C Web Authentication (WebAuthn)’ tab.

On a Windows computer **without** any hardware authenticator attached click on Add New Authenticator.

The browser asks you to plug in an authenticator.

You can enter your PIN / show your face / use a fingerprint scanner to register Windows Hello as an authenticator.

### Further testing

Delete the authenticator and try adding it again in the user profile page in the frontend of the site. It should still work.

Make sure that in the frontend you can delete an authenticator you added in the backend.

Make sure that in the backend you can delete an authenticator you added in the frontend.

Please make sure you can add more than one authenticators. **IMPORTANT!** You cannot add the same authenticator twice (in the past you could; it was a bug that went unnoticed). You can only test this if you have more than one authenticators, e.g. Windows Hello, a FIDO or FIDO2 hardware authenticator, an Android phone and so on.

Please make sure that you can edit the name of the authenticator. This was broken in #37464.

Please make sure you can log into the front- and backend of the site.

Please test on as many platforms as you have: Android (works on Android 9 and later if you have a fingerprint scanner but only on Chrome as far as I know), iOS/iPadOS (both TouchID and FaceID), macOS (TouchID, if you have a MacBook Air/Pro or an iMac/Mac Studio with Apple Silicon and the Apple keyboard with a TouchID sensor) as well as various FIDO and FIDO2 authenticators. I have tested all of these and Linux EXCEPT for Android due to lack of hardware running Android (my Android phone's battery bloated, I had to decommission it before it spontaneously turned into an incendiary grenade).

### Translation strings

The following language strings were added:

* `PLG_SYSTEM_WEBAUTHN_ERR_XHR_INITCREATE`
* `PLG_SYSTEM_WEBAUTHN_FIELD_ATTESTATION_SUPPORT_DESC`
* `PLG_SYSTEM_WEBAUTHN_FIELD_ATTESTATION_SUPPORT_LABEL`
* `PLG_SYSTEM_WEBAUTHN_LBL_DEFAULT_AUTHENTICATOR`

The following language strings were changed:

* `PLG_SYSTEM_WEBAUTHN_LBL_DEFAULT_AUTHENTICATOR_LABEL`

As of Joomla! __DEPLOY_VERSION__ the WebAuthn plugin has attestation enabled by default. This means that only authenticators with publicly verifiable cryptographic signatures can be registered with WebAuthn starting with this version of Joomla.

The publicly verifiable certification authorities for authenticators are retrieved from the FIDO Alliance site, namely the URL `https://mds.fidoalliance.org/`.

This default setting **will** prevent some cheaper authenticators which are not FIDO-certified from being used with WebAuthn. Moreover, some sites may be unable to download and/or cache the root certificates from FIDO Alliance, or it might take so long that the plugin aborts the operation to prevent your site from timing out. If you encounter any problems with registering authenticators with WebAuthn please edit the plugin settings and disable the Attestation Support option.

The Attestation Support feature requires the following prerequisites to work:

1. Your site must be able to access directly the URL `https://mds.fidoalliance.org/`.
2. Your cache folder (`administrator/cache`) must be writeable by PHP.
3. The system temporary directory must be writeable by PHP.
4. The OpenSSL extension must be installed and enabled on your site — this is a requirement for WebAuthn as a whole.

If these prerequisites are not met the WebAuthn plugin will proceed _without_ verifying the cryptographic signatures of the authenticators against the publicly verifiable certification authorities published by the FIDO Alliance. This is still secure — in fact far more secure than using a password and Two Factor Authentication. The only downside is that you may experience a short delay, up to 5 seconds, once a month when the plugin attempts to download the root certification authority information from the FIDO Alliance.

If your site meets all of the prerequisites except the first one you may download the information from `https://mds.fidoalliance.org/` and place them in the file `administrator/cache/fido.jwt`. In this case the WebAuthn plugin can operate with attestation support. This is very useful if your site is behind a firewall or disconnected from the Internet (e.g. on a high security intranet handling sensitive material). You need to remember to update this file once every month to avoid any problems.

Enabling the Attestation Support feature also allows Joomla to identify the maker and model of most FIDO2 certified authenticators. If you register an authenticator _after_ enabling this option you will see an icon of the maker's logo next to the Authenticator Name when viewing the list of authenticators. Furthermore, registering a new authenticator will have a more user-friendly default name, e.g. “Yubikey 5Ci added on 28 April 2022, 18:00” instead of “Authenticator added on 28 April 2022, 18:00”. 

If you disable the Attestation Support option the logo and the authenticator type will be hidden.

Finally, do note that authenticators added with previous versions of Joomla or while the Attestation Support feature is disabled or while the Attestation Support feature is enabled but its prerequisites not met will _always_ be displayed as “Generic Authenticator” as the necessary information to determine the make and model of the authenticator will have not been relayed to Joomla when you registered your authenticator.


brianteeman:
Working my way through the tests. 
When I try to remove an authenticator it fails with the following js error in the console

```
ncaught TypeError: Cannot read properties of undefined (reading 'querySelectorAll')
    at Object.e.plgSystemWebauthnDelete (management.min.js?e42299c3645fe9dd5d2ac5c502af8bce:formatted:172:44)
    at HTMLButtonElement.e.plgSystemWebauthnDeleteOnClick (management.min.js?e42299c3645fe9dd5d2ac5c502af8bce:formatted:216:7)
```

brianteeman:
>  IMPORTANT! You cannot add the same authenticator twice (in the past you could; it was a bug that went unnoticed). 

Is it really a bug if I want to use windows hello fingerprint and windows hello pin?

And if you insist it is can there be a more friendly error message please instead of 
`InvalidStateError: The user attempted to register an authenticator that contains one of the credentials already registered with the relying party.`

richard67:
@nikosdion It seems drone doesn't like your PR due to the many emojis in the description text.

nikosdion:
@brianteeman I will look into the removal of the authenticator, I think it's the same issue as the edit button.

Regarding the authenticator, you are viewing this the wrong way. The **authenticator** is one and the same: Windows Hello, the platform authenticator provided by Windows. How it's unlocked — be it a PIN or a fingerprint or a face scan or a PIV card or... — is an implementation detail of the authenticator which is NOT communicated to the WebAuthn consumer (the browser). The browser asks the authenticator “yo, authenticators, gather round and tell me if you know any of these registered public keys the server sent me” and the Windows Hello authenticator replies “eeeyup, one of these were issued by me, I can unlock it with my private key, don't ask me to do that again”.

The same goes for macOS where the platform authenticator can be unlocked by TouchID or double-clicking the side button of my paired Apple Watch. The unlocking gesture I use is an implementation detail. The browsers only see “macOS platform authenticator”.

nikosdion:
@richard67 Hahahaha! **I BROKE Drone, AGAIN!** It's a special talent. No worries, I will remove the Emojis.

brianteeman:
> Regarding the authenticator

I understand what you are saying  just not sure i agree. If I have **two** computers one with fingerprint and one without then it would be useful to have both fingerprint and pin setup

brianteeman:
... or as you know the state of my laptop the fingerprint doesnt always work so could use the pin as an alternative.

thats exactly how i login to the computer

nikosdion:
@brianteeman This is really up to the implementation on the OS side. 

Seriously, what we were doing before wouldn't change that behaviour at all. We were just not sending to the browser the list of registered public keys when asking for attestation (adding authenticators), therefore the browser couldn't ask the authenticators if they already knew about them. On assertion (login) we were sending this list. Only the first public key of each authenticator was taken into account by the authenticator itself, though. All additional keys? Ignored. You could see that if you dumped the information from the database table and checked the stored signature counter. It only ever increased in the first authenticator.

Anyway, this is a moot point not up to discussion as the WebAuthn specification mandates the corrected behaviour.

brianteeman:
> Only the first public key of each authenticator was taken into account by the authenticator itself, though. All additional keys? Ignored. 

To avoid misunderstanding are you saying that on 4.1 if i setup windows hello fingerprint and then windows hello pin only the fingerprint will work as a login method? That is not my experience

> You could see that if you dumped the information from the database table and checked the stored signature counter. 

Which table has the counter?

nikosdion:
As per https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer/create 

> `excludeCredentials` (Optional): An [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array) of descriptors for existing credentials. This is provided by the relying party to avoid creating new public key credentials for an existing user who already have some.

We pass the existing credentials for the user when we ask for a new attestation. Also see https://www.w3.org/TR/webauthn-2/#dom-publickeycredentialcreationoptions-excludecredentials 

What will happen next is between the **BROWSER** and the **AUTHENTICATOR(S)** neither of which is under our control. That's the entire point of WebAuthn! We the Relying Party (site) know exactly jack shit about the internal workings of the Authenticator and the Client. We don't have the Authenticator's private keys, we don't know how the Authenticator interacts with the user to unlock itself for cryptographic operations, we do not care how the Client communicates with the Authenticator, we don't care how the Client decides that an Authenticator has already been used to create credentials for our Relying Party ID (our site's URL). We are oblivious to all of that _by design of the WebAuthn specification itself_.

If Windows Hello or any other authenticator uses _multiple PRIVATE keys_ depending on how you unlock it, therefore rejects asserting a credential if you use the “wrong” way to interact with this Authenticator, this is not our business. We cannot and **must not** anticipate this as per https://www.w3.org/TR/webauthn-2/#sctn-security-considerations-rp points 4 and 5:
> 4. The [Relying Party](https://www.w3.org/TR/webauthn-2/#relying-party) can automatically support multiple types of [user verification](https://www.w3.org/TR/webauthn-2/#user-verification) - for example PIN, biometrics and/or future methods - with little or no code change, and can let each user decide which they prefer to use via their choice of [authenticator](https://www.w3.org/TR/webauthn-2/#authenticator).
> 5. The [Relying Party](https://www.w3.org/TR/webauthn-2/#relying-party) does not need to store additional secrets in order to gain the above benefits.

In the end of the day, our responsibility as a Relying Party (site) begins and ends by sending the excluded credentials to the Client (browser) and letting it figure out the gory details.

micker:
very cool feature ! i will test ASAP

brianteeman:
> You could see that if you dumped the information from the database table and checked the stored signature counter.

Which table has the counter?

> @brianteeman I will look into the removal of the authenticator, I think it's the same issue as the edit button.

Confirmed its the same issue and the same fix

nikosdion:
@brianteeman The credentials are stored in `#__webauthn_credentials`. However this the base64 representation of an encrypted JSON document. You should set a breakpoint at the last line of `\Joomla\Plugin\System\Webauthn\CredentialRepository::findAllForUserEntity` to see what is loaded from the database. Good luck!

brianteeman:
You are missing the point I am making about windows hello. Unless I am completely misunderstanding everything then windows hello is tied to the physical computer. So if I want to use windows hello with webauthn on both my work computer AND my home computer I need to be able to set up two instances of the windows hello authenticator. Doesnt matter if one is pin and the other fingerprint or both are fingerprint they are different because the harddware is different on the different pcs.

nikosdion:
Brian, this is **TOTALLY NOT** the use case you were discussing.

Each **physical device's** Windows Hello is a **different authenticator**. Likewise, each Apple device's TouchID/FaceID is a different authenticator. You can of course register different authenticators, therefore you'd be able to register different devices.

I actually have a test site where I've registered Windows Hello, a Yubikey 4, a Yubikey 5C, a Security Key NFC, two **different** Security Key (old FIDO version), my iPhone's FaceID, my iPad's FaceID, and my MacBook Pro's TouchID.

What you were saying before is that on the _same_ device Windows Hello behaves differently when unlocked by PIN and fingerprint which sounds implausible but I can't test because of lack of hardware. Regardless of how you unlock it, Windows Hello will interface the TPM (hardware or software) which has a singular, device-specific private key. I don't see why the unlock method would make it work differently; if it does, that's a problem with Microsoft's implementation. Considering that they were among the parties to help define the standard I find it implausible.

Have you actually tried with different Windows computers or are you just speculating?

brianteeman:
ok remotely logged in to a second computer and that works fine with its own windows hello - so all good there. and I get it now.

it was the error message that really confused me and led me into the path of brainfailure and the way that windows presents its options

> The user attempted to register an authenticator that contains one of the credentials already registered with the relying party.

brianteeman:
so other than the js error everything seems to be working fine - now i have my brain in gear - and as soon as you fix that one line of js I can mark it as a successful test

nikosdion:
Were you able to use a platform authenticator over RDP?! Windows 11 wouldn't let me do it, I had to be using a physical interface when I tried it yesterday. If that's the case I might resurrect Windows 10 on my old laptop to give this a go.

brianteeman:
I used nomachine to connect to my remote windows 10 computer and setup windows hello pin 

would it be possible instead of giving the name as "authenticator added on date" to include the name of the actual authenticator eg "windows hello authenticator added on date"

nikosdion:
@brianteeman We do not get a human readable authenticator name. All you get is the AAGUID but we _really_ don't want to have to trawl the Internet for all possible AAGUIDs and map them to human readable names (there is currently no registry; you have to check with each authenticator manufacturer).

brianteeman:
ok well no harm in asking

nikosdion:
It was a good question, one I had asked myself when I was implementing this 😉

plugins/system/webauthn/src/CredentialRepository.php
```diff
@@ -485,4 +584,67 @@ private function getEncryptionKey(): string
...
+	 *
+	 * @param   string|\DateTime  $date     The date to format
+	 * @param   string|null       $format   The format string, default is Joomla's DATE_FORMAT_LC6 (usually "Y-m-d
+	 *                                      H:i:s")
+	 * @param   bool              $tzAware  Should the format be timezone aware? See notes above.
+	 *
+	 * @return  string
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private static function formatDate($date, ?string $format = null, bool $tzAware = true): string
```
laoneo: Can you please make the functions in this pr not static and work with objects which are going to be injected?

nikosdion: Sure. I'm working on it now.

nikosdion:
@brianteeman @laoneo Ready for testing.

brianteeman:
FILE: ...rc/plugins/system/webauthn/src/PluginTraits/EventReturnAware.php
FOUND 1 ERROR AFFECTING 1 LINE
 16 | ERROR | Missing class doc comment
    |       | (Joomla.Commenting.ClassComment.Missing)

layouts/plugins/system/webauthn/manage.php
```diff
@@ -79,15 +78,15 @@
  * that problem.
  */
 $randomId    = 'plg_system_webauthn_' . UserHelper::genRandomPassword(32);
+/** @noinspection PhpInternalEntityUsedInspection */
+$authHelper  = new Authentication;
```
laoneo: Guess you can pass that from the plugin to $displayData

layouts/plugins/system/webauthn/manage.php
```diff
@@ -111,8 +110,8 @@
 		<?php // phpcs:ignore
 		foreach ($credentials as $method): ?>
 			<tr data-credential_id="<?php echo $method['id'] ?>">
-				<th scope="row"><?php echo htmlentities($method['label']) ?></th>
-				<td>
+				<th scope="row" class="webauthnManagementCell"><?php echo htmlentities($method['label']) ?></th>
```
laoneo: We never used camel case CSS classes, my preference is to do it BEM style.

plugins/system/webauthn/src/Authentication.php
```diff
@@ -0,0 +1,502 @@
...
+	/**
+	 * Public constructor.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct()
+	{
+		try
+		{
+			$this->app = Factory::getApplication();
```
laoneo: Inject the app

plugins/system/webauthn/src/Authentication.php
```diff
@@ -0,0 +1,502 @@
...
+		}
+		catch (\Throwable $e)
+		{
+			$this->app = null;
+		}
+
+		try
+		{
+			$this->session = ($this->app instanceof CMSApplication) ? $this->app->getSession() : null;
```
laoneo: We can also get the session from the container

plugins/system/webauthn/src/Authentication.php
```diff
@@ -0,0 +1,502 @@
...
+		}
+
+		if (!is_object($publicKeyCredentialCreationOptions) || !($publicKeyCredentialCreationOptions instanceof PublicKeyCredentialCreationOptions))
+		{
+			throw new RuntimeException(Text::_('PLG_SYSTEM_WEBAUTHN_ERR_CREATE_NO_PK'));
+		}
+
+		// Retrieve the stored user ID and make sure it's the same one in the request.
+		$storedUserId = $this->session->get('plg_system_webauthn.registration_user_id', 0);
+		$myUser       = $this->app->getIdentity() ?? new User;
```
laoneo, line 345: There is a user factory to create user objects, or when #37498 you just have to use getCurrentUser

plugins/system/webauthn/src/Authentication.php
```diff
@@ -0,0 +1,502 @@
...
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private function getCredentialsRepository(): CredentialRepository
+	{
+		if ($this->credentialsRepository !== null)
+		{
+			return $this->credentialsRepository;
+		}
+
+		$this->credentialsRepository = new CredentialRepository;
```
laoneo: Can this also be injected?

nikosdion:
The file does have a class comment, it's lines 17 through 21. What are you talking about?

plugins/system/webauthn/src/CredentialRepository.php
```diff
@@ -410,6 +410,105 @@ public function getHandleFromUserId(int $id): string
...
+	 */
+	public function getUserIdFromHandle(?string $userHandle): ?int
+	{
+		if (empty($userHandle))
+		{
+			return null;
+		}
+
+		/** @var DatabaseDriver $db */
+		$db = Factory::getContainer()->get('DatabaseDriver');
```
laoneo: You can use the databaseaware trait from the database package and pass it all the way down from the service provider. Accessing the db object should really be avoided, otherwise we have to refactor for 5 anyway.

richard67:
@nikosdion Brian is referring to the result of the PHP code style check of Drone: https://ci.joomla.org/joomla/joomla-cms/53685/1/6

plugins/system/webauthn/src/CredentialRepository.php
```diff
@@ -485,4 +584,67 @@ private function getEncryptionKey(): string
...
+			try
+			{
+				$tzDefault = Factory::getApplication()->get('offset');
+			}
+			catch (\Exception $e)
+			{
+				$tzDefault = 'GMT';
+			}
+
+			$user = Factory::getUser($userId);
```
laoneo: Use the user factory

nikosdion:
Groan! It's not in the committed file. Let me fix that.

richard67:
> Let me fix that.

I would be silly if I won't let you 😄

plugins/system/webauthn/src/Extension/Webauthn.php
```diff
@@ -0,0 +1,183 @@
...
+	 *
+	 * @return  array
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function getSubscribedEvents(): array
+	{
+		try
+		{
+			$app = Factory::getApplication();
```
laoneo, line 173: On the caching pr, you moved this checks to the respective function, or?

nikosdion:
Apparently when I was chunking my changes into separate commits in phpStorm I forgot to click on a hunk. All right, there we go, now the DocBlock is committed.

richard67:
Yep, PHPCS is happy now.

laoneo:
Instead of using traits, I would work here with inheritance. The traits have references to properties and functions which are not declared in the trait. So it looks for me like an abuse just for the sake of splitting the code.

nikosdion:
@laoneo The traits are used _exactly because_ we need to split the code into manageable chunks. Working with inheritance is a bad and far more precarious solution. The problem is that Joomla requires all of these event handlers to be in the same object. So you can either take this solution or I can make it into one supermassive class. Since I will be the one maintaining this code I can't work with a supermassive class and I choose to split it using Traits. Contorting the code in unnatural ways will just make it harder to maintain. Also, if you are using a decent IDE (such as phpStorm) it knows that the Trait is used by the extension class and it can autocomplete and type hint the properties just fine. Don't try to substitute a good IDE with a subpar code contortion solution. It will end in tears.

I will reply to everything else inline.

layouts/plugins/system/webauthn/manage.php
nikosdion: This file is not loaded by the plugin, it is loaded by the Field class which is loaded by the Form object of the com_users page.

layouts/plugins/system/webauthn/manage.php
nikosdion: I don't care. This is a fix to what Brian broke by changing the TD to a TH. I am not going to write even more precarious JavaScript to appease a certain aesthetic.

plugins/system/webauthn/src/Authentication.php
nikosdion: Absolutely not. This is not a random or lazy line of code. There was a lot of thought going into it.

System plugins are always loaded by all applications — including arbitrary .php files which implement their own WebApplication subclass. However, this plugin and especially this object will only work with CMSApplication objects. 

This object is created in the service provider very early in the application lifecycle (when it's booting the system plugins). We would have to move the entire application type check to the service provider to control whether the plugin will load _at all_. However, the way Joomla is written, this would just cause a fatal error because the plugin boot up would return an invalid object. It's also architecturally incorrect to do that.

Why not allow $app in this object accept any kind of ApplicationInterface object? Well, because it expects a CMSApplication and eventually we're going to type it statically once our min PHP requirement goes to PHP 7.4 or later. If it's typed as a CMSApplication trying to boot the plugin on a non-CMSApplication will result in a fatal error. Typing it as ApplicationInterface is wrong because the object does not work with anything other than a CMSApplication.

The only other approach is to have the object created in the plugin extension class' constructor, passing `$this->app` to it. But where does the plugin gets its app object from? That's right, by going to the Factory. This cannot change because to injection in any Joomla version it'd be a hard b/c break for all plugins in a way that would not allow the Joomla version introducing this change to provide b/c with older plugins. Therefore the Factory::getApplication will remain as long as Joomla exists so it's safe to use.

You see what's in front of you. I am taking into account all the edge cases you might have never seen and what Joomla will look like in the next 10 years.

plugins/system/webauthn/src/Authentication.php
nikosdion: No, sorry, what you say makes no sense.

The user factory takes an ID. What I want to do here is check whether the user ID in the session matches _the currently logged in user_ as a security measure (we don't want someone manipulating the session data to underhandedly link an Authenticator under their control to a user account other than their own).

The only sensible way to do this is to check if the application identity's ID matches the ID in the session.

However, the malicious user may be a bloody bastard, using a custom application object (e.g. through some vulnerability we have not found out) which does not have an identity. Or something might be broken when the malicious user manipulated the session. In this case I want to deterministically fall back to a Guest user object with ID 0 to **FORCE** the check to fail and generate an error.

That is to say, since this is a SECURITY CONTROL I want to only succeed when I am absolutely sure, beyond any shred of doubt, that the application's identity ID matches the ID I have in the session.

Using your trait would open a backdoor to this security control. If a user can maliciously manipulate the session who's to say that they can't use the same or a similar vulnerability to force a user object through your trait's public methods? We cannot have that! It would be a violation of security. Remember that this is **critical** for login security; once you link an Authenticator under your control to a user account you can log in as that user without anything stopping you; not even Two Factor Authentication / Two Step Verification.

So, no, I'd rather be super duper ultra extra cautious here and make sure that I am not opening a backdoor for no good reason.

plugins/system/webauthn/src/Extension/Webauthn.php
nikosdion: This is a different deal altogether.

This is a very special system plugin which acts as an authentication plugin. However, due to the interactive nature of the login ceremony (and the management pages) it can only ever work in the `site` and `administrator` applications.

However, it is still a system plugin. A system plugin can be booted from any WebApplication, including custom objects — and yes, I have absolutely seen real world extensions doing that (and yes, they should be using `com_ajax` but I am here to address real world use cases, not be a pendant and say that if an extension developer does something they needn't do but allowed to do we should throw a fatal error when we can avoid it). These application objects may NOT register themselves in the Factory. Hence the try/catch. If the application object cannot be fetched from the Factory we have one of these bad uses cases and we must not let our plugin load.

The next if-block checks that our application is indeed the core site or administrator application.

Since this static method is called early in the plugin boot process and we know we need none of the event handlers in these cases we can add this check right here and write the rest of our code knowing it cannot execute unless we are inside a CMSApplication.

plugins/system/webauthn/src/Authentication.php
nikosdion: I am closing this as the reasoning is explained in the replies to your comment on the lines following it.

plugins/system/webauthn/src/CredentialRepository.php
nikosdion: This I can fix; I missed it on the refactoring passes.

plugins/system/webauthn/src/Authentication.php
nikosdion: I cannot because this needs to be instantiated by the Field which is loaded by a Form object outside the plugin.

plugins/system/webauthn/src/CredentialRepository.php
nikosdion: I cannot because this object needs to be instantiated by the Field which is loaded by a Form object outside the plugin.

brianteeman:
i just applied the changes and rebuild the js
i can confirm that the rename/delete buttons now work
but after deleting the existing aauthenticators and trying to create new ones from scraatch I just get an error of 500

image

brianteeman:
If I revert the last commit to use the user factory  0bc36d6bf36ccbd99bec86823deb9651bcbec501 then all seems to be good

brianteeman:
I have successfully tested the following

1. admin create and use windows hello and android
2. site create and use windows hello and android

The only oddity is something that I suspect cannot be fixed as below - but marking this as successful test

if you login using an authenticator and then delete the authenticator and **without saving the user** try to re-add the same authenticator then it fails with the cant use the same authenticator again. Which is slightly confusing but understandable.

brianteeman:
I have tested this item :white_check_mark: successfully on b80b3b9fb9eb72da6e51f63cc9529fafe4c2b828<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37675">issues.joomla.org/tracker/joomla-cms/37675</a>.</sub>

nikosdion:
@brianteeman I actually ran into the same issue when I was fixing the problem you reported earlier. It can be fixed and I am fixing it now. It's almost 11pm, though, so I am not sure if I will finish tonight.

Fixing that will also fix the comments @laoneo made about the manage.php file. And you know what? I might be able to inject most of the stuff he asked me to inject since the critical stuff will be going through the plugin instead of being accessed through a form object.

plugins/system/webauthn/src/Authentication.php
```diff
@@ -0,0 +1,502 @@
...
+		return new PublicKeyCredentialUserEntity(
+			$user->username,
+			$repository->getHandleFromUserId($user->id),
+			$user->name,
+			$this->getAvatar($user, 64)
+		);
+	}
+
+	/**
+	 * Get the user's avatar (through Gravatar)
```
brianteeman, line 472: is this actually used?

brianteeman:
I have not tested this item.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37675">issues.joomla.org/tracker/joomla-cms/37675</a>.</sub>

brianteeman:
> @brianteeman We do not get a human readable authenticator name. All you get is the AAGUID but we _really_ don't want to have to trawl the Internet for all possible AAGUIDs and map them to human readable names (there is currently no registry; you have to check with each authenticator manufacturer).

Did a bit of research and for the hardware keys there are only some incomplete and [unofficial lists](https://blog.workinghardinit.work/2021/11/01/fido2-aaguid-lists/) available. However for windows hello they are published and there are only 4 so at least those could be used

> Windows Hello employs a variety of ways to protect user credentials. You can check which method has been used to protect a credential by consuming the [AAGUID](https://w3c.github.io/webauthn#sec-attested-credential-data) field in the attestation object returned at credential creation. The following is the list of AAGUIDs that Windows Hello may return:
**Software backed authenticators**

- Windows Hello software authenticator: 6028B017-B1D4-4C02-B4B3-AFCDAFC96BB2
- Windows Hello VBS software authenticator: 6E96969E-A5CF-4AAD-9B56-305FE6C82795

**Trusted Platform Module (TPM) backed authenticators**

- Windows Hello hardware authenticator: 08987058-CADC-4B81-B6E1-30DE50DCBE96
- Windows Hello VBS hardware authenticator: 9DDD1817-AF5A-4672-A2B9-3E3DD95000A9

brianteeman:
Keeping this seperate from previous post which I wrote last night but forgot to hit save.

> @brianteeman We do not get a human readable authenticator name. All you get is the AAGUID but we really don't want to have to trawl the Internet for all possible AAGUIDs and map them to human readable names (there is currently no registry; you have to check with each authenticator manufacturer).

It would appear that there is a central list now maintained by the fido alliance as a jwt blob that is updated on a monthly basis. https://fidoalliance.org/metadata/

After decoding the jwt blob I can confirm that the windows aaguid in the previous post are all present so it looks like this can be used as a trusted data source.

_(prior to mds v3 i dont think this was a public data source)_

nikosdion:
Okay, it's been a wild ride but I think I have things which will please both @brianteeman and @laoneo in these commits.

All the injection stuff is fixed, there are no more magic object creations that I can find. Phew.

You can now see the make and model of your authenticator as well as the logo of its manufacturer **IF AND ONLY IF** you add it AFTER applying the latest changes. The AAGUID is not communicated to Joomla unless we ask for full attestation which is something I added in these commits.

To implement this feature I had to enable full attestation support. This means that the authenticator signatures are also checked against the root certificates known to the FIDO Alliance (h/t to Brian for the link). This makes Joomla's WebAuthn implementation _even more secure_ and suitable for use in some high security environments (we don't have a feature to lock down the authenticator types and/or add custom certificates which would be required in very high security applications of the government, military, financial institution etc type — then again, Joomla isn't FIPS certified so it can't be used there anyway).

You CAN of course disable the attestation support if you like as it may prevent cheaper, not certified FIDO keys from being used and/or may cause problems on some sites behind firewalls, on slow connections, on subpar server configurations and the like. While I try to fail safe in these cases the only thing I cannot do is allow an uncertified key to be used after the attestation ceremony. If you need to use this kind of keys, no problem, disable Attestation Support in the plugin options and Bob's your uncle.

There was a massive bug with U2F keys (the old FIDO keys) and Android Keys on PHP 8 in the version of the WebAuthn library we are using. Since we cannot upgrade the library I had to fork three objects from the library. I marked these `@deprecated 5.0` since we can get rid of them once we upgrade the library in Joomla 5.

I have so far tried this with a small army of Yubico hardware authenticators, macOS / iOS / iPadOS built-in authenticators and Windows Hello. I have NOT tried this with Android due to lack of hardware; if you have an Android phone please do test that too.

Oof. I think that's all. Please test, test again and test some more. Since we're putting all this effort into this plugin let's do it right.

plugins/system/webauthn/src/Authentication.php
nikosdion: Yes, but currently the browsers don't display the user's image. I am keeping this with the assumption that this will change in the future. Apple's Passkeys seem to be moving towards that direction (still in alpha, still not working very well, time will tell). So, for now, this stays in.

plugins/system/webauthn/src/Authentication.php
nikosdion: It involved a lot of bending over backwards and using the service provider in ways we don't use it elsewhere in the core but I did it so I am resolving this conversation.

brianteeman:
awesome work - thanks. Will do plenty of android tests tonight and report back.

administrator/language/en-GB/plg_system_webauthn.ini
```diff
@@ -29,7 +29,8 @@ PLG_SYSTEM_WEBAUTHN_FIELD_N_AUTHENTICATORS_REGISTERED="%d WebAuthn authenticator
 PLG_SYSTEM_WEBAUTHN_FIELD_N_AUTHENTICATORS_REGISTERED_0="No WebAuthn authenticator has been set up yet"
 PLG_SYSTEM_WEBAUTHN_FIELD_N_AUTHENTICATORS_REGISTERED_1="One WebAuthn authenticator already set up: %2$s"
 PLG_SYSTEM_WEBAUTHN_HEADER="W3C Web Authentication (WebAuthn) Login"
-PLG_SYSTEM_WEBAUTHN_LBL_DEFAULT_AUTHENTICATOR_LABEL="Authenticator added on %s"
+PLG_SYSTEM_WEBAUTHN_LBL_DEFAULT_AUTHENTICATOR="Generic Authenticator"
+PLG_SYSTEM_WEBAUTHN_LBL_DEFAULT_AUTHENTICATOR_LABEL="%s added on %s"
```
brianteeman, line 35: please see https://github.com/joomla/joomla-cms/issues/37485 and update accordingly

plugins/system/webauthn/src/MetadataRepository.php
```diff
@@ -0,0 +1,246 @@
+<?php
+/**
+ * @package         Joomla.Plugin
+ * @subpackage      System.Webauthn
+ *
+ * @copyright   (C) 2020 Open Source Matters, Inc. <https://www.joomla.org>
```
brianteeman: is this a new file? if so then 2022

nikosdion: My code on my local repo states clearly:
```php
/**
 * @package         Joomla.Plugin
 * @subpackage      System.Webauthn
 *
 * @license         GNU General Public License version 2 or later; see LICENSE.txt
 */
```

I actually made a separate commit for this: https://github.com/joomla/joomla-cms/pull/37675/commits/786d1659421fca1775f3e70adc1fcf54bf8322f2

The diff on GitHub also shows 2022: https://github.com/joomla/joomla-cms/pull/37675/files#diff-0ebd52c447bfcbffa4c58f3219f821d0bcc84f31814e30abafb85c8ff3fc5506

How is it even possible that you see something else?!!

administrator/language/en-GB/plg_system_webauthn.ini
nikosdion: I won't for the reasons I stated there. It's the translator's job to add positional arguments **IF AND ONLY IF** they are needed. Anything else is busywork for the developers, making language strings even less readable.

plugins/system/webauthn/src/MetadataRepository.php
brianteeman: i blame github cache.

brianteeman: you can mark this as resolved

brianteeman:
image


is it worth using the images? wont people just comment that they have a missing image for non-fido?

nikosdion:
@brianteeman This is not a missing image. There are never missing images. It's me doing a stupid thing and using a white background when the Windows Hello image is a white smiley face on transparent background. Shoot. Let me fix that very quickly.

nikosdion:
That should do the trick.

brianteeman:
That did fix it but to be honest with you I would drop the image. Doesnt look great (not your fault and you're not a designer). The inline styles will trigger some people and I'm pretty sure we dont need an ALT, Title and Aria

Otherwise everything in my tests worked well. Note that I have not tested what would happen when upgrading on a site that already has authenticators

nikosdion:
The images are those provided by the vendors 🤷🏽‍♂️ 

I think it's an okay solution to displaying the authenticator type when you have full attestation, maybe not so much if you're using multiples of the same authenticator. If people really hate it we can remove it and replace it with a small piece of text under the authenticator name.

brianteeman:
> f people really hate it we can remove it and replace it with a small piece of text under the authenticator name.

I really hate it ;) But maybe the authenticators you have are all the same size image
What would the small piece of text say - can't imagine anything of value

image

brianteeman:
I have tested this item :white_check_mark: successfully on 71815ff82acca78f380602f3afa8ede266dfa786<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37675">issues.joomla.org/tracker/joomla-cms/37675</a>.</sub>

richard67:
Appveyor (which does the ci checks on Windows) fails. It reports problems with the composer.lock file. See https://ci.appveyor.com/project/release-joomla/joomla-cms/builds/43381337 . Any idea what the problem could be?

brianteeman:
looks like ext-sodium is missing from the version of php on appveyor

richard67:
> looks like ext-sodium is missing from the version of php on appveyor

Yes, that’s one of the reported problems.

brianteeman:
> Yes, that’s one of the reported problems.

I _think_ they are all related

richard67:
I‘ll try to get someone who knows more about the test set up than I do.

nikosdion:
Brian is right. The problem is that the Windows build machine doesn’t have Sodium. 

The library it complains about is used for EdDSA support. However, our plugin disables the EdDSA algorithm for attestation and assertion (linking authenticator and logging in) unless Sodium is installed. Of course the dependency needs to be installed to be usable in runtime when server conditions permit and that — installing the dependency via Composer — is what exposed the problem with the automation. 

Hey, is this, dunno, the third time I break the automation with this PR by doing something it didn’t anticipate? I have a knack for doing that. Sorry!

brianteeman:
> Brian is right.

nikosdion:
@richard67 There you go! I followed the suggestion in AppVeyor's output and it's working now. Basically, I tell it to ignore the fact that a dependency requires Sodium which isn't enabled and install it anyway. This is safe as we don't actually use the code that uses Sodium in AppVeyor.

richard67:
> @richard67 There you go! I followed the suggestion in AppVeyor's output and it's working now. Basically, I tell it to ignore the fact that a dependency requires Sodium which isn't enabled and install it anyway. This is safe as we don't actually use the code that uses Sodium in AppVeyor.

@nikosdion I see, and I see it works. Thanks.

richard67:
@brianteeman I've restored your successful test result in the issue tracker since the commit which invalidated the test count has not changed anything on the tested code but only changed the test setup, see commit https://github.com/joomla/joomla-cms/pull/37675/commits/d2e20ceb6ac91abf94fb5a06a1396a598d3823a0 .

plugins/system/webauthn/src/Authentication.php
```diff
@@ -0,0 +1,586 @@
...
+	 *
+	 * @param   ApplicationInterface|null                 $app       The app we are running in
+	 * @param   SessionInterface|null                     $session   The app session object
+	 * @param   PublicKeyCredentialSourceRepository|null  $credRepo  Credentials repo
+	 * @param   MetadataStatementRepository|null          $mdsRepo   Authenticator metadata repo
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct(
+		?ApplicationInterface                $app = null,
```
laoneo, line 99: I really like the direction this pr is heading to. But what is the benefit to make the dependencies not mandatory?

brianteeman: > we don't actually use the code that uses Sodium in AppVeyor.

nikosdion: @laoneo I was working from everything is automatic to everything is explicitly provided. I am also not sure if someone would try to reuse that helper outside the plugin. So I left it like that.

What definitely needs to be optional is the metadata statement repository. This object is only required when requiring full attestation information from the client when doing the initial attestation ceremony.

If there's a good reason to make the app and session mandatory I can of course remove the default values and the code which handles them when null. It's easier to remove than to add :)

nikosdion:
@richard67 You may want to restore Brian's test again. I fixed a bug in the (old) WebAuthn library we are using and invalidated his test.

Also, I am continuing development on this as a third party plugin since it takes a very long time to get anything included in the core 😉 Some cool things I have implemented and will be able to contribute after this PR is merged:

* **Support for Apple Passkeys / synchronised authenticators**. The WebAuthn specification says that clients _may_ choose to synchronise authenticator credentials among several devices. Google, Microsoft and Apple have already come up with the plumbing and, shockingly, decided on a common standard. macOS 12.3 and iOS/iPadOS 15.4 have added preliminary support for this standard and call it Apple Passkeys (marketing name) / Syncing Platform Authenticator (developer-facing documentation). Enabling it is currently restricted to those with an Apple Developer account which I have, so I can tell you I figured out how to enabled support for it. Now, get this. This feature also allows a Mac computer to use an Android device as the authenticator, even in Safari! Reports of flying pigs are unconfirmed but deemed plausible in the light of this development.
* **Support for iOS/iPadOS 15.4 and later**. Apple changed the attestation statement of its mobile devices to use the `apple` format instead of the `packed` format. Our very old copy of the WebAuthn library (2.x) does not support it _but_ I just finished backporting it from version 4.x — and make it PHP 7.2 compatible. It wasn't that hard and solves the problem of WebAuthn not working properly on mobile Apple devices.
* **Support for resident authentication credentials (passkeys)**. Right now, our WebAuthn plugin only supports simple authentication credentials. These do not carry a user identification in them which is why you need to enter your username for password less authentication. Resident authentication credentials (marketing name: Passkeys) are different. They are stored _in the authenticator device_ and contain the user identifier. Therefore you can log in without a password _and without a username_. It's magical! Caveat: hardware authenticators like FIDO2 USB-attached keys have limited memory and can only store 5 to 20 credentials. Before late 2021 you couldn't remove them; not the case anymore. Platform Authenticators, through, such as Windows Hello, Apple's TouchID/FaceID and Android biometrics have **unlimited storage capacity**. This means that you can reduce logging in to all of your Joomla sites to a biometric, PIN or device password check. No usernames.

All of these are features I implemented today in my copy of the plugin. I don't want to put them in this PR here as it'll get into massive scope creep, testability problems and we'll miss yet another Joomla version to ship improvements to WebAuthn. Once this PR is merged I'll do two PRs for these features. One for Apple stuff, one for the resident credentials. 

I believe that this will put Joomla well ahead of any other CMS in terms of secure and easy authentication options. THE END (of passwords as means of authentication) IS NIGH!

plugins/system/webauthn/src/PluginTraits/EventReturnAware.php
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * Adds a result value to an event
+	 *
+	 * @param   Event   $event  The event we were processing
+	 * @param   mixed   $value  The value to append to the event's results
+	 *
+	 * @return  void
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private function returnFromEvent(Event $event, $value = null): void
```
HLeithner, line 32: As discussed in the Event PR, I would say that should be moved to the (immutable?) Event class because this needs to be cause by all event listeners which needs/adds a result.

nikosdion: When I wrote this PR a month ago the Events PR wasn't merged :) I will circle back to this PR tomorrow and update it to remove the now no longer necessary trait. I suppose this gives us enough time to merge it in time for 4.2, right?

nikosdion: So, after looking at this, I cannot remove this trait.

Right now the Joomla core code (and the third party event code) is trying to create a GenericEvent which does not implement ResultAware for the various generic events we handle (onUserAfterDelete, onUserLoginButtons, onContentPrepareForm, onContentPrepareData). We need to create concrete events for these core events _before_ I can refactor this code.

What I **can** do as part of this PR is create concrete events for the various `onAjaxWebauthn*` events. I will do it now.

I suggest that the rest of the events are migrated in a different PR. We don't want to expend the scope of this PR to encompass completely unrelated features (scope creep is bad).

nikosdion:
I am no longer interested in contributing to a project which allows toxic people to call into question my integrity as a business owner and my ability as a software developer to write working software. Good luck figuring this out by yourself.

**WARNING: DO NOT EVEN TRY TO COPY MY OWN CODE (WHICH IS LICENSED UNDER GPLv3).**

richard67:
@nikosdion Please think it over and restore your branch and reopen this PR. If it is really like you say, do you really think one person can damage your reputation? I don’t think so. Yes, in a community driven open source project there are all kinds of people, and some are not so easy to deal with. But as long as they don’t violate CoC, what shall we do? Let me know if there is anything I can do to make you change your mind. If necessary you can contact me via the contact form on my website and we can have a talk. I want peace and I do not want to lose your expertise for the project. Am saying this as a normal contributor, not speaking for anybody else, but I am sure many others fell like I do.

nikosdion:
@richard67 I am sorry, what?! Putting someone's integrity into question (i.e. calling someone a liar), even passively-aggressively, and attacking their experience is not a CoC violation? It's an ad hominem attack!

The same thing happened to me ten years ago when Andrew Eddie personally attacked me. The end result is that he got a bit less than a slap on the wrist and allowed to continue contributing (and abusing people) just fine while I was handed a six month ban for defending myself against an uncalled for personal attack. Unfortunately, nothing has changed in the last ten years. Victim blaming abound, toxic behaviour left unchecked, normalised and encouraged.

Let me put it this way. Would you like to work in a company where your co-workers call you a fraud and a liar? Then why do that without being paid and have this unfounded accusations levelled against you in public?! If contributions are punished then I shall not contribute.

nikosdion:
I redid this PR in https://github.com/joomla/joomla-cms/pull/37910

PR #37701, opened 2022-04-30
[4.2] Mail Templates htmlbody

brianteeman:
### Summary of Changes
html emails are much much bigger than plain text emails. Using some email templates from beefree.io was almost impossible as they were all too big to be saved. Checking the database the field type was TEXT and yet for the fulltext field in #__content MEDIUMTEXT was used. So it wasnt even possible to use an entire article in the mail template.

This PR changes the field to MEDIUMTEXT in mysql for updates and new installations. Note postgres doesnt need any change as it only has one datatype for text.

### Testing Instructions
Upgrade with the prebuilt test packages and check the datatype of the htmlbody field in the #__mail_templates table
Clean install with he prebuilt test packages and check the datatype of the htmlbody field in the #__mail_templates table

check of the database checker in the system dashboard. 
. In case of an update to the patched package for the PR, the checker should only show the usual error about not matching CMS version when using such packages, and in case of a new installation no errors should be shown.

TEXT

MEDIUMTEXT

none


administrator/components/com_admin/sql/updates/mysql/4.1.3-2022-04-30.sql
```diff
@@ -0,0 +1,5 @@
+--
+-- Increase the size of the htmlbody field in the #__mail_templates table
+--
+ALTER TABLE `#__mail_templates`
+	CHANGE COLUMN `htmlbody` `htmlbody` MEDIUMTEXT NOT NULL COLLATE 'utf8mb4_unicode_ci';
```
richard67: Please use MODIFY and not CHANGE when not renaming a column.

It is good practice to do so because less error prune, no risk for a typo in the new column name.

I know we don’t follow that good practice everywhere, but we should not establish new bad examples.

richard67: Find an example here which is also understood by the DB checker: ~~https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_admin/sql/updates/mysql/4.0.0-2018-07-19.sql#L5~~

Sorry, wrong one, I mean https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_admin/sql/updates/mysql/4.0.0-2018-08-29.sql#L2

richard67: I think currently the DB checker doesn’t understand the optional keyword COLUMN in a MODIFY or CHANGE statement.

richard67: See code beginning here: https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Schema/ChangeItem/MysqlChangeItem.php#L155 and then down to the end of the next case.

richard67: It also doesn’t need to specify collation in the update SQL because it should be right and. It be modified when changing the column. We also don’t do that in the example I‘ve linked in one of my comments above.

brianteeman: just tell me exactly what you want - i'm happy as long as i can use my mail templates

richard67: Ok, will resolve this conversation and make a new one with a proposed code change.

brianteeman: thank you

richard67:
```suggestion
ALTER TABLE `#__mail_templates` MODIFY `htmlbody` mediumtext NOT NULL COLLATE 'utf8mb4_unicode_ci';
```

brianteeman: that is giving me an error with my existing mail template records
> /* SQL Error (1366): Incorrect string value: '\xE2\x98\xB0</s...' for column 'htmlbody' at row 12 */

but if I add the COLLATE then it works perfectly

even after running it once with collate and then trying again without collate it gives the same error so collate must be required ?

richard67: Hmm ... does your mail templates table still have the right default charset and collation like stated here https://github.com/joomla/joomla-cms/blob/4.1-dev/installation/sql/mysql/supports.sql#L406 ? And does the column have that charset and collation before running the SQL statement?

Am just curious what causes that error. I would have to check where else we modify text columns and if so add that collate expression.

To be on the safer side I add it to my suggestion above.

richard67: Not sure though if adding the COLLATE would not cause problems on databases where people used things like Akeeba Admin Tools to change their collations.

richard67: Collate is definitely not required regarding the syntax just by the way.

brianteeman: i'm attaching the table

[yle.sql.txt](https://github.com/joomla/joomla-cms/files/8596853/yle.sql.txt)

richard67: @brianteeman In that file, your table is created with latin1 as default charset:

```
CREATE TABLE IF NOT EXISTS `fiqip_mail_templates_copy` (
  `template_id` varchar(127) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `extension` varchar(127) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `language` char(7) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `subject` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `body` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  `htmlbody` mediumtext CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  `attachments` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  `params` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  PRIMARY KEY (`template_id`,`language`) USING BTREE
) ENGINE=InnoDB DEFAULT CHARSET=latin1 ROW_FORMAT=DYNAMIC;
```

I think that's the problem.

Joomla creates tables with utf8mb4 charset and utf8mb4_unicode_ci collations, but it does not touch the default character set of the database because that might require higher privileges than we have.

Now when you create a new table and do not explicitly specify default character set and default collation for that table, the database defaults are used, which still might be latin1 on older database servers, and these table defaults are then used when modifying the column.

Something like that must have happened when you created your copy of the core table.

richard67: That means when we modify the column and not use the COLLATE part, it defaults to the table default now, which is latin1.

If your table would be created like core Joomla tables, that would not happen:

```
CREATE TABLE IF NOT EXISTS `fiqip_mail_templates_copy` (
  `template_id` varchar(127) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `extension` varchar(127) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `language` char(7) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `subject` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `body` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  `htmlbody` mediumtext CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  `attachments` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  `params` text CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  PRIMARY KEY (`template_id`,`language`) USING BTREE
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
```

brianteeman: The database was created by the joomla installer

richard67: The database default character set and collation of the database is not specified in the create database statement in this case, as far as I remember, and so the values from server settings are used in this case.

Anyway, the table has the wrong default character set, wherever that might come from, as your text file clearly shows.

If it has the right one the problem will not happen.

Up to now we do not use COLLATE in any of our update SQL scripts in any of the "ALTER TABLE ... CHANGE ..." or "ALTER TABLE ... MODIFY ..." statements for columns with text data types.

See e.g. here https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_admin/sql/updates/mysql/4.0.0-2020-03-10.sql#L2-L7

or here https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_admin/sql/updates/mysql/4.0.0-2018-03-05.sql#L80 .

For the latter change I would expect similar issues when the table has utf8mb4 data in that column but has default character set latin1.

Not sure now what's better, remove the COLLATE from my suggestion above, or add it at all these places.

If people have changed their collation e.g. with Admin Tools to some other utf8mb4 collation, it should not be a problem to force utf8mb4_unicode_ci here. Things like illegal mix of collations I would only expect when mixing with utf8 or a 1 byte character set like latin1.

On the other hand we would never run into your problem if your table had the right default character set and collation, and we do not have COLLATE at all the other places I've found.

~~So I think I would prefer not to use it here.~~

brianteeman: image

richard67: Well then at least change CHANGE to MODIFY ... I've added the COLLATE back to my proposal.

richard67: P.S.: Also the optional keyword "COLUMN" shall be removed, like I proposed, because the database checker doesn't understand it. Am working since a while on a future PR to change that.

brianteeman: was there a reason you changed MEDIUMTEXT to mediumtext?

richard67: Consistency with other update SQL script where data types are in lowercase.

brianteeman: what lovely consistency as the data types are all uppercase in the installation sql

richard67: yes but that would be a bigger clean up.

richard67:
@brianteeman To make it perfect you could add the check of the database checker in each case (update or new install) to the testing instructions. In case of an update to the patched package for the PR, the checker should only show the usual error about not matching CMS version when using such packages, and in case of a new installation no errors should be shown.

richard67 (approved):

Kostelano:
I have tested this item :white_check_mark: successfully on 594d6b98b31a2f5eca719da237692032693b4c92<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37701">issues.joomla.org/tracker/joomla-cms/37701</a>.</sub>

richard67:
@brianteeman It needs to rename the update SQL script. Assuming that this PR remains on the 4.1-dev branch, it should be a name with version 4.1.5, and when 4.1.5 is released before this PR is merged, it needs a name with version 4.1.6. But I think the PR should be rebased to 4.2-dev. If this is the case, it needs a name with version 4.2.0 and a later date than 2022-05-15 (because "4.2.0-2022-05-15.sql" is the latest 4.2.0 update SQL script).

@roland-d Shall this PR be rebased to 4.2-dev?

brianteeman:
I await your response before making any changes. To be clear this is NOT a new feature it is fixing a bug

richard67:
> I await your response before making any changes. To be clear this is NOT a new feature it is fixing a bug

@brianteeman I know. But 4.1.5 is already in RC phase, so @bembelimen will merge only extremely urgent bug fixes (or release blocker), and very likely will be the last 4.1.x release, as far as I know, so it will go in 4.2-dev anyway.

brianteeman:
Rebased to 4.2 and changed the filename

richard67:
I’ve restored the previous human test result in the issue tracker because the changes which invalidated the test count were just a clean rebase and the rename of the update SQL script (which I reviewed).

alikon:
I have tested this item :white_check_mark: successfully on b033c55459163379e639e2dc1d5a50a9cd07e854<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37701">issues.joomla.org/tracker/joomla-cms/37701</a>.</sub>

alikon:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37701">issues.joomla.org/tracker/joomla-cms/37701</a>.</sub>

roland-d:
Thanks everybody

Issue #31530, opened 2020-11-29
Tag links are pointing to a nonSEF URL in multilanguage mode


hakanara:
### Executive summary
Tag links on **Tags Popular module** and **articles** are pointing to a **nonSEF URL** in multilanguage mode like www.mysite.com/component/tags/tag/joomla.html even if there is a menu item that is set to "List all tags".

### Steps to reproduce the issue
- Setup a new MULTILANGUAGE web site by following these steps (Fresh setup for multilanguage site):
  https://docs.joomla.org/J3.x:Setup_a_Multilingual_Site/Installing_New_Language

- Create 2 "List All Tags" menu items for each language (In my case Turkish and English)
  (Where the created menu items lead to: index.php?option=com_tags&view=tags)

- In Global Configuration, say yes to these 3 settings:
  Search Engine Friendly URLs
  Use URL Rewriting
  Add Suffix to URL 

- Rename htaccess file on the server as .htaccess.

### Expected result
Tags on the "Tags Popular" module pointing to a menu like:
www.mysite.com/tr/myTurkishTagsAlias/joomla and 
www.mysite.com/en/myEnglishTagsAlias/joomla.
(Joomla is the only available tag after the setup.)

### Actual result
www.mysite.com/en/component/tags/tag/joomla.html
www.mysite.com/tr/component/tags/tag/joomla.html
(This issue is the same for blog post tags and the tags in "tags popular" module.)

_If there is only one tag menu item assigned to all languages, it works as I have expected above._

### System information (as much as possible)
Joomla 3.9.23 (Latest as of today)

### Additional comments
You can read the same issue here that was written many years ago:
https://forum.joomla.org/viewtopic.php?t=859288

It might be a router issue because the same issue arises on the blog view. The tags of the articles also lead to a link like component/tags/tag/joomla.html when there is more than one tags menu item that are assigned to different languages.

Thank you very much to all the members of Joomla team who are putting their love to this wonderful CMS.

hakanara:
After making some more research, I realize that this conflict only arises when you set the language of "List all tags" menu item to a specific language. If the language of the menu item is set to "all" the links look OK. So also in this case, I couldn't assign two different alias to tags for two different languages.

infograf768:
have you set Tags Options to use "Current"?
[...]

Issue #31754, opened 2020-12-22
[4.0] Tags Module shows tags from other languages


sakiss:
### Steps to reproduce the issue
1. A multi-lingual site is required (for the example i will use English and German).
2. Create a tag and set "English" in the language field.
3. Create an article in German and assign the previously created tag.
Btw. Loading tags in different language than the article's one, seems buggy.
4. Create and enable a Tags module.
5. Check your front-end in the German language.

Tags
![tags_be](https://user-images.githubusercontent.com/1838041/102916719-ac31ff00-448c-11eb-8404-d9100e0068a0.png)

Article assignment
![article_tags](https://user-images.githubusercontent.com/1838041/102916726-b05e1c80-448c-11eb-83f0-3002b4751488.png)

### Expected result
The tags in languages other than the site's language not to be loaded.

### Actual result
It shows tags from all the languages, that were assigned at least in 1 article.

![tags_fe](https://user-images.githubusercontent.com/1838041/102916736-b48a3a00-448c-11eb-8f37-9bf4de519b3b.png)

### System information (as much as possible)

Joomla4.0 beta6

### Additional comments
Language Filter : Current in the com_tags configuration

chmst:
As long as an atricle is not saved, the system cannot know for which language you are writing, so how should it select a language for tags? I think that this is normal behaviour.

bembelimen:
Go to the Tags options in the backend and set the "language filter" to "Current" to have the expected behaviour (ofc the language detection should set to "browser" in the language plugin):

![grafik](https://user-images.githubusercontent.com/1229869/103181804-8306d980-48a5-11eb-8f7e-808124390768.png)

sakiss:
@bembelimen I already pointed that the "Language Filter" is set to current.

Plz read the post and try to reproduce the issue, before closing it.

bembelimen:
I did and it works as expected:

![grafik](https://user-images.githubusercontent.com/1229869/103213400-db71c180-490d-11eb-8229-172103c1c4c4.png)

German:
![grafik](https://user-images.githubusercontent.com/1229869/103213434-ed536480-490d-11eb-866a-a7474929d0f4.png)

English:
![grafik](https://user-images.githubusercontent.com/1229869/103213448-fc3a1700-490d-11eb-8d30-1990fd1bc780.png)

sakiss:
@bembelimen plz follow the instructions as mentioned in my post.
The tag needs to be in different language than the article.

[...]

PR #37721, opened 2022-05-02
Make the Tags RouteHelper language aware

roland-d:
### Summary of Changes
While building a multi-lingual site with tags we found out that the mod_tags_popular module was not producing SEF URLs due to missing the language string. When the menu items had their language set to All (*) it worked fine but when changing them to English (en-GB) and Dutch (nl-NL) the links are no longer working. After some investigation I found out that the Tags RouteHelper does not take the current language into account. This fix resolves that.

### Testing Instructions
1. Create a Joomla site with 2 languages (here I assume EN and NL)
2. Create some Tags associated with the English language
3. Create some Tags associated with the Dutch language
4. Create some Articles associated with the English language and select some English Tags
5. Create some Articles associated with the Dutch language and select some Dutch Tags
6. In the English Home menu create a menu entry of the type `List All Tags` and set the language to English
7. In the English Home menu create a menu entry of the type `Compact List of Tagged Items` and set the language to English, select a tag used by the English articles and set `Content Type` to `Articles`
8. In the Dutch Home menu create a menu entry of the type `List All Tags` and set the language to Dutch
9. In the Dutch Home menu create a menu entry of the type `Compact List of Tagged Items` and set the language to Dutch, select a tag used by the Dutch articles and set `Content Type` to `Articles`
10. Create a front-end module of the type `Tags Popular` and set the language to English and assign it to a menu item that is set in step 7
11. Create a front-end module of the type `Tags Popular` and set the language to Dutch and assign it to a menu item that is set in step 9
12. Go the URL that links to the menu item created in step 7
13. Click on an article shown there
14. You should see the module from step 10 on this page
15. Notice that the links in the module are not as expected but they contain `component/tags/tag`
16. Go the URL that links to the menu item created in step 9
17. Click on an article shown there
18. You should see the module from step 11 on this page
19. Notice that the links in the module are not as expected but they contain `component/tags/tag`
20. Apply the patch
21. Redo step 12 - 14
22. You should see the correct URL in the module
23. Redo step 16 - 18
24. You should see the correct URL in the module

The URLs in the module do not show the correct SEF URL but instead contain `component/tags/tag`

The URLs in the module do show the correct SEF URL according to their menu item

None


wilsonge:
LGTM overall.

In terms of the b/c. what's the change of `*` compared to the query not existing before (i.e. if people in the JLayout don't update their template override is there a behaviour change?)

roland-d:
@wilsonge There is no b/c break here because the default value mimics the old behavior. So by not specifying the language the function works like it currently does.

Quy:
Fixes #31530 #31754 ?

roland-d:
It sounds like it, they should test this

components/com_tags/src/Helper/RouteHelper.php
```diff
@@ -81,17 +81,19 @@ public static function getItemRoute($contentItemId, $contentItemAlias, $contentC
 	/**
 	 * Tries to load the router for the component and calls it. Otherwise calls getRoute.
 	 *
-	 * @param   integer  $id  The ID of the tag
+	 * @param   string   $language  The language of the tag
+	 * @param   integer  $id        The ID of the tag
```
Quy:
```suggestion
	 * @param   integer  $id        The ID of the tag
	 * @param   string   $language  The language of the tag
```

wilsonge (approved):

components/com_tags/src/Helper/RouteHelper.php
```diff
@@ -81,17 +81,19 @@ public static function getItemRoute($contentItemId, $contentItemAlias, $contentC
...
 	 */
-	public static function getTagRoute($id)
+	public static function getTagRoute($id, string $language = '*')
 	{
-		$needles = array(
-			'tag'  => array((int) $id)
-		);
+		$needles = [
+			'tag'      => [(int) $id],
+			'language' => $language
```
Quy:
```suggestion
			'language' => $language,
```

modules/mod_tags_popular/tmpl/cloud.php
```diff
@@ -49,7 +49,7 @@
 		endif;
 ?>
 		<span class="tag">
-			<a class="tag-name" style="font-size: <?php echo $fontsize . 'em'; ?>" href="<?php echo Route::_(RouteHelper::getTagRoute($item->tag_id . ':' . $item->alias)); ?>">
+			<a class="tag-name" style="font-size: <?php echo $fontsize . 'em'; ?>" href="<?php echo Route::_(RouteHelper::getTagRoute($item->tag_id . ':' . $item->alias, $module->language)); ?>">
```
joomdonation: Could you explain why you are using module language for routing instead of the tag's language?

roland-d: @joomdonation I am using the module language here because it is the module being rendered and this way the tag adheres to the module language.

joomdonation: I'm unsure that is the right behavior. By default, the module display tags from all languages. So for example:

- A site has two languages English and German
- Create an instance of Popular Tags module, assign it to English language
- The module (by default) display tags from all languages. If a tag is assigned to German language, you are using English language (from the module) for routing for that German language tag and it could not be right

Isn't it better if we use the tag language instead? I don't think anywhere in Joomla we use module language for routing like you are using in this PR.

Maybe I am wrong? (haven't tested the behavior yet, just comment base on code review only)

roland-d: I was not sure either, so I will follow you and changed the code to use the tag language.

libraries/src/Helper/ModuleHelper.php
```diff
@@ -425,7 +425,7 @@ public static function getModuleList()
 		$query   = $db->getQuery(true);
 		$nowDate = Factory::getDate()->toSql();
 
-		$query->select($db->quoteName(['m.id', 'm.title', 'm.module', 'm.position', 'm.content', 'm.showtitle', 'm.params', 'mm.menuid']))
+		$query->select($db->quoteName(['m.id', 'm.title', 'm.module', 'm.position', 'm.content', 'm.showtitle', 'm.params', 'm.language', 'mm.menuid']))
```
joomdonation: The change here should be reverted.

modules/mod_tags_popular/tmpl/default.php
```diff
@@ -24,7 +24,7 @@
 	<ul>
 	<?php foreach ($list as $item) : ?>
 	<li>
-		<a href="<?php echo Route::_(RouteHelper::getTagRoute($item->tag_id . ':' . $item->alias)); ?>">
+		<a href="<?php echo Route::_(RouteHelper::getTagRoute($item->tag_id . ':' . $item->alias, $module->language)); ?>">
```
joomdonation: `$module->language` should be replaced with `$item->language`

joomdonation (commented):
There are still few minor changes needed, then we can test the PR.

modules/mod_tags_popular/tmpl/default.php
roland-d: @joomdonation Agree, changes pushed.

libraries/src/Helper/ModuleHelper.php
roland-d: Change also pushed.

modules/mod_tags_popular/tmpl/default.php
joomdonation: Thanks @roland-d . It should be all OK now. I will perform real test on tomorrow and report my test result.

joomdonation:
@roland-d I spent sometime to test this PR today and it works good for me. The only thing left is that in `RouteHelper` class, we have `getTagsRoute()` method, I think we should add support for $language parameter, too. Could you please look at it? This method is not used in the core, but I think the code should be consistent. After that, I can mark my test success for the PR. Thanks.

roland-d:
@joomdonation I have added the language as option to the `getTagsRoute()` as well. I did not look at it before as I am not using it either :)

joomdonation:
I have tested this item :white_check_mark: successfully on 4f446c2fc862c4452e7735e219b921dd85605f64<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37721">issues.joomla.org/tracker/joomla-cms/37721</a>.</sub>

joomdonation:
All good now, thanks @roland-d

hakanara:
I had the same issue for a very long time. I would like to thank to all the contributors. I have a question: Would this fix be affecting Joomla 3.10 as well in the upcoming release? Or only the 4.1?

chmst:
I have tested this item :white_check_mark: successfully on 4f446c2fc862c4452e7735e219b921dd85605f64<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37721">issues.joomla.org/tracker/joomla-cms/37721</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37721">issues.joomla.org/tracker/joomla-cms/37721</a>.</sub>

bembelimen:
> @wilsonge There is no b/c break here because the default value mimics the old behavior. So by not specifying the language the function works like it currently does.

In 8.1 this breaks. No fix for a bugfix release. In my opinion something for 5.0 because of 8.1 (see: https://onlinephp.io/c/e5394), but moving to 4.2 so you can decide.

hakanara:
I would like to ask one more time, will this fix be applied to 3.10 as well?

laoneo:
If possible, then the deprecated function should call the new function and not have duplicated code.

roland-d:
@laoneo Of course, good point. Code cleaned up.

joomdonation:
So the change is now backward-compatible. Do you want to have it in 4.1 @bembelimen ? Or should we re-test it for 4.2?

laoneo:
Back to pending as some code got changed.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37721">issues.joomla.org/tracker/joomla-cms/37721</a>.</sub>

components/com_tags/src/Helper/RouteHelper.php
```diff
@@ -81,17 +82,38 @@ public static function getItemRoute($contentItemId, $contentItemAlias, $contentC
...
+	 *
+	 * @param   integer  $id        The ID of the tag
+	 * @param   string   $language  The language of the tag
+	 *
+	 * @return  string  URL link to pass to the router
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  Exception
+	 */
+	public static function getComponentTagRoute(int $id, string $language = '*'): string
```
joomdonation: You cannot use `int` for `$id` parameter here because when we call this method, we pass value for this parameter as string (in this format `TAG_ID:TAG_ALIAS`). Force int data type here causes wrong tag URL generated:

- The right URL is http://localhost/jmt/index.php/en/list-all-tags-en/tag-2-en (tag-2-en is Tag Alias)
- The wrong URL (with the current code) is http://localhost/jmt/index.php/en/list-all-tags-en/3  (3 is ID of the tag)

components/com_tags/src/Helper/RouteHelper.php
```diff
@@ -124,13 +149,33 @@ public static function getTagRoute($id)
...
+	 *
+	 * @return  string  URL link to pass to the router
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  Exception
+	 */
+	public static function getComponentTagsRoute(string $language = '*'): string
+	{
+		$needles = [
+			'tag'      => [0],
```
joomdonation: Typo here, should be **tags**, not **tag**

joomdonation (changes requested):
There are some issues in the code need to be fixed.

components/com_tags/src/Helper/RouteHelper.php
roland-d: Thank you, fixed.

components/com_tags/src/Helper/RouteHelper.php
roland-d: Thank you, fixed.

components/com_tags/src/Helper/RouteHelper.php
```diff
@@ -81,17 +82,38 @@ public static function getItemRoute($contentItemId, $contentItemAlias, $contentC
...
+	 *
+	 * @return  string  URL link to pass to the router
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 * @throws  Exception
+	 */
+	public static function getComponentTagRoute(string $id, string $language = '*'): string
+	{
+		$needles = [
+			'tag'      => [$id],
```
joomdonation: Could you please change this to `'tag'      => [(int) $id],` like in the original code? Then all is fine.

roland-d: I changed that but also removed the type casting in the `getTagRoute()` so the `$id` is passed just as-is.

joomdonation: Ah, Yes. Luckily you did that :), otherwise, we would get issue override layouts use `getTagRoute()`

joomdonation:
I have tested this item :white_check_mark: successfully on 762ae51e21b88adaaa6e3da30e17e8b1e1fb85c5<br /><br />Works great now, thanks !<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37721">issues.joomla.org/tracker/joomla-cms/37721</a>.</sub>

joomdonation:
@chmst Could you please re-test ? It's nice to have this issue fixed.

viocassel:
I have tested this item :white_check_mark: successfully on 7891e0ece7774ae48e8b432186a07d0459d3f78e<br /><br />👍<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37721">issues.joomla.org/tracker/joomla-cms/37721</a>.</sub>

hakanara:
I have opened [this same issue](https://github.com/joomla/joomla-cms/issues/31530) 1,5 years ago. I am still looking forward to a solution for this since then. But I have to use Joomla 3.10 for one more year for some specific reasons. Please create an update for Joomla 3.10 as well since you already know how to modify the code. Thank you for everyone who have contributed for the solution.

wilsonge:
Two good tests. Let's get this in. Thankyou!

simbus82:
No one who answered @hakanara.
Multilanguage tags URL routing in Joomla 3.10 is broken like hell.
https://github.com/joomla/joomla-cms/issues/31530

Remember that Joomla 3.x today is the 75% of all Joomla installed.
image

Is there a possibility that it can be solved on Joomla 3.10?
Can these changes be adapted?

hakanara:
Please hear us and create a solution for Joomla 3.10. Thank you for mentioning this @simbus82 
I have opened this issue https://github.com/joomla/joomla-cms/issues/31530 1,5 years ago when there was no Joomla 4.

PR #37724, opened 2022-05-02
[4.2] System Information: Available Icons tab

pe7er:
Pull Request to **add an overview of** all **available Font Awesome Icons** to the System Information tab.

#### Joomla 3: Icomoon Icons
For Joomla 3 we have a nice overview of all Icomoon Icons that are in the Joomla core: https://docs.joomla.org/J3.x:Joomla_Standard_Icomoon_Fonts
I find that page very useful:
- Users can add such Icons in Articles.
- Extension Developers can use those in their Extensions without adding additional Icon Font Libraries. 

#### Joomla 4: Font Awesome Icons
I needed some default Font Awesome Icons in J4. There wasn't an overview of all Joomla 4 Font Awesome Icons. So I wrote some PHP code to display all the Joomla 4 icons. I used the supplied in the SVGs in ``/media/vendor/fontawesome-free/webfonts/`` to get all the icon glyph names.

The overview that I generated I used to create a J4 "Joomla Standard Font Awesome Fonts" documentation page: 
https://docs.joomla.org/J4.x:Joomla_Standard_Font_Awesome_Fonts
However, the documentation page in the Joomla MediaWiki does not support Font Awesome icons and most images are not displayed.

Furthermore I think that it would be useful to have an overview of those Icons in Joomla itself. So that you don't have to visit the documentation website to find some proper icons.

### Testing Instructions
After installing this PR a new "Available Icons" tab will be visible:
In System > System Information > "Available Icons" tab
icons

### Discussion
- Do other people also think that it would be useful to have an overview of those Font Awesome Icons in Joomla itself?
- Is the "System Information" a suitable place to add such an overview, as only Super Admins have access to that component?.  Or would is another place be more suitable?


brianteeman:
Personally I would just fix the wiki

administrator/components/com_admin/src/Extension/AdminComponent.php
```diff
@@ -18,6 +18,7 @@
 use Joomla\Component\Admin\Administrator\Service\HTML\Directory;
 use Joomla\Component\Admin\Administrator\Service\HTML\PhpSetting;
 use Joomla\Component\Admin\Administrator\Service\HTML\System;
+use Joomla\Component\Admin\Administrator\Service\HTML\DefaultIcons;
```
Quy, line 21:
```suggestion
use Joomla\Component\Admin\Administrator\Service\HTML\DefaultIcons;
use Joomla\Component\Admin\Administrator\Service\HTML\Directory;
use Joomla\Component\Admin\Administrator\Service\HTML\PhpSetting;
use Joomla\Component\Admin\Administrator\Service\HTML\System;
```

richard67:
As far as I know it’s a template‘s decision if it loads fontawesome. What if a 3rd party backend template doesn’t load it?

Besides this, I think the system info is not really the right place, and for me it is sufficient to have it in the documentation.

chmst:
A nice idea - but it is template related (atum and cassiopeia) not system related. 
Adding such a  list to cassiopea and atum would be useful for template developers.

pe7er:
Thank you all for your feedback. I understand that not all templates support Font Awesome and that the documentation wiki seems indeed the best place.
I'm closing this issue.
Thanks!

Issue #37380, opened 2022-03-25
[4.1] Image and Figure User experience


brianteeman:
### Steps to reproduce the issue
![image](https://user-images.githubusercontent.com/1296369/160187876-b02546f4-9999-40d1-a4ee-3f90d1ae62be.png)

This screen is confusing or doesnt do what I thought it would do.

#### Example 1
Enter an Image class AND a figure class

**Expected result.** that the image would be inserted with that image class and then wrapped in a figure element with the figure class

**Actual result.** The image is inserted with the image class only and no figure element

#### Example 2
Enter a figure class
**Expected result.** that the image would be inserted  and then wrapped in a figure element with the figure class

**Actual result.** The image is inserted with no class or figure element

## What actually happens
The image is ONLY wrapped in a figure element when there is a figure caption. So the figure class is completely ignored in all other uses.

## What should happen
Maybe the results are what a well educated person would expect. I didn't and thought the code was broken. It should be much clearer that the figure element is only created when there is a figure caption AND so there is no point adding a figure class if you dont have a figure caption.

Also when the figure element is not created then anything entered as a figure class is lost forever

_(updated)_
## Part 2
![image](https://user-images.githubusercontent.com/1296369/160190573-6bf62cfe-931a-4f01-9b3e-84954dded141.png)

Now it becomes more clear and more confusing.

When you add an intro or full image then you dont get asked for a figure class at all. As you can see you are only asked for an image class and a caption.

So based on the experience from Part 1 I expect that if I only add a caption then a figure element will wrap the image and caption. ✔️ it does

And I expect that if I only add an image class then the image will be inserted with an image element and class only. ❌😢
What actually happens is that the image is inserted without the class but it is wrapped in a figure element that is using the image class.

- The logic for why intro/full images always use figure even without a caption escapes me.
- The logic behind having both an image and figure class in media manager and only an image class in intro/full escapes me.
- The logic behind having a field where you enter an image class that is either an image or a figure class depending on the screen you are on escapes me.
- 
_updated again_
## Part 3
Because of the behaviour in **part 2** it is not possible to use many of the bootstrap image classes. There is for example no way to center a figure element using a bs image class.

Vanshitasingh:
@brianteeman What is a figure class?

drmenzelit:
> @brianteeman What is a figure class?

A css class that would be added to the figure element

brianteeman:
[...]

PR #37752, opened 2022-05-06
Media wrap images in figure

chmst:
Pull Request for Issue #37380 part1 .

### Summary of Changes
Images are always wrapped in a figure tag.

### Testing Instructions
see #37380 part 1. 

Open an article and add images from media. Add mage class, caption, caption class in all variants.

see #37380 part 1. 

very image has is wrapped i a figure.
All classes are set, 
Caption is only set when a caption is entered


chmst:
accidentally submitted

PR #37789, opened 2022-05-13
Add position event in content after list

korenevskiy:
Pull Request for Issue # .

### Summary of Changes
Added position in the lists of articles of categories, archives selected. This position is displayed after this list of articles. 

This is necessary for the subsequent withdrawal of information about this list of articles.
It will be convenient to use the list of category articles to create a form of ordering goods. Each material may have a selection field. And at the very end you can place an order form with delivery fields and contact details. And all this will be able to perform one field.

### Testing Instructions
You need to create the `onContentAfterItems($context, $item, $params)` method in any content plugin
Then in this method return additional text.
This text will be inserted into the list of articles of the Blog or other list of articles . 

```PHP
public function onContentAfterItems($context, $item, $params) : string{
	return "<h3>This insert text in list articles</h3>";
}
```
The event of a trigger is caused by this code inserted at the end of the category blog.
```PHP
$results = $app->triggerEvent('onContentAfterItems', array($this->category->extension . '.categories', &$this, &$this->params));
$afterDisplayItems = trim(implode("\n", $results));
echo $afterDisplayItems;
```
Before all positions allowed to add and process text before to  location position the list of articles.
This approach is convenient for the output of ordinary material, but not for categories and lists of articles .
But when displaying a list of articles, the list itself is important, so you need to display data before the list and after the list of articles.

https://docs.joomla.org/Plugin/Events/Content
https://docs.joomla.org/J3.x:Creating_a_content_plugin


laoneo:
Personally I think we have enough display events.

If you still think we should add these events to core, then they must be triggered in the view and the layout files (like default.php) have then to render only the output. An example can be found here https://github.com/joomla/joomla-cms/blob/4.1-dev/components/com_content/src/View/Article/HtmlView.php#L245-L253

korenevskiy:
> Personally I think we have enough display events.
> 
> If you still think we should add these events to core, then they must be triggered in the view and the layout files (like default.php) have then to render only the output. An example can be found here https://github.com/joomla/joomla-cms/blob/4.1-dev/components/com_content/src/View/Article/HtmlView.php#L245-L253

You are absolutely right. I think so too.
 But, in the already existing Joomla code, everything is different.
Exactly in half the cases it was done as you say, but in the other half of the cases it has already been done in the render file.
In all cases, I would do as you advise, but then it will be correct for the first part of the calls, but for the second part of the calls it will not be correct.
 Therefore, for the second half of the calls where the event calls are located in the render, I did the same, and added my corrections there.
 My correction does not correct the structure of the Joomla render.
My correction is only made by the call trigger.
Structural renderetions must be made by separate PR. And conduct numerous tests. Perhaps structural fixes are no longer possible to fix, except in Joomla 5.

So what should I do? I cannot violate the existing render structure.
You have indicated a link over the view of the article. But if you look at the view of the category, then everything is done incorrectly.

HLeithner:

HLeithner:
Sorry to say but this PR is conceptional wrong, we don't have events in layouts/templates if you want to change a layout then please override it.
Events are expensive so we have to be very careful using them.

HLeithner:
I got notified that my closing comment maybe sounds rude. This wasn't my intention and I apologies for this.

korenevskiy:
Why everything is being destroyed. Please understand me, I completely agree with you. I don't want to write an event call in layouts. But I couldn't do otherwise. because the other event calls are exactly there. I already wrote it above! The event call in the article is where it should be. But the category event calls are not there, this is a big mistake.

@HLeithner  
`components/com_content/tmpl/category/blog.php`
https://github.com/joomla/joomla-cms/blob/4.2-dev/components/com_content/tmpl/category/blog.php#L21-L32
**Look at this file yourself and see all the event calls. What can I do in this case?**
I created this PR code, not because I don't know Joomla well, you made a mistake, I know Joomla well enough, and I know PHP well. I had no other choice.

korenevskiy:
@nibra 
https://github.com/joomla/joomla-cms/blob/4.2-dev/components/com_content/tmpl/category/blog.php#L21-L32
Look at this file  and see all the event calls.
Why deleted this PR?

HLeithner:
> @nibra https://github.com/joomla/joomla-cms/blob/4.2-dev/components/com_content/tmpl/category/blog.php#L21-L32 Look at this file and see all the event calls. Why deleted this PR?

Because they are wrong, a layout file shouldn't call any event. The code you have linked as more then one problem and we don't want to continue to make more broken code. If you look at the event refactor discussion at https://github.com/joomla/joomla-cms/pull/36578 you may notice where the way should go with the event/message system.

And once again the Events manipulates the items (edit: ok not all of them) not the layout (adding text like your event) for this a layout override fits better and is more generic.

edit: We could add many other events for the layout stuff but that doesn't make sense in my opinion, the events we have are already too much.

PR #37794, opened 2022-05-14
Implement 'requireon' attribute to form fields for conditional requiredness

stephan-ansems:
Pull Request for a new Feature 

### Summary of Changes
Implementation for an new attribute `requireon` to the form field spec to that accepts conditional statements identical to that of the `showon` attribute. However the effect is to toggle the `required` attribute on form fields and show/hode the little star next to the label that indicates that the field is required.

#### Why is this useful?
Some argue that having to toggle the requiredness of a field is a design flaw in the form, however I believe that it is not. For example a form to specify OAuth2 settings could have a dropdown to indicate the grant type. Depending on the grant type chosen, the form can change quite drastically. Some fields are not relevant for certain grant types and can be hidden, others are no longer required. The problem with only `showon` is that you can hide these fields, but a hidden field still remains required and form validation will not pass succesfully. 

### Testing Instructions
Create a form with a listfield with two options and a field that has the new attribute, for example this content:
```xml
  <field
      name="foo"
      type="list"
      >
      <option value="1">MANDATORY</option>
      <option value="0">OPTIONAL</option>
  </field>
  <field
      name="bar"
      type="text"
      requireon="foo:1"
  />
</form>
```
Open the form and change the selection of the `foo` field and see the `bar` field become mandory or optional.

Any form field that uses the `requireon` attribute remain required or optiona depending on value of the `required` attribute.

Any fields that are conditionally required will be required when the condition evaluates to true. The logic works the same as for the `showon` attribute.

(https://docs.joomla.org/Form_field#Modal_form_field_types) can probably be documented in one fell swoop with the ShowOn property explanations.


richard67:
As this is a new feature, it has to be made for the 4.2-dev branch and not for 4.1-dev.

stephan-ansems:
I'll close this one then and move it to 4.2... I'm learning a lot here.

PR #37714, opened 2022-05-01
[4.1] Update deleted files list in script.php for upcoming 4.1.3 release and add missing form validation rule from 3.10

richard67:
Pull Request for Issue # .

### Summary of Changes

This pull request (PR) updates the lists of files and folders to be deleted on update by script.php for the upcoming 4.1.3 release - details see below - and adds file "libraries/src/Form/Rule/CssIdentifierSubstringRule.php" from the 3.10-dev branch.

The "CssIdentifierSubstringRule.php" file has been added due to a security patch with commit https://github.com/joomla/joomla-cms/commit/44b7b8977d9bdf2b408b21ed074dfe637501b915 .

In Joomla 4.x this file is not really needed because the same security fix uses the "CssIdentifierRule" from this file https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Form/Rule/CssIdentifierRule.php , which is more strict.

But it makes sense to keep that file, and so I add it with this PR. It can be used by 3rd party developers to validate form fields for **_parts_** of CSS identifiers like we had them in past (J3) in the core with the "CSS class suffix" fields, while the "CssIdentifierRule" only make sense for validating **_complete_** CSS identifiers.

In detail following changes are made on the lists of files and folders to be deleted on update by script.php:

- Move file "/administrator/components/com_users/src/Field/PrimaryauthprovidersField.php" from the "$folders" to the "$files" array. It was added by mistake to the "$folders" array with PR #37415 . Possibly my mistake with my code suggestion for that PR.
- Add missing leading slash to file "/libraries/vendor/webmozart/assert/.php_cs" - the missing "/" was my mistake from PR #37458 .

There have not been detected any files or folders to be added to the lists in script.php by my tool.

### Testing Instructions

Code review should be enough.

### Actual result BEFORE applying this Pull Request

File "libraries/src/Form/Rule/CssIdentifierSubstringRule.php" doesn't exist.

In file "administrator/components/com_admin/script.php":
- File "/administrator/components/com_users/src/Field/PrimaryauthprovidersField.php" is at the end of the "$folders" array in section "// From 4.1.1 to 4.1.2".
- File "libraries/vendor/webmozart/assert/.php_cs" - mind the missing leading "/" - is at the end of the "$files" array in section "// From 4.1.2 to 4.1.3".

### Expected result AFTER applying this Pull Request

[...]

PR #37826, opened 2022-05-18
[4.2] Fix mistakes in lists of deleted files and folders in script.php

richard67:
Pull Request for Issue # .

### Summary of Changes

With the last merge from 4.1-dev to 4.2-dev, PR #37714 was not completely merged up.

The files which had been moved from the list of folders to the list of files have been added in the list of files but not removed in the list of folders.

Furthermore, with this commit, another files was added to the list of folders and not the list of files in 4.2-dev: https://github.com/joomla/joomla-cms/commit/61c6f1fe8792fb350034f495a5112d71cfeb96c8

This PR here corrects these mistakes and fixes code style so that the last element in an array is followed by a comma, which is one of our code style rules.

In addition, the version comment "// From 4.1.1 to 4.2.0" is changed to "// From 4.1 to 4.2.0" because that section does not depend on the patch version of 4.1, it applies to any 4.1.

### Testing Instructions

Code review should be enough.

Files are in the list of folders to be deleted, and '/administrator/components/com_users/src/Field/PrimaryauthprovidersField.php' appears in both lists of files and folders.

Files are in the list of files where they should be.

None.

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -6387,6 +6387,8 @@ class JoomlaInstallerScript
 			'/administrator/components/com_users/src/Field/PrimaryauthprovidersField.php',
 			// From 4.1.2 to 4.1.3
 			'/libraries/vendor/webmozart/assert/.php_cs',
+			// From 4.1 to 4.2.0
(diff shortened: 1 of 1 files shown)


roland-d:
Thank you @richard67

PR #37458, opened 2022-04-02
[4.1] Remove composer dependencies' development artifacts from packages and from installations on updates

richard67:
Pull Request for Issue # .

### Summary of Changes

This pull request (PR) adds removal of .git folders and .php_cs files from 3rd party libraries in general and of the "tests" folder or the composer dependency "tobscure/json-api" in particular to the build script "build/build.php" so that these files and folders which are not needed on a J4 installation will not be included in the full installation or update packages.

It also adds the "libraries/vendor/webmozart/assert/.php_cs" file to the list of files to be deleted on update in administrator/components/com_admin/script.php.

### Reasons for Changes

When having updated the list of files and folders to be deleted on update with my PR #37287 before the 4.0.1 release, I had done that as usual based on a comparison of a full package built with the build script on the 4.1-dev branch at that time and the full installation package of the previous 4.0.0 release (and of course also another comparison with a 3.10-dev package, but that doesn't matter here).

The result was that the file ".php_cs" and subfolders ".git" and "tests" of the "tobscure/json-api" dependency were detected as removed and so added to script.php with PR #37287 as this file and these folders are not needed on a J4 installation.

See here

https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_admin/script.php#L6344-L6385

and here

https://github.com/joomla/joomla-cms/blob/4.1-dev/administrator/components/com_admin/script.php#L7698-L7718

It's development stuff from of this dependency and so it was right to remove it.

But now the same check for the current 4.1-dev branch and the 4.0.2 release has shown that this stuff was included again in the release packages, while the 4.1-dev nightly builds and also the 4.2-dev nightly builds and the 4.2.0-alpha1 packages do not include that.

See here a comparison of the latest 4.1-dev nightly installation package (left side) with the 4.0.2 release installation package (right side):

2022-04-02_j4 1-tobscure-json-api-files

The ".git" and "tests" subfolders are not expanded in the above image, they contain a lot of files.

With the 4.1 update packages it's the same.

The 4.2 packages are all ok, regardless if nightly or latest alpha.

So it seems to depend on the build environment if that stuff is in the package or not. I was never able to get it into a package when I made local builds with `php ./build/build.php --remote=HEAD --exclude-gzip --exclude-bzip2`, so up to now only @bembelimen could reproduce the issue.

I have now decided to do it like we did with other vendor stuff in past and added code to the build script for removing these files and folders from the build result before we pack the packages.

[...]

PR #37488, opened 2022-04-05
[4.1] Use to PHP 8.1 fixes for debug bar dependency

laoneo:
Pull Request for Issue #37274.

### Summary of Changes
Updates the PHP debugbar dependency to the commit with the latest PHP 8.1 fixes. It is basically the release 1.18.0 with the fixes.

### Testing Instructions
- Run Joomla on PHP 8.1
- Enable debugging in the global Joomla configuration
- Set error reporting to maximum in the global Joomla configuration
- Make sure the debug plugin is enabled by default (is the case when doing a new installation)

### Actual result BEFORE applying this Pull Request
The following error is shown:
_Implicit conversion from float -INF to int loses precision in_

### Expected result AFTER applying this Pull Request
No error.

Issue #37274, opened 2022-03-14
PHP 8.1 Deprecation


exstreme:
### Steps to reproduce the issue
Turn on error displaying

### System information (as much as possible)

Joomla 4.1 PHP 8.1

Deprecated
: substr(): Passing null to parameter #1 ($string) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: substr(): Passing null to parameter #1 ($string) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: substr(): Passing null to parameter #1 ($string) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: substr(): Passing null to parameter #1 ($string) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

Deprecated
: substr(): Passing null to parameter #1 ($string) of type string is deprecated in
C:\OpenServer\domains\joomla.test\libraries\vendor\maximebf\debugbar\src\DebugBar\JavascriptRenderer.php
on line
792

[...]

PR #37827, opened 2022-05-18
[4.1] Remove maximebf debugbar developer stuff from builds and on update

richard67:
Pull Request for Issue # .

### Summary of Changes

This pull request (PR) adds removal of .bowerrc and bower.json files from 3rd party libraries in general and of the "build", "demo", "docs" and "tests" folder or the composer dependency "maximebf/debugbar" in particular to the build script "build/build.php" so that these files and folders which are not needed on a J4 installation will not be included in the full installation or update packages.

Furthermore it adds the mentioned files and all files in the mentioned folders to script.php to the list of files to be removed on update and the folders to the list of folders to be removed on update.

The reason for this is that again - like it was before with another dependency and fixed with my PR #37458 - these developer files have been released with the 4.1.3 release in the installation package, while they are not in the 4.1.4-dev nightly builds and also not in packages which I build here locally on Linux.

### Testing Instructions

Code review by someone who knows the build process should be enough this time.

Ping @wilsonge .

I am not really willing to write again such detailed testing instructions like I did for my PR #37458 .

Depending on your environment (OS), there might or there might not be all these files and folders in the full installation package when you build it locally.

None of these files and folders is in the full installation package when you build it locally, as it is with the nightly builds.

None.


wilsonge (approved):

wilsonge:
//cc @bembelimen

bembelimen:
Yep, thx, I'm aware of it :+1:

laoneo:
Keep in mind, once the debug bar dependency will be reverted back to the official release, then these changes can probably be reverted. This was introduced with #37488 and the open issue is #37274.

richard67:
> Keep in mind, once the debug bar dependency will be reverted back to the official release, then these changes can probably be reverted. This was introduced with #37488 and the open issue is #37274.

@laoneo The change in build.php maybe, but the change in script.php will still be necessary if we want to clean up these files when updating from 4.1.3.

bembelimen:
Thx

Issue #37823, opened 2022-05-17
Use of PHP reserved word in function name


ProfDavidMora:
### Problem identified
This was initially discussed in the Inteliphense project in [this issue](https://github.com/bmewburn/vscode-intelephense/issues/2227).
The tool is detecting an instance of a function using the name `match` in the CMS core code. This is a PHP keyword since version 8.0 according to [the manual](https://www.php.net/manual/en/reserved.keywords.php) 

### Proposed solution
Change the name of this functions both at definition:
https://github.com/joomla/joomla-cms/blob/a8adadf982677c141686dc2b5560aabd2f944e76/libraries/src/Environment/Browser.php#L514

And the calling location (there seems to be only one place):
https://github.com/joomla/joomla-cms/blob/a8adadf982677c141686dc2b5560aabd2f944e76/libraries/src/Environment/Browser.php#L477

Some candidates for the new name:
- _match
- match_useragent
- match_features

### Open questions
I am not fully aware of all applicable coding rules (naming conventions), so I prefer to have someone with more expertise address the final code and include in the project if relevant. Does this need to be changed and what is the preferred final code for it?

wilsonge:
As far as I can tell from the PHP RFC https://wiki.php.net/rfc/match_expression_v2  

```
match was added as a keyword (reserved_non_modifiers). This means it can’t be used in the following contexts anymore:

- namespaces
- class names
- function names
- global constants

Note that it will continue to work in method names and class constants.
```

We're using it in a method name here which means everything works fine. Given that this is a public method available for extension developers to use I'm not sure that we really want to change the function name - certainly we would have to keep the original in J4 as an alias even if we did set a new name

eeshaanSA:
Hi, i would like to work on this issue, if allowed, as @wilsonge pointed out. I assume this is a good issue for beginners. Please point me out if i am wrong.

ProfDavidMora:
@wilsonge, thanks for the reference to the RFC. It indeed works on the web server, most definitely, otherwise the webapp would have been broken during testing. I will contact the guys at Intelephense again so that they tweak the code sensing and allow this particular one in the stated contexts.

Given that fact, I understand there is no **need** for code changes, but maybe it should follow the path of deprecation so that extension developers have time to move away from it and eventually it can be taken out altogether and align with the evolution of PHP. Possibly not a top priority, understandably, but also something that can be done specially having a volunteer already jumping in. Thoughts?

laoneo:
[...]

PR #37853, opened 2022-05-22
Modified function name to avoid conflict with PHP keyword.

eeshaanSA:
Pull Request for Issue  #37823 
Link for the issue: 
[https://github.com/joomla/joomla-cms/issues/37823](url)

### Summary of Changes
Changed the function name to avoid conflict with PHP keyword. Previously it was "match", but match is a PHP keyword since PHP 8.0, so changed it to "_match". 

### Open Questions:
I am unware of of the naming standards followed by the Joomla! team, so if you want me to change the name to something more 'understandable', let me know. 
Other function names that I have in mind:
- match_useragent
- match_features

nil

nil

None to my knowledge.

### DIFF
--- a/libraries/src/Environment/Browser.php
+++ b/libraries/src/Environment/Browser.php
@@ -474,7 +474,7 @@ class Browser
 	 */
 	public function __construct($userAgent = null, $accept = null)
 	{
-		$this->match($userAgent, $accept);
+		$this->_match($userAgent, $accept);
 	}
 
 	/**
@@ -511,7 +511,7 @@ class Browser
 	 *
 	 * @since   1.7.0
 	 */
-	public function match($userAgent = null, $accept = null)
+	public function _match($userAgent = null, $accept = null)
 	{
 		// Set our agent string.
 		if (\is_null($userAgent))
--- a/tests/Unit/Libraries/Cms/Environment/BrowserTest.php
+++ b/tests/Unit/Libraries/Cms/Environment/BrowserTest.php
@@ -175,7 +175,7 @@ class BrowserTest extends UnitTestCase
 	 */
 	public function testBrowserMatching($userAgent, $expectedBrowser, $expectedPlatform, $expectedMajorVersion, $expectedMobile)
 	{
-		$this->browser->match($userAgent);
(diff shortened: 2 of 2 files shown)


richard67:
@eeshaanSA It seems you either have missed or not understood this comment in the issue: https://github.com/joomla/joomla-cms/issues/37823#issuecomment-1129313947 . It says that there is no need for this change here, but if you change it you have to keep the method with the old name and let it can a method with a new name, and the old method is to be deprecated and can be removed not before version 5.

So this PR here is wrong as it is now.

richard67:
P.S.: See also this doc about Joomla‘s semantic versioning https://developer.joomla.org/roadmap.html .

eeshaanSA:
Sorry i missed the comment. So there are no changes required as of now?

richard67:
@eeshaanSA There are no changes REQUIRED. But what still CAN be done is to add a new method with a good name, let the old method with the „bad“ name call the new one so the old one is just a proxy to the new one, and add a deprecation note to the documentation block of the old method.

richard67:
So either close this PR or change it as requested in the above comment.

eeshaanSA:
understood @richard67 . will try to work on the change.

eeshaanSA:
@richard67 i would like some help on the deprecation note. along with the deprecated tag and the version, is there anything else I am supposed to mention in the documentation block of the method?

richard67:
> @richard67 i would like some help on the deprecation note. along with the deprecated tag and the version, is there anything else I am supposed to mention in the documentation block of the method?

@eeshaanSA No, there isn't anything else. Just search for sting "@deprecated" in PHP files and you will find examples. I suggest you change this PR, then we (me or other maintainers or contributors) can suggest changes if necessary.

ditsuke:
@eeshaanSA

See this for how to log the deprecation:
https://github.com/joomla/joomla-cms/blob/360409f51cd81de6ca302ee3a51e718f26d3fa7b/libraries/src/Application/CMSApplication.php#L422-L426

eeshaanSA:
sorry, commited the same code by mistake. Deleted the whole PR. will submit a new one shortly and provide a link to it.

PR #37865, opened 2022-05-23
[4.2] Improving parallelization of Drone jobs

Hackwar:
This further improves the Drone build process by running jobs in parallel and fixing the artifacts path.


HLeithner:
Thanks

Issue #37806, opened 2022-05-16
[4] scheduler  web cron not working


alikon:
### Steps to reproduce the issue

go to system -> scheduled task -> web cron   enable it and save

### Expected result

![image](https://user-images.githubusercontent.com/181681/168556725-b7945ede-790d-46f2-8675-0b9a687c504c.png)

### Actual result
![image](https://user-images.githubusercontent.com/181681/168553348-90a75cde-c04d-4954-9376-959b43f219df.png)

brianteeman:
Confirmed

rgblogs:
How to solve this?

joomdonation:
If I understand the issue correctly, PR https://github.com/joomla/joomla-cms/pull/37868 should fix it. Please help testing. Thanks !

PR #37387, opened 2022-03-27
[4.x] Name attribute in xml - code style

brianteeman:
All the `<name>` elements in the xml files use lowercase except for these two.

Doesn't break anything having these in a different case but it does look odd when you look at the #__extensions table

Code review should be enough

PR #37868, opened 2022-05-23
[4.1] Fix generate web cron key

joomdonation:
Pull Request for Issue #37806.

### Summary of Changes
This PR fixes a small typo in the code cause web cron key not being generated when Web Cron enabled

### Testing Instructions
- See https://github.com/joomla/joomla-cms/issues/37806 , confirm the issue
- Apply patch, confirm the issue fixed.

Please note that I have zero experience with web cron feature, just try to fix this issue by reading the code quickly.

### DIFF
--- a/plugins/system/schedulerunner/schedulerunner.php
+++ b/plugins/system/schedulerunner/schedulerunner.php
@@ -374,8 +374,7 @@ class PlgSystemSchedulerunner extends CMSPlugin implements SubscriberInterface
 		/** @var Extension $table */
 		[$context, $table] = $event->getArguments();
 
-		if ($context !== 'com_config.component'
-			|| ($table->name ?? '') !== 'com_scheduler')
+		if ($context !== 'com_config.component' || $table->name !== 'com_scheduler')
 		{
 			return;
 		}
--- a/plugins/system/schedulerunner/schedulerunner.php
+++ b/plugins/system/schedulerunner/schedulerunner.php
@@ -375,7 +375,7 @@ class PlgSystemSchedulerunner extends CMSPlugin implements SubscriberInterface
 		[$context, $table] = $event->getArguments();
 
 		if ($context !== 'com_config.component'
-			|| ($table->name ?? '') !== 'COM_SCHEDULER')
+			|| ($table->name ?? '') !== 'com_scheduler')
 		{
 			return;
 		}


plugins/system/schedulerunner/schedulerunner.php
```diff
@@ -375,7 +375,7 @@ public function generateWebcronKey(EventInterface $event): void
 		[$context, $table] = $event->getArguments();
 
 		if ($context !== 'com_config.component'
-			|| ($table->name ?? '') !== 'COM_SCHEDULER')
+			|| ($table->name ?? '') !== 'com_scheduler')
```
laoneo: This can also be moved to one line, or?

joomdonation: You mean change it to, or ?

```php
if ($context !== 'com_config.component' || ($table->name ?? '') !== 'com_scheduler')
```

brianteeman:
Strange that this is the fix as the code has always been this way and it obviously worked previously. I wonder what changed (elsewhere) to break this. Whatever it was, if it broke this then we should treat that as a red flag that it could have broken code elsewhere.

plugins/system/schedulerunner/schedulerunner.php
laoneo: Yes

joomdonation:
@brianteeman I don't have any clue about it because I haven't looked at the code of schedule tasks feature before. From what I see (by just code reading), `$table->name` will return name of the component stored in #__extensions table . By looking at the table, it is **com_scheduler**, so compare it with **COM_SCHEDULER** is wrong and that's the reason the code below that is not executed. So I just fix this small typo here, tested and it solved the issue.

plugins/system/schedulerunner/schedulerunner.php
```diff
@@ -374,8 +374,7 @@ public function generateWebcronKey(EventInterface $event): void
 		/** @var Extension $table */
 		[$context, $table] = $event->getArguments();
 
-		if ($context !== 'com_config.component'
-			|| ($table->name ?? '') !== 'COM_SCHEDULER')
+		if ($context !== 'com_config.component' || $table->name !== 'com_scheduler')
```
wilsonge, line 377: Can we do a strtolower on this please - covers both cases with minimal overhead. If this used to work - it's safer

joomdonation: @wilsonge We can but is it really needed? The install and update SQL, from my quick check, is just **com_scheduler**. Name of all of our components in extensions table is also lower case, so I don't see the need for that. Strange that this feature worked before.

toivo:
I have tested this item :white_check_mark: successfully on b09ee6b0c776b43a8f2d284d098bcce34fe4113f<br /><br />Tested successfully in Joommla 4.1.4-rc2-dev of 23 May in Wampserver 3.2.8 using PHP 8.0.15<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37868">issues.joomla.org/tracker/joomla-cms/37868</a>.</sub>

plugins/system/schedulerunner/schedulerunner.php
richard67: It might depend on the collation if a string comparison in MySQl is case sensitive or not. See e.g. https://stackoverflow.com/questions/5629111/how-can-i-make-sql-case-sensitive-string-comparison-on-mysql or https://stackoverflow.com/questions/5629111/how-can-i-make-sql-case-sensitive-string-comparison-on-mysql . So for people using the Joomla standard "utf8mb4_unicode_ci" it might have worked, and for people using some collation ending with "_cs" it moght not have worked.

richard67: I'm not sure with my above comment, it's just an idea.

joomdonation: Here we just compare data which is read from database, not from an SQL query, so I think it is not related.

richard67: Right, silly me. The comparison happens in PHP and not in a query in SQL. Seems I was blind.

brianteeman:
~Just did a quick test on an existing 4.1.3 site where the webcron key is not generated. I applied the patch from here. No change. Still cant generate the key.~ my mistake

joomdonation:
Hmm. The patch worked for me for both 4.1 and 4.2, but for my local site, it is a new installation, not updating from old version of Joomla.

@brianteeman Could you please changed that line of code to the code below, then test it again to see if it works?

```php
if ($context !== 'com_config.component' || strtolower($table->name) !== 'com_scheduler')
```

brianteeman:
I have tested this item :white_check_mark: successfully on b09ee6b0c776b43a8f2d284d098bcce34fe4113f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37868">issues.joomla.org/tracker/joomla-cms/37868</a>.</sub>

joomdonation:
RTC. Thanks all !<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37868">issues.joomla.org/tracker/joomla-cms/37868</a>.</sub>

heelc29:
> Strange that this is the fix as the code has always been this way and it obviously worked previously. I wonder what changed (elsewhere) to break this. Whatever it was, if it broke this then we should treat that as a red flag that it could have broken code elsewhere.

@brianteeman It broke with your PR #37387

brianteeman:
@heelc29 and now we know why - thanks

bembelimen:
Thx

stefanwillig:
confirmed, it works! Thank you!

PR #32980, opened 2021-04-03
[4.0] Module Custom: remove inline style

Fedik:
### Summary of Changes

this changes how background added to the Custom module, to avoid conflict with CSP rules.

### Testing Instructions
Apply patch.
Click "Enable" in CSP component.
Create a custom module with background, and some content.
Check the module rendering.

### Actual result BEFORE applying this Pull Request
Background not visible,
in browser console CSP related error.

### Expected result AFTER applying this Pull Request
Background visible, 
in browser console no CSP related error.

### Documentation Changes Required
none

PR #37872, opened 2022-05-23
[4] Cassipopeia. Banner image and nonces. Move Inline Css to HEAD

ReLater:
### Summary of Changes
- Adapt module layout code like found in layout `default.php` of `mod_custom`.

### Testing Instructions
- Current Firefox. Should be no difference with other browsers(?).
- After installation of Joomla 4.1.3 you'll see a banner image in frontend. See image 1 below.
  - It's inserted by a site module named "image" of type `mod_custom` that uses layout `Cassiopeia:banner.php`.
- Activate plugin `System - HTTP Headers` and configure it.
  - Activate `Content-Security-Policy (CSP)`
  - Activate `Nonce`. Nothing else.
  - Add a `style-src` Policy Directive: `{nonce} 'self' 'unsafe-inline'`

- Go to frontend. Banner image is gone. See image 2 below.

- Apply patch.
- Image comes back. See image 1 below.

**Image 1**

grafik

**Image 2**

grafik

### DIFF
--- a/templates/cassiopeia/html/mod_custom/banner.php
+++ b/templates/cassiopeia/html/mod_custom/banner.php
@@ -8,10 +8,23 @@
  */
 
 defined('_JEXEC') or die;
+
+use Joomla\CMS\HTML\HTMLHelper;
+use Joomla\CMS\Uri\Uri;
+
+$modId = 'mod-custom' . $module->id;
+
+if ($params->get('backgroundimage'))
+{
+	/** @var Joomla\CMS\WebAsset\WebAssetManager $wa */
+	$wa = $app->getDocument()->getWebAssetManager();
+	$wa->addInlineStyle('
+#' . $modId . '{background-image: url("' . Uri::root(true) . '/' . HTMLHelper::_('cleanImageURL', $params->get('backgroundimage'))->url . '");}
+', ['name' => $modId]);
+}
 ?>
 
-
-<div class="mod-custom custom banner-overlay" <?php if ($params->get('backgroundimage')) : ?> style="background-image:url(<?php echo $params->get('backgroundimage'); ?>)"<?php endif; ?> >
+<div class="mod-custom custom banner-overlay" id="<?php echo $modId; ?>">
 	<div class="overlay">
 		<?php echo $module->content; ?>
 	</div>


ChristineWk:
I have tested this item :white_check_mark: successfully on 35d16b749da86ea7760ae6fff8e5c29932dacd77<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37872">issues.joomla.org/tracker/joomla-cms/37872</a>.</sub>

wilsonge (approved):

dgrammatiko:
@ReLater although your approach seems to solve to problem, reality is that this approach is not gonna scale. Eg try editing the module in the backend and in the tinyMCE try to set the text alignment to right for the paragraph (it will be again broken). The solution obviously cannot be adding inline css declaration all around the layouts, this is a hack!

brianteeman:
@dgrammatiko in general you are correct. However in this specific case you are not. The css here is not user created but hardcoded into the template layout

dgrammatiko:
> However in this specific case you are not.

Actually the current implementation is outdated. It can be done using an actual image tag instead of css background-image, ie: https://nystudio107.com/blog/the-css-background-image-property-as-an-anti-pattern

ReLater:
It was absolutely clear for me that there will not come a comment on the nonce issues over days, but when a pr is provided and I've wasted my time.

@dgrammatiko Can you tell me why you've tested the same solution successfully here? https://github.com/joomla/joomla-cms/pull/32980

Kiss my ass! @dgrammatiko and Joomla! I think you don't need people like me. Keep the Joomla bugs and block security settings until one of your godfathers provides complicated solutions that nobody can follow and are complete bullshit behind the scene.

HLeithner:
I'm reopen this pull request, because I would accept it because it solves the problem in the first place. Improvements are always welcome.

@ReLater please be less rude in the future, you know we are all volunteers. Also you should know that maintainers are in charge for decline or approving pull requests. So it doesn't make sense to blame dimitries or brian for something they are not in the position to do. I hope we get this merged into 4.1 branch soon and get released with 4.1.5 (for 4.1.4 it's too late this will be released in a couple of hours).

brianteeman:
hey- what did I do? I said dmitris was wrong

HLeithner:
> hey- what did I do? I said dmitris was wrong

sorry it wasn't directly related on your comment it was only because you also joined the discussion.

jwaisner:
I have tested this item :white_check_mark: successfully on 35d16b749da86ea7760ae6fff8e5c29932dacd77<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37872">issues.joomla.org/tracker/joomla-cms/37872</a>.</sub>

jwaisner:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37872">issues.joomla.org/tracker/joomla-cms/37872</a>.</sub>

Fedik:
I have redo my fix for mod_custom to work without ID, also I include the banner.php layout 
Please test #37976

bembelimen:
I'm aware, that the IDs are not correct in modules (not only mod_custom but also in the chromes), but because of B/C issues, I will merge this and it can then improved in minor/mayor releases.

bembelimen:
Thx

PR #37888, opened 2022-05-25
[4.1] Do not throw a warning when the image url is null on clean

laoneo:
### Summary of Changes
On PHP 8.1 a warning is displayed when the image to clean is null. This happens on empty media form fields.

### Testing Instructions
- Open the back end article form
- Open the source of the HTML page
- Search for deprecated

A deprecated message is shown for images.

No deprecated message.


alikon:
did you mean  this ?
image

laoneo:
Nope, there should be a Notice in the source itself around the subform

joomdonation:
I can see this notice from PHP error logs. It comes from from our media form field. I wonder if we should fix the code which call this method instead? Shouldn't we  only call the method if the url is not empty ?

alikon:
i've replicated the issue  opening System -> Global Configuration  cause of offline image

laoneo:
> I can see this notice from PHP error logs. It comes from from our media form field. I wonder if we should fix the code which call this method instead? Shouldn't we only call the method if the url is not empty ?

What would be the advantage?

joomdonation:
> What would be the advantage?

Better code, save the system from having to execute unnecessary commands. It is logical that before displaying image, we check that the image is available before running more code to prepare data to display that image.

laoneo:
Then make your own pr with your better code. I prefer to fix it on the source and make a stable library instead of telling everyone "please do the null check" when using this function.

libraries/src/HTML/HTMLHelper.php
```diff
@@ -654,6 +654,11 @@ public static function cleanImageURL($url)
 			'height' => 0,
 		];
 
+		if ($url === null)
+		{
+			$url = '';
```
richard67, line 659:
```suggestion
			$obj->url = '';

			return $obj;
```

Maybe this would be better? Just an idea, I don't insist on it.

laoneo: Then we have duplicated code, but made it lighter which avoids the strpos

richard67: Hmm with your latest change the url property of the returned object will not be an empty string but null. Is that intended?

laoneo: Yes sure

laoneo: Reverted it

bembelimen:
As this PR fixes it confirmed, I will merge it.

bembelimen:
Thx

Issue #37801, opened 2022-05-15
[4.1.3] PHP8.1.5: Warnings on Updater.php


ghost:
### Steps to reproduce the issue

1. Use PHP8.1.5
2. Login screen at .../administrator/index.php
3. this 2 informations are shown (Error Reporting is 'Maximum')

```
Warning
: Trying to access array offset on value of type null in
.../libraries/src/Updater/Updater.php
on line
347


Deprecated
: version_compare(): Passing null to parameter #2 ($version2) of type string is deprecated in
.../libraries/src/Updater/Updater.php
on line
347
```

### Expected result
No Warnings are shown.

### Actual result
See above.

### System information (as much as possible)
![image](https://user-images.githubusercontent.com/89642616/168476544-01fa1a71-4b56-4741-9090-6aa181dc2036.png)

### Additional comments
On PHP 7.4.29 this warnings  aren't shown.

eeshaanSA:
where are these informations shown? in the php log, error log or the warings section?

ghost:
> Login screen at .../administrator/index.php
<img width="692" alt="Untitled" src="https://user-images.githubusercontent.com/89642616/170853890-65196741-1d69-47d0-9f18-b5fe26bbfbff.png">

brianteeman:
Unable to replicate in 4.2

ghost:
Closed as issue resolved. Thanks for looking at this @brianteeman

PR #37894, opened 2022-05-25
Modify Function name to avoid conflict with PHP keyword. 

eeshaanSA:
Pull Request for Issue  #37823 

### Summary of Changes
Added a new function '_match' with contents same as the original function. The 'match' function is now just a proxy to the '_match' function, and will be deprecated before version 5.0

### Testing Instructions

-

-

-

Nothing i am aware of. please let me know if any.

### DIFF
--- a/libraries/src/Environment/Browser.php
+++ b/libraries/src/Environment/Browser.php
@@ -474,7 +474,7 @@ class Browser
 	 */
 	public function __construct($userAgent = null, $accept = null)
 	{
-		$this->match($userAgent, $accept);
+		$this->_match($userAgent, $accept);
 	}
 
 	/**
@@ -511,7 +511,7 @@ class Browser
 	 *
 	 * @since   1.7.0
 	 */
-	public function match($userAgent = null, $accept = null)
+	public function _match($userAgent = null, $accept = null)
 	{
 		// Set our agent string.
 		if (\is_null($userAgent))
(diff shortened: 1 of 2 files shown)


eeshaanSA:
I still haven't logged the deprecation because i am still a little confused about that. Now that the change is pushed, @richard67 you can suggest any more required changes.

richard67:
> I still haven't logged the deprecation because i am still a little confused about that. Now that the change is pushed, @richard67 you can suggest any more required changes.

@eeshaanSA It wouldn't really be good for your learning curve if I would suggest the complete code. One thing I see is that the new method does not have a doc block. As this PR is not really necessary I think you should not waste much time with it and maybe focus on some other issue.

eeshaanSA:
i am working on the doc block, although what do you suggest? Should i close this PR? OR try working on the doc block and deprecation log?

richard67:
> i am working on the doc block, although what do you suggest? Should i close this PR? OR try working on the doc block and deprecation log?

Depends on if you want to spend time with it. I would be ok with both.

laoneo:
If I can make a suggestion. I would leave the name as it is as there is no need to make a new one. There are many other issues where we need your help. For example this one https://github.com/joomla/joomla-cms/issues/37801 is a good one for starters.

eeshaanSA:
sure @laoneo, will explore that issue.

laoneo:
@eeshaanSA I'm closing this pr for now. Thanks for your contribution on making Joomla better and would love to see more pr's from you.

PR #37898, opened 2022-05-25
Update MailTemplate.php

rigin:
The method does not bind the template to the extension. And I propose to correct this omission.

Pull Request for Issue # .

### Summary of Changes
When using this method, the 'extension' field remains empty and, accordingly, the template is not visible in the list of templates.
I added a parameter to insert it.If it is not used, then the 'extension' field is filled with the contents of the 'name' field before the '.' separator.


libraries/src/Mail/MailTemplate.php
```diff
@@ -411,17 +411,19 @@ public static function getTemplate($key, $language)
...
 	 * @since   4.0.0
 	 */
-	public static function createTemplate($key, $subject, $body, $tags, $htmlbody = '')
+	public static function createTemplate($key, $subject, $body, $tags, $htmlbody = '',$extension = null)
 	{
 		$db = Factory::getDbo();
 
 		$template = new \stdClass;
 		$template->template_id = $key;
+		$template->extension = $extension??(explode('.', $key)[0]);
```
richard67:
```suggestion
		$template->extension = $extension ?? (explode('.', $key)[0]);
```
Fix failing PHP code style checks in drone ci.

libraries/src/Mail/MailTemplate.php
```diff
@@ -411,17 +411,19 @@ public static function getTemplate($key, $language)
...
 	 * @param   array   $tags      Associative array of tags to replace
 	 * @param   string  $htmlbody  A default htmlbody (normally a translatable string)
+	 * @param   string  $extension The extension to which the template belongs
 	 *
 	 * @return  boolean  True on success, false on failure
 	 *
 	 * @since   4.0.0
 	 */
-	public static function createTemplate($key, $subject, $body, $tags, $htmlbody = '')
+	public static function createTemplate($key, $subject, $body, $tags, $htmlbody = '',$extension = null)
```
Quy:
```suggestion
	public static function createTemplate($key, $subject, $body, $tags, $htmlbody = '', $extension = null)
```

HLeithner:
Adding a parameter is a b/c break and can't go into a minor version.

brianteeman:
> Adding a parameter is a b/c break and can't go into a minor version.

even when added at the end and with a default value?

HLeithner:

Hackwar (approved):

Fedik:
To keep backward compatibility we should extract extension from the `$key` for now.
Adding extra parameter can be done in 5.x version, or we should review how all this thing works.

Kaushik1216 (approved):

ReLater:
@rigin
Because we have another pr (https://github.com/joomla/joomla-cms/pull/39540) of @Kaushik1216 just the question (no pushing!) if you will continue your pr here; whenever you'll have the time. Thank you!

Quy:
Closing in favor of #39540. Thank you.

Issue #37909, opened 2022-05-27
Installation folder cannot be deleted on Windows


chmst:
### Steps to reproduce the issue
This occurs when installing a full package of J4.1 on Win11, xampp (installed on device D:, not on C:) , php Version 8.1.5.
Same with package 3.10.x and 4.1.3 and 4.1.4. 

Installation runs without errors, but at the end appears the messaga "cannot delete the installation folder, please do that manually".
Both buttons for opening site or backend are disabled.

The installation folder is empty (no hidden or protected files in the folder), can be deleted and all works fine. 

Update by @richard67 ; It happens also with developer versions, e.g. nightly builds, where you have a button for removing the installation folder. If you use that button, the installation folder cannot be deleted and is empty. It happens only on Windows, and only with PHP 8.x. See my comments with my findings below.

### Expected result
Installation runs without errors, the Installation folder is deleted automatically.

### Actual result
@joomdonation got the same error. with PHP Warning  rmdir(D:\...): Directory not empty.

### System information (as much as possible)
As the directory IS empty  - maybe something the combination of win11 with php 8.1.5 causes this error?

richard67:
I can reproduce the issue with latest XAMPP 8.1.6 with PHP 8.1.6 on Windows 10, but not on my domain's shared hosting with Linux and the same PHP version.

richard67:
I think I can rule out file system permission issues which you could have with a standard setup of XAMPP in some cases, depending on your user status (administrator or not) and your UAC settings. I have checked it also with the actual Windows user having full access to the "d:\xampp" folder and all its subfolders (including the one there the Joomla installation is) and files inheriting that.

As @joomdonation found out, you get a PHP warning about the installation folder not being empty when removing the error output suppressing "@" from the "rmdir" statement here: https://github.com/joomla/joomla-cms/blob/4.1-dev/libraries/src/Filesystem/Folder.php#L376 .

But the folder is definitely empty, there are also no hidden files or no protected system files which are hidden by default but could be seen with the right folder options.

eeshaanSA:
I myself have faced the issue while installing Joomla! - CMS on my local system. My Xampp is installed in the C:\ Drive. 
Installation runs without errors, but at the end a pop-up appears: "cannot delete the installation folder, please do that manually".

Both buttons for opening site or backend are unresponsive. Reloading the page does no good either. 

The installation folder is empty, and deleting it manually makes everything work fine. 

XAMPP 8.1.6
Winn 11
[...]

PR #37938, opened 2022-05-31
[4.x] Fix installation JS for failed folder delete

richard67:
Pull Request for Issue #37909 (UX part) .

### Summary of Changes

Let the installation JS disable fields which cannot be used anymore when there was an error with deleting the installation folder and make sure that the buttons to go to the admin or the site still work in this case.

#### Test 1 without the patch applied to reproduce the issue

This test should be made on Windows because on Linux file system permissions are fixed by the PHP code with "chmod", and Linux still can handle open file system handles of deleted files and folders, so it's hard to cause the error with deleting the installation folder.

1. Make a new Joomla 4.1.x installation and proceed untill the last step where you can see the buttons to to go admin or site.

2. If you can not reproduce issue #37909 or don't know if you can:
Open a Windows Command Shell (CMD) Window and there change current directory to the installation folder, e.g. `cd /d "D:\xampp\htdocs\joomla-41\installation"`.
Leave that CMD Window open during the following steps.
Alternatively you could set the permissions of some other file than the index.php, e.g. the "localise.xml" file, so that it cannot be modified or deleted anymore.
The result would be the same in both cases:
There is an error when deleting the installation folder, but the index.php in that folder and other necessary files to run the installer are deleted.

3. Depends on if you are using a stable or a development version.
- Stable version:
Use one of the buttons to go to admin or site.
Result: See 1st screenshot in section "Actual result BEFORE applying this Pull Request" below.
- Development version (e.g. nightly build or local git clone):
Use the button to delete the installation folder and then confirm in the confirmation dialogue.
Result: See 2nd screenshot in section "Actual result BEFORE applying this Pull Request" below.

4. Only if you are using a development version:
Repeat step 3, i.e. use the button and confirm again.
Result: See 3rd screenshot in section "Actual result BEFORE applying this Pull Request" below.

5. Use the button to install additional languages, select some language(s) and use the "Next" button to proceed with installing the language(s).
Result: 404 not found.

#### Test 2 with the patch applied to test the fix

This test should be made on Windows for the same reasons as stated above for test 1.

1. Apply the patch of this PR. Since the JS of the installer is not handled by the build.js, the patch can be applied with Patchtester.

2. Repeat steps 1 to 3 of the previous test "Test 1 without the patch applied to reproduce the issue".
Result: Depends on if you are using a stable or a development version.
- Stable version:
See 1st screenshot in section "Expected result AFTER applying this Pull Request" below.
- Development version (e.g. nightly build or local git clone):
See 2nd screenshot in section "Expected result AFTER applying this Pull Request" below.

3. Use one of the buttons to go to admin or site.
Result: Button works.

4. Repeat steps 2 and 3 but this time use the other button in step 3.
Result: Button works.

#### Test 3 with the patch applied to test if everything else still works as before

This test should be made on an environment where you do not run into issue #37909 , i.e. not on Windows with PHP 8.1.

1. Apply the patch of this PR. Since the JS of the installer is not handled by the build.js, the patch can be applied with Patchtester.

2. Make a new Joomla 4.1.x installation and proceed untill the last step where you can see the buttons to to go admin or site.

3. Use one of the buttons to go to admin or site.
Result: Button works.

4. Repeat steps 2 and 3 but this time use the other button in step 3.
Result: Button works.

5. Repeat step 2.

6. Check everything works with installing additional languages, e.g. the "Skip" and the "Next" button both work.

#### Stable version

When using a stable version, the installation folder should be automatically deleted when using one of the 2 buttons to go to admin or site.

When deleting the folder has failed, you see the right error alert, and the 2 buttons buttons to go to admin or site are disabled.

The button to install additional languages is not disabled.

j41-installer-stable-before_1

You can start to install additional languages, but as soon as you have selected a language and use the "Next" button, you get a 404 error.

The only way out is to manually enter the right URL in the browser's URL field to go to the admin or site. If you don't know that you might think the installation is broken somehow.

#### Development version

When using a development version, there is an extra button to delete the installation folder.

When using that button and deleting the installation folder fails, you see the right error alert, and all buttons are still enabled.

j41-installer-development-before_1

When using the same button again, you see an error alert about fetching JSON data failed:

j41-installer-development-before_2

With installing languages you will have the same issue as described above for the stable version.

The only way out is to manually enter the right URL in the browser's URL field to go to the admin or site. If you don't know that you might think the installation is broken somehow.

When deleting the installation folder has failed, the button to install additional languages is disabled, and the buttons to go to admin or site just are still enabled.

When using a development version, the button to delete the installation folder is also disabled in this case.

Stable version:
j41-installer-stable-after_1

Development version:
j41-installer-development-after_1

The buttons to go to admin or site work because they just redirect to the right place and not try to call any installation stuff again which does not exist anymore.

None.


richard67:
@brianteeman Should we extend the message in the warning alert by a hint that you can use the buttons to go to the admin or the site after having deleted the folder manually?

brianteeman:
I dont think this is correct.

The error message could occur if the folder was not removed and it really wasnt empty. In that case you wouldnt want someone to be able to click on the two buttons to continue.

richard67:
> I dont think this is correct.
> 
> The error message could occur if the folder was not removed and it really wasnt empty. In that case you wouldnt want someone to be able to click on the two buttons to continue.

As soon as something could have been removed, we do not really know if the installation folder is complete or not. So we can not trust in the installation still working, and the only way to continue are these 2 buttons which now will work.

The user has to delete the folder manually so or so, it doesn't make a difference.

richard67:
P.S.: Right now without the PR you already can go to the admin and site via the right URL without having deleted the installation folder manually after an error has happened with the automatic deletion. So this PR here does not really change that. It is still on the user to manually delete the folder in that case.

chmst:
I have tested this item :white_check_mark: successfully on 8f08d1374114f604984debdaa9fea574a9a9c004<br /><br />Tested on Win11, xampp, php 8.1.5. 

With Patch, the folder can be deleted and buttons for backend / Fronted are enabled. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37938">issues.joomla.org/tracker/joomla-cms/37938</a>.</sub>

viocassel:
I have tested this item :white_check_mark: successfully on 8f08d1374114f604984debdaa9fea574a9a9c004<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37938">issues.joomla.org/tracker/joomla-cms/37938</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37938">issues.joomla.org/tracker/joomla-cms/37938</a>.</sub>

bembelimen:
Thx

PR #27162, opened 2019-11-26
Use Y-m-d format for creation dates in extension XML. Adds date sorting to extension manager and localised date formats.

Bakual:
The translators decided to standardise the date format in their language packs to use the ISO 8601 date format. Thus it makes sense that core uses the same format for core language XML as well.

The build script currently only knows one creation date for all XML files, so it will affect also `administrator/manifests/files/joomla.xml`, not only language ones.

Initially this PR only changed the build script and the XMLs updated by it. I've expanded it now so it contains all core XML files.

Also in the extension manager, you can sort by date and the date is formatted according to the active language.

### Summary of Changes
Changing the format of the creation date to use eg 2019-11-26 instead of November 2019

### Testing Instructions
You can try and run the build/bump.php and see the creation date in the XML files. That script is used when a new release is built.

administrator/language/en-GB/en-GB.xml
administrator/language/en-GB/install.xml
administrator/manifests/files/joomla.xml
administrator/manifests/packages/pkg_en-GB.xml
installation/language/en-GB/en-GB.xml
language/en-GB/en-GB.xml
language/en-GB/install.xml

In the extension manager, try sorting by date.
It may be necessary to select all extensions and hit the "Refresh Cache" button so the creation dates are updated from the XMLs.
Also try with different languages and see if the date format changes.

### Expected result
Date format like 2019-11-26 (or 26.11.2019 or whatever is specific to the active language)

### Actual result
Date format like November 2019

### Documentation Changes Required
None

PR #37949, opened 2022-06-01
[4.2] Tinymce 5.10.5

brianteeman:
Update to the latest release in the 5 series.

https://www.tiny.cloud/docs/release-notes/release-notes5105/


brianteeman:
I have no idea why drone is failing. There were no problems at all on the branch this was submitted against

HLeithner:
> I have no idea why drone is failing. There were no problems at all on the branch this was submitted against

drone restarted and solved

plugins/editors/tinymce/tinymce.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <extension type="plugin" group="editors" method="upgrade">
 	<name>plg_editors_tinymce</name>
-	<version>5.10.3</version>
-	<creationDate>2005-08</creationDate>
+	<version>5.10.5</version>
+	<creationDate>2005-2021</creationDate>
```
laoneo: This here should be reverted, otherwise it would not be sortable. More information can be found in #27162.

brianteeman: This was caused by @roland-d merging in 4.2. 
The code was correct when it was submitted

roland-d:
Thank you

PR #37943, opened 2022-05-31
[4.1] added deprecation note for 5.0 signature change

SniperSister:
### Summary of Changes
Added a deprecation note to the doc block of the AES class to emphasize the upcoming signature change in 5.x

### Actual result BEFORE applying this Pull Request
Note is missing

### Expected result AFTER applying this Pull Request
Note is there

### Documentation Changes Required
-

PR #37950, opened 2022-06-01
Revert "added deprecation note for 5.0 signature change (#37943)"

laoneo:
This reverts commit e98815ce93f4329c168b81cb6600189c69c0265b as deprecations should be done on minors only.


laoneo:
@SniperSister can you make the pr again for 4.2? Thanks.

PR #37963, opened 2022-06-02
[4.x] Cassiopeia - logo/brand and navigation in one line

reilldesign:
Pull Request for Issue # .

### Summary of Changes

div container and additional class added in index.php

### Testing Instructions

call website in frontend.

logo/brand and navigation are on top of each other

Static Layout
Cassiopeia-Default-Static

Fluid Layout
Cassiopeia-Default-Fluid

using css it becomes possible to get logo/brand and navigation in one line. without css there is no difference in the display.

`.container-brand-nav {
	display: grid;
	margin-left: auto;
	margin-right: auto;
	max-width: 1320px;
	width: 100%;
	/* grid-template-columns: repeat(2, auto); */
	grid-template-columns: 15em auto;
	grid-column-gap: 1em;
	grid-template-areas:
		"brand nav";
}

body.wrapper-fluid .container-brand-nav {
	max-width: none;
}

.container-header .container-brand {
	grid-area: brand;
}

.container-header .container-nav {
	grid-area: nav;
	padding-top: 1.5em;
}

.container-header nav {
	margin: .25em auto 0 0;
}`

Static Layout
Cassiopeia-Single-Static

Fluid Layout
Cassiopeia-Single-Fluid

in the future a template parameter and the css could be added to turn this on and off in the site template backend

### DIFF
--- a/templates/cassiopeia/index.php
+++ b/templates/cassiopeia/index.php
@@ -153,8 +153,9 @@ $wa->getAsset('style', 'fontawesome')->setAttribute('rel', 'lazy-stylesheet');
 			</div>
 		<?php endif; ?>
 
+		<div class="container-brand-nav">
 		<?php if ($this->params->get('brand', 1)) : ?>
-			<div class="grid-child">
+			<div class="grid-child container-brand">
 				<div class="navbar-brand">
 					<a class="brand-logo" href="<?php echo $this->baseurl; ?>/">
 						<?php echo $logo; ?>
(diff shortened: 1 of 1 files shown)


richard67:
Why does it need this change in the core if it can be done it a user.css file for each individual taste?

we cannot make everyone happy with the core. For example I don’t like this change.

reilldesign:
The div with class="container-brand-nav" is needed to set a max-width for the container and the "container-brand" makes css styling easier.

ChristineWk:
> Why does it need this change in the core if it can be done it a user.css file for each individual taste?
> 
> we cannot make everyone happy with the core. For example I don’t like this change.

Agree.

I haven't looked at the codes. It is up to each user himself whether he/she changes something in the user.css.

If you want the main menu next to the logo:
create a custom module for the logo and normal main menu module. Both on position: menu.

chmst:
Sorry @reilldesign but I close this issue now. 
With user.css you can do any change you like in the template and in a child template you can change the index.php if you need own classes.

Issue #37969, opened 2022-06-03
Mitigate information disclosure on profile edit form


alexandreelise:
### Is your feature request related to a problem? Please describe.

When you are on edit profile page want to prevent leaking information when showing oteps backup codes.

### Describe the solution you'd like

Add a details/summary to show/hide on demand

### Additional context

I did a pull request addressing this feature

richard67:
@alexandreelise I think I have explained you already in past that it does not need to create an issue when creating a pull request.

richard67:
Closing as having a pull request. See #37970 .

alexandreelise:
Hi @richard67 I thought I needed to create an issue as in the pull request template it still requires an issue number. Maybe someone should remove this requirement in the template as it is misleading. I tried to strictly follow the rules and the procedure. Next time i'll not do an issue. Thanks for reminding Richard.

PR #37970, opened 2022-06-03
Attempt to prevent information disclosure

alexandreelise:
For example if someone is looking over your shoulder while you have backup oteps shown. This little modification tries to mitigate this possibility.
One side note. The details text should be a language string.

Pull Request for Issue #37969.

### Summary of Changes

add details/summary html5 tags

### Testing Instructions

Enable 2FA in administrator for a user in Registered group
Login in frontend with this Registered user
Go to edit profile
Scroll down to see Oteps backup codes

Oteps backup codes show immediately and might disclose information

Oteps Backup codes don't show immediately. Rather, they show only if we click on Details

Possibly.


components/com_users/tmpl/profile/edit.php
```diff
@@ -100,9 +100,12 @@
 						<?php echo Text::_('COM_USERS_PROFILE_OTEPS_WAIT_DESC'); ?>
 					</div>
 				<?php else : ?>
+					<details>
+						<summary>Details</summary>
```
brianteeman: do not hard code english. use a language string

alexandreelise: Yes you are right @brianteeman I put it in a side note as I know it is best practice to use language strings but I wasn't sure what language string to add and I the translation team had a better idea for the language string name that's why I hardcoded the English. Next time I won't do the same mistake.

richard67: That’s not a reason not to do it right here. If you don’t modify your PR so it uses a language string here, it can’t be accepted.

language/en-GB/com_users.ini
```diff
@@ -51,6 +51,7 @@ COM_USERS_PROFILE_NAME_LABEL="Name"
 COM_USERS_PROFILE_NEVER_VISITED="This is the first time you have visited this site."
 COM_USERS_PROFILE_NOCHANGE_USERNAME_DESC="If you want to change your username, please contact a site administrator."
 COM_USERS_PROFILE_OTEPS="One time emergency passwords"
+COM_USERS_PROFILE_OTEPS_DETAILS_DESC="Details"
```
wojsmol: Use COM_USERS_PROFILE_OTEPS_DETAILS_LABEL="Details"

alexandreelise: Thanks for the suggestion @wojsmol . It has been fixed

richard67:
@alexandreelise Thanks for this PR. Unfortunately I am not sure if it will make sense in 4.1 because there will be only one more 4.1.x release before 4.2.0 will come, and in 4.2. the 2FA will be completely reworked and renamed to MFA (multi-factor authentication). See PR #37912 for the ongoing work.

I haven't checked yet if that PR will bring back the thing which you want to solve with this PR here. But that might be the case, and your changes would be lost.

@brianteeman As you have checked Nik's PR already: Do you know if it already works like that, show the backup codes only on demand?

brianteeman:
if there is even just one more release then this should be included even if the replacement code in 4.2 will not use it

Of course if you use the global string of JDETAILS then there is no problem at all.

brianteeman:
As there will now not be another 4.1 release this can be closed as it is not relevant for 4.2

PR #37873, opened 2022-05-23
[4] Fix for 37859. mod_feed. inline style and nonces.

ReLater:
Fixes https://github.com/joomla/joomla-cms/issues/37859

### Testing Instructions
See opening post of issue https://github.com/joomla/joomla-cms/issues/37859

### Expected result AFTER applying this Pull Request
Feed output right aligned.

PR #37976, opened 2022-06-04
[4.2] Remove ID attribute from mod_custom

Fedik:
### Summary of Changes

It redo of #32980, but without use of ID attribute

### Testing Instructions
Select background Image for mod_custom,
And check module on the site, also the module HTML markup

Background visible, module have an ID attribute

Background visible, module do not have an ID attribute

none


brianteeman:
The changes to administrator/components/com_installer/src/Controller/InstallController.php look unrelated to me

Fedik:
That because it was for 4.1 first.
Should be good now.

simbus82:
Why the ID needs to be removed?

chmst:
@simbus82 This is why: https://github.com/joomla/joomla-cms/pull/37873#issuecomment-1146187075  - multiple IDs are possible. 

@Fedik 
Do we need an extra classname? Other modules don't.
The variablename $modid sounds like and ID but makes a classname.

Fedik:
> Do we need an extra classname? Other modules don't.

The code changes use of ID to class-[id]

simbus82:
> @simbus82 This is why: [#37873 (comment)](https://github.com/joomla/joomla-cms/pull/37873#issuecomment-1146187075) - multiple IDs are possible.

Ok, it's clear that we can't have multiple ID in a single DOM (but only for validation): in 2022 is only a problem if we try to pointing them wrongly with CSS and JS.
Removing an ID (in the core since one year) can break lot's of templates that rely on this method to apply some css styles.
I hope users will be informed about this B/C.

brianteeman:
As pointed out before by others removing the id will be an isssue for any site that is using the id for css and/or js (I am) and while it can be argued that this is not covered by the b/c promise is it really necessary?

The only way that you can have one module (with the same id) published more than once is to additionally load the module in some content with the loadmodule plugin. As that is a deliberate action by the web developer to publish the module twice on the same page when it makes no sense to do so I would expect the occurance of this to be infinitely less than the use of the id or js/css

Fedik:
I just sugested a solution to avoid duplicated ids (even if it rare possible) :)
It can be accepted or closed, I am fine with both.

HLeithner:
Mod_custom is the module used as reference for new module that should not suggest to use wrong syntax.

Fedik:
For keep b/c I can revert ID (for people who use js), but the styling (background image) will be done with class. And will remove ID in j5 (if no one forget).
Deal? :)

Mrsuperstar304:
Hey everyone, I am new here. Please tell me how to start here and what are the skills required to contribute here.<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/37976">issues.joomla.org/tracker/joomla-cms/37976</a>.</sub>

chmst:
@Mrsuperstar304 thank you for your interest in contributing to Joomla. Maybe you start with this video https://www.youtube.com/watch?v=30JYBOeNVTA. Or search for "how to contribute " in joomla documentation.

simbus82:
> Mod_custom is the module used as reference for new module that should not suggest to use wrong syntax.

And it's a wrong syntax to not have an unique ID for every instance of rendered mod_custom in the DOM 😄

Fedik:
I have restored ID, and added a comment that it should be removed in next Major version

PR #37978, opened 2022-06-04
Improved MVC events

Denitz:
### Summary of Changes

1. `events_map` property moved from `FormModel` to `BaseDatabaseModel` because the property is shared by multiple inheritors.
2. All `eventClass` params of `AbstractEvent::create()` calls are shortened to use `Classname::class` notation instead of using string event classname (easier to find all event class usages vide IDE in 2022).
3. `onAfterDisplay` event can modify the output now.
4. New `ListQueryEvent` class and `onContentListQuery` event to modify the list query before executing.
5. New `AfterGetListEvent` class and `onContentAfterGetList` event to modify the list of items after load from database.

### Testing Instructions

1. Apply patch, try to play with custom events via your plugin, i.e.

```
		// All views return 'my custom HTML'
		public function onAfterDisplay(DisplayEvent $event)
		{
			$event->setArgument('source', 'my custom HTML');
		}

		// All lists are ordered by `a.id`
		public function onContentListQuery(DisplayEvent $event)
		{
			$event->getArgument('query')->order('a.id');
		}

		// All lists' titles are 'OK'
		public function onContentAfterGetList(AfterGetListEvent $event)
		{
			foreach($event->getArgument('list') as $item)
			{
				$item->title = 'OK';
			}
		}
```

Core Joomla is barely extendable.

Core Joomla is extendable.

No.


brianteeman:
Is this backwards compatible?

libraries/src/MVC/Model/ListModel.php
```diff
@@ -216,11 +262,71 @@ protected function _getListQuery()
...
+	/**
+	 * Applies plugins for the list loaded from database.
+	 *
+	 * @param   object[]  $list  The list of item.
+	 *
+	 * @return  object[]  An array of results.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function afterGetList($list)
```
nibra: Why not private?

libraries/src/MVC/Model/ListModel.php
```diff
@@ -121,6 +124,22 @@ class ListModel extends BaseDatabaseModel implements FormFactoryAwareInterface,
 	 */
 	protected $listForbiddenList = array('select');
 
+	/**
+	 * The event to trigger on list query
+	 *
+	 * @var    string
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $event_list_query;
```
nibra: Follow naming conventions: `$eventListQuery`

libraries/src/MVC/Model/ListModel.php
```diff
@@ -121,6 +124,22 @@ class ListModel extends BaseDatabaseModel implements FormFactoryAwareInterface,
...
+
+	/**
+	 * The event to trigger after get list
+	 *
+	 * @var    string
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $event_after_get_list;
```
nibra: Follow naming conventions: `$eventAfterGetList`

libraries/src/MVC/Model/ListModel.php
```diff
@@ -157,6 +176,33 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
...
 		}
+
+		if (isset($config['event_list_query']))
+		{
+			$this->event_list_query = $config['event_list_query'];
+		}
+		elseif (empty($this->event_list_query))
+		{
+			$this->event_list_query = 'onContentListQuery';
+		}
```
nibra: Use [null coalescing operator](https://www.php.net/manual/en/language.operators.comparison.php#example-121):

```php
$this->event_list_query = $config['event_list_query'] ?? $this->event_list_query ?? 'onContentListQuery';
```

libraries/src/MVC/Model/ListModel.php
```diff
@@ -157,6 +176,33 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
...
+
+		if (isset($config['event_after_get_list']))
+		{
+			$this->event_after_get_list = $config['event_after_get_list'];
+		}
+		elseif (empty($this->event_after_get_list))
+		{
+			$this->event_after_get_list = 'onContentAfterGetList';
+		}
```
nibra: Same here

libraries/src/MVC/View/HtmlView.php
```diff
@@ -224,17 +225,17 @@ public function display($tpl = null)
...
-					'source'     => $result
+					'source'     => $result,
 				]
 			)
 		);
 
 		$eventResult->getArgument('used', false);
 
-		echo $result;
+		echo $eventResult->getArgument('source');
```
nibra: This seems to fix a bug that has been undiscovered for a long time.

administrator/components/com_installer/src/Model/InstallerModel.php
```diff
@@ -80,6 +80,10 @@ protected function _getList($query, $limitstart = 0, $limit = 0)
 			// Get results from database and translate them.
 			$db->setQuery($query);
 			$result = $db->loadObjectList();
+
+			// Apply plugins
+			$result = $this->afterGetList($result);
+
```
nibra: Are you sure that you caught all occurances?

nibra (changes requested):
Looks good so far; I don't immediately see any b/c breaks. However, this PR should be targeted at 4.2, as it introduces a new feature.

libraries/src/MVC/Model/ListModel.php
Denitz: `private` and `final` should be avoided in open source, 3rd-party developers can extend classes.

libraries/src/MVC/Model/ListModel.php
Denitz: Currently, all event names are snake-cased, it's better to keep the same style.

libraries/src/MVC/Model/ListModel.php
Denitz: Currently, all event names are snake-cased, it's better to keep the same style.

libraries/src/MVC/Model/ListModel.php
Denitz: I used the same code approach as all other core classes, better keep same.

libraries/src/MVC/Model/ListModel.php
Denitz: Same

libraries/src/MVC/View/HtmlView.php
Denitz: Show I change something?

administrator/components/com_installer/src/Model/InstallerModel.php
Denitz: Yes

Denitz:
> Is this backwards compatible?

@brianteeman Yes, it only adds the new functionality. I can add the same code in Joomla 3 but won't be accepted :)

richard67:
@Denitz New functionality should go into 4.2, if backwards compatible. So this PR should be rebased to the 4.2-dev branch. Let us know if you need help with that.

After rebase, the PR might show unrelated changes on GitHub, but that will go away after the next update of the 4.2-dev branch with the latest changes on the 4.1-dev branch (so-called "upmerge") and after that an update of your branch for this PR to the latest 4.2-dev here.

Denitz:
@richard67 Rebased

administrator/components/com_installer/src/Controller/InstallController.php
```diff
@@ -49,16 +49,17 @@ public function install()
...
-			$redirect_url = base64_decode($this->input->getBase64('return'));
+			$redirect_url = base64_decode($return);
 		}
 
 		// Don't redirect to an external URL.
-		if (!Uri::isInternal($redirect_url))
+		if ($redirect_url && !Uri::isInternal($redirect_url))
 		{
 			$redirect_url = '';
 		}
```
nibra: How is this related to this PR?

administrator/components/com_installer/src/Model/InstallerModel.php
nibra: Ok

libraries/src/MVC/Model/ListModel.php
nibra: Like `$listForbiddenList`? Nevertheless: Follow the required code style. It makes no sense to repeat failures from the past.

libraries/src/MVC/Model/ListModel.php
nibra: Same as above

libraries/src/MVC/View/HtmlView.php
nibra: `DispatcherInterface::dispatch()` does not have a return value, so `$eventResult` should be `null` in any case. However, plugins may change parameters provided by the event. Thus, `$result` has probably been changed by plugins, so `echo $result` would yield the expected result. `$eventResult` is not reliable, since it only could contain values different from `null`, if the Dispatcher violates the interface - which would be a bug.

nibra (changes requested):

libraries/src/Event/Model/AfterGetListEvent.php
```diff
@@ -0,0 +1,77 @@
...
+	 * list			\stdClass[]			Item list.
+	 *
+	 * @param   string  $name       The event name.
+	 * @param   array   $arguments  The event arguments.
+	 *
+	 * @throws  BadMethodCallException
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct($name, array $arguments = array())
```
Quy:
```suggestion
	public function __construct($name, array $arguments = [])
```

libraries/src/Event/Model/ListQueryEvent.php
```diff
@@ -0,0 +1,77 @@
...
+	 * query		QueryInterface		Database query.
+	 *
+	 * @param   string  $name       The event name.
+	 * @param   array   $arguments  The event arguments.
+	 *
+	 * @throws  BadMethodCallException
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function __construct($name, array $arguments = array())
```
Quy:
```suggestion
	public function __construct($name, array $arguments = [])
```

libraries/src/MVC/Model/ListModel.php
nibra: That old style was used in J3, because it was supposed to work with PHP5. That restriction has gone, so it can be done right now.

libraries/src/MVC/View/HtmlView.php
Denitz: Nope, `\Joomla\Event\DispatcherInterface::dispatch()` returns the event itself and it should! Even PSR-14 requires this, see https://github.com/php-fig/event-dispatcher/blob/aa4f89e91c423b516ff226c50dc83f824011c253/src/EventDispatcherInterface.php#L17

Another issue here:
`DisplayEvent` is immutable (why? the purpose of event to be able to modify the output)!
But we can change the argument via `$event->setArgument('source', $html)` but not via offset like `$event['source'] = $html;`
Is it expected behavior for immutable event?

libraries/src/MVC/Model/ListModel.php
Denitz: Done.

Denitz:
Drone fails because `xmlreader` is not installed for PHP 8.1, can somebody fix it?

libraries/src/MVC/View/HtmlView.php
nibra: You're right, I went into the trap and looked at `\Joomla\CMS\Dispatcher\DispatcherInterface`.

> But we can change the argument via `$event->setArgument('source', $html)`

That's obviously an oversight. The event is immutable. The event handler should create a new event with the desired values as a return object.

administrator/components/com_installer/src/Controller/InstallController.php
richard67: Possibly one of the unrelated changes shown after branch rebase which are caused by the 4.2-dev branch being a commit behind the 4.1-dev branch. If this is the case, it was wrong to revert that change here in this PR.

libraries/src/MVC/Model/ListModel.php
nibra: > `private` and `final` should be avoided in open source

Surely not. Exactly because 3PD can extend the classes, both should be used wherever possible.
The purpose of this method is solely to apply the plugins to make the class extensible.
If the method is `protected`, this is an invitation for 3PD to destroy this functionality.

administrator/components/com_installer/src/Controller/InstallController.php
richard67: Anyway it can be finally solved after the next upmerge from 4.1-dev to 4.2-dev and then the branch update here. We should leave this review comment open as reminder and set it to resolved when that has been done.

richard67:
> Drone fails because `xmlreader` is not installed for PHP 8.1, can somebody fix it?

Have contacted maintainers to check.

libraries/src/MVC/View/HtmlView.php
Denitz: The purpose of event is to be able to modify output, I think we should use usual mutable event (DisplayEvent extends AbstractEvent) and view will return the event argument as I've already added, please confirm.

nibra: Keep the `echo $eventResult->getArgument('source');`, but let the event alone for now. There are currently considerations to improve the event system for Joomla 5. It's early enough to address that (mutable vs. immutable) once decisions are made for the improvements.

Denitz: OK.

Denitz:
Is it possible to restart drone?

richard67:
> Is it possible to restart drone?

@Denitz I have done that recently a few times, but the tests fail all at the same step: Logging in to admin after a new installation. Have you tested this yourself, making a new installation with your PR applied, and then logging in to backend? I'll restart drone now but I do not really expect new results.

libraries/src/Event/Model/AfterGetListEvent.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class AfterGetListEvent extends AbstractEvent
+{
+	/**
+	 * Mandatory arguments:
+	 * subject		ListModelInterface	The model instance we are operating on.
+	 * context		string			    The model context.
+	 * list			\stdClass[]			Item list.
```
richard67:
```suggestion
	 * subject  ListModelInterface  The model instance we are operating on.
	 * context  string              The model context.
	 * list     \stdClass[]         Item list.
```

libraries/src/Event/Model/ListQueryEvent.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @since  __DEPLOY_VERSION__
+ */
+class ListQueryEvent extends AbstractEvent
+{
+	/**
+	 * Mandatory arguments:
+	 * subject		ListModelInterface	The model instance we are operating on.
+	 * context		string			    The model context.
+	 * query		QueryInterface		Database query.
```
richard67:
```suggestion
	 * subject  ListModelInterface  The model instance we are operating on.
	 * context  string              The model context.
	 * query    QueryInterface      Database query.
```

administrator/components/com_installer/src/Controller/InstallController.php
Denitz: 4.1-dev was merged via https://github.com/joomla/joomla-cms/commit/c297caa4c0b3368d7f89af44c2bec1b353c37619
I will try to update branch

libraries/src/Event/Model/AfterGetListEvent.php
```diff
@@ -0,0 +1,77 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  (C) 2020 Open Source Matters, Inc. <https://www.joomla.org>
```
heelc29:
```suggestion
 * @copyright  (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

libraries/src/Event/Model/ListQueryEvent.php
```diff
@@ -0,0 +1,77 @@
+<?php
+/**
+ * Joomla! Content Management System
+ *
+ * @copyright  (C) 2020 Open Source Matters, Inc. <https://www.joomla.org>
```
heelc29:
```suggestion
 * @copyright  (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
```

libraries/src/MVC/Model/ListModel.php
```diff
@@ -157,6 +176,18 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
 		{
 			$this->listForbiddenList = array_merge($this->listBlacklist, $this->listForbiddenList);
 		}
+
+		$this->eventListQuery = $config['eventListQuery'] ?? $this->eventListQuery ?? 'onContentListQuery';
+		$this->eventAfterGetList = $config['eventAfterGetList'] ?? $this->eventAfterGetList ?? 'onContentAfterGetList';
```
heelc29:
```suggestion
		$this->eventListQuery    = $config['eventListQuery'] ?? $this->eventListQuery ?? 'onContentListQuery';
		$this->eventAfterGetList = $config['eventAfterGetList'] ?? $this->eventAfterGetList ?? 'onContentAfterGetList';
```

libraries/src/MVC/Model/ListModel.php
```diff
@@ -157,6 +176,18 @@ public function __construct($config = array(), MVCFactoryInterface $factory = nu
...
+
+		$config['events_map'] = $config['events_map'] ?? array();
+
+		$this->events_map = array_merge(
+			array(
+				'list_query' => 'content',
+				'after_get_list' => 'content',
+			), $config['events_map']
```
heelc29:
```suggestion
			[
				'list_query'     => 'content',
				'after_get_list' => 'content',
			],
			$config['events_map']
```

libraries/src/MVC/Model/ListModel.php
```diff
@@ -216,11 +247,71 @@ protected function _getListQuery()
...
 			$this->query            = $this->getListQuery();
+
+			PluginHelper::importPlugin($this->events_map['list_query']);
+
+			$event = new ListQueryEvent(
+				$this->eventListQuery,
+				[
+					'subject' => $this,
+					'context' => $this->context,
+					'query' => $this->query,
```
heelc29:
```suggestion
					'subject' => $this,
					'context' => $this->context,
					'query'   => $this->query,
```

libraries/src/MVC/Model/ListModel.php
```diff
@@ -216,11 +247,71 @@ protected function _getListQuery()
...
+	protected function afterGetList($list)
+	{
+		PluginHelper::importPlugin($this->events_map['after_get_list']);
+
+		$event = new AfterGetListEvent(
+			$this->eventAfterGetList,
+			[
+				'subject' => $this,
+				'context' => $this->context,
+				'list' => $list,
```
heelc29:
```suggestion
				'subject' => $this,
				'context' => $this->context,
				'list'    => $list,
```

HLeithner:
There is a reason why we don't have a listquery event because it creates more then it solves. This has discussed several times.

Also the modification of the result is questionable. Because we easily generate more problems then expected.

libraries/src/MVC/Model/BaseDatabaseModel.php
```diff
@@ -64,6 +64,14 @@ abstract class BaseDatabaseModel extends BaseModel implements DatabaseModelInter
 	 */
 	protected $event_clean_cache = null;
 
+	/**
+	 * Maps events to plugin groups.
+	 *
+	 * @var    array
+	 * @since  __DEPLOY_VERSION__
+	 */
+	protected $events_map;
```
laoneo: Please revert this change as a class should not have properties which are not used by itself.

administrator/components/com_installer/src/Model/UpdateModel.php
```diff
@@ -235,7 +235,12 @@ protected function _getList($query, $limitstart = 0, $limit = 0)
 		{
 			$db->setQuery($query);
 			$result = $db->loadObjectList();
+
+			// Apply plugins
+			$result = $this->afterGetList($result);
```
laoneo: Would that mean that every model which extends the listmodel and does overwrite the _getList function needs to call this.

Denitz: Ideally, it should if the model wants to apply the event.

libraries/src/MVC/Model/BaseDatabaseModel.php
Denitz: Reverted.

HLeithner (changes requested):
Please remove the new Events for query manipulation, this has been discussed several times for more information look at https://github.com/joomla/joomla-cms/pull/26179

Denitz:
> Please remove the new Events for query manipulation, this has been discussed several times for more information look at #26179

Sorry and thanks for your time. The query manipulation was the main feature of this PR.

laoneo:
Would you mind to open a new pr which removes the hard codec class paths?

Denitz:
I removed onContentListQuery event, though I totally disagree.

Issue #37956, opened 2022-06-01
Joomla\CMS\Table bind() improperly binds null fields (without a reset)


aforgh:
### Steps to reproduce the issue
The issue occurs when you have more than one record to bind and save.
eg.
```
foreach($recordset as $record)
{
	$table->save($record);
}
```

If the record contains fields that admit null and that are set to null, they don't get bound properly by the bind() function.
The first record to be saved works properly, because the Table object is freshly instantiated and initialized.
During subsequent bind() operations, fields containing null values don't overwrite table properties already set to non-null.

I created a new PHPunit method in **\tests\Integration\Libraries\Cms\Table\TableTest.php**, to better explain the issue

```
/**
 * Test that bind() properly bind a null field 
 *
 * @return  void
 * @since   4.1.0
 */	public function testBindNullFieldsWithoutReset()
{
	$data_1 = [
		'title' => 'Title One',
		'checked_out' => 1
	];
	
	$data_2 = [
		'title' => 'Title Two',
		'checked_out' => null
	];
	
	// Check if nullable int property is 1
	$this->object->bind($data_1);
	$this->assertEquals(1, $this->object->checked_out);
	
	// Check if nullable int property is null
	$this->object->bind($data_2);
	$this->assertEquals(null, $this->object->checked_out);
}
```

### Expected result
When I run the test above, I expect that bind($data_2) sets 'checkout' property to null.

### Actual result
The actual result is that the 'checkout' property remains set to 1.

### System information (as much as possible)
PHP Version	8.0.19
Web Server	Apache/2.4.38 (Debian)
WebServer to PHP Interface	fpm-fcgi
Joomla! Version	Joomla! 4.1.0 Stable [ Kuamini ] 15-February-2022 09:30 GMT

### Additional comments (proposed solution)

I traced the root of the problem to line 695 of file **libraries\src\Table\Table.php**

```
	// Bind the source value, excluding the ignored fields.
	foreach ($this->getProperties() as $k => $v)
	{
		// Only process fields not in the ignore array.
		if (!\in_array($k, $ignore)) 
		{
			if (isset($src[$k])) //line 695
			{
				$this->$k = $src[$k];
			}
		}
	}
```
The `isset()` function does return false if $src\[$k\] exists, but contains null.

See:
https://www.php.net/manual/en/function.isset.php

> isset() will return false when checking a variable that has been assigned to null. Also note that a null character ("\0") is not equivalent to the PHP null constant.

[...]

PR #37983, opened 2022-06-04
Table bind() improperly binds fields

eeshaanSA:
Pull Request for Issue #37956 

### Summary of Changes
Made the changes as suggested by @aforgh. Please checkout the issue #37956. All necessary information, required code, changes made, and their effects have been mentioned neatly there. 

Checkout Property remains set to 1.

Checkout property is set to null.

--

### DIFF
--- a/libraries/src/Table/Table.php
+++ b/libraries/src/Table/Table.php
@@ -694,7 +694,7 @@ abstract class Table extends CMSObject implements TableInterface, DispatcherAwar
 			// Only process fields not in the ignore array.
 			if (!\in_array($k, $ignore))
 			{
-				if (isset($src[$k]))
+				if (array_key_exists($k, $src))
 				{
 					$this->$k = $src[$k];
 				}


eeshaanSA:
I only made the changes suggested by the owner of this issue, @aforgh. Please let me know if there anything else to this issue, except the changes made above.

chmst:
The first error messag is : Save failed with the following error: Column 'level' cannot be null. I don't see how this is related to this PR. Maybe @richard67 can help?

richard67:
> The first error messag is : Save failed with the following error: Column 'level' cannot be null. I don't see how this is related to this PR. Maybe @richard67 can help?

I think that error is related because now with this change any array element of the dev array which has a null value will be used, regardless if that column allows null values or not.

@eeshaanSA i don’t think it is a good idea to make a pull request with a change suggested by someone else without really understanding that change.

roland-d:
@eeshaanSA Thank you for your pull request but I am sorry this cannot be used. This is a change in behavior of how the function works and is not backwards compatible. Such a change can only be done in a major release if we want this change. This is the reason I am closing the PR.

eeshaanSA:
> > The first error messag is : Save failed with the following error: Column 'level' cannot be null. I don't see how this is related to this PR. Maybe @richard67 can help?
> 
> 
> 
> I think that error is related because now with this change any array element of the dev array which has a null value will be used, regardless if that column allows null values or not.
> 
> 
> 
> @eeshaanSA i don’t think it is a good idea to make a pull request with a change suggested by someone else without really understanding that change.

Sorry. Will take care from now onwards.

PR #37989, opened 2022-06-05
[4.2] Fix fatal error on update to Joomla 4.2

joomdonation:
Pull Request for Issue # .

### Summary of Changes
In Joomla 4.2, we convert some plugins to new structure, so the namespace map will need to be re-generated before avoid Joomla calls com_joomlaupdate to finalize upgrade process to avoid fatal error.

### Testing Instructions
1. Install Joomla 4.1.
2. Try to Update that 4.1 website to Joomla 4.2 nightly build. During the update, you would get fatal error saying class `Joomla\Plugin\Actionlog\Joomla\Extension\Joomla` not found (something like that, I don't remember exactly the error message)
3. Install Joomla 4.1 again
4. Try to update to the update package generated for this PR. Make sure upgrade works properly

Fatal error while upgrading to Joomla 4.2

Upgrade to Joomla 4.2 works well.


joomdonation:
Sorry, wrong branch

PR #38011, opened 2022-06-08
[4.2] Use the app identity in the FormController

laoneo:
### Summary of Changes
Use the app identity in the `FormController`.

~When the controller needs a current user, it must be set in the MVCFactory. This is the case on all Form based controllers.~

### Testing Instructions
- Log in on the back end
- Go to the articles manager
- Click on New
- Give the article a title
- Save the article

All works.

All works.


toivo:
I have tested this item :white_check_mark: successfully on cf172a9ebb05527c349ddef4ac6a84a8f6dc0ac4<br /><br />Tested successfully in Joomla 4.2.0-beta1-dev of 9 June using Wampserver 3.2.9 and PHP 8.0.15<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38011">issues.joomla.org/tracker/joomla-cms/38011</a>.</sub>

joomdonation:
In controller, we can already access to current user object via `$this->app->getIdentity();` . Could you please explain the reason for injecting it again when controller is created using MVCFactory?

laoneo:
YOu are right, changed the code to use the identity.

wilsonge (approved):

libraries/src/MVC/Controller/FormController.php
```diff
@@ -200,9 +200,14 @@ public function add()
 	 *
 	 * @since   1.6
 	 */
-	protected function allowAdd($data = array())
+	protected function allowAdd($data = [])
 	{
-		$user = Factory::getUser();
+		$user = $this->app->getIdentity();
+
+		if (!$user)
```
joomdonation: The app identity is loaded very early in our application life cycle (at initialize phase, even before we import and trigger system plugin), so $this->app->getIdentity() will always return a `User` object, thus the check here is not needed. Same for the code which you added in `allowEdit` method.

laoneo: I'm not sure if the identity is always set in the app. @wilsonge is it?

laoneo: Can confirm, removed the check. Thanks for the hint!

joomdonation:
I have tested this item :white_check_mark: successfully on 4a36bdebcce17e0968a9545bb22fc79a9ba5bba7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38011">issues.joomla.org/tracker/joomla-cms/38011</a>.</sub>

joomdonation:
@laoneo Unrelated to this PR but you made similar check here https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/MVC/Controller/BaseController.php#L598 and here https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/MVC/Controller/BaseController.php#L630 in PR #37498 . The check `$this->app->getIdentity()` in these lines could be removed, too.

laoneo:
Can you make a pr removing them?

chmst:
I have tested this item :white_check_mark: successfully on 4a36bdebcce17e0968a9545bb22fc79a9ba5bba7<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38011">issues.joomla.org/tracker/joomla-cms/38011</a>.</sub>

chmst:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38011">issues.joomla.org/tracker/joomla-cms/38011</a>.</sub>

roland-d:
Thanks everybody

PR #38022, opened 2022-06-09
[4.2] Remove unnecessary $this->app->getIdentity() check

joomdonation:
Pull Request for Issue # .

### Summary of Changes
As pointed out in my comment https://github.com/joomla/joomla-cms/pull/38011#discussion_r893650429, `$this->app->getIdentity()` will always return a `User` object, so the check here could be removed.

### Testing Instructions
1. Use Joomla 4.2 nightly build
2. Apply patch
3. Access to Articles Management screen from backend, make sure nothing is broken.

Works but there is unnecessary code.

Works

### DIFF
--- a/libraries/src/MVC/Controller/BaseController.php
+++ b/libraries/src/MVC/Controller/BaseController.php
@@ -595,7 +595,7 @@ class BaseController implements ControllerInterface, DispatcherAwareInterface
 			return false;
 		}
 
-		if ($model instanceof CurrentUserInterface && $this->app->getIdentity())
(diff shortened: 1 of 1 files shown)


richard67:
@joomdonation The installation at the beginning of the api tests in drone fails. I’ r restarted drone 3 times now but always the same result. Are you sure the removed code was useless? Have you tested to make a new installation with your changes?

joomdonation:
@richard67 Sorry, I did not check installation. The installation app does not initialize the identity object like our other apps, so the change in this PR is causing an error.

PR #38030, opened 2022-06-11
[4.x] Joomla.request return Promise, optionally

Fedik:
### Summary of Changes

Allow `Joomla.request` to return a `Promise`.
It more like RFC, not for 4.2, but we can put it in one of future version.

I have added new option `promise`, when it is `true` then `Joomla.request` will return a `Promise`.
So it can be used as:
```javascript
Joomla.request({
    url: 'blabla/file.json'
    promise: true
})
  .then((xhr) => {
    console.log('ok', xhr.responseText);
  })
  .catch((xhr) => {
    console.error('error', xhr.statusText);
  });
```

### Testing Instructions
Run `npm install`.
Make sure all AJAX stuf works as before, for example extensions upload.

Then add to index.php of any template:
```html
<script type="module">
  Joomla.request({
    url: Joomla.getOptions('system.paths').base + '/index.php?option=com_ajax&type=plugin',
    promise: true
  })
    .then(() => console.log('success'))
    .catch(() => console.log('error'));
    
  Joomla.request({
    url: Joomla.getOptions('system.paths').base + '/index.php?option=com_ajax&type=plugin&format=raw',
    promise: true
  })
    .then(() => console.log('success'))
    .catch(() => console.log('error'));
</script>
```
Watch result in the browser console.

2 message (1 per request):
```
error
success
```

2 message (1 per request):
```
error
success
```

yeah, if it exists somwhere,
I have updated comment for `Joomla.request` method description.

### DIFF
--- a/build/media_source/system/js/core.es6.js
+++ b/build/media_source/system/js/core.es6.js
@@ -577,6 +577,8 @@ window.Joomla.Modal = window.Joomla.Modal || {
    *    perform: true,        Perform the request immediately
    *              or return XMLHttpRequest instance and perform it later
    *    headers: null,        Object of custom headers, eg {'X-Foo': 'Bar', 'X-Bar': 'Foo'}
+   *    promise: false        Whether return a Promise instance.
+   *              When true then next options is ignored: perform, onSuccess, onError, onComplete
    *
    *    onBefore:  (xhr) => {}            // Callback on before the request
    *    onSuccess: (response, xhr) => {}, // Callback on the request success
@@ -598,18 +600,18 @@ window.Joomla.Modal = window.Joomla.Modal || {
    * @see    https://developer.mozilla.org/docs/Web/API/XMLHttpRequest
    */
   Joomla.request = (options) => {
-    let xhr;
     // Prepare the options
     const newOptions = Joomla.extend({
       url: '',
       method: 'GET',
       data: null,
       perform: true,
+      promise: false,
     }, options);
 
-    // Set up XMLHttpRequest instance
-    try {
-      xhr = new XMLHttpRequest();
+    // Setup XMLHttpRequest instance
+    const createRequest = (onSuccess, onError) => {
+      const xhr = new XMLHttpRequest();
 
       xhr.open(newOptions.method, newOptions.url, true);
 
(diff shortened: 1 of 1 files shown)


HLeithner:
@Fedik can you please fix the conflict, thanks

Fedik:
Hm, currently it not really possible, because #38019

HLeithner:
Because of the queue handling in this function? I didn't read the whole thread so sorry if I miss something.

Wouldn't make more sense to have an own function that handles the queue and return a promise which then uses the Request method?

Fedik:
> Because of the queue handling in this function?

yes

> Wouldn't make more sense to have an own function that handles the queue and return a promise which then uses the Request method?

And that first what I have sugested in that PR: "use own method that handles the queue" :)

Well, I will look later, maybe will find another way.

Fedik:
@nikosdion what do you think about reimplementing #38019 using https://github.com/joomla/joomla-cms/pull/38019#issuecomment-1151327526 with changes from this PR?

yeah I remember I talked about new method for promise, but seems it not that hard in existing method :)

nikosdion:
@Fedik When I tried this method it did not work AT ALL. I explained why this is the case in https://github.com/joomla/joomla-cms/pull/38019#issuecomment-1151464935 You seem to also be forcing perform=true when a promise is returned, just like the Fetch API, therefore we'd never be able to set up the queue. Everything would execute at one, so we're back where we started as if 38019 never happened at all.

Fedik:
Yeah, the javascript promisses is eagar not lazy.

hmhm, I thought this one would work, where you create a next promise only after previous is resolved:

```javascript
let lastRequestPromise = Promise.resolve();

Joomla.enqeueRequest = (options) => {
   options.promise = true;
   lastRequestPromise = lastRequestPromise.then(() => Joomla.Request(options));
}
```

Well, okay

nikosdion:
You _can_ create a sequence with promises **but** it requires knowing all of the queue items in advance, see https://web.dev/promises/#creating-a-sequence  

Our problem is that we cannot know our queue items in advance. They might even be added at runtime as a result of a previous queue item executing or some conditional JavaScript. As a result we cannot use this trick for creating a sequence of web requests with promises.

Adding Promises support to Joomla.Request is not useless, though! There are some cases where you want to run several requests in parallel and take action when all of them have completed, the JS equivalent of using curl_multi_init() in PHP code.

PR #28884, opened 2020-04-30
Fix for Broken html in Date of Birth field in User Profile Form

AndySDH:
Pull Request for Issue #28883, see that for code details.

### Summary of Changes

This fixes the Date of Birth field in User Profile Edit Form that generates a broken html.

The issue is to due the extra info about the date format being added with broken html around it.

### Previous result

The previous "hacky" HTML creates an extra control-group, and the actual "controls" is outside of it, so two problems for the price of one.

The goal was to put the info language string on a separate line but I don't think that is worth breaking the HTML structure of the form.

### New result

Now the info sits near the label and doesn't break the HTML of the control-group. It does occupy a bit more space (with the default layout) but at least the HTML is correct.

### Testing Instructions

Simply see the HTML of the edit form of a user profile page with a DOB field enabled.

### Documentation Changes Required

None

Issue #28883, opened 2020-04-30
[3.9] Broken html in Date of Birth field in User Profile Form


AndySDH:
I noticed the Date of Birth field in User Profile Edit Form generates a broken html.

The issue is to due the extra info about the date format being added with broken html around it.

Might be related to this: https://github.com/joomla/joomla-cms/pull/7345

### Expected result

It should generate something like this (like all the other fields):

```
<div class="control-group">
    <div class="control-label">
            // Label here
    </div>
    <div class="controls">
            // Calendar Input here
            <span>The date of birth entered should use the format Year-Month-Day, ie 0000-00-00</span>
    </div>
</div>
```

### Actual result

Instead it generates a clearly broken HTML like this:

```
<div class="control-group">
    <div class="control-label">
    </div>
</div>
<div class="controls">The date of birth entered should use the format Year-Month-Day, ie 0000-00-00</div>
<div class="control-group">
    <div class="control-label">
      // Label here
    </div>
    <div class="controls">
      // Calendar Input here
    </div>
</div>
```

As you can see it creates an extra control-group, and the actual "controls" is outside of it, so two problems for the price of one.

Should be an easy fix probably but not sure in which file to look, couldn't find it.

SharkyKZ:
https://github.com/joomla/joomla-cms/blob/2bac089d1e508548fab61720c16328dcd7e535b9/layouts/plugins/user/profile/fields/dob.php#L17

This is done on purpose but it looks very hacky. Maybe there's a better way.

AndySDH:
Oh nice job for finding the culprit file! I will see to come up with a better solution for it and open a PR.

AndySDH:
See pull request https://github.com/joomla/joomla-cms/issues/28883

AndySDH:
See pull request https://github.com/joomla/joomla-cms/pull/28884

PR #38057, opened 2022-06-14
[4.2 re-do] Fix for Broken html in Date of Birth field in User Profile Form

AndySDH:
As requested, here's the Joomla 4.2 Re-Do of Pull Request #28884 (which was for Issue #28883).

### Summary of Changes

This fixes the Date of Birth field in User Profile Edit Form that generates a broken html.

The issue is to due the extra info about the date format being added with broken html around it.

I've also update the language string to be shorter in order to fit the new positioning better. See the original PR #28884 for more details/discussion. 

### Previous result

The previous "hacky" HTML creates an extra control-group, and the actual "controls" is outside of it, so two problems for the price of one.

The goal was to put the info language string on a separate line but this breaks the HTML structure of the form.

### New result

Now the info sits near the label and doesn't break the HTML of the control-group. It does occupy a bit more space (with the default layout) but now the HTML is valid

### Testing Instructions

Simply see the HTML of the edit form of a user profile page with a DOB field enabled.

None


administrator/language/en-GB/plg_user_profile.ini
```diff
@@ -12,7 +12,7 @@ PLG_USER_PROFILE_FIELD_ADDRESS1_LABEL="Address 1"
 PLG_USER_PROFILE_FIELD_ADDRESS2_LABEL="Address 2"
 PLG_USER_PROFILE_FIELD_CITY_LABEL="City"
 PLG_USER_PROFILE_FIELD_COUNTRY_LABEL="Country"
-PLG_USER_PROFILE_FIELD_DOB_LABEL="Date of Birth"
+PLG_USER_PROFILE_FIELD_DOB_LABEL="Birth Date"
```
brianteeman: This should be reverted - it is always Date of Birth unless its a bad translation

AndySDH: Ok, sorry, will do.

brianteeman:
What am I missing? Surely this just needs a regular field of type calendar and no extra fields or layouts

image

image

or could also include a hint

image


image

AndySDH:
To be honest I agree. But I wasn't / am not sure how to go about completely removing a layout. So if you know how to do so, please be my guest, I do agree with you that a simple hint would be enough.

Kostelano:
Don't add `hint="YYYY-MM-DD"` because `DD.MM.YYYY` is correct for Russian/Ukrainian and many other languages.

Screenshot_1

brianteeman:
The hint is a translatable string

AndySDH:
@brianteeman Can you make a new PR with your idea of making it a calendar field with the hint?

brianteeman:
alternative created as requested #38065

AndySDH:
I'll close this in favor of @brianteeman's PR.

PR #38092, opened 2022-06-19
[4.2] [GSoC 21] Keyboard-Shortcut Plugin

fancyFranci:
### Summary of Changes
This is a new plugin, developed in the GSoC 21 Accessibility project. 

It introduces keyboard shortcuts to the administrator site like pressing J + F to jump into the search field or J + S to save. Other extensions are able to add their shortcuts as well by hooking in on the `onLoadShortcuts` event. All(!) of them are listed in a modal window, which shows up by pressing J + X. 

The current state is just the start. Further steps would be the editing of shortcuts, the possibility to disable them, changing the overview-text-hint to a button in the corner etc. The editing was there already but it had some flaws, so it will come later (without flaws) :)

Thanks to @Krshivam25 the student behind this GSoC project, @chmst, @bembelimen, and @brianteeman for alle the work and feedback!

### Testing Instructions
Make sure the shortcut plugin is enabled and start by pressing J + X (in sequence, not at the same time!) for the overview. Have a look at the hint, which tells you the same at the bottom of the page. Then visit different pages and try out the shortcuts :)

If your motivated you can let your plugin add new shortcuts and test them! Add "onLoadShortcuts" to the getSubscribedEvents and write something like this:
```
public function onLoadShortcuts(EventInterface $event): void {
   $shortcuts = $event->getArgument('shortcuts');
   $testShortcuts = array('test' => (object)['shortcut' => 'J + T', 'title' => 'Great Test', 'selector' => '#menu-collapse']);
   $event->setArgument('shortcuts',  array_merge($shortcuts, $testShortcuts));
}
```

You needed to click on buttons to do something

You can use your keyboard to do something (besides typing text)

Yes, I will create a page to describe the build up of the shortcuts, how extensions can add them and what the default joomla shortcuts are. Part of it would be:

The 3rd party plugin needs to provide an array of shortcuts which are objects:
`{ shortcut: string, selector: string, title: string }`
The shortcut needs to be a keyboard input, separated by a plus e.g. "J + Y" or "ALT + Z + 7" (currently there is really no filtering).
Selectors are like CSS selectors to target an element. When it is an input element, the shortcut gives focus, otherwise it will be clicked.
The Title will be displayed in the overview. It could be the name of the target.


brianteeman:
You need to add update sql scripts

brianteeman:
image

brianteeman:
> Other extensions are able to add their shortcuts as well by hooking in on the onLoadShortcuts event. 

Can you provide an example please

> All of them are listed in a modal window, which shows up by pressing J + X.

Would that include these additional extension provided shortcuts

brianteeman:
When the keyboard focus is on a select then the hotkeys do not work

image

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
```diff
@@ -0,0 +1,3 @@
+-- See https://github.com/joomla/joomla-cms/pull/38092
+INSERT INTO `#__extensions` (`name`, `type`, `element`, `folder`, `client_id`, `enabled`, `access`, `protected`, `manifest_cache`, `params`, `custom_data`, `checked_out`, `checked_out_time`, `ordering`, `state`) VALUES
+('plg_system_shortcut', 'plugin', 'shortcut', 'system', 0, 1, 1, 0, '', '{}', '', 0, '0000-00-00 00:00:00', 0, 0);
```
brianteeman: Please look at other examples as this should be use NULL for dates

administrator/components/com_admin/sql/updates/postgresql/4.2.0-2022-06-19.sql
```diff
@@ -0,0 +1,3 @@
+-- See https://github.com/joomla/joomla-cms/pull/38092
+INSERT INTO `#__extensions` (`name`, `type`, `element`, `folder`, `client_id`, `enabled`, `access`, `protected`, `manifest_cache`, `params`, `custom_data`, `checked_out`, `checked_out_time`, `ordering`, `state`) VALUES
+('plg_system_shortcut', 'plugin', 'shortcut', 'system', 0, 1, 1, 0, '', '{}', '', 0, '1970-01-01 00:00:00', 0, 0);
```
brianteeman: Please look at other examples as this should be use NULL for dates

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
richard67: For checked_out (the flag of checked out) we also use null and not 0 meanwhile if I remember right. The problem ist that this change happens during the 4.0 update SQL in the middle, so if you copy such stuff from the very first update SQL scripts you still find the old way.

administrator/components/com_admin/sql/updates/postgresql/4.2.0-2022-06-19.sql
richard67: And zero for checked_out, see my previous comment.

brianteeman:
I am sorry I made a big accessibility mistake when I suggested using J as the modifier key.

If you are unable to press or have difficulty pressing shortcut key combinations then (on windows) you can enable sticky keys.

image

However the only sticky keys are the traditional modifiers. 

The way github does it using the G key is that they appear to use it as a keyboard sequence and not a modifier.

In other words I can press G followed by I to go to issues
In this PR I have to press and hold J while pressing the S to go to the Search bar

plugins/system/shortcut/shortcut.php
```diff
@@ -0,0 +1,147 @@
...
+		PluginHelper::importPlugin('shortcut');
+
+		$context = $this->app->input->get('option') . '.' . $this->app->input->get('view');
+
+		$shortcuts = [];
+
+		$event = new GenericEvent('onLoadShortcuts', [
+			'context' => $context,
+			'shortcuts' => $shortcuts
+		]);
```
heelc29:
```suggestion
		$event = new GenericEvent(
			'onLoadShortcuts',
			[
				'context'   => $context,
				'shortcuts' => $shortcuts,
			]
		);
```
https://ci.joomla.org/joomla/joomla-cms/55527/1/6

plugins/system/shortcut/shortcut.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<extension type="plugin" group="system" method="upgrade">
+	<name>plg_system_shortcut</name>
+	<author>Joomla! Project</author>
+	<creationDate>Jun 2022</creationDate>
```
heelc29:
```suggestion
	<creationDate>2022-06</creationDate>
```
#27162

plugins/system/shortcut/shortcut.xml
```diff
@@ -0,0 +1,23 @@
...
+	<copyright>(C) 2022 Open Source Matters, Inc.</copyright>
+	<license>GNU General Public License version 2 or later; see LICENSE.txt</license>
+	<authorEmail>admin@joomla.org</authorEmail>
+	<authorUrl>www.joomla.org</authorUrl>
+	<version>__DEPLOY_VERSION__</version>
+	<description>PLG_SYSTEM_SHORTCUT_XML_DESCRIPTION</description>
+	<namespace path="src">Joomla\Plugin\System\Shortcut</namespace>
+	<media destination="plg_system_shortcut" folder="media">
+	  <folder>js</folder>
```
heelc29:
```suggestion
		<folder>js</folder>
```

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,13 @@
+; Joomla! Project
+; (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt, see LICENSE.php
```
heelc29:
```suggestion
; License GNU General Public License version 2 or later; see LICENSE.txt
```

administrator/language/en-GB/plg_system_shortcut.sys.ini
```diff
@@ -0,0 +1,7 @@
+; Joomla! Project
+; (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt, see LICENSE.php
```
heelc29:
```suggestion
; License GNU General Public License version 2 or later; see LICENSE.txt
```

administrator/language/en-GB/plg_system_shortcut.sys.ini
```diff
@@ -0,0 +1,7 @@
...
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcut"
+PLG_SYSTEM_SHORTCUT_XML_DESCRIPTION="<p>Enables keyboard shortcuts on the administrator site, which can be provided by other plugins and includes directly the following list of shortcuts:<p><ul><li>Save: J + S</li><li>Cancel: J + Q</li><li>New: J + N</li><li>Save & New: J + N</li><li>Save & Close: J + W</li><li>Search: J + S</li><li>Option: J + O</li><li>Help: J + H</li><li>Overview: J + X</li></ul>"
```
heelc29:
```suggestion
PLG_SYSTEM_SHORTCUT_XML_DESCRIPTION="<p>Enables keyboard shortcuts on the administrator site, which can be provided by other plugins and includes directly the following list of shortcuts:</p><ul><li>Save: J + S</li><li>Cancel: J + Q</li><li>New: J + N</li><li>Save & New: J + N</li><li>Save & Close: J + W</li><li>Search: J + S</li><li>Option: J + O</li><li>Help: J + H</li><li>Overview: J + X</li></ul>"
```

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,13 @@
...
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcut"
+PLG_SYSTEM_SHORTCUT_ACTIONS_FORM_LABEL="Form View Actions"
+PLG_SYSTEM_SHORTCUT_ACTIONS_GENERAL_LABEL="General Actions"
+PLG_SYSTEM_SHORTCUT_ACTIONS_LIST_LABEL="List View Actions"
+PLG_SYSTEM_SHORTCUT_CORE_SHORTCUTS_TAB="Core Shortcuts"
+PLG_SYSTEM_SHORTCUT_OVERVIEW_HINT="Press J + X for our keyboard shortcut overview"
+PLG_SYSTEM_SHORTCUT_OVERVIEW_TITLE="Available Joomla Shortcuts"
+PLG_SYSTEM_SHORTCUT_XML_DESCRIPTION="<p>Enables keyboard shortcuts on the administrator site, which can be provided by other plugins and includes directly the following list of shortcuts:<p><ul><li>Save: J + S</li><li>Cancel: J + Q</li><li>New: J + N</li><li>Save & New: J + N</li><li>Save & Close: J + W</li><li>Search: J + S</li><li>Option: J + O</li><li>Help: J + H</li><li>Overview: J + X</li></ul>"
```
heelc29:
```suggestion
PLG_SYSTEM_SHORTCUT_XML_DESCRIPTION="<p>Enables keyboard shortcuts on the administrator site, which can be provided by other plugins and includes directly the following list of shortcuts:</p><ul><li>Save: J + S</li><li>Cancel: J + Q</li><li>New: J + N</li><li>Save & New: J + N</li><li>Save & Close: J + W</li><li>Search: J + S</li><li>Option: J + O</li><li>Help: J + H</li><li>Overview: J + X</li></ul>"
```

administrator/language/en-GB/plg_system_shortcut.sys.ini
fancyFranci: Thank you, eagle eye :)

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
fancyFranci: Exactly, I copied it from an old file -.-

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
```diff
@@ -0,0 +1,3 @@
+-- See https://github.com/joomla/joomla-cms/pull/38092
+INSERT INTO `#__extensions` (`name`, `type`, `element`, `folder`, `client_id`, `enabled`, `access`, `protected`, `manifest_cache`, `params`, `custom_data`, `checked_out`, `checked_out_time`, `ordering`, `state`) VALUES
```
richard67: @fancyFranci The update SQL is still not right. There is a new column "locked" which has to be set to 1 for core extensions so that they cannot be uninstalled. This column was added in the 4.0 update SQL scripts in the middle, so when you copied your statement from an older one, that column was missing. In addition, it does not need to have the checked_out and the checked_out_time at all in this statement with their null values since they default to NULL anyway. An actual example on how extensions should be added with an update script can be found here for MySQL: https://github.com/joomla/joomla-cms/blob/4.2-dev/administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-05-15.sql#L25-L30 . For PostgreSQL applies the same, see https://github.com/joomla/joomla-cms/blob/4.2-dev/administrator/components/com_admin/sql/updates/postgresql/4.2.0-2022-05-15.sql#L28-L33 .

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
richard67: Still not ok, see my later comment below.

fancyFranci:
> When the keyboard focus is on a select then the hotkeys do not work

From my side this is intended. Users should be able to jump to a select value with a character, depending on the available values. Like pressing "u" in the "-select featured-" select, I want to jump to "unfeatured". If there is a value starting with "j", I want to select that.

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
richard67: P.S. And as you can see in the linked example, there is also another column "custom_data" which was removed at the beginning of J4 and later was added back.

fancyFranci: Thank you, I took the example you provided. Is it ok to set ordering on zero, unlike the example did it?

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
richard67: Looks ok now. Thanks @brianteeman for the find.

administrator/components/com_admin/sql/updates/postgresql/4.2.0-2022-06-19.sql
richard67: Looks ok now. Thanks @brianteeman for the find.

administrator/components/com_admin/sql/updates/mysql/4.2.0-2022-06-19.sql
richard67: @fancyFranci Looks ok now. Thanks @brianteeman for the first find.

richard67: > Is it ok to set ordering on zero, unlike the example did it?

@fancyFranci I would have used 1 and no 0, but I think it doesn't matter.

richard67: @fancyFranci P.S.: In Nik's case it was a completely new plugin group so he could specify the right ordering within that group. In your case you add a new plugin to the existing system plugins so you can't really know about ordering on update because this might have been changed by the user, so I think zero is ok.

fancyFranci:
> > All of them are listed in a modal window, which shows up by pressing J + X.
> 
> Would that include these additional extension provided shortcuts

Yes, as long as the provided shortcuts have a title, they are added to the overview.

@brianteeman Like your screenshot displayed it shortcuts can trigger different things, depending on the context. I did not find a screen where the Search and the Save & Close button are both present, so I let them use the same shortcut. Same with New and Save & New. Do you have concerns or is it their presentation in the overview which is an issue?

brianteeman:
> Do you have concerns or is it their presentation in the overview which is an issue?

Both. 

In other systems only the valid shortcuts are displayed (see github for examples)

It is not normal to use the same shortcut for two different things. It really makes memorising them very difficult.

brianteeman:
> > When the keyboard focus is on a select then the hotkeys do not work
> 
> From my side this is intended. Users should be able to jump to a select value with a character, depending on the available values. Like pressing "u" in the "-select featured-" select, I want to jump to "unfeatured". If there is a value starting with "j", I want to select that.

you are correct that they should be able to find a value beginning with J. But you are using J as a modifier key. In other words press J on its own and it will find the word beginning with J. Press J in combination with another key and the shortcut should happen.

brianteeman:
Please add the plugin to the list of core plugins
https://github.com/joomla/joomla-cms/blob/f94b46b0c23ba242baee2fdf8121584c013dcf96/libraries/src/Extension/ExtensionHelper.php#L271

fancyFranci:
> [...]
> The way github does it using the G key is that they appear to use it as a keyboard sequence and not a modifier.
> 
> In other words I can press G followed by I to go to issues In this PR I have to press and hold J while pressing the S to go to the Search bar

Ooh thats the reason github does that... I hoped we would not need that, as it means splitting up the shortcuts, adding a timeout, and so on.

Is this something we could add to the future-feature-list or is it stopping this PR?

brianteeman:
> Is this something we could add to the future-feature-list or is it stopping this PR?

Well we can't call it an accessibility feature if it is not usable by someone who cannot press more than one key at a time. it needs to be a key sequence and not a key combination.

brianteeman:
_I think you can do it by setting a **scope** of J_. But I could be completely wrong.

fancyFranci:
> > Do you have concerns or is it their presentation in the overview which is an issue?
> 
> Both.
> 
> In other systems only the valid shortcuts are displayed (see github for examples)
> 
> It is not normal to use the same shortcut for two different things. It really makes memorising them very difficult.

Ok, the only reason for doubling them were the matching characters. Search => S, Save => S, but yes, it is not optimal. Then I need to set "random" chars or do you have another idea? :/

plugins/system/shortcut/src/Extension/Shortcut.php
```diff
@@ -0,0 +1,130 @@
...
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function initialize()
+	{
+		if (!$this->getApplication()->isClient('administrator'))
+		{
+			return;
+		}
+
+		PluginHelper::importPlugin('shortcut');
```
laoneo: This looks wrong to me to make a system plugin which then load other plugins. This would mean, that there is a system plugin loaded on every request, just to load another group of plugins.

brianteeman: I assume that this is how an extension developer can add their pwn shortcuts?

brianteeman:
>  Then I need to set "random" chars or do you have another idea? :/

well F (find) is usually used in browsers to prevent conflicts with S (save)

installation/sql/mysql/base.sql
```diff
@@ -342,6 +342,7 @@ INSERT INTO `#__extensions` (`package_id`, `name`, `type`, `element`, `folder`,
 (0, 'plg_system_sef', 'plugin', 'sef', 'system', 0, 1, 1, 0, 1, '', '', '', 18, 0),
 (0, 'plg_system_sessiongc', 'plugin', 'sessiongc', 'system', 0, 1, 1, 0, 1, '', '', '', 19, 0),
 (0, 'plg_system_skipto', 'plugin', 'skipto', 'system', 0, 1, 1, 0, 1, '', '{}', '', 20, 0),
+(0, 'plg_system_shortcut', 'plugin', 'shortcut', 'system', 0, 1, 1, 0, 0, '', '{}', '', 0, 0),
```
heelc29: Can you move `plg_system_shortcut` before `plg_system_skipto` (alphasort)?
And set `locked` to 1 as in the sql update?

richard67: @fancyFranci It needs the same also for the base.sql for postgresql.

installation/sql/postgresql/base.sql
```diff
@@ -348,6 +348,7 @@ INSERT INTO "#__extensions" ("package_id", "name", "type", "element", "folder",
 (0, 'plg_system_sef', 'plugin', 'sef', 'system', 0, 1, 1, 0, 1, '', '', '', 18, 0),
 (0, 'plg_system_sessiongc', 'plugin', 'sessiongc', 'system', 0, 1, 1, 0, 1, '', '', '', 19, 0),
 (0, 'plg_system_skipto', 'plugin', 'skipto', 'system', 0, 1, 1, 0, 1, '', '{}', '', 20, 0),
+(0, 'plg_system_shortcut', 'plugin', 'shortcut', 'system', 0, 1, 1, 0, 0, '', '{}', '', 0, 0),
```
richard67: It needs the same change here as done before for the MySQL file (locked = 1 and alpha ordering).

plugins/system/shortcut/src/Extension/Shortcut.php
fancyFranci: I removed it and saw no difference, other plugins can still add shortcuts. Good that you questioned that.

brianteeman:
image

This might look better if the keys come before the text?

brianteeman:
It works much better now with the scope and using the keys in sequence. Its finally accessible.

However this means that it can not be called a modifier key. A modifier key is a key thata you hold down to change another key. And it doesnt work if you try to do that. Not sure what the correct terminology is. Will check that in the morning.

fancyFranci:
Indeed, the other way around looks better! Happy testing everyone

brianteeman:
Note to self (or anyone else) of additional things to test tomorrow

- [x] English strings
- [x] RTL
- [x] non english keyboard
- [ ] accessibility with screen reader
- [ ] order the keys in the overview (and how to make that sensible in other languages?)

brianteeman:
Can I request one additional ccore shortcut J D
takes you to the home dashboard

bembelimen:
> 
> However this means that it can not be called a modifier key. A modifier key is a key thata you hold down to change another key. And it doesnt work if you try to do that.

grafik

This works for me, first pressing "J" and then "ALT + N" at the same time.

bembelimen:
>     * [ ]  accessibility with screen reader

I wonder if we should show some kind of modal/text when "J" was pressed with aria-attributes to make aware, that we're in the shortcut mode?

bembelimen:
> Can I request one additional ccore shortcut J D takes you to the home dashboard

- Create Article
- Goto articles
- [...] menus
- [...] users

From everywhere...

brianteeman:
> > ```
> > * [ ]  accessibility with screen reader
> > ```
> 
> I wonder if we should show some kind of modal/text when "J" was pressed with aria-attributes to make aware, that we're in the shortcut mode?

I will check what screen readers do with their own keys

brianteeman:
### Definition Lists
The code is currently 

- Defnition Term (dt) = Save
- Definition Description (dd) = J+A

I _think_ this is the wrong way around. @chmst what do you think?

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,12 @@
+; Joomla! Project
+; (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcut"
```
brianteeman:
```suggestion
PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcuts"
```

administrator/language/en-GB/plg_system_shortcut.sys.ini
```diff
@@ -0,0 +1,7 @@
+; Joomla! Project
+; (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcut"
```
brianteeman:
```suggestion
PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcuts"
```

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,12 @@
...
+PLG_SYSTEM_SHORTCUT_OVERVIEW_DESC="Press <span class=\"badge bg-dark px-2\">J</span> to access the shortkey mode followed by the shortkey."
+PLG_SYSTEM_SHORTCUT_OVERVIEW_HINT="Press J + X for our keyboard shortcut overview"
```
brianteeman:
```suggestion
PLG_SYSTEM_SHORTCUT_OVERVIEW_HINT="J + X - Keyboard Shortcuts"
```

I think this is better simplified like this

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,12 @@
...
+PLG_SYSTEM_SHORTCUT_OVERVIEW_TITLE="Available Joomla Shortcuts"
```
brianteeman:
```suggestion
PLG_SYSTEM_SHORTCUT_OVERVIEW_TITLE="Joomla Keyboard Shortcuts"
```

We an't say available as the shortcut might not be available on the page where this is displayed

chmst:
Example here: Key combination first: https://webaim.org/resources/shortcuts/nvda
Also on Visual studio Code.

chmst:
What Screenreaders announce depends on screenreader and OS. Will build a test installation where more people can perform tests.

plugins/system/shortcut/src/Extension/Shortcut.php
```diff
@@ -0,0 +1,135 @@
...
+		$shortcuts = array_merge($shortcuts, [
+			'applyKey'   => (object) ['selector' => 'joomla-toolbar-button .button-apply', 'shortcut' => 'A', 'title' => Text::_('JAPPLY')],
+			'cancelKey'  => (object) ['selector' => 'joomla-toolbar-button .button-cancel', 'shortcut' => 'Q', 'title' => Text::_('JCANCEL')],
+			'helpKey'    => (object) ['selector' => 'joomla-toolbar-button .button-help', 'shortcut' => 'H', 'title' => Text::_('JHELP')],
+			'newKey'     => (object) ['selector' => 'joomla-toolbar-button .button-new', 'shortcut' => 'N', 'title' => Text::_('JTOOLBAR_NEW')],
+			'optionKey'  => (object) ['selector' => 'joomla-toolbar-button .button-options', 'shortcut' => 'O', 'title' => Text::_('JOPTIONS')],
+			'saveKey'    => (object) ['selector' => 'joomla-toolbar-button .button-save', 'shortcut' => 'S', 'title' => Text::_('JTOOLBAR_SAVE')],
+			'searchKey'  => (object) ['selector' => 'input[placeholder=' . Text::_('JSEARCH_FILTER') . ']', 'shortcut' => 'F', 'title' => Text::_('JSEARCH_FILTER')],
+			'toggleMenu' => (object) ['selector' => '#menu-collapse', 'shortcut' => 'M', 'title' => Text::_('JTOGGLE_SIDEBAR_MENU')],
+		]);
```
richard67:
```suggestion
		$shortcuts = array_merge(
			$shortcuts,
			[
				'applyKey'   => (object) ['selector' => 'joomla-toolbar-button .button-apply', 'shortcut' => 'A', 'title' => Text::_('JAPPLY')],
				'cancelKey'  => (object) ['selector' => 'joomla-toolbar-button .button-cancel', 'shortcut' => 'Q', 'title' => Text::_('JCANCEL')],
				'helpKey'    => (object) ['selector' => 'joomla-toolbar-button .button-help', 'shortcut' => 'H', 'title' => Text::_('JHELP')],
				'newKey'     => (object) ['selector' => 'joomla-toolbar-button .button-new', 'shortcut' => 'N', 'title' => Text::_('JTOOLBAR_NEW')],
				'optionKey'  => (object) ['selector' => 'joomla-toolbar-button .button-options', 'shortcut' => 'O', 'title' => Text::_('JOPTIONS')],
				'saveKey'    => (object) ['selector' => 'joomla-toolbar-button .button-save', 'shortcut' => 'S', 'title' => Text::_('JTOOLBAR_SAVE')],
				'searchKey'  => (object) ['selector' => 'input[placeholder=' . Text::_('JSEARCH_FILTER') . ']', 'shortcut' => 'F', 'title' => Text::_('JSEARCH_FILTER')],
				'toggleMenu' => (object) ['selector' => '#menu-collapse', 'shortcut' => 'M', 'title' => Text::_('JTOGGLE_SIDEBAR_MENU')],
			]
		);
```
Fix PHPCS failing in drone.

brianteeman:
Grrrh.
Using a keyboard sequence doesnt seem to work as I expected. It doesnt work at all in NVDA or Windows Narrator.

In both cases the combination J + X tries to go to the next checkbox.

brianteeman:
successfully tested in arabic that both the display in RTL is correct and that the shortcut keys work in chrome for windows when chrome and joomla are set to use arabic language for display and arabic language for keyboard

fancyFranci:
> > Can I request one additional ccore shortcut J D takes you to the home dashboard
> 
>     * Create Article
> 
>     * Goto articles
> 
>     * [...] menus
> 
>     * [...] users
> 
> 
> From everywhere...

I added the possibility in general and dashboard works. I can include more shortcuts of course, I only need details about urls and selectors.

build/media_source/plg_system_shortcut/js/shortcut.es6.js
```diff
@@ -0,0 +1,169 @@
...
+
+    const modal = `
+      <div class="modal fade" id="shortcutOverviewModal" tabindex="-1" role="dialog" data-bs-backdrop="static" aria-labelledby="shortcutOverviewModalLabel" aria-hidden="true">
+        <div class="modal-dialog" role="document">
+          <div class="modal-content">
+            <div class="modal-header">
+              <h3 id="shortcutOverviewModalLabel" class="modal-title">
+                ${Joomla.Text._('PLG_SYSTEM_SHORTCUT_OVERVIEW_TITLE')}
+              </h3>
+              <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
```
Quy: Please use language string.

fancyFranci: You're right

Quy:
Maybe sort in the same order as in the plugin and move text next to keys?

PLUGIN:
shortkey-plugin

MODAL:
shortkey-modal

Quy:
Instead of using `.badge`, maybe use `<kbd>`. https://getbootstrap.com/docs/5.2/content/reboot/#user-input

plugins/system/shortcut/src/Extension/Shortcut.php
```diff
@@ -0,0 +1,142 @@
...
+ * @package     Joomla.Plugin
+ * @subpackage  System.shortcut
+ *
+ * @copyright   (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+namespace Joomla\Plugin\System\Shortcut\Extension;
+
+defined('_JEXEC') or die;
```
Quy:
```suggestion
\defined('_JEXEC') or die;
```

fancyFranci:
> Instead of using `.badge`, maybe use `<kbd>`. https://getbootstrap.com/docs/5.2/content/reboot/#user-input

Thanks for the hint, it makes the html much more readable.

fancyFranci:
> [...] and move text next to keys?

Currently it is a dl and thats the default style of that kind of list. Previously it was a ul and all in line but wishes were to use a dl. Long story short I let the style like it is for now, but as more improvements are planned it could be part of them.
But I aligned the order at least :)

fancyFranci:
Ok guys, my target is to bring this feature into 4.2 beta-2 which will be released today in the evening. Otherwise shortcuts would need to wait until 4.3. So every further improvement will find its way into beta-3 and this PR here will get merged after 2 tests. I hope thats ok, @brianteeman ? :) Would be great when we can create a list of things needed to be finished before beta-3.

Quy:
Make the legend less prominent?

keyboard-shortcuts-legend

```
<span class="icon-keyboard fa-keyboard me-2" aria-hidden="true"></span> <kbd>J</kbd>+<kbd>X</kbd> <small>Keyboard Shortcuts</small>
```

fancyFranci:
> Make the legend less prominent?

Done :)

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,12 @@
+; Joomla! Project
+; (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcuts"
+PLG_SYSTEM_SHORTCUT_OVERVIEW_DESC="Press <kbd>J</kbd> to access the shortkey mode followed by the shortkey."
+PLG_SYSTEM_SHORTCUT_OVERVIEW_HINT="<span class='icon-keyboard fa-keyboard me-2' aria-hidden='true'></span> <kbd>J</kbd> + <kbd>X</kbd> <small>Keyboard Shortcuts</small>"
```
Quy:
```suggestion
PLG_SYSTEM_SHORTCUT_OVERVIEW_HINT="<span class=\"icon-keyboard fa-keyboard me-2\" aria-hidden=\"true\"></span> <kbd>J</kbd> + <kbd>X</kbd> <small>Keyboard Shortcuts</small>"
```

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,12 @@
...
+PLG_SYSTEM_SHORTCUT_OVERVIEW_TITLE="Joomla Keyboard Shortcuts"
+PLG_SYSTEM_SHORTCUT_TIMEOUT_DESC="The time in milliseconds how long a shortcut can be pressed after \"J\" has been pressed."
+PLG_SYSTEM_SHORTCUT_TIMEOUT_LABEL="Timeout"
```
Quy:
```suggestion
PLG_SYSTEM_SHORTCUT_TIMEOUT_DESC="Maximum time that a shortcut can be pressed after pressing <kbd>J</kbd>."
PLG_SYSTEM_SHORTCUT_TIMEOUT_LABEL="Timeout (in milliseconds)"
```

administrator/language/en-GB/plg_system_shortcut.ini
```diff
@@ -0,0 +1,12 @@
+; Joomla! Project
+; (C) 2022 Open Source Matters, Inc. <https://www.joomla.org>
+; License GNU General Public License version 2 or later; see LICENSE.txt
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SHORTCUT="System - Keyboard Shortcuts"
+PLG_SYSTEM_SHORTCUT_OVERVIEW_DESC="Press <kbd>J</kbd> to access the shortkey mode followed by the shortkey."
```
Quy:
```suggestion
PLG_SYSTEM_SHORTCUT_OVERVIEW_DESC="Press <kbd>J</kbd> to access the shortcut mode followed by the shortcut."
```

bembelimen:
I have tested this item :white_check_mark: successfully on b2238449b360437329da71abec0c98f40572d33c<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38092">issues.joomla.org/tracker/joomla-cms/38092</a>.</sub>

chmst:
I have tested this item :white_check_mark: successfully on b2238449b360437329da71abec0c98f40572d33c<br /><br />Great!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38092">issues.joomla.org/tracker/joomla-cms/38092</a>.</sub>

brianteeman:
> Grrrh. Using a keyboard sequence doesnt seem to work as I expected. It doesnt work at all in NVDA or Windows Narrator.
> 
> In both cases the combination J + X tries to go to the next checkbox.

fancyFranci:
> > Grrrh. Using a keyboard sequence doesnt seem to work as I expected. It doesnt work at all in NVDA or Windows Narrator.
> > In both cases the combination J + X tries to go to the next checkbox.

Sorry, that one got lost in all the comments. I'll have a look on that, maybe preventing defaults after J was pressed (not X alone) could be enough.

Quy:
I find the legend in the footer a little distracting. My focus draws to it. It is useful in the beginning but I can see it becoming annoying after a while.

shortcut-footer

plugins/system/shortcut/src/Extension/Shortcut.php
```diff
@@ -0,0 +1,142 @@
...
+		Text::script('JCLOSE');
+
+		$document = $this->getApplication()->getDocument();
+		$wa       = $document->getWebAssetManager();
+		$wa->useScript('bootstrap.modal');
+		$wa->registerAndUseScript('script', 'plg_system_shortcut/shortcut.min.js', ['dependencies' => ['hotkeys.js']]);
+
+		$plugin = PluginHelper::getPlugin('system', 'shortcut');
+
+		$timeout = (new Registry($plugin->params))->get('timeout', 2000);
```
heelc29:
```suggestion
		$timeout = $this->params->get('timeout', 2000);
```
Should that also be possible here?

brianteeman:
jx is a dl but the description in the plugin is a ul

plugins/system/shortcut/src/Extension/Shortcut.php
fancyFranci: Indeed!

fancyFranci:
> I find the legend in the footer a little distracting. My focus draws to it. It is useful in the beginning but I can see it becoming annoying after a while.

Now the hint only appears on cpanel-pages like the home dashboard. As you said - its only useful at the beginning

fancyFranci:
> Grrrh. Using a keyboard sequence doesnt seem to work as I expected. It doesnt work at all in NVDA or Windows Narrator.
> In both cases the combination J + X tries to go to the next checkbox.

It seems like NVDA interrupt the keyboard input before js can do/prevent anything. Even when I change the shortcuts to something NVDA does not use, nothing happens besides making a sound. When we configure modifiers like ALT, they do not work in sequence because the browser interrupts (at least in my tests). What do you think @brianteeman ? :/

brianteeman:
(not at my pc to test) what happens on github when you use nvda and the github shortcuts?

fancyFranci:
I tried out s for githubs search and I'm not jumping into the search field as long as nvda is running. Same with g and n, which should normally open notifications.

fancyFranci:
...but the command palette (which opens on ctrl+k) works with nvda

brianteeman:
It looks like its a known issue with the nvda single character shortcuts
https://www.nvaccess.org/files/nvda/documentation/userGuide.html#SingleLetterNavigation

brianteeman:
yes looks like this is one of those cases where a real screen reader user knows what to do.


"In browse mode, you can now toggle single letter navigation on and off by pressing NVDA+shift+space. When off, single letter keys are passed to the application, which is useful for some web applications such as Gmail, Twitter and Facebook."

or you switch from browse to focus mode.

Havng read the nvda manual link I am happy enough that the current code is ok to merge

fancyFranci:
Thanks everybody! :)

brianteeman:
thank you all. 

Cant wait to read the docs how this can be extended by extensions

fancyFranci:
I wrote an example in the test instructions which I used myself to test added shortcuts. But yes, docs will follow.

PR #38095, opened 2022-06-19
[4.2] Remove mocking XMLReader as it's bad practice. Fix parsing version

wilsonge:
### Summary of Changes
Fixes tests to run on PHP 8.1. Generally we shouldn't be mocking core PHP classes so this fixes:

- Everywhere we mocked and then didn't even mock any values/return types etc. Just insert a real dummy instance
- The one place we did test was mocking versions. Here however this covered up a bug because when turning to concrete tests it turned out we parsed the XML version in the XML doc type rather than the version of the feed being used :D This bug is fixed as part of this and one extra test case in atom added

### Testing Instructions
Check Drone 8.1 tests

All tests related to feed do not pass

All tests related to feed do pass (note session tests do fail - but that will be the subject of a different PR)

None


rdeutz (approved):
looks good to me

HLeithner (approved):
Looks good to me but I think needs real tests for our RSS feed.

wilsonge:
Agreed but probably outside the scope of this PR

HLeithner:
I mean a really test in an installed cms to merge this PR ;) not only codereview

wilsonge:
Screenshot 2022-06-19 at 21 35 22
Screenshot 2022-06-19 at 21 35 17

Took a 3rd party one in too for good measure

Hackwar (approved):

HLeithner:
Thanks

wilsonge:
Thankyou!

PR #38105, opened 2022-06-20
[4.2] npm update

brianteeman:
### Summary of Changes
CodeMirror updated to[ 5.65.6](https://codemirror.net/5/doc/releases.html)
and various dev dependencies

All scripts that are not dev dependencies have already been updated or are semver compatible

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "bootstrap": "^5.1.3",
         "choices.js": "^9.0.1",
         "chosen-js": "^1.8.7",
-        "codemirror": "^5.64.0",
+        "codemirror": "^5.65.0",
         "cropperjs": "^1.5.12",
         "diff": "^5.0.0",
         "dragula": "^3.7.3",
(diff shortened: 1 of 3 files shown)


brianteeman:
testing should be quite simple. Apply the pr. then `npm ci`

There should be no new files to commit

then run `npm outdated` and the output will be
image

laoneo:
Vue should be updated as well as it is only a patch shift.

brianteeman:
@laoneo thanks - vue was locked to a specific patch release. Its not any more

laoneo:
Then I'm wondering why it got locked. It was done in #36295. @dgrammatiko any reason you locked vue to a specific release?

dgrammatiko:
> @dgrammatiko any reason you locked vue to a specific release?

Yes, the reason is security and assuring that the person in charge of releases will **NOT** accidentally distribute a version that was not tested/peer reviewed/etc. FWIW for years I kept asking for an automated solution for both npm+composer but gloriously ignored. Anyways, both package.json and composer.json **SHOULD** point to **specific** versions for the dependencies (joomla is not a weekend project, so maintainers should treat it respectively).

An implementation of renovate bot in this repo will apply this before start doing any PRs for updating any of the dependencies (you can check it in one of my many repos where I use it for quite some time, eg: [package.josn](https://github.com/ttc-freebies/plugin-responsive-images/blob/4.0.0/package.json) [renovate.josn](https://github.com/ttc-freebies/plugin-responsive-images/blob/4.0.0/renovate.json) ).

laoneo:
The lock file does the job of pointing to a specific version. There is no need to do that in the config file as well, except when there is a bug/incompatibility/whatever in the library and we have to stick to a specific version.

dgrammatiko:
@laoneo sure but everybody is using specific version **on their package.json**, eg: https://github.com/WordPress/wordpress-develop/blob/40c4f11a81ee28b1ec1869c9842064ac0bf137c2/package.json#L78-L156

brianteeman:
if everybody was doing that then there would be zero point in the ^ or ~ functionality existing.

brianteeman:
closed. i'm not wasting my time resolving conflicts

PR #38107, opened 2022-06-20
[5.2] Don't convert to absolute URLs in MailTemplate when in CLI

bembelimen:
Pull Request for Issue # .

### Summary of Changes
This PR prevents the execution of the relative to absolute URL conversation for Mail Templates when in CLI mode because this method uses (J)Uri which tries to load the URL from the server:

1. https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Mail/MailTemplate.php#L266
2. https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Mail/MailHelper.php#L209
3. https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Uri/Uri.php#L204
4. https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Uri/Uri.php#L142
5. https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Uri/Uri.php#L103

So sending mail templates in the cli does not work properly

### Testing Instructions
Open [`libraries/src/Console/SiteDownCommand.php` Line 97](https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Console/SiteDownCommand.php#L97) and add:

```
$mailtemplate = new \Joomla\CMS\Mail\MailTemplate('com_contact.mail', 'en-GB');
$mailtemplate->addRecipient('test@example.com');
$mailtemplate->send();
exit;
``` 

open the terminal/cmd/powershell, navigate to your joomla folder and run:
```php cli/joomla.php site:down```

PHP Warning:  Undefined array key "HTTP_HOST" in libraries/src/Uri/Uri.php on line 103

No warning


wilsonge:
I think it's more complicated than this I'm afraid. Because if you have the `live_site` parameter set in global config then this would work. We also probably want it to work when in CLI too...

bembelimen:
It still does not work, not because of the live site (this makes it nearly work) but because of the isSsl check line 143 (Uri.php) and getting the server instance above it outputs the warning...

laoneo:
I think you should elaborate a bit more why it fails, because I see it here the same as @wilsonge, that it should work with the live site. Just disabling it on cli is a hard measure. Pretty sure there are real world use cases where mail templates will be used on cli.

wilsonge:
I understand what he's saying - the ssl detection also fails - because it still tries to parse the site to determine that. even if you used global config force ssl (which although it could be coded to work - not all sites even have ssl - and others will have had it setup with nginx/apache from the start without the joomla params - and we need to not hard fail on those sites).

HLeithner:
This pull requests has been automatically converted to the PSR-12 coding standard.

HLeithner:

tomsrocket:
Testers beware: you can only test this issue if you have a (local) mail server! (which I don't)

TLWebdesign:
I have tested this item :red_circle: unsuccessfully on 0799bdd0b63604ac60d46cc1b5b4b4136c71a422<br /><br />I added the code to line 102 (has changes since instructions were written) of a J4.4.3 installation. Ran the site:down command and did not get an error. So can't replicate the issue. <hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38107">issues.joomla.org/tracker/joomla-cms/38107</a>.</sub>

pe7er:
I have tested this item :red_circle: unsuccessfully on 0799bdd0b63604ac60d46cc1b5b4b4136c71a422<br /><br />I've tested this issue on Joomla! 4.4.3
There is no difference before and after the patch

After ``php cli/joomla.php site:down``
the CLI states: 

Site Offline

and I received an email (on local system using Mailhog) with non translated language strings:
Subject	COM_CONTACT_ENQUIRY_SUBJECT
To	"test@example.com" <test@example.com>
Plain text
COM_CONTACT_ENQUIRY_TEXT

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38107">issues.joomla.org/tracker/joomla-cms/38107</a>.</sub>

HLeithner:

Issue #37643, opened 2022-04-24
[4.x] update:extensions:check 


brianteeman:
### Steps to reproduce the issue
Install older versions of a few extensions
Check that the updates are found 

![image](https://user-images.githubusercontent.com/1296369/164998661-c7f8d52f-d8e0-4889-9ddf-6408ad7a2c79.png)

From the cli run

`php cli/joomla.php update:extensions:check 

### Expected result
feedback that updates are available

### Actual result
 [OK] There are no available updates                 
![image](https://user-images.githubusercontent.com/1296369/164998698-5b8964f6-4036-419c-b8f6-e5f20184def2.png)

### System information (as much as possible)
Database Type	mysql
Database Version	8.0.19
Database Collation	utf8mb4_unicode_ci
Database Connection Collation	utf8mb4_0900_ai_ci
Database Connection Encryption	None
Database Server Supports Connection Encryption	Yes
PHP Version	8.0.10
Web Server	Apache/2.4.41 (Win64) mod_fcgid/2.3.10-dev

brianteeman:
see #38119

PR #38119, opened 2022-06-22
[4.2] cli update extension check

brianteeman:
Pull Request for Issue #37643 .

### Summary of Changes
The cache timeout was incorrectly set at 3600 x 6 HOURS

Install older versions of a few extensions
Check that the updates are found 

image

From the cli run

`php cli/joomla.php update:extensions:check`

### Expected result
 ! [NOTE] There are available updates to apply

image

### Actual result
 [OK] There are no available updates                 
image

### DIFF
--- a/libraries/src/Console/CheckUpdatesCommand.php
+++ b/libraries/src/Console/CheckUpdatesCommand.php
@@ -52,7 +52,7 @@ class CheckUpdatesCommand extends AbstractCommand
 		// Get the update cache time
 		$component = ComponentHelper::getComponent('com_installer');
 
-		$cache_timeout = 3600 * (int) $component->getParams()->get('cachetimeout', 6);
+		$cache_timeout = (int) $component->getParams()->get('cachetimeout', 6);
 
 		// Find all updates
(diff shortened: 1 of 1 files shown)


Quy:
I have tested this item :white_check_mark: successfully on 036d85cb552b174b9c5d0dd006af6611a17e094f<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38119">issues.joomla.org/tracker/joomla-cms/38119</a>.</sub>

alikon:
humm..
the cahce timeout is set correctly 6 hours  x 3600 seconds = 21600 seconds (both cli & admin)

https://github.com/joomla/joomla-cms/blob/54ed9907d08908f4163cc7b9caddbb2ab6904dd0/libraries/src/Updater/Updater.php#L110

therefore this pr seems to works cause you are setting 6 seconds

in the admin side the cached results are covered by the component querying tables

imho on the cli side we don't need cached results

so  my choiche would be https://github.com/brianteeman/joomla-cms/pull/271

Issue #38117, opened 2022-06-22
<supported_databases> in updates XML always results in an error (which also has wrong wording).


nikosdion:
Ever since @richard67 introduced me to the 2-in-1 bug specials I only find this kind of bugs. This morning we will see that an update with `<supported_databases>` always results in an error and the error message is misleading.

### Steps to reproduce the issue

* Install an older version of an extension with `<supported_databases>` in its XML update file, e.g. https://github.com/lucid-fox/social-magick/releases/download/1.0.1/socialmagick-1.0.1.zip
* Go to System, Update, Extensions
* Click on Check for Updates

### Expected result

Update found, no warning.

### Actual result

A bunch of **misleading** warnings:

* For the extension Social Magick version 1.0.2 is available, but your current database MySQL (PDO) is not supported anymore.
* For the extension Social Magick version 1.0.1 is available, but your current database MySQL (PDO) is not supported anymore.
* For the extension Social Magick version 1.0.0 is available, but your current database MySQL (PDO) is not supported anymore.

### System information (as much as possible)

Joomla 4.1.5 on PHP 8.1 but this might be irrelevant as the issue seems to exist since #12355

### Additional comments

They say a picture is worth one thousand words. I agree:
<img width="802" alt="Screenshot 2022-06-22 at 09 29 03" src="https://user-images.githubusercontent.com/256041/174962139-57b1dd10-d09f-4c31-9bb9-279434976893.png">

In libraries/src/Updater/Adapter/ExtensionAdapter.php:159 we are trying to see if the site's database connector object's `$dbType` (all **lowercase**) matches one of the keys of the `$supportedDbs` array. However, the keys in the `$supportedDbs` array are **upppercase**. Therefore we will _never_ match a database type...

This is bug number one.

We then display the JLIB_INSTALLER_AVAILABLE_UPDATE_DB_MINIMUM message _but_ instead of reporting the database _server type_ (e.g. MySQL) we report the database _connector name_ (e.g. MySQLi or PDO MySQL). This causes the asinine and misleading messages sampled above.

This is bug number two.

#### How to fix this

For the first bug change line 159 from
```php
if (\array_key_exists($dbType, $supportedDbs))
```
to
```php
if (in_array(strtolower($dbType), array_map('strtolower', array_keys($supportedDbs))))
```

For the second bug, change lines 167 to 174 from
```php
								$dbMsg = Text::sprintf(
[...]

PR #38121, opened 2022-06-22
in updates XML always resulted in an error

nikosdion:
Pull Request for Issue #38117 .

### Summary of Changes

Updates `\Joomla\CMS\Updater\Adapter\ExtensionAdapter` to support the `<supported_databases>` tag in XML update files correctly.

Moreover, fixes the message when there is no supported database to display the DB server technology type, not the DB connector type (therefore mysqli and pdomysql / mysql will all be referred to as “MySQL”, whereas the same connectors on a MariaDB server will be referred to as “MariaDB”).

### Testing Instructions

* Install an older version of an extension with <supported_databases> in its XML update file, e.g. https://github.com/lucid-fox/social-magick/releases/download/1.0.1/socialmagick-1.0.1.zip
* Go to System, Update, Extensions
* Click on Check for Updates

A bunch of misleading warnings:

* For the extension Social Magick version 1.0.2 is available, but your current database MySQL (PDO) is not supported anymore.
* For the extension Social Magick version 1.0.1 is available, but your current database MySQL (PDO) is not supported anymore.
* For the extension Social Magick version 1.0.0 is available, but your current database MySQL (PDO) is not supported anymore.

Update found, no warnings.

None

### Translation impact

Added six new translation strings: JLIB_DB_SERVER_TYPE_MARIADB, JLIB_DB_SERVER_TYPE_MYSQL, JLIB_DB_SERVER_TYPE_POSTGRESQL, JLIB_DB_SERVER_TYPE_ORACLE, JLIB_DB_SERVER_TYPE_SQLITE, JLIB_DB_SERVER_TYPE_MSSQL

However, these strings are product names and do not need to be translated; they just need to be copied verbatim into language files by the translation teams.

### DIFF
--- a/administrator/language/en-GB/lib_joomla.ini
+++ b/administrator/language/en-GB/lib_joomla.ini
@@ -758,3 +758,11 @@ JLIB_SIZE_PB="PiB"
 JLIB_SIZE_EB="EiB"
 JLIB_SIZE_ZB="ZiB"
 JLIB_SIZE_YB="YiB"
+
+; Database server technology types in human readable terms. Used in the Updater package.
+JLIB_DB_SERVER_TYPE_MARIADB="MariaDB"
+JLIB_DB_SERVER_TYPE_MSSQL="Microsoft SQL Server"
+JLIB_DB_SERVER_TYPE_MYSQL="MySQL"
+JLIB_DB_SERVER_TYPE_ORACLE="Oracle"
+JLIB_DB_SERVER_TYPE_POSTGRESQL="PostgreSQL"
+JLIB_DB_SERVER_TYPE_SQLITE="SQLite"
--- a/libraries/src/Updater/Adapter/ExtensionAdapter.php
+++ b/libraries/src/Updater/Adapter/ExtensionAdapter.php
@@ -155,10 +155,13 @@ class ExtensionAdapter extends UpdateAdapter
 							}
 						}
 
+						// $supportedDbs has uppercase keys because they are XML attribute names
+						$dbTypeUcase = strtoupper($dbType);
+
 						// Do we have an entry for the database?
-						if (in_array($dbType, array_map('strtolower', array_keys($supportedDbs))))
+						if (\array_key_exists($dbTypeUcase, $supportedDbs))
 						{
-							$minimumVersion = $supportedDbs[$dbType];
+							$minimumVersion = $supportedDbs[$dbTypeUcase];
 							$dbMatch        = version_compare($dbVersion, $minimumVersion, '>=');
 
 							if (!$dbMatch)
@@ -168,7 +171,7 @@ class ExtensionAdapter extends UpdateAdapter
 									'JLIB_INSTALLER_AVAILABLE_UPDATE_DB_MINIMUM',
 									$this->currentUpdate->name,
 									$this->currentUpdate->version,
-									Text::_($db->name),
+									Text::_('JLIB_DB_SERVER_TYPE_' . $dbTypeUcase),
 									$dbVersion,
 									$minimumVersion
 								);
@@ -183,7 +186,7 @@ class ExtensionAdapter extends UpdateAdapter
 								'JLIB_INSTALLER_AVAILABLE_UPDATE_DB_TYPE',
 								$this->currentUpdate->name,
 								$this->currentUpdate->version,
-								Text::_('JLIB_DB_SERVER_TYPE_' . $dbType)
+								Text::_('JLIB_DB_SERVER_TYPE_' . $dbTypeUcase)
 							);
 
 							Factory::getApplication()->enqueueMessage($dbMsg, 'warning');
--- a/libraries/src/Updater/Adapter/ExtensionAdapter.php
+++ b/libraries/src/Updater/Adapter/ExtensionAdapter.php
@@ -156,7 +156,7 @@ class ExtensionAdapter extends UpdateAdapter
 						}
 
 						// Do we have an entry for the database?
-						if (\array_key_exists($dbType, $supportedDbs))
+						if (in_array($dbType, array_map('strtolower', array_keys($supportedDbs))))
 						{
 							$minimumVersion = $supportedDbs[$dbType];
 							$dbMatch        = version_compare($dbVersion, $minimumVersion, '>=');
@@ -183,7 +183,7 @@ class ExtensionAdapter extends UpdateAdapter
 								'JLIB_INSTALLER_AVAILABLE_UPDATE_DB_TYPE',
 								$this->currentUpdate->name,
 								$this->currentUpdate->version,
-								Text::_($db->name)
+								Text::_('JLIB_DB_SERVER_TYPE_' . $dbType)
 							);
 
 							Factory::getApplication()->enqueueMessage($dbMsg, 'warning');


administrator/language/en-GB/lib_joomla.ini
```diff
@@ -758,3 +758,11 @@ JLIB_SIZE_PB="PiB"
 JLIB_SIZE_EB="EiB"
 JLIB_SIZE_ZB="ZiB"
 JLIB_SIZE_YB="YiB"
+
+; Database server technology types in human readable terms. Used in the Updater package.
+JLIB_DB_SERVER_TYPE_MARIADB="MariaDB"
```
brianteeman, line 763: do you need to create new strings? These all already exist in joomla.ini

nikosdion: They do not, really.

The strings in `joomla.ini` are for the PHP database connectors. The strings I added are for database products.

The difference is easy to spot with MySQL and PostgreSQL.

The `joomla.ini` file has strings for MySQLi (note the `i` at the end), PDO MySQL, PostgreSQL and PDO PostgreSQL.

However, the database technologies are MySQL (no trailing `I`) and PostgreSQL. Moreover, when talking about MySQL databases, Joomla distinguishes between MariaDB and MySQL (proper and derivatives, such as Percona), targeting them differently.

A developer might choose to support MySQL proper _but not MariaDB_, or vice versa. In both cases the site would be connected to the DB server using, say, the MySQLi connector. So the user would be told that the developer no longer supports the MySQLi connector which is a lie.

So, yes, we do need to have different strings for PHP database connectors and database product names. They are not the same.

richard67:
> However, these strings are product names and do not need to be translated; they just need to be copied verbatim into language files by the translation teams.

In most cases this might be right, but in some languages like e.g. Chinese there **_might_** be transcripts of product names in their writing system, so I think it's good to have them translatable.

nikosdion:
@richard67 They are language strings already, as I said. If a language needs to transliterate the product names, they can.

richard67:
> @richard67 They are language strings already, as I said. If a language needs to transliterate the product names, they can.

@nikosdion Yes, I understood that. I only wanted to say why I think this PR is really useful. Am preparing for tests now. Will create some own updatesite for that extension and modify it to use that so I can play around a bit with the supported_databases values.

nikosdion:
@richard67 Oh! Sorry, I misunderstood what you wrote before. All clear now :)

richard67:
Unfortunately the extension linked in the description doesn't have 4.2 yet in the targetplatform tag of the update site https://raw.githubusercontent.com/lucid-fox/social-magick/main/update/socialmagick.xml , so on a 4.2-dev the test will not work as described with that extension. So I have to create an own update site anyway.

brianteeman:
>  So I have to create an own update site anyway.

or modify the version number of your install. One file to edit. Very quick

richard67:
> > So I have to create an own update site anyway.
> 
> or modify the version number of your install. One file to edit. Very quick

@brianteeman For that purpose only yes, but I also want to play with the supported_databases values of the update site.

nikosdion:
@richard67 You can download the update XML and host it locally, amending the supported platforms. Then edit your DB to point its update site to your local file and Bob's your uncle.

richard67:
> @richard67 You can download the update XML and host it locally, amending the supported platforms. Then edit your DB to point its update site to your local file and Bob's your uncle.

@nikosdion Yes, I'm doing it that way, change the URL of the update site in the db and also in the manifest XML of the plugin.

What I found out up to now is that your PR works as described, but I did not get the version check work up to now here https://github.com/joomla/joomla-cms/blob/e7e376547471f81a57d9b4ce669309613478e5c7/libraries/src/Updater/Adapter/ExtensionAdapter.php#L162 .

I had a `mysql="9.0.0"` in the `supported_databases` for the 1.0.2 version of the plugin, but I was offered the update.

When I removed the "mysql" completely with `<supported_databases oracle="10.1" />` for plugin version 1.0.2, then I got the expected message telling that my database type "MySQL" is not supported anymore.

Am investigating what the reason could be. Will let you know soon.

richard67:
@nikosdion The problem is that the array keys of `$supportedDbs` are in uppercase for whatever reason, while `$dbType` is lowercase. You catch that well in line 159, but in line 161 the `$minimumVersion = $supportedDbs[$dbType];` fails.

If we can rely on the keys always being uppercase we could just change that to `$minimumVersion = $supportedDbs[strtoupper($dbType)];`, but maybe it would be better just to map or copy the `$supportedDbs` so it has lowercase keys?

P.S. What we should not to is to change `$dbType` to uppercase.

richard67:
Hmm, it is an XML element attribute ... if we can rely on it always being made uppercase by the XML parser, we can just use `$supportedDbs[strtoupper($dbType)];` in line 161.

libraries/src/Updater/Adapter/ExtensionAdapter.php
```diff
@@ -156,7 +156,7 @@ protected function _endElement($parser, $name)
 						}
 
 						// Do we have an entry for the database?
-						if (\array_key_exists($dbType, $supportedDbs))
+						if (in_array($dbType, array_map('strtolower', array_keys($supportedDbs))))
 						{
 							$minimumVersion = $supportedDbs[$dbType];
```
richard67:
```suggestion
						$dbTypeUcase = strtoupper($dbType);

						if (\array_key_exists($dbTypeUcase, $supportedDbs))
						{
							$minimumVersion = $supportedDbs[$dbTypeUcase];
```
If we do it this way, we can also use the `$dbTypeUcase` below for making the language string key.

libraries/src/Updater/Adapter/ExtensionAdapter.php
```diff
@@ -183,7 +183,7 @@ protected function _endElement($parser, $name)
 								'JLIB_INSTALLER_AVAILABLE_UPDATE_DB_TYPE',
 								$this->currentUpdate->name,
 								$this->currentUpdate->version,
-								Text::_($db->name)
+								Text::_('JLIB_DB_SERVER_TYPE_' . $dbType)
```
richard67:
```suggestion
								Text::_('JLIB_DB_SERVER_TYPE_' . $dbTypeUcase)
```
Can be done if my previous suggestion gets accepted.

richard67:
@nikosdion In addition to my 2 previous review comments, I found that it needs to change the error message also here  https://github.com/joomla/joomla-cms/blob/e7e376547471f81a57d9b4ce669309613478e5c7/libraries/src/Updater/Adapter/ExtensionAdapter.php#L171 from `Text::_($db->name),` to `Text::_('JLIB_DB_SERVER_TYPE_' . $dbTypeUcase),` if you follow my suggestions, or to `Text::_('JLIB_DB_SERVER_TYPE_' . $dbType),` otherwise.

richard67:
For testing I have created a few update sites:

- For testing as described on 4.2-dev https://test5.richard-fath.de/test_pr-38121.xml

- For database type being not supported https://test5.richard-fath.de/test_pr-38121_test-2.xml - This has `<supported_databases oracle="12.0" />`, so you can test with whatever you have and it will not be supported.

- For database version being not supported https://test5.richard-fath.de/test_pr-38121_test-3.xml - This has `<supported_databases mysql="10.0.0" mariadb="20.0"  postgresql="20.0" />`, so you can test with whatever you have and it will be a too old version of a supported type.

Just change the `location` column of the record with name 'Social Magick Updates' in the `#__update_sites` table to the mentioned URL, and to be sure it survives an update site rebuild change it also in line 637 of file "plugins/system/socialmagick/socialmagick.xml" of your testing site.

richard67:
Another thing I found is that there is very similar functionality here https://github.com/joomla/joomla-cms/blob/4.2-dev/libraries/src/Updater/Update.php#L357-L388 . I'm not sure, but I think that's used for the core update. It would make sense to use the new language strings also there when this PR here has been merged. I could make a follow up or for that in this case.

Finally, as the language strings are only used for reporting the current database type on which Joomla is running, and not the supported one from the update site, and Joomla 4 runs only on MySQL or PostgreSQL, we currently do not really need the language strings for the other database types (Oracle, MS SQL, SQLite). But as the drivers support to connect with them, I think it is good to have them for future use at other places. But that's just my opinion, others might disagree.

richard67:
@nikosdion If you want I can make a PR for you in your repo to apply my above suggestions.

nikosdion:
Yup, please do a PR.

As for the lang strings, best to have all keys than ending up with an untranslated string 5 years down the line if, say, we decide that a site can have multiple DB connections and people write integration plugins which talk to the secondary database which might be of a different technology than what is supported for the primary database.

richard67:
> Yup, please do a PR.

@nikosdion Done, see https://github.com/nikosdion/joomla-cms/pull/10 .

Regarding the language strings I fully agree with you. I just mentioned it in case others may disagree.

nikosdion:
Thank you! I merged your changes.

richard67:
I have tested this item :white_check_mark: successfully on a4be380722e4e79a0234e574bf7a14314ac278f8<br /><br />I have tested as follows.

I have created 3 update sites for 3 tests:

1. https://test5.richard-fath.de/test_pr-38121.xml for testing the PR as currently described in the instructions on a 4.2-dev.
2. https://test5.richard-fath.de/test_pr-38121_test-2.xml for having no supported database type.
3. https://test5.richard-fath.de/test_pr-38121_test-3.xml for having a supported database type but with unsupported version.

Then I have installed the plugin linked in the description.

Then I have updated the `location` column of the record with name 'Social Magick Updates' in the `#__update_sites` table to the URL of the particular update site and then have used the "Check For Updates" button.

I've done this for all 3 sites without and with the PR applied.

Results will come with screenshot in my next comment.

Other testers please follow my inscrutions above.

<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38121">issues.joomla.org/tracker/joomla-cms/38121</a>.</sub>

richard67:
Here the details to my test results for the 3 update sites I've mentioned in the comment to my test.

Without the PR applied, I get with all 3 update sites the same result as described in the issue:
test-pr-38121_before

With the PR applied, everything is as expected:

When using the first update site, an update is found:
test-pr-38121_after-1

When using the second update site, I get a message that the database is not supported anymore:
test-pr-38121_after-2

When using the third update site, I get a message that the database **_version_** is not supported anymore:
test-pr-38121_after-3

Quy:
I have tested this item :white_check_mark: successfully on a4be380722e4e79a0234e574bf7a14314ac278f8<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38121">issues.joomla.org/tracker/joomla-cms/38121</a>.</sub>

Quy:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38121">issues.joomla.org/tracker/joomla-cms/38121</a>.</sub>

HLeithner:
thanks

Issue #15833, opened 2017-05-05
Better SEO options in edit articles


JoomlaFan:
I think this may improve the SEO option, grouped in one tab.

![screen shot 2017-05-05 at 11 20 34](https://issues.joomla.org/uploads/1/ba939b87f14827f9a64bf859ebc9b87c.png)

tomekomel:
Nice improvement!<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/15833">issues.joomla.org/tracker/joomla-cms/15833</a>.</sub>

brianteeman:
I dont personally see the benefit of moving these options from the publishing tab to a dedicated tab - but if you do then feel free to submit a PR to make this change. Note that for consistency you would need to make the same change in all the core components. Also you should ensure that the option to hide the publishing tab still works etc

dgt41:
8 tabs that someone needs to go through to publish an article? Really?
We are doing it wrong!

bertmert:
Why ist this when or where "SEO"???

- Author
- Content Rights
- Browser Page Title
...

I think this should not go into core. It's easy to handle things like this with a short custom plugin.

brianteeman:
@bertmert this is just moving items from the publishing tab to their own tab

bertmert:
@brianteeman 

> this is just moving items from the publishing tab to their own tab

Yes, I know. But these options are not all "SEO" or not only "SEO" or maybe sometimes "SEO". The tabulator label is misleading nonsense.

JoomlaFan:
These options are now scattered. For example, see where in Joomla 3.7 was added override the article title meta - in the article parameters tab - at the bottom. It is really difficult to find in the article options. If this is not improved, the opinion about SEO in Joomla will still be bad.

Meta title, description, keywords and robots should be together. Not necessarily on one card. But a separate tab is in my opinion the best one.

Sorry for my english ;-)<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/15833">issues.joomla.org/tracker/joomla-cms/15833</a>.</sub>

photodude:
I'm not for moving it to a new tab, but not against grouping as an "SEO" section  under the existing publishing tab.

Quy:
Similar to #12043

[...]

PR #38124, opened 2022-06-22
Added Separate Section For SEO

pritam825:
It's Not an Issue, It's an suggestion

### Testing Instructions
go to administrator->Banners->Banners->New

No Separate Section for SEO and  Metadata section is in publishing section

image

### 
If this is ok for everyone then will create for others also.


brianteeman:
None of those fields are anything to do with SEO

pritam825:
> None of those fields are anything to do with SEO

Metadata comes under seo

brianteeman:
You should look at how the data is used. None of those fields are used for seo

pritam825:
@brianteeman 
image
so, according to you this should be in separate section because of meta description part?

Quy:
See discussion #15833

pritam825:
@Quy I have read the discussion but not seeing any end-up of discussion , some people says it's good , some not, if you all are ok with this then I will create pr for every component, otherwise I will close this, Thanks :)

Quy:
I don't see the point of moving just 3 fields to its own tab especially if they are not updated often.
What is your rationale to justify its own tab?
Why not do a text override to "Publishing/SEO" in your installation?

pritam825:
> I don't see the point of moving just 3 fields to its own tab especially if they are not updated often. What is your rationale to justify its own tab? Why not do a text override to "Publishing/SEO" in your installation?

For example if I will have to add the meta description or keywords so first thing which comes in mind that there will be an separate SEO section where I can find the all functionality for SEO, how will I know that it will come under publishing

Quy:
It may not be obvious at first, but once you know where these settings are under `Publishing` then I don't see the issue. Create a language override if you need a reminder.

HLeithner:

web-eau-net:
I have tested this item :white_check_mark: successfully on 1e8b03b7256b2e602d1f41fadbfd9898cab0b1fb<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38124">issues.joomla.org/tracker/joomla-cms/38124</a>.</sub>

Quy:
Thank you for the PR. Closing for now for reason stated. When there are additional SEO settings added, we will revisit this.

PR #38128, opened 2022-06-22
[4.4] package.json

brianteeman:
in another project I maintain we always update the version number here with each release


richard67:
Shouldn't this then be done either by the build or by the bump script?

brianteeman:
yes it should be automated - _if maintainers think it should be changed at all_ which I wasnt sure about as its never been updated

brianteeman:
rebased to 4.4

laoneo:
This should either be bumped with the bump script or removed all together as the CMS is not a library where other npm packages depend on.

brianteeman:
Updated as requested to be managed by the bump script #40513

PR #38131, opened 2022-06-23
Update ExtensionAdapter.php

mgilkes:
Line 142 needs to be reverted back to 'strtoupper'. It can't be 'strtolower' as the keys in array $supportedDbs are all upper case. Thus, when checking for the $dbType (value of $dbType was 'mysql' in my testing)  in $supportedDbskey array (
  'MYSQL' => '5.5.3',
  'POSTGRESQL' => '9.1',
  'MSSQL' => '10.50.1600.1',
  'MARIADB' => '10.1',
), it won't be found, because the case is different. The user will see the message: "For the extension SOME_EXTENSION_NAME_VERSION_XXX is available, but your current database MySQLi is not supported anymore."

Pull Request for Issue # .

### Summary of Changes

On line 142, revert "strtolower" to "strtoupper"

### Testing Instructions

Install an extension that has the following line in its update file: "<supported_databases mysql="5.5.3" postgresql="9.1" mssql="10.50.1600.1" mariadb="10.1"/>", and make sure your mysql database version is greater than 5.5.3 (most are 5.7+), then go to System > Extensions Update, then click on "Check for Updates"

You will get a notice which says "For the extension SOME_EXTENSION_NAME_VERSION_XXX is available, but your current database MySQLi is not supported anymore." You are unable to update the extension via the updater. So the site admin thinks that the extension is not compatible any more, but this is not the case. 

No notice is shown, and you can update the extension via the updater.

### DIFF
--- a/libraries/src/Updater/Adapter/ExtensionAdapter.php
+++ b/libraries/src/Updater/Adapter/ExtensionAdapter.php
@@ -139,7 +139,7 @@ class ExtensionAdapter extends UpdateAdapter
 					if (isset($this->currentUpdate->supported_databases))
 					{
 						$db           = Factory::getDbo();
-						$dbType       = strtolower($db->getServerType());
+						$dbType       = strtoupper($db->getServerType());
 						$dbVersion    = $db->getVersion();
 						$supportedDbs = $this->currentUpdate->supported_databases;
 


richard67:
Please check if #38121 solves this in a better way, and if so, close your PR in favor of that one.

richard67:
P.S. In general it is expected that before you make a PR you check if there is already one.

richard67:
P.P.S.: it is also expected that you test your change before making a PR. Your code cannot work without any additional changes, see just one example here:  https://github.com/joomla/joomla-cms/blob/f85324ae4c1b3551df83f15d12acb13a86a68e0d/libraries/src/Updater/Adapter/ExtensionAdapter.php#L147 . There are more of that fir her down in the code. So finally I would say this PR cannot be accepted as it it because it cannot work and so for sure has not been tested right by the author.

mgilkes:
@richard67 Thanks for reviewing my pull request. You are right. #38121 solves this issue in a better way.

richard67:
@mgilkes Thanks for checking and reporting back.

PR #38133, opened 2022-06-24
Skip Session Manager tests on PHP 8.1 as old version of PHPUnit

wilsonge:
Works around https://github.com/sebastianbergmann/phpunit/issues/4879 by skipping these tests for now. At least the remaining tests after this patch run successfully

### DIFF
--- a/tests/Unit/Libraries/Cms/Session/SessionManagerTest.php
+++ b/tests/Unit/Libraries/Cms/Session/SessionManagerTest.php
@@ -46,6 +46,7 @@ class SessionManagerTest extends UnitTestCase
 	 */
 	protected function setUp(): void
 	{
+		// @todo remove this after upgrading phpunit to 9+ see https://github.com/sebastianbergmann/phpunit/issues/4879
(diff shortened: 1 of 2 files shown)


tests/Unit/Libraries/Cms/Session/SessionManagerTest.php
```diff
@@ -46,6 +46,15 @@ class SessionManagerTest extends UnitTestCase
 	 */
 	protected function setUp(): void
 	{
+		if (version_compare(phpversion(), '8.1.0', '>='))
```
HLeithner, line 49: // @todo remove this after upgrading phpunit to 9+ see https://github.com/sebastianbergmann/phpunit/issues/4879

HLeithner, line 50:
```suggestion
		// @todo remove this after upgrading phpunit to 9+ see https://github.com/sebastianbergmann/phpunit/issues/4879
		if (version_compare(phpversion(), '8.1.0', '>='))
```

HLeithner:
Thanks

PR #38156, opened 2022-06-27
[4.2] codemirror update

brianteeman:
update to v 5.65.6

20-06-2022: [Version 5.65.6](https://codemirror.net/5/codemirror-5.65.6.zip):

- Avoid firing beforeCursorEnter callbacks twice for cursor selections.
- Improve support for auto-hiding macOS scrollbars.
- [show-hint addon](https://codemirror.net/5/doc/manual.html#addon_show-hint): Fix an issue where the tooltip could be placed to the left of the screen.
- [swift mode](https://codemirror.net/5/mode/swift): Support structured concurrency keywords.


richard67:
Thanks. Merging by review and after having tested that codemirror still works.

PR #38166, opened 2022-06-28
[4.2] remove ruleset.xml

brianteeman:
This pr adds ruleset.xml to the list of files that will not be distributed
 code review


build/build.php
```diff
@@ -394,6 +394,7 @@ function clean_composer(string $dir)
 	'plugins/sampledata/testing/testing.xml',
 	'plugins/sampledata/testing/language/en-GB/en-GB.plg_sampledata_testing.ini',
 	'plugins/sampledata/testing/language/en-GB/en-GB.plg_sampledata_testing.sys.ini',
+    'ruleset.xml',
```
richard67, line 391:
```suggestion
	'ruleset.xml',
```
As the build.php is not psr12 yet we should use a tab here like in the lines above and below, and not 4 spaces.

brianteeman: You cant do that. .editorconfig contains the rules that ide should follow. It will be a nightmare if we dont follow it.

richard67: The build folder is configured to be skipped in the ruleset.xml!!! The file is still with tabs. Check the diff on GH and you will see.

brianteeman: I am changing it but this is an issue that needs to be resolved by the maintainers

richard67: Yes, I agree. I have already started to discuss that we should convert the code style of the PHP files in the build folder, even if we not include it in the automated checks for some reason.

richard67: For the code style thing see #38167 .

HLeithner:
sorry with merging https://github.com/joomla/joomla-cms/pull/38167 I broke this pr...

richard67:
@HLeithner @brianteeman I've fixed the conflict so the PR is ok and uses psr-12.

HLeithner:
thanks

brianteeman:
thanks

PR #38167, opened 2022-06-28
Change PHP code style to PSR-12 for the build folder

richard67:
Pull Request for Issue # .

### Summary of Changes

Converts the PHP files in the build folder to PSR-12 and but does **_not_** include them into the cs check yet.

The files in subfolder "psr12" are already PSR-12, but this PR fixes a few smaller code style errors in these files, too.

This PR here makes sure that people who use the right .editorconfig for PSR-12 can edit the files in the build folder without getting mad.

### Testing Instructions

On a clean, current 4.2-dev run `composer install` if not done before.

Then run `./libraries/vendor/bin/phpcs --extensions=php -p --standard=ruleset.xml ./build`

Lots of code style errors and warnings.

Only the following warning is remaining for some files:

```
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------------------------------------
 1 | WARNING | A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it should execute logic
   |         | with side effects, but should not do both. The first symbol is defined on line 25 and the first side effect is on line 26.
------------------------------------------------------------------------------------------------------------------------------------------------------
```

None.

### DIFF
--- a/build/build.php
+++ b/build/build.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Script used to build Joomla distribution archive packages
  * Builds packages in tmp/packages folder (for example, 'build/tmp/packages')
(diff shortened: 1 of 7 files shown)


HLeithner:
thanks

laoneo:
You can add there like what I did in https://github.com/joomla/joomla-cms/blob/4.2-dev/tests/Unit/bootstrap.php#L13

laoneo:
I would also add it to the CS check as this should be for all files.

richard67:
@laoneo Ok, I will make a follow up PR. Thanks for the hint.

HLeithner:
When we do exceptions I think we should describe the reason for the code style exception, especially for then `PSR1.Files.SideEffects` which has security implications.

richard67:
> When we do exceptions I think we should describe the reason for the code style exception, especially for then `PSR1.Files.SideEffects` which has security implications.

@HLeithner For the scripts in the build folder we can add a comment that they are not shipped and so not relevant.

laoneo:
Same for tests

richard67:
I think we should not do that in particular files but have exclusion sets for that rule in the ruleset.xml for the build and the tests folder.

laoneo:
I would not do that for a particular folder, more for the individual files. For the build and tests folder should also go under the same standard as the rest of the code base.

richard67:
> I would not do that for a particular folder, more for the individual files. For the build and tests folder should also go under the same standard as the rest of the code base.

@laoneo Then why have you added an exception for a folder here with PR #38161 ? https://github.com/joomla/joomla-cms/blob/4.2-dev/ruleset.xml#L247

HLeithner:
In the end we should have "no" exceptions, of course that's not possible but I'm pretty sure we now have a punch of cs ignore rules in our codebase which could be removed after the move to the psr-12 standard.

HLeithner:
When someone has time we have to reduce the char count to about 150 but that wasn't possible with our current code base so I set the longest line I found.

richard67:
I think we should do that for particular rules in the ruleset.xml because there are already comments for particular sections, like `<!-- expected, may move exception to file or refactor implementation -->`. I will make my PR like that.

HLeithner:
what pr would you like to create?!

richard67:
The one to include the build folder into the regular tests, i.e. remove it from the exception in ruleset.xml

richard67:
@HLeithner See #38168 .

richard67:
My PR #38168 worked locally on Linux and failed in Drone (also on Linux), and I have no idea why, so I gave it up.

PR #38168, opened 2022-06-28
Add build folder to PHPCS checks

richard67:
Pull Request for Issue # .

### Summary of Changes

Removes the build folder from the exceptions in ruleset.xml so the PHPCS checks are run on that folder by default, too, and adds the necessary exceptions to the "PSR1.Files.SideEffects.FoundWithSymbols" rule to the right section in the ruleset.xml file.

In addition, the exception to that rule for the "tests/Unit/bootstrap.php" file has been moved from the file to the ruleset.xml.

### Testing Instructions

On a clean, current 4.2-dev run `composer install` if not done before.

Then run `./libraries/vendor/bin/phpcs --extensions=php -p --standard=ruleset.xml .`

The "./build" folder is not checked. There are no errors or warnings.

When running `./libraries/vendor/bin/phpcs --extensions=php -p --standard=ruleset.xml ./build`, the build folder will be checked, and there will be warnings as mentioned in PR #38167 .

The "./build" folder is checked. There are no errors or warnings.

None.

### DIFF
--- a/ruleset.xml
+++ b/ruleset.xml
@@ -293,5 +293,6 @@
         <exclude-pattern type="relative">libraries/import\.php</exclude-pattern>
         <exclude-pattern type="relative">libraries/import\.legacy\.php</exclude-pattern>
         <exclude-pattern type="relative">libraries/loader\.php</exclude-pattern>
+        <exclude-pattern type="relative">tests/Unit/bootstrap\.php</exclude-pattern>
     </rule>
 </ruleset>
--- a/ruleset.xml
+++ b/ruleset.xml
@@ -243,13 +243,7 @@
         <exclude-pattern type="relative">plugins/content/pagebreak/pagebreak\.php</exclude-pattern>
         <exclude-pattern type="relative">plugins/editors/none/none\.php</exclude-pattern>
         <exclude-pattern type="relative">plugins/user/joomla/joomla\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Codeception/acceptance/administrator/components/com_content/ContentCategoryListCest\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Codeception/acceptance/administrator/components/com_content/ContentListCest\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Codeception/acceptance/administrator/components/com_content/ContentFieldListCest\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Codeception/api/com_contact/ContactCest\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Codeception/api/com_banners/BannerCest\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Codeception/api/com_content/ContentCest\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/Unit/Libraries/Cms/MVC/Model/DatabaseModelTest\.php</exclude-pattern>
+        <exclude-pattern type="relative">tests/</exclude-pattern>
     </rule>
 
     <rule ref="PSR1.Classes.ClassDeclaration.MultipleClasses">
--- a/ruleset.xml
+++ b/ruleset.xml
@@ -3,7 +3,6 @@
     <description>The Joomla CMS PSR-12 exceptions.</description>
 
     <!-- Exclude folders not containing production code -->
-    <exclude-pattern type="relative">build/*</exclude-pattern>
     <exclude-pattern type="relative">cache/*</exclude-pattern>
     <exclude-pattern type="relative">docs/*</exclude-pattern>
     <exclude-pattern type="relative">logs/*</exclude-pattern>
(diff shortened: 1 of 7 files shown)


ruleset.xml
```diff
@@ -271,6 +276,11 @@
         <exclude-pattern type="relative">api/includes/defines\.php</exclude-pattern>
         <exclude-pattern type="relative">api/includes/framework\.php</exclude-pattern>
         <exclude-pattern type="relative">api/index\.php</exclude-pattern>
+        <exclude-pattern type="relative">build/build\.php</exclude-pattern>
```
laoneo, line 273: Can this file be added here and the inline check being removed https://github.com/joomla/joomla-cms/blob/4.2-dev/tests/Unit/bootstrap.php#L13?

richard67: Yes, will do so.

richard67: Done. Test still succeeds.

ruleset.xml
```diff
@@ -244,7 +243,13 @@
         <exclude-pattern type="relative">plugins/content/pagebreak/pagebreak\.php</exclude-pattern>
         <exclude-pattern type="relative">plugins/editors/none/none\.php</exclude-pattern>
         <exclude-pattern type="relative">plugins/user/joomla/joomla\.php</exclude-pattern>
-        <exclude-pattern type="relative">tests/</exclude-pattern>
```
laoneo: I'v added it here, because every time when _before is implemented, the file needs to be added here.

richard67:
The strange thing is that PHPCS fails in drone here for this PR: https://ci.joomla.org/joomla/joomla-cms/56045/1/6 . But when I run the test locally it works. @HLeithner Do you have an idea what the problem could be? My PR doesn't even touch that "build/psr12/phpcs.joomla.report.php" file.

ruleset.xml
HLeithner: no please don't do that, that's exactly what I mean in the other pr. We should explicit write the reason for this exception into the file because this file doesn't have a jexec || die and if it has one it means it is mixed with executable code and that's a security issue which we should get rid of (the bootstrap file with constant definition and if statements is an exception) but for example all files in the  should be fixed.

I wasn't able to do this in the initial pr because it may great merge conflicts which I tried to avoid.

HLeithner:
> The strange thing is that PHPCS fails in drone here for this PR: https://ci.joomla.org/joomla/joomla-cms/56045/1/6 . But when I run the test locally it works. @HLeithner Do you have an idea what the problem could be? My PR doesn't even touch that "build/psr12/phpcs.joomla.report.php" file.

it's possible that the file pattern matcher works different under windows and linux

ruleset.xml
richard67: > all files in the should be fixed.

@HLeithner All files in the what?

HLeithner: lol sorry
< !-- should be reconsidered -- >

richard67:
> > The strange thing is that PHPCS fails in drone here for this PR: https://ci.joomla.org/joomla/joomla-cms/56045/1/6 . But when I run the test locally it works. @HLeithner Do you have an idea what the problem could be? My PR doesn't even touch that "build/psr12/phpcs.joomla.report.php" file.
> 
> it's possible that the file pattern matcher works different under windows and linux

@HLeithner Here I have Linux, and Drone where it fails has Linux, too, or am I wrong?

HLeithner:
> > > The strange thing is that PHPCS fails in drone here for this PR: https://ci.joomla.org/joomla/joomla-cms/56045/1/6 . But when I run the test locally it works. @HLeithner Do you have an idea what the problem could be? My PR doesn't even touch that "build/psr12/phpcs.joomla.report.php" file.
> > 
> > 
> > it's possible that the file pattern matcher works different under windows and linux
> 
> @HLeithner Here I have Linux, and Drone where it fails has Linux, too, or am I wrong?

hard to say can you fix the file?

ruleset.xml
richard67: So what shall I do now? I have my stuff in the `<!-- expected, may move exception to file or refactor implementation -->` section, I thought that would be suitable.

richard67:
> > > > The strange thing is that PHPCS fails in drone here for this PR: https://ci.joomla.org/joomla/joomla-cms/56045/1/6 . But when I run the test locally it works. @HLeithner Do you have an idea what the problem could be? My PR doesn't even touch that "build/psr12/phpcs.joomla.report.php" file.
> > > 
> > > 
> > > it's possible that the file pattern matcher works different under windows and linux
> > 
> > 
> > @HLeithner Here I have Linux, and Drone where it fails has Linux, too, or am I wrong?
> 
> hard to say can you fix the file?

@HLeithner I see no mistake in the file. The  "build/psr12/phpcs.joomla.report.php" file is not in any of the exceptions, so it is checked for all. Can it be that this file is used by the checker itself? In this case we would have to exclude it.

ruleset.xml
HLeithner: the comment describes my thoughts, the exclude should be moved to the file directly or it should be refactored and a reason should be added.

HLeithner:
no the report is not used in the check, but you can use phpcs directly on this file if it is included for unknown reasons...

richard67:
@HLeithner Another thing is that if you prefer to have the exception in the single files and have some comment there, I have no idea what kind of comment you except e.g. for the build/build.php file. Maybe I should close this PR and leave that to someone else. Or you make me some suggestion for the right comment.

richard67:
> no the report is not used in the check, but you can use phpcs directly on this file if it is included for unknown reasons...

@HLeithner Here it works when I let phpcs check that file only.

ruleset.xml
richard67: I've reverted the change for the "tests/Unit/bootstrap.php" file.

richard67:
I give it up. The tests work on my local environment and fail in drone.

HLeithner:
for example the build/build.php needs the exception because it has 3 functions and a constant declaration.

All of them can be removed because they are use less (functions used only once and the PHP_TAB constant doesn't makes it better)

PR #38171, opened 2022-06-28
[4.2] Cleanup some inline phpcs:ignore

laoneo:
Removes some inline CS ignores which can be removed since the transition to PSR 12 is done.


laoneo:
@nikosdion you might want to have a look here as if I'm not mistaken, then I removed some of your recent cs:ignore statements.

nikosdion:
The statements were just there to get the PR checks to pass as I was violating Joomla's code style which did not take several things into account, e.g. the fact that some legacy properties which we still have to use were snake_case instead of camelCase. Since your checks pass here, including the PHPCS ones, there's no problem with me :)

HLeithner:
Thanks

Issue #38173, opened 2022-06-28
[4.2] No longer possible to install extensions - Release blocker


joeforjoomla:
### Steps to reproduce the issue
Install whatever extension on Joomla 4.2, for example Acymailing, JEvents, etc

In more details, it seems that the class Joomla\CMS\Installer\Installer is now DatabaseAwareTrait and if an extension instantiate it in the install script as usual, it will be missing the setDatabase causing the B/C issue.
It's common that an install script also install other extensions such as plugins, modules, etc

### Expected result
No errors

### Actual result
Installation fails:
![image](https://user-images.githubusercontent.com/17835460/176318907-ffa7a64b-27ad-4473-a0d2-c953c9e8d2b7.png)

### System information (as much as possible)
Joomla 4.2 Beta 2

brianteeman:
~Almost certainly those extensions have only been set as compatible with 4.0 and 4.1~

joeforjoomla:
@brianteeman an extension has to be set compatible with Joomla 3 and Joomla 4 on the JED and it's supposed to be compatible with a whole major release without this kind of breaking changes between minor releases.

brianteeman:
~1. 4.2 is not releasd yet~
~2. This is nothing new. An extension has always been responsible with setting which version it is compatible with.~
3. I might hav been confused with th targetversion in the update xml

laoneo:
Is it a public extension you try to install? If yes, can you share here the link so I can try it out?

laoneo:
I just tried to install JEvents on the latest 4.2 code and it worked. Can you test if they work with the latest changes from the 4.2-dev branch?

laoneo:
Acymailing failed with the first installation but a second one succeeded.

joeforjoomla:
@laoneo the problem happens when an extension instantiate the class Joomla\CMS\Installer\Installer in the install script file without being aware that now it requires ->setDatabase() to inject the db object

joeforjoomla:
You can try the free package of acymailing: https://www.acyba.com/index.php?option=com_updateme&ctrl=download&component=acymailing&level=starter&ext=tar.gz

This one can't be installed at all, nor the first nor the second time

OctavianC:
Can confirm - `new JInstaller` now needs `$installer->setDatabase($db);` to work. ~~Also, a world of pain comes up when calling the JInstaller to uninstall another extension from the uninstall script:~~
```php
class com_testInstallerScript
{
[...]

PR #38177, opened 2022-06-29
Workaround for removed _db, fix backward compatibility without __get

Fedik:
### Summary of Changes

This is redo #37636
Create a `_db` without use of magic `__get`.

The magic `get` changes behavior of the object A LOT, example if any model have created a property "on fly" it crashes.
Like:
```php
class FooBar{
  public function __get($name){ return null; }
}

$c = new FooBar;
$c->_cache[1] = 'foobar';

var_dump($c->_cache[1] ); // Will be NULL instead of "foobar"
```

That what happen in https://github.com/joomla/joomla-cms/blob/dd91072a956a19a43798515239892edd2b49ac7b/administrator/components/com_modules/src/Model/ModuleModel.php#L676-L679

### Testing Instructions

Because it is an alternative fix, it still should work as before.
Addittionaly, please test any 3d component, wich use `_db` in its model.

As part of #37095

### DIFF
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -401,6 +401,7 @@ abstract class BaseDatabaseModel extends BaseModel implements
 
     /**
      * Set the database.
+     * @todo: This is a workaround for backward compatibility, when $this->_db still in use. Should be removed in 5.0
      *
      * @param   DatabaseInterface  $db  The database.
      *
--- a/libraries/src/MVC/Model/BaseDatabaseModel.php
+++ b/libraries/src/MVC/Model/BaseDatabaseModel.php
@@ -48,7 +48,9 @@ abstract class BaseDatabaseModel extends BaseModel implements
     CurrentUserInterface,
     CacheControllerFactoryAwareInterface
 {
-    use DatabaseAwareTrait;
+    use DatabaseAwareTrait {
+        setDatabase as private setDatabaseTrait;
+    }
     use MVCFactoryAwareTrait;
     use DispatcherAwareTrait;
     use CurrentUserTrait;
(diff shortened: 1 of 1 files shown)


Fedik:
hm, how to ignore this in PHPCS?
```
Property name "$_db" should not be prefixed with an
underscore to indicate visibility
```

richard67:
> hm, how to ignore this in PHPCS?
> 
> ```
> Property name "$_db" should not be prefixed with an
> underscore to indicate visibility
> ```

@Fedik You can add your file to the list of exceptions here https://github.com/joomla/joomla-cms/blob/4.2-dev/ruleset.xml#L37 or you can put a comment at the top of your file at the same place as here https://github.com/joomla/joomla-cms/blob/4.2-dev/tests/Unit/bootstrap.php#L13 , but of course with another rule:

```
// phpcs:disable PSR2.Classes.PropertyDeclaration.Underscore
```

I would prefer to do it in the XML, but a recent discussion in my closed PR #38168 has shown that at least @HLeithner would prefer to do it in the PHP file but with an explaining comment why it's needed. @laoneo was also involved in the discussion. Maybe they can tell you in which of the 2 ways it shall be done.

richard67:
@Fedik Will this solve issue #38173 ? Or does it solve only a part?

Fedik:
Nope, there need separated fix.

libraries/src/MVC/Model/BaseDatabaseModel.php
```diff
@@ -7,6 +7,9 @@
  * @license    GNU General Public License version 2 or later; see LICENSE.txt
  */
 
+// Disable Underscore, because we need $_db for backward compatibility
```
HLeithner, line 10: I really like this comment thanks

libraries/src/MVC/Model/BaseDatabaseModel.php
```diff
@@ -381,27 +400,18 @@ public function setDbo(DatabaseInterface $db = null)
     }
 
     /**
-     * Proxy for _db variable.
+     * Set the database.
```
laoneo, line 403: I would add her a remark that this is a workaround.

laoneo:
There are a couple of core models which are still using _db like the back end plugins model.

libraries/src/MVC/Model/BaseDatabaseModel.php
Fedik: Added

PR #38178, opened 2022-06-29
Remove subform in subform limitation for custom fields

HLeithner:
### Summary of Changes
Remove the limitation to select a subform as a subform field in custom fields.

There is no reason we have this limit, subforms can already be used as field of another subform.

### Testing Instructions
* Create a number of custom fields
* Create a subform !1
* Add the fields to the subform
* Create a subform !2
* Add subform !1 to subform !2
* check if everything works

It's not possible to select subform !1 in subform !2

You can select subform !1 in subform !2 and you also can edit it

### DIFF
--- a/administrator/components/com_fields/src/Field/SubfieldsField.php
+++ b/administrator/components/com_fields/src/Field/SubfieldsField.php
@@ -71,11 +71,6 @@ class SubfieldsField extends ListField
 
         // Iterate over the custom fields for this context
         foreach (static::$customFieldsCache[$this->context] as $customField) {
-            // Skip our own subform type. We won't have subform in subform.
-            if ($customField->type == 'subform') {
-                continue;
-            }
-
             $options[] = HTMLHelper::_(
                 'select.option',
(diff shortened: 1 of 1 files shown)


wilsonge:
I think we need to consider a nesting limit to stop the UI from falling apart.

brianteeman:
a limit of zero - aka the current situation seems perfect to me

HLeithner:
> I think we need to consider a nesting limit to stop the UI from falling apart.

didn't looked at it yet but someone already came up with the question and I hope this protection is already in the field^^

wilsonge:
I *think* that protection is the code your deleting ;)

crystalenka:
I would be very happy to have this in core, personally.

HLeithner:

HLeithner:

Quy:
Alternate #43204 that appears to be more complete.

Quy:
Closing in favor of #43204. Thanks.

PR #38185, opened 2022-06-30
[4.2] Revert the injected app changes in webauthn plugin

laoneo:
### Summary of Changes
As pointed out in #38060 and #38153 changing a view template file can cause an update issue. Despite it is not part of our BC policy. So this pr reverts the changes in it, to not cause any issue when web sites do an upgrade to 4.2 and have an override for plugins/multifactorauth/webauthn/tmpl/default.php.

### Testing Instructions
Run the testing instructions from #38132.

All works.

All works.


Fedik:
Now I really confused :wink: 

> So this pr reverts the changes in it, to not cause any issue when web sites do an upgrade to 4.2 and have an override for plugins/multifactorauth/webauthn/tmpl/default.php.

`multifactorauth` is a new feature in 4.2, there cannot be such fail,
What we talked in adjacent issues, is that it can affect other plugins which use similar approach :wink: 

And we got conclusion, kind of: that it better to define `$app` property, if it not defined and plugin has `getApplication()`. 
Kind of this #38177, but with override for `setApplication()`.
As workaround untill j5, then we drop it.

I can try prepare PR, sometime.

richard67:
> `multifactorauth` is a new feature in 4.2, there cannot be such fail

@Fedik Not a completely new feature. It replaces the 2 factor auth and so has some kind of a migration path from that. But you are right, there should not be any override (except if people update from previous 4.2 alpha or beta versions and had such).

laoneo:
Ok, then I close it. Not sure what I found at that time in the 4.1 branch.

PR #38203, opened 2022-07-02
[4.2] Use the app language translate function in plugins

laoneo:
### Summary of Changes
Reverts the translate function in the `CMSPlugin` class and uses the translate function from the application language object.

### Testing Instructions
Create a GET request task and run it manually in the back end from the web interface.

An output is displayed with the string "Response Code: 200".

An output is displayed with the string "Response Code: 200".

### DIFF
--- a/libraries/src/Plugin/CMSPlugin.php
+++ b/libraries/src/Plugin/CMSPlugin.php
@@ -173,34 +173,6 @@ abstract class CMSPlugin implements DispatcherAwareInterface, PluginInterface
     }
 
     /**
-     * Translates the given key with the local applications language. If arguments are available, then
-     * injects them into the translated string.
-     *
-     * @param   string   $key        The key to translate
-     * @param   mixed[]  $arguments  The arguments
-     *
-     * @return  string  The translated string
-     *
-     * @since   4.2.0
-     *
-     * @see     sprintf
-     */
-    protected function translate(string $key): string
-    {
-        $language = $this->getApplication()->getLanguage();
-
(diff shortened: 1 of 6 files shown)


brianteeman:
Why not revert fully to `$this->logTask(Text::sprintf(` ?

or maybe a better commit message?

laoneo:
Why using static function when we have everything from the local application object?

Another reason is, because `Text` uses the application from the Factory class and then we can't do correct unit testing. Because code should be autonomic. As soon as the testable class loads it's dependency by it's own, then it violates the inversion of control principle which leads to a dependency chain hell. And no, this is not super theoretic high school stuff, it is a basic principle of software development which we started with the development of Joomla 4. 

In this case it would be simple to just set the application on the Factory class but there are places where we have much more complex dependency chains which we want to avoid (this is the reason why so many "unit test" got removed in Joomla 4). When we start to go the way that classes do define their dependencies and not load them by themself, then the code will become simpler, more predictable and easily unit testable. When we start doing excuses, here and there, then it will become messy again over time.

brianteeman:
Thank you for the explanation. Please consider a better commit message then please

laoneo:
As the pr will be squashed it uses the pr title. Any suggestion?

brianteeman:
It was the title I referred to as that is the commit message.

It is not a revert - that is confusing - perhaps "update"?

brianteeman:
Thank you

roland-d:
Thank you

PR #38209, opened 2022-07-02
Update Joomla Version and dependencies for 5.0

HLeithner:
This Pull Request update the Joomla Version and the minium PHP requirements.
Additionally all composer packages and npm package has been updated based on the semver rule we set.
We have to check all packages to be updated to the latest current version.

For example symfony 6.1 is our target version (at this time) we have to upgrade and validate compatibility.
Some applies for all other 3rd party packages.


HLeithner:
@nikosdion since we support only 8.1 and the current webauthn lib (2.x) in't compatible can you have a look someone in the future to update this libraries then we can have a proper composer.lock again.

thanks

nikosdion:
@HLeithner There is no other WebAuthn library for PHP. That's the only one which actually works, the work to produce one is way too much, therefore nobody else really tries to.

Every few months, when a maintainer tries to update dependencies, this discussion keeps coming back. 

I had said ten years ago that Composer is NOT a good idea for mass–distributed software which needs to run in a far wider range of supported environments, especially PHP versions, than third party library developers are willing to support. I was told I was against progress or that I don't understand how PHP software development works — despite the fact that I was only saying that because it had been just a few months since I had tried using the official Amazon AWS SDK for PHP through Composer and ran into these problems myself.

You keep bumping onto that unsolvable problem

You have exactly three options:

1. Invent everything yourselves. This is impossible, especially when we're talking about WebAuthn which has very specific domain knowledge requirements we can't find easily in the PHP community, let alone the Joomla community.
2. Fork dependencies on every major release and maintain them for future PHP versions. This is impractical because you have neither the manpower nor the domain–specific knowledge for most of these libraries to be anywhere near confident that you're not introducing new security issues. Nevermind the fact that not being able to update a library for 2 to 4 years will expose Joomla users to security issues third party dependencies may have.
3. Just accept the fact that every minor version of Joomla needs to drop PHP versions which have become end–of–life.

You can't have your pie and eat it too. I've tried. It doesn't work. I ended up reinventing the wheel to avoid third party dependencies in my software but that only works for me because I use a tiny subset of features. If it's something more substantial and I absolutely need a third party dependency, like WebAuth, I go with the third option. If you have a fourth option I'd like to hear it.

PS: Using Rector to “upgrade” the code to be compatible with a future PHP version falls under case number 2. Been there, done that, it gets exponentially harder. I am only doing this for ONE dependency (Horde IMAP) and only because nobody and nothing else supports XOAUTH2 for fetching and sending email with G Suite accounts. It sucks exactly as much as you think and even more. As I said, I've been there and done that — all of that — already. There is no magic solution I know of but if anyone has one I'd love to hear it! It'd save me a lot of pain.

nikosdion:
BTW, for automated tests **and** for building Joomla releases you already have the minimum version in composer.json as the platform environment. This means that the old WebAuthn library version we use (version 2, when they are now on 4...) will install. This is the recommended way to do this when you have a minimum supported PHP version which does not match your local build environment. If you are referring to that, it does not need changing and it's not a workaround, it's legitimately how it should be done and the reason Composer offers that config option.

Despite what its `composer.json` declares, the WebAuthn library v2 _does_ work on PHP 8.1 — or at least does not break in any show–stopper way (IIRC there are still some deprecated notices, whatever, Joomla core does the same). My local development sites are on PHP 8.1 and 8.2-dev. I am using WebAuthn for MFA using TouchID on macOS, FaceID on iOS/iPadOS, and a WebAuthn dongle as a backup when my laptop is docked. I log into these sites at least two dozen times a day (I have short session times). I can tell you it definitely works. Dunno what will happen with 8.2 proper. Until there's an 8.2-RC1 I won't call it working. Fair?

HLeithner:
> @HLeithner There is no other WebAuthn library for PHP. That's the only one which actually works, the work to produce one is way too much, therefore nobody else really tries to.

I thought we can upgrade to 4.1.x https://github.com/web-auth/webauthn-framework and was hoping that you can do the "migration" not sure if it is one.

I'm not interested to maintaining our own version of web-authn-framework if not needed.

Sorry if I expressed my self wrong.

nikosdion:
@HLeithner Aaaah! Now I get you. I thought you were worried about the maximum supported PHP version of the current library version :D 

Sure, we can definitely upgrade to version 4 of the library. If I don't already have the if-blocks in the current code I definitely have them in my repositories (LoginGuard where the MFA feature was forked from and PasswordlessLogin where the WebAuthn plugin was forked from). It's just a matter of me having some spare time...

...which means not in the next month :( I have a backlog of deep refactoring to do on my software which was pushed behind due to the Joomla 4.0 release and the need to migrate everything and make sure it's all polished before doing any deep architectural work. I also need to work on the developer docs and the Rector rules to auto-refactor Joomla 3 components. I _think_ that come October or November at the latest I can work on the WebAuthn migration for both login and MFA with an inclusion target of the first betas of Joomla 4.3 at the latest. Does that sound in line with what you had in mind?

nikosdion:
I just realised you are talking about 5.0. So I guess my plan is in line with what you have in mind, never mind me :D

HLeithner:
Yes I'm talking about Joomla 5 so you have enough time ;-) upgrading composer is a pain but I think that can be solved differently in the mean time for the ci.

Thanks and just take your time.

nikosdion:
@HLeithner So, I was taking a look at this a couple of days ago. Are you okay with me doing a MEGA-PR which updates the dependencies, MFA and WebAuthn in one go? It's the only way I can reasonably think will result in something testable. Right now all libraries are out of date and `composer install` won't even work.

nikosdion:
@HLeithner Well, we have a problem.

`web-auth/webauthn-lib` versions 2 and 3 are not _nominally_ compatible with PHP 8.1 — even though they do work fine, PHP 8.1 is not listed in their composer.json files.

Version 4 of the library requires `psr/log` version 2 or 3.

However, the Joomla Framework's `joomla/application` package requires `psr/log` 1.

Therefore we cannot upgrade the dependencies unless EITHER `joomla/application` upgrades its dependency (b/c break, runs afoul of the b/c promise you gave in August) OR we fork the WebAuthn library and all its dependencies to make them compatible with the rest of our dependencies.

I know what I would do, but it's ultimately not my call, it's yours. Tell me how to proceed. As things are right now I cannot update Joomla 5's dependencies to even make it installable…

HLeithner:
I created https://github.com/joomla/joomla-cms/pull/39123 which solves the dependency issues and upgraded webauthn to latest version.

Issue #38214, opened 2022-07-03
CLI finder:index


brianteeman:
### Steps to reproduce the issue
`php .\cli\joomla.php finder:index `

### Expected result
no errors

### Actual result
```
In PluginHelper.php line 46:
                                                                                     
  Call to undefined method Joomla\CMS\Application\ConsoleApplication::getTemplate()
```

### System information (as much as possible)
php 8.1

alikon:
please test #38222

PR #36747, opened 2022-01-20
[4.2] Smart Search: Index custom fields

Hackwar:
Pull Request for Issue #12829.

### Summary of Changes
This PR implements indexing of the custom fields for a component. This is controlled on a per field basis, where you can select if the content should be just added to the search index or handled as a taxonomy or both or actually none of the above. You can then implement this feature for your respective finder plugin with a single line of code.
Each custom field has a new option in its edit screen where you can select how the indexer should treat this field. The finder plugin then has to call `Helper::addCustomFields($item, 'context');` and the rest is just magic. ;)
Right now this is implemented for com_content and com_contacts.

### Testing Instructions
1. Your installation should have some custom fields for com_content or com_contact in it.
2. Edit a custom field and set the "Smart Search" option to "Make searchable and add as taxonomy".
3. Run the indexing process either by running the mass indexer in the smart search component or by editing an item and saving it again.
4. See that the custom field shows up as a taxonomy and can be find by searching for its value.

### Documentation Changes Required
The documentation needs to be amended to contain how to implement this feature for third party finder plugins.

### Notes
1. Since 4.1 is in RC, this will have to go into 4.2. Since there is no 4.2-dev branch yet, I'm pointing this towards 4.1-dev for now.

PR #38222, opened 2022-07-04
[4][cli] finder:index

alikon:
Pull Request for Issue #38214 .

### Summary of Changes
get the Template in the Console Application (even if not needed) but like API application 

### Testing Instructions
from cli run
`php cli/joomla.php finder:index`

`In PluginHelper.php line 46:
  Call to undefined method Joomla\CMS\Application\ConsoleApplication::getTemplate()`

finder indexer works

### DIFF
--- a/libraries/src/Application/ConsoleApplication.php
+++ b/libraries/src/Application/ConsoleApplication.php
@@ -463,4 +463,29 @@ class ConsoleApplication extends Application implements DispatcherAwareInterface
 
         return Router::getInstance($name, $options);
     }
+
+    /**
+     * Gets the name of the current template.
+     *
+     * @param   boolean  $params  True to return the template parameters
+     *
+     * @return  string|\stdClass
+     *
+     * @since   __DEPLOY_VERSION__
+     */
+    public function getTemplate($params = false)
+    {
+        // The CLI application should not need to use a template
+        if ($params) {
+            $template              = new \stdClass();
+            $template->template    = 'system';
+            $template->params      = new Registry();
+            $template->inheritable = 0;
+            $template->parent      = '';
+
+            return $template;
+        }
+
+        return 'system';
+    }
 }


brianteeman:
I have tested this item :white_check_mark: successfully on 19e0d039fe8d64dcb6cf732dc625c1ce2a991af5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38222">issues.joomla.org/tracker/joomla-cms/38222</a>.</sub>

viocassel:
I have tested this item :white_check_mark: successfully on 19e0d039fe8d64dcb6cf732dc625c1ce2a991af5<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38222">issues.joomla.org/tracker/joomla-cms/38222</a>.</sub>

alikon:
@SharkyKZ  the same method is in the API application as well https://github.com/joomla/joomla-cms/blob/365d316e291dbf30b95e893013a27fdbca1d1ae5/libraries/src/Application/ApiApplication.php#L194

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38222">issues.joomla.org/tracker/joomla-cms/38222</a>.</sub>

wilsonge:
This one's a bit different to the api because we can check whether the application implements CMSApplicationInterface instead of CMSWebApplicationInterface (although in truth obviously would be better if the api didn't have that dummy getTemplate method).

richard67:
> This one's a bit different to the api because we can check whether the application implements CMSApplicationInterface instead of CMSWebApplicationInterface (although in truth obviously would be better if the api didn't have that dummy getTemplate method).

@wilsonge So should I revert RTC and the PR needs to be changed or replaced by a new one?

wilsonge:
I'm not sure. it depends how intrenched the usage is. Like I don't see `PluginHelper::getLayoutPath` being called by the finder plugins themselves. So what is actually calling this function and what's the effect of it looking for the system template?

Fedik:
This is a general issue in our override system, which is require Template (it not only about Finder).
Plugin may want to render some `tmpl` or `layout` in CLI/API Application context.

Defaulting to 'system' template will work, but it just hide the issue.
The fix is okay, as long as we do not have a better idea what to do here :)

alikon:
@wilsonge  this is the call stack 
image

wilsonge:
OK So comes from this https://github.com/joomla/joomla-cms/pull/36747 PR which is a new 4.2 feature. So first of all that's caused a regression. Is this CLI working in 4.1?

Second of all that PR is setting prepareValue to true. But this is presumably going to index different based on whether it's frontend or backend driven edits causing the index because the active template will be different. Seems like that entire feature is very broken to me? Beyond just the CLI. 

Tagging @Hackwar for insight as he wrote the code and might be able to give an understanding and @roland-d because looks like 4.2 currently is causing a regression.

Hackwar:
I haven't written the CLI part here. I would have to look into this more deeply.

Fedik:
It not a Finder issue, it "Template Override" issue.
The same will happen if call `LayoutHelper::render('foo.bar')` in CLI context.
If we fix it, then a rest will work automatically :)

wilsonge:
> I haven't written the CLI part here. I would have to look into this more deeply.

This isn't a purely CLI issue. Your rendering the custom field on index - which uses the active template. Not the frontend template. In the CLI that crashes but even if that update comes from the backend seems dodgy to me

roland-d:
@wilsonge Thanks for the heads-up. 

> So first of all that's caused a regression. Is this CLI working in 4.1?

It is not a regression I guess because it fails with the same error on Joomla 4.1.5 for me.

```
$ php cli/joomla.php finder:index
Finder Indexer
==========================

 Starting Indexer
 Setting up Smart Search plugins
PHP Notice:  Undefined index: HTTP_HOST in .../j4/libraries/src/Uri/Uri.php on line 103
 Setup 127 items in 0.177 seconds.

In image.php line 57:
                                                                                     
  Call to undefined method Joomla\CMS\Application\ConsoleApplication::getTemplate()  
```
```
$ php cli/joomla.php 
Joomla! 4.1.5 (debug: Yes)
```

Regression or not, this is something we should fix, ideally before the last beta 3 release tomorrow.

@Hackwar Do you have time to look into this?

wilsonge:
OK So sat down with the dev laptop tonight and actually tried doing some reproduction steps and debugging. First of all - with latest 4.2-dev and the test data set installed I can index from CLI just fine.

If I then create a custom field and make it searchable it fails. So i'm very confident in 4.2-dev it's related to that fields PR (at least for now - if it fails in 4.1-dev then possible there might be other scenarios we have to look at too). I can't install the same dataset there (because I run 8.1 locally and 4.1.5 doesn't have the PHP8.1 fixes for the test sample set)

4.2-dev:

Screenshot 2022-07-04 at 23 01 30

4.1.5:

Screenshot 2022-07-04 at 23 42 34


I think that PR needs reverting if it's not fixed because:

1. Your going to get different results depending on if you are frontend or backend editing (because it may or may not load the correct template overrides for a custom field).
2. I don't think it should work (searchable) for things like the media field. It certainly doesn't at the moment anyhow and I'm not sure we should show people the option
3. If you make a subform field searchable (not such a totally crazy thing to do) then it hard crashes the index process.

Screenshot 2022-07-04 at 23 39 52

alikon:
@roland-d  with 4.1.5 stable it works well for me under the same settings for 4.2 (blog sample data)
image

roland-d:
Thank you @wilsonge for your detailed response. I agree, at this point we should revert this PR and take some time to further polish it.

@alikon Perhaps it comes from 3PD custom fields, I did not check why it fails, just that it fails.

This PR can be closed I guess and the final solution should be in the new PR for indexing custom fields.

brianteeman:
so you will be reverting #36747 and closing this and waiting for a new pr

roland-d:
@brianteeman That is correct, a new PR needs to be made and give @Hackwar some time to look into this.

roland-d:
Thanks everybody for your work on this.

PR #38225, opened 2022-07-04
[4.2] NPM build tools cleanup

dgrammatiko:
Pull Request for Issue # .

### Summary of Changes
- remove some I/O operations on both the JS and SCSS minification (eg use the code output directly)
- Ran `npm lint:js -- --fix` and committed the changes
  - Removed the IIFE from [build/media_source/com_content/js/articles-status.es6.js](https://github.com/joomla/joomla-cms/compare/4.2-dev...dgrammatiko:joomla-cms:4.2-dev-buildtools?expand=1#diff-4afb4f910136391957e62b8afcb5f0d6b240ee226cd69b4960777bc15ddfe668) as the file is correctly loaded either as module or when in `nomodule` it's using the defer attribute, thus adding the listener for `DOMContentLoaded` is pure overhead. At that point the DOM is already there fully parsed. This needs to be applied **ON EVERY SCRIPT**

### Testing Instructions
On a fresh installation of Joomla (ie `git clone`) do the composer install and then the npm install. Copy the media folder somewhere in your hardisk. (check the time needed)
Now clone this PR and do both the composer and npm installations. Once done do a folder compare of the current media folder against the folder that you saved in the previous step. You should have the same files apart from the `build/media_source/com_content/js/articles-status.es6.js` which will have different content (expected).

Compare the times and do a brief check that both front end and back still work as expected

Screenshot 2022-07-04 at 18 36 53

Screenshot 2022-07-04 at 18 35 50

NO

@HLeithner have somebody merge this so I could do the changes in the v5.0. Thanks


brianteeman:
the change to image.vue is not correct. The span element is not self closing. Both parts are mandatory.

dgrammatiko:
> the change to image.vue is not correct. The span element is not self closing. Both parts are mandatory.

On your repo (4.2) try `npm run lint:js -- --fix`, you will get the same. The HTML inside the .vue files is totally fine with this change and actually it's the recommendation, (it's the v- part that makes the difference)

btw here's the eslint rule and the explanation: https://eslint.vuejs.org/rules/html-self-closing.html

build/build-modules-js/stylesheets/handle-scss.es6.js
```diff
@@ -20,28 +20,23 @@ module.exports.handleScssFile = async (file) => {
...
+  // Auto prefixing
+  const cleaner = Postcss([Autoprefixer()]);
+  cleaner.process(compiled.css.toString(), { from: undefined })
+    .then((res) => {
+      writeFile(cssFile, res.css, { encoding: 'utf8', mode: 0o644 });
+      return Postcss([CssNano]).process(res.css, { from: undefined });
+    })
+    .then((cssMin) => writeFile(cssFile.replace('.css', '.min.css'), cssMin.css, { encoding: 'utf8', mode: 0o644 }))
+    // eslint-disable-next-line no-console
+    .then(() => console.log(`✅ SCSS File compiled: ${cssFile}`))
```
brianteeman: is there a reason why this one is 
icon text file
and all the others are
text file icon

dgrammatiko: I have no clue but I can switch it

brianteeman:
### Before
image

### After
image

PR #38227, opened 2022-07-05
[4.2] Empty error message in package install

brianteeman:
Pull Request for Issue #37575 .

### Summary of Changes
Display an error message with text not just an empty error

### Testing Instructions
npm run build:js 
or install prebuilt package

Go to the package installer and try to drag & drop a folder

image

and js error in the console

image

and no console errors

none


build/media_source/plg_installer_packageinstaller/js/packageinstaller.es6.js
```diff
@@ -120,6 +120,11 @@ Joomla = window.Joomla || {};
       const file = files[0];
       const data = new FormData();
 
+      if (!file.length) {
```
Fedik: The `file` object does not have property `length`, this `if` will prewent any upload :wink: 
You probably have to check for type: `if (!file.type)`.

Ideally it should check `if (file.type !== 'application/zip' .... and other types)`, 
but it will be fine just `if (!file.type)`.

richard67:
Unfortunately I cannot reproduce the issue here on a clean, current 4.2-dev without this PR. I don't get an empty error message when trying to upload a folder. I get a message "No Install Type Found". I've tried an empty folder and a folder with one empty file and also a folder with files and subfolders. In all cases the same result. What am I doing wrong?

P.S. The message I get is of type information or has no type. It has blue background.

brianteeman:
were you using drag and drop?

richard67:
> were you using drag and drop?

Yes, was on `administrator/index.php?option=com_installer&view=install` and dragged the folder from my Windows 10 Desktop into the drop area of that page.

richard67:
After I've dropped the folder, it shows "Uploading 0%" or something like that for a while, and then:


Client (browser) is Windows, and the webserver is Linux.

richard67:
What a pity because I think this PR is right now, and I wanted to test it with success, only I can't reproduce the issue.

Fedik:
Maybe some OS specific, I have "empty error" as in description, when doing "folder drag and drop"  in Linux in Chrome

brianteeman:
recorded on windows 11

### Before
before

richard67:
Hmm, no, same result when I use Firefox on my Linux server directly.

richard67:
@pearlpriyanka Did you have your issue #37575 with Windows 11 as client OS where the browser was used?

richard67:
Hmm, anyway, I believe you guys of course, no need for a video or animated gif. I just can't reproduce it here and would like to know why.

Fedik:
I have tested this item :white_check_mark: successfully on e57fc28a1bb924cfedf1c4626e7ba58608516267<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38227">issues.joomla.org/tracker/joomla-cms/38227</a>.</sub>

joomdonation:
I have tested this item :white_check_mark: successfully on e57fc28a1bb924cfedf1c4626e7ba58608516267<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38227">issues.joomla.org/tracker/joomla-cms/38227</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38227">issues.joomla.org/tracker/joomla-cms/38227</a>.</sub>

build/media_source/plg_installer_packageinstaller/js/packageinstaller.es6.js
```diff
@@ -120,6 +120,11 @@ Joomla = window.Joomla || {};
       const file = files[0];
       const data = new FormData();
 
+      if (!file.type) {
```
dgrammatiko, line 123: @brianteeman you could be way more restrictive here. The only accepted files for packages are `.zip and .tar` files so something like 

```suggestion
      if (!['application/zip', 'application/x-tar'].includes(file.type)) {
```

acts also as a client side validation for the uploads

brianteeman: not that simple. tar and gz aree not official mimetype

roland-d: @brianteeman Where is the list of official mimetypes? The [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) does list these mimetypes.

dgrammatiko: You can refer to the mime types that apache is using: https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

brianteeman: image

Fedik: There much variants https://superuser.com/a/960710/190987, 
We can stay with just checking that the type actually exist (I hope server doing proper validation already :smile: ).
If we want to validate it on client side, then have to include ALL variants, for each file type supported in upload here.

roland-d:
Thank you. As for the discussion of which mime-types to check, that can be a separate issue.

PR #38231, opened 2022-07-05
[4.2] Ruleset.xml

brianteeman:
Combine two rules as they are the same

cc @hleithner


HLeithner:
thanks

brianteeman:
np spotted it while configuring my new system

PR #38240, opened 2022-07-07
[4] remove api-basic-auth-plg

alikon:
### Summary of Changes
as per https://docs.joomla.org/Joomla_Api_Specification#Authentication
_There is also a legacy basic authentication plugin for simple local development however this is likely to be removed before the release of Joomla 4.0.0 stable._

### Testing Instructions
code review

there is still a basic authentication plugin

no more basic authentication plugin

?


richard67:
As it is now, the PR doesn't handle removal of the plugin on core updates.

richard67:
Besides that I think we can't do that with a minor version as it's a b/c break.

alikon:
it should have been removed before 4.0 stable,  how it can be a b\c break ?

brianteeman:
because it wasnt removed and people could be using it

laoneo:
In j3 we still have the GMail plugin which is totally broken. So I think we should not remove extension in a minor version as we didn't do it in the past. It would be too disrupting.

@richard67 I don't think so that removing extensions is covered by our BC policy.

alikon:
if really there is someone that even use it in the wild, which i strongly doubt, even then, they must be use the token one, just for security reason 😄 

> "_**errare humanum est perseverare autem diabolicum**_"

richard67:
> @richard67 I don't think so that removing extensions is covered by our BC policy.

@laoneo I know that removing core extensions is not covered by our b/c promise. But having them as option in an API maybe is.

PR #38245, opened 2022-07-08
Display array properties of JConfig in System Information

Denitz:
Pull Request for Issue # .

### Summary of Changes

JConfig can contain array properties, they are not properly displayed in backend System Information

### Testing Instructions

Add nested array property in `configuration.php`:

```
public $test = [1, 2, 3 => [4, 5]];
```

Open "System Information > Configuration File" tab

See `Warning: Array to string conversion` warnings and no array values are displayed.

See json of arrays

No.


administrator/components/com_admin/src/Model/SysinfoModel.php
```diff
@@ -287,6 +287,12 @@ public function &getConfig(): array
             $this->config[$key] = 'xxxxxx';
         }
 
+        foreach ($this->config as $key => $value) {
+			if (\is_array($value)) {
+				$this->config[$key] = \json_encode($value);
+			}
+		}
```
richard67:
```suggestion
            if (\is_array($value)) {
                $this->config[$key] = \json_encode($value);
            }
        }
```
Fix code style: On the 4.2-dev branch we have recently changed to use the PSR-12 standard. This uses spaces and not tabs for indentation in PHP files. Each indentation level is 4 tabs.

chmst:
@Denitz could you give an example to improve your test istructions?

Denitz:
Sorry, I was "inspired" by Joomla3.

Joomla4 has `<?php echo HTMLHelper::_('configuration.value', $value); ?>` and it's enough for usual array, but multidimensional  arrays will cause the issue.

ReLater:
I have tested this item :white_check_mark: successfully on aa14dab8f6148417cc62b0c5b3ea112fd0441ed6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38245">issues.joomla.org/tracker/joomla-cms/38245</a>.</sub>

viocassel:
I have tested this item :white_check_mark: successfully on aa14dab8f6148417cc62b0c5b3ea112fd0441ed6<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38245">issues.joomla.org/tracker/joomla-cms/38245</a>.</sub>

richard67:
The file still has old, meanwhile wrong code style (tabs used for indentation) at the places which were not changed by this PR, while the changes of this PR use the right PSR-12 style.

I have no idea why the PHPCS check is successful in Drone with these tabs.

@Denitz Would be cool if you could fix the code style at the other, unchanged places of the file. It would not require new human tests after such a change. Otherwise, if you don't want, we can make a PR for that later.

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38245">issues.joomla.org/tracker/joomla-cms/38245</a>.</sub>

brianteeman:
> I have no idea why the PHPCS check is successful in Drone with these tabs.

I asked the same question before and @hleithner said it wasnt currently something being done

HLeithner:
Sorry ?? '' is not the way to go, please make a validation for string and throw a deprecation warning and add a todo that we can throw an exception or better set a type in the function signature.

HLeithner:
another question is why should we have array information in the configuration.php?

richard67:
Back to pending due to comment https://github.com/joomla/joomla-cms/pull/38245#issuecomment-1190331057 .<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38245">issues.joomla.org/tracker/joomla-cms/38245</a>.</sub>

ReLater:
> another question is why should we have array information in the configuration.php?

Because one can override `configuration.php` (other file location). Not too obvious but without forbidden hacks possible.

HLeithner:
tbh this looks more like a band ad then a proper solution. If this is really needed (and I mean there are better solutions out there) then we should look at them. For example everything beside a scalar value and an array is a problem. So this should be intercepted properly.

Anyway for allowing a null value it should not be hidden behind a `?? ''` a proper if statement with maybe a `NULL` string as return value makes more sense.

Denitz:
Sorry, which band ad? 3rd-party extensions can keep own data in global config in order to not perform own data loading operations.

Added NULL string.

HLeithner:
> Sorry, which band ad? 3rd-party extensions can keep own data in global config in order to not perform own data loading operations.

json_encode is not a nice visualization of a value for example.

laoneo:
> Sorry, which band ad? 3rd-party extensions can keep own data in global config in order to not perform own data loading operations.

How? I really see no use case where an extension should add it's own data to the configuration.php.

laoneo:
> Sorry, which band ad? 3rd-party extensions can keep own data in global config in order to not perform own data loading operations.

How? I really see no use case where an extension should add it's own data to the configuration.php.

Denitz:
I have a custom plugin which specially adds special fields to global config with multiple values and I do see the use case of having custom app-related properties.

laoneo:
Why do you not store the settings in the plugin settings? Can you share a bit more details why you add your own settings to the global Joomla configuration?

laoneo:
You can write your own system plugin and then overwrite the service in the registry. Like that you can then visualize your date in the way you want.

Hackwar:
I would vote this as not a bugfix and thus this would have to go into 4.3.

Denitz:
> Why do you not store the settings in the plugin settings?

I need the settings to be available before the plugins are loaded from the database.

> Can you share a bit more details why you add your own settings to the global Joomla configuration?

I have a complex custom integration service, sorry, can't inform more details. The main issue is that I need to execute own code with custom settings before the plugins are actually loaded, hence I can't afford getting settings from the database.

> You can write your own system plugin and then overwrite the service in the registry. Like that you can then visualize your date in the way you want.

Yes, I can, but all modern PHP frameworks allow nested data structures in config and are able to dump them without errors.
I want Joomla to be modern with just a couple lines of code :)

laoneo:
As I already said, when this is something specifically to your use case, then overwrite the HTML service of the sysconfig view in the plugin or when it is only about one site, then make a template override of that view and display your custom data properly. I'm closing this as there will come too many implications when people start adding arrays to the configuration.php.

PR #38249, opened 2022-07-10
[4.x] SEF, Implement Case insensitive url, but only for path part of URL

Fedik:
Pull Request for Issue #38145 .

### Summary of Changes

Implement Case insensitive url (only for path part of URL), by use of 301 redirect in our SEF plugin.

### Testing Instructions

Apply patch,
Try open any url on the site, but change the case of URL path manually

Error 404, page not found

Page loaded, after redirect to lowercase counterpart (but only for path part of URL)

Probably, not sure

@Hackwar please review

### DIFF
--- a/plugins/system/sef/sef.php
+++ b/plugins/system/sef/sef.php
@@ -13,13 +13,15 @@
 use Joomla\CMS\Plugin\CMSPlugin;
 use Joomla\CMS\Router\Route;
 use Joomla\CMS\Uri\Uri;
+use Joomla\Event\SubscriberInterface;
+use Joomla\String\StringHelper;
 
 /**
  * Joomla! SEF Plugin.
  *
  * @since  1.5
  */
-class PlgSystemSef extends CMSPlugin
+class PlgSystemSef extends CMSPlugin implements SubscriberInterface
 {
     /**
      * @var    \Joomla\CMS\Application\CMSApplication
@@ -27,6 +29,45 @@ class PlgSystemSef extends CMSPlugin
      * @since  3.5
      */
     protected $app;
+
+    /**
+     * Returns an array of events this subscriber will listen to.
+     *
+     * @return  array
+     *
+     * @since  __DEPLOY_VERSION__
+     */
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            'onAfterInitialise' => 'onAfterInitialise',
+            'onAfterDispatch'   => 'onAfterDispatch',
+            'onAfterRender'     => 'onAfterRender',
+        ];
+    }
+
+    /**
+     * Implement Case insensitive url, but only for path part of URL
+     *
+     * @return  void
+     * @since  __DEPLOY_VERSION__
+     */
+    public function onAfterInitialise()
+    {
+        if (!$this->app->isClient('site')) {
+            return;
+        }
+
+        // Get current URI
+        $uri   = Uri::getInstance();
+        $lPath = StringHelper::strtolower($uri->getPath());
+
+        // Always redirect to lowercase counterpart
+        if ($lPath !== $uri->getPath()) {
+            $uri->setPath($lPath);
+            $this->app->redirect($uri, 301);
+        }
+    }
 
     /**
      * Add the canonical uri to the head.
--- a/plugins/system/sef/sef.php
+++ b/plugins/system/sef/sef.php
@@ -60,10 +60,10 @@ class PlgSystemSef extends CMSPlugin implements SubscriberInterface
 
         // Get current URI
         $uri   = Uri::getInstance();
-        $lPath = StringHelper::strtolower($uri->getPath());
+        $lPath = $uri->getPath() ? StringHelper::strtolower($uri->getPath()) : false;
 
         // Always redirect to lowercase counterpart
-        if ($lPath !== $uri->getPath()) {
+        if ($lPath && $lPath !== $uri->getPath()) {
             $uri->setPath($lPath);
             $this->app->redirect($uri, 301);
         }


Webdongle:
I have tested this item :white_check_mark: successfully on e39e518be051d6f94b6397684261d44ac2d4b875<br /><br />Tested successfully<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38249">issues.joomla.org/tracker/joomla-cms/38249</a>.</sub>

plugins/system/sef/sef.php
```diff
@@ -28,6 +30,45 @@ class PlgSystemSef extends CMSPlugin
...
+     * Implement Case insensitive url, but only for path part of URL
+     *
+     * @return  void
+     * @since  __DEPLOY_VERSION__
+     */
+    public function onAfterInitialise()
+    {
+        if (!$this->app->isClient('site')) {
+            return;
+        }
```
sandewt, line 59: Proper use of braces:
```
if (!$this->app->isClient('site')) 
{
       return;
}
```

richard67: @sandewt You are wrong. The code style has been recently changed on the 4.2-dev branch to PSR-12 standard.

sandewt: Thanks @richard67 , I see #37686

weeblr:
Hi

I'm afraid this cannot go as is. This is an oversimplification of the issue and as proposed, **it will kill a large number of existing sites** (in the dozens of thousands of them): **all those which use uppercase letters as part of the canonical URL**.

The question at hand is NOT to redirect or canonicalize to the lower-case version of the page URL. What needs to be done is redirect/canonicalize to the **canonical version** of the URL.

- On Joomla routing-only sites, this will be a URL with the lower-cased version of the path - or an entirely different URL, that you know nothing of.
- On non-Joomla-routing based sites, you have no idea what the canonical URL is and you should not redirect under any circumstances.

### Possible solution:

**- when the Joomla router decides the request is a 404, it does the lower-case test and does the redirect if there are some upper case**. This should also be disableable.

### Less correct solution:

- keep doing it as proposed here, in the SEF plugin, **but it has to be disableable by 3rd-parties** (programmatically), so that a site with an up to date SEF extension can disable the redirect (I wish the auto-canonical inserted by the SEF plugin would also be disableable, it's also so often so wrong).

### In all cases

The current PR is missing the basic condition test to allow redirecting: method must be GET, you cannot redirect on POST, PUT, DELETE, etc

**I reiterate here that using a redirect is likely to be too harsh and cause large SEO issues in real life. "Just" adding a canonical is by far the preferred method in our use-case.** 

Note this does not contradict the information survey done by @simbus82 : **the question asked was biased because it assume the correct URL was the lower-cased one**. This assumption is **incorrect** in the Joomla world.

Cheers

Webdongle:
@weeblr 
Are you saying
If sNakecASe can not be found convert to lowercase?

weeblr:
> If sNakecASe can not be found convert to lowercase?

It certainly can - if sNakecASe is the correct URL for the page. But on some sites, the correct URL is my-article and one some Joomla sites the URLs are MY-ARTICLE or My-Article.


**Lower case is NOT always the correct URL on a Joomla site.**

Webdongle:
@weeblr 
I think you are miss reading my question.  It is a Yes/No question not a statement that requires an agreement like "It certainly can".

I will rephrase the question.

Are you saying
That when sNakecASe can not be found it should be converted to lowercase?

e.g.
IF
sNakecASe not found
THEN
look for
snakecase

It is a Yes/No question not a statement that requires an agreement like "It certainly can"

weeblr:
Hi

It absolutely cannot. It will kill dozens of thousands of Joomla websites.

This cannot be merged as is under any circumstances.

weeblr:
@Webdongle Also I did not mis-read your question. Simply, your question is not a Yes/No question.

Or the question is a Yes / No, but the answer is not.

simbus82:
> **the question asked was biased because it assume the correct URL was the lower-cased one**. This assumption is **incorrect** in the Joomla world.

Maybe it's the idea that you have about Joomla that's wrong.
You may have forgotten that the alias field of menu items, articles, categories, etc. turns everything into lowercase.
So Joomla pushes not to have "non-lowercase" URLs in any way.

And it's a worldwide approved assumption that lowercase for URL it's the best error-proof mode to creating urls. For backlinking too, that is really important for SEO.
Also Google give the advice to use LC urls, right to not have any type of errors.

Why do you have Joomla websites with non-lowercase "path" in pages urls?
I have done more than 400 websites (true business website, not for the pub down the road). I have never had a mixedcase url or a single uppercase url with Mambo, Joomla 1.x, 2,x, 3.x and 4.x.

weeblr:
@simbus82 Really? me? wrong idea about Joomla routing?

> You may have forgotten that the alias field of menu items, articles, categories, etc. turns everything into lowercase.

You may have forgotten about SEF extensions. What you state here is only valid for sites using Joomla routing. 

> I have done more than 400 websites (true business website, not for the pub down the road). I have never had a mixedcase url or a single uppercase url with Mambo, Joomla 1.x, 2,x, 3.x and 4.x.

Don't want to get into my is bigger than yours, but there are several **dozens of thousands** - at the very least, a more reasonable estimate is well above 100,000 - with such URLs. And that's just for sh404SEF. JoomSEF and others have to be added. Even if you talk about Joomla 4, because they are all migrating now - and keeping their mixed-case URLs, because we never want to change URL.

> And it's a worldwide approved assumption that lowercase for URL it's the best error-proof mode to creating urls. For backlinking too, that is really important for SEO.

Which URLs are created has nothing to do with our discussion. The problem is backward compatibility.

If this PR goes as is, any site using mixed-case URL is dead: all their pages will be redirected to the lower-case version, which - with sh404SEF will cause a redirect to the correct, canonical, mixed-case one. Infinite loop. Dead site.

simbus82:
> @simbus82 Really? me? wrong idea about Joomla routing?
> 
> > You may have forgotten that the alias field of menu items, articles, categories, etc. turns everything into lowercase.
> 
> You may have forgotten about SEF extensions. What you state here is only valid for sites using Joomla routing.
> 
> > I have done more than 400 websites (true business website, not for the pub down the road). I have never had a mixedcase url or a single uppercase url with Mambo, Joomla 1.x, 2,x, 3.x and 4.x.
> 
> Don't want to get into my is bigger than yours, but there are several **dozens of thousands** - at the very least, a more reasonable estimate is well above 100,000 - with such URLs. And that's just for sh404SEF. JoomSEF and others have to be added. Even if you talk about Joomla 4, because they are all migrating now - and keeping their mixed-case URLs, because we never want to change URL.
> 
> > And it's a worldwide approved assumption that lowercase for URL it's the best error-proof mode to creating urls. For backlinking too, that is really important for SEO.
> 
> Which URLs are created has nothing to do with our discussion. The problem is backward compatibility.
> 


Joomla world or Joomla routing? Make a decision.

I guess you didn't understand.
I am not a developer, I decide how the sites (lead generation, ecommerce) should be done to perform at a technical, commercial, sales and branding level.
No competition who has the longest. The numbers are based on the projects followed, you develop plugins, I do digital marketing projects for companies.
In over 400 sites followed (which is an excellent statistical basis), NO SITE has ever generated (or wanted to generate) url with a casing other than lowercase.
Among other things, I think I used a SEF extension in Joomla maybe in 2012, a well structured site does not absolutely need it.

Again, even an old site in Joomla 2.5 cannot have URLs other than the lowercase. They can only be generated by poorly made plugins and extensions.

> If this PR goes as is, any site using mixed-case URL is dead: all their pages will be redirected to the lower-case version, which - with sh404SEF will cause a redirect to the correct, canonical, mixed-case one. Infinite loop. Dead site.

Not a Joomla core problem so. Isn't it that with your plugins you have allowed anyone (who it's bad at structuring a site) to generate URLs of this type?

PS: i like the only canonical solution without forced redirect, but the wrong url (upper or mixed case) come from outside, not from the core.

weeblr:
Hi

> Joomla world or Joomla routing? Make a decision.

Joomla routing is definitely part of the Joomla world, and in a discussion about Joomla URLs it seems appropriate to have knowledge about both. I am rather experienced about both.

> In over 400 sites followed (which is an excellent statistical basis), 

Sorry to be blunt but no, that's an extremely tiny number of websites.

> No competition who has the longest. The numbers are based on the projects followed, you develop plugins, I do digital marketing projects for companies.

Not sure what "develop plugins" means in your mind, but I can guarantee you my job is not really about code. It's about what the code should be doing. You realize I am the developer of sh404SEF, 4SEF and 4SEO, right?

> Among other things, I think I used a SEF extension in Joomla maybe in 2012, a well structured site does not absolutely need it.

Don't see how this is relevant to the discussion. What's relevant is if there are Joomla websites, today, including Joomla 4 ones, that use them.

> Again, even an old site in Joomla 2.5 cannot have URLs other than the lowercase. They can only be generated by poorly made plugins and extensions.

This is incorrect and likely due to the small number of sites you base this assumption on, or rather the fact the sites you work on have hired you, which of course is something an infinitely tiny number of Joomla site can do. 

But again and more importantly, this is  irrelevant. The question is not whether it's good or bad, but whether it is so and what to do about it.

> Not a Joomla core problem so. Isn't it that with your plugins you have allowed anyone (who it's bad at structuring a site) to generate URLs of this type?

So you'd say a few hundred thousands website is not Joomla core problem? Consideting the state of the "Joomla world", I'd say it is (and always was as a matter of fact).

Let's backtrack a bit and remember that all this started because Joomla 4 routing was modified so that only lower-case URLs are now accepted, mixed-case will trigger a 404. 

I'm pretty experienced in that field, and I'd say that this needs to be handled with much care. Yes, it may work in the general case but there's a large number of sites, using extensions, for which this will not work at all. So there needs to be a hook, a way to disable this (programmatically) so that extension can prevent it from happening.

PS: and also, add the GET method check

joeforjoomla:
@weeblr actually the Joomla core has never allowed to create not lowercase URLs. And this is correct and the right way of dealing with URLs. The problem is only related to you that are the developer of sh404SEF. Indeed in many years i've never agreed to the fact the sh404SEF allowed to create bad sNakecASe URLs.
So if this PR get merged the problem to the Joomla community and to broken websites is caused by your extension.

brianteeman:
Sorry Joe you are wrong!!!

When marketing my Joomla 3 sites I could write the url as EXAMPLE.COM/NEWS and it worked
In Joomla 4 the same url will produce a 404

This is nothing to do with any extension. It is about a change in behaviour between 3 and 4 that has significant impact

weeblr:
@joeforjoomla and the other extensions doing this. 

Can I ask what exactly is your argument? A few dozens thousands Joomla site broken is fine because they used a specific extension in a specific way some years back?

What's happening here? what is this discussion?

simbus82:
> Sorry Joe you are wrong!!!
> 
> When marketing my Joomla 3 sites I could write the url as EXAMPLE.COM/NEWS and it worked In Joomla 4 the same url will produce a 404

@brianteeman, i think you didn't understand what he meant.
**You can't CREATE non-lowercase ALIAS/URL insides Joomla** (since 1.x), but obviously as you say if you "type in the browser" (call an url) in uppercase, the j3 site responds correctly. Joe wrote correctly.
URL Creation vs URL Calling.

> This is nothing to do with any extension. It is about a change in behavior between 3 and 4 that has significant impact

Sure, but the problem now is the detractors.
So now the problem is with the plugin! And @joeforjoomla got it right.

sh404 probably changes the rules of URL creation in Joomla and how Joomla responds to URL calls in the browser.
If i remember well, you can create short url like www.mysites/MiXeD23 and www.mysites/mixed23  and if you call these url they work like different url, but it's a thing promoted by the plugin.
Should be good as an upper layer of url management, but the Joomla core needs to stay away from these things.

> A few dozens thousands Joomla site broken is fine because they used a specific extension in a specific way some years back?

Are you serious?
Yes, it's fine, because ALL webmasters expect that is the plugin dev's that adapt 3rd party plugins behavior in a migration.
Are you asking Joomla to change a good behavior in the core (this PR) to manage a weird behavior of your plugin? Really?
@weeblr if thousands of websites (with sh404 or similar plugins installed) have pages identified by URLs in mixed-case, unfortunately it's just their problem, **not a Joomla core problem.**
This means they have built the website url structure inside a plugin, not with the core logic. And this is wrong.
It's like using Gantry Particles and then crying because if you change the template you lose all the work done.

Or it's like complaining about the fact that switching to Joomla 4 from a J3 site with a "multi-category per article plugin" (which may not yet exist for J4), no longer works and the menu and navigation tree needs to be rebuilt.

We daily do redirection and url migration projects for clients that migrate from a website cms to another one: we have never blamed a CMS if we realize that a plugin has been used to manage the site structure differently than the standard.
It's a problem that every SEO, armed with Screaming Frog and an Excel sheet, can solve adding redirects to an htaccess.

We are really talking about nothing.

It's all a matter of how the CMS must NATIVELY treat (create) URLs and IN ANY CASE according to logic and the UX must always use lowercase URLs with a system that handles "URL call" errors (like J3 which anyway returns the correct page in any case).

Joomla must let us "create" url in lowercase, must prevent url creation in uppercase/mixedcase and must instead respond correctly to URL calls in mixedcase/uppercase (as you say, @brianteeman), without returning 404 and considering different url casing the same content, like J3 do (maybe with some SEO updates obviously).

simbus82:
I have tested this item :white_check_mark: successfully on e39e518be051d6f94b6397684261d44ac2d4b875<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38249">issues.joomla.org/tracker/joomla-cms/38249</a>.</sub>

Works well and respect query strings and url parameters
image

Good for respect the J3 behavior, maybe in future we can let users manage it with different settings (show the right content based on lowercased alias with a canonical without forced redirect)

Webdongle:
> @Webdongle Also I did not mis-read your question. Simply, your question is not a Yes/No question.
> 
> Or the question is a Yes / No, but the answer is not.

Hmmm
It is a yes/no answer

If mIxedcASe not found then convert to lower case.

Either the answer is Yes that would work
or No that would not work.

weeblr:
@simbus82 

> Are you serious?

Yes, because...

> Yes, it's fine, because ALL webmasters expect that is the plugin dev's that adapt 3rd party plugins behavior in a migration.

If this PR goes in, **it is technically impossible for any plugin developer to fix the problem** - because that redirect cannot be disabled or altered in any way. 

**As soon as a user update their site is dead.**. Nothing can be done by any one. Nada.

> but it's a thing promoted by the plugin.

No, it's a thing used by a large number of people who created their website and used my extensions or other developers extensions to achieve their goals. Which are not yours.

> Are you asking Joomla to change a good behavior in the core (this PR) to manage a weird behavior of your plugin? Really?
@weeblr if thousands of websites (with sh404 or similar plugins installed) have pages identified by URLs in mixed-case, unfortunately it's just their problem, not a Joomla core problem.

I cannot disagree more. Yes, my extension is involved. But honestly, if you think this way, you may have missed some important points in what Joomla is (you know, a community). 

> We daily do redirection and url migration projects for clients that migrate from a website cms to another one: we have never blamed a CMS if we realize that a plugin has been used to manage the site structure differently than the standard.
It's a problem that every SEO, armed with Screaming Frog and an Excel sheet, can solve adding redirects to an htaccess.

This is not about or what you do. You very obviously work on a specific type of (Joomla) sites where owners hire SEO and agencies. 

> This means they have built the website url structure inside a plugin, not with the core logic. And this is wrong.

Whether it's wrong is irrelevant. It is what it is. There is no reason to break all these sites for all these people just because this PR is too simple and does not take a global approach.

> It's all a matter of how the CMS must NATIVELY treat (create) URLs and IN ANY CASE according to logic and the UX must always use lowercase URLs with a system that handles "URL call" errors (like J3 which anyway returns the correct page in any case).

No, it has nothing to do with that. That point is well established and lowercase has been the standard for many years. Nobody has argued against it at any point. 
This discussion is about how the core should handle requests that depart from this standard. And how the proposed solution will break a large number of real, current, up to date Joomla websites if published.


Also, I think you missed along that **I do not advocate against this PR in any way**. I actually very much favor it. **It's just not ready to go in because it has:**

- a significant technical issue (not checking for GET request) - this may damage a LOT of just core Joomla sites
- a heavy handed approach disallowing a significant number of Joomla websites to continue operating, without any reason

**So in short, this PR needs to be fixed before being accepted.**

> Joomla must let us "create" url in lowercase, must prevent url creation in uppercase/mixedcase and must instead respond correctly to URL calls in mixedcase/uppercase (as you say, @brianteeman), without returning 404 and considering different url casing the same content, like J3 do (maybe with some SEO updates obviously).

Yes, as every one here agreed, suggested, since day one, and since the previous PR that @brianteeman opened. 

What exactly have you understood I suggested? 


@Webdongle 

> Either the answer is Yes that would work or No that would not work.

Well, it would work for any who used only Joomla core on their site and it would kill a large number of Joomla sites which used sh404SEF, JoomSEF, MiJoSEF, AceSEF and such.

@richard67 Please check the code here. This PR does a  redirect without checking the HTTP request, ie it will redirect POST, DELETE, etc

richard67:
> @richard67 Please check the code here. This PR does a redirect without checking the HTTP request, ie it will redirect POST, DELETE, etc

@weeblr I've set the RLDQ label, not the RTC label.

Fedik:
I am closing it, before someone get heart attack :neckbeard: 

The fix is fine for "private use". 

For the core we need to think something else.
It probably should be in Router level, kind of: 
```php
$doc->setCanonical($componentRouter->getCanonical());
```
or 
```php
if ($requestUri !== $componentRouter->getCanonical()) {
  $app->redirect($componentRouter->getCanonical());
}
```

Fedik:
I have opened issue. Can continue disscussion there #38257.
But would be nice with some technical and practical suggestion, not just general talk.
Thanks.

weeblr:
@Fedik Working on that. Indeed I also do it with an added parseRule in the router. No reason to do it elsewhere. Will post code.

joeforjoomla:
> Sorry Joe you are wrong!!!
> 
> When marketing my Joomla 3 sites I could write the url as EXAMPLE.COM/NEWS and it worked In Joomla 4 the same url will produce a 404
> 
> This is nothing to do with any extension. It is about a change in behaviour between 3 and 4 that has significant impact

@brianteeman sorry but there are also situations in which also Joomla 3 throws a 404 page changing the case of a URL. Indeed the com_content router does not clear $segments and the $uri path.
I've still not 100% clear in which situation this can happen, but it happens.

See:

image

AND:

image

Issue #38265, opened 2022-07-13
[4.2.0] JCategories call results in no DatabaseInterface


OctavianC:
### Steps to reproduce the issue
```php
$categories = new JCategories(array('extension' => 'com_content', 'table' => '#__content'));
print_r($categories->get('root'));
```

### Expected result
```
Joomla\CMS\Categories\CategoryNode Object
(
    [id] => root
    [asset_id] => 0
    [parent_id] => 0
...
```

### Actual result
`Resource 'Joomla\CMS\Categories\DatabaseInterface' has not been registered with the container. `

### System information (as much as possible)
Latest nightly
These packages were last built: Wednesday, 13 July 2022 00:11:27 UTC

### Additional comments
Can be fixed by calling `$categories->setDatabase(JFactory::getDbo());` but that seems a B/C break. Code works on Joomla! 4.1.5 fine.

chmst:
Where did you find this line? Which Version of Joomla, which Version of which extension?

$categories = new JCategories(array('extension' => 'com_content', 'table' => '#__content'));``

OctavianC:
It's just an example. Obivously it's an extension that uses the [Categories class](https://docs.joomla.org/Categories_and_CategoryNodes_API_Guide) but that's not the point.

Here, use this code which is according to the docs:

```php
use Joomla\CMS\Categories\Categories;

class TestCategories extends Categories
{
    public function __construct($options = array())
    {
        $options['table']      = '#__test_items';
        $options['extension']  = 'com_test';
        $options['statefield'] = 'published';

        parent::__construct($options);
    }
}

$categories = JCategories::getInstance('Test');
var_dump($categories->get('root'));
```
### 4.2.0 Result
`0 Resource 'Joomla\CMS\Categories\DatabaseInterface' has not been registered with the container. `

### 4.1.5 Result
`object(Joomla\CMS\Categories\CategoryNode)[894]`

OctavianC:
Found it

brianteeman:
and?

OctavianC:
... and I've fixed it. Please test #38269

PR #38269, opened 2022-07-13
DatabaseInterface missing namespace causing JCategories call to fail

OctavianC:
Pull Request for Issue #38265 .

### Testing Instructions
See #38265

     0 Resource 'Joomla\CMS\Categories\DatabaseInterface' has not been registered with the container. 

Works


wilsonge:
Clearly correct. Merging on review

PR #38290, opened 2022-07-17
[4.2] New shortcuts

brianteeman:
Following feedback this PR adds two new shortcut keys
Save & New
Save as Copy

image


Kostelano:
I can not "catch" the work of new combinations.

Moreover, they can work quite well, but the speed / quality of the plugin is poor. By pressing other combinations about 8-10 times, the combination works for me a maximum of a couple of times. In this situation, by pressing `J+SN`, I caught `J+S` a couple of times and that's it.

Given this, I would prefer to use a combination of 2 buttons at most.

Kostelano:
Just thoughts. Given the prospect of expanding combinations for the plugin in the future + simplifying so that any combination consists of simple 2 buttons, we could divide them into groups.

For example, group `J` (base):

`J + S` Search,
`J + O` Options,
`J + H` Help,
`J + M` Toggle Menu,
`J + Х` Overview,
`J + D` Home Dashboard,

Group `H` (conditionally ... The letter `A` from the word `actions` will be too close to labels.):

`H + S` Save,
`H + C` Save & Close,
`H + N` Save & New,
`H + D` Save & Copy,
`H + Q` Cancel/Close

This will make it easier to select a paired letter (label).

Again, all of this only makes sense if the plugin is going to be extended in the core.

brianteeman:
can't say i am a fan. much better to just have one key act as the trigger. Also I cant replicate your issue with activating the shortcuts

chmst:
> Moreover, they can work quite well, but the speed / quality of the plugin is poor. By pressing other combinations about 8-10 times, the combination works for me a maximum of a couple of times.

Probably you pressed the keys too fast? There must be a small time between. You can manipulate the time in the plugin. 

About grouping the keys: For normal users the grouping is not obvious (remember: don't make me think). 
My opinion is, that users have a kind of reflex in pressing shortcut keys ... always first J ..and should not think about which one could be the first key.

HLeithner:

PR #38337, opened 2022-07-28
full width spacer hr line

svanschu:
Pull Request for Issue # . No issue created for it.

My first Joomla! PR ...

### Summary of Changes
In the current Joomla 4 backend the space hr form field only creates a line partially.
Bildschirmfoto vom 2022-07-28 19-50-15
With this changes the hr line is again on the full width and looks much better as spacer.
The changes are only made for the hr=true case, because if nothing is shown it doesn't matter. 
If text is used it only gets displayed as label and hr=false need to be set.
For bigger text the note form field should be used.
Bildschirmfoto vom 2022-07-29 00-31-17

Open Modules: Articles - Category Filtering-Options and you see a small line space left aligned.

After the patch you should see a full width line

### DIFF
--- a/build/media_source/templates/administrator/atum/scss/blocks/_form.scss
+++ b/build/media_source/templates/administrator/atum/scss/blocks/_form.scss
@@ -49,14 +49,15 @@
   }
 }
 
-.stack .control-label {
-  width: 100%;
+.field-spacer {
+  .control-label {
+    padding: 0;
+    width: 100%;
+  }
 }
 
-.spacer {
-  hr {
-    width: 380px;
-  }
+.stack .control-label {
+  width: 100%;
 }
 
 td .form-control {
--- a/build/media_source/templates/administrator/atum/scss/blocks/_form.scss
+++ b/build/media_source/templates/administrator/atum/scss/blocks/_form.scss
@@ -51,8 +51,8 @@
 
 .field-spacer {
   .control-label {
-    padding: 0;
     width: 100%;
+    padding: 0;
   }
 }
 
--- a/libraries/src/Form/Field/SpacerField.php
+++ b/libraries/src/Form/Field/SpacerField.php
@@ -31,6 +31,14 @@ class SpacerField extends FormField
     protected $type = 'Spacer';
 
     /**
+     * Hide the description when rendering the form field.
+     *
+     * @var    boolean
+     * @since  4.0.0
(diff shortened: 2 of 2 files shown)


RickR2H:
I have tested this item :white_check_mark: successfully on 4568771a58326159bff877bcc638310283294f44<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38337">issues.joomla.org/tracker/joomla-cms/38337</a>.</sub>

drmenzelit:
Above the function `getInput()` there is a comment:
`Method to get the field input markup for a spacer. The spacer does not have accept input.`
so adding code here seems wrong to me. And the html code is already being created in the function `getLabel()`
I think a change of the CSS should be enough to get a 100% wide line

`.spacer hr {
width: 100%;
}
.control-group.field-spacer .control-label {
width: 100%;
}`

svanschu:
> Above the function `getInput()` there is a comment: `Method to get the field input markup for a spacer. The spacer does not have accept input.` so adding code here seems wrong to me. And the html code is already being created in the function `getLabel()` I think a change of the CSS should be enough to get a 100% wide line
> 
> `.spacer hr { width: 100%; } .control-group.field-spacer .control-label { width: 100%; }`

Pure CSS is not enough. Either this patch or you need to rewrite the spacer field completely to be like the note field. with own HTML without the normal FormField behavior.

drmenzelit:
grafik

Content / Options
With CSS only

Can you give an example where that doesn't work?

chmst:
Css works for me. 
Seems to be a "old" fild. Is there a reason for the fixed width of 380px? And for bootstrap popOver?

svanschu:
@drmenzelit yes you're right sorry. 
@chmst yes that's a pretty old field. I thought about to rewrite it so it really only makes a space or a line without the option to enter text, because since J3.1 we have the note field which adds text full width.
What do you think?

chmst:
@svanschu thank you very much for all your work and commitment here. 
Unfortunately thsi PR cannot be accepted, it would be a b/c break, as until now  the text was displayed and could be used by 3rd party developers in their forms. 

Now as you have made your first PR, you know how this works and we are looking forward to new PRs from you :)

PR #38343, opened 2022-07-30
[4.2] Codemirror Update

brianteeman:
20-07-2022: [Version 5.65.7](https://codemirror.net/5/codemirror-5.65.7.zip):

Fix several references to the global document/window, improving support for creating an editor in another frame.
[vim bindings](https://codemirror.net/5/demo/vim.html): Use [upstream](https://github.com/replit/codemirror-vim/) code instead of keeping our own copy.
[tern addon](https://codemirror.net/5/demo/tern.html): Properly HTML escape variable names in rename dialog.

20-08-2022: [Version 5.65.8](https://codemirror.net/5/codemirror-5.65.8.zip):

Include direction override and isolate characters in the default set of special characters.
Fix an issue that could cause document corruption when mouse-selecting during composition.
[foldgutter addon](https://codemirror.net/5/doc/manual.html#addon_foldgutter): Refresh markers when the editor’s mode changes.
[merge addon](https://codemirror.net/5/doc/manual.html#addon_merge): Fix syntax that prevented the addon from loading in IE10.

Once again we missed merging an update so this pr now has two


brianteeman:
Closing as we no longer need this manual pr

PR #38378, opened 2022-08-03
Content Plugins are not triggered upon batch copy

grantg182:
Content plugins are not being triggered when an article is batch copied, despite the model `cleanupPostBatchCopy` function triggering the content plugins.

Content plugins are not triggered without importing the content plugins, so now import the content plugins before triggering the event.

### Summary of Changes

Import content plugins before triggering them.

### Testing Instructions

Batch copy an article, onContentAfterSave ought to be triggered upon saving the new article.

onContentBeforeSave is not triggered.

onContentBeforeSave is triggered.

### DIFF
--- a/administrator/components/com_content/models/article.php
+++ b/administrator/components/com_content/models/article.php
@@ -87,7 +87,8 @@ class ContentModelArticle extends JModelAdmin
 				$fieldsData['com_fields'][$field->name] = $field->rawvalue;
 			}
 		}
-
+		
+		JPluginHelper::importPlugin('content');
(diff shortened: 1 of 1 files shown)


HLeithner:
@zero-24 since it's not merged yet (before final non security release) this can be rejected?

Hackwar:
@grantg182 Thank you for this PR. As Brian wrote, we aren't accepting any more bugfixes for Joomla 3 anymore. However, this is still an issue in Joomla 4. I created a PR in #39013 and would be very happy if you could test this, so that we can merge it as soon as possible.

PR #38398, opened 2022-08-04
[5.0] Rector

nibra:
### Summary of Changes

* Add Rector
* Update dependencies (PHP:^8.1)
* Apply Rector rule set LevelSetList::UP_TO_PHP_81
* Fix issues detected during automated refactoring

### Testing Instructions

TBD

TBD


administrator/components/com_actionlogs/src/Field/LogsdaterangeField.php
```diff
@@ -35,14 +35,7 @@ class LogsdaterangeField extends PredefinedlistField
...
      */
-    protected $predefinedOptions = array(
-        'today'       => 'COM_ACTIONLOGS_OPTION_RANGE_TODAY',
-        'past_week'   => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_WEEK',
-        'past_1month' => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_1MONTH',
-        'past_3month' => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_3MONTH',
-        'past_6month' => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_6MONTH',
-        'past_year'   => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_YEAR',
-    );
+    protected $predefinedOptions = ['today'       => 'COM_ACTIONLOGS_OPTION_RANGE_TODAY', 'past_week'   => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_WEEK', 'past_1month' => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_1MONTH', 'past_3month' => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_3MONTH', 'past_6month' => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_6MONTH', 'past_year'   => 'COM_ACTIONLOGS_OPTION_RANGE_PAST_YEAR'];
```
wilsonge, line 38: There has to be a setting for this kinda codestyle. this isn't readable and how array's are structured aren't part of PSR/PER's

administrator/components/com_actionlogs/src/Helper/ActionlogsHelper.php
```diff
@@ -171,10 +159,10 @@ public static function getLogContentTypeParams($context)
      */
     public static function getHumanReadableLogMessage($log, $generateLinks = true)
     {
-        static $links = array();
+        static $links = [];
 
         $message     = Text::_($log->message_language_key);
-        $messageData = json_decode($log->message, true);
+        $messageData = json_decode((string) $log->message, true, 512, JSON_THROW_ON_ERROR);
```
wilsonge, line 165: This is a b/c change and without appropriate try/catch's we shouldn't be doing as part of this round of codestyle. If this. is going to be a b/c change in 5 it. at least requires documentation.

administrator/components/com_actionlogs/src/Field/LogsdaterangeField.php
nibra: That step is in preparation

administrator/components/com_actionlogs/src/Model/ActionlogsModel.php
```diff
@@ -378,7 +370,7 @@ public function purge()
      *
      * @since   3.9.0
      */
-    public function getFilterForm($data = array(), $loadData = true)
+    public function getFilterForm($data = [], $loadData = true): Form|bool
```
wilsonge, line 373: I'm confused by this change in relation to the return type hints - aren't they supposed to match the parent ones or give a strict error? Also why is it being added here but not in the parent in that's the case. Just seems inconsistent.

administrator/components/com_actionlogs/src/Helper/ActionlogsHelper.php
nibra: This is not so much about code style, but about rule-based refactoring.

administrator/components/com_actionlogs/src/Plugin/ActionLogPlugin.php
```diff
@@ -76,17 +79,17 @@ protected function addLog($messages, $messageLanguageKey, $context, $userId = nu
             }
 
             if (\array_key_exists('type', $message)) {
-                $message['type'] = strtoupper($message['type']);
+                $message['type'] = strtoupper((string) $message['type']);
```
wilsonge, line 82: Do we really want to typehint this? If people are actually pushing in non-strings here I'd assume we're better of erroring than pushing a stringified object into strtoupper. Feels like this should be on a case by case basis rather than applied globally.

administrator/components/com_admin/src/Model/SysinfoModel.php
```diff
@@ -202,22 +201,21 @@ protected function cleanPrivateData(array $dataArray, string $dataType = 'other'
...
-     * @return  string|array
      *
      * @since   3.5
      */
-    protected function cleanSectionPrivateData($sectionValues)
+    protected function cleanSectionPrivateData($sectionValues): string|array
     {
         if (!\is_array($sectionValues)) {
-            if (strstr($sectionValues, JPATH_ROOT)) {
+            if (strstr((string) $sectionValues, (string) JPATH_ROOT)) {
```
wilsonge, line 210: Another example. Maybe rector needs typehints for JPATH_ROOT - because we know this will always be a string and you physically can't reach this method if it's not because the application would never have booted.

administrator/components/com_associations/layouts/joomla/searchtools/default.php
```diff
@@ -58,18 +58,7 @@
...
-    'orderFieldSelector'  => '#list_fullordering',
-    'showNoResults'       => !empty($noResultsText),
-    'noResultsText'       => !empty($noResultsText) ? $noResultsText : '',
-    'formSelector'        => !empty($data['options']['formSelector']) ? $data['options']['formSelector'] : '#adminForm',
-);
+$customOptions = ['filtersHidden'       => isset($data['options']['filtersHidden']) && $data['options']['filtersHidden'] ? $data['options']['filtersHidden'] : $hideActiveFilters, 'filterButton'        => isset($data['options']['filterButton']) && $data['options']['filterButton'] ? $data['options']['filterButton'] : $showFilterButton, 'defaultLimit'        => $data['options']['defaultLimit'] ?? Factory::getApplication()->get('list_limit', 20), 'searchFieldSelector' => '#filter_search', 'selectorFieldName'   => $selectorFieldName, 'showSelector'        => $showSelector, 'orderFieldSelector'  => '#list_fullordering', 'showNoResults'       => !empty($noResultsText), 'noResultsText'       => !empty($noResultsText) ? $noResultsText : '', 'formSelector'        => !empty($data['options']['formSelector']) ? $data['options']['formSelector'] : '#adminForm'];
```
wilsonge, line 61: Another place where the array flatenning becomes nearly impossible to read....

administrator/components/com_actionlogs/src/Plugin/ActionLogPlugin.php
nibra: Before, `null` was accepted and treated as an empty string. This just mimics the behavior.

administrator/components/com_actionlogs/src/Helper/ActionlogsHelper.php
wilsonge: I understand. But if all json decodes previously returned false and now throw an exception - that's a b/c break. I don't think it's one that you intended else you'd wouldn't have put 'nothing' in the docs required part of the PR :)

wilsonge:
Big fan overall of adding rector in and know it's in draft so isn't ready yet - but first impressions are there's quite a long way to go yet. happy to help carrying on reviews though to make this mergable.

administrator/components/com_admin/src/Model/SysinfoModel.php
nibra: Sure. As soon as there is chance for Rector to infer the type, it will remove the type cast again.

administrator/components/com_associations/layouts/joomla/searchtools/default.php
nibra: I know, phpcbf is still running :(

brianteeman:
could you prefix the PR title with [5.0] please.

administrator/components/com_actionlogs/src/Helper/ActionlogsHelper.php
```diff
@@ -68,14 +65,7 @@ public static function getCsvData($data): Generator
...
 
-            yield array(
-                'id'         => $log->id,
-                'message'    => self::escapeCsvFormula(strip_tags(static::getHumanReadableLogMessage($log, false))),
-                'extension'  => self::escapeCsvFormula(Text::_($extension)),
-                'date'       => (new Date($log->log_date, new \DateTimeZone('UTC')))->format('Y-m-d H:i:s T'),
-                'name'       => self::escapeCsvFormula($log->name),
-                'ip_address' => self::escapeCsvFormula($log->ip_address === 'COM_ACTIONLOGS_DISABLED' ? $disabledText : $log->ip_address)
-            );
+            yield ['id'         => $log->id, 'message'    => self::escapeCsvFormula(strip_tags(static::getHumanReadableLogMessage($log, false))), 'extension'  => self::escapeCsvFormula(Text::_($extension)), 'date'       => (new Date($log->log_date, new \DateTimeZone('UTC')))->format('Y-m-d H:i:s T'), 'name'       => self::escapeCsvFormula($log->name), 'ip_address' => self::escapeCsvFormula($log->ip_address === 'COM_ACTIONLOGS_DISABLED' ? $disabledText : $log->ip_address)];
```
brianteeman, line 68: This is not very readable. Plus if it really is to be flattened like this shouldnt the spaces be trimmed?

administrator/components/com_associations/src/Helper/AssociationsHelper.php
```diff
@@ -10,6 +10,8 @@
 
 namespace Joomla\Component\Associations\Administrator\Helper;
 
+use Joomla\CMS\Association\AssociationExtensionHelper;
```
brianteeman, line 13: This change is not following the code style of alphasorting the use statements

administrator/components/com_associations/src/View/Associations/HtmlView.php
```diff
@@ -10,6 +10,8 @@
 
 namespace Joomla\Component\Associations\Administrator\View\Associations;
 
+use Joomla\CMS\Pagination\Pagination;
```
brianteeman, line 13: another example of the alphasorting being broken

administrator/components/com_banners/tmpl/banners/default.php
```diff
@@ -192,7 +192,7 @@
                             $this->loadTemplate('batch_body')
                         ); ?>
                     <?php endif; ?>
-                <?php endif; ?>
+<?php endif; ?>
```
brianteeman, line 195: I was playing with rector this morning and saw this type of change - which doesn't appear to make any sense

administrator/components/com_banners/tmpl/clients/default.php
```diff
@@ -180,7 +180,7 @@
                     <?php // Load the pagination. ?>
                     <?php echo $this->pagination->getListFooter(); ?>
 
-                <?php endif; ?>
+<?php endif; ?>
```
brianteeman, line 183: another example of the removal of indentation without any apparent logic?

administrator/components/com_actionlogs/src/Helper/ActionlogsHelper.php
nibra: Code style fixes have not been applied yet

administrator/components/com_admin/src/Model/SysinfoModel.php
wilsonge: Yeah but point being this is just in the repo so you'll never have that type. Doesn't mean adding a type hint is right either though.

libraries/src/Version.php
```diff
@@ -253,7 +251,7 @@ public function getMediaVersion(): string
         // Load the media version and cache it for future use
         if (self::$mediaVersion === null) {
             self::$mediaVersion = $this->getMediaVersionCache()
-                ->get([$this, 'generateMediaVersion'], [], md5('_media_version' . $this->getLongVersion()));
+                ->get($this->generateMediaVersion(...), [], md5('_media_version' . $this->getLongVersion()));
```
richard67, line 254: Does this really work with `...`?

When going to the prebuilt packages for this PR, the page shows a syntax error for the Version.php file: https://ci.joomla.org/artifacts/joomla/joomla-cms/5.0-dev/38398/downloads/56678/

For other PRs for the 5.0-dev branch the prebuilt packages page does not show that syntax error: https://ci.joomla.org/artifacts/joomla/joomla-cms/5.0-dev/38396/downloads/56677/

Could it be related to this change here?

nibra: https://www.php.net/manual/en/functions.first_class_callable_syntax.php

richard67: Hmm but then there must be a different problem.

administrator/components/com_actionlogs/src/Controller/ActionlogsController.php
```diff
@@ -85,7 +85,7 @@ public function exportLogs()
         if (\count($data)) {
             try {
                 $rows = ActionlogsHelper::getCsvData($data);
-            } catch (InvalidArgumentException $exception) {
```
laoneo, line 88: Looks wrong

nibra: But it's not. `$exception` is not used in the catch branch and no longer needed to be declared. Works like 'case instanceof ...'

laoneo: Forgot that none capturing catches are already available in 8.1

HLeithner:
I'm closing this for now till you have a better solution for this ;-)

PR #38399, opened 2022-08-04
Codestyle indent

brianteeman:
Small change to fix codestyle indent

code review

### DIFF
--- a/administrator/components/com_banners/tmpl/banners/default.php
+++ b/administrator/components/com_banners/tmpl/banners/default.php
@@ -181,7 +181,7 @@ if ($saveOrder && !empty($this->items)) {
                     $user->authorise('core.create', 'com_banners')
                         && $user->authorise('core.edit', 'com_banners')
                         && $user->authorise('core.edit.state', 'com_banners')
-) : ?>
+                    ) : ?>
                         <?php echo HTMLHelper::_(
                             'bootstrap.renderModal',
                             'collapseModal',
--- a/administrator/components/com_banners/tmpl/banners/default.php
+++ b/administrator/components/com_banners/tmpl/banners/default.php
@@ -177,8 +177,9 @@ if ($saveOrder && !empty($this->items)) {
                     <?php echo $this->pagination->getListFooter(); ?>
 
                     <?php // Load the batch processing form. ?>
-                    <?php if (
-                    $user->authorise('core.create', 'com_banners')
+                    <?php
+                    if (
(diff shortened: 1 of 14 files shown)


richard67:
It seems that the PHPCS check in drone doesn’t like your changes.

brianteeman:
The failure is what caused the problem in the first place - /me confused

richard67:
Me also confused because your code style looks right to me and the other not.

@HLeithner Do you have any idea why PHPCS currently accepts wrong code style in files in the drone checks but complains about this PR here?

brianteeman:
fyi the current lack of indent comes from the psr12 work

HLeithner:
the reason is that phpcs starts to count after `<?php ` so the if is a position 0, I would move the if into the next line.

HLeithner:
> fyi the current lack of indent comes from the psr12 work

for phpcs it's correct for us it's bad...

brianteeman:
thanks for the explanation. Suprised that the same situation doesnt arise with the echo directly afterwards

brianteeman:
I can see there about 15 instances of this. If the new updated change with your recommendation passes then I will make the same change for them all

chmst (approved):

Quy:
I have tested this item :white_check_mark: successfully on 8cd6df696b3090b3ff53db5bfc97629a40c05470<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38399">issues.joomla.org/tracker/joomla-cms/38399</a>.</sub>

viocassel:
I have tested this item :white_check_mark: successfully on 8cd6df696b3090b3ff53db5bfc97629a40c05470<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38399">issues.joomla.org/tracker/joomla-cms/38399</a>.</sub>

richard67:
RTC<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38399">issues.joomla.org/tracker/joomla-cms/38399</a>.</sub>

wilsonge:
Thanks!

brianteeman:
thanks

Issue #35381, opened 2021-08-26
[4.0] Override update info improvement


bembelimen:
### Is your feature request related to a problem? Please describe.
When making an update where overriden files were changes, you get a message.
![grafik](https://user-images.githubusercontent.com/1229869/130933978-560bee74-48d0-4ee6-ae94-e9ef90fee1ff.png)

### Describe the solution you'd like
It would be helpful to have a link to click, where the overrides could be checked.

@astridx ?

RussW:
is it not sufficient that on the "Home Dashboard" there is a link to the overrides, the template style with the change then has a badge stating it has a change, the file list for the template style with the change also has a tab "Updated Files" that then can be selected/checked and acknowledged that you have seen it and taken any appropriate action, which in turn removes the notification.

brianteeman:
A link in the message would be helpful even if it just goes to /administrator/index.php?option=com_templates&view=templates

Wherever possible we should aim to have a "solution" in the message. In this case the "solution" is the link

Scrabble96:
I have just upgraded a site to 4.0.5 and had a message that 4 overrides had changed, but absolutely no indication as to which of the many overrides I have created had changed. What use is a message like that?

And, presumably, it is not the overrides that have changed but the original core files that the overrides were based on?

astridx:
@Scrabble96 
> I have just upgraded a site to 4.0.5 and had a message that 4 overrides had changed, but absolutely no indication as to which of the many overrides I have created had changed. What use is a message like that?
And, presumably, it is not the overrides that have changed but the original core files that the overrides were based on?

For me, the procedure is fine.

After the update, I see a note:
![1](https://user-images.githubusercontent.com/9974686/146353465-188f6616-7be4-41b8-9c16-0f0e106cbd47.PNG)

The red symbol "Updates to check" leads me per click to the Template Manager:
![2](https://user-images.githubusercontent.com/9974686/146353470-d103ab1b-cedb-4ae8-8199-d676bd2ff428.PNG)

In the Template Manager I can see which templates are affected and I can open the template for editing per click on the name:
![2a](https://user-images.githubusercontent.com/9974686/146353473-0b93c52c-b94e-4021-8f85-a984434afb09.png)

When I edit the template, I see in the tab `updatet files`, a link through which I can open each individual file to change.
![2b](https://user-images.githubusercontent.com/9974686/146353476-e1e380a5-3109-44c5-9df0-44640369db74.PNG)

[...]

PR #38401, opened 2022-08-05
template override diff - a11y ux

brianteeman:
### Summary of Changes
The "Diff between original and overridden file" uses color alone to indicate the changes. 

- This is an accessibility failure as you should never rely on a single visual indicator for sighted users.
- This is an accessibility failure as there is no indicator at all for non-sighted users.
- This is a UX failure as some users (#35381) have reported they expected the red/green to be reversed
- This is a semantic failure as we should use the `<del>` and `<ins>` elements as they are specifically created for diffs to indicate [text added and removed](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#demarcating_edits)

This PR addresses all these failures by 
- adding a plus or minus symbol at the beginning of the line
- using the del/ins elements instead of the span
- uses the browser defauly styling for these elements of strikethrough and underline

### Testing Instructions
This is a js change so either use a prebuilt package or `npm run build:js`

Create a template override and make some changes.
Save the override and then display the differences.

## Before
image

## After
image

### Note
js isn't my thing so maybe it can be improved cc @fedik


build/media_source/com_templates/js/admin-template-compare.es6.js
```diff
@@ -37,18 +39,27 @@
...
-        pre = document.createElement('pre');
+
+        pre = document.createElement(elm);
         pre.style.backgroundColor = color;
         pre.className = 'diffview';
         pre.appendChild(document.createTextNode(decodeHtmlspecialChars(part.value)));
         fragment.appendChild(pre);
+        pre.insertAdjacentHTML('afterbegin', `<span class="indicator">${indicator}</span>`);
       });
+      display.appendChild(fragment);
```
Fedik, line 62: this is duplicated :)
few lines down the same code

brianteeman:
In the 5 months since this was created and completed ignored by the @joomla/joomla-accessibility-team-jat  it has been superceded by #38823

PR #38417, opened 2022-08-08
Enable testing of more than Libraries

tatankat:
### Summary of Changes
Remove Libraries subdirectory from directory configuration in phpunit config, so Component and Plugin tests are also executed.

### Testing Instructions
run phpunit and see more tests being done

765 tests executed

786 tests executed

None that I know of.


rdeutz:
1) If you want to add more tests than you can add more testsuites.
2) Our CI fails on you change, what is not a good sign.

tatankat:
> 1. If you want to add more tests than you can add more testsuites.

I do not want to add more tests. The tests are already written and part of the repository, they are just not executed.

>  2. Our CI fails on you change, what is not a good sign.

The CI fails on the tests that were already there. Problem is they do not fail when I run them locally. I have no idea why they fail on the CI.

laoneo:
The problem is that in drone the tests are executed as root, so it can create the invalid folders. I made an alternative pr where I fixed the tests in #38421. Can you have a look on this pr @tatankat?

tatankat:
@laoneo Nice, thanks for looking into it and the work done! I will close this one in favor of #38421

Issue #28800, opened 2020-04-25
[4.0] Make captcha plugins use plugin events


richard67:
### Steps to reproduce the issue

Pull Request (PR) #28785 has fixed the recaptcha plugin in the 4.0-dev branch so that it works again.

The fix was to use change onInit(), onDisplay(), onCheckAnswer() from events to callbacks.

But according to [https://github.com/joomla/joomla-cms/pull/28785#issuecomment-619394249](https://github.com/joomla/joomla-cms/pull/28785#issuecomment-619394249) this is not like it was meant to be in J4.

Therefore it needs to investigate what has to be done so the callbacks can be changed back to events.

This issue is just a reminder for that so it is not forgotten.

joomdonation:
What should we do with this Release Blocker? @Fedik Look like revert your PHP changes on your original PR https://github.com/joomla/joomla-cms/pull/28785 and add the missing ```$this->_captcha->registerListeners();``` at the end of **_load** method of the Captcha class will have captcha still works, too. Could that be the solution?

Fedik:
tbh, I have no idea.
It not only about **captcha** but also about **editor** plugins.

Obviously it will be wrong to have `onInit`, `onDisplay` (etc. captcha, editor event) as global events.
at least it need a new event names, for captcha and editor.

It sounds like need to rewrite whole `Joomla\CMS\Captcha` and `Joomla\CMS\Editor` API. 
It also should deal somehow with multiple return result from event dispatching.

I would leave it in peace until better times :smile:

richard67:
Well, I had created this issue here on @wilsonge 's request ... so maybe he should check the above comment 😄

joomdonation:
Ah ah, so editor plugins also use **onInit** and **onDisplay** events. If so, Yes, I agree that we should leave it as how it is for now :D.

wilsonge:
Fine leave it as it is :( but i really dislike the fact that not all our plugins are actually using the event system. As mentioned it's easy to move it to an events system but needs something more complicated to ensure we only get back the item from the correct plugin (passing in the name of the active editor/captcha system or similar)

Fedik:
We can review it in future again.

I think, when the event dispatcher would allow to call a specific plugin, it could help much here.
Something like:
```php
// Load all editors and captcha
PluginHelper::importPlugin('editors'); 
PluginHelper::importPlugin('captcha'); 

// Selected editor in the configuration
$context = 'editors.tinymce'; // $group + $name 
[...]

PR #38441, opened 2022-08-12
[4.3] Convert recaptcha plugin to services and add captcha plugin interface

laoneo:
### Summary of Changes
The pr introduces a new interface for captcha based plugins so operations on them is more clear. Additionally it deprecates the `Captcha` class as it basically duplicates the code of the `bootPlugin` operation and does some method checks for plugins which do not implement the interface.

### Testing Instructions
- Setup the recaptcha and invisible captcha plugin in the plugin manager with the proper keys
- Select recaptcha in the contact options in the form tab
- Create a single contact menu item
- Open that contact menu item and send a mail as not logged in user
- Select invisible captcha in the contact options in the form tab
- Open that contact menu item and send a mail as not logged in user

Both times is a captcha challenge executed and two mails are sent out.

Both times is a captcha challenge executed and two mails are sent out.

When deprecated messages are already documented, then this one needs to be added.

### DIFF
--- a/libraries/src/Captcha/Captcha.php
+++ b/libraries/src/Captcha/Captcha.php
@@ -9,21 +9,22 @@
 
 namespace Joomla\CMS\Captcha;
 
+use Joomla\CMS\Extension\DummyPlugin;
 use Joomla\CMS\Factory;
 use Joomla\CMS\Filter\InputFilter;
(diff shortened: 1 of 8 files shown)


laoneo:
Closing this as it should be reverted to plugin events, as discussed here #28800.

PR #38244, opened 2022-07-07
[4.2] Menu item custom field

brianteeman:
### Summary of Changes
adds a custom field to show a list of all menu items that can be selected and then displayed as a link in the frontend

### Testing Instructions
Install a prebuilt package
or
Install this PR, discover the extension and publish the plugin

Create a menuitem custom field. Check that it shows all menuitems from all menus in the select ;ist and then check on the front end that the link is displayed.

### Expected result AFTER applying this Pull Request
image

image

image

### Documentation Changes Required
New plugin

PR #38457, opened 2022-08-14
Fixed custom field menuitem warning

roland-d:
Pull Request for Issue #38244

### Summary of Changes
This PR fixes a few issues with the menu item field:
- Cast value to integer as menu item IDs are always numbers
- Check if the menu item exists in case the number is wrong or unpublished
- Use single quotes around strings
- Use HTMLHelper for rendering link

### Testing Instructions
1. Create a custom field of the type `List Of Menu Items (menuitem)`
2. Save the custom field
3. Create an article and select a menu item on the Fields tab
4. Create a menu item of the type `Single Article` and select the new created article
5. Visit the article on the front-end
6. You should see the menu item listed
7. Unpublish the menu item that was selected in the article
8. Refresh the page
9. You see a warning `Attempted to read property title on null`
10. Apply patch
11. Refresh the page
12. The link and warning should be gone

A warning `Attempted to read property title on null` when menu item is not available

No warning is displayed and no link is displayed

None

### DIFF
--- a/plugins/fields/menuitem/menuitem.php
+++ b/plugins/fields/menuitem/menuitem.php
@@ -10,11 +10,13 @@
  * @phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace
  */
 
+use Joomla\Component\Fields\Administrator\Plugin\FieldsPlugin;
+
 /**
  * Fields Menuitem Plugin
  *
  * @since  4.2.0
  */
-class PlgFieldsMenuitem extends \Joomla\Component\Fields\Administrator\Plugin\FieldsPlugin
+class PlgFieldsMenuitem extends FieldsPlugin
 {
 }
--- a/plugins/fields/menuitem/tmpl/menuitem.php
+++ b/plugins/fields/menuitem/tmpl/menuitem.php
@@ -14,13 +14,14 @@ use Joomla\CMS\Router\Route;
 
 defined('_JEXEC') or die;
 
-$value = (int) $field->value;
+$value = (int)$field->value;
+$menu  = Factory::getApplication()->getMenu()->getItem($value);
 
-if ($value == '') {
+if ($value === 0 || !$menu) {
     return;
 }
 
 $url   = Route::_('index.php?Itemid=' . $value);
-$title = Factory::getApplication()->getMenu()->getItem($value)->title;
+$title = $menu->title;
 
 echo HTMLHelper::_('link', $url, $title);
--- a/plugins/fields/menuitem/tmpl/menuitem.php
+++ b/plugins/fields/menuitem/tmpl/menuitem.php
@@ -10,6 +10,7 @@
 
 use Joomla\CMS\Factory;
 use Joomla\CMS\HTML\HTMLHelper;
+use Joomla\CMS\Menu\MenuItem;
 use Joomla\CMS\Router\Route;
 
 defined('_JEXEC') or die;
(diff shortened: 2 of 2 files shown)


chmst:
I have tested this item :white_check_mark: successfully on a066d43e615ef423737a9f1cccc1c627c7207869<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38457">issues.joomla.org/tracker/joomla-cms/38457</a>.</sub>

brianteeman:
For my education can you please explain the changes regarding use statements.

Previously I was told that if its used only once in the file then it should be 
`class PlgFieldsMenuitem extends \Joomla\Component\Fields\Administrator\Plugin\FieldsPlugin
`
and not
`use Joomla\Component\Fields\Administrator\Plugin\FieldsPlugin;

class PlgFieldsMenuitem extends FieldsPlugin
`

But you have changed this in the file

superknutsel:
I have tested this item :white_check_mark: successfully on a066d43e615ef423737a9f1cccc1c627c7207869<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38457">issues.joomla.org/tracker/joomla-cms/38457</a>.</sub>

brianteeman:
Sorry but this is not the correct fix.

The custom field should work the same as the standard fields see https://docs.joomla.org/Menuitem_form_field_type

As you can see I missed adding the published option and just used the default
> published (optional) determines whether all menu items are listed or only published menu items. If state is '0' then all menu items will be listed. If state is '1' then only published menu items will be listed. You also can use comma separated values like '1,2'.

Removing the ability to select an unpublished menu item severely limits the usability of this field ie you can not create links in advance of publishing the menu item

brianteeman:
I have tested this item :red_circle: unsuccessfully on a066d43e615ef423737a9f1cccc1c627c7207869<hr /><sub>This comment was created with the <a href="https://github.com/joomla/jissues">J!Tracker Application</a> at <a href="https://issues.joomla.org/tracker/joomla-cms/38457">issues.joomla.org/tracker/joomla-cms/38457</a>.</sub>

roland-d:
> For my education can you please explain the changes regarding use statements.

For me you always add namespaces to the list of use statements, so you have a complete overview of the namespaces you include.

> Previously I was told that if its used only once in the file then it should be

I am not aware of such rule nor do I understand why it should be that way. The only time you could use the full path is when you have a naming conflict but even then namespaces can be aliased. If it is a rule, then it should be reverted.

> Sorry but this is not the correct fix.

Fine, I am going to revert the custom field and we can apply proper fixes and get it ready for 4.3.

brianteeman:
There are over 50 instances of 
class xxxxxx extends \JoomlaXXXXXXXXXXXXXX

brianteeman:
but if you want to change that and always use namespaces then that is fine by me but then dont you also need to remove
` * @phpcs:disable PSR1.Classes.ClassDeclaration.MissingNamespace`

brianteeman:
The problem must be in the parent groupedlistfield and not here as if you are using the usergroup list field then the exact same testing instructions do not produce the error

brianteeman:
actually I misread part of this and you missed it in my comment

I read it as this pr removed the ability to SELECT an unpublished menu item which is what I said was desired.

> Removing the ability to select an unpublished menu item severely limits the usability of this field ie you can not create links in advance of publishing the menu item

I see now that the check is only for the output of the link

PR #38471, opened 2022-08-16
[5.3] Drone: Adding check for PHP version

Hackwar:
Right now we actually aren't checking the compatibility of our code for a certain PHP version statically. This PR should change that.


Hackwar:
When PRs #38528 and #38527 are merged, the hard errors would be solved in our codebase. The issues left would be several warnings and 3 errors regarding soft reserved keywords which we are using. Those issues couldn't be solved and should be part of the exclusion list.

HLeithner:
@Hackwar what can we do here? clean the error or mark as exception for phpcs?

Hackwar:
Most of these should be fixed. Some would have to be marked as exclusions. Will look at that soon.

Hackwar:
This needs a thorough code review. All changes are outdated code which we never removed.

laoneo:
Either this needs to be split into chunks with propper teating instructions or rebased to 5.0 and merged there early. But I wouldn't merge this in a minor only by code review.

HLeithner:
Please proper error handling using error_get_last() like used in the framework package and rebase on j5.

HLeithner:
@Hackwar can you solve the merge conflicts please

Hackwar:
I've split this PR up into 5 PRs, see above. When those are merged and they are upmerged into this branch, that should leave only the code to introduce the version check in phpcs. I don't think it is usefull to solve the conflicts before those are merged, since each will introduce new conflicts...

composer.json
```diff
@@ -104,6 +104,8 @@
         "phpunit/phpunit": "^8.5",
         "friendsofphp/php-cs-fixer": "^3.0",
         "squizlabs/php_codesniffer": "~3.0",
+        "phpcompatibility/php-compatibility": "*",
+        "phpcompatibility/phpcompatibility-joomla":"*",
```
jrfnl:
```suggestion
        "phpcompatibility/phpcompatibility-joomla":"^2",
```

No need to `require-dev` PHPCompatibility, better to let PHPCompatibilityJoomla manage that as its that package' dependency.

Also, please be aware that PHPCompatibilityJoomla 2.x is based on PHPCompatibility 9.x, PHPCompatibilityJoomla 3.x will be based on PHPCompatibility 10.x (expected soonish). As that's a major, it will bring breaking changes (and some new dependencies), so you may want to do a controlled upgrade when the time comes.

ruleset.xml
```diff
@@ -21,6 +21,32 @@
...
+    <rule ref="PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue">
+        <exclude-pattern type="relative">libraries/src/Date/Date\.php</exclude-pattern>
+        <exclude-pattern type="relative">libraries/src/Application/CliApplication\.php</exclude-pattern>
+        <exclude-pattern type="relative">libraries/src/Application/ConsoleApplication\.php</exclude-pattern>
+        <exclude-pattern type="relative">libraries/src/Document/HtmlDocument\.php</exclude-pattern>
+    </rule>
+
+    <rule ref="PHPCompatibility.FunctionNameRestrictions.NewMagicMethods">
+        <exclude-pattern type="relative">administrator/components/com_finder/src/Indexer/Result\.php</exclude-pattern>
+    </rule>
```
jrfnl, line 47: Please be aware that these excludes will exclude the issues for the complete file, which may hide new issues being introduced in the future.

If you want more selective "ignores", you can use the [PHPCS native inline ignore annotations](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-parts-of-a-file) in the files themselves, typically ignoring only one particular line or a small group of lines.
This will more effectively prevent people from accidentally introducing new issues.

ruleset.xml
```diff
@@ -21,6 +21,32 @@
```
jrfnl, line 20: ☝🏻 As a general remark - I see top-level `exclude-patterns` in this config file, but no `<file>` _include_ ? That also may make the `type="relative"` directives unstable as it is unclear what this should be _relative to_...

ruleset.xml
```diff
@@ -21,6 +21,32 @@
 
     <rule ref="PSR12" />
 
+    <rule ref="PHPCompatibilityJoomla">
+        <exclude-pattern type="relative">libraries/src/Encrypt/AES/Mcrypt\.php</exclude-pattern>
+    </rule>
```
jrfnl, line 25:
```suggestion
    <rule ref="PHPCompatibilityJoomla">
        <include-pattern>*\.php$</include-pattern>
        <exclude-pattern type="relative">libraries/src/Encrypt/AES/Mcrypt\.php</exclude-pattern>
    </rule>
```

PHPCS will scan PHP, JS and CSS files by default. As there is no top-level [`extensions` setting](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#specifying-valid-file-extensions) included in the ruleset to change this, you may want to ensure that the PHPCompatibility sniffs are only run against PHP files.

jrfnl: Also: as the `Mcrypt` file is excluded completely - is that a polyfill for the Mcrypt extension ? And if so, should the PHPCompatibilityJoomla ruleset be updated to exclude errors and warnings from the polyfilled functionality completely ?

jrfnl (commented):
Left some inline notes at the request of @Hackwar. Hope it helps.

ruleset.xml
Hackwar: The Mcrypt file is a wrapper and basically currently broken in 4.x. We are working on removing it...

libraries/src/Language/LanguageHelper.php
```diff
@@ -425,12 +416,7 @@ public static function parseIniFile($fileName, $debug = false)
             $strings = @parse_ini_file($fileName);
         }
 
-        // Restore error tracking to what it was before.
-        if ($debug === true) {
-            ini_set('track_errors', $trackErrors);
-        }
-
-        return \is_array($strings) ? $strings : [];
+        return \is_array($strings) ? $strings : array();
```
HLeithner: I'm surprised about this change, why do we use `array()` now instead of `[]`

Hackwar: most likely a wrong merge...

HLeithner:

brianteeman:
Is anything happening with this? It has not been touched for over a year and it is unclear if anything will ever happen. It refers to it being partialy split into 5 pr but some of those have been closed as abandoned.

HLeithner:

PR #38523, opened 2022-08-18
[4.2] Remove hotkeys.js by script.php

fancyFranci:
Pull Request for Issue # .

### Summary of Changes
While the development of 4.2 hotkeys.js was renamed to hotkeysjs to avoid issues. Currently both ~~are in the build~~ may be on sites, but the one with the dot should be removed with all its files and folders.

@richard67 please have a look

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -6438,7 +6438,7 @@ class JoomlaInstallerScript
             // From 4.2.0 to 4.2.1
             '/media/vendor/hotkeys.js/js/hotkeys.js',
             '/media/vendor/hotkeys.js/js/hotkeys.min.js',
-            '/media/vendor/hotkeys.js/js/hotkeys.min.js.gz'
+            '/media/vendor/hotkeys.js/js/hotkeys.min.js.gz',
             '/media/vendor/hotkeys.js/LICENSE',
         );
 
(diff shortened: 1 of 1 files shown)


richard67:
> Currently both are in the build

@fancyFranci That should not be the case. But the old folder and files may still be there on sites updated from pervious versions, and that's what will be deleted on update with your change.

administrator/components/com_admin/script.php
```diff
@@ -6435,6 +6435,10 @@ public function deleteUnexistingFiles($dryRun = false, $suppressOutput = false)
             '/plugins/system/webauthn/webauthn.php',
             '/plugins/task/checkfiles/checkfiles.php',
             '/plugins/task/demotasks/demotasks.php',
+            // From 4.2.0 to 4.2.1
+            '/media/vendor/hotkeys.js/js/hotkeys.js',
+            '/media/vendor/hotkeys.js/js/hotkeys.min.js',
```
richard67, line 6440:
```suggestion
            '/media/vendor/hotkeys.js/js/hotkeys.min.js',
            '/media/vendor/hotkeys.js/js/hotkeys.min.js.gz'
```
The pre-zipped file should be removed, too.

That's why I always base my deleted files and folders checks on comparing the full installation zip packages of the 3 releases, and that's the only it can be done, because in the build/tmp folder, the build.php script deletes files from the build between building the full installation packages and the update packages, so checking the folder with the build after that will not cover all files.

administrator/components/com_admin/script.php
```diff
@@ -7804,6 +7808,8 @@ public function deleteUnexistingFiles($dryRun = false, $suppressOutput = false)
             '/plugins/system/webauthn/src/Exception',
             // From 4.2.0 to 4.2.1
             '/libraries/vendor/symfony/string/Resources/bin',
+            '/media/vendor/hotkeys.js/js',
```
richard67, line 7811: In the list of folders we use reverse (i.e. descending) alphabetical order because that makes sure that we delete subfolders before their parent. That means that you should move your new 2 lines to before the '/libraries/vendor/symfony/string/Resources/bin' line, because "m" comes before "l" when using **reverse** ordering.

This applies only to the folders list for the mentioned reason. The files list is as usual sorted in alphabetical order ascending, not reverse (descending).

richard67 (approved):

